Issue #2109, opened 2023-04-13
Cloud UI burndown


ericpgreen2:
Any more feedback? Feel free to comment in [this Notion page](https://www.notion.so/rilldata/Cloud-UI-feedback-70b060531caa43d4b949260a41fdb114?pvs=4).

[Bug bash](https://www.notion.so/rilldata/Rill-cloud-Bug-bash-2413071d9b734a9285ba677369874b69).

## P1
- [ ] Rill Developer: Controls to resize a model's preview table show up on the dashboard
- [ ] Login/Signup: Click the back button after logging in through the UI, and you get to an Auth0 error page (@AdityaHegde, @djbarnwal)
- [ ] Home: Clicking on the home page in UI when on home page doesn’t refresh the page (@AdityaHegde)
- [ ] Dashboard: Fixing an errored dashboard and clicking on the link from project page shows “something went wrong page” briefly before redirecting to the dashboard page (@AdityaHegde)
- [ ] Dashboard: Missing validation while restoring from url state (@AdityaHegde)
- [ ] Dashboard: Figure out Anshul's 404 from the bug bash (@ericpgreen2, pending steps to reproduce)
- [ ] Errors: Add telemetry for errors
- [ ] https://github.com/rilldata/rill-developer/issues/2150 (@AdityaHegde)
- [ ] https://github.com/rilldata/rill-developer/issues/2149 (@AdityaHegde)

## P2
- [ ] Rill Developer: Add "Local files will be duplicated into your project’s `/data` directory." to the local file upload modal
- [ ] Meta: Make fonts smoother / antialiased to match the mocks (@djbarnwal) (#2359)
- [ ] Meta: Avoid fetching a Project’s Organization just for the name, given https://github.com/rilldata/rill-developer/pull/2085
- [ ] Meta: Use Tailwind grays from [extended color palette](https://tailwindcss.com/docs/upgrade-guide#renamed-gray-scales) 
- [ ] Meta: `CTALayoutContainer.svelte` should use responsive margin, not fractional height
- [ ] Login/Signup: We should have two different routes `/login` and `/signup` (@djbarnwal)
- [ ] Top bar: Add text ellipses when names are long (@djbarnwal)
- [ ] Home: The “if you” Sharing CTA language should instead look at the `GetOrganization.manageMembers` permission
- [ ] Home: When logging out on the home page, a 403 Forbidden flashes before redirecting to the login screen (@ericpgreen2)
- [ ] Home: Add collapsible carets
[...]

PR #2178, opened 2023-04-20
Add Github connection section to project page

ericpgreen2:
Contributes to #2109


web-admin/src/routes/[organization]/[project]/-/deployment/+page.svelte
```diff
@@ -13,7 +16,7 @@
 <div class="flex flex-col items-center">
   <div class="flex space-x-10 border-b border-gray-200 w-full px-12 py-5">
```
Comment (hamilton) [inline web-admin/src/routes/[organization]/[project]/-/deployment/+page.svelte:17]: Would suggest you use grid here, not flex, and explicitly set these column widths according to the mocks. It'll be much more usable and easier to tweak

web-admin/src/components/projects/ProjectGithubConnection.svelte
```diff
@@ -1,3 +1,54 @@
...
+          <a
+            href={$proj.data?.project?.githubUrl}
+            class="flex items-center gap-x-1 text-gray-800"
+            target="_blank"
+            rel="noreferrer"
+          >
+            <Github className="inline-block w-4 h-4" />
+            <span
+              class="font-semibold text-[12px] leading-5"
+              style="font-family: Source Code Variable">{prettyGithubRepo}</span
```
hamilton, line 36: you can use `var(--monospace)` here I believe, in case we change it later

hamilton (approved):
Overall looks good!

We will definitely need to enforce stronger constraints on column sizes, but this PR can merge first & we can address the remaining layout changes in another PR.

I left a couple of small comments. Feel free to merge after addressing

web-admin/src/routes/[organization]/[project]/-/deployment/+page.svelte
Comment (ericpgreen2) [reply web-admin/src/routes/[organization]/[project]/-/deployment/+page.svelte:17]: Good call. I'll change the layout in a different PR, but just added a max-width & truncate here, so there's no excessive overflow.

PR #2180, opened 2023-04-20
Generate tree-structured CLI documentation for Docusaurus

egor-ryashin:
image


docs/docs/references/CLI/rill/docs/docs.md
```diff
@@ -0,0 +1,16 @@
+## rill docs
+
+Manage documentation
```
nishantmonu51, line 3: current behavior for `rill docs` is that it opens the documentation.

docs/docs/references/CLI/rill/docs/docs.md
```diff
@@ -0,0 +1,16 @@
...
+
+```
+  -h, --help   Print usage
+```
+
+### SEE ALSO
+
+* [rill](../rill.md)	 - Rill CLI
+* [rill docs generate](generate.md)	 - Generate CLI documentation
+* [rill docs open](open.md)	 - Open docs.rilldata.com
```
nishantmonu51, line 15: @begelundmuller : should we hide docs generate and open from referencing here ?

docs/docs/references/CLI/rill/project/project.md
```diff
@@ -0,0 +1,30 @@
...
+
+### Options
+
+```
+      --api-token string   Token for authenticating with the admin API
+      --api-url string     Base URL for the admin API (default "https://admin.rilldata.io")
+      --org string         Organization Name
+```
+
+### Options inherited from parent commands
```
nishantmonu51, line 13: I wonder if we should merge options inherited and normal options under same heading ? 
Also should we call them parameters instead of options? 
+cc @begelundmuller

nishantmonu51 (commented):
LGTM, 👍

docs/docs/references/CLI/rill/docs/docs.md
begelundmuller: Yeah – and also the `rill runtime` and `rill admin` commands (used internally to start those services). In general, all the commands where `Hidden: !cfg.IsDev()` is set should not be generated in the docs.

docs/docs/references/CLI/rill/project/project.md
begelundmuller: In the Cobra `--help` menus, it renders as:
```
Flags:
      --api-token string   Token for authenticating with the admin API
      --api-url string     Base URL for the admin API (default "http://localhost:9090")

Global Flags:
  -h, --help   Print usage
```
Maybe we should mirror that here?

At least, we should use the word "flag" instead of "options".

I think it's helpful to keep the global flags separate from the individual flags, since the flag ordering might be weird otherwise (we currently order flags by expected frequency of use, not alphabetically)

nishantmonu51: sounds good, lets mirror the --help menus here.

PR #2182, opened 2023-04-20
2169 cli burndown list link

rakeshsharma14317:
## Org Management

- [x]  `rill org create --name <org_name>`  @Rakesh Sharma
    - [x]  creates org with specified name, prompt for name if not already provided
    - [x]  Graceful failure when org already exist instead of database: violates unique constraint
    - [ ]  Limit handling ?? More than X orgs?
    
- [x]  `rill org list`
    - [x]  Add an indication to show default org @Anshul Khandelwal
- [x]  `rill org show` - rill list already shows this info remove this command @Rakesh Sharma
- [x]  `rill org close` remove that command, `rill org delete` is already present  @Rakesh Sharma
- [x]  `rill org edit --org <org_name> --description <new_description>` @Rakesh Sharma
    - [x]  Change org_name to flag
    - [x]  Handle case when org doesn’t exist, Suggest running `rill org list` to list available orgs
- [x]  `rill org delete` @Rakesh Sharma
    - [x]  Add warn, deleting an org will remove all metadata associated with the org
    - [x]  Add a `--force` flag which skips the confirmation
- [x]  `rill org rename --org name --new_name new`
    - [x]  Add a warning stating that renaming an org would invalidate dashboard URLs.
    - [x]  Add better error handling when the renamed org already exist, Current Error: rpc error: code = Internal desc = database: violates unique constraint
- [x]  `rill org members` list all members of an org
- [x]  `rill org switch` - sets a default org

- [x]  `rill project status --project <proj_name> --org <org_name>`  @Rakesh Sharma
    - [ ]  Also print URL for project UI
- [x]  `rill project delete --project <proj_name> --org <org_name>` @Rakesh Sharma
    - [x]  Change project to flag
    - [x]  Add confirmation and make user enter project name
    - [x]  Add a line about implications of deleting a project
    - [x]  Add a `--force` flag which skips the confirmation
- [x]  `rill project rename --project <old_name> --new_name <new_name>` @Rakesh Sharma
    - [x]  interactive mode is working fine, just change to flags
    - [x]  Add a warning that this will change dashboard URLs
- [x]  `rill project edit --project <project_name> --description --prod-branch --public --repo_path <path_for_monorepo>`   @Rakesh Sharma
- [ ]  `rill project reconcile`
    - [ ]  reconciles a project by default
    - [ ]  Flag `—refresh` to also refresh sources
    - [ ]  Flag `--source <name>`  to refresh a selected source
    - [ ]  Flag `--reset` to redeploy the project
- [x]  `rill project members` - List all members of a project
- [x]  Infer `--project` from git remote repo name for project level commands @k-anshul 
    - [x]  Do a lookup based on `git remote` and find project name
    - [x]  No inference, if there are more than one projects with same remote URLs - Add to docs of the commands.


cli/cmd/project/delete.go
```diff
@@ -22,17 +25,48 @@ func DeleteCmd(cfg *config.Config) *cobra.Command {
 			}
 			defer client.Close()
 
+			// Get the new project name from user if not provided in the flag
+			if !cmd.Flags().Changed("project") {
```
k-anshul, line 29: This will not prompt for a name if empty name is given by the user. Is that expected ?

rakeshsharma14317: @begelundmuller Your thoughts on this? We can check for `==""` for projects but not for `orgs` as we set the default org for projects.

Also we have checks for `value length` for empty in API validate. So any suggestions on this?

cli/cmd/org/create.go
```diff
@@ -11,21 +12,40 @@ import (
...
 
+			if !cmd.Flags().Changed("name") {
+				// Get the new org name from user if not provided in the flag
+				err := cmdutil.PromptIfUnset(&name, "Org Name", "")
+				if err != nil {
+					return err
+				}
+			}
+
+			exist, err := cmdutil.OrgExists(ctx, client, name)
```
k-anshul, line 38: We can avoid this RPC call and can directly call `CreateOrganization` which will fail if org name already exists.

cli/cmd/org/rename.go
```diff
@@ -50,34 +43,22 @@ func RenameCmd(cfg *config.Config) *cobra.Command {
...
 					return err
 				}
 			}
 
+			fmt.Println("Warn: Renaming an org would invalidate dashboard URLs")
+
 			confirm := false
 			prompt := &survey.Confirm{
-				Message: fmt.Sprintf("Do you want to rename org \"%s\" to \"%s\"?", color.YellowString(currentName), color.YellowString(newName)),
+				Message: fmt.Sprintf("Do you want to rename org \"%s\" to \"%s\"?", color.YellowString(name), color.YellowString(newName)),
```
k-anshul, line 61: Can use `cmdutil.ConfirmPrompt` here ?

cli/cmd/org/rename.go
```diff
@@ -114,6 +95,9 @@ func RenameCmd(cfg *config.Config) *cobra.Command {
 			return nil
 		},
 	}
+	renameCmd.Flags().SortFlags = false
+	renameCmd.Flags().StringVar(&name, "org", cfg.Org, "Current Org Name")
```
k-anshul, line 99: Setting defaults here is not required since we are always taking input from the user if flag is not set.

k-anshul (changes requested):

cli/cmd/project/edit.go
```diff
@@ -24,20 +28,54 @@ func EditCmd(cfg *config.Config) *cobra.Command {
...
+			}
+
+			resp, err := client.GetProject(ctx, &adminv1.GetProjectRequest{OrganizationName: cfg.Org, Name: name})
 			if err != nil {
 				return err
 			}
 
 			proj := resp.Project
 
+			if !cmd.Flags().Changed("description") {
```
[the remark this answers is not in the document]
k-anshul: @nishantmonu51 What do you think ?

cli/cmd/project/rename.go
```diff
@@ -49,34 +42,22 @@ func RenameCmd(cfg *config.Config) *cobra.Command {
...
-							name := any.(string)
-							if name == "" {
-								return fmt.Errorf("empty name")
-							}
-
-							return nil
-						},
-					},
-				}
+				name = cmdutil.SelectPrompt("Select project to rename", projectNames, "")
```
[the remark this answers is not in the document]
rakeshsharma14317: yeah lets make this same at every places, @nishantmonu51 what we should prefer if org/project not provided we should taking text input or show drop-down?

cli/cmd/org/edit.go
```diff
@@ -23,7 +25,32 @@ func EditCmd(cfg *config.Config) *cobra.Command {
 			}
 			defer client.Close()
 
-			resp, err := client.GetOrganization(ctx, &adminv1.GetOrganizationRequest{Name: args[0]})
+			if !cmd.Flags().Changed("org") {
+				// Get the new org name from user if not provided in the flag
+				err := cmdutil.PromptIfUnset(&orgName, "Org Name", orgName)
```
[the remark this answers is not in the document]
rakeshsharma14317: As in flag it will be set to default org so it won't be empty, it will always be default if not provided

cli/cmd/org/edit.go
```diff
@@ -45,7 +72,8 @@ func EditCmd(cfg *config.Config) *cobra.Command {
 		},
 	}
 	editCmd.Flags().SortFlags = false
-	editCmd.Flags().StringVar(&description, "description", "", "Description")
+	editCmd.Flags().StringVar(&orgName, "org", cfg.Org, "Organization name")
```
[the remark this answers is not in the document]
rakeshsharma14317: keeping a default value as it will show up in the prompt if user has not explicitly provided 
and user can just enter for default one

cli/cmd/org/rename.go
```diff
@@ -50,34 +43,22 @@ func RenameCmd(cfg *config.Config) *cobra.Command {
...
-							name := any.(string)
-							if name == "" {
-								return fmt.Errorf("empty name")
-							}
-
-							return nil
-						},
-					},
-				}
+				name = cmdutil.SelectPrompt("Select org to rename", orgNames, "")
```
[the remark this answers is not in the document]
rakeshsharma14317: Yeah, we need to decide on either of them, @nishantmonu51 @begelundmuller can suggest on this

cli/cmd/org/rename.go
rakeshsharma14317: keeping a default value as it will show up in the prompt if user has not explicitly provided
and user can just enter for default one

cli/cmd/project/edit.go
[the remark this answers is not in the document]
rakeshsharma14317: Yeah we can take a decision here, Also i am putting older values as default in prompt, so if user can see that value and decide if needs to change or just press enter for default one.

cli/cmd/org/edit.go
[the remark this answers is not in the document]
k-anshul: May be we should just put this default only in the prompt config? Its slightly counter-intuitive to see a default here.

cli/cmd/org/rename.go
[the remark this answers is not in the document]
nishantmonu51: For objects we already know the options, I would prefer using select prompt, e.g. selecting from existing projects or orgs. 
For objects where we don't know the inputs to an input prompt, e.g. email id of user to invite.

cli/cmd/project/status.go
```diff
@@ -25,9 +27,16 @@ func StatusCmd(cfg *config.Config) *cobra.Command {
...
 			defer client.Close()
 
+			if !cmd.Flags().Changed("project") {
+				name, err = inferProjectName(cmd.Context(), client, cfg.Org, path)
+				if err != nil {
+					return err
+				}
+			}
+
 			proj, err := client.GetProject(context.Background(), &adminv1.GetProjectRequest{
```
k-anshul, line 37: use cmd.Context() here

cli/cmd/org/create.go
```diff
@@ -11,25 +12,37 @@ import (
...
+				Name:        name,
 				Description: description,
 			})
 			if err != nil {
-				return err
+				if !cmdutil.IsNameExistsErr(err) {
+					return err
+				}
+
+				return fmt.Errorf("Org name %q already exists", name)
```
k-anshul, line 45: We can just print this msg and return nil.

cli/cmd/org/edit.go
```diff
@@ -23,9 +27,32 @@ func EditCmd(cfg *config.Config) *cobra.Command {
 			}
 			defer client.Close()
 
-			resp, err := client.GetOrganization(ctx, &adminv1.GetOrganizationRequest{Name: args[0]})
+			if !cmd.Flags().Changed("org") {
```
k-anshul, line 30: We agreed to use default org in cases when `--org` is not passed. Also applied to `rill org edit`.
Also if you decide to prompt in these cases, please pass the default org as default in SelectPrompt.

cli/cmd/org/edit.go
```diff
@@ -23,9 +27,32 @@ func EditCmd(cfg *config.Config) *cobra.Command {
...
+			resp, err := client.GetOrganization(ctx, &adminv1.GetOrganizationRequest{Name: orgName})
 			if err != nil {
-				return err
+				if st, ok := status.FromError(err); ok {
+					if st.Code() != codes.NotFound {
+						return err
+					}
+				}
+
+				return fmt.Errorf("Org name %q not exists, please run `rill org list` to list available orgs", orgName)
```
k-anshul, line 55: Nit: `Org name %q doesn't exist...`

cli/cmd/project/delete.go
```diff
@@ -22,17 +25,44 @@ func DeleteCmd(cfg *config.Config) *cobra.Command {
...
-			cmdutil.SuccessPrinter(fmt.Sprintf("Deleted project: %v\n", args[0]))
+			cmdutil.SuccessPrinter(fmt.Sprintf("Deleted project: %v\n", name))
 			return nil
 		},
 	}
+
+	deleteCmd.Flags().SortFlags = false
+	deleteCmd.Flags().BoolVar(&force, "force", false, "Delete forcefully, skips the confirmation")
+	deleteCmd.Flags().StringVar(&name, "project", "", "Name")
+	deleteCmd.Flags().StringVar(&path, "path", ".", "Project directory")
```
k-anshul, line 65: The default `org` should be `cfg.Org` but there should be a flag to override that. Applied to other project sub commands as well.

k-anshul (changes requested):

cli/cmd/org/edit.go
```diff
@@ -23,9 +27,32 @@ func EditCmd(cfg *config.Config) *cobra.Command {
...
+			resp, err := client.GetOrganization(ctx, &adminv1.GetOrganizationRequest{Name: orgName})
 			if err != nil {
-				return err
+				if st, ok := status.FromError(err); ok {
+					if st.Code() != codes.NotFound {
+						return err
+					}
+				}
+
+				return fmt.Errorf("Org name %q doesn't exist, please run `rill org list` to list available orgs", orgName)
```
k-anshul, line 55: This can also be changed to a msg on CLI and can return `nil`.

cli/cmd/org/edit.go
```diff
@@ -45,7 +72,8 @@ func EditCmd(cfg *config.Config) *cobra.Command {
 		},
 	}
 	editCmd.Flags().SortFlags = false
-	editCmd.Flags().StringVar(&description, "description", "", "Description")
+	editCmd.Flags().StringVar(&orgName, "org", cfg.Org, "Organization name")
+	editCmd.Flags().StringVar(&description, "description", "Unknown", "Description")
```
k-anshul, line 76: Default description should be empty string

cli/cmd/org/rename.go
```diff
@@ -114,6 +76,9 @@ func RenameCmd(cfg *config.Config) *cobra.Command {
 			return nil
 		},
 	}
+	renameCmd.Flags().SortFlags = false
+	renameCmd.Flags().StringVar(&name, "org", cfg.Org, "Current Org Name")
+	renameCmd.Flags().StringVar(&newName, "new_name", cfg.Org, "New Org Name")
```
k-anshul, line 81: Default value of new_name can't be default org name

k-anshul (approved):

PR #2185, opened 2023-04-20
fixes console error

bcolloran:
This fixes a console error that is showing up all the time and driving me crazy while debugging other code. There is a subtle bug here where the `container` item may exist when the the `setTimeout` is set, but it may not exist by the time the callback is triggered. This adds a guard _within_ the callback to make sure that the `container` elt exists at the time the callback is fired (just added the `if (container)`, the rest is indentation change)

### DIFF
--- a/web-common/src/components/data-graphic/marks/MultiMetricMouseoverLabel.svelte
+++ b/web-common/src/components/data-graphic/marks/MultiMetricMouseoverLabel.svelte
@@ -134,17 +134,19 @@ It is probably not the most up to date code; but it works very well in practice.
   $: if (container && locations && $xScale && $yScale) {
     clearTimeout(transitionalTimeoutForCalculatingLabelWidth);
     transitionalTimeoutForCalculatingLabelWidth = setTimeout(() => {
-      labelWidth = Math.max(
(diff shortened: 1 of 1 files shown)


bcolloran:
This should be a trivial review. Assigning @ericpgreen2 but anyone can take it.

web-common/src/components/data-graphic/marks/MultiMetricMouseoverLabel.svelte
```diff
@@ -134,17 +134,19 @@ It is probably not the most up to date code; but it works very well in practice.
   $: if (container && locations && $xScale && $yScale) {
     clearTimeout(transitionalTimeoutForCalculatingLabelWidth);
     transitionalTimeoutForCalculatingLabelWidth = setTimeout(() => {
-      labelWidth = Math.max(
-        ...Array.from(container.querySelectorAll(".widths")).map(
-          (q: SVGElement) => q.getBoundingClientRect().width
-        )
-      );
+      if (container) {
```
djbarnwal, line 137: Quickly went through this. We are already checking if `container` is valid on line 134 right? Adding another check in the same block looks counter-intutive.

djbarnwal (approved):

PR #2194, opened 2023-04-21
Utilize any parameter of DuckDB CSV/JSON/Parquet "read" function

esevastyanov:
Pass all properties (located in "duckdb" section) of a source config to corresponding DuckDB functions on ingestion
CSV ingestion props are under `duckdb -> csv`
Parquet ingestion props are under `duckdb -> parquet`
JSON ingestion props are under `duckdb -> json`
Decided to split props per format as we may want to pass some more properties like extensions `duckdb -> extension`


runtime/services/catalog/artifacts/yaml/objects.go
```diff
@@ -145,9 +143,6 @@ func fromSourceArtifact(source *Source, path string) (*drivers.CatalogEntry, err
 	if source.Region != "" {
 		props["region"] = source.Region
 	}
-	if source.CsvDelimiter != "" {
-		props["csv.delimiter"] = source.CsvDelimiter
-	}
```
begelundmuller, line 150: Can we keep reading `csv.delimiter` from the source artifact, and writing it to `duckdb.delim` here instead – for backwards compatibility

begelundmuller (changes requested):

runtime/services/catalog/artifacts/artifacts_test.go
```diff
@@ -137,23 +137,23 @@ timeseries: time
...
-  description: Dim0_D
-- label: Dim1_L
-  property: dim1
-  description: Dim1_D
+    - label: Dim0_L
+      property: dim0
+      description: Dim0_D
+    - label: Dim1_L
+      property: dim1
+      description: Dim1_D
```
esevastyanov, line 145: yaml.v3 doesn't support setting indentation for sequences
https://github.com/go-yaml/yaml/issues/661

runtime/services/catalog/artifacts/yaml/yaml.go
```diff
@@ -8,7 +8,7 @@ import (
 
 	"github.com/rilldata/rill/runtime/drivers"
 	"github.com/rilldata/rill/runtime/services/catalog/artifacts"
-	"gopkg.in/yaml.v2"
+	"gopkg.in/yaml.v3"
```
esevastyanov, line 11: yaml.v2 deserializes nested maps as `map[interface{}]interface{}` instead of `map[string]interface{}`
https://github.com/go-yaml/yaml/issues/139

runtime/drivers/duckdb/connectors.go
```diff
@@ -201,3 +150,111 @@ func resolveLocalPath(env *connectors.Env, path, sourceName string) (string, err
...
+	for _, target := range targets {
+		if strings.Contains(source, target) {
+			return true
+		}
+	}
+	return false
+}
+
+func generateReadCsvStatement(paths []string, properties map[string]interface{}) (string, error) {
+	ingestionProps := collectDuckDBIngestionProperties("duckdb.csv.", properties)
```
begelundmuller, line 187: Instead of `duckdb.csv`, should we just put it under `duckdb`? I guess the user will already know from context what kind of format they're ingesting?

runtime/drivers/duckdb/connectors.go
```diff
@@ -201,3 +150,111 @@ func resolveLocalPath(env *connectors.Env, path, sourceName string) (string, err
...
+	// backward compatibility: csv.delimiter might be passed separately from duckdb.csv.delim and has a priority
+	if csvDelimiter, csvDelimiterDefined := properties["csv.delimiter"]; csvDelimiterDefined {
+		ingestionProps["delim"] = fmt.Sprintf("'%v'", csvDelimiter)
+	}
```
begelundmuller, line 191: I would put these backwards compatibility rules in `runtime/services/catalog/artifacts/yaml/objects.go` instead. So right at artifact parse time, `csv.delimiter` gets mapped to `duckdb.delim`.

runtime/drivers/duckdb/connectors.go
```diff
@@ -201,3 +150,111 @@ func resolveLocalPath(env *connectors.Env, path, sourceName string) (string, err
...
+func generateReadParquetStatement(paths []string, properties map[string]interface{}) (string, error) {
+	ingestionProps := collectDuckDBIngestionProperties("duckdb.parquet.", properties)
+	// set hive_partitioning to true by default
+	if _, hivePartitioningDefined := ingestionProps["hive_partitioning"]; !hivePartitioningDefined {
+		ingestionProps["hive_partitioning"] = true
+	}
+	// backward compatibility: hive_partitioning might be passed separately from duckdb.parquet.hive_partitioning
+	if hivePartitioning, hpDefinedSeparately := properties["hive_partitioning"]; hpDefinedSeparately {
+		ingestionProps["hive_partitioning"] = hivePartitioning
+	}
```
begelundmuller, line 205: Comments above also apply here

runtime/drivers/duckdb/connectors.go
```diff
@@ -201,3 +150,111 @@ func resolveLocalPath(env *connectors.Env, path, sourceName string) (string, err
...
+	return fmt.Sprintf("read_parquet(%s)", convertToStatementParamsStr(paths, ingestionProps)), nil
+}
+
+func generateReadJSONStatement(paths []string, properties map[string]interface{}) (string, error) {
+	ingestionProps := collectDuckDBIngestionProperties("duckdb.json.", properties)
```
begelundmuller, line 210: Comment above also applies here

begelundmuller (changes requested):

runtime/drivers/duckdb/connectors.go
esevastyanov: switched back to `duckdb`

runtime/drivers/duckdb/connectors.go
esevastyanov: moved to `objects.go`

runtime/drivers/duckdb/connectors.go
esevastyanov: also moved to `objects.go`

runtime/drivers/duckdb/connectors.go
esevastyanov: done

begelundmuller (approved):
Looks great!

PR #2197, opened 2023-04-21
Fix 2075 removable list menu closing when clicked

bcolloran:


bcolloran:
@hamilton -- I've spent hours and hours on #2075, and I'm stumped. Either I'm much misunderstanding something fundamental about Svelte, or we've hit some weird edge case.

The bug itself is fixed I think -- it was just a matter of removing a listener for `focusin` in the menu component -- but there is now a much weirder bug in the menu itself: it seems that when the `excludeMode` prop is updated in the RemovableListChip component, it's not being passed down to the RemovableListMenu component, so the menu content doesn't update. I've googled everything I can think of, but I'm at my wit's end, so hopefully you have an idea of what to try next.

Perhaps we should spend a minute pairing on this on Tuesday.

bcolloran:
closing in favor of #2227

PR #2201, opened 2023-04-24
Show line chart when no prev data

djbarnwal:
Prev - 

image

Now -

image


ericpgreen2 (approved):
I'll approve since this is definitely better than before. However, we probably want to revisit the line chart to better convey there's no data in the prior period.

PR #2204, opened 2023-04-24
RBAC fixes: Public projects and full project permissions for org admins

begelundmuller:
- Enables anonymous access to projects marked `public`
- Gives org admins (people with `org_roles.manage_projects = true`) full project-level permissions on all projects in the org
- Lets only org admins list all projects in an org – others just get to see those they are a member of

### DIFF
--- a/admin/server/projects.go
+++ b/admin/server/projects.go
@@ -27,48 +27,60 @@ func (s *Server) ListProjectsForOrganization(ctx context.Context, req *adminv1.L
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 
-	// add public projects
-	orgProjects := map[string]*database.Project{}
-	publicProjects, err := s.admin.DB.FindPublicProjectsInOrganization(ctx, org.ID)
-	if err != nil {
-		return nil, status.Error(codes.Internal, err.Error())
-	}
-	for _, proj := range publicProjects {
-		orgProjects[proj.Name] = proj
-	}
-
+	// If user has ManageProjects, return all projects
 	claims := auth.GetClaims(ctx)
-	if !claims.OrganizationPermissions(ctx, org.ID).ReadProjects {
-		// check if the user is an outside member of a project in the org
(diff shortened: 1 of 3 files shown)


admin/server/projects.go
```diff
@@ -124,14 +136,20 @@ func (s *Server) GetProject(ctx context.Context, req *adminv1.GetProjectRequest)
...
+		permissions.ReadProject = true
+		permissions.ReadProd = true
+	}
+
+	if !permissions.ReadProject {
 		return nil, status.Error(codes.PermissionDenied, "does not have permission to read project")
 	}
 
-	if proj.ProdDeploymentID == nil {
+	if proj.ProdDeploymentID == nil || !permissions.ReadProd {
```
pjain1, line 149: should this be `permissions.ReadProd` ? whats the meaning of `!permissions.ReadProd` ?

admin/server/projects.go
```diff
@@ -317,7 +333,7 @@ func (s *Server) ListProjectMembers(ctx context.Context, req *adminv1.ListProjec
 	}
 
 	claims := auth.GetClaims(ctx)
-	if !claims.Can(ctx, proj.OrganizationID, auth.ReadOrgMembers, proj.ID, auth.ReadProjectMembers) {
+	if !claims.ProjectPermissions(ctx, proj.OrganizationID, proj.ID).ReadProjectMembers {
```
pjain1, line 336: now an org viewer will not be able to list project members, is that correct ?

admin/server/projects.go
begelundmuller: If this `if` clause is triggered, it will only return the project info, not the prod deployment info and JWT. So if user only has `ReadProject`, but not `ReadProd` (kind of weird, but possible with current permissions), they will get the project info, but not the prod deployment info.

admin/server/projects.go
pjain1: makes sense, however we add project viewer role to the all_user_group of the org upon project creation.

PR #2205, opened 2023-04-24
Fix editor gutter and fontSize

### LINKED ISSUE #2074: the editor CSS appears to be broken
Looks like the styles in the sql editor aren't getting applied.

Screen Shot 2023-04-11 at 6 44 37 AM

djbarnwal:
Fixes  #2074 

Don't see any commits which change the existing Editor CSS. Also codemirror packages haven't been updated recently. Not sure what was going wrong. Have added new styles which fixes the gutter background. Also increased font size to 13px.


ericpgreen2 (approved):
Looks good!

PR #2206, opened 2023-04-24
Handle dashboard yaml changes outside of our app

AdityaHegde:
Any changes done to dashboard.yaml when the editor is open will overwrite it if further changes are made. This is because we do not have a periodic fetch the yaml file.

Adding `refetchOnWindowFocus` to make it refetch when user came back to the app. Periodic refetch might add some other inconstancies when the user is editing the dashboard on the app.


ericpgreen2 (approved):
Nice, I like this solution for now.

PR #2220, opened 2023-04-25
fix for expired and malformed auth cookie

### LINKED ISSUE #2163: Admin: Fix error `failed to get session: securecookie: expired timestamp`
We should generally not fail if cookies have expired or are malformed (perhaps with some exceptions). Refactor to clear cookies that are expired, and then continue as if the cookie was empty.

k-anshul:
closes #2163


cli/cmd/auth/login.go
```diff
@@ -64,7 +64,7 @@ func LoginCmd(cfg *config.Config) *cobra.Command {
 
 				fmt.Printf("Set default organization to %q. Change using `rill org switch`.\n", defaultOrg)
 			} else {
-				warn.Println("You are not part of any org. Run `rill org create` or `rill deploy` to create one.")
+				warn.Println("You are not part of any org. Run `rill org create` to create one.")
```
begelundmuller, line 67: Restore this – `rill deploy` also creates orgs

admin/server/auth/handlers.go
```diff
@@ -53,6 +55,18 @@ func (a *Authenticator) authLogin(w http.ResponseWriter, r *http.Request) {
...
+			// clear cookie
+			sess.Options.MaxAge = -1
+			err = sess.Save(r, w)
+			if err != nil {
+				http.Error(w, fmt.Sprintf("failed to clear session: %s", err), http.StatusInternalServerError)
+				return
+			}
+			http.Redirect(w, r, r.URL.RequestURI(), http.StatusTemporaryRedirect)
+			return
+		}
```
begelundmuller, line 69: We have `a.cookies.Get` in several places where we need this check.

I think instead of clear+redirect, it should just set `sess` to an empty/fresh session if the cookie fails to parse. The cookie will usually be updated shortly after anyway. And maybe log the error?

begelundmuller (changes requested):

cli/cmd/auth/login.go
k-anshul: OOPS got confused with multiple iterations.

admin/server/auth/handlers.go
k-anshul: The usual [recommendation](https://github.com/gorilla/securecookie/blob/9865a2b6137913abf1982cc4b1e8acdc35b2f052/securecookie.go#L39) seems to be to reject the request completely. Hence I am clearing the cookie and redirecting to login page.
From what I understood this should only be happening in cases when user is trying to edit the cookies manually hence its okay to not log this error. The expired timestamp happened bcz of `max-age` configuration mismatch which is fixed now.
Since the recommendation is to reject the request, I am not taking any action anywhere and the client(UI in this case) will redirect the user to login page but we can redirect the user to login page in other cases as well.

begelundmuller: I agree, but currently it's not redirecting to the login page, it's redirecting to the same page (which I think makes sense). So in this case, what I'm talking about will be the exact same behavior, just without a redirect and extra return case.

Everywhere, we could just have `sess := a.getCookie(r, name)` (without the error case, since if there's an error, it will just log and return a clean cookie). Would be cleaner?

begelundmuller: Btw, this can also happen if we hard-reset session keys due to a compromise. Or if we reduce `max-age` in the future.

begelundmuller (commented):
Instead of `sessionutil`, consider adding a type that wraps `sessions.CookieStore` in `admin/server/cookies` and takes a logger (since the config here is arguably too specific to this server to be considered a generic util in `pkg`). 

So to initialize it, we might have `cookies.New(logger, ...)` and in the code, you can just do `s.cookies.Get(r, cookieName` like before, but without the error handling

begelundmuller (approved):

PR #2224, opened 2023-04-25
enforce basic quotas on admin apis

pjain1:
Resolves https://github.com/rilldata/rill-developer/issues/2207

### DIFF
--- a/admin/database/database.go
+++ b/admin/database/database.go
@@ -129,6 +129,11 @@ type DB interface {
 	FindProjectInvite(ctx context.Context, projectID, userEmail string) (*ProjectInvite, error)
 	InsertProjectInvite(ctx context.Context, email, projectID, roleID, invitedByID string) error
 	DeleteProjectInvite(ctx context.Context, id string) error
+
+	CountOrganizationProjects(ctx context.Context, orgID string) (int, error)
+	CountOrganizationDeploymentsAndSlots(ctx context.Context, orgID string) (*OrganizationDeploymentsAndSlots, error)
+	CountOrganizationOutstandingInvitations(ctx context.Context, orgID string) (int, error)
+	CountSingleUserOrganizationForUser(ctx context.Context, userID string) (int, error)
 }
 
 // Tx represents a database transaction. It can only be used to commit and rollback transactions.
@@ -150,12 +155,17 @@ var ErrNotUnique = errors.New("database: violates unique constraint")
 
 // Organization represents a tenant.
 type Organization struct {
-	ID             string
-	Name           string
-	Description    string
-	AllUsergroupID *string   `db:"all_usergroup_id"`
-	CreatedOn      time.Time `db:"created_on"`
-	UpdatedOn      time.Time `db:"updated_on"`
+	ID                          string
+	Name                        string
+	Description                 string
+	AllUsergroupID              *string   `db:"all_usergroup_id"`
+	CreatedOn                   time.Time `db:"created_on"`
+	UpdatedOn                   time.Time `db:"updated_on"`
+	QuotaProjects               int       `db:"quota_projects"`
+	QuotaDeployments            int       `db:"quota_deployments"`
+	QuotaSlotsTotal             int       `db:"quota_slots_total"`
+	QuotaSlotsPerDeployment     int       `db:"quota_slots_per_deployment"`
+	QuotaOutstandingInvitations int       `db:"quota_outstanding_invitations"`
 }
 
 // InsertOrganizationOptions defines options for inserting a new org
@@ -279,13 +289,14 @@ type RuntimeSlotsUsed struct {
 // User is a person registered in Rill.
 // Users may belong to multiple organizations and projects.
 type User struct {
-	ID             string
-	Email          string
-	DisplayName    string    `db:"display_name"`
-	PhotoURL       string    `db:"photo_url"`
-	GithubUsername string    `db:"github_username"`
-	CreatedOn      time.Time `db:"created_on"`
-	UpdatedOn      time.Time `db:"updated_on"`
+	ID                  string
+	Email               string
+	DisplayName         string    `db:"display_name"`
+	PhotoURL            string    `db:"photo_url"`
+	GithubUsername      string    `db:"github_username"`
+	CreatedOn           time.Time `db:"created_on"`
+	UpdatedOn           time.Time `db:"updated_on"`
+	QuotaSingleuserOrgs int       `db:"quota_singleuser_orgs"`
 }
 
 // InsertUserOptions defines options for inserting a new user
@@ -445,4 +456,9 @@ type Invite struct {
 	Email     string
 	Role      string
 	InvitedBy string `db:"invited_by"`
+}
+
+type OrganizationDeploymentsAndSlots struct {
+	Deployments int
+	Slots       int
 }
--- a/admin/database/database.go
+++ b/admin/database/database.go
@@ -70,14 +70,14 @@ type DB interface {
 	InsertProject(ctx context.Context, opts *InsertProjectOptions) (*Project, error)
 	DeleteProject(ctx context.Context, id string) error
 	UpdateProject(ctx context.Context, id string, opts *UpdateProjectOptions) (*Project, error)
-	CountOrganizationProjects(ctx context.Context, orgID string) (int, error)
+	CountProjectsForOrganization(ctx context.Context, orgID string) (int, error)
 
 	FindDeployments(ctx context.Context, projectID string) ([]*Deployment, error)
 	FindDeployment(ctx context.Context, id string) (*Deployment, error)
 	InsertDeployment(ctx context.Context, opts *InsertDeploymentOptions) (*Deployment, error)
 	DeleteDeployment(ctx context.Context, id string) error
 	UpdateDeploymentStatus(ctx context.Context, id string, status DeploymentStatus, logs string) (*Deployment, error)
-	CountOrganizationDeploymentsAndSlots(ctx context.Context, orgID string) (*OrganizationDeploymentsAndSlots, error)
+	CountDeploymentsForOrganization(ctx context.Context, orgID string) (*DeploymentsCount, error)
 
 	ResolveRuntimeSlotsUsed(ctx context.Context) ([]*RuntimeSlotsUsed, error)
 
@@ -113,7 +113,7 @@ type DB interface {
 	InsertOrganizationMemberUser(ctx context.Context, orgID, userID, roleID string) error
 	DeleteOrganizationMemberUser(ctx context.Context, orgID, userID string) error
 	UpdateOrganizationMemberUserRole(ctx context.Context, orgID, userID, roleID string) error
-	CountSingleUserOrganizationsForMemberUser(ctx context.Context, userID string) (int, error)
+	CountSingleuserOrganizationsForMemberUser(ctx context.Context, userID string) (int, error)
 
 	FindProjectMemberUsers(ctx context.Context, projectID string) ([]*Member, error)
 	InsertProjectMemberUser(ctx context.Context, projectID, userID, roleID string) error
@@ -126,7 +126,7 @@ type DB interface {
 	FindOrganizationInvite(ctx context.Context, orgID, userEmail string) (*OrganizationInvite, error)
 	InsertOrganizationInvite(ctx context.Context, email, orgID, roleID, invitedByID string) error
 	DeleteOrganizationInvite(ctx context.Context, id string) error
-	CountOrganizationOutstandingInvites(ctx context.Context, orgID string) (int, error)
+	CountInvitesForOrganization(ctx context.Context, orgID string) (int, error)
 
 	FindProjectInvites(ctx context.Context, projectID string) ([]*Invite, error)
 	FindProjectInvitesByEmail(ctx context.Context, userEmail string) ([]*ProjectInvite, error)
@@ -457,7 +457,7 @@ type Invite struct {
 	InvitedBy string `db:"invited_by"`
 }
 
-type OrganizationDeploymentsAndSlots struct {
+type DeploymentsCount struct {
 	Deployments int
 	Slots       int
 }
--- a/admin/database/database.go
+++ b/admin/database/database.go
@@ -70,12 +70,14 @@ type DB interface {
 	InsertProject(ctx context.Context, opts *InsertProjectOptions) (*Project, error)
 	DeleteProject(ctx context.Context, id string) error
 	UpdateProject(ctx context.Context, id string, opts *UpdateProjectOptions) (*Project, error)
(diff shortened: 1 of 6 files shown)


admin/database/postgres/postgres.go
```diff
@@ -802,6 +802,56 @@ func (c *connection) DeleteProjectInvite(ctx context.Context, id string) error {
...
+		return 0, parseErr(err)
+	}
+	return count, nil
+}
+
+func (c *connection) CountSingleUserOrganizationForUser(ctx context.Context, userID string) (int, error) {
+	var count int
+	err := c.getDB(ctx).QueryRowxContext(ctx, `
+	SELECT SUM(total_count) as total_count FROM (
+	    SELECT CASE WHEN COUNT(*) = 1 THEN 1 ELSE 0 END as total_count FROM users_orgs_roles GROUP BY org_id HAVING org_id IN (
```
begelundmuller, line 844: Would it be cheaper with `WHERE` instead of `HAVING`?

admin/server/organizations.go
```diff
@@ -86,7 +86,20 @@ func (s *Server) CreateOrganization(ctx context.Context, req *adminv1.CreateOrga
 		return nil, status.Error(codes.Unauthenticated, "not authenticated as a user")
 	}
 
-	org, err := s.admin.CreateOrganizationForUser(ctx, claims.OwnerID(), req.Name, req.Description)
+	user, err := s.admin.DB.FindUser(ctx, claims.OwnerID())
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	// check single user org limit for this user
+	count, err := s.admin.DB.CountSingleUserOrganizationForUser(ctx, user.ID)
```
begelundmuller, line 94: nit: move the comment to the top of the block

admin/server/organizations.go
```diff
@@ -86,7 +86,20 @@ func (s *Server) CreateOrganization(ctx context.Context, req *adminv1.CreateOrga
...
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	if count >= user.QuotaSingleuserOrgs {
+		return nil, status.Errorf(codes.InvalidArgument, "user has reached the limit of %d single user orgs", user.QuotaSingleuserOrgs)
```
begelundmuller, line 99: Make error `quota exceeded: you can only create %d single-user orgs`

admin/server/organizations.go
```diff
@@ -199,6 +212,14 @@ func (s *Server) AddOrganizationMember(ctx context.Context, req *adminv1.AddOrga
 		return nil, status.Error(codes.PermissionDenied, "not allowed to add org members")
 	}
 
+	count, err := s.admin.DB.CountOrganizationOutstandingInvitations(ctx, org.ID)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	if count >= org.QuotaOutstandingInvitations {
+		return nil, status.Errorf(codes.FailedPrecondition, "cannot invite more users, org %q already %d outstanding invitations", org.Name, org.QuotaOutstandingInvitations)
```
begelundmuller, line 220: Make error: `quota exceeded: org can at most have %d outstanding invitations`

Including the org name when it's in the request payload is not needed. That applies multiple places

admin/server/projects.go
```diff
@@ -207,10 +207,36 @@ func (s *Server) CreateProject(ctx context.Context, req *adminv1.CreateProjectRe
...
+		return nil, status.Errorf(codes.ResourceExhausted, "quota exceeded: org %q is limited to %d projects", org.Name, org.QuotaProjects)
+	}
+
 	if !claims.OrganizationPermissions(ctx, org.ID).CreateProjects {
 		return nil, status.Error(codes.PermissionDenied, "does not have permission to create projects")
 	}
 
+	// check per deployment slots limit
+	if int(req.ProdSlots) > org.QuotaSlotsPerDeployment {
+		return nil, status.Errorf(codes.ResourceExhausted, "quota exceeded: org %q is limited to %d slots per deployment", org.Name, org.QuotaSlotsPerDeployment)
```
begelundmuller, line 225: Make it `quota exceeded: org can't provision more than %d slots per deployment`

admin/server/projects.go
```diff
@@ -207,10 +207,36 @@ func (s *Server) CreateProject(ctx context.Context, req *adminv1.CreateProjectRe
...
+	}
+
+	// check per project deployments and slots limit
+	deploymentsSlots, err := s.admin.DB.CountOrganizationDeploymentsAndSlots(ctx, org.ID)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	if deploymentsSlots.Deployments >= org.QuotaDeployments {
+		return nil, status.Errorf(codes.ResourceExhausted, "quota exceeded: org %q is limited to limited to %d deployments", org.Name, org.QuotaDeployments)
```
begelundmuller, line 234: nit: spelling

admin/server/projects.go
```diff
@@ -377,6 +403,19 @@ func (s *Server) AddProjectMember(ctx context.Context, req *adminv1.AddProjectMe
...
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	// TODO can be optimized by fetching org level quotas in FindProjectByName
+	org, err := s.admin.DB.FindOrganization(ctx, proj.OrganizationID)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	if count >= org.QuotaOutstandingInvitations {
+		return nil, status.Errorf(codes.FailedPrecondition, "cannot invite more users, org %q already %d outstanding invitations", org.Name, org.QuotaOutstandingInvitations)
+	}
```
begelundmuller, line 417: 1. Can remove the TODO (many queries can theoretically be optimized, but let's wait until we see latency problems before optimizing). 
2. Add a comment at top of the block like other places (like `// Check outstanding invites quota`)

admin/database/database.go
```diff
@@ -129,6 +129,11 @@ type DB interface {
 	FindProjectInvite(ctx context.Context, projectID, userEmail string) (*ProjectInvite, error)
 	InsertProjectInvite(ctx context.Context, email, projectID, roleID, invitedByID string) error
 	DeleteProjectInvite(ctx context.Context, id string) error
+
+	CountOrganizationProjects(ctx context.Context, orgID string) (int, error)
+	CountOrganizationDeploymentsAndSlots(ctx context.Context, orgID string) (*OrganizationDeploymentsAndSlots, error)
+	CountOrganizationOutstandingInvitations(ctx context.Context, orgID string) (int, error)
+	CountSingleUserOrganizationForUser(ctx context.Context, userID string) (int, error)
```
begelundmuller, line 136: Missing plural for organizations

admin/database/database.go
```diff
@@ -129,6 +129,11 @@ type DB interface {
 	FindProjectInvite(ctx context.Context, projectID, userEmail string) (*ProjectInvite, error)
 	InsertProjectInvite(ctx context.Context, email, projectID, roleID, invitedByID string) error
 	DeleteProjectInvite(ctx context.Context, id string) error
+
+	CountOrganizationProjects(ctx context.Context, orgID string) (int, error)
+	CountOrganizationDeploymentsAndSlots(ctx context.Context, orgID string) (*OrganizationDeploymentsAndSlots, error)
+	CountOrganizationOutstandingInvitations(ctx context.Context, orgID string) (int, error)
```
begelundmuller, line 135: Invites instead of invitations

admin/database/database.go
begelundmuller, line 136: Also, consider grouping by object (project, deployment, invites, etc.) instead of verb (`Count`), like the other functions in this interface.

admin/database/database.go
```diff
@@ -446,3 +457,8 @@ type Invite struct {
 	Role      string
 	InvitedBy string `db:"invited_by"`
 }
+
+type OrganizationDeploymentsAndSlots struct {
+	Deployments int
+	Slots       int
+}
```
begelundmuller, line 464: Subjective, but might be simpler to just return (int, int, error) from handler instead of having a separate type

begelundmuller (changes requested):
Additionally: Align use of error status codes, it's a little mixed. I think `codes.FailedPrecondition` is better than `codes.ResourceExhausted` because the latter is usually used for rate limits, which can be retried later (whereas this should not be retried without changing a precondition first).

admin/database/postgres/migrations/0005.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE orgs ADD COLUMN quota_projects INTEGER DEFAULT 5;
+ALTER TABLE orgs ADD COLUMN quota_deployments INTEGER DEFAULT 10;
+ALTER TABLE orgs ADD COLUMN quota_slots_total INTEGER DEFAULT 20;
+ALTER TABLE orgs ADD COLUMN quota_slots_per_deployment INTEGER DEFAULT 5;
+ALTER TABLE orgs ADD COLUMN quota_outstanding_invitations INTEGER DEFAULT 200;
+
+ALTER TABLE users ADD COLUMN quota_singleuser_orgs INTEGER DEFAULT 3;
```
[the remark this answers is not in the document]
pjain1: If application level defaults are non zero then what is the use of setting DB defaults to -1 when it will be used ? I mean when will I know the -1 is set by a user and not a DB default.

admin/database/postgres/postgres.go
```diff
@@ -802,6 +802,56 @@ func (c *connection) DeleteProjectInvite(ctx context.Context, id string) error {
...
+
+func (c *connection) CountOrganizationOutstandingInvitations(ctx context.Context, orgID string) (int, error) {
+	var count int
+	// count outstanding org invites as well as project invites for this org
+	err := c.getDB(ctx).QueryRowxContext(ctx, `
+		SELECT SUM(total_count) as total_count FROM (
+  			SELECT COUNT(*) as total_count FROM org_invites WHERE org_id = $1
+  			UNION ALL
+  			SELECT COUNT(*) as total_count FROM project_invites WHERE project_id IN (SELECT id FROM projects WHERE org_id = $1)
+		) as subquery
```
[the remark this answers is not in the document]
pjain1: postgres needs an alias

admin/database/postgres/postgres.go
pjain1: yes will fix

admin/server/organizations.go
pjain1: will change the error message. Default org name will get added to request most of the time without user entering it, so added just so that its in front of user.

admin/database/database.go
pjain1: ok this looked cleaner as return types are explicit.

admin/database/postgres/migrations/0005.sql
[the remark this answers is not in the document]
begelundmuller: I meant that when using `ADD COLUMN` and `NOT NULL`, you have to specify a `DEFAULT` clause. But this is not the place to configure defaults, so just put a fixed value for all of them here.

admin/database/database.go
```diff
@@ -70,12 +70,14 @@ type DB interface {
 	InsertProject(ctx context.Context, opts *InsertProjectOptions) (*Project, error)
 	DeleteProject(ctx context.Context, id string) error
 	UpdateProject(ctx context.Context, id string, opts *UpdateProjectOptions) (*Project, error)
+	CountOrganizationProjects(ctx context.Context, orgID string) (int, error)
```
begelundmuller, line 73: `CountProjectsForOrg`

admin/database/database.go
```diff
@@ -70,12 +70,14 @@ type DB interface {
...
 
 	FindDeployments(ctx context.Context, projectID string) ([]*Deployment, error)
 	FindDeployment(ctx context.Context, id string) (*Deployment, error)
 	InsertDeployment(ctx context.Context, opts *InsertDeploymentOptions) (*Deployment, error)
 	DeleteDeployment(ctx context.Context, id string) error
 	UpdateDeploymentStatus(ctx context.Context, id string, status DeploymentStatus, logs string) (*Deployment, error)
+	CountOrganizationDeploymentsAndSlots(ctx context.Context, orgID string) (*OrganizationDeploymentsAndSlots, error)
```
begelundmuller, line 80: `CountDeploymentsForOrg`

admin/database/database.go
```diff
@@ -123,6 +126,7 @@ type DB interface {
 	FindOrganizationInvite(ctx context.Context, orgID, userEmail string) (*OrganizationInvite, error)
 	InsertOrganizationInvite(ctx context.Context, email, orgID, roleID, invitedByID string) error
 	DeleteOrganizationInvite(ctx context.Context, id string) error
+	CountOrganizationOutstandingInvites(ctx context.Context, orgID string) (int, error)
```
begelundmuller, line 129: `CountInvitesForOrg`

admin/database/database.go
```diff
@@ -111,6 +113,7 @@ type DB interface {
 	InsertOrganizationMemberUser(ctx context.Context, orgID, userID, roleID string) error
 	DeleteOrganizationMemberUser(ctx context.Context, orgID, userID string) error
 	UpdateOrganizationMemberUserRole(ctx context.Context, orgID, userID, roleID string) error
+	CountSingleUserOrganizationsForMemberUser(ctx context.Context, userID string) (int, error)
```
begelundmuller, line 116: Other places, it's `Singleuser` instead of `SingleUser`, so pick one (I think `Singleuser` looks cleanest to show separation)

admin/database/database.go
```diff
@@ -446,3 +456,8 @@ type Invite struct {
 	Role      string
 	InvitedBy string `db:"invited_by"`
 }
+
+type OrganizationDeploymentsAndSlots struct {
```
begelundmuller, line 460: Would suggest renaming it to `DeploymentsCount` then

admin/database/postgres/postgres.go
```diff
@@ -802,6 +802,56 @@ func (c *connection) DeleteProjectInvite(ctx context.Context, id string) error {
 	return nil
 }
 
+func (c *connection) CountOrganizationProjects(ctx context.Context, orgID string) (int, error) {
```
begelundmuller, line 805: Place implementations in same relative location in file as in interface, makes them easier to navigate

begelundmuller (changes requested):

admin/database/postgres/migrations/0005.sql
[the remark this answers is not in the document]
begelundmuller: Can also just have default 0 here. I just suggested `-1` as a way of setting infinite, instead of setting a high number – but we can just keep the checks simple, we can anyway just set a very high number if needed.

begelundmuller (approved):

PR #2227, opened 2023-04-25
Fix 2075 removable list menu closing when clicked 3

### LINKED ISSUE #2075: the `RemovableListMenu` component closes after I select options; rather than staying open

bcolloran:
Fixes #2075 by making 2 changes:

1. passing exclude toggle state down in a store rather than as props. I don't know why exactly this workaround is needed, but it seems like there might be something complicated happening with the interplay of the `Portal` and `Menu` elements used by the chip menu, and the that svelte compiler is not able to correctly invalidate the props passed to the `RemovableListMenu`. See https://github.com/rilldata/rill-developer/pull/2197 for more details on the non-working behavior.

2. Removing the `focusin` listener fixes the `Menu` component closing immediately when clicking something within the menu. I clicked around in every menu I could find in the app, and removing this didn't seem to break anything. I'm not sure why this listener was here in the first place. @hamilton do you know of any specific need for that listener?


bcolloran:
@djbarnwal can you review?

djbarnwal (approved):
Tested the menu, works as expected. I am not sure what are the repercussions of removing the `focusin` event listener but since all the menus are working I suppose we can merge this for now. We can come back to this later if we are missing something cc @hamilton

PR #2231, opened 2023-04-25
Add remove_from_projects flag to RemoveOrganizationMember

rakeshsharma14317:


admin/server/organizations.go
```diff
@@ -328,6 +328,24 @@ func (s *Server) RemoveOrganizationMember(ctx context.Context, req *adminv1.Remo
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
+	// delete from projects if RemoveFromProjects flag is set
+	if req.RemoveFromProjects {
+		projects, err := s.admin.DB.FindProjectsForOrgAndUser(ctx, req.Organization, user.ID)
```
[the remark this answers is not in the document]
pjain1: cc @begelundmuller

cli/cmd/user/remove.go
```diff
@@ -54,6 +56,7 @@ func RemoveCmd(cfg *config.Config) *cobra.Command {
 	removeCmd.Flags().StringVar(&cfg.Org, "org", cfg.Org, "Organization")
 	removeCmd.Flags().StringVar(&projectName, "project", "", "Project")
 	removeCmd.Flags().StringVar(&email, "email", "", "Email of the user")
+	removeCmd.Flags().BoolVar(&removeFromProjects, "remove-from-projects", false, "Remove user from projects")
```
begelundmuller, line 59: This should be `true` by default. Maybe flip it to `--keep-project-roles` with a default of `false`? And a description of "Keep roles granted directly on projects in the org"

admin/server/organizations.go
```diff
@@ -333,6 +333,35 @@ func (s *Server) RemoveOrganizationMember(ctx context.Context, req *adminv1.Remo
...
+			if err != nil {
+				return nil, status.Error(codes.Internal, err.Error())
+			}
+		}
+
+		err = tx.Commit()
+		if err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+	}
```
begelundmuller, line 363: Move this into a single DB call, like `s.admin.DB.DeleteAllProjectMemberUserForOrg`, and call it inside the existing transaction used for deleting the org membership.

begelundmuller (changes requested):

admin/server/organizations.go
```diff
@@ -328,6 +328,15 @@ func (s *Server) RemoveOrganizationMember(ctx context.Context, req *adminv1.Remo
 	if err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
 	}
+
+	// delete from projects if RemoveFromProjects flag is set
```
pjain1, line 332: flag name seems wrong, it should be `KeepProjectRoles`

admin/server/organizations.go
```diff
@@ -328,6 +328,15 @@ func (s *Server) RemoveOrganizationMember(ctx context.Context, req *adminv1.Remo
...
+	if req.KeepProjectRoles {
```
pjain1, line 333: if `KeepProjectRoles` is true then we are deleting the roles, logic should be flipped ?

admin/database/postgres/postgres.go
```diff
@@ -696,6 +696,11 @@ func (c *connection) DeleteProjectMemberUser(ctx context.Context, projectID, use
 	return checkDeleteRow("project member", res, err)
 }
 
+func (c *connection) DeleteAllProjectMemberUserForOrg(ctx context.Context, orgID, userID string) error {
```
pjain1, line 699: nit: `DeleteAllProjectMemberUserForOrg` -> `DeleteAllProjectMemberUserForOrganization` for consistency

admin/server/organizations.go
rakeshsharma14317: Yeah it should be `if !req.KeepProjectRoles` then delete, 

But @begelundmuller should we keep `keep-project-roles` as false by default? it means we will delete the roles if this flag is not set, it that fine?

admin/database/postgres/postgres.go
```diff
@@ -696,6 +696,11 @@ func (c *connection) DeleteProjectMemberUser(ctx context.Context, projectID, use
 	return checkDeleteRow("project member", res, err)
 }
 
+func (c *connection) DeleteAllProjectMemberUserForOrganization(ctx context.Context, orgID, userID string) error {
+	res, err := c.getDB(ctx).ExecContext(ctx, "DELETE FROM users_projects_roles upr WHERE upr.user_id = $1 AND upr.project_id IN (SELECT p.id FROM projects p WHERE p.org_id = $2)", userID, orgID)
+	return checkDeleteRow("project member", res, err)
```
begelundmuller, line 701: Use `parseErr`, this one only works when exactly 1 row will be deleted (in this case, 0 to many rows may be deleted)

admin/server/organizations.go
```diff
@@ -328,6 +328,15 @@ func (s *Server) RemoveOrganizationMember(ctx context.Context, req *adminv1.Remo
 	if err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
 	}
+
+	// delete from projects if KeepProjectRoles flag is set
+	if !req.KeepProjectRoles {
+		err = s.admin.DB.DeleteAllProjectMemberUserForOrganization(ctx, req.Organization, user.ID)
```
begelundmuller, line 334: `req.Organization` is the org name, not ID. Must use `org.ID`

begelundmuller (changes requested):

begelundmuller (approved):

PR #2234, opened 2023-04-25
Cleanup number formatting phase 1

bcolloran:
@djbarnwal this PR just removes some dead code following from the number formatting work a couple months ago. This is the low hanging fruit from some refactors around that code that @hamilton and I discussed when those changes landed, just getting back to it now. I should have some more stuff on this topic forthcoming.


djbarnwal (approved):
Looks Good, thanks for the PR

PR #2245, opened 2023-04-26
removes unused component props that were causing console noise

bcolloran:
quick pr while working on other stuff -- fixes all this console noise
image


bcolloran:
@djbarnwal  can you rubberstamp this 2 second review, which removes unknown props in 2 component invocations? Thanks!

djbarnwal (approved):
Looks good!

PR #2249, opened 2023-04-27
user addition notification email

pjain1:
Send notification email upon addition of an exiting user to an org or project.

### DIFF
--- a/admin/email/email.go
+++ b/admin/email/email.go
@@ -30,8 +30,8 @@ func (c *Client) SendOrganizationAdditionNotification(toEmail, toName, orgName, 
 	err := c.sender.Send(
 		toEmail,
 		toName,
-		"Addition to an organization",
-		fmt.Sprintf("You've been added to organization <b>%s</b> as <b>%s</b>. Follow this <a href=\"%s\">link</a> to visit it.", orgName, roleName, c.frontendURL),
+		fmt.Sprintf("You've been added to %q", orgName),
+		fmt.Sprintf("You've been added to the organization <b>%s</b> as <b>%s</b>. Follow this <a href=\"%s\">link</a> to visit it.", orgName, roleName, c.frontendURL),
 	)
 	return err
 }
@@ -50,8 +50,8 @@ func (c *Client) SendProjectAdditionNotification(toEmail, toName, projectName, r
 	err := c.sender.Send(
 		toEmail,
 		toName,
-		"Addition to a project",
-		fmt.Sprintf("You've been added to project <b>%s</b> as <b>%s</b>. Follow this <a href=\"%s\">link</a> to visit it.", projectName, roleName, c.frontendURL),
+		fmt.Sprintf("You've been added to %q", projectName),
+		fmt.Sprintf("You've been added to the project <b>%s</b> as <b>%s</b>. Follow this <a href=\"%s\">link</a> to visit it.", projectName, roleName, c.frontendURL),
 	)
 	return err
 }
--- a/admin/email/email.go
+++ b/admin/email/email.go
@@ -26,12 +26,32 @@ func (c *Client) SendOrganizationInvite(toEmail, toName, orgName, roleName strin
 	return err
 }
 
+func (c *Client) SendOrganizationAdditionNotification(toEmail, toName, orgName, roleName string) error {
+	err := c.sender.Send(
+		toEmail,
+		toName,
+		"Addition to an organization",
+		fmt.Sprintf("You've been added to organization <b>%s</b> as <b>%s</b>. Follow this <a href=\"%s\">link</a> to visit it.", orgName, roleName, c.frontendURL),
+	)
+	return err
+}
+
 func (c *Client) SendProjectInvite(toEmail, toName, projectName, roleName string) error {
 	err := c.sender.Send(
 		toEmail,
 		toName,
 		"Invitation to join Rill",
 		fmt.Sprintf("You have been invited to project <b>%s</b> as <b>%s</b>. Please sign into Rill <a href=\"%s\">here</a> to accept invitation.", projectName, roleName, c.frontendURL),
+	)
+	return err
+}
+
+func (c *Client) SendProjectAdditionNotification(toEmail, toName, projectName, roleName string) error {
+	err := c.sender.Send(
+		toEmail,
+		toName,
+		"Addition to a project",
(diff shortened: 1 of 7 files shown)


admin/user.go
```diff
@@ -149,6 +149,21 @@ func (s *Service) InviteUserToOrganization(ctx context.Context, email, inviterID
 	return nil
 }
 
+func (s *Service) NotifyUserAdditionToOrganization(ctx context.Context, email, orgName, roleName string) error {
+	// Validate email address
+	_, err := mail.ParseAddress(email)
```
pjain1, line 154: Should I move this validation to start of `AddOrganizationMember` and `AddProjectMember` so that its not need in downstream methods like this and invitation method.

pjain1: Also it will be good to validate before entering it in the database.

pjain1: moved it to Add methods

admin/user.go
```diff
@@ -149,15 +143,17 @@ func (s *Service) InviteUserToOrganization(ctx context.Context, email, inviterID
 	return nil
 }
 
-func (s *Service) InviteUserToProject(ctx context.Context, email, inviterID, projectID, roleID, projectName, roleName string) error {
-	// Validate email address
-	_, err := mail.ParseAddress(email)
+func (s *Service) NotifyUserAdditionToOrganization(ctx context.Context, email, orgName, roleName string) error {
```
pjain1, line 146: This can be inlined but we will need to export `email` from `Service` in order to use it in `server` package

admin/server/organizations.go
```diff
@@ -212,6 +213,12 @@ func (s *Server) AddOrganizationMember(ctx context.Context, req *adminv1.AddOrga
 		return nil, status.Error(codes.PermissionDenied, "not allowed to add org members")
 	}
 
+	// Validate email address
+	_, err = mail.ParseAddress(req.Email)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "invalid email address: %q", req.Email)
+	}
+
```
begelundmuller, line 221: If you refactor `InviteUserToOrganization` to accept an `opts` parameter instead, you can add a validation tag for `email` there instead (see `database.InsertProjectOptions` as an example).

Alternatively (or additionally), you can use the validation tags in the proto file to enforce it at the request parsing stage.

admin/email/email.go
```diff
@@ -26,6 +26,16 @@ func (c *Client) SendOrganizationInvite(toEmail, toName, orgName, roleName strin
 	return err
 }
 
+func (c *Client) SendOrganizationAdditionNotification(toEmail, toName, orgName, roleName string) error {
+	err := c.sender.Send(
+		toEmail,
+		toName,
+		"Addition to an organization",
```
begelundmuller, line 33: Make it `"You've been added to %q", orgName`

admin/email/email.go
```diff
@@ -26,6 +26,16 @@ func (c *Client) SendOrganizationInvite(toEmail, toName, orgName, roleName strin
...
+		fmt.Sprintf("You've been added to organization <b>%s</b> as <b>%s</b>. Follow this <a href=\"%s\">link</a> to visit it.", orgName, roleName, c.frontendURL),
```
begelundmuller, line 34: Put "the" before "organization"

admin/email/email.go
```diff
@@ -35,3 +45,13 @@ func (c *Client) SendProjectInvite(toEmail, toName, projectName, roleName string
 	)
 	return err
 }
+
+func (c *Client) SendProjectAdditionNotification(toEmail, toName, projectName, roleName string) error {
+	err := c.sender.Send(
+		toEmail,
+		toName,
+		"Addition to a project",
```
begelundmuller, line 53: Make it `"You've been added to \"%s/%s\"", orgName, projectName`

admin/email/email.go
```diff
@@ -35,3 +45,13 @@ func (c *Client) SendProjectInvite(toEmail, toName, projectName, roleName string
...
+		fmt.Sprintf("You've been added to project <b>%s</b> as <b>%s</b>. Follow this <a href=\"%s\">link</a> to visit it.", projectName, roleName, c.frontendURL),
```
begelundmuller, line 54: 1. Put "the" before "project"
2. Format project name as `orgName/projectName`

admin/server/organizations.go
```diff
@@ -274,6 +281,11 @@ func (s *Server) AddOrganizationMember(ctx context.Context, req *adminv1.AddOrga
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
+	err = s.admin.NotifyUserAdditionToOrganization(ctx, req.Email, org.Name, role.Name)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
```
begelundmuller, line 287: This feels a little redundant. Let's just export the emailer, so it becomes `s.admin.Email...`.

Ideally, we would move the whole thing into a `s.admin.AddOrgMember` func, but don't have to refactor like that now.

admin/server/projects.go
```diff
@@ -403,6 +404,12 @@ func (s *Server) AddProjectMember(ctx context.Context, req *adminv1.AddProjectMe
 		return nil, status.Error(codes.PermissionDenied, "not allowed to add project members")
 	}
 
+	// Validate email address
+	_, err = mail.ParseAddress(req.Email)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "invalid email address: %q", req.Email)
+	}
```
begelundmuller, line 411: Same comment as for org

admin/server/projects.go
```diff
@@ -452,6 +459,11 @@ func (s *Server) AddProjectMember(ctx context.Context, req *adminv1.AddProjectMe
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
+	err = s.admin.NotifyUserAdditionToProject(ctx, req.Email, proj.Name, role.Name)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
```
begelundmuller, line 465: Same comment as for org

begelundmuller (changes requested):

admin/server/organizations.go
pjain1: validating before call to `InviteUserToOrganization` will prevent from doing extra work like fetching org and role from db. proto file already has `[(validate.rules).string.email = true]` so I think we are good, we can remove the check from here.

admin/email/email.go
pjain1: orgName not present, will need to be fetched using orgId

begelundmuller (approved):

PR #2254, opened 2023-04-27
Project layout changes

djbarnwal:
Add dashboard list to project and change layout according to mocks


ericpgreen2 (approved):
Looks good. Just added one commit to handle a long list of dashboards. Added an `overflow-y-auto` and `max-h-[170px]` to match the mocks.

PR #2256, opened 2023-04-27
Admin: Add tests for RBAC

pjain1:
Contains tests of Org APIs. Tests for project APIs requires mocking github client, need to investigate that further.


admin/user.go
```diff
@@ -128,8 +128,14 @@ func (s *Service) CreateOrganizationForUser(ctx context.Context, userID, orgName
 }
 
 func (s *Service) InviteUserToOrganization(ctx context.Context, email, inviterID, orgID, roleID, orgName, roleName string) error {
+	// Validate email address
```
pjain1, line 131: adding back the validation

admin/admin.go
```diff
@@ -94,3 +94,14 @@ func (s *Service) Close() error {
...
+// NewMock for testing
+func NewMock(db database.DB, githubClient *github.Client, prov provisioner.Provisioner, issuer *auth.Issuer, emailClient *email.Client) *Service {
+	return &Service{
+		DB:          db,
+		github:      githubClient,
+		provisioner: prov,
+		issuer:      issuer,
+		Email:       emailClient,
+	}
+}
```
begelundmuller, line 107: It would be better if we could just use `New` and pass in mock dependencies where applicable. It would ensure better consistency, especially as we increase the functionality in this service in the future.

The only thing preventing that appears to be the Github client. However, it's used very little, and we use a very small part of its functionality. What about creating a `type Github interface` (in `admin/github.go`) with just the two or three functions we need, having a real and a mock implementation of it, and then passing `github Github` as a parameter to `admin.New` instead?

begelundmuller (changes requested):

admin/server/auth/auth.go
```diff
@@ -64,3 +64,9 @@ func NewAuthenticator(logger *zap.Logger, adm *admin.Service, cookieStore *cooki
 
 	return a, nil
 }
+
+func NewMockAuthenticator(adminService *admin.Service) *Authenticator {
+	return &Authenticator{
+		admin: adminService,
+	}
+}
```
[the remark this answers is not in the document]
pjain1: `NewAuthenticator` actually creates 
```
oidcProvider, err := oidc.NewProvider(context.Background(), "https://"+opts.AuthDomain+"/")
```
which will fail so thats why create a new mock one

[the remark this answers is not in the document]
pjain1: Don't we do the same thing with mocking libraries ? We mock the object and return values of methods on that object that we expect to be called. Upon adding new tests or change in existing flow, tests break and then we add new methods that are expected to be called now.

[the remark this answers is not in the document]
pjain1: I tried `gorillio-stage.auth0.com` as AuthDomain and it worked, was not sure if tests would have internet access.

admin/admin.go
pjain1: tried to abstract out Github using interface
```
type Github interface {
	GetGithubInstallation(ctx context.Context, githubURL string) (int64, error)
	LookupGithubRepoForUser(ctx context.Context, installationID int64, githubURL, gitUsername string) (*github.Repository, error)
	ProcessGithubEvent(ctx context.Context, rawEvent any) error
}
```
Implementation of this creates the github client using github app id and key. However this calls needs DB as well. DB can be create before admin.Service creation and can be passed to both service and Github implementation. However, to process github push it needs to reconcile for which it needs `admin.Service` thus creating cyclic dependency. Any thoughts here ?

begelundmuller: I would make the `Github` interface even simpler, just pure proxy to Github API. So anything involving our DB stays in the `admin.Service`. Is that not possible?

admin/server/auth/auth.go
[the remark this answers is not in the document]
begelundmuller: Yes it's expected with mocks. But in this case, it's not really a mock authenticator (responding with mocked values), but a real one that will just fail on certain code paths. And we actually want to test parts of this authenticator – so it would be nicer if we got the "real" one from NewAuthenticator.

Generally, if testing something that has code paths that shouldn't run in tests (in this case, the oidc provider stuff), we need to then figure out a way to prevent that – might involve refactoring, to decouple the parts we want to test from those that should be replaced with a mock dependency (e.g. by introducing interfaces and using dependency injection like in the Github interface case). 

In this case, if it seems to work by passing the stage auth domain, it's probably fine for now if you don't want to spend more time on it.

admin/user.go
```diff
@@ -129,43 +129,39 @@ func (s *Service) CreateOrganizationForUser(ctx context.Context, userID, orgName
...
-	if err != nil {
-		return fmt.Errorf("invalid user email address %q", email)
+func (s *Service) InviteUserToOrganization(ctx context.Context, opts *database.InviteUserToOrganizationOptions) error {
+	if err := database.Validate(opts); err != nil {
+		return err
 	}
 
 	// Create invite
-	err = s.DB.InsertOrganizationInvite(ctx, email, orgID, roleID, inviterID)
+	err := s.DB.InsertOrganizationInvite(ctx, opts.Email, opts.OrgID, opts.RoleID, opts.InviterID)
```
begelundmuller: Sorry, I think I misrepresented in another comment. The options types in the `database` package should directly match the database functions, e.g. `InsertOrganizationInviteOptions`. 

I would probably keep the previous func signature on the `Service` here. The call to `s.Email...` will already validate the email, so just need it to happen on the `InsertOrganizationInvite` call

admin/admin.go
pjain1: change, please check

begelundmuller (approved):
Looks good.

Refactored the Github interface slightly to expose the clients instead, and to make the installation cache concurrency safe (since it's used in request handlers).

PR #2257, opened 2023-04-27
Moving the url parsing to dashboard component

AdityaHegde:
In cloud since we make a network call for project status, there is a delay in getting the dashboard name. This will lead to some race condition between creating a store entry from meta api vs parsing from url.

Moving the parsing of url to Dashboard component to make sure that data is available when dashboard is navigated to.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -41,6 +41,7 @@ README.md
 docs
 *.json
 *.cjs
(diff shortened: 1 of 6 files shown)


ericpgreen2 (approved):
Looks good 👍

PR #2263, opened 2023-04-27
Common error handling of CLI errors

### LINKED ISSUE #2176: CLI: Don't print CLI usage for common errors
e.g. user clicks on reject in the login flow. 
We also print the error message twice.

k-anshul:
Fixes following:

- [x] Network errors render as rpc error: code = XXX desc = YYY in CLI instead of just showing the error message. 
- [x] Error messages get shown twice in the CLI on error
- [x] Rill login still says rill auth login
- [x] Give better error messages to handle cases when auth token has been invalidate/expired from backend.
- [x] `rill user list` command fails early if not logged in.
- [x] `rill user list` command returns error if neither org flag is passed nor default flag is set.
- [x] closes #2176


cli/cmd/root.go
```diff
@@ -116,9 +136,15 @@ func runCmd(ctx context.Context, ver config.Version) error {
 	}
 	for _, cmd := range adminCmds {
 		cmd.PersistentFlags().StringVar(&cfg.AdminURL, "api-url", cfg.AdminURL, "Base URL for the admin API")
+		if err := cmd.PersistentFlags().MarkHidden("api-url"); err != nil {
+			panic(err)
+		}
```
begelundmuller, line 141: Only when `!cfg.IsDev()`

cli/cmd/root.go
```diff
@@ -116,9 +136,15 @@ func runCmd(ctx context.Context, ver config.Version) error {
...
 		cmd.PersistentFlags().StringVar(&cfg.AdminTokenOverride, "api-token", "", "Token for authenticating with the admin API")
 		rootCmd.AddCommand(cmd)
 	}
-
+	// silence usage, usage string will only show up if missing arguments/flags
+	rootCmd.SilenceUsage = true
+	// we want to override some error messages
+	rootCmd.SilenceErrors = true
```
begelundmuller, line 148: nit: Move to top of func where other properties on `rootCmd` (like `Version`) is configured

begelundmuller (approved):

PR #2264, opened 2023-04-27
Rill cloud bug bash list

rakeshsharma14317:


cli/cmd/org/org.go
```diff
@@ -41,14 +41,12 @@ func toTable(organizations []*adminv1.Organization, defaultOrg string) []*organi
 
 func toRow(o *adminv1.Organization) *organization {
 	return &organization{
-		Name:        o.Name,
-		Description: o.Description,
-		CreatedAt:   o.CreatedOn.AsTime().String(),
+		Name:      o.Name,
+		CreatedAt: o.CreatedOn.AsTime().String(),
```
begelundmuller, line 45: Time should also be formatted here

begelundmuller (changes requested):
Also please fix merge conflicts

begelundmuller (approved):

PR #2265, opened 2023-04-27
Generate tree-structured CLI documentation for Docusaurus (#2180)

egor-ryashin:


cli/cmd/build/build.go
```diff
@@ -37,12 +37,14 @@ func BuildCmd(cfg *config.Config) *cobra.Command {
...
-	buildCmd.Flags().BoolVar(&verbose, "verbose", false, "Sets the log level to debug")
-	buildCmd.Flags().StringSliceVarP(&variables, "env", "e", []string{}, "Set project variables")
+	flags := buildCmd.Flags()
+	flags.SortFlags = false
+	flags.StringVar(&projectPath, "project", ".", "Project directory")
+	_ = flags.MarkHidden("project")
+	flags.StringVar(&olapDSN, "db", local.DefaultOLAPDSN, "Database DSN")
+	flags.StringVar(&olapDriver, "db-driver", local.DefaultOLAPDriver, "Database driver")
+	flags.BoolVar(&verbose, "verbose", false, "Sets the log level to debug")
+	flags.StringSliceVarP(&variables, "env", "e", []string{}, "Set project variables")
```
begelundmuller, line 47: Why this change? We don't even use this command, should probably deprecate it

cli/cmd/docs/generate.go
```diff
@@ -0,0 +1,193 @@
...
+func GenerateCmd(rootCmd *cobra.Command) *cobra.Command {
+	docsCmd := &cobra.Command{
+		Use:   "generate",
+		Short: "Generate CLI documentation",
+		Args:  cobra.MaximumNArgs(1),
+		Run: func(cmd *cobra.Command, args []string) {
+			dir := "docs/docs/references/CLI/"
+			if len(args) > 0 {
+				dir = args[0]
+			}
```
begelundmuller, line 26: Just make it require exactly one arg – don't need any user affordances here

cli/cmd/docs/generate.go
```diff
@@ -0,0 +1,193 @@
...
+			err := GenMarkdownTree(rootCmd, dir)
+			if err != nil {
+				log.Fatal(err)
+			}
+		},
+	}
+	return docsCmd
+}
+
+func GenMarkdownTree(cmd *cobra.Command, dir string) error {
```
begelundmuller, line 37: 1. Make lowercase, not used outside package
2. Add comment before explaining how this works at a high level and with links to relevant Cobra docs

cli/cmd/docs/generate.go
```diff
@@ -0,0 +1,193 @@
...
+	return nil
+}
+
+type byName []*cobra.Command
+
+func (s byName) Len() int           { return len(s) }
+func (s byName) Swap(i, j int)      { s[i], s[j] = s[j], s[i] }
+func (s byName) Less(i, j int) bool { return s[i].Name() < s[j].Name() }
+
+func GenMarkdownCustom(cmd *cobra.Command, w io.Writer, linkHandler func(string) string) error {
```
begelundmuller, line 128: Make lowercase, not used outside the package

cli/cmd/docs/generate.go
```diff
@@ -0,0 +1,193 @@
...
+		buf.WriteString("```\n\n")
+	}
+	return nil
+}
+
+type byName []*cobra.Command
+
+func (s byName) Len() int           { return len(s) }
+func (s byName) Swap(i, j int)      { s[i], s[j] = s[j], s[i] }
+func (s byName) Less(i, j int) bool { return s[i].Name() < s[j].Name() }
```
begelundmuller, line 126: Move helpers to after `GenMarkdownCustom` (so functions roughly follow call order)

cli/cmd/docs/open.go
```diff
@@ -0,0 +1,24 @@
...
+		Short: "Open docs.rilldata.com",
+		Run: func(cmd *cobra.Command, args []string) {
+			err := browser.Open(docsURL)
+			if err != nil {
+				fmt.Printf("Could not open browser. Copy this URL into your browser: %s\n", docsURL)
+			}
+		},
+	}
+	return docsCmd
+}
```
begelundmuller, line 24: This should still happen just on `rill docs` if possible. If not possible, we should probably move the generator to `scripts/generate_cli_docs` or something.

docs/docs/references/CLI/CLI.md
```diff
@@ -43,7 +43,7 @@ to initialize an empty project.
```
begelundmuller, line 43: Keep this filename and folder name lowercase – they go in the URL. You can change the title or sidebar case in the front matter of the file if needed.

begelundmuller (changes requested):

cli/cmd/build/build.go
egor-ryashin: @nishantmonu51 could also comment on `project` command? I remember we decided to hide it

nishantmonu51: Yes, we don't need to add project here, lets be consistent with `start` and `init`

Makefile
```diff
@@ -1,3 +1,8 @@
+.PHONY: docs
+docs: cli
+	rm -fr docs/docs/references/cli/rill*
+	./rill docs generate docs/docs/references/cli/
```
begelundmuller, line 4: 1. Target should be `docs.generate`
2. Use `go run ./cli docs generate docs/docs/references/cli/` instead, and remove dependency on `cli` target (to prevent full frontend build, which is slow)

docs/docs/references/cli/cli.md
```diff
@@ -43,7 +43,7 @@ to initialize an empty project.
```
begelundmuller, line 43: Move this file to `references/cli-cheat-sheet`

cli/cmd/source/add.go
```diff
@@ -97,6 +97,7 @@ func AddCmd(cfg *config.Config) *cobra.Command {
 	addCmd.Flags().StringVar(&sourceName, "name", "", "Source name (defaults to file name)")
 	addCmd.Flags().BoolVarP(&force, "force", "f", false, "Overwrite the source if it already exists")
 	addCmd.Flags().StringVar(&projectPath, "project", ".", "Project directory")
+	_ = addCmd.Flags().MarkHidden("project")
```
begelundmuller, line 100: Should be changed to `--path`, not hidden

cli/cmd/source/drop.go
```diff
@@ -60,6 +60,7 @@ func DropCmd(cfg *config.Config) *cobra.Command {
 
 	dropCmd.Flags().SortFlags = false
 	dropCmd.Flags().StringVar(&projectPath, "project", ".", "Project directory")
+	_ = dropCmd.Flags().MarkHidden("project")
```
begelundmuller, line 63: Should be changed to `--path`, not hidden

cli/cmd/build/build.go
```diff
@@ -16,8 +16,9 @@ func BuildCmd(cfg *config.Config) *cobra.Command {
 	var variables []string
 
 	buildCmd := &cobra.Command{
-		Use:   "build",
-		Short: "Build project without starting web app",
+		Use:    "build",
+		Short:  "Build project without starting web app",
+		Hidden: !cfg.IsDev(),
```
begelundmuller, line 21: Just delete this file, not used

cli/cmd/docs/open.go
```diff
@@ -0,0 +1,24 @@
+package docs
```
begelundmuller, line 1: This file is now not used, can be removed

begelundmuller (changes requested):
Also, move the generated docs to `references/rill-cli-usage` instead of `references/cli/rill/` to prevent too much nesting

Makefile
egor-ryashin: Item 2 conflicts with
> Also, move the generated docs to references/rill-cli-usage instead of references/cli/rill/ to prevent too much nesting

docs/docs/references/cli/cli.md
egor-ryashin: A separate dir or .md file?

Makefile
begelundmuller: Oh, I just meant to use `go run ./cli` instead of `./rill` (a prod build), not that the arg there needs to match exactly

docs/docs/references/cli/cli.md
egor-ryashin: The name of this file associates with the folder and folder takes the name from the file `TITLE`. If I rename it, then the folder name will change to lowercase `cli`.

egor-ryashin: See https://docusaurus.io/docs/sidebar/autogenerated#category-index-convention

egor-ryashin: changing to this
image

begelundmuller (approved):

PR #2270, opened 2023-04-28
Account for invalid vs. valid dashboards

ericpgreen2:
This PR:
- Includes invalid dashboards in listings (home page, project page, breadcrumbs)
- Allows for navigation to valid dashboards in errored projects

To compose the list of valid & invalid dashboards, we have to synthesize the results from two different runtime queries. Currently, the Tanstack Query code is verbose. We should figure out how to make custom hooks from custom hooks, and that'll clean it up.

Aesthetics can probably be improved, but that can be a follow-on item.

Contributes to #2109.

### DIFF
--- a/web-admin/src/components/navigation/Breadcrumbs.svelte
+++ b/web-admin/src/components/navigation/Breadcrumbs.svelte
@@ -3,13 +3,14 @@
   import { page } from "$app/stores";
   import {
     createRuntimeServiceListCatalogEntries,
-    V1CatalogEntry,
+    createRuntimeServiceListFiles,
   } from "@rilldata/web-common/runtime-client";
   import { runtime } from "@rilldata/web-common/runtime-client/runtime-store";
   import {
     createAdminServiceListOrganizations,
     createAdminServiceListProjectsForOrganization,
   } from "../../client";
+  import { getDashboardListItemsFromFilesAndCatalogEntries } from "../projects/dashboards";
   import BreadcrumbItem from "./BreadcrumbItem.svelte";
   import OrganizationAvatar from "./OrganizationAvatar.svelte";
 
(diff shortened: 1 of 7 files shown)


web-admin/src/routes/[organization]/[project]/[dashboard]/+page.svelte
```diff
@@ -42,21 +47,60 @@
...
+      type: "OBJECT_TYPE_METRICS_VIEW",
+    },
+    {
+      query: {
+        placeholderData: undefined,
+        enabled: !!project && !!$runtime?.instanceId,
+      },
+    }
+  );
+  $: dashboardListItems = getDashboardListItemsFromFilesAndCatalogEntries(
```
Comment (AdityaHegde) [inline web-admin/src/routes/[organization]/[project]/[dashboard]/+page.svelte:83]: Why do we need the full list here? Should be enough to just fetch a single file and single catalog entry and get the validity status from that.

web-admin/src/routes/[organization]/[project]/[dashboard]/+page.svelte
```diff
@@ -42,21 +47,60 @@
   }
 
   async function getDashboardsAndInvalidate() {
-    return invalidateProjectQueries(
-      queryClient,
-      await getDashboardsForProject($project.data)
-    );
+    const dashboardListings = await getDashboardsForProject($project.data);
+    const dashboardNames = dashboardListings.map((listing) => listing.name);
+    return invalidateProjectQueries(queryClient, dashboardNames);
```
Comment (AdityaHegde) [inline web-admin/src/routes/[organization]/[project]/[dashboard]/+page.svelte:52]: Since we now have the file and catalog entry we need to update `invalidateProjectQueries` to invalidate them.

web-admin/src/components/navigation/Breadcrumbs.svelte
```diff
@@ -21,24 +22,38 @@
...
-        select: (data) => {
-          currentDashboard = data?.entries?.find(
-            (entry) => entry.name === $page.params.dashboard
-          );
-          return data;
-        },
       },
     }
   );
+  $: dashboardListItems = getDashboardListItemsFromFilesAndCatalogEntries(
```
AdityaHegde, line 50: Same here. We just need a single catalog entry and file.

web-admin/src/routes/[organization]/[project]/[dashboard]/+page.svelte
Comment (AdityaHegde) [reply web-admin/src/routes/[organization]/[project]/[dashboard]/+page.svelte:83]: This can be fixed after the release. It is just optimization.

web-admin/src/components/navigation/Breadcrumbs.svelte
AdityaHegde: Can be taken up later

web-admin/src/routes/[organization]/[project]/[dashboard]/+page.svelte
Comment (AdityaHegde) [reply web-admin/src/routes/[organization]/[project]/[dashboard]/+page.svelte:52]: This is needed since there could be stale data @ericpgreen2

AdityaHegde (approved):

PR #2271, opened 2023-04-28
Fixing dimension table not navigatable

### LINKED ISSUE #2269: cannot access dimension detail table when clicking on leaderboard header
I went through a bunch of commits on main, and i think this bug was introduced in #2257 -- it seems like maybe it has to do with this chunk of new code in `web-common/src/features/dashboards/workspace/Dashboard.svelte` :

```
  $: metricsExplorer = useDashboardStore(metricViewName);

  $: stateSyncManager = new StateSyncManager(metricViewName);
  $: if ($metricsExplorer) {
    stateSyncManager.handleStateChange($metricsExplorer);
  }
  $: if ($page) {
    stateSyncManager.handleUrlChange();
  }
``` 

It looks like the `selectedDimensionName` variable is not being toggled, and therefore the dimension table component is never shown

AdityaHegde:
closes #2269

### DIFF
--- a/web-common/src/features/dashboards/workspace/Dashboard.svelte
+++ b/web-common/src/features/dashboards/workspace/Dashboard.svelte
@@ -55,7 +55,7 @@
     stateSyncManager.handleUrlChange();
   }
 
-  $: selectedDimensionName = metricsExplorer?.selectedDimensionName;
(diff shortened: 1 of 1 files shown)


nishantmonu51 (approved):

nishantmonu51:
One note - if possible, we should also aim to add test cases as well.

PR #2272, opened 2023-04-28
change role of invited user

pjain1:
Change role of invited user using `rill user set-role`

### DIFF
--- a/admin/database/postgres/postgres.go
+++ b/admin/database/postgres/postgres.go
@@ -817,6 +817,11 @@ func (c *connection) CountInvitesForOrganization(ctx context.Context, orgID stri
 	return count, nil
 }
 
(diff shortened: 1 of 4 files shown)


admin/database/postgres/postgres.go
```diff
@@ -817,6 +817,11 @@ func (c *connection) CountInvitesForOrganization(ctx context.Context, orgID stri
 	return count, nil
 }
 
+func (c *connection) UpdateOrganizationInviteRole(ctx context.Context, id, roleID string) error {
+	_, err := c.getDB(ctx).ExecContext(ctx, `UPDATE org_invites SET org_role_id = $1 WHERE id = $2`, roleID, id)
```
pjain1, line 821: not need to check rows affected I guess as we know invite id is correct.

PR #2275, opened 2023-04-28
Fixing home page list getting cut off

AdityaHegde:
Using justify-center can break scrolling and make some parts inaccessible while using flex. instead using margin auto to achieve similar behaviour.

Ref: https://bhch.github.io/posts/2021/04/centring-flex-items-and-allowing-overflow-scroll/


djbarnwal (approved):
Looks good, thanks

PR #2276, opened 2023-04-28
Reconcile deployment on project update and env changes

k-anshul:
- [x] Reconciles deployment on update on project options like branch, variables etc
- [x] Prompt user to reconcile deployment on env changes

### DIFF
--- a/admin/database/database.go
+++ b/admin/database/database.go
@@ -245,6 +245,7 @@ type UpdateProjectOptions struct {
 	ProdBranch           string
 	ProdVariables        map[string]string
 	ProdDeploymentID     *string
+	ReconcileDeployments bool
 }
 
 // DeploymentStatus is an enum representing the state of a deployment
--- a/admin/database/database.go
+++ b/admin/database/database.go
@@ -245,7 +245,6 @@ type UpdateProjectOptions struct {
 	ProdBranch           string
 	ProdVariables        map[string]string
 	ProdDeploymentID     *string
-	ReconcileDeployments bool
 }
 
 // DeploymentStatus is an enum representing the state of a deployment
--- a/admin/projects.go
+++ b/admin/projects.go
@@ -122,7 +122,7 @@ func (s *Service) TeardownProject(ctx context.Context, p *database.Project) erro
 }
 
 // UpdateProject updates a project and any impacted deployments.
-// It does not run a reconcile, even if deployment parameters (like branch or variables) have been changed.
+// It run a reconcile if deployment parameters (like branch or variables) have been changed and ReconcileDeployments option is set.
 func (s *Service) UpdateProject(ctx context.Context, proj *database.Project, opts *database.UpdateProjectOptions) (*database.Project, error) {
 	impactsDeployments := (proj.ProdBranch != opts.ProdBranch ||
 		!reflect.DeepEqual(proj.GithubURL, opts.GithubURL) ||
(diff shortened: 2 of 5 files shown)


k-anshul:
@begelundmuller Should we even propagate the envs to runtime instance if user choose not to reconcile ? May be a good behaviour is to ask users if they want to reset deployment on env change and reset the deployment and not reconcile.

begelundmuller:
> @begelundmuller Should we even propagate the envs to runtime instance if user choose not to reconcile ? May be a good behaviour is to ask users if they want to reset deployment on env change and propagate the env changes only on reset(redeploy).

I think we shouldn't overthink this – if a user updates variables, we should update them on the relevant deployments. Updating certain variables will have a chance of inconsistency, but I think the user would be aware of that.

Giving users the option to not reconcile/reset when changing variables gives them some flexibility, e.g. if they are updating variables just before pushing a change that depends on them.

admin/database/database.go
```diff
@@ -245,6 +245,7 @@ type UpdateProjectOptions struct {
 	ProdBranch           string
 	ProdVariables        map[string]string
 	ProdDeploymentID     *string
+	ReconcileDeployments bool
```
begelundmuller, line 248: This doesn't belong in the database options struct – not used in database call.

admin/projects.go
```diff
@@ -148,6 +148,12 @@ func (s *Service) UpdateProject(ctx context.Context, proj *database.Project, opt
 				// TODO: This may leave things in an inconsistent state. (Although presently, there's almost never multiple deployments.)
 				return nil, err
 			}
+			if opts.ReconcileDeployments {
+				if err := s.triggerReconcile(ctx, d); err != nil {
+					// TODO :: check if trigger reconcile error should fail project update
+					s.logger.Error("failed to trigger reconcile", zap.String("deployment_id", d.ID), zap.String("project_id", proj.ID))
+				}
+			}
```
begelundmuller, line 156: This should be moved into `updateDeployment`

cli/cmd/env/configure.go
```diff
@@ -105,15 +106,33 @@ func ConfigureCmd(cfg *config.Config) *cobra.Command {
...
+				cmdutil.SuccessPrinter("Triggered reconcile successfully.")
+			}
 			return nil
 		},
 	}
 
 	configureCommand.Flags().SortFlags = false
 	configureCommand.Flags().StringVar(&projectPath, "path", ".", "Project directory")
 	configureCommand.Flags().StringVar(&projectName, "project", "", "")
+	configureCommand.Flags().BoolVar(&triggerReconcile, "reconcile", true, "Reconcile deployment")
```
begelundmuller, line 135: Since it's going to prompt, I think putting default `false` here makes more sense

cli/cmd/env/configure.go
```diff
@@ -105,15 +106,33 @@ func ConfigureCmd(cfg *config.Config) *cobra.Command {
...
+					return err
+				}
+
+				_, err = client.TriggerReconcile(ctx, &adminv1.TriggerReconcileRequest{DeploymentId: project.ProdDeployment.Id})
+				if err != nil {
+					warn.Printf("Reconcile failed. Trigger reconcile again with `rill project reconcile` if required.")
+					return err
+				}
+				cmdutil.SuccessPrinter("Triggered reconcile successfully.")
+			}
```
begelundmuller, line 127: I'm not sure triggering reconcile makes sense. Reconcile will do nothing if sources are already ingested and models deployed, because the variables are not part of the dependency DAG (we should change that in the next reconcile).  You would need to use `TriggerRefreshSources` to re-ingest sources or use `TriggerRedeploy` to re-build the whole project.

We should either give people the option to chose between a) refresh sources or b) trigger redeploy. Or just ask people if they want to redeploy or not.

begelundmuller (changes requested):

admin/database/database.go
k-anshul: This problem arises bcz of choice of keeping `database.UpdateProjectOptions` in service methods. Ideally service should take its own options where `database.UpdateProjectOptions` is one field. For now I will add `ReconcileDeployments` as additional method argument.

cli/cmd/env/configure.go
k-anshul: I was wrongly under the impression that we look at the file content to check if artifact has changed but looks like we are only looking at the file timestamps so yeah in this case `reconcile` will make no sense. I will change to always `redeploy`.

begelundmuller (approved):

PR #2278, opened 2023-04-28
Format logs page

djbarnwal:
- Format logs page to match the mocks
- Add a copy for non admin roles

Couple of questions @ericpgreen2 
- In the mocks it says the copy for viewer should be - "This project has 11 errors. Ask your organization’s admin for help resolving these." But the `logs` key is empty for the viewer. Is there a may to retrieve the number of log errors?

- For checking if the user has permission to view logs I am using the `manageProject` permission, although I saw that there were various other permissions as well. Not sure if `manageProject` is the right one or I should use `readProdStatus` here.

### DIFF
--- a/web-admin/src/components/projects/ProjectDeploymentLogs.svelte
+++ b/web-admin/src/components/projects/ProjectDeploymentLogs.svelte
@@ -5,6 +5,8 @@
   export let project: string;
 
   $: proj = createAdminServiceGetProject(organization, project);
+  $: hasAdminAccess = $proj.data?.projectPermissions?.manageProject;
+
   $: errors = parseLogs($proj.data?.prodDeployment?.logs);
 
   interface Error {
@@ -23,7 +25,11 @@
 
 {#if $proj.isSuccess && errors}
   <ul class="w-full">
-    {#if errors.length === 0}
+    {#if !hasAdminAccess}
+      <li class="px-12 py-2 font-semibold text-gray-500 border-b">
+        You don't have permission to view project logs
+      </li>
+    {:else if errors.length === 0}
       <li class="px-12 py-2 font-semibold text-gray-500 border-b">
         No logs present
       </li>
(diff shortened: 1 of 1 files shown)


web-admin/src/components/projects/ProjectDeploymentLogs.svelte
```diff
@@ -22,16 +24,22 @@
...
-    <ul class="w-full">
+  <ul class="w-full">
+    {#if !hasAdminAccess}
+      <li class="px-12 py-2 font-semibold text-gray-500 border-b">
+        You don't have permission to view project logs
+      </li>
+    {:else if errors.length === 0}
+      <li class="px-12 py-2 font-semibold text-gray-500 border-b">
+        No logs present
+      </li>
```
ericpgreen2, line 35: These aren't really "list items," so I'd move this code up above the `<ul>` tag.

ericpgreen2:
Good questions.

> * In the mocks it says the copy for viewer should be - "This project has 11 errors. Ask your organization’s admin for help resolving these." But the `logs` key is empty for the viewer. Is there a may to retrieve the number of log errors?

Hmm, the mocks are off here. There's no other API to retrieve the number of log errors, and anyways, given the logs are private, it would be mysterious to expose that number to the viewer. Let's go with "This project has errors" omitting the number for now. Then we can circle back to double check with Janet.

> * For checking if the user has permission to view logs I am using the `manageProject` permission, although I saw that there were various other permissions as well. Not sure if `manageProject` is the right one or I should use `readProdStatus` here.

`readProdStatus` is the right one. Benjamin actually mentioned to me that that property might be renamed to `readProdLogs` in the future to be more descriptive.

ericpgreen2 (approved):
Approving so you can merge after addressing the comments

web-admin/src/components/projects/ProjectDeploymentLogs.svelte
djbarnwal: The mocks make it look like a list with a single element so I would prefer to have it this way. Also it makes the component structure much simpler.

PR #2280, opened 2023-04-28
Use CTA components for ProjectBuilding

djbarnwal:
Use CTA components for ProjectBuilding


ericpgreen2 (approved):
FYI I made a couple styling edits that I discussed with Janet, and I tweaked a couple of the prop semantics.

PR #2283, opened 2023-04-28
Adding empty url for default dashboard

AdityaHegde:
If starting from a blank filters in a dashboard the default state will not populate the url state.

But if starting from a url then we cannot get the default state right now. We need to do a refactor of time controls to get the default state in one place and address this in a later PR.

Further changes handled in #2285


djbarnwal (approved):
Looks good for a stopgap solution. As you mentioned, would be good to have the defaults defined properly. Can you create a issue for the same and link this PR?

PR #2287, opened 2023-04-29
Fix homepage layout

ericpgreen2:
This PR:
- Ensures the "Hi, {user}!" text is consistently placed at the top of the screen
- Brings the Share CTA closer to the middle of the screen horizontally
- Spaces out the list of organizations and projects to match mocks

Contributes to #2109.


web-admin/src/routes/+page.svelte
```diff
@@ -22,26 +22,24 @@
 </svelte:head>
 
 <AuthRedirect>
-  <section class="flex flex-col w-4/5 mx-auto h-2/5">
-    <h1 class="text-4xl leading-10 font-light mb-2 my-auto">
+  <section class="flex flex-col w-4/5 mx-auto h-5/6 my-36">
```
hamilton, line 25: It appears this use of `h-5/6` may be creating a scrollbar. It'd be a good idea to _not_ set the height like this; it's pretty unstable.

Kapture 2023-05-03 at 20 35 41

hamilton: additionally, it would be a good idea to add some gap between the two columns, since in slightly smaller screens the left column looks too close to the right column.

hamilton (changes requested):
Thanks for the PR. I got caught trying to re-deploy a project, and that's not working. But here are a few changes needed based on what i see so far:
- the body is too big on my laptop, so I have to scroll even if not required. Would suggest not setting the height the way you are; tends to be pretty brittle.
- I see "Hi Hamilton Ulmer!" I believe the figma mock only uses the first name. Can we fix that in this PR?

web-admin/src/routes/+page.svelte
ericpgreen2: Yeah good call – just went with a responsive margin approach

skokenes (approved):
LGTM!

A couple of design things I noticed that we may want to revisit later
- The columns collapse into rows around ~1350px, which does not seem very narrow
landing-flex
- pretty thin margin between two columns when the project title is long
Screenshot 2023-05-08 at 2 38 34 PM

ericpgreen2:
> A couple of design things I noticed that we may want to revisit later
> * The columns collapse into rows around ~1350px, which does not seem very narrow
> * pretty thin margin between two columns when the project title is long

Good calls. I increased the margin now. I'll leave the breakpoint to be revisited later.

PR #2289, opened 2023-05-01
common error handling fix

k-anshul:
`status.FromError(err)` doesn't work for wrapped errors. We are wrapping rpc errors in lot of places.

### DIFF
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -2,9 +2,9 @@ package cmd
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/rilldata/rill/cli/cmd/admin"
 	"github.com/rilldata/rill/cli/cmd/auth"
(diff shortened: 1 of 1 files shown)


cli/cmd/root.go
```diff
@@ -59,8 +60,9 @@ func Execute(ctx context.Context, ver config.Version) {
 				fmt.Println("Please log in or sign up for Rill with `rill login`.")
 			default:
 				// no known message
+				// printing the full error message so that wrapped context is not lost
 				// todo :: add trace id as well
-				fmt.Printf("Error: %s (%v)\n", s.Message(), s.Code())
+				fmt.Printf("Error: %s (%v)\n", err, s.Code())
```
begelundmuller, line 65: This will cause the error formatting to be the ugly format again when `err` is an RPC error

begelundmuller (changes requested):
Consider just checking the message cases using `strings.Contains` instead of unwrapping

cli/cmd/root.go
```diff
@@ -45,25 +46,23 @@ var rootCmd = &cobra.Command{
...
+			fmt.Println("Auth token is invalid/expired. Run `rill logout` and login again with `rill login`.")
+		} else if strings.Contains(errMsg, "not authenticated as a user") {
+			fmt.Println("Please log in or sign up for Rill with `rill login`.")
+		} else {
+			if s, ok := status.FromError(err); ok {
+				// rpc error
 				fmt.Printf("Error: %s (%v)\n", s.Message(), s.Code())
+			} else {
+				fmt.Printf("Error: %s\n", err.Error())
 			}
```
begelundmuller, line 65: Could flatten the nested `if/else`

begelundmuller (approved):

cli/cmd/root.go
k-anshul: Keeping it separate shows this is a separate case of `unknown` messages. Flattening it would mean that `if/else-if` blocks handle different cases. 
Don't have strong opinion but prefer this one.

PR #2292, opened 2023-05-01
Upgrade Go dependencies

begelundmuller:


runtime/pkg/observability/middleware.go
```diff
@@ -62,16 +64,47 @@ func LoggingUnaryServerInterceptor(logger *zap.Logger) grpc.UnaryServerIntercept
...
+	})
+}
+
+// tracingFieldsFromCtx picks tracing-related fields from the ctx for use in the gRPC logging interceptor.
+func tracingFieldsFromCtx(ctx context.Context) logging.Fields {
+	sctx := trace.SpanFromContext(ctx).SpanContext()
+	if !sctx.IsValid() {
+		return nil
+	}
+	return []any{"trace_id", sctx.TraceID().String()}
```
k-anshul, line 107: Do we also want `span_id` ?

k-anshul (approved):

runtime/pkg/observability/middleware.go
begelundmuller: I also initially thought so, but the examples for otel grpc did not have it (only `trace_id`). @kaspersjo, what do you think?

PR #2293, opened 2023-05-01
Fix 400 error in dashboards without a time series

ericpgreen2:
This PR fixes the 400 error that's triggered when a dashboard calls the Query Service's `time-range-summary` endpoint when a time series is *not* present.

Because fetching and constructing a model's "all time range" is a common operation, we've created a custom selector hook `useModelAllTimeRange` to contain the transformation logic. To this custom selector hook, I've added the ability to pass TanStack query options (such as the `enabled` flag for dependent queries). I've mirrored the pattern that Orval uses in its generated hooks.

Contributes to #2109.

### DIFF
--- a/web-common/src/features/dashboards/selectors.ts
+++ b/web-common/src/features/dashboards/selectors.ts
@@ -1,14 +1,17 @@
 import { TimeRangePreset } from "@rilldata/web-common/lib/time/types";
 import {
   RpcStatus,
+  V1MetricsView,
+  V1MetricsViewFilter,
   createQueryServiceColumnTimeRange,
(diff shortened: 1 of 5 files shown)


web-common/src/features/dashboards/selectors.ts
```diff
@@ -139,8 +142,13 @@ export const useModelHasTimeSeries = (
 export function useModelAllTimeRange(
```
AdityaHegde, line 142: Since this is a custom hook we can just pass the enabled in here than repeating in all components

AdityaHegde (changes requested):

AdityaHegde (approved):

PR #2299, opened 2023-05-02
version check changes

rakeshsharma14317:


cli/pkg/update/update.go
```diff
@@ -64,12 +65,12 @@ func LatestVersion(ctx context.Context) (string, error) {
 			return "", err
 		}
 
-		updatedAt, err := time.Parse("2006-01-02 15:04", cachedVersionUpdatedAt)
+		updatedAt, err := time.Parse(cmdutil.TSFormatLayout, cachedVersionUpdatedAt)
```
begelundmuller, line 68: This should not use the user-facing timestamp format, it's for internal persistance. Use `time.RFC3339` instead

cli/pkg/update/update.go
```diff
@@ -64,12 +65,12 @@ func LatestVersion(ctx context.Context) (string, error) {
...
 		if err != nil {
 			return "", err
 		}
 
-		if time.Since(updatedAt).Hours() < versionCheckTTL.Hours() {
+		if time.Since(updatedAt.UTC()).Hours() < versionCheckTTL.Hours() {
```
begelundmuller, line 73: 1. `UTC` will not be needed here when using `time.RFC3339`
2. Don't do `Hours()`, just compare the time deltas

cli/pkg/update/update.go
```diff
@@ -80,7 +81,7 @@ func LatestVersion(ctx context.Context) (string, error) {
 		return "", err
 	}
 
-	err = dotrill.SetVersionUpdatedAt(time.Now().Format("2006-01-02 15:04"))
+	err = dotrill.SetVersionUpdatedAt(time.Now().UTC().Format(cmdutil.TSFormatLayout))
```
begelundmuller, line 84: Don't do `UTC` and use `time.RFC3339`

cli/cmd/root.go
```diff
@@ -87,7 +87,7 @@ func runCmd(ctx context.Context, ver config.Version) error {
 	// Check version
 	err := update.CheckVersion(ctx, cfg.Version.Number)
 	if err != nil {
-		return err
+		fmt.Printf("check version failed, Error: %v\n", err)
```
begelundmuller, line 90: Make it `Warning: version check failed: %v`

begelundmuller (changes requested):

cli/pkg/update/update.go
rakeshsharma14317: Thanks, i was checking for some better approach for this, But yeah this looks good for internal persistance :)

begelundmuller (approved):

PR #2305, opened 2023-05-02
Fixing valid dashboard refresh

AdityaHegde:
We had a reactive statement dependant on `$runtime`. Since it changes every time jwt changes this lead to constant refresh of the dashboard.

### DIFF
--- a/web-common/src/runtime-client/RuntimeProvider.svelte
+++ b/web-common/src/runtime-client/RuntimeProvider.svelte
@@ -16,7 +16,7 @@
   // Re-run all runtime queries when `host` changes
   // By default, a new `instanceId` triggers a re-run because it's in the queryKeys
   const queryClient = useQueryClient();
-  $: $runtime.host && invalidateRuntimeQueries(queryClient);
+  $: host && invalidateRuntimeQueries(queryClient);
 </script>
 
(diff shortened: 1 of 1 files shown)


AdityaHegde:
All i am doing is not depending on the whole of `$runtime` here. This is essentially memoisation of host. The `runtime.set` is getting called since jwt refreshes. Which changes the contents of `$runtime`. Which triggers the reactive statement that refetches the queries.

Not sure I understand your concern here.

AdityaHegde:
Ah its the race condition that is the problem. Lemme check

ericpgreen2:
To go with your approach, but still avoid the race condition, maybe a `derived` store for the runtime host would work.

ericpgreen2:
Alternatively, I _think_ making the `useProject` hook that's used on the dashboard page into a `useProjectDeploymentStatus` custom hook (w/ a selector for the status) would work & is a more correct use of TanStack.

AdityaHegde:
Changing `useProject` should be in a different PR. It is unrelated. I have just removed the reactive statement. This assumed instanceId always changes when the deployment for a project changes.

ericpgreen2 (approved):

PR #2306, opened 2023-05-02
Avoiding refetching catalogs while reconciling

AdityaHegde:
While reconciling we unnecessarily make calls to fetch files and catalogs. Adding the status check to `enabled` and moving the logic to a derived store.

Needs changes from #2305 to fully work.


web-admin/src/components/home/DashboardList.svelte
```diff
@@ -16,6 +16,13 @@
   }
 
   async function updateDashboardsForProject() {
```
ericpgreen2, line 18: It'd be better if this function took `projectData` as a parameter, rather than closing over the global value.

web-admin/src/components/home/DashboardList.svelte
```diff
@@ -16,6 +16,13 @@
...
+    const status = $proj.data?.prodDeployment.status;
+    if (
+      status === "DEPLOYMENT_STATUS_PENDING" ||
```
ericpgreen2, line 21: These shouldn't be hardcoded strings, but rather e.g. `V1DeploymentStatus.DEPLOYMENT_STATUS_RECONCILING`

web-admin/src/components/navigation/Breadcrumbs.svelte
```diff
@@ -29,7 +25,7 @@
   $: isOrganizationPage = $page.route.id === "/[organization]";
 
   $: projectName = $page.params.project;
-  $: project = createAdminServiceGetProject(orgName, projectName);
+  $: project = useProject(orgName, projectName);
```
ericpgreen2, line 28: This is a good example why I think `useProject` should be renamed to `useProjectDeploymentStatus`. The name `useProject` on its own doesn't connote how it's different from the vanilla call to `createAdminServiceGetProject` that was here previously.

web-admin/src/components/navigation/Breadcrumbs.svelte
```diff
@@ -43,34 +39,8 @@
 
   // Here, we compose the dashboard list via two separate runtime queries.
   // We should create a custom hook to hide this complexity.
```
ericpgreen2, line 41: We can remove this comment now. I like the custom hook!

web-admin/src/components/projects/dashboards.ts
```diff
@@ -72,3 +78,66 @@ export function getDashboardListItemsFromFilesAndCatalogEntries(
...
+): Readable<{
+  items: DashboardListItem[];
+  success: boolean;
+}> {
+  let isProfiling = false;
+  if (project) {
+    const status = get(project)?.data?.prodDeployment?.status;
+    if (
+      status === "DEPLOYMENT_STATUS_PENDING" ||
+      status === "DEPLOYMENT_STATUS_RECONCILING"
```
ericpgreen2, line 94: Again, should be e.g. `V1DeploymentStatus.DEPLOYMENT_STATUS_RECONCILING` rather than hardcoded strings.

web-admin/src/components/projects/dashboards.ts
```diff
@@ -72,3 +78,66 @@ export function getDashboardListItemsFromFilesAndCatalogEntries(
 
   return dashboardListings;
 }
+
+export function useDashboardListItems(
+  instanceId: string,
+  project: CreateQueryResult<V1GetProjectResponse>
```
ericpgreen2, line 84: Can this be more specifically `projectStatus`?

web-admin/src/components/projects/dashboards.ts
```diff
@@ -72,3 +78,66 @@ export function getDashboardListItemsFromFilesAndCatalogEntries(
...
+            placeholderData: undefined,
+            enabled: !isProfiling && !!project && !!instanceId,
+          },
+        }
+      ),
+    ],
+    ([dashboardFiles, dashboardCatalogEntries]) => {
+      if (!dashboardFiles.isSuccess || !dashboardCatalogEntries.isSuccess)
+        return {
+          success: false,
```
ericpgreen2, line 130: It'd be good to mirror the TanStack return types as much as possible. So `isSuccess` rather than `success`.

web-admin/src/routes/[organization]/[project]/[dashboard]/+page.svelte
```diff
@@ -72,39 +70,10 @@
   // As is the case in `Breadcrumbs.svelte`, there are two queries we compose to get the dashboard list items,
   // and we should hide this complexity in a custom hook.
```
Comment (ericpgreen2) [inline web-admin/src/routes/[organization]/[project]/[dashboard]/+page.svelte:71]: We can remove this comment too. (But not the next line that concerns the 404)

ericpgreen2:
Every time I navigate to the home screen, I see two canceled network requests: one to list files, and one to the catalog. Can we avoid those?

web-admin/src/components/navigation/Breadcrumbs.svelte
AdityaHegde: Lets hold on this for now. I think there are a lot of these that we have punted on.

web-admin/src/components/navigation/Breadcrumbs.svelte
ericpgreen2: Comment hasn't been removed

ericpgreen2:
Approving so you can merge after fixing: `svelte-check` issue, `Breadcrumbs.svelte` comment still present, typo

ericpgreen2 (approved):

PR #2307, opened 2023-05-02
update updated on for user role update

pjain1:


begelundmuller (commented):
I'm not sure it makes sense to update the `users` table when an org gets a new member. This leads to arbitrary transitive rules – like for example, when there's a change to a deployment, should it also update the project and org timestamps?

I think it makes sense to keep `updated_on` changes to only the table in question. In this case, it would mean tracking timestamp on the `users_orgs_roles` and `users_projects_roles` tables.

But since we don't have such fields now, this seems less like an issue. Feel free to not do it at this stage. (I thought we had such fields, but they were not updated – which I see is the case for other `UPDATE` statements, which we should address).

pjain1:
As of now `SetOrganizationMemberRole` is only called when we change role of an existing user from cli not sure what you mean by 
> I'm not sure it makes sense to update the users table when an org gets a new member.

And as per our discussion and issue `Admin: When changing a user's role, update updated_at in the many-to-many table` we wanted to update this on role change ?

PR #2312, opened 2023-05-02
Toplist comparison

egor-ryashin:


proto/rill/runtime/v1/queries.proto
```diff
@@ -173,6 +180,58 @@ message MetricsViewToplistResponse {
...
+message TimeRange {
+  google.protobuf.Timestamp start = 1;
+  google.protobuf.Timestamp end = 2;
+}
+
+enum ComparisonSortType {
+  COMPARISON_SORT_TYPE_UNSPECIFIED = 0;
+  COMPARISON_SORT_TYPE_COMPARISON_VALUE = 1;
+  COMPARISON_SORT_TYPE_BASE_VALUE = 2;
+  COMPARISON_SORT_TYPE_DELTA = 3;
```
begelundmuller, line 212: 1. nit: put base value before comparison value
2. Should support both `delta_rel` and `delta_abs`?

runtime/queries/metricsview_comparision_toplist.go
```diff
@@ -0,0 +1,280 @@
...
+
+func (q *MetricsViewComparisonToplist) Resolve(ctx context.Context, rt *runtime.Runtime, instanceID string, priority int) error {
+	olap, err := rt.OLAP(ctx, instanceID)
+	if err != nil {
+		return err
+	}
+
+	if olap.Dialect() != drivers.DialectDuckDB {
+		return fmt.Errorf("not available for dialect '%s'", olap.Dialect())
+	}
```
begelundmuller, line 67: Should also support Druid like in `MetricsViewToplist`:
```go
	if olap.Dialect() != drivers.DialectDuckDB && olap.Dialect() != drivers.DialectDruid {
		return fmt.Errorf("not available for dialect '%s'", olap.Dialect())
	}
```

runtime/queries/metricsview_comparision_toplist.go
```diff
@@ -0,0 +1,280 @@
...
+		return err
+	}
+
+	if mv.TimeDimension == "" {
+		return fmt.Errorf("metrics view '%s' does not have a time dimension", q.MetricsViewName)
+	}
+
+	if q.BaseTimeStart == nil || q.BaseTimeEnd == nil || q.ComparisonTimeStart == nil || q.ComparisonTimeEnd == nil {
+		return fmt.Errorf("undefined time range for comparison on '%s' metrics view ", q.MetricsViewName)
+	}
```
begelundmuller, line 80: Can we make it work like the normal toplist query if the comparison time range is empty? I.e. just return 0 for all the comparison values if no comparison time range is specified?

runtime/server/queries_metrics.go
```diff
@@ -36,6 +36,33 @@ func (s *Server) MetricsViewToplist(ctx context.Context, req *runtimev1.MetricsV
...
+	}
+
+	q := &queries.MetricsViewComparisonToplist{
+		MetricsViewName:     req.MetricsViewName,
+		DimensionName:       req.DimensionName,
+		MeasureNames:        req.MeasureNames,
+		BaseTimeStart:       req.BaseTimeRange.Start,
+		BaseTimeEnd:         req.BaseTimeRange.End,
+		ComparisonTimeStart: req.ComparisonTimeRange.Start,
+		ComparisonTimeEnd:   req.ComparisonTimeRange.End,
```
begelundmuller, line 52: Will panic if `req.BaseTimeRange` or `req.ComparisonTimeRange` is `nil`

runtime/queries/metricsview_comparision_toplist.go
```diff
@@ -0,0 +1,280 @@
...
+		case runtimev1.ComparisonSortType_COMPARISON_SORT_TYPE_DELTA:
+			pos = 4 + i*4
+		default:
+			return "", nil, fmt.Errorf("undefined sort type for measure %s", s.MeasureName)
+		}
+		orderClause += fmt.Sprint(pos)
+		if !s.Ascending {
+			orderClause += " DESC"
+		}
+		orderClause += " NULLS LAST"
```
begelundmuller, line 246: Need a conditional to only add this for DuckDB, it's not supported by Druid (see old toplist query)

runtime/queries/metricsview_comparision_toplist.go
```diff
@@ -0,0 +1,280 @@
...
+		measureMap[n] = i
+		found := false
+		for _, m := range mv.Measures {
+			if m.Name == n {
+				expr := fmt.Sprintf(`%s as "%s"`, m.Expression, m.Name)
+				selectCols = append(selectCols, expr)
+				finalSelectCols = append(
+					finalSelectCols,
+					fmt.Sprintf(
+						"base.%[1]s, comparison.%[1]s, comparison.%[1]s - base.%[1]s, (comparison.%[1]s - base.%[1]s)/base.%[1]s::DOUBLE",
```
begelundmuller, line 180: I think the subtraction, division and cast to `::DOUBLE` will fail for measures that produce non-numeric values (suppose someone, weirdly, has a measure like `ANY_VALUE(name)`). Maybe also for `HUGEINT` or `DECIMAL` in some cases?

We need a way to protect against that case – any ideas? I guess it's preferable to guard against it in-query vs. running an extra query to pre-check measure result types.

begelundmuller (changes requested):

proto/rill/runtime/v1/queries.proto
egor-ryashin: 2. Ordering of abs delta and relative delta is always the same

runtime/queries/metricsview_comparision_toplist.go
egor-ryashin: We cannot do delta on non-numeric measures, I think we should report it's not supported.

egor-ryashin: Timeseries also fail for `ANY(name)`

runtime/queries/metricsview_comparision_toplist.go
```diff
@@ -0,0 +1,413 @@
...
+	if err != nil {
+		return "", nil, err
+	}
+
+	dimName := safeName(q.DimensionName)
+	selectCols := []string{dimName}
+	measureMap := make(map[string]int)
+	for i, m := range ms {
+		measureMap[m.Name] = i
+		expr := fmt.Sprintf(`%s as "%s"`, m.Expression, m.Name)
```
begelundmuller, line 241: Should we use `safeName` for `m.Name` instead?

egor-ryashin: No need, `m.Name` is generated automatically (not a user input).

egor-ryashin: No need, `m.Name` is generated automatically (not a user input).

PR #2320, opened 2023-05-03
Moving validation of time controls to store

AdityaHegde:
- [x] Move validation of time ranges from TimeControl to dashboard store and to other files.
- [ ] Move setting of default time range and time range from url outside of TimeControl
- [ ] Add tests to this code since they are now outside of svelte component.


AdityaHegde:
Closing in favour of #2760

PR #2324, opened 2023-05-03
Redirect project breadcrumb selection to project's first dashboard

ericpgreen2:
Contributes to #2109.

### DIFF
--- a/web-admin/src/routes/[organization]/[project]/-/redirect/+page.svelte
+++ b/web-admin/src/routes/[organization]/[project]/-/redirect/+page.svelte
@@ -18,7 +18,7 @@
     }
   );
 
-  // Avoid a race condition: ensure the runtime store has been updated with the new project's runtime instance ID
+  // Avoid a race condition: make sure the runtime store has been updated (with the host, instanceID, and jwt).
   $: isRuntimeStoreReady =
     $proj?.data &&
     $proj.data.prodDeployment.runtimeInstanceId === $runtime.instanceId;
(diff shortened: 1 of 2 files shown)


web-admin/src/routes/[organization]/[project]/-/redirect/+page.svelte
```diff
@@ -1,13 +1,33 @@
...
+  );
+
+  // Avoid a race condition: ensure the runtime store has been updated with the new project's runtime instance ID
+  $: isRuntimeStoreReady =
+    $proj?.data &&
+    $proj.data.prodDeployment.runtimeInstanceId === $runtime.instanceId;
+
+  let dashboardsQuery;
+  $: if (isRuntimeStoreReady) {
+    dashboardsQuery = useDashboardNames($runtime.instanceId);
```
Comment (AdityaHegde) [inline web-admin/src/routes/[organization]/[project]/-/redirect/+page.svelte:28]: Why not use `$proj.data.prodDeployment.runtimeInstanceId` here instead of dealing with race condition?

Comment (ericpgreen2) [reply web-admin/src/routes/[organization]/[project]/-/redirect/+page.svelte:28]: Because `useDashboardNames`'s `httpClient` needs `$runtime.host` and `$runtime.jwt`. If we use `$proj.data.prodDeployment.runtimeInstanceId`, then the `httpClient` might reference `$runtime` before the new `host` and `jwt` are set.

Comment (ericpgreen2) [reply web-admin/src/routes/[organization]/[project]/-/redirect/+page.svelte:28]: Just tried to clarify the comment a bit

AdityaHegde (approved):

PR #2326, opened 2023-05-04
pagination in list apis

k-anshul:
Closes https://github.com/rilldata/rill-developer/issues/2302

### DIFF
--- a/admin/provisioner/provisioner_test.go
+++ b/admin/provisioner/provisioner_test.go
@@ -1,180 +1,182 @@
 package provisioner
 
-import (
-	"context"
-	"fmt"
-	"testing"
+// NOTE: Uncommenting temporarily as it appears to cause conflicts of multiple Postgres containers running at the same time.
 
-	"github.com/c2h5oh/datasize"
-	"github.com/google/uuid"
-	"github.com/rilldata/rill/admin/database"
-	"github.com/stretchr/testify/require"
-	"github.com/testcontainers/testcontainers-go"
-	"github.com/testcontainers/testcontainers-go/wait"
+// import (
+// 	"context"
+// 	"fmt"
+// 	"testing"
 
-	_ "github.com/rilldata/rill/admin/database/postgres"
-)
+// 	"github.com/c2h5oh/datasize"
+// 	"github.com/google/uuid"
+// 	"github.com/rilldata/rill/admin/database"
+// 	"github.com/stretchr/testify/require"
+// 	"github.com/testcontainers/testcontainers-go"
+// 	"github.com/testcontainers/testcontainers-go/wait"
 
-func Test_staticProvisioner_Provision(t *testing.T) {
-	ctx := context.Background()
-	container, err := testcontainers.GenericContainer(ctx, testcontainers.GenericContainerRequest{
-		Started: true,
-		ContainerRequest: testcontainers.ContainerRequest{
-			Image:        "postgres:14",
-			ExposedPorts: []string{"5432/tcp"},
-			WaitingFor:   wait.ForListeningPort("5432/tcp"),
-			Env: map[string]string{
-				"POSTGRES_USER":     "postgres",
-				"POSTGRES_PASSWORD": "postgres",
-				"POSTGRES_DB":       "postgres",
-			},
-		},
-	})
-	require.NoError(t, err)
-	defer container.Terminate(ctx)
+// 	_ "github.com/rilldata/rill/admin/database/postgres"
+// )
 
-	host, err := container.Host(ctx)
-	require.NoError(t, err)
-	port, err := container.MappedPort(ctx, "5432/tcp")
-	require.NoError(t, err)
-	databaseURL := fmt.Sprintf("postgres://postgres:postgres@%s:%d/postgres", host, port.Int())
+// func Test_staticProvisioner_Provision(t *testing.T) {
+// 	ctx := context.Background()
+// 	container, err := testcontainers.GenericContainer(ctx, testcontainers.GenericContainerRequest{
+// 		Started: true,
+// 		ContainerRequest: testcontainers.ContainerRequest{
+// 			Image:        "postgres:14",
+// 			ExposedPorts: []string{"5432/tcp"},
+// 			WaitingFor:   wait.ForListeningPort("5432/tcp"),
+// 			Env: map[string]string{
+// 				"POSTGRES_USER":     "postgres",
+// 				"POSTGRES_PASSWORD": "postgres",
+// 				"POSTGRES_DB":       "postgres",
+// 			},
+// 		},
+// 	})
+// 	require.NoError(t, err)
+// 	defer container.Terminate(ctx)
 
-	db, err := database.Open("postgres", databaseURL)
-	require.NoError(t, err)
-	require.NotNil(t, db)
-	defer db.Close()
+// 	host, err := container.Host(ctx)
+// 	require.NoError(t, err)
+// 	port, err := container.MappedPort(ctx, "5432/tcp")
+// 	require.NoError(t, err)
+// 	databaseURL := fmt.Sprintf("postgres://postgres:postgres@%s:%d/postgres", host, port.Int())
 
-	require.NoError(t, db.Migrate(ctx))
+// 	db, err := database.Open("postgres", databaseURL)
+// 	require.NoError(t, err)
+// 	require.NotNil(t, db)
+// 	defer db.Close()
 
-	org, err := db.InsertOrganization(ctx, &database.InsertOrganizationOptions{
-		Name: "test",
-	})
-	require.NoError(t, err)
+// 	require.NoError(t, db.Migrate(ctx))
 
-	p1, err := db.InsertProject(ctx, &database.InsertProjectOptions{
-		OrganizationID: org.ID,
-		Name:           "p-q",
-		Region:         "us-east-1",
-		ProdBranch:     "main",
-		ProdSlots:      1,
-	})
-	require.NoError(t, err)
+// 	org, err := db.InsertOrganization(ctx, &database.InsertOrganizationOptions{
+// 		Name: "test",
+// 	})
+// 	require.NoError(t, err)
 
-	// insert data
-	_, err = db.InsertDeployment(ctx, &database.InsertDeploymentOptions{
-		ProjectID:         p1.ID,
-		Slots:             2,
-		Branch:            "main",
-		RuntimeHost:       "host_1",
-		RuntimeInstanceID: uuid.NewString(),
-	})
+// 	p1, err := db.InsertProject(ctx, &database.InsertProjectOptions{
+// 		OrganizationID: org.ID,
+// 		Name:           "p-q",
+// 		Region:         "us-east-1",
+// 		ProdBranch:     "main",
+// 		ProdSlots:      1,
+// 	})
+// 	require.NoError(t, err)
 
-	require.NoError(t, err)
-	_, err = db.InsertDeployment(ctx, &database.InsertDeploymentOptions{
-		ProjectID:         p1.ID,
-		Slots:             5,
-		Branch:            "main",
-		RuntimeHost:       "host_2",
-		RuntimeInstanceID: uuid.NewString(),
-	})
-	require.NoError(t, err)
+// 	// insert data
+// 	_, err = db.InsertDeployment(ctx, &database.InsertDeploymentOptions{
+// 		ProjectID:         p1.ID,
+// 		Slots:             2,
+// 		Branch:            "main",
+// 		RuntimeHost:       "host_1",
+// 		RuntimeInstanceID: uuid.NewString(),
+// 	})
 
-	_, err = db.InsertDeployment(ctx, &database.InsertDeploymentOptions{
-		ProjectID:         p1.ID,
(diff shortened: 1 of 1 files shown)


proto/rill/admin/v1/api.proto
```diff
@@ -555,3 +569,9 @@ message UserInvite {
   string role = 2;
   string invited_by = 3;
 }
+
+
+message PageToken {
+  // cursor is kept as an array since it may not be possible to use one field to unqiuely identify in all cases
+  repeated string cursor = 1;
+}
```
begelundmuller, line 577: 1. Since this isn't part of the external API, move to a file `internal.proto` in `proto/rill/admin/v1`
2. Don't need to take a generic approach to the fields – empty fields are omitted in protobuf serialization, so it's a needless optimization. Better to have explicit fields like `string string_value = 1`, and can add other fields if needed in the future (like `int64 int_value = 2`, `repeated string string_values = 3`, etc.).

admin/server/projects.go
```diff
@@ -24,95 +23,48 @@ func (s *Server) ListProjectsForOrganization(ctx context.Context, req *adminv1.L
...
 	}
 
+	opts, err := paginationOptions(req.PageToken, int(req.PageSize))
+	if err != nil {
+		return nil, err
+	}
+
 	// If user has ManageProjects, return all projects
 	claims := auth.GetClaims(ctx)
+	projs := make([]*database.Project, 0)
```
begelundmuller, line 33: Redundant allocation – do `var projs []*database.Project`

admin/server/projects.go
```diff
@@ -24,95 +23,48 @@ func (s *Server) ListProjectsForOrganization(ctx context.Context, req *adminv1.L
...
-		dtos := make([]*adminv1.Project, len(projs))
-		for i, p := range projs {
-			dtos[i] = s.projToDTO(p, org.Name)
-		}
-
-		return &adminv1.ListProjectsForOrganizationResponse{
-			Projects: dtos,
-		}, nil
+		projs, err = s.admin.DB.FindProjectsForOrganization(ctx, org.ID, opts)
+	} else if claims.OwnerType() == auth.OwnerTypeUser {
```
begelundmuller, line 36: Missing another `else` conditions (for everyone else, i.e. anon users) – should return result of `FindPublicProjectsInOrganization`

admin/database/database.go
```diff
@@ -47,8 +47,8 @@ type DB interface {
 	Migrate(ctx context.Context) error
 	FindMigrationVersion(ctx context.Context) (int, error)
 
-	FindOrganizations(ctx context.Context) ([]*Organization, error)
-	FindOrganizationsForUser(ctx context.Context, userID string) ([]*Organization, error)
+	FindOrganizations(ctx context.Context, paginationOpts *PaginationOptions) ([]*Organization, error)
```
begelundmuller, line 50: Replace the `PaginationOptions` with the actual query parameters – in this case something like `limit int, afterName string`. 

The `database` functions are all very close to the actual query parameters, and having a generic name like `Cursor` makes it easy to accidentally pass a value that doesn't match the underlying column name (like an ID instead of a name).

admin/database/postgres/postgres.go
```diff
@@ -67,8 +67,8 @@ func (c *connection) FindOrganizationsForUser(ctx context.Context, userID string
 		UNION
 		SELECT o.* FROM orgs o JOIN projects p ON o.id = p.org_id
 		JOIN users_projects_roles upr ON p.id = upr.project_id
-		WHERE upr.user_id = $1) u order by u.name
-	`, userID)
+		WHERE upr.user_id = $1) u where lower(u.name) > $2 order by lower(u.name) limit $3
```
begelundmuller, line 70: Break line, use capitalized SQL keywords

admin/database/postgres/postgres.go
```diff
@@ -180,25 +180,31 @@ func (c *connection) FindProjectsForUser(ctx context.Context, userID string) ([]
...
+				UNION
+				SELECT ugpr.project_id FROM usergroups_projects_roles ugpr JOIN usergroups_users uug ON ugpr.usergroup_id = uug.usergroup_id WHERE uug.user_id = $2
+			)
 			UNION
-			SELECT ugpr.project_id FROM usergroups_projects_roles ugpr JOIN usergroups_users uug ON ugpr.usergroup_id = uug.usergroup_id WHERE uug.user_id = $2
-		)
-	`, orgID, userID)
+			SELECT p.* FROM projects p WHERE p.org_id = $1 AND p.public = true
+		) up where lower(up.name) > $3 ORDER BY lower(up.name) LIMIT $4
+	`, orgID, userID, paginationOpts.Cursor, paginationOpts.PageSize)
```
begelundmuller, line 207: 1. Wrapping `SELECT` and second union is redundant – just move to `(p.public = true OR p.id IN ...`
2. Capitalize SQL keywords

admin/database/postgres/postgres.go
```diff
@@ -657,11 +655,11 @@ func (c *connection) CountSingleuserOrganizationsForMemberUser(ctx context.Conte
 	return count, nil
 }
 
-func (c *connection) FindProjectMemberUsers(ctx context.Context, projectID string) ([]*database.Member, error) {
+func (c *connection) FindProjectMemberUsers(ctx context.Context, projectID string, paginationOpts *database.PaginationOptions) ([]*database.Member, error) {
 	var res []*database.Member
 	err := c.getDB(ctx).SelectContext(ctx, &res, `SELECT u.id, u.email, u.display_name, u.created_on, u.updated_on, r.name FROM users u 
     	JOIN users_projects_roles upr ON u.id = upr.user_id
-		JOIN project_roles r ON r.id = upr.project_role_id WHERE upr.project_id=$1`, projectID)
+		JOIN project_roles r ON r.id = upr.project_role_id WHERE upr.project_id=$1 AND lower(u.email) > $2 ORDER BY lower(u.email) limit $3`, projectID, paginationOpts.Cursor, paginationOpts.PageSize)
```
begelundmuller, line 662: 1. Capitalize SQL keywords
2. Also please fix odd line breaks/indentation in this query

admin/server/organizations.go
```diff
@@ -471,3 +536,40 @@ func organizationToDTO(o *database.Organization) *adminv1.Organization {
...
+
+func validPageSize(pageSize int) int {
+	if pageSize <= 0 {
+		return _defaultPageSize
+	}
+	if pageSize > _maxPageSize {
+		return _maxPageSize
+	}
+	return pageSize
+}
```
begelundmuller, line 575: 1. Replace `paginationOptions` with `unmarshalPageToken(token string) (*adminv1.PageToken, error)`
2. Remove `validPageSize` and validate using `(validate.rules)` annotations in the proto file instead.
3. Replace `nextPageToken` with something like `marshalPageToken(val string) string`. There's no need to return an `error` here, it can just `panic` (would be a dev error).
4. Move this code to `server.go` or a `page_token.go` file, this is not specific to organizations

begelundmuller (changes requested):

admin/server/organizations.go
k-anshul: 2. Wouldn't that mean that we will reject the request if `page_size` > max. Is there a way in `validate.rules` to change the value if `gt` some value ?

proto/rill/admin/v1/api.proto
k-anshul: 2. It will be very weird to see a `PageToken` proto in future with both `string_value` and `string_values`. 
```
message PageToken {
  string string_value = 1;
  repeated string string_values = 2;
}
```
This will just add tons of confusion. 

The necessity of having multiple columns as cursor isn't very far fetched either. Imagine getting all projects for users across all their organisations.
Also I find keeping different datatype fields confusing as well. Its neither generic nor specific.
I would either choose an API specific token or a generic token with repeated fields(even if that meant I have to cast in the code).

admin/server/organizations.go
begelundmuller: 2. It's best to just reject the request instead of silently changing behavior. It would be discovered at frontend dev time anyway, so they could fix it immediately.

k-anshul: From what I understood this is definitely not a best practice to reject the request in this case and a server may always return less records than what has been requested

proto/rill/admin/v1/api.proto
begelundmuller: I agree that some cases might warrant an API specific format.

But if going with a generic format, it's better to at least use the type system as much as possible. E.g. if you need an int offset, coercing it to a string array is weird. Not sure what the right solution is, some other ideas:

- Use a `oneof` type, for example: https://github.com/protocolbuffers/protobuf/blob/0ae328929ee2b6378d8cb3bb2c26700922098910/src/google/protobuf/struct.proto#L62
- Have specific messages for different types, like:
```proto
message StringCursor {
  string value = 1;
}

// Etc.
```
- Have something like `repeated google.protobuf.Value`, but you'd need a custom type since `google.protobuf.Value` can't contain ints.
- Use semantic field names (note this is exactly what you would implicitly do if serializing a map):
```proto
message PageToken {
  oneof property {
    string id = 1;
    string name = 2;
    int32 offset = 3;
  }
}
```

k-anshul: Also to add its not uncommon to have an array of string for such use cases. Take an example of [grpc-metdata](https://github.com/grpc/grpc-go/blob/master/Documentation/grpc-metadata.md#constructing-metadata) the key is `string` and the value is a `[]string`. 

I would not add `offset` in the cursor but would instead add two new fields for `offset` and `page`. I would expect to set `cursor` when using cursor based pagination or `offset` and `page` when using offset based pagination. Alternatively I would create another type of PageToken and would probably use that. The goal with cursor as `[]string` is not to create a generic token for all use cases but to be able to adapt for different cursors that's why the field name cursor.

1. `oneof` can't be repeated and we may need a string and int cursor both at the same time. I would need to wrap this in another message and it will be same as the third idea in the list.
2. I like this idea. Its simple and self explanatory that the token is a `stringcursor`.
3. This one works but feel its complicated for the use case. It makes sense to use this if I am not marshalling the data as is the case with `json`. Here I clearly know what `type` I will be getting. 
4. Sorry I didn't really understand this.

begelundmuller: I think we're a little confused on terminology here – a "cursor" can be anything that marks the next place to fetch from, can also be an "offset". 

We would do best to avoid using both "cursor" and "page token" since they are basically the same. I think introducing a `StringPageToken` containing one `string value = 1` field makes sense for now. In the future, we can then introduce `StringsPageToken`, `IntPageToken`, `APISpecificPageToken`, etc. types as needed.

How does that sound?

k-anshul: Fundamentally agree but have always found `offset` to be considered different from `cursor`.

Yeah `StringPageToken ` sounds good and we can add more types if required. Will change it to `StringPageToken`.

admin/server/organizations.go
k-anshul: Page token invalid and page size higher than limit are totally different cases. The handling needs to be looked on a case to case basis.
Github APIs return response capping to max limit even if higher page size is requested for example : 
```
curl -L \
  -H "Accept: application/vnd.github+json" \
  -H "Authorization: Bearer <token>"\
  -H "X-GitHub-Api-Version: 2022-11-28" \
  https://api.github.com/repos/OWNER/REPO/issues?per_page=100000
```
Not returning an error and limiting to max page size would allow us to reduce these max limits in future if required without breaking the client.

admin/database/postgres/postgres.go
```diff
@@ -225,34 +226,43 @@ func (c *connection) FindProjectsForUser(ctx context.Context, userID string) ([]
...
-		)
-	`, orgID, userID)
+		SELECT up.* FROM ( SELECT p.* FROM projects p
+			WHERE p.org_id = $1 AND (p.public = true OR p.id IN (
+				SELECT upr.project_id FROM users_projects_roles upr WHERE upr.user_id = $2
+				UNION
+				SELECT ugpr.project_id FROM usergroups_projects_roles ugpr JOIN usergroups_users uug ON ugpr.usergroup_id = uug.usergroup_id WHERE uug.user_id = $2
+			))
+		) up WHERE lower(up.name) > $3 ORDER BY lower(up.name) LIMIT $4
+	`, orgID, userID, afterProjectName, limit)
```
begelundmuller, line 252: The outer `SELECT` is still redundant, can be flattened, like:
```sql
WHERE p.org_id=$1 AND lower(p.name)>$2 AND (p.public=true OR p.id IN (...))
ORDER BY lower(name) LIMIT $4
```

admin/server/page_token.go
```diff
@@ -0,0 +1,43 @@
...
+	token := &adminv1.StringPageToken{Val: val}
+	bytes, err := proto.Marshal(token)
+	if err != nil {
+		return "", err
+	}
+
+	return base64.URLEncoding.EncodeToString(bytes), nil
+}
+
+func validPageSize(pageSize int) int {
```
begelundmuller, line 38: Take `uint32` input to save cast everywhere it's called

begelundmuller (changes requested):

begelundmuller (approved):

PR #2331, opened 2023-05-04
API returning model rows matching dashboard filters

begelundmuller:
Requirements details [here](https://www.notion.so/rilldata/API-to-return-model-rows-based-on-dashboard-filters-and-time-range-9288ec0a7c894ec685d795179dac5471).

The API is generated as `createQueryServiceMetricsViewRows`.


egor-ryashin (approved):
LGTM

PR #2334, opened 2023-05-04
Set sample_size=200000 for CSV/JSON ingestion to scan enough events for sampling

esevastyanov:


begelundmuller (approved):
LGTM – but please hold pending discussion [here](https://rilldata.slack.com/archives/C01190F1R1D/p1683200458728579)

begelundmuller:
Update: Can we set it to a default of `200000`, instead of `-1`? Then the PR is good to merge.

PR #2358, opened 2023-05-07
Add auth template to monorepo

djbarnwal:
Move old Auth template to svelte

Tagging @ericpgreen2 for web-admin/CTA components related changes and @hamilton for overall structure of the package. 

- [x] CI dependent on #2361


web-auth/package.json
```diff
@@ -0,0 +1,48 @@
...
+		"svelte-check": "^3.0.1",
+		"tailwindcss": "^3.2.7",
+		"tslib": "^2.4.1",
+		"typescript": "^4.9.3",
+		"vite": "^4.0.0"
+	},
+	"type": "module",
+	"dependencies": {
+		"auth0-js": "^9.20.2",
+		"dotenv": "^16.0.3"
```
ericpgreen2, line 46: Can these be `devDependencies`? We should only use `devDependencies` if possible.

.prettierignore
```diff
@@ -49,3 +49,7 @@ sql/
 
 # a temporary fix until we refactor css classes
 *.css
+
+# generated by runtime
```
ericpgreen2, line 53: generated by Orval

web-auth/.gitignore
ericpgreen2, line 1: This looks like the same `.gitignore` that's at the top-level? Can you trim this down to only ignore things relevant to `web-auth`?

web-auth/src/app.css
ericpgreen2, line 1: Can you instead use the `RillTheme` component, which imports `web-common/app.css`, so we don't need this file?

web-auth/package.json
djbarnwal: Auth0 is a core dependency and it's APIs are directly used in the codebase. Have moved `dotenv` to `devDependencies`

ericpgreen2:
Ah one more thing. Can we add `web-auth` to `.github/web-test.yml` our Github Action configuration? We can mirror the same format as the other `web-x` directories.

ericpgreen2 (approved):

PR #2363, opened 2023-05-08
Implement deployment hibernation

rakeshsharma14317:
- Deletes a project's prod deployment after 14 days of inactivity
- Inactivity is measured as no calls to `GetProject`
- A hibernated project can be rehydrated by calling `rill project reconcile`
- A hibernated project will have `prodDeployment == nil`. The frontend code has been updated to support this.
- Attaching a screenshot of the frontend experience when accessing a hibernated project. It's stable and refinement of the experience can happen in a separate PR.

Screenshot 2023-07-04 at 14 27 27


admin/database/database.go
```diff
@@ -207,17 +209,19 @@ type Project struct {
...
+	GithubInstallationID *int64        `db:"github_installation_id"`
+	Subpath              string        `db:"subpath"`
+	ProdBranch           string        `db:"prod_branch"`
+	ProdVariables        Variables     `db:"prod_variables"`
+	ProdOLAPDriver       string        `db:"prod_olap_driver"`
+	ProdOLAPDSN          string        `db:"prod_olap_dsn"`
+	ProdSlots            int           `db:"prod_slots"`
+	ProdDeploymentID     *string       `db:"prod_deployment_id"`
+	ProductionTTL        time.Duration `db:"production_ttl"`
+	PreviewTTL           time.Duration `db:"preview_ttl"`
```
begelundmuller, line 222: Make it `ProdTTL` instead to align with other fields. And we can remove `PreviewTTL` until we implement previews (didn't add it for other fields either yet)

admin/database/database.go
```diff
@@ -269,6 +273,7 @@ const (
 	DeploymentStatusOK          DeploymentStatus = 2
 	DeploymentStatusReconciling DeploymentStatus = 3
 	DeploymentStatusError       DeploymentStatus = 4
+	DeploymentStatusHibernated  DeploymentStatus = 3
```
begelundmuller, line 276: Number should be 5

admin/database/database.go
```diff
@@ -85,6 +86,7 @@ type DB interface {
 	DeleteDeployment(ctx context.Context, id string) error
 	UpdateDeploymentStatus(ctx context.Context, id string, status DeploymentStatus, logs string) (*Deployment, error)
 	UpdateDeploymentBranch(ctx context.Context, id, branch string) (*Deployment, error)
+	UpdateDeploymentTS(ctx context.Context, ids []string) (*Deployment, error)
```
begelundmuller, line 89: Make it `UpdateDeploymentUsedOn` and add a new timestamptz field `used_on` for it. The hibernation checks should use that field, not `updated_on`

go.mod
```diff
@@ -47,6 +47,7 @@ require (
 	github.com/prometheus/client_golang v1.15.0
 	github.com/rs/cors v1.9.0
 	github.com/spf13/cobra v1.7.0
+	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.8.2
```
begelundmuller, line 51: Not sure why this has changed, it's not a dependency. Try running `go mod tidy` on this branch

admin/admin.go
```diff
@@ -16,6 +17,11 @@ type Options struct {
 	ProvisionerSpec string
 }
 
+type updateTSCache struct {
+	cache map[string]bool
+	lock  sync.Mutex
+}
```
begelundmuller, line 23: Some ways to make this cleaner:

- Rename to `usedFlusher` and move to a separate file `used_flusher.go`
- Rename `cache` to `deployments` (in the future, we may use this same object to track `used_on` for other objects, like access tokens)
- Add the flushing related methods on this struct instead of the `Service`
- Add a `newUsedFlusher(logger, db)` method that initializes it and starts the background flushing
- Add a `.close() error` method that stops the background flushing. Call it from `Service.Close`

admin/deployments.go
```diff
@@ -170,6 +170,46 @@ func (s *Service) updateDeployment(ctx context.Context, depl *database.Deploymen
...
+			continue
+		}
+
+		if depl.Status == database.DeploymentStatusReconciling && time.Since(depl.UpdatedOn) < 30*time.Minute {
+			fmt.Printf("skipping because it is already running\n")
+			continue
+		}
+
+		if proj.ProductionTTL < time.Since(depl.UpdatedOn) {
+			err := s.teardownDeployment(ctx, proj, depl)
```
begelundmuller, line 194: `teardownDeployment` completely deletes the deployment, so we can't update its status after – not sure what the best path forward is here

admin/deployments.go
```diff
@@ -170,6 +170,46 @@ func (s *Service) updateDeployment(ctx context.Context, depl *database.Deploymen
...
+	if err != nil {
+		return err
+	}
+
+	for _, proj := range projects {
+		depl, err := s.DB.FindDeployment(ctx, *proj.ProdDeploymentID)
+		if err != nil {
+			fmt.Printf("depl not found for proj %s", proj.Name)
+			continue
+		}
```
begelundmuller, line 186: Instead of `FindAllProjects`, add a DB call that returns all deployments where `used_on + prod_ttl < now()`

begelundmuller (changes requested):
Also please add a new background job in `worker` that calls `HibernateDeployments` every 5 minutes

admin/database/database.go
```diff
@@ -264,6 +269,7 @@ type UpdateProjectOptions struct {
 	ProdVariables        map[string]string
 	ProdDeploymentID     *string
 	ProdSlots            int
+	ProdTTL              *int64
```
begelundmuller, line 272: Make sure every place in the code (except tests) that uses `UpdateProjectOptions` (search for it in editor) passes a value for `ProdTTL`

begelundmuller (changes requested):
I would also suggest a higher level refactor:
1. Remove the `HIBERNATED` status from deployments. Instead, if a deployment is to be hibernated, just delete it altogether. Then set `prod_deployment_id` on the project to `null`.
2. Then you can just use the existing `TriggerRedeploy` to re-hydrate a project. You just need to slightly refactor that function to only delete the old deployment if one exists.
3. You need to go over all places where `prod_deployment_id` is used and make sure the code there supports it being `nil`.

admin/database/postgres/postgres.go
```diff
@@ -363,6 +363,19 @@ func (c *connection) CountProjectsForOrganization(ctx context.Context, orgID str
...
 }
 
+// FindExpiredDeployments returns all the deployments which are expired as per prod ttl
+func (c *connection) FindExpiredDeployments(ctx context.Context) ([]*database.Deployment, error) {
+	var res []*database.Deployment
+	err := c.getDB(ctx).SelectContext(ctx, &res, `
+	SELECT d.* FROM deployments d 
+	JOIN projects p ON d.project_id = p.id
+	WHERE p.prod_ttl_seconds is not null AND d.used_on + p.prod_ttl_seconds * interval '1 second' < now()`)
+	if err != nil {
```
begelundmuller, line 373: nit: indent the query one level more

admin/database/database.go
```diff
@@ -80,13 +80,16 @@ type DB interface {
 	UpdateProjectVariables(ctx context.Context, id string, variables map[string]string) (*Project, error)
 	CountProjectsForOrganization(ctx context.Context, orgID string) (int, error)
 
+	FindExpiredDeployments(ctx context.Context) ([]*Deployment, error)
 	FindDeploymentsForProject(ctx context.Context, projectID string) ([]*Deployment, error)
 	FindDeployment(ctx context.Context, id string) (*Deployment, error)
 	FindDeploymentByInstanceID(ctx context.Context, instanceID string) (*Deployment, error)
 	InsertDeployment(ctx context.Context, opts *InsertDeploymentOptions) (*Deployment, error)
+	UpdateDeployment(ctx context.Context, id string, opts *UpdateDeploymentOptions) (*Deployment, error)
```
begelundmuller, line 88: Can remove this, not used anymore

admin/deployments.go
```diff
@@ -177,6 +177,54 @@ func (s *Service) updateDeployment(ctx context.Context, depl *database.Deploymen
 	return nil
 }
 
+// HIBERNATE free deployments
+func (s *Service) HibernateDeployments(ctx context.Context) error {
```
begelundmuller, line 181: nit: Update docstring to `// HibernateDeployments tears down unused deployments`

admin/deployments.go
```diff
@@ -177,6 +177,54 @@ func (s *Service) updateDeployment(ctx context.Context, depl *database.Deploymen
...
+	if err != nil {
+		return err
+	}
+
+	for _, depl := range depls {
+		fmt.Println("expired deployments are: ", depl)
+		if depl.Status == database.DeploymentStatusReconciling && time.Since(depl.UpdatedOn) < 30*time.Minute {
+			fmt.Printf("skipping because it is already running\n")
+			continue
+		}
```
begelundmuller, line 192: Use `s.logger.Info` instead of `fmt.Printf`

admin/deployments.go
```diff
@@ -177,6 +177,54 @@ func (s *Service) updateDeployment(ctx context.Context, depl *database.Deploymen
...
+		if err != nil {
+			s.logger.Error("skipping because error while find project", zap.String("projectId", proj.ID), zap.String("deplId", depl.ID), zap.Error(err))
+			continue
+		}
+
+		err = s.teardownDeployment(ctx, proj, depl)
+		if err != nil {
+			s.logger.Error("skipping because error while hibernate deployment", zap.String("projectId", proj.ID), zap.String("deplId", depl.ID), zap.Error(err))
+			continue
+		}
```
begelundmuller, line 204: Use snake case for log field keys, like `project_id` and `deployment_id`

admin/server/projects.go
```diff
@@ -209,6 +213,14 @@ func (s *Server) CreateProject(ctx context.Context, req *adminv1.CreateProjectRe
 		return nil, err
 	}
 
+	// Add prod TTL as 7 days if not a public project else infinite
+	prodTTL := int64(prodDeplTTL.Seconds())
+	var prodTTLPtr *int64
+	prodTTLPtr = &prodTTL
+	if req.Public {
+		prodTTLPtr = nil
+	}
```
begelundmuller, line 222: Can simplify if you do:
```go
var prodTTL *int64
if !req.Public {
  tmp := int64(prodDeplTTL.Seconds())
  prodTTL = &tmp
}
```

admin/worker/worker.go
```diff
@@ -42,6 +42,11 @@ func (w *Worker) Run(ctx context.Context) error {
 	group.Go(func() error {
 		return w.schedule(ctx, "delete_expired_device_auth_codes", w.deleteExpiredDeviceAuthCodes, 6*time.Hour)
 	})
+
+	group.Go(func() error {
+		return w.schedule(ctx, "hibernateExpiredDeployments", w.hibernateExpiredDeployments, 15*time.Minute)
+	})
```
begelundmuller, line 48: Use snake case like other background task names

cli/cmd/project/edit.go
```diff
@@ -71,6 +72,11 @@ func EditCmd(cfg *config.Config) *cobra.Command {
 				req.Public = &public
 			}
 
+			if cmd.Flags().Changed("prod-ttl-seconds") {
+				promptFlagValues = false
+				req.ProdTtlSeconds = &prodTTL
+			}
```
begelundmuller, line 78: Can we make it never prompt for this? If not explicitly set, it should just stay unchanged

proto/rill/admin/v1/api.proto
```diff
@@ -668,6 +670,7 @@ enum DeploymentStatus {
   DEPLOYMENT_STATUS_OK = 2;
   DEPLOYMENT_STATUS_RECONCILING = 3;
   DEPLOYMENT_STATUS_ERROR = 4;
+  DEPLOYMENT_STATUS_HIBERNATED = 5;
```
begelundmuller, line 673: Can remove this status now

admin/used_flusher.go
```diff
@@ -0,0 +1,84 @@
...
+func (u *usedFlusher) Deployment(id string) {
+	u.lock.Lock()
+	defer u.lock.Unlock()
+
+	u.deployments[id] = true
+}
+
+func (u *usedFlusher) runBackground(ctx context.Context) {
+	ticker := time.NewTicker(20 * time.Second)
+	go func() {
```
begelundmuller, line 47: Can remove the `go func()` scope here, since you're already calling `go used.runBackground`, so it will already run in the background

admin/used_flusher.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func (u *usedFlusher) Deployment(id string) {
+	u.lock.Lock()
+	defer u.lock.Unlock()
+
+	u.deployments[id] = true
+}
+
+func (u *usedFlusher) runBackground(ctx context.Context) {
```
begelundmuller, line 45: Don't need to pass `ctx` here since you have `u.ctx` already

admin/used_flusher.go
```diff
@@ -0,0 +1,84 @@
...
+	ticker := time.NewTicker(20 * time.Second)
```
begelundmuller, line 46: Move the `20 * time.Second` to a const at the top of the file

admin/used_flusher.go
```diff
@@ -0,0 +1,84 @@
...
+		for {
+			select {
+			case <-ticker.C:
+				u.logger.Info(`Inside LastusedFluser`, zap.Int("no of deployments", len(u.deployments)), observability.ZapCtx(ctx))
+				if len(u.deployments) > 0 {
+					err := u.updateDeplToDB(ctx)
+					if err != nil {
+						fmt.Printf("Error while flush update timestamp map into db, error: %v", err)
+					}
+				}
```
begelundmuller, line 57: Move the `if` check to `updateDeplToDB`

admin/used_flusher.go
```diff
@@ -0,0 +1,84 @@
...
+	_, err := u.db.UpdateDeploymentUsedOn(ctx, deplIds)
+	u.logger.Info(`Updated deployment status`, zap.Strings("Depl Ids", deplIds), observability.ZapCtx(ctx))
+
+	u.deployments = make(map[string]bool)
+	return err
+}
+
+func (u *usedFlusher) Close() {
+	u.cancel()
+}
```
begelundmuller, line 84: Since `Close` is an external func (capital first letter), move it up above the internal funcs (lowercase first letter)

begelundmuller: Also, close should also do a final call to `flush` (with a `context.Background()`) to ensure any final data gets flushed before stopping.

admin/used_flusher.go
```diff
@@ -0,0 +1,84 @@
...
+				}
+			case <-u.ctx.Done():
+				ticker.Stop()
+				return
+			}
+		}
+	}()
+}
+
+func (u *usedFlusher) updateDeplToDB(ctx context.Context) error {
```
begelundmuller, line 66: Just call this func `flush`

admin/used_flusher.go
```diff
@@ -0,0 +1,84 @@
...
+	deplIds := make([]string, 0, len(u.deployments))
+	for k := range u.deployments {
+		deplIds = append(deplIds, k)
+	}
+
+	_, err := u.db.UpdateDeploymentUsedOn(ctx, deplIds)
+	u.logger.Info(`Updated deployment status`, zap.Strings("Depl Ids", deplIds), observability.ZapCtx(ctx))
+
+	u.deployments = make(map[string]bool)
+	return err
```
begelundmuller, line 79: The lock may stay locked for a long time, potentially blocking calls to `Deployment`, slowing user requests down.

Instead, copy `u.deployments` into a tmp var, then set a new empty one, then unlock immediately, and then continue the DB write afterwards (where the lock is unlocked)

admin/used_flusher.go
begelundmuller: Actually, also move the info and error logging to that func, so it also gets applied when `flush` is called in `Close`

begelundmuller (changes requested):
Also please check all the code for places where the `prodDeploymentId` or `prodDeployment` is accessed and ensure it's able to handle them being `nil`

cli/cmd/project/edit.go
rakeshsharma14317: It is just setting the values here, We won't prompt for this you can see below, we just set it if this flag is `changed`

admin/used_flusher.go
```diff
@@ -30,55 +44,43 @@ func newUsedFlusher(logger *zap.Logger, db database.DB) *usedFlusher {
...
+			deplIds = append(deplIds, k)
+		}
 
-	u.deployments = make(map[string]bool)
-	return err
-}
+		_, err := u.db.UpdateDeploymentUsedOn(ctx, deplIds)
+		if err != nil {
+			fmt.Printf("Error while flush update timestamp map into db, error: %v", err)
+		}
```
begelundmuller, line 82: Use the logger!

admin/used_flusher.go
```diff
@@ -20,6 +22,18 @@ type usedFlusher struct {
...
+	u.lock.Lock()
+	defer u.lock.Unlock()
+
+	u.deployments[id] = true
+}
+
+func (u *usedFlusher) Close() {
+	u.flush(context.Background())
+	u.cancel()
+}
```
begelundmuller, line 35: There's a raise condition where if `flush` is called in `runBackground` and takes a long time to flush, and then `Close` is called while the first `flush` is running, the second `flush` here will exit quickly (since there's nothing to flush), and then `Close` will return even though the first `flush` is still running (which may lead to process exit and the first flush being stopped).

begelundmuller:
@ericpgreen2 Made some frontend changes to support hibernated projects (`prodDeployment` will be `nil`). Did a pretty thorough experience pass, but would appreciate a review from you of the frontend code.

admin/used_flusher.go
```diff
@@ -50,39 +40,57 @@ func newUsedFlusher(logger *zap.Logger, db database.DB) *usedFlusher {
...
-		u.lock.Lock()
-		deployments := u.deployments
-		u.deployments = make(map[string]bool)
-		u.lock.Unlock()
 
-		deplIds := make([]string, 0, len(deployments))
+	u.mu.Lock()
+	deployments := u.deployments
+	u.deployments = make(map[string]bool)
+	u.mu.Unlock()
```
rakeshsharma14317, line 76: Why we are not checking this inside `len(u.deployments) > 0`? if no deployments i guess we should not lock it every time this runs

begelundmuller: It would not be faster – you can't call `len(u.deployments)` without locking it first (locks are not just for writes, they're also needed for reads that can happen during a write).

ericpgreen2:
> @ericpgreen2 Made some frontend changes to support hibernated projects (`prodDeployment` will be `nil`). Did a pretty thorough experience pass, but would appreciate a review from you as well of the frontend code.

I found we didn't have a redirect for all routes underneath the `/org/project` route. Namely, the dashboards weren't redirecting to the project page. I just added a redirect in `[project]/+layout.svelte`. @begelundmuller, do you agree with it? If we want to display something on the dashboard page, that's probably for another PR.

(The code is confusing: That `org/project/-/redirect` page is only used for magic breadcrumb navigation when you're navigating from a dashboard. Feels like this code should be moved elsewhere to make that more clear.)

Otherwise, all looks good to me.

web-admin/src/routes/[organization]/[project]/+layout.svelte
```diff
@@ -1,34 +1,20 @@
...
-    }
+    $page.params.project
   );
+
+  $: isRuntimeHibernating = $projRuntime.isSuccess && !$projRuntime.data;
+
+  $: if (isRuntimeHibernating) {
+    // Redirect any nested routes (notably dashboards) to the project page
+    goto(`/${$page.params.organization}/${$page.params.project}`);
+  }
```
Comment (begelundmuller) [inline web-admin/src/routes/[organization]/[project]/+layout.svelte:17]: If already on the project page, is `goto` then a no-op or could this lead to some weird re-render cycles?

begelundmuller (approved):
Put a question below, but it's probably insignificant, so approving

PR #2364, opened 2023-05-08
fix: Show model preview runtime errors

skokenes:
Previously, we were not displaying runtime errors in the modeler that may be triggered when the model gets the table preview. See https://github.com/rilldata/rill-developer/issues/1570

This PR adds a check for the runtime errors at the ModelBody level and displays the runtime error if it exists, along with any other errors.

image

### DIFF
--- a/web-common/src/features/models/workspace/ModelBody.svelte
+++ b/web-common/src/features/models/workspace/ModelBody.svelte
@@ -45,8 +45,6 @@
   export let modelName: string;
   export let focusEditorOnMount = false;
 
-  let tableQuery;
-
   const queryClient = useQueryClient();
 
   const queryHighlight: Writable<QueryHighlightState> = getContext(
@@ -56,13 +54,13 @@
   $: runtimeInstanceId = $runtime.instanceId;
   const updateModel = createRuntimeServicePutFileAndReconcile();
 
-  let limit = 150;
+  const limit = 150;
 
   $: tableQuery = createQueryServiceTableRows(runtimeInstanceId, modelName, {
     limit,
   });
 
-  $: runtimeError = $tableQuery.error?.response.data;
+  $: runtimeError = ($tableQuery.error as any)?.response.data;
 
   // track innerHeight to calculate the size of the editor element.
   let innerHeight: number;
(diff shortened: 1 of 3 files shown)


web-common/src/features/models/workspace/ModelBody.svelte
```diff
@@ -44,6 +45,8 @@
   export let modelName: string;
   export let focusEditorOnMount = false;
 
+  let tableQuery;
```
djbarnwal, line 48: This is redundant. The reactive statement `$: tableQuery = ..` will initialize the variable for us.

web-common/src/features/models/workspace/ModelBody.svelte
```diff
@@ -53,6 +56,14 @@
   $: runtimeInstanceId = $runtime.instanceId;
   const updateModel = createRuntimeServicePutFileAndReconcile();
 
+  let limit = 150;
```
djbarnwal, line 59: Let's make this `const` if we are not changing this anywhere

web-common/src/features/models/workspace/ModelBody.svelte
```diff
@@ -53,6 +56,14 @@
...
+
+  $: tableQuery = createQueryServiceTableRows(runtimeInstanceId, modelName, {
+    limit,
+  });
+
+  $: runtimeError = $tableQuery.error?.response.data;
```
djbarnwal, line 65: I believe the type for `response` is not present. There's already a TODO note in the file. For now we can just use `($tableQuery.error as any)?.response.data`

djbarnwal (approved):
Looks good, have left some nit comments. Feel free to merge after those have been resolved. Approving.

web-common/src/features/models/workspace/ModelBody.svelte
skokenes: Done

web-common/src/features/models/workspace/ModelBody.svelte
skokenes: Done

web-common/src/features/models/workspace/ModelBody.svelte
skokenes: Done

PR #2377, opened 2023-05-09
Fix error code styling issue

djbarnwal:
The extra bold fontface was not added to the pre-load list. As a result, the error code text (which was using extra bold styling) had a delay in it's styling.

That being said, preloading it on all the pages also has it issues. It can slow down the load time of all the pages in the application

The other option is to use `font-display:block` ([more here](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-display)), but this hides the error code until the font is fetched while showing other text on the page which is an odd experience.

I am more inclined to keep this as is. Not sure if slight delay in styling is worth loading 100kb on every page.

### DIFF
--- a/web-admin/src/app.html
+++ b/web-admin/src/app.html
@@ -38,6 +38,13 @@
       type="font/woff2"
       crossorigin
     />
+    <link
+      rel="preload"
+      href="%sveltekit.assets%/fonts/inter/Inter-ExtraBold.woff2?v=3.19"
+      as="font"
+      type="font/woff2"
+      crossorigin
+    />
     <meta charset="utf-8" />
     <link rel="icon" href="%sveltekit.assets%/favicon.png" />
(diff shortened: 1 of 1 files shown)


bcolloran:
Looks good, I'll approve and merge. 

Regarding the concern about page load, [see this comment](https://github.com/rilldata/rill-developer/pull/1969#issue-1636203204
).  When @hamilton and I discussed this a few months ago, we decided to prioritize correct styling over snappy load, but there are options that we can use to improve this if page load is starting to feel heavy: (1) preload only subsets of critical fonts (2) use built fonts styled to look like the desired font until the desired font is loaded (3) make sure we're only using Inter Var for everything so that we don't need all these weights. These approaches can work in concert.

So @djbarnwal  @hamilton @ericpgreen2 -- make a mental a note that we can dig into this more if we feel that we need to optimize.

PR #2378, opened 2023-05-09
2366 cli burndown part ii

rakeshsharma14317:
This PR also contains the cli changes from bug bash list : https://www.notion.so/rilldata/Rill-cloud-Bug-bash-2413071d9b734a9285ba677369874b69#5ab78daed16a4274bf8ca55642bf1378


cli/cmd/project/delete.go
```diff
@@ -25,24 +25,32 @@ func DeleteCmd(cfg *config.Config) *cobra.Command {
...
+			if len(args) > 0 {
+				name = args[0]
+			}
+
+			if !cmd.Flags().Changed("project") && len(args) == 0 && cfg.Interactive {
 				name, err = inferProjectName(cmd.Context(), client, cfg.Org, path)
 				if err != nil {
 					return err
 				}
 			}
```
begelundmuller, line 37: This is missing the select list for picking projects in interactive state. A new util for that could be something like:
```go
err = cmdutil.RequireSelectArgString(cmd, cfg, &name, args, 0, "project", "Select project", func () ([]string, error) {
  return cmdutil.ProjectNamesByOrg(ctx, client, cfg.Org)
  })
```

It would be like the previous suggested util, except it takes a func at the end that returns a list of `[]string` options to select from.

Then the code in this command (and many of the other `rill project` commands could be):

```go
if len(args) == 0 && !cmd.Flags().Changed("project") {
  name, err = inferProjectName(...)
  if err != nil {
    return err
  }
}

if name == "" {
  err = cmdutil.RequireSelectArgString(cmd, cfg, &name, args, 0, "project", "Select project", func () ([]string, error) {
    return cmdutil.ProjectNamesByOrg(ctx, client, cfg.Org)
  })
  if err != nil {
    return err
  }
}
```

cli/cmd/project/project.go
```diff
@@ -72,6 +72,6 @@ func inferProjectName(ctx context.Context, adminClient *client.Client, org, path
 type project struct {
 	Name         string `header:"name" json:"name"`
 	Public       bool   `header:"public" json:"public"`
-	GithubURL    string `header:"github" json:"github"`
+	GithubURL    string `header:"git url" json:"github"`
```
begelundmuller, line 75: I know this was suggested in the bug bash, but IMO this should be "github" – the URLs we return here are *not* valid Git URLs (they don't end in `.git`)

cli/cmd/project/delete.go
```diff
@@ -25,24 +25,32 @@ func DeleteCmd(cfg *config.Config) *cobra.Command {
...
 
+			if name == "" {
+				return fmt.Errorf("please provide valid project name, Run `rill project list` for available projects")
```
begelundmuller, line 40: Fix grammar: `"Please provide a valid project name. Run `rill project list` to see the available projects."`

cli/pkg/cmdutil/cmdutil.go
```diff
@@ -374,6 +375,44 @@ func DefaultProjectName() string {
...
+				prompt := &survey.Confirm{
+					Message: fmt.Sprintf("Confirm \"%s\"?", color.YellowString(f.Usage)),
+				}
+
+				err = survey.AskOne(prompt, &public)
+				if err != nil {
+					return
+				}
+
+				value = fmt.Sprintf("%t", public)
```
begelundmuller, line 402: `public` is probably a copy paste err

cli/pkg/cmdutil/cmdutil.go
```diff
@@ -374,6 +375,44 @@ func DefaultProjectName() string {
...
+				if err != nil {
+					fmt.Println("error while input prompt, error:", err)
+					return
+				}
+			}
+
+			if f.Value.Type() == "bool" {
+				var public bool
+				prompt := &survey.Confirm{
+					Message: fmt.Sprintf("Confirm \"%s\"?", color.YellowString(f.Usage)),
```
begelundmuller, line 394: Doesn't need to be yellow

cli/cmd/org/delete.go
```diff
@@ -13,20 +13,32 @@ import (
...
+			if len(args) > 0 {
+				name = args[0]
+			}
+
+			if len(args) == 0 && cfg.Interactive {
+				err = cmdutil.SetFlagsByInputPrompts(*cmd, "name")
+				if err != nil {
+					return err
+				}
+			}
```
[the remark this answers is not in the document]
begelundmuller: The `SetFlagsByInputPrompts` can still be useful for populating optional values that only have flags. Maybe just not for params that can also be populated from `arg` or that require select lists.

begelundmuller (changes requested):

cli/cmd/org/edit.go
```diff
@@ -73,7 +75,7 @@ func EditCmd(cfg *config.Config) *cobra.Command {
 		},
 	}
 	editCmd.Flags().SortFlags = false
-	editCmd.Flags().StringVar(&orgName, "org", cfg.Org, "Organization name")
+	editCmd.Flags().StringVar(&orgName, "name", cfg.Org, "Organization name")
```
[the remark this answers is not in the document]
rakeshsharma14317: Actually i got the couple comments on this to have a same flag name for every subcommand, it was previously `--name` for `org create` and `--org` for others. an example of some comments is in bug bash https://www.notion.so/rilldata/Rill-cloud-Bug-bash-2413071d9b734a9285ba677369874b69?pvs=4#8e9f0c91579040a298b9bbe0e3a47c73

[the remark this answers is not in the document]
rakeshsharma14317: Also i guess with `rill org` having flag as `--name` looks better than `rill org delete --org`. We have `--org` with all the commands for project i.e `rill project --org`

So i guess we can keep it `--name` for org cmd

cli/cmd/project/delete.go
rakeshsharma14317: Yeah we can have some common util function select as well, but in above case `inferProjectName` has the SelectPrompt inside that function. So user will be prompted with project list when in interactive state.

cli/cmd/project/delete.go
rakeshsharma14317: making it as `please provide a valid project name. Run \`rill project list\` to see the available projects`

cli/cmd/org/edit.go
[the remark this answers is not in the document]
begelundmuller: I think with the new positional support, calling the flag `org` everywhere except `create` feels most consistent. Since that's what it's called everywhere. Similarly, `rill project edit` would take `--project` to override the name, not `--name`. 

It's fine to have `rill org create` as an exception since here the name is not yet acting as an ID.

cli/cmd/project/project.go
begelundmuller: Yes, see my comment there. It does not generalize to other platforms than Github anyway, so we should keep the old title

begelundmuller (approved):

PR #2381, opened 2023-05-09
Toplist inline measure api test (QUESTION, not a real PR)

bcolloran:
Hey @hamilton and @begelundmuller, I started looking at the inline measure API per Hamilton's request, but could use you all's help since I have not worked with the query APIs yet. Thought the easiest way to ask a question would be to post a PR that we can delete, since that allows me to comment on just the things I added relative to @begelundmuller 's branch. I *think* I am invoking this api correctly as part of a `totalsQuery` created with `createQueryServiceMetricsViewTotals` based on the new types in @begelundmuller 's branch and the invocation pattern I see elsewhere, but I'm not sure I'm doing it right, because I'm not seeing the what I expected in the query response. 

-  I created an `inlineMeasures` object as:
``` 
let inlineMeasures: V1InlineMeasure[] = [
      {
        name: "name_that_must_never_collide",
        expression: "COUNT(*)",
      },
    ];
```
- I added this as an `inlineMeasures` section to the `totalsQueryParams: QueryServiceMetricsViewTotalsBody` arg of the `createQueryServiceMetricsViewTotals` call.
- I expected to see an additional measure in the return data with name "name_that_must_never_collide", and with numeric value matching the user-defined `count(*)` measure in this metrics definition.

I actually see only the user defined measure's I've created. See the screenshot below where I log the query return for a test dashboard with 11 user-defined measures -- it has only those 11 measure in the `data` and `meta` sections.

Not sure if this branch is not fully implemented and I jumped the gun on testing, or if I'm just doing something wrong  when iinvoking the API (I suspect the latter 😅), so thought I'd document what I tried so you all can let me know where I've gone astray.

Also cc @AdityaHegde in case you see what I'm doing wrong. Thanks all for your help as I get spun up in this area.

image

### DIFF
--- a/web-common/src/features/dashboards/leaderboard/LeaderboardDisplay.svelte
+++ b/web-common/src/features/dashboards/leaderboard/LeaderboardDisplay.svelte
@@ -10,6 +10,8 @@
   import {
     createQueryServiceMetricsViewTotals,
     MetricsViewDimension,
+    QueryServiceMetricsViewTotalsBody,
+    V1InlineMeasure,
     V1MetricsViewTotalsResponse,
   } from "@rilldata/web-common/runtime-client";
   import { CreateQueryResult, useQueryClient } from "@tanstack/svelte-query";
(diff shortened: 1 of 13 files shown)


web-common/src/features/dashboards/leaderboard/LeaderboardDisplay.svelte
```diff
@@ -62,7 +64,18 @@
...
+        name: "name_that_must_never_collide",
+        expression: "COUNT(*)",
+      },
+    ];
+
+    let totalsQueryParams: QueryServiceMetricsViewTotalsBody = {
+      measureNames: selectedMeasureNames,
+      inlineMeasures,
+    };
+
```
bcolloran, line 78: Added this on top of @begelundmuller 's branch to try to add an inline measure

web-common/src/features/dashboards/leaderboard/LeaderboardDisplay.svelte
```diff
@@ -79,6 +92,8 @@
     );
   }
 
+  $: console.log("$totalsQuery?.data", $totalsQuery?.data);
```
bcolloran, line 95: just logging the query return

PR #2386, opened 2023-05-10
allow schema relaxation

### LINKED ISSUE #2337: Runtime: Support schema relaxation when later partitions have new columns for glob patterns
Initial thoughts documented here : https://www.notion.so/rilldata/Support-schema-relaxation-a831129b82b747f08c44f08d6183bfc4

k-anshul:
closes #2337


runtime/drivers/duckdb/connectors.go
```diff
@@ -14,13 +14,24 @@ import (
...
 )
 
+var dateTypeChangeMapping = map[string][]string{
+	"BOOLEAN":   {"BIGINT", "DOUBLE", "VARCHAR"},
+	"BIGINT":    {"DOUBLE", "VARCHAR"},
+	"DOUBLE":    {"VARCHAR"},
+	"TIME":      {"VARCHAR"},
+	"DATE":      {"TIMESTAMP", "VARCHAR"},
+	"TIMESTAMP": {"VARCHAR"},
+	"VARCHAR":   {},
```
nishantmonu51, line 32: do we also need to add TINYINT -> INTEGER -> BIGINT as well ?

k-anshul: I only added a conversion mapping of types which are inferred during auto detection. The table is mentioned here : https://duckdb.org/docs/data/csv/auto_detection#type-detection

k-anshul: okay but seems like its there only for csv. For parquet such a sequence is possible.

runtime/drivers/duckdb/connectors.go
```diff
@@ -75,20 +110,86 @@ func (c *connection) ingest(ctx context.Context, env *connectors.Env, source *co
...
 	}
 
-	var query string
-	if appendToTable {
-		query = fmt.Sprintf("INSERT INTO %q (SELECT * FROM %s);", source.Name, from)
-	} else {
-		query = fmt.Sprintf("CREATE OR REPLACE TABLE %s AS (SELECT * FROM %s);", source.Name, from)
+	currentSchema = oldSchema
+	if currentSchema == nil {
+		currentSchema, err = c.scanSchemaFromQuery(ctx, fmt.Sprintf("Describe %s;", source.Name))
```
begelundmuller, line 134: nit: uppercase `DESCRIBE`

runtime/drivers/duckdb/connectors.go
```diff
@@ -197,6 +316,10 @@ func generateReadCsvStatement(paths []string, properties map[string]any) (string
 	if _, sampleSizeDefined := ingestionProps["sample_size"]; !sampleSizeDefined {
 		ingestionProps["sample_size"] = 200000
 	}
+	// set union_by_name to unify the schema of the files
+	if _, defined := ingestionProps["union_by_name"]; !defined {
+		ingestionProps["union_by_name"] = true
+	}
```
begelundmuller, line 322: If the user explicitly set `union_by_name` to false, we should check that allow column addition is false

begelundmuller (commented):
A separate note, doesn't need to be addressed here. This file is getting pretty ugly (some places there are four levels of nesting), and it's not that complicated. There's probably ways to make the control flows simpler here.

runtime/drivers/duckdb/connectors.go
```diff
@@ -59,13 +60,47 @@ func (c *connection) ingest(ctx context.Context, env *connectors.Env, source *co
...
+			tableSchema = newSchema
+			if srcSchema != nil {
+				// add columns and their datatypes to ensure the datatypes are not inferred again
+				var ingestionProps map[string]any
+				if duckDBProps, ok := source.Properties["duckdb"].(map[string]any); ok {
+					ingestionProps = copyMap(duckDBProps)
+				} else {
+					ingestionProps = map[string]any{}
+				}
+				ingestionProps["columns"] = schemaToDuckDBColumnsProp(srcSchema)
```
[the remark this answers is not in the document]
k-anshul: Do we want to allow column addition in this case ?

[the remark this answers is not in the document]
k-anshul: I remember now, I didn't handle this case since if `columns` (or `types` or `dtypes`) is specified the inference would ensure that the returned schema is same as what is specified in `columns` in cases when there are more columns in new file. In cases when new file has less columns it would just error out (not allowed by duckdb).
Hence I feel its just convenient to not add additional checks and increase complexity here and rather let the queries run (which should be fast since column inference doesn't need to happen).

runtime/drivers/duckdb/connectors.go
```diff
@@ -59,13 +60,47 @@ func (c *connection) ingest(ctx context.Context, env *connectors.Env, source *co
...
+				from, err = sourceReader(files, ingestionProps)
+				if err != nil {
+					return nil, err
+				}
+			}
+
+			// doing a union with existing schema to handle cases when order of columns is changed or count of columns in new csv is less
+			query = fmt.Sprintf("INSERT INTO %q ((SELECT * FROM %s LIMIT 0) UNION ALL BY NAME (SELECT * FROM %s));", source.Name, source.Name, from)
```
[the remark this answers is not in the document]
k-anshul: Yeah makes sense.

k-anshul:
> A separate note, doesn't need to be addressed here. This file is getting pretty ugly (some places there are four levels of nesting), and it's not that complicated. There's probably ways to make the control flows simpler here.

I agree on this. I think a lot of ugliness is bcz of repeated parsing of `source.Properties` and taking action on its basis. 
May be we need to add a local object called `Ingestor` which the connector can use. It needs to do following things :
 - parses the properties once and then we just use object fields.
 - The source reader and related functions can then be transferred to the `Ingestor` struct. 
 - Some complexity like keeping track of append vs insert can be shifted there which should reduce these nested if-elses.
 - **May be** keep an `interface` for reader and use separate `Readers` for parquet vs csv vs json.

runtime/drivers/duckdb/connectors.go
```diff
@@ -71,20 +123,82 @@ func (c *connection) ingest(ctx context.Context, env *connectors.Env, source *co
...
+			return nil, nil, errors.New("new files change datatypes of some columns and column relaxation not allowed")
+		}
+	}
+
+	if len(newCols) != 0 && !allowAddition {
+		c.logger.Error("new files have new columns and column addition not allowed",
+			zap.String("files", strings.Join(names(fileNames), ",")),
+			zap.String("columns", strings.Join(keys(newCols), ",")))
+		return nil, nil, errors.New("new files have new columns and column addition not allowed")
+	}
```
begelundmuller, line 184: Can we make the errors a little richer? Imagine if a user sees this error, it should help them resolve the problem. Maybe show the first filename and all the column names?

begelundmuller (changes requested):
Two other comments:
- It looks like `read_csv_auto` will still use `sample_size` to detect delimiters and timestamp formats even if we pass `columns`. So we will still run two detections for each file. I wonder if we should just try to ingest new partitions normally, and only retry with schema relaxation if we get an error on the regular insert? (We could check for error messages matching mismatched schemas.)
- In the docs, `read_csv_auto` lists `dtypes` and `columns` separately. Are we sure that passing `columns` only is sufficient to disable the data type detection?

runtime/drivers/duckdb/connectors.go
k-anshul: I am adding that information in the log entry so a user should be able to get this information. These logs should be available on the UI as well right ?
Seems weird to put all that information in the error message.

k-anshul:
> * In the docs, `read_csv_auto` lists `dtypes` and `columns` separately. Are we sure that passing `columns` only is sufficient to disable the data type detection?

The description of `column` mentions `column names` and `column types` explicitly so it mostly should.

runtime/drivers/duckdb/connectors.go
begelundmuller: For cloud deployments people don't have access to the log messages.

So if a source breaks one day, and a dashboard stops refreshing, the errors shown in the UI will not be informative

runtime/drivers/duckdb/connectors.go
```diff
@@ -123,14 +156,158 @@ func (c *connection) ingestLocalFiles(ctx context.Context, env *connectors.Env,
...
+	from, err := sourceReader(files, format, a.ingestionProps)
+	if err != nil {
+		return err
+	}
+
+	query := fmt.Sprintf("INSERT INTO %q (SELECT * FROM %s);", a.source.Name, from)
+	err = a.Exec(ctx, &drivers.Statement{Query: query, Priority: 1})
+	if err == nil || !containsAny(err.Error(), []string{"binder error", "conversion error"}) {
+		return err
+	}
```
begelundmuller, line 212: I might be missing something, but after the first insertion error that leads to relaxation, won't this always lead to an error? I think we can solve that by using `INSERT INTO %q BY NAME (SELECT * FROM %s)` – that would also enable support for column reordering in partitions. Does that sound right?

runtime/drivers/duckdb/connectors.go
```diff
@@ -123,14 +156,158 @@ func (c *connection) ingestLocalFiles(ctx context.Context, env *connectors.Env,
...
+		}
+	}
+
+	if len(newCols) != 0 && !a.allowColAddition {
+		fileNames := strings.Join(names(fileNames), ",")
+		columns := strings.Join(missingMapKeys(srcSchema, a.tableSchema), ",")
+		a.logger.Error("new files have new columns and column addition not allowed", zap.String("files", fileNames),
+			zap.String("columns", columns))
+		return nil, fmt.Errorf("new files %q have new columns %q and column addition not allowed", fileNames, columns)
+	}
```
begelundmuller, line 292: Can remove duplicate log statements, and just log any error returned from `updateSchema` when calling it in `appendData` (unless there's something even higher in the call chain that already logs errors?)

runtime/drivers/duckdb/connectors.go
begelundmuller: Would it solve it if we only add `BY NAME` if `allow_field_relaxation` is true? By a slight stretch, you could argue that column reordering is also a kind of relaxation

begelundmuller (approved):

begelundmuller (approved):

PR #2390, opened 2023-05-10
Track bytes ingested using BIGINT

begelundmuller:


runtime/drivers/duckdb/connectors.go
```diff
@@ -92,34 +88,48 @@ func (c *connection) ingestIteratorFiles(ctx context.Context, source *connectors
...
+		return nil, err
 	}
 	if len(localPaths) == 0 {
-		return fmt.Errorf("file does not exist at %s", conf.Path)
+		return nil, fmt.Errorf("file does not exist at %s", conf.Path)
+	}
+
+	// Calculate bytes that will be ingested
+	var bytesIngested int64
+	for _, p := range localPaths {
```
k-anshul, line 113: can use `fileSize` method from the same class

k-anshul (approved):

PR #2392, opened 2023-05-10
[Dashboard] Hide/Show measure and dimensions

AdityaHegde:
This is a working branch for enabling hide/show measure and dimensions.

Things in this PR
- [x] Add a unique identifier `name` to dimensions.
- [x] Switch `property` field to `column` in dashboard config.
- [x] Use `name` instead of `column` in dashboard APIs for dimensions.
- [x] Re-enable hide/show feature.
- [x] Update the dimension display table to only show the selected measures.
- [x] Move the hide/show feature to a separate file for ease of testing and compartmentalisation.
- [x] What should happen to the selected measure in leaderboard and dimension table (for sorting) when the measure is hidden?
- [x] Backwards compatibility for older projects with and without duplicate columns selected in dimensions. It just mirrors the `column` to `name`. For duplicates it should use `<column>_1` and so on. Done in backend. Pending in UI for cloud projects that are not reconciled after release.
- [ ] ~~E2E testing using existing project.~~ This has extensive unit tests. E2E tests not necessary at this point.


runtime/services/catalog/artifacts/yaml/objects.go
```diff
@@ -338,6 +339,16 @@ func fromMetricsViewArtifact(metrics *MetricsView, path string) (*drivers.Catalo
...
 	}
 
+	// backwards compatibility where name was used as property
+	for i, dimension := range apiMetrics.Dimensions {
+		if dimension.Name == "" {
+			dimension.Name = fmt.Sprintf("dimension_%d", i)
+		}
+		if dimension.Property == "" {
+			dimension.Property = dimension.Name
+		}
```
begelundmuller: If `name == ""`, it will try to call `dimension_%d` in the underlying model during validation. That might be weird?

proto/rill/runtime/v1/catalog.proto
```diff
@@ -85,6 +85,7 @@ message MetricsView {
     string name = 1;
     string label = 2;
     string description = 3;
+    string property = 4;
```
begelundmuller, line 88: Consider `column` or `field` instead – I think those are more commonly used for table column names

runtime/services/catalog/migrator/metricsviews/metrics_views.go
```diff
@@ -74,12 +74,24 @@ func (m *metricsViewMigrator) Validate(ctx context.Context, olap drivers.OLAPSto
 
 	var validationErrors []*runtimev1.ReconcileError
 
+	dimensionNames := make(map[string]bool)
 	for i, dimension := range mv.Dimensions {
-		if _, ok := fieldsMap[strings.ToLower(dimension.Name)]; !ok {
+		if _, ok := dimensionNames[dimension.Name]; ok {
```
begelundmuller, line 79: Should we make this check case insensitive?

begelundmuller (commented):
Additionally, it probably also needs to change the metrics APIs to not use `dimension.Name` anymore

runtime/services/catalog/migrator/metricsviews/metrics_views.go
AdityaHegde: Same applies to measures as well

ericpgreen2:
I'm seeing this on the Dashboard page: 
image

Additionally, for existing dashboards, the metrics definition page freezes when I try to add a dimension.

ericpgreen2:
Now I'm no longer seeing that errored Dashboard page above, but the metrics definition page still freezes on me when i try to add a dimension to an existing dashboard. For all _new_ dashboards, adding dimensions works.

ericpgreen2:
Here's a new error I found:

1. Make a dashboard from the [Our World In Data COVID data](https://raw.githubusercontent.com/owid/covid-19-data/master/public/data/owid-covid-data.csv). Go to the dashboard.
2. Filter for any `Continent`
3. Navigate to the metrics page
4. Navigate back to the dashboard
5. The dashboard breaks and I see: 
image

begelundmuller (approved):
Approving from backend perspective

bcolloran (changes requested):
Hey Aditya, just a couple high level things so far:

First off, I'm not able to get the app to load on this branch. Video below shows that if I `run clean` and than `run dev` on this branch the app does not load (I actually go through it 3 times in the video, You don't have to watch the whole thing -- once on this branch, which doesn't work; then on main which works ok; and then on this branch again which doesn't work).

https://github.com/rilldata/rill-developer/assets/2380975/17a9bee2-5d14-4ff9-bfcf-146801bd055a

It looks like it's hanging somewhere in the welcome page redirect possibly:
image

---

Then the second thing I noticed was that a number of the "reducer"  methods attached to the `metricViewReducers` seem to be obsolete and can be removed. For example, when I string search for `toggleMeasureVisibilityByKey`, `hideAllMeasures`, `setMultipleMeasuresVisible`, those methods string names don't show up anywhere in the code base where they are being called. So please do pass through the code to remove those and look for other things like that that might be dangling.

---

AdityaHegde:
Hey @bcolloran I have updated the PR to remove old code and move tests to run on new code. Also fixed any issues with using existing project from older code. Please take a look.

@nishantmonu51 @magorlick This PR is ready for a UX experience. A couple of questions (also the reason for the test failure),
- What should happen when a new measure or dimension is added to the dashboard through the editor and the user navigates to the dashboard? Should it be visible automatically or be hidden?
- What should happen when a measure or dimension was added and a url from an older dashboard is loaded? Should the visible items be taken from the state in the url or should the new measure/dimension be shown?

bcolloran (commented):
Code looks good, but did not do an experience pass -- @nishantmonu51 @magorlick, handing this off to you two for that and answers to Aditya's product questions, happy to do another code review of any changes following from those decisions.

bcolloran (approved):

nishantmonu51:
My initial thoughts would be - 

- Read the list of dimensions/metrics from the URL state when it is present in the state, otherwise show all by default. 
- List of dimensions/metrics is only appended to the URL state when a user has explicitly interacted with the dimension/metrics selector. If a user has not taken any actions to explicitly hide any dim/metric, any newly dimensions/metric would be visible by default on dashboard view.
- Remove the state from the URL when a user has selected all dims/metrics to be shown or a user clicks on `select all`  

@AdityaHegde : Does this help handle the boundary cases you asked for ?

magorlick:
Thanks for this work @AdityaHegde .  I did an experience pass and the feedback I provided on my last pass has been addressed.  I agree with @nishantmonu51 's feedback on your open questions. We should show newly added measures and dimensions by default in addition to whatever the prior state was on the dashboard when they navigate back.

I left a few more [notes on the experience](https://www.notion.so/rilldata/Ability-to-hide-measures-and-dimensions-from-the-dashboard-f416bfaddda74e5285ec8bc33161a2c0?pvs=4#c4beb00ceee94fcb84a48fbb3f370e51) and how it could be cleaner when no measure or no dimensions were selected, but those notes are P2 and not blocking merging this in.

AdityaHegde:
Thanks @bcolloran @nishantmonu51 @magorlick 

Yes @nishantmonu51 That should be enough for all edge cases. I have made the changes, please take a look.

PR #2398, opened 2023-05-10
resolve scroll issue. fixes #2395

### LINKED ISSUE #2395: Dashboard: Make layout scrolling independent of container size
The cloud version of teh dashboard has a scrollbar for the entire page, while Rill Developer doesn't. That's because we're not using the same layout logic in these two interfaces.

@ericpgreen2 attempted to fix by making the change below

```
section {
    height: 100vh; // should be 100%
    ...
}
```

But in making that change, the line charts aren’t scrolling correctly. That's because we'll have to change other parts of the layout CSS to make the measure and leaderboard boxes fill the available screen and overflow auto.

This shouldn't be a lot of work to get right most likely but will require some refactor.

bcolloran:
@ericpgreen2  and @hamilton, in my local testing I think this fixes the scrolling problem in cloud while preserving the behavior we want in developer. In the video below, when my mouse is hovering above the header I'm going nuts with the scroll wheel, and there is no scroll available at that level of the hierarchy.

https://github.com/rilldata/rill-developer/assets/2380975/860d1eaa-5932-4dcb-a9d0-edfc2520e91e

So I think this should be a simple fix, but with that said please please do a QA pass on cloud and dev to make sure I haven't missed anything.

### DIFF
--- a/web-admin/src/routes/+page.svelte
+++ b/web-admin/src/routes/+page.svelte
@@ -27,29 +27,31 @@
   <title>Home - Rill</title>
 </svelte:head>
 
-<AuthRedirect>
-  <section
-    class="flex flex-col mx-8 my-8 sm:my-16 sm:mx-16 lg:mx-32 lg:my-24 2xl:mx-64 mx-auto"
-  >
-    <h1 class="text-4xl leading-10 font-light mb-2">
-      Hi {getFirstNameFromDisplayName($user.data.user.displayName)}!
-    </h1>
-    <div class="flex flex-row gap-x-7 flex-wrap">
-      <div class="md:w-1/2">
-        {#if $orgs.isSuccess}
-          {#if !hasAnOrganization}
-            <WelcomeMessage />
-          {:else}
-            <h3 class="text-base leading-6 font-normal text-gray-500 mb-3">
-              Check out your dashboards below.
-            </h3>
-            <OrganizationList />
+<div class="overflow-auto h-full">
+  <AuthRedirect>
+    <section
+      class="flex flex-col mx-8 my-8 sm:my-16 sm:mx-16 lg:mx-32 lg:my-24 2xl:mx-64 mx-auto"
+    >
+      <h1 class="text-4xl leading-10 font-light mb-2">
+        Hi {getFirstNameFromDisplayName($user.data.user.displayName)}!
+      </h1>
+      <div class="flex flex-row gap-x-7 flex-wrap">
+        <div class="md:w-1/2">
+          {#if $orgs.isSuccess}
+            {#if !hasAnOrganization}
+              <WelcomeMessage />
+            {:else}
+              <h3 class="text-base leading-6 font-normal text-gray-500 mb-3">
+                Check out your dashboards below.
+              </h3>
+              <OrganizationList />
+            {/if}
           {/if}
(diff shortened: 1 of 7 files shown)


ericpgreen2:
Hey Brendan, on Cloud I tested non-dashboard screens, like the homepage (`/`) and the project page (`/{org}/{project}`), and this removes scroll behavior on them. It feels like it's due to your changes in `web-admin/src/routes/+layout.svelte`

Here's a project page with a bunch of fake logs, and no scroll.
image

Otherwise works on Cloud and Local dashboards!

bcolloran:
@ericpgreen2 Thanks so much for catching that! Deeply appreciate the thorough QA on this, because I'm not yet at all familiar with the different routes and edge cases that that can come up in the admin and cloud environments.

I've updated the PR to make scrolling the responsibility of the sub pages at the different routes. Unfortunately, we can't just setting everything other than the header to overflow-auto because of the dashboard. The dashboard needs to be in a container of fixed size so that it can manage the scrolling of its child elements, so the container for everything other than the header needs to be a fixed size and have overflow set to hidden, which limits the size the dashboard container can expand to fill. But since the dashboard component needs this in order to take responsibility for it's scrolling behavior, we also need to make sure that other components that can end up in that slot take responsibility for their scrolling behavior. In this case, pushing that responsibility to the sub page routes is seeming to do the trick, except for in the case of the 404 pages.

I'm punting on the 404 pages for for now because it turns out that the CTA container style that you mentioned is actually pretty tricky, and I don't want to dig into that container since it's used in other places. I also think that in the case of the 404 in particular the scrolling is less of an issue because it does cause a broken layout at a very low Y height for the page, but that is something of an edge case, whereas the other scrolling issues for the dashboard and for the list of errors that you found seem like quite plausible mainstream cases. So we should loop back and look at the CTA container in all of its contexts because I bet it breaks scrolling layouts in other places that we haven't noticed yet, but for now I think it makes sense to try to get this in. That's just my $0.02 though -- if you want to do it all in one go I'm happy to keep digging here.

Please do let me know if you find anything else on any routes or conditions that I've missed-- again, not feeling confident that I know all of the surface area for this. Thanks so much! :-)

web-admin/src/routes/+page.svelte
```diff
@@ -27,29 +27,31 @@
...
-      <div class="md:w-1/2">
-        {#if $orgs.isSuccess}
-          {#if !hasAnOrganization}
-            <WelcomeMessage />
-          {:else}
-            <h3 class="text-base leading-6 font-normal text-gray-500 mb-3">
-              Check out your dashboards below.
-            </h3>
-            <OrganizationList />
+<div class="overflow-auto h-full">
```
ericpgreen2, line 30: Given you're using this in multiple places, what about making this a component `VerticalScrollContainer` (or something similarly named)?

```svelte
<div class="overflow-auto h-full">
    <slot />
</div>
```

Then, in this case, it'd be clear that the `<AuthRedirect>` should be a parent of `<VerticalScrollContainer>`, not a child of it.

ericpgreen2:
All pages look good on QA! Thanks for tackling.

I'll approve so you can address my code cleanup comment and merge.

ericpgreen2 (approved):

PR #2400, opened 2023-05-10
adds a priority number for the time series query, which makes it appear much faster

hamilton:


hamilton:
Aditya approves but can't merge since he's afk, merging it as is.

PR #2404, opened 2023-05-11
Fix "organization is undefined" bug

ericpgreen2:
PR #2297 affected all our usage of TanStack Query. Now, by design, the `data` object is `undefined` until data is returned from the server. In many places previously, our code assumed the `data` object would be `{}` until data is returned from the server.

This PR fixes one file where we didn't switch the code to account for `data = undefined`.


bcolloran (approved):
always love some optional chaining :-)

PR #2414, opened 2023-05-12
Adding user id to telemetry if logged in

AdityaHegde:
We want to track user id in telemetry if the user has logged in.

- Adding user_id to deploy command
- Adding user_id to rill developer if user has logged in

### DIFF
--- a/cli/pkg/local/app.go
+++ b/cli/pkg/local/app.go
@@ -298,7 +298,7 @@ func (a *App) ReconcileSource(sourcePath string) error {
 	return nil
 }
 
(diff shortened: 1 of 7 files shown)


cli/pkg/local/app.go
```diff
@@ -298,7 +298,7 @@ func (a *App) ReconcileSource(sourcePath string) error {
 	return nil
 }
 
-func (a *App) Serve(httpPort, grpcPort int, enableUI, openBrowser, readonly bool) error {
+func (a *App) Serve(httpPort, grpcPort int, enableUI, openBrowser, readonly bool, userID string) error {
```
AdityaHegde, line 301: Added only in server since it is only used during `rill start`

PR #2417, opened 2023-05-12
Adding Vitest to web-common

AdityaHegde:
Right now our jest tests on web-common do not work. Taking this opportunity to move to vitest. This will drastically cut down our test configuration.

This disabled a few tests that are either failing or using jest's mocking feature.


djbarnwal (approved):
Looks good, thanks for the PR

PR #2418, opened 2023-05-12
Fixing restoring dashboard state from url

AdityaHegde:
One of the recent commits broke restoring dashboard state from url. This fixes it.


ericpgreen2 (approved):
I'll approve since it fixes the bug, but it feels fragile. What if we created a wrapper component named something like `URLStateProvider` or `URLStateManager` that contained this logic? It might also make sense to include the code in `StateSyncManager.ts`.

And we could use it like:
```svelte
<URLStateProvider>
  <Dashboard />
</URLStateProvider>
```

web-common/src/features/dashboards/proto-state/DashboardStateProvider.svelte
```diff
@@ -0,0 +1,18 @@
...
+
+  export let metricViewName: string;
+
+  $: metricsExplorer = useDashboardStore(metricViewName);
+  $: stateSyncManager = new StateSyncManager(metricViewName);
+  $: if ($metricsExplorer) {
+    stateSyncManager.handleStateChange($metricsExplorer);
+  }
+  $: if ($page) {
+    stateSyncManager.handleUrlChange();
```
ericpgreen2, line 14: Not for this PR, but just a note we should make this a pure function.

ericpgreen2 (approved):

PR #2419, opened 2023-05-12
Recover from invalid dashboard state

AdityaHegde:
~~Needs changes from #2417 for tests~~ Done

Dashboard url can get stale when dimensions or measures are removed. This PR makes sure we recover from such invalid states.


web-common/src/features/dashboard-stores.spec.ts
```diff
@@ -136,4 +144,88 @@ describe("dashboard-stores", () => {
       AdBidsBidPriceMeasure
     );
   });
+
+  describe("Restore invalid state", () => {
+    it("Restore invalid filter", () => {
+      createAdBidsInStore();
+      metricsExplorerStore.toggleFilter(
+        AdBidsName,
```
ericpgreen2, line 152: This should be an all-caps constant. And a more descriptive name could be `ADBIDS_DASHBOARD_NAME`.

web-common/src/features/dashboard-stores.spec.ts
```diff
@@ -136,4 +144,88 @@ describe("dashboard-stores", () => {
...
+        AdBidsPublisherDimension,
```
ericpgreen2, line 153: Should be an all-caps constant, too.

web-common/src/features/dashboards/dashboard-stores.ts
```diff
@@ -177,6 +180,26 @@ const metricViewReducers = {
         metricsExplorer.visibleDimensionKeys = new Set(
           metricsView.dimensions.map((dim) => dim.name)
         );
+
+        const dimensionsMap = getMapFromArray(
```
ericpgreen2, line 184: I'm not following why we need a map. Can't we just use the array? Can you add a comment if this is an optimization?

web-common/src/features/dashboards/dashboard-stores.ts
ericpgreen2, line 1: The `sync` function is getting long and hard to follow. Can you break it up either functionally or with comments to describe the different sections?

ericpgreen2:
Added some comments.

And so I'm 100% clear –  what are the steps to reproduce the bug on `main` that this PR fixes?

AdityaHegde:
These steps shouldnt break the UI
1. Select some filters, copy the URL, remove the dimension, visit the copied URL
2. View dimension table, copy the URL, remove the dimension, visit the copied URL
3. Select a measure for leaderboard, copy the URL, remove the measure, visit the copied URL

ericpgreen2 (approved):
Nice, those comments in `syncFromURL` help a lot.

PR #2423, opened 2023-05-12
Set the mobile viewport width to 1024px

ericpgreen2:
Currently, the Cloud UI is a broken experience on mobile because we haven't designed a fully responsive app. To make the dashboards responsive would require a big effort and constant maintanence.

This PR is a stop-gap solution for mobile users.

Contributes to #2109.


begelundmuller (commented):
Aren't many smaller laptops screens smaller than 1400px? Wondering if this impacts rendering on these

begelundmuller:
I don't really know what the right trade-off is here :)

ericpgreen2:
I think `1024px` is a good pick. Just updated the PR.

Currently, depending on the width of the leaderboards, neither `1024px` nor `1150px` looks good 100% of the time. `1024px` is more conventional than an arbitrary number that matches the state of the UI, which will be a moving target as it evolves over time.

begelundmuller (approved):
LGTM

PR #2425, opened 2023-05-15
add dash in user code for readability

pjain1:


begelundmuller (commented):
I think the spec notes the use of `-` as a way to make it look nicer, but that it should be optional. So instead of storing it in the DB, should we instead add/trim the `-` at the API layer?

pjain1:
yeah this seems fasted way, otherwise has to add and remove dash during retrieval and checking

pjain1:
will add at api layer

admin/server/auth/device_code.go
```diff
@@ -145,7 +148,9 @@ func (a *Authenticator) handleUserCodeConfirmation(w http.ResponseWriter, r *htt
 	}
 	userID := claims.OwnerID()
 
-	authCode, err := a.admin.DB.FindPendingDeviceAuthCodeByUserCode(r.Context(), userCode)
+	// Remove "-" from user code
+	dbUserCode := userCode[:4] + userCode[5:]
```
begelundmuller, line 152: The user code might be shorter (it's user input), so it's safer to just do: `userCode = strings.ReplaceAll(userCode, "-", "")`

begelundmuller (changes requested):

admin/server/auth/device_code.go
pjain1: In lines below to this, there are multiple error messages in which `userCode` so keeping it as it is and doing the replace to get a new string

pjain1: not sure if you wanted to use same `userCode` variable or think `ReplaceAll` is more safer (why ?) because I think slicing is more efficient than replacing.

pjain1: I see what you mean, to prevent against wrong user input, makes sense.

begelundmuller (approved):

PR #2427, opened 2023-05-15
e2e test config and sources

skokenes:
This PR is the first step in re-enabling our e2e tests and fortifying them. It contains the following:
- a PR template for the applications team, which we will later set up as default when creating a PR
- turns the sources e2e tests back on
- Implements a build step of the local repo version of Rill in a jest globalSetup function, as well as cleans up with a jest globalTeardown function
- sets up web-common testing with a coverage report. For now we will not commit the coverage report to git


web-local/.jest/setup.js
```diff
@@ -0,0 +1,17 @@
+import { spawn } from "child_process";
```
AdityaHegde, line 1: We should move this to package.json and avoid dependancy on jest's features. Something like `setup_binary.sh && (run tests || cleanup_binary.sh)`

web-local/.jest/setup-e2e-rill.sh
```diff
@@ -0,0 +1,17 @@
...
+
+while getopts "s" arg; do
+  case $arg in
+    s) skipUIBuild=true;;
+  esac
+done
+
+if ! $skipUIBuild ; then
+    ECHO "Building UI"
+    make cli.prepare -C ../..
```
AdityaHegde, line 12: This ends up building the UI twice in CI. We can remove the build from the job config.

web-local/.jest/setup.js
skokenes: Not sure I follow - why do you want to avoid using jest's features? Isn't that the point of having a testing framework?

We are already using jest's features, see our existing jest config https://github.com/rilldata/rill-developer/blob/main/web-local/jest.config.json#LL17C2-L20C5

web-local/.jest/setup-e2e-rill.sh
skokenes: For CI, you can use the environment variable SKIP_UI_BUILD=true to skip it, like `SKIP_UI_BUILD=true npm run test`

AdityaHegde: Yes that part is missing in this PR. The job config needs to be updated.

web-local/.jest/setup.js
AdityaHegde: I think we should be moving to vitest. It is a lot better since we are already using vite.

AdityaHegde (approved):

web-local/.jest/setup-e2e-rill.sh
skokenes: Done

PR #2429, opened 2023-05-15
hides the measure and dimension selector until the pending state work is complete

### LINKED ISSUE #2428: temp fix: hide the show/hide measure & dimension selector
We're pretty behind schedule on implementing the state work for the show/hide measures & dims feature, so until then, this PR will disable showing the menus.

See here for the remaining work:

https://www.notion.so/rilldata/Ability-to-hide-measures-and-dimensions-from-the-dashboard-f416bfaddda74e5285ec8bc33161a2c0

hamilton:
closes #2428


ericpgreen2:
A couple scenarios I'm seeing:
- When I select a time range, refresh the page, the dashboard resets to the default time range
- When I filter for a dimension, refresh the page, the dashboard resets to no-dimensions-filtered

But maybe the above behavior is fixed by https://github.com/rilldata/rill-developer/pull/2418.

It's hard to review this with the above PR outstanding. I'd have more confidence in the PR if it were a pure `git revert` on whatever commit added the functionality you're trying to remove.

ericpgreen2:
Approving since I did a little QA pass and looks good, but these code files are really long, so it's hard to follow exactly what you've commented out. The fact that we can comment-out this code to turn off a feature is a good hint that maybe we can move this code to standalone components / its own file in the next pass.

ericpgreen2 (approved):

hamilton:
@ericpgreen2 I agree, these files are way, way too long. We need to start from scratch

PR #2430, opened 2023-05-15
Druid-based unit tests should use prebuilt Docker image

egor-ryashin:
Depends on https://github.com/rilldata/rill-developer/pull/2432


runtime/drivers/druid/druid_test.go
```diff
@@ -86,13 +86,9 @@ func TestDruid(t *testing.T) {
...
 		Started: true,
 		ContainerRequest: testcontainers.ContainerRequest{
-			ExposedPorts: []string{"8081/tcp", "8082/tcp"},
 			WaitingFor:   wait.ForHTTP("/status/health").WithPort("8081"),
-			FromDockerfile: testcontainers.FromDockerfile{
-				Context:       ".",
-				Dockerfile:    "Dockerfile",
-				PrintBuildLog: true,
-			},
+			Image:        "druid-micro:latest",
```
begelundmuller, line 90: I think this will only work on your local because the image name is cached. It needs the full path of the image. Let's also pin to an exact version to prevent caching issues. So I think it will be `gcr.io/rilldata/druid-micro:25.0.0`?

begelundmuller (changes requested):

begelundmuller (approved):

PR #2432, opened 2023-05-16
Build Druid image for unit tests

egor-ryashin:


.github/workflows/druid-test-image.yml
```diff
@@ -0,0 +1,33 @@
...
+      - name: Build & Publish Rill docker image
+        run: |-
+          cd drivers/druid
+
+          gcloud auth configure-docker
+
+          docker build -t gcr.io/rilldata/test-druid:${GITHUB_SHA} .
+          docker tag gcr.io/rilldata/test-druid:${GITHUB_SHA} gcr.io/rilldata/test-druid
+          docker push gcr.io/rilldata/test-druid:${GITHUB_SHA}
+          docker push gcr.io/rilldata/test-druid
```
himadrisingh, line 33:
```suggestion
          docker push gcr.io/rilldata/test-druid

```

.github/workflows/druid-test-image.yml
```diff
@@ -0,0 +1,33 @@
+name: Build Druid image for tests
+
+on:
+  pull_request:
+    paths:
+      - "drivers/druid/Dockerfile"
```
himadrisingh, line 6: do we need it only for PRs?

egor-ryashin: Yes, it should be rebuilt on change, building on merging doesn't matter.

.github/workflows/druid-test-image.yml
```diff
@@ -0,0 +1,36 @@
+name: Build Druid image for tests
+
+on:
+  pull_request:
+    paths:
+      - "scripts/test-druid-micro.dockerfile"
+      - ".github/workflows/druid-test-image.yml"
```
begelundmuller, line 7: Should only run on merge to main, something like:
```
on:
  push:
    branch:
      - main
    paths:
      - ...
```

.github/workflows/druid-test-image.yml
```diff
@@ -0,0 +1,36 @@
...
+      - name: Build & Publish Rill docker image
+        run: |-
+          cd scripts
+
+          gcloud auth configure-docker
+
+          docker build -f test-druid-micro.dockerfile --build-arg DRUID_VERSION=${{ env.DRUID_VERSION }} -t gcr.io/rilldata/test-druid-micro:${{ env.DRUID_VERSION }} .
+          docker tag gcr.io/rilldata/test-druid-micro:${{ env.DRUID_VERSION }} gcr.io/rilldata/test-druid-micro
+          docker push gcr.io/rilldata/test-druid-micro:${{ env.DRUID_VERSION }}
+          docker push gcr.io/rilldata/test-druid-micro
```
begelundmuller, line 34: What about `druid-micro`, seems like that's what the image contains? Having `test` in the name sounds like the image runs tests, but it doesn't.

If you want `test` in the name, can we have it after the `druid` prefix (like `druid-micro-test`)? Having domain-scoped prefixes makes things easier to navigate/discover in listings

begelundmuller (changes requested):

.github/workflows/druid-test-image.yml
egor-ryashin: That way if one changes tests that require image update then it will require 2 separate PRs (one for the Dockerfile to be merged and the other for test changes). Instead if you build on push to PR the changes can be packaged in a single PR.

.github/workflows/druid-test-image.yml
egor-ryashin: `test` describes the purpose actually (ie it doesn't affect production functionality).

begelundmuller (approved):

PR #2434, opened 2023-05-16
Handle Githubapp uninstalls/suspends

### LINKED ISSUE #2411: Runtime: Fix issues when Github installation has been removed
We are seeing some odd behavior in the runtime for instances where the Github installation has been removed. Examples:

1. There are calls to `ListCatalogEntries` that run for an insane amount of time. For example, [here's one](https://rill.datadoghq.com/apm/traces?query=%40_top_level%3A1%20service%3Aruntime-server%20env%3Astage%20status%3Aerror&cols=core_service%2Ccore_resource_name%2Clog_duration%2Clog_http.method%2Clog_http.status_code&historicalData=true&messageDisplay=inline&sort=desc&spanID=12800487505332662021&spanType=service-entry&timeHint=1683661872286&trace=AgAAAYgCEVyeqCzGUAAAAAAAAAAYAAAAAEFZZ0NFWGQxQUFCNmUxQjdoell2NWtPegAAACQAAAAAMDE4ODAyMTctNWU2Mi00ZDEyLWIzMzgtOThmMjRhMjZhNmIz&traceID=12917386887318019536&start=1683281672699&end=1683886472699&paused=false) that failed after 53 minutes!
2. Here's a place I encountered the bug from the CLI:
```
> rill project reconcile --project snl-demo --reset
Error: rpc error: code = InvalidArgument desc = could not refresh installation id 37079251's token: received non 2xx response status "404 Not Found" when fetching https://api.github.com/app/installations/37079251/access_tokens (InvalidArgument)
```
3. When visiting these projects in the UI, the UI just hangs, showing no content or error messages

Task list:

- [ ] There's some exponential retry logic wrapping Github in `runtime/drivers/github`. It should not trigger on 404 errors like this, and it should never be able to run for more than ~1 minute (how is 53 minutes even possible?). Maybe we should remove the retries altogether, return the Github error, and instead let the user manually run `rill project reconcile` again?
- [ ] We should not trigger a `git pull` (calls to `RepoStore.Sync`) in other calls than `Reconcile`. In `ListCatalogEntries` and `ListFiles`, it should just return whatever is available. If no clone/pull has run it, the results of those RPCs should just be empty.
- [ ] When an installation is removed, we should receive a webhook from Github. Use it to delete any related projects.

(Please consider alternative approaches or other possible related bugs to the above.)

k-anshul:
closes #2411


admin/database/postgres/postgres.go
```diff
@@ -277,6 +277,15 @@ func (c *connection) FindProjectsByGithubURL(ctx context.Context, githubURL stri
 	return result, nil
 }
 
+func (c *connection) FindProjectsByGithubInstallationID(ctx context.Context, id int64) ([]*database.Project, error) {
+	result := make([]*database.Project, 0)
```
begelundmuller, line 281: Should just be `var res []*database.Project` (no allocation)

admin/github.go
```diff
@@ -216,15 +221,88 @@ func (s *Service) processGithubPush(ctx context.Context, event *github.PushEvent
...
+func (s *Service) deleteProjectsForInstallation(ctx context.Context, id int64) error {
+	// Find Rill project for installationID
+	projects, err := s.DB.FindProjectsByGithubInstallationID(ctx, id)
+	if err != nil {
+		if errors.Is(err, database.ErrNotFound) {
+			// App is removed from repo not currently deployed. Do nothing.
+			return nil
+		}
+		return err
+	}
```
begelundmuller, line 268: I don't think `database.ErrNotFound` can/should be returned from DB queries returning a list (since the empty list is not an error when selecting multiple)

admin/github.go
```diff
@@ -216,15 +221,88 @@ func (s *Service) processGithubPush(ctx context.Context, event *github.PushEvent
...
+	// Find Rill project matching the repo that was pushed to
+	githubURL := *repo.HTMLURL
+	projects, err := s.DB.FindProjectsByGithubURL(ctx, githubURL)
+	if err != nil {
+		if errors.Is(err, database.ErrNotFound) {
+			// App is removed from repo not currently deployed. Do nothing.
+			return nil
+		}
+		return err
+	}
```
begelundmuller, line 294: Same note about the not found err

admin/github.go
```diff
@@ -216,15 +221,88 @@ func (s *Service) processGithubPush(ctx context.Context, event *github.PushEvent
...
+
+	var multiErr error
+	for _, p := range projects {
+		err := s.TeardownProject(ctx, p)
+		if err != nil {
+			if !errors.Is(err, database.ErrNotFound) {
```
begelundmuller, line 300: Maybe don't skip this case here? If the project was returned from DB, then it would be nice to log if it's missing here (inconsistent)

admin/github.go
```diff
@@ -216,15 +221,88 @@ func (s *Service) processGithubPush(ctx context.Context, event *github.PushEvent
...
+		// the github installation ID will change if user re-installs the app deleting the project for now
+		installation := event.GetInstallation()
+		if installation == nil {
+			return fmt.Errorf("nil installation")
+		}
 
+		if err := s.deleteProjectsForInstallation(ctx, installation.GetID()); err != nil {
+			s.logger.Error("failed to delete project for installation", zap.Int64("installation_id", installation.GetID()), zap.Error(err))
+			return err
+		}
```
begelundmuller, line 236: 1. Wrap with start/stop info logs
2. Maybe prefix all the new logs in this file with "github webhook: ...`
3. Missing `observability.ZapCtx(ctx)` field in all the logs

admin/github.go
```diff
@@ -216,15 +221,88 @@ func (s *Service) processGithubPush(ctx context.Context, event *github.PushEvent
...
+		// no handling as of now
+	case "removed":
+		var multiErr error
+		for _, repo := range event.RepositoriesRemoved {
+			if err := s.deleteProjectsForRepo(ctx, repo); err != nil {
+				multiErr = multierr.Combine(multiErr, err)
+				s.logger.Error("failed to delete project for repo", zap.String("repo", *repo.HTMLURL), zap.Error(err))
+			}
+		}
+		return multiErr
```
begelundmuller, line 254: 1. Also wrap with info logs, maybe something like `github webhook: processing removed repositories`

admin/github.go
```diff
@@ -216,15 +221,88 @@ func (s *Service) processGithubPush(ctx context.Context, event *github.PushEvent
...
+		return err
+	}
+
+	var multiErr error
+	for _, p := range projects {
+		err := s.TeardownProject(ctx, p)
+		if err != nil {
+			if !errors.Is(err, database.ErrNotFound) {
+				multiErr = multierr.Combine(multiErr, err)
+				s.logger.Error("unable to delete project", zap.String("project_id", p.ID), zap.Error(err))
```
begelundmuller, line 302: This log is redundant because the caller is logging the returned multierr. You can add the `project_id` to the error message if you want it to get printed (something like `fmt.Errorf("unable to delete project %q: %w", p.ID, err)`)

runtime/drivers/github/github.go
```diff
@@ -137,38 +124,40 @@ func (c *connection) MigrationStatus(ctx context.Context) (current, desired int,
...
-	_, err = git.PlainClone(c.tempdir, false, &git.CloneOptions{
-		URL:           cloneURL,
-		ReferenceName: plumbing.NewBranchReferenceName(c.dsn.Branch),
-		SingleBranch:  true,
-	})
-	if err != nil {
-		return err
+	// no need to pull again if clone tried within same request
+	if cloneTried {
+		return cloneErr
```
begelundmuller, line 147: Not sure what this is supposed to do, but pretty sure it doesn't work. The goroutine will likely get scheduled after the `if` check and it will try to pull instead of clone.

I don't think we need to overthink this, just use a lock on the whole `pull` function. If you want to deduplicate requests, use a `TryLock` first and if it fails, you can do a normal `Lock` and return immediately when you get it (maybe there's an even simpler way).

runtime/drivers/github/repo.go
```diff
@@ -99,10 +101,19 @@ func (c *connection) Delete(ctx context.Context, instID, filePath string) error
...
 func (c *connection) Sync(ctx context.Context, instID string) error {
-	r := retrier.New(retrier.ExponentialBackoff(retryN, retryWait), nil)
-	err := r.Run(func() error { return c.pull(ctx) })
+	r := retrier.New(retrier.ExponentialBackoff(retryN, retryWait), retrier.BlacklistClassifier{errUnauthorized})
+	err := r.Run(func() error {
+		err := c.pull(ctx)
+		ghinstallationErr := &ghinstallation.HTTPError{}
+		if errors.As(err, &ghinstallationErr) && ghinstallationErr.Response != nil && ghinstallationErr.Response.StatusCode == 404 {
+			return errUnauthorized
+		}
```
begelundmuller, line 110: I thought we should not retry on any 4xx except for rate limit error?

begelundmuller (changes requested):

admin/github.go
k-anshul: This was more for the case when project doesn't have any active deployment. As noted in other comments it won't return this error so removing this check.

runtime/drivers/github/repo.go
k-anshul: I thought we concluded on only failing for unauthorised errors. I can change this.

runtime/drivers/github/github.go
k-anshul: > Not sure what this is supposed to do, but pretty sure it doesn't work. The goroutine will likely get scheduled after the if check and it will try to pull instead of clone.

I think you misunderstood. The clone happens in the same goroutine. 

The expectation is that only the first request should clone the project and subsequent requests pull. 
Even if we take lock on the whole pull function, we would still need to keep a boolean flag to denote that its been cloned already so it really makes sense to just use an atomic boolean and check its value instead. 
I can remove that `cloneTried` thing and put all of this logic in the clone function. Here is how the impl would like (Although this will both clone and pull for the first request which should be fine I guess) :


```

func (c *connection) pull(ctx context.Context) error {
	if !c.cloned.Load() {
		if err := c.clone(ctx); err != nil {
			return err
		}
	}
	**** pull logic ***
}

// clone runs the initial clone of the repo.
func (c *connection) clone(ctx context.Context) error {
	c.mu.Lock()
	defer c.mu.Unlock()
	if c.cloned.Load() {
		// cloned already
		return nil
	}

	cloneURL, err := c.cloneURL(ctx)
	if err != nil {
		return err
	}

	_, err = git.PlainClone(c.tempdir, false, &git.CloneOptions{
		URL:           cloneURL,
		ReferenceName: plumbing.NewBranchReferenceName(c.dsn.Branch),
		SingleBranch:  true,
	})
	if err != nil {
		return err
	}

	c.cloned.Store(true)
	return nil
}

```

begelundmuller: I see now there's no `go` call, but it still doesn't really make sense. There is no need here for anything fancy apart from locking the whole call (it's just as important to have a lock around the pull as it is for the clone), and maybe doing a `TryLock` to deduplicate calls as described earlier.

k-anshul: The reason there is a lock on the clone is that we don't want clone to run multiple times. It wasn't required previously because it was happening on `open` call and there was no chance for it to run concurrently. 
For pull I didn't add any lock since it wasn't there previously as well. I think you are right here. We need to also lock around pull. There is a chance that it can leave the local clone in a weird state. 

I don't need understand the use of `TryLock`. For clone we just don't want `deduplication` we only want it to run just once(successfully) during the complete lifecycle.

begelundmuller: I mean it can just be something like this:

```go
func (c *connection) pull(ctx context.Context) error {
  var deduplicated bool
  if !c.mu.TryLock() {
    deduplicated = true
    c.mu.Lock()
  }
  defer c.mu.Unlock()

  if deduplicated {
    return nil
  }

  if !c.cloned {
    // do clone
    return nil
  }

  // do pull
  return nil
}
```

begelundmuller: To return deduplicated errors, you can just store the last pull err as a var

begelundmuller: You can also use a singleflight for it with a const key if you want

begelundmuller (approved):

PR #2446, opened 2023-05-17
Adding a null check for start and end for comparisions

AdityaHegde:
Since we have a lot of code in TimeControls there is a race condition which ends with adding a name for time range but not start and end. This fixes the hangup by adding a null check.

We should really move the code out of that component. Prerequisite PR for it: https://github.com/rilldata/rill-developer/pull/2320


hamilton:
Thanks – this unfortunately won't fix the underlying issue which is causing other bugs. We'll be reviewing + merging #2449

PR #2449, opened 2023-05-17
fixes non-reactive time range initialization

hamilton:
This is a temporary fix that resolves a few bugs relating to the dashboard. Previously, we were not correctly initializing the start / end date, causing issues when using the time controls. This was a recent issue introduced in the last week or so.

We will be refactoring this component to remove rough edges around state initialization. Until then, we need a patch release.


ericpgreen2 (approved):
Looks good 👍

PR #2451, opened 2023-05-17
remove useless var `leaderboards` and useless dispatch that sets it

bcolloran:
While working on other stuff, I noticed that in the file `LeaderboardDisplay.svelte`, the variable `leaderboards` is set, but never passed to anything else nor used to derive anything else. Search ctrl+f in that file to convince yourself the variable only show up in 4 places, and that these 4 places just update the variable without any other effects.

Moreover, in  `Leaderboard.svelte`, we dispatch an event that does nothing other than set the value of `leaderboards` in  `LeaderboardDisplay.svelte`, so that is useless too. Search all files for "leaderboard-value" to convince yourself that this event is only captured by `LeaderboardDisplay.svelte` and only sets the useless variable.

Assigning @djbarnwal -- This should be quick.

### DIFF
--- a/web-common/src/features/dashboards/leaderboard/Leaderboard.svelte
+++ b/web-common/src/features/dashboards/leaderboard/Leaderboard.svelte
@@ -22,7 +22,6 @@
     MetricsViewMeasure,
   } from "@rilldata/web-common/runtime-client";
   import { useQueryClient } from "@tanstack/svelte-query";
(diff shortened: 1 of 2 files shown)


djbarnwal (approved):
Thanks for the PR. I recall we were using these for formatting the leaderboard and before that for something else. Thanks for finding these.

PR #2456, opened 2023-05-18
Sync repo in file APIs

begelundmuller:


runtime/services/catalog/catalog.go
```diff
@@ -61,7 +63,9 @@ func (s *Service) FindEntry(ctx context.Context, name string) (*drivers.CatalogE
 	if err != nil {
 		return nil, err
 	}
+	s.Meta.lock.RLock()
 	s.Meta.fillDAGInEntry(entry)
+	s.Meta.lock.RUnlock()
```
k-anshul, line 68: Should we add this as part of defer in case `fillDAGInEntry` panics ?

k-anshul (approved):

PR #2461, opened 2023-05-19
separate file logger for local

k-anshul:
- Adds a separate file logger for local application
- Adds logs for ingestion


nishantmonu51:
tested and works as expected, Will look at more detailed logs.

nishantmonu51 (approved):
LGTM, have tested it on my local rill setup as well.

runtime/connectors/blob/blobdownloader.go
```diff
@@ -182,11 +185,13 @@ func (it *blobIterator) NextBatch(n int) ([]string, error) {
 			// Collect metrics of download size and time
 			startTime := time.Now()
 			defer func() {
+				duration := time.Since(startTime)
+				it.logger.Info("download complete", zap.String("object", obj.obj.Key), zap.Float64("took_seconds", duration.Seconds()))
```
begelundmuller, line 189: Use `zap.Duration` instead, and make field `duration` instead of `took_seconds`

runtime/drivers/duckdb/connectors.go
```diff
@@ -61,11 +62,15 @@ func (c *connection) ingest(ctx context.Context, env *connectors.Env, source *co
...
+		st := time.Now()
+		c.logger.Info("ingesting files", zap.String("source", source.Name), zap.Strings("files", files))
 		if err := c.ingestIteratorFiles(ctx, source, files, appendToTable); err != nil {
 			return nil, err
 		}
 
-		summary.BytesIngested += fileSize(files)
+		size := fileSize(files)
+		summary.BytesIngested += size
+		c.logger.Info("ingested files", zap.String("source", source.Name), zap.Strings("files", files), zap.Int64("bytes_ingested", size), zap.Float64("took_seconds", time.Since(st).Seconds()))
```
begelundmuller, line 73: Use `zap.Duration`

begelundmuller (changes requested):
1. Also missing `observability.ZapCtx(ctx)` in most logs
2. Add `With(zap.String("instance", instanceID), zap.String("driver", "duckdb"))` or something similar to the logger used in DuckDB (add once so it's in all log messages, including those on the logger passed down to the connectors)

begelundmuller (approved):

PR #2465, opened 2023-05-19
Improving the Rill CLI first-run experience

rakeshsharma14317:


Makefile
```diff
@@ -4,7 +4,7 @@ docs.generate:
 	# Sets version to the latest tag to simulate a production build, where certain commands are hidden.
 	rm -rf docs/docs/reference/cli
 	if [ -f ~/.rill/config.yaml ]; then mv ~/.rill/config.yaml ~/.rill/config.yaml.tmp; fi;
-	go run -ldflags="-X main.Version=$(shell git describe --tags `git rev-list --tags --max-count=1`)" ./cli docs generate docs/docs/reference/cli/
+	go run -ldflags="-X main.Version=$(bash git describe --tags `git rev-list --tags --max-count=1`)" ./cli docs generate docs/docs/reference/cli/
```
begelundmuller, line 7: Hey, this did not work. It has set the version to an empty string, as you see, the generated CLI docs now include hidden commands, such as `rill admin`, which they shouldn't.

begelundmuller (changes requested):

Makefile
rakeshsharma14317: This is done as part of https://github.com/rilldata/rill-developer/pull/2483

cli/cmd/start/start.go
```diff
@@ -50,28 +51,35 @@ func StartCmd(cfg *config.Config) *cobra.Command {
...
 				if !cmd.Flags().Changed("example") {
-					if err := survey.Ask(questions, &projectPath); err != nil {
+					currentDir, err := filepath.Abs("")
+					if err != nil {
+						return err
+					}
+
+					displayPath := currentDir
+					defval := true
+					projectPath = currentDir
```
begelundmuller, line 62: nit: move `displayPath` and `defval` to above the `strings.Contains` check that first uses them

cli/cmd/start/start.go
```diff
@@ -50,28 +51,35 @@ func StartCmd(cfg *config.Config) *cobra.Command {
...
+
+					homeDir, err := os.UserHomeDir()
+					if err != nil {
 						return err
 					}
+
+					if strings.Contains(currentDir, homeDir) {
```
begelundmuller, line 69: Instead of `Contains`, use `HasPrefix`

begelundmuller (changes requested):

PR #2467, opened 2023-05-19
Redirect signups with unverified emails to UI error page

begelundmuller:
We can merge this when there's a UI page to match it :)


ericpgreen2 (approved):
Looks good. Here's a PR to add the new UI page: https://github.com/rilldata/rill-developer/pull/2473

PR #2471, opened 2023-05-19
Timeouts for gRPC and OLAP

egor-ryashin:


runtime/server/server.go
```diff
@@ -88,17 +89,29 @@ func (s *Server) Ping(ctx context.Context, req *runtimev1.PingRequest) (*runtime
 	return resp, nil
 }
 
+func TimeoutSelector(service, method string) time.Duration {
```
begelundmuller, line 92: Should not be exported (not to be used outside the package). Also please move it below `ServeGRPC`

runtime/server/server.go
```diff
@@ -246,3 +259,74 @@ func mapGRPCError(err error) error {
...
+func (w *serverStream) Context() context.Context {
+	return w.ctx
+}
+
+func wrapServerStream(ctx context.Context, ss grpc.ServerStream) *serverStream {
+	return &serverStream{
+		ServerStream: ss,
+		ctx:          ctx,
+	}
+}
```
begelundmuller, line 277: In other interceptors, we use the `grpc_middleware` package for this. You can replace it with just:
```
wrapped := grpc_middleware.WrapServerStream(ss)
wrapped.WrappedContext = newCtx

return handler(srv, wrapped)
```

runtime/server/server.go
```diff
@@ -246,3 +259,74 @@ func mapGRPCError(err error) error {
...
+			duration = fn(service, method)
+		}
+
+		ctx, cancel := context.WithTimeout(ss.Context(), duration)
+		defer cancel()
+		return handler(srv, wrapServerStream(ctx, ss))
+	}
+}
+
+func DeadlineUnaryServerInterceptor(fn func(service, method string) time.Duration) grpc.UnaryServerInterceptor {
```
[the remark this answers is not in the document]
begelundmuller: Same applies for the streaming variant

runtime/server/server.go
```diff
@@ -246,3 +259,74 @@ func mapGRPCError(err error) error {
...
+	) (interface{}, error) {
+		duration := time.Minute * 2
+		if fn != nil {
+			service, method, err := ParseFullMethod(info.FullMethod)
+			if err != nil {
+				return nil, err
+			}
+
+			duration = fn(service, method)
+		}
```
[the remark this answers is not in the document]
begelundmuller: Same applies for the streaming variant

begelundmuller (changes requested):

admin/server/server.go
```diff
@@ -20,6 +20,7 @@ import (
 	"github.com/rilldata/rill/admin/server/auth"
 	"github.com/rilldata/rill/admin/server/cookies"
 	adminv1 "github.com/rilldata/rill/proto/gen/rill/admin/v1"
+	"github.com/rilldata/rill/runtime/middleware"
```
begelundmuller, line 23: Move to `runtime/pkg/middleware` (`pkg` is usually used for generic utils)

admin/server/server.go
```diff
@@ -102,10 +103,15 @@ func New(logger *zap.Logger, adm *admin.Service, issuer *runtimeauth.Issuer, opt
 	}, nil
 }
 
+func timeoutSelector(service, method string) time.Duration {
+	return time.Minute
+}
```
begelundmuller, line 108: nit: put after public functions

runtime/drivers/druid/olap.go
```diff
@@ -43,6 +43,12 @@ func (c *connection) Execute(ctx context.Context, stmt *drivers.Statement) (*dri
 		return nil, prepared.Close()
 	}
 
+	var cancelFunc context.CancelFunc
+	if stmt.ExecutionTimeout != 0 {
+		ctx, cancelFunc = context.WithTimeout(ctx, stmt.ExecutionTimeout)
+		_ = cancelFunc
+	}
```
begelundmuller, line 50: This and setting it in the result is not needed. Firstly, the `ctx` is used only to resolve the query, not for reading the result, so there it doesn't make a difference to delay its cancellation until after the rows have been read. Secondly, the execution timeout is for execution, not for reading the results (that would be covered by the API-level timeout)

begelundmuller: So can just do 
```go
if stmt.ExecutionTimeout != 0 {
	cctx, cancelFunc := context.WithTimeout(ctx, stmt.ExecutionTimeout)
	defer cancelFunc()
	ctx = cctx
}
```

runtime/drivers/olap.go
```diff
@@ -30,17 +31,19 @@ type OLAPStore interface {
...
 }
 
 // Result wraps the results of query.
 type Result struct {
 	*sqlx.Rows
-	Schema    *runtimev1.StructType
-	cleanupFn func() error
+	Schema     *runtimev1.StructType
+	cleanupFn  func() error
+	CancelFunc context.CancelFunc
```
begelundmuller, line 46: See earlier note – this is redundant

begelundmuller (changes requested):

runtime/drivers/druid/olap.go
egor-ryashin: Nope, once the timeout is cancelled the rows cannot be read.

egor-ryashin: For example:
```
func Execute() {
        ...
	if stmt.ExecutionTimeout != 0 {
		ctx, cancel = context.WithTimeout(ctx, stmt.ExecutionTimeout)
		cancel()
	}

	rows, err := conn.QueryxContext(ctx, stmt.Query, stmt.Args...)
        ...
        return rows // the context is canceled at this point
}

func processing() {
        rows, err := Execute()
        ...
        if rows.next() {
          err = rows.Scan(&timeGrainString) // we get context-cancelled error here
        }
}


```

begelundmuller: Okay, I see. Slightly odd behavior from the library there, please add a comment before `context.WithTimeout` that the context is not released because it will also be used for reading rows.

Also, instead of the `CancelFunc` on `Result`, can't you just use the `SetCleanupFunc` that's already there? Instead of just passing `release` to it, you would pass a func that calls both `release` and `cancel`

runtime/drivers/duckdb/olap.go
```diff
@@ -130,7 +136,15 @@ func (c *connection) Execute(ctx context.Context, stmt *drivers.Statement) (res
...
 
 	res = &drivers.Result{Rows: rows, Schema: schema}
-	res.SetCleanupFunc(release) // Will call release when res.Close() is called.
+	res.SetCleanupFunc(func() error {
+		if cancelFunc != nil {
+			cancelFunc()
+		}
+		if release != nil {
+			return release()
+		}
```
begelundmuller, line 145: release is never `nil`

runtime/drivers/duckdb/olap.go
```diff
@@ -130,7 +136,15 @@ func (c *connection) Execute(ctx context.Context, stmt *drivers.Statement) (res
...
+		return nil
+	}) // Will call release when res.Close() is called.
```
begelundmuller, line 147: Can remove this comment now

runtime/drivers/duckdb/olap.go
```diff
@@ -116,6 +116,12 @@ func (c *connection) Execute(ctx context.Context, stmt *drivers.Statement) (res
 	// NOTE: We can't just "defer release()" because release() will block until rows.Close() is called.
 	// We must be careful to make sure release() is called on all code paths.
 
+	var cancelFunc context.CancelFunc
+	if stmt.ExecutionTimeout != 0 {
+		ctx, cancelFunc = context.WithTimeout(ctx, stmt.ExecutionTimeout)
+		_ = cancelFunc
```
begelundmuller, line 122: If not using `defer`, the `cancelFunc` must manually be called on all the error conditions

begelundmuller (changes requested):

begelundmuller (approved):

PR #2473, opened 2023-05-19
Add `/-/auth/verify-email` page

ericpgreen2:
Given #2467, this PR implements the `/-/auth/verify-email` page that Janet has mocked up in Figma.


himadrisingh:
Can we redirect to this page whenever we get the error `email not verified`

PR #2478, opened 2023-05-22
Add CSP & Cache Control headers

himadrisingh:
Please go the the `Preview` tab and select the appropriate sub-template:

* [Others](?expand=1&template=fallback_template.md)


begelundmuller (commented):
Why are the headers different for `/*` and `/-/*`? It's a SPA, and there are user-facing pages on both paths, so I would think everything should have the same headers?

netlify.toml
```diff
@@ -0,0 +1,19 @@
+# This should be used for docs also
+[[headers]]
+  for = "/*"
+  [headers.values]
+    Content-Security-Policy = "default-src https:; script-src https: 'unsafe-inline'; style-src https: 'unsafe-inline'; img-src https: data: blob:"
```
begelundmuller, line 5: - Shouldn't we use `script-src 'self'`? The application shouldn't be loading remote scripts
- Same for styles
- Images may come from auth providers, so we need wider permissions there

netlify.toml
```diff
@@ -0,0 +1,19 @@
...
+    Permissions-Policy = "geolocation=(),midi=(),sync-xhr=(self),microphone=(),camera=(),magnetometer=(),gyroscope=(),fullscreen=(self),payment=()"
+    Referrer-Policy = "no-referrer"
+    X-Frame-Options = "SAMEORIGIN"
+    X-Xss-Protection = "1; mode=block"
```
begelundmuller, line 9: Looks like this has been deprecated: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-XSS-Protection

netlify.toml
```diff
@@ -0,0 +1,19 @@
...
+    X-Content-Type-Options = "nosniff"
+
+[[headers]]
+  for = "/-/*"
+  [headers.values]
+    cache-control = '''
+    max-age=0,
+    no-cache,
+    no-store,
+    must-revalidate'''
```
begelundmuller, line 19: I don't think this is needed. The application is a SPA, so it serves static content on all routes. Any dynamic parameters in the query string will only be consumed client-side, not rendered in the content.

begelundmuller (changes requested):

netlify.toml
himadrisingh: script-src are inline and we have to calculate nonce of them, which can be problematic for developers. Sometimes we use open source libraries. But defailt-src is `https` which gets applied to all.

netlify.toml
himadrisingh: ok, we can remove this as its not working as expected.

himadrisingh, line 19:
```suggestion
```

netlify.toml
himadrisingh: This came from https://docs.netlify.com/routing/headers/#syntax-for-the-headers-file

himadrisingh: https://securityheaders.com/?q=ui.rilldata.io&followRedirects=on this doesn't require this so we can remove it.

begelundmuller (approved):

PR #2484, opened 2023-05-23
Move project init to APIs

### LINKED ISSUE #2453: Runtime: APIs for listing and unpacking examples
We want to add support for listing and unpacking example projects from within the UI. See details [here](https://www.notion.so/rilldata/Onboarding-screen-make-it-possible-to-jump-start-an-example-project-into-an-existing-empty-projec-6a8b813e85734eb3a6b7be2aacd19f8f).

`cli/pkg/examples` currently contains code that embeds example projects in the binary and allows unpacking them in `rill init` and `rill start`. 

Proposed task list – please suggest changes or improvements:

- [ ] Move the `cli/pkg/examples` logic into `runtime/pkg/examples` (adapt the `Makefile` accordingly)
- [ ] Adapt the `Makefile` to embed examples from https://github.com/rilldata/rill-examples instead of `examples` (delete that directory)
- [ ] Add a new runtime RPC `ListExamples` that returns example names (their directory name), titles and descriptions (from their `rill.yaml`) extracted from the embedded examples
- [ ] Add a new runtime RPC `UnpackExample` that unpacks an example project into an instance's `drivers.RepoStore` (should use the interface APIs, not the file system directly)
    - It should take a parameter `force=true`. It should first scan for conflicts and return an error instead of unpacking unless `force=true`.
- [ ] Adapt logic in `rill init` and `rill start` to unpack examples by calling functions on `runtime.Runtime`
- [ ] (Maybe: rill start should no longer initialize an empty project on start)

k-anshul:
This PR makes the following changes:

- Adds three new APIs:`ListExamples`, `UnpackExample`, and `UnpackEmpty`
- Pulls examples from https://github.com/rilldata/rill-examples
- No longer initializes an empty project in `rill start`

Expected frontend modifications:

- Do not rely on the project directory containing a `rill.yaml` or other Rill-related files
- Show the welcome screen if `rill.yaml` is not found
- Call `UnpackExample` to initialize a project (make sure to run `make cli.prepare` before starting the backend to setup the example projects)
- After unpacking an example, trigger reconcile (unpacking an example does *not* reconcile it)

Closes #2453.


nishantmonu51 (commented):
LGTM, did a high level pass over the APIs.

runtime/server/examples.go
```diff
@@ -0,0 +1,72 @@
...
+		paths, err := repo.ListRecursive(ctx, req.InstanceId, "{sources,models}/*.{yaml,yml,sql}")
+		if err != nil {
+			return nil, err
+		}
+
+		// Should we check content or filenames ??
+		if len(paths) != 0 {
+			return nil, fmt.Errorf("repo is not empty %s", paths)
+		}
+	}
```
begelundmuller, line 53: Instead of hardcoding paths to check, check that no paths in the example project exist in the directory

begelundmuller (changes requested):
In addition to the below suggestions, the scope of this work was expanded overnight. Additional changes needed:

- We should no longer initialize projects on `rill start`. So it needs to remove the initialization code around `app.InitProject` in `rill start`. It will also be easier to entirely remove the `rill init` command – that let's you delete this functions entirely (less old code that can cause issues).
  - If there is already a `rill.yaml`, it should continue to call reconcile on the project before starting the application.
- Since we don't immediately initialize projects anymore, we can no longer assume that an instance has a `rill.yaml` file. I audited the code, and I think there's only one backend change required to make this work:
  - In `runtime/registry.go:CreateInstance`, move the call to `ProjectConfig` and assignment of `ProjectVariables` to a new step during reconciliation instead – maybe just a new call from `runtime/services/catalog/migrations.go:Reconcile`.

proto/rill/runtime/v1/api.proto
```diff
@@ -84,6 +84,16 @@ service RuntimeService {
...
   }
 
+  // ListExamples lists all the examples embedded into binary
+  rpc ListExamples(ListExamplesRequest) returns (ListExamplesResponse) {
+    option (google.api.http) = {get: "/v1/examples"};
+  }
+  
+  // UnpackExample unpacks an example project
+  rpc UnpackExample(UnpackExampleRequest) returns (UnpackExampleResponse) {
+    option (google.api.http) = {get: "/v1/instances/{instance_id}/files/unpack-example"};
```
begelundmuller, line 94: Should be `post` instead of `get` (note: see other examples, it then also needs a `body` param)

runtime/pkg/examples/example.go
```diff
@@ -0,0 +1,57 @@
...
+			Name:        entry.Name(),
+			Title:       rillYaml.Title,
+			Description: rillYaml.Description,
+		})
+	}
+
+	return exampleList, nil
+}
+
+func Unpack(name string) (fs.FS, error) {
```
begelundmuller, line 55: `Get` would be more suitable – the unpacking happens at caller

begelundmuller (changes requested):

begelundmuller:
This PR has been approved, but should not be merged yet. It requires frontend modifications first. @ericpgreen2 Will merge when the frontend changes are in place.

cli/cmd/start/start.go
```diff
@@ -51,36 +49,26 @@ func StartCmd(cfg *config.Config) *cobra.Command {
...
+						},
+					},
+				}
 
-					msg := fmt.Sprintf("Rill will create project files in %q. Do you want to continue?", displayPath)
-					confirm := cmdutil.ConfirmPrompt(msg, "", defval)
-					if !confirm {
-						cmdutil.WarnPrinter("Aborted")
-						return nil
-					}
```
begelundmuller, line 81: Bad merge here

proto/rill/runtime/v1/api.proto
```diff
@@ -363,6 +384,40 @@ message RenameFileRequest {
...
+  bool force = 3;
+}
+
+// Response message for RuntimeService.UnpackExample
+message UnpackExampleResponse {}
+
+// Request message for RuntimeService.UnpackEmpty
+message UnpackEmptyRequest {
+  string instance_id = 1 [(validate.rules).string = {pattern: "^[_\\-a-zA-Z0-9]+$"}];
+  string title = 2 [(validate.rules).string.min_len = 1];
```
begelundmuller, line 415: Titles can be empty/unset

begelundmuller (changes requested):

begelundmuller (approved):
Backend looks good now

PR #2486, opened 2023-05-23
Add Go test coverage

begelundmuller:
It's currently just an ad-hoc output you can get by running `make coverage.go`. We can investigate connecting to a coverage monitoring service in a later PR.

It reports the current coverage at 50.5% of statements.


nishantmonu51 (approved):
👍

PR #2488, opened 2023-05-23
Trim space in file path for gcs credentials

nishantmonu51:
Trim extra space in file path, if any


runtime/connectors/gcs/gcs.go
```diff
@@ -75,11 +75,10 @@ var spec = connectors.Spec{
 					return ""
 				}
 
-				path, err := fileutil.ExpandHome(val)
+				path, err := fileutil.ExpandHome(strings.TrimSpace(val))
```
k-anshul, line 78: This is also required in `ValidateFunc`

k-anshul (approved):

PR #2503, opened 2023-05-25
Slot and region edit in project

### LINKED ISSUE #2452: Admin/CLI: Support changing a project's `region` or `slots`
- [ ] Add this support to `rill project edit` 
- [ ] Make `slots` an undocumented flag

k-anshul:
- closes #2452 
- fixes project and org already exist bug in rill deploy


admin/projects.go
```diff
@@ -124,6 +126,16 @@ func (s *Service) TeardownProject(ctx context.Context, p *database.Project) erro
 // UpdateProject updates a project and any impacted deployments.
 // It runs a reconcile if deployment parameters (like branch or variables) have been changed and reconcileDeployment is set.
 func (s *Service) UpdateProject(ctx context.Context, proj *database.Project, opts *database.UpdateProjectOptions, reconcileDeployment bool) (*database.Project, error) {
+	if proj.Region != opts.Region || proj.ProdSlots != opts.ProdSlots { // require new deployments
```
begelundmuller, line 129: This adds a redundancy where it will recreate the deployment and then afterwards update it in `impactsDeployments`. The call to `updateDeployment` may even failed because the new deployment is still reconciling (it crudely disregards concurrent reconciles because they're not supported under the hood).

The low down is that any change happening in `UpdateProject` must at most trigger one reconcile on a deployment.

begelundmuller (changes requested):

admin/projects.go
k-anshul: We are returning on creating new deployment here so its not triggering another reconcile. Although I think this leads to a bug where other filed aren't updated. I will fix it to read fields from opts.

admin/projects.go
```diff
@@ -124,6 +137,16 @@ func (s *Service) TeardownProject(ctx context.Context, p *database.Project) erro
...
+		s.logger.Info("update project: recreating project", zap.String("project_id", proj.ID),
+			zap.String("region", opts.Region),
+			zap.Int("slots", opts.ProdSlots),
+			observability.ZapCtx(ctx),
+		)
+		// create new deployment for project
+		return s.recreateDeployment(ctx, proj, opts)
```
begelundmuller, line 147: The control flow of `recreateDeployment` is a bit hard to reason about, since inside it also updates the project, though it sounds like it just triggers a redeploy.

There are no great options, but I think it would be better if this just updates the project first, and then calls `TriggerRedeploy`. And keep the logic in this function, so at least the control flow is easy to reason about

begelundmuller: Also, the log statement here may be useful, but no need to duplicate the params. We'll be logging request params in the request logs with the new logging changes anyway. And please also add a similar brief log message under `if impactsDeployments`.

The trace_ids ties all this together. They actually also encapsulate the runtime calls, it's pretty neat. Here's an example: https://rill.datadoghq.com/logs?query=%40grpc.method%3A%2Frill.admin.v1.AdminService%2FUpdateProjectVariables%20&cols=host%2Cservice&event=AgAAAYhKFAjY7tYnEQAAAAAAAAAYAAAAAEFZaEtGQTlDQUFCSTRBMmtPbVg1X0FBQgAAACQAAAAAMDE4ODRhMzAtMTg3MC00NmJlLWEyZDQtYWRiMTA2MTNiYzQw&index=%2A&messageDisplay=inline&stream_sort=desc&viz=stream&from_ts=1684419394629&to_ts=1685024194629&live=true

begelundmuller (changes requested):

admin/projects.go
k-anshul: > I think it would be better if this just updates the project first, and then calls TriggerRedeploy

This can leave the project in an inconsistent state in cases when `createDeployment` fails in the sense that project entry would be updated reflecting the new slots but the deployment with respect to project would still be deployed as per older number of slots.
Can be solved with db update in a `transaction` but it wouldn't be great to keep the transaction open till the new deployment is created. 
May be I can move the entire function to `updateProject` itself ?

k-anshul: Another option is to update rest of the properties first and update `slot` and `region` later and change it to take `slot` and `region` as parameters signifying project will be updated with these parameters plus new depl ID.

begelundmuller: No matter the order, there can be inconsistencies – if the new deployment is updated, but the Postgres call fails, there will also be inconsistency. For a redeploy, I think it's okay to update the Postgres first.

Without a background task system with retries, we just have to rely on the user's ability to retry themselves if things goes wrong (which they shouldn't too often here anyway).

k-anshul: The idea is not to solve for all inconsistencies but to ensure that if a deployment ID is updated and the deployment is active the slots in the deployment matches that of project. 
> if the new deployment is updated, but the Postgres call fails (or server crashes), there will also be inconsistency

The project will still not be updated with the new deployment ID and the new slots so if the user will run `rill project status` he will still see old slots. There will still be a phantom deployment (that remained bcz server crashed) but it should get deleted once user retries. 

> Without a background task system with retries, we just have to rely on the user's ability to retry themselves if things goes wrong (which they shouldn't too often here anyway).

Agreed but we must make sure that the errored state is correctly stored in the system and presented to the user. 
Imagine the case where we update project with new slots but the new deployment can't be created due to issues like slots not available (likely). The user is presented with an error `deployment failed` so the user runs `rill project status` to see the current state of the project and sees project status as `DEPLOYMENT_STATUS_OK ` with the new slots updated. This is highly incorrect state to present to the user and likely they may not even retry in this case.

Having said that I agree there are still cases when we may end up in the scenario that I described with the current impl but those will be very rare (cases like db update was successful but server received failure and the deployment delete was also unsuccessful) and can only be solved by making sure that we create deployment and update project within same transaction.

begelundmuller: You're right, that makes sense – in that case, can we just inline `recreateDeployment` in `UpdateProject` so the control flow is very clear (even if a bit long)? And only teardown the current prod deployment, not all deployments for the project (to prevent weird states like trying to teardown something that is already tearing down, or if we do other features where non-prod deployments may hang around)

begelundmuller (approved):
Perfect

PR #2510, opened 2023-05-25
Leaderboard row count percentage in tooltip 2

bcolloran:
This prototype PR updates the leaderboard tooltip in two ways:
1) removes the row label from the top of the tooltip, since this is redundant information (the row label is always shown)
2) replaces the row label with the number of rows and the percentage of filtered rows that each leaderboard row represents.

If no filters are active, then the row count and percentage are out of all available data:
image

If a filter is active, then we only use the filtered row count, and the tooltip mentions "filtered rows"
image


bcolloran:
@hamilton this PR is ready for review. FYI, the tricky bug I mentioned in slack that @skokenes tests caught seemed to be caused by what I will call, for lack of a better term "slot drilling" -- passing target component down slots through multiple layers of other components, which was preventing the target component from getting rerendering/getting updated props even though I could see in the console that the parent of this target component was getting (and should have been passing on) new props. Flattening out `DimensionLeaderboardEntrySet.svelte` fixed this problem, though it does means I touched more files than I might have desired.

(I recall that about a month ago I encountered a similar problem of a slotted component not updating correctly, so it's something to watch out for)

@jkhwu per your request, tagging you here so you can see @hamilton's proposed approach re the tooltip.

---

@hamilton I also want to point out one behavior detail and get your take -- in the image below, when I hover on the row for "Google", the leaderboard entry shows a value (which in this case happen to be the same as the row count) even though the row is actually not included in the filter. This also means that the row count inline query data is available, so we are able to show the number of rows for Google in the tooltip. However, since Google is not actually included in the 32k filtered rows, it doesn't really make sense to say that google is "18k of 32k filtered rows", since its really **0** of 32k filtered rows
image

I think I'd recommend that if we want to indicate the proportion of filtered rows, we should simply say "excluded from current filter set" or something along those lines, but this could be confusing in other ways. In general, the semantics of items that are not included in the filter set but are shown in a leaderboard could bear some more detailed thought wrt to old denominator problem, since they are shown to the user but not included in any summary number

skokenes:
> the tricky bug I mentioned in slack that @skokenes tests caught seemed to be caused by what I will call, for lack of a better term "slot drilling" -- passing target component down slots through multiple layers of other components, which was preventing the target component from getting rerendering/getting updated props

Interesting, I saw a similar bug in our UI a couple of weeks ago that was not completely solved by, but was helped by, flattening some "slot drilling" happening. It makes me question where there is some underlying bug with slots in Svelte that we keep hitting

bcolloran:
@skokenes I talked to @hamilton this morning and he suggested that I tag you for review on this one, though we will hold off on merging for a moment while some product and design discussions are happening

web-common/src/features/dashboards/dashboard-stores.ts
skokenes, line 1: This may or may not be a better pattern than we use today to get `useDashboardStore(metricName)`, but are we sure we want to introduce this new pattern here in just 1 place, without broader architecture scoped out?
If we merge this, now we have code accessing dashboard stores in multiple ways.
It'd be best if we could stick with 1 pattern

skokenes: Same comment goes for introducing selectors in this way. To be clear, this is not me advocating against the pattern itself; more me wondering about introducing new patterns like this in a feature PR before we've solidified our overall strategy and plan to unify our approach across the codebase

skokenes (changes requested):
I left a couple of comments, but generally given that I've never touched any of these components, I don't feel like I have enough information to review any of this code.

I'd suggest:
1) Schedule a meeting to walk me through what we've done here and why
2) Let's also tag someone who is familiar with these components to date to take a look

Thanks

bcolloran:
closing as obsolete -- we put this info in the context column in the leaderboards

PR #2519, opened 2023-05-29
Source telemetry

djbarnwal:
## Todo

- [x] Add Source Trigger Telemetry
- [x] Create store for current active page
- [ ] Fix previous navigation event bugs  #2542
- [ ] Handle and parse large file source errors


web-common/src/features/sources/add-source/FileDrop.svelte
```diff
@@ -53,14 +64,31 @@
...
-        await createSource(
+        const errors = await createSource(
           queryClient,
           runtimeInstanceId,
           tableName,
           yaml,
           $createSourceMutation
         );
+
+        if (createSourceMutationError.isError || errors.length) {
```
AdityaHegde, line 75: We should get the error for the source file path instead of assuming that the 1st error is for the source.

web-common/src/features/sources/add-source/LocalSource.svelte
```diff
@@ -104,6 +114,24 @@
         // if the upload didn't work, delete the source file.
         handleDeleteSource(tableName);
       }
+
+      if ($createSourceMutation.isError || errors.length) {
+        sourceErrorTelemetryHandler(
+          MetricsEventSpace.Modal,
+          $appScreen,
+          createSourceMutationError?.message ?? errors[0]?.message,
```
AdityaHegde, line 122: Get the error based on file path.

web-common/src/features/sources/add-source/RemoteSource.svelte
```diff
@@ -123,6 +142,23 @@
               false
             );
           }
+          if ($createSourceMutation.isError || error) {
```
AdityaHegde, line 145: Get this error based on file path, including updating the error shown to the user.

web-common/src/features/sources/sourceUtils.ts
```diff
@@ -45,3 +53,56 @@ export function inferSourceName(connector: V1Connector, path: string) {
...
+  if (!fileType) return "";
+
+  if (fileType === "gz") {
+    return fileName.split(".").slice(-2).shift();
+  }
+
+  return fileType;
+}
+
+export function sourceErrorTelemetryHandler(
```
AdityaHegde, line 69: Would something like `emitSourceErrorTelemetry` make more sense here? Same for the next method.

web-common/src/features/sources/sourceUtils.spec.ts
```diff
@@ -83,3 +89,12 @@ describe("inferSourceName", () => {
     });
   });
 });
+
+describe("getFileTypeFromPath", () => {
```
AdityaHegde, line 93: Great to see tests for this!

AdityaHegde (approved):

PR #2523, opened 2023-05-30
remove credentials configuration from deploy command

k-anshul:
CLI: Remove credentials configuration from rill deploy – print message instead

### DIFF
--- a/cli/cmd/deploy/deploy.go
+++ b/cli/cmd/deploy/deploy.go
@@ -14,7 +14,6 @@ import (
 	adminclient "github.com/rilldata/rill/admin/client"
 	"github.com/rilldata/rill/admin/pkg/urlutil"
 	"github.com/rilldata/rill/cli/cmd/auth"
-	"github.com/rilldata/rill/cli/cmd/env"
 	"github.com/rilldata/rill/cli/cmd/org"
 	"github.com/rilldata/rill/cli/pkg/browser"
 	"github.com/rilldata/rill/cli/pkg/cmdutil"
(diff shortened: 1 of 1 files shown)


cli/cmd/deploy/deploy.go
```diff
@@ -282,6 +275,10 @@ func DeployCmd(cfg *config.Config) *cobra.Command {
 
 			// Success!
 			success.Printf("Created project \"%s/%s\". Use `rill project rename` to change name if required.\n\n", cfg.Org, res.Project.Name)
+			// Run flow to check connector credentials
+			if err := variablesFlow(ctx, fullProjectPath, name); err != nil {
+				return err
+			}
 			success.Printf("Rill projects deploy continuously when you push changes to Github.\n")
```
begelundmuller, line 282: 1. Move the log message up before the variables flow
2. If the variables flow fails, probably just log the error instead of returning/exiting with it

cli/cmd/deploy/deploy.go
```diff
@@ -478,6 +475,39 @@ func createProjectFlow(ctx context.Context, client *adminclient.Client, req *adm
...
+		if _, ok := src.Properties["path"]; ok {
+			// print URL wherever applicable
+			fmt.Printf(" - %s\n", src.Properties["path"])
+		} else {
+			fmt.Printf(" - %s\n", src.Name)
+		}
+	}
+	fmt.Printf("\nRun `rill env configure --project %s` to provide access to the data store.\n", projectName)
+	warn := color.New(color.Bold).Add(color.FgYellow)
+	warn.Printf("The project will remain in an error state until this is done.\n\n")
```
begelundmuller, line 506: 1. Maybe print it all in warn color
2. Merge last log into the first – "Could not ingest all sources. Rill requires credentials for the following sources:"
3. Then make the last log: "Run `rill env configure --project %s` to provide credentials."

begelundmuller (changes requested):

begelundmuller (approved):

PR #2524, opened 2023-05-30
Scheduled background job for deleting expired device auth codes

begelundmuller:


admin/worker/worker.go
```diff
@@ -33,6 +33,9 @@ func New(logger *zap.Logger, adm *admin.Service) *Worker {
 func (w *Worker) Run(ctx context.Context) error {
 	group, ctx := errgroup.WithContext(ctx)
 	group.Go(func() error { return w.schedule(ctx, "check_slots", w.checkSlots, 15*time.Minute) })
+	group.Go(func() error {
+		return w.schedule(ctx, "delete_expired_device_auth_codes", w.deleteExpiredDeviceAuthCodes, 15*time.Minute)
```
pjain1, line 37: Not sure if we should run it this frequently as in the normal flow device code will be cleaned up after confirmation code approval or rejection. Only in cases where flow is abandoned in the middle code will be not be deleted. Probably running once in few hours should be sufficient ?

pjain1 (commented):
There is a race condition where device code expires and it is cleaned up by the worker before the UI confirms/rejects the user code or cli asks for the access token. In these cases user will get error `code not found` instead of `expired token`. Its a rare case not affecting correctness so mostly not worth handling.

begelundmuller:
Thank you, very valid points. I have addressed both comments in the most recent commit.

pjain1 (approved):

PR #2526, opened 2023-05-30
Connector preflight APIs

### LINKED ISSUE #2341: Runtime: Connector preflight APIs

k-anshul:
closes #2341


runtime/connectors/s3/s3.go
```diff
@@ -190,10 +195,121 @@ func (c connector) HasAnonymousAccess(ctx context.Context, env *connectors.Env,
...
+	}
+
+	bucket, err := openBucket(ctx, &Config{AWSRegion: req.Region}, req.Bucket, creds)
+	if err != nil {
+		return nil, "", err
+	}
+	defer bucket.Close()
+
+	var pageToken []byte
+	if req.GetPageToken() == "" {
```
begelundmuller, line 256: Let's just access the properties directly instead of through getters, they're not commonly used in Go (except for in interfaces)

runtime/connectors/s3/s3.go
```diff
@@ -190,10 +195,121 @@ func (c connector) HasAnonymousAccess(ctx context.Context, env *connectors.Env,
...
+	}
+
+	sess, err := getAwsSessionConfig(ctx, &Config{}, req.GetBucket(), creds)
+	if err != nil {
+		return "", err
+	}
+
+	if sess.Config.Region != nil {
+		return *sess.Config.Region, nil
+	}
```
begelundmuller, line 309: Same question as earlier – is it necessary to set region?

runtime/connectors/gcs/gcs.go
```diff
@@ -191,6 +200,100 @@ func (c connector) HasAnonymousAccess(ctx context.Context, env *connectors.Env,
...
+
+	projectID := credentials.ProjectID
+	if projectID == "" {
+		f := &credentialsFile{}
+		if err := json.Unmarshal(credentials.JSON, f); err != nil {
+			return nil, "", err
+		}
+
+		projectID = f.getProjectID()
+	}
```
begelundmuller, line 223: Is `credentials.JSON` guaranteed to be set in this case?

begelundmuller (changes requested):
Apart from the comments below, I would suggest moving the handlers to `gcs/api.go` and `s3/api.go` respectively to provide some light context about how the functions are used

runtime/connectors/s3/s3.go
```diff
@@ -190,10 +195,121 @@ func (c connector) HasAnonymousAccess(ctx context.Context, env *connectors.Env,
...
+		},
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	// no region found, default to us-east-1
+	if sess.Config.Region == nil || *sess.Config.Region == "" {
+		sess = sess.Copy(&aws.Config{Region: aws.String("us-east-1")})
+	}
```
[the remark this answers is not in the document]
k-anshul: 1. The region here is not used for the API. It is used to specify where we want to send our requests to (https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/configuring-sdk.html)
We can in fact send it to any location and the API should ideally succeed so that's why I just fallback to `us-east-1` if nothing is found in user configurations.
2. Good catch. This is a bug in the current resolution of region for a bucket as well. I will fix at both places.

runtime/connectors/s3/s3.go
```diff
@@ -190,10 +195,121 @@ func (c connector) HasAnonymousAccess(ctx context.Context, env *connectors.Env,
...
+	for _, bucket := range output.Buckets {
+		if bucket.Name != nil {
+			buckets = append(buckets, *bucket.Name)
+		}
+	}
+	return buckets, nil
+}
+
+func (c Connector) ListObjects(ctx context.Context, req *runtimev1.S3ListObjectsRequest, env *connectors.Env) ([]*runtimev1.S3Object, string, error) {
+	// todo :: check for cases when accessing public buckets but env configured
```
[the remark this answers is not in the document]
k-anshul: Yeah its a open question. So currently when user is trying to access public buckets but have their env configured we have a fallback to try again with anonymous credentials. Do we want same behaviour here as well ? I am not sure how the e2e flow in UI will look like. The user won't see a public buckets in `ListBucket` API so not sure if this case is possible for connector preflight experience.
If their is an alternate flow then I guess once user inputs any bucket name we can also ask user to toggle the switch if the bucket is public.

runtime/connectors/s3/s3.go
k-anshul: Replied in the other comment.

runtime/connectors/gcs/gcs.go
k-anshul: Yes. I took a look at the credentials resolution. There seem to be four cases where credentials can be retrieved from the host system. 
1. Via env variable : Json will be set
2. via well known file (application_default_credentials.json) : Json will be set
3. We're on a Google App Engine : project ID will be set, json will not be
4. We're on Google Compute Engine : project ID will be set, json will not be

runtime/connectors/s3/s3.go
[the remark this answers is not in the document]
begelundmuller: I think we should try fallback to anon credentials here as well. If user manually enters a public bucket name, the UI will probably want to "just work" and not require the user to specify if it's public or not.

runtime/connectors/gcs/gcs.go
begelundmuller: Sounds good. Just to be defensive for future changes, let's add a check for `credentials.JSON == nil` also, and return a nicer response for that case

PR #2532, opened 2023-05-31
Data exports

egor-ryashin:
How to use:

- Call `createQueryServiceExport` to get a download URL, passing parameters for either `MetricsViewToplist` or `MetricsViewRows`
- Concatenate the returned `downloadUrlPath` to the runtime host to get the download URL


cli/pkg/local/app.go
```diff
@@ -272,6 +273,9 @@ func (a *App) Serve(httpPort, grpcPort int, enableUI, openBrowser, readonly bool
 			mux.Handle("/local/config", a.infoHandler(inf))
 			mux.Handle("/local/version", a.versionHandler())
 			mux.Handle("/local/track", a.trackingHandler(inf))
+			mux.Handle("/v1/downloads", &downloads.DownloadHandler{
+				Runtime: a.Runtime,
+			})
```
begelundmuller, line 278: This download API won't just be for local, but also in cloud. So mount it around `runtime/server/server.go:165` instead

cli/cmd/runtime/start.go
```diff
@@ -130,11 +132,17 @@ func StartCmd(cliCfg *config.Config) *cobra.Command {
...
+					Runtime: rt,
+				})
+			}
+
 			// Run server
 			ctx := graceful.WithCancelOnTerminate(context.Background())
 			group, cctx := errgroup.WithContext(ctx)
 			group.Go(func() error { return s.ServeGRPC(cctx) })
-			group.Go(func() error { return s.ServeHTTP(cctx, nil) })
+			group.Go(func() error { return s.ServeHTTP(cctx, fn) })
```
begelundmuller, line 145: Change here not needed given other comment

runtime/queries/downloads/downloads.go
```diff
@@ -0,0 +1,110 @@
...
+		}
+	default:
+		http.Error(w, fmt.Sprintf("Unsupported request type: %s", reflect.TypeOf(v).Name()), http.StatusBadRequest)
+		return
+	}
+
+	w.WriteHeader(http.StatusOK)
+}
+
+func (h *DownloadHandler) executeToplist(ctx context.Context, writer http.ResponseWriter, req *runtimev1.MetricsViewToplistRequest) error {
```
begelundmuller, line 73: Since the export logic will be custom to each query, how about we implement it on the `Query` type instead? E.g. we could add a new func to the `Query` interface:

```go
	Export(ctx context.Context, rt *Runtime, instanceID string, priority int, format runtimev1.DownloadFormat, w io.Writer) error
```

Internally, it could decide whether to call `.Resolve` on itself, or whether to run an entirely different implementation (e.g. using DuckDB to build the CSV using the `COPY` statement).

Most queries will not support exports. We should just have a const error `runtime.ErrExportNotSupported` that they can return.

runtime/queries/downloads/downloads.go
```diff
@@ -0,0 +1,110 @@
...
+	Format           string = "format"
+	Request          string = "request"
+	downloadPriority int    = 0
+)
+
+type DownloadHandler struct {
+	Runtime *runtime.Runtime
+}
+
+func (h *DownloadHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
```
begelundmuller, line 30: Let's implement this in `runtime/server` instead of a separate package. There's already a file `runtime/server/export.go` that has an old export implementation (just dumps a whole table), we could put it there

proto/rill/runtime/v1/queries.proto
```diff
@@ -605,3 +612,28 @@ message TableRowsRequest {
...
+enum DownloadFormat {
+  DOWNLOAD_FORMAT_UNSPECIFIED = 0; // assumes CSV
+  DOWNLOAD_FORMAT_CSV = 1;
+  DOWNLOAD_FORMAT_PARQUET = 2;
+}
+
+enum DownloadCompression {
+  DOWNLOAD_COMPRESSION_UNSPECIFIED = 0; // assumes no compression
+  DOWNLOAD_COMPRESSION_ZIP = 1;
+}
```
begelundmuller, line 635: To keep our backend simple, I would suggest not having compression be separate from format. For Parquet, compression is supposed to be applied at columnar level (i.e. inside the format, not at file level), so there we should just have an opinionated choice.

For CSV, we could have `DOWNLOAD_FORMAT_CSV` and `DOWNLOAD_FORMAT_CSV_GZIP` if necessary, though I think initially, if possible, we can just support CSV and instead apply gzip compression on the HTTP response (`Content-Encoding: gzip`) to minimize bandwidth

begelundmuller (changes requested):

runtime/queries/downloads/downloads.go
egor-ryashin: I think using a package for each handler is a more decoupled solution. I always assumed `server` is for core functionality, an additional handler acts as an extension to the `server`.

runtime/queries/downloads/downloads.go
egor-ryashin: Makes sense, though the long parameter list says we don't structure properly somewhere.

egor-ryashin: Should we move `rt *runtime.Runtime, instanceID string, priority int` into `Context`?

begelundmuller: I agree the `Query` (and now `Export`) funcs are doing too much now. I don't think hiding the params in the context is the right way to clean it up, though. I have a few ideas for how we could simplify it, but let's keep it like this for now, and consider larger refactors afterwards.

runtime/queries/downloads/downloads.go
begelundmuller: Almost all the APIs are currently implemented in `server`, I'm not convinced this needs to be an exception – it'll be less than 100 lines of code when the formatting logic is moved to the query

cli/cmd/runtime/start.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/rilldata/rill/runtime"
 	"github.com/rilldata/rill/runtime/pkg/graceful"
 	"github.com/rilldata/rill/runtime/pkg/observability"
+	"github.com/rilldata/rill/runtime/queries/downloads"
```
begelundmuller, line 15: Import appears unused?

proto/rill/runtime/v1/queries.proto
```diff
@@ -605,3 +612,22 @@ message TableRowsRequest {
...
+  int32 limit = 1;
+  DownloadFormat format = 2;
+  oneof request {
+    MetricsViewToplistRequest metrics_view_toplist_request = 3;
+    MetricsViewRowsRequest metrics_view_rows_request = 4;
+  }
+}
+
+enum DownloadFormat {
+  DOWNLOAD_FORMAT_UNSPECIFIED = 0; // assumes CSV
```
begelundmuller, line 626: Let's not assume CSV, and instead return an error if it's unspecified. It's not a user-facing API, so better to require the client to be explicit

proto/rill/runtime/v1/queries.proto
```diff
@@ -605,3 +612,22 @@ message TableRowsRequest {
 message TableRowsResponse {
   repeated google.protobuf.Struct data = 1;
 }
+
+message DownloadLinkRequest {
+  int32 limit = 1;
```
begelundmuller, line 617: Put validation rule here that it's less than or equal to 10k rows

runtime/queries/metricsview_rows.go
```diff
@@ -91,6 +94,45 @@ func (q *MetricsViewRows) Resolve(ctx context.Context, rt *runtime.Runtime, inst
...
+		return err
+	}
+
+	switch format {
+	case runtimev1.DownloadFormat_DOWNLOAD_FORMAT_UNSPECIFIED, runtimev1.DownloadFormat_DOWNLOAD_FORMAT_CSV:
+		return writeCSV(q.Result.Meta, q.Result.Data, writer)
+	}
+
+	return nil
+}
```
begelundmuller, line 109: Should return unsupported error for formats that are not supported

runtime/queries/metricsview_rows.go
```diff
@@ -91,6 +94,45 @@ func (q *MetricsViewRows) Resolve(ctx context.Context, rt *runtime.Runtime, inst
...
+	}
+	if err := w.Write(record); err != nil {
+		return err
+	}
+	record = record[:0]
+
+	for _, structs := range data {
+		for _, field := range structs.Fields {
+			record = append(record, field.GetStringValue())
+		}
```
begelundmuller, line 126: `GetStringValue()` will return nil if the field doesn't have a string type. I think `AsInterface()` or `String()` are what you need

begelundmuller: Also, are you certain that slicing to 0 and then calling `append` will re-use the same array? I know slicing keeps the underlying array, but couldn't immediately find info about whether `append` includes safety measures to prevent overriding underlying data that can be referenced by other slices.

I think it should work as expected, just want to be sure

runtime/queries/metricsview_rows.go
```diff
@@ -91,6 +94,45 @@ func (q *MetricsViewRows) Resolve(ctx context.Context, rt *runtime.Runtime, inst
...
+
+	switch format {
+	case runtimev1.DownloadFormat_DOWNLOAD_FORMAT_UNSPECIFIED, runtimev1.DownloadFormat_DOWNLOAD_FORMAT_CSV:
+		return writeCSV(q.Result.Meta, q.Result.Data, writer)
+	}
+
+	return nil
+}
+
+func writeCSV(meta []*runtimev1.MetricsViewColumn, data []*structpb.Struct, writer io.Writer) error {
```
begelundmuller, line 111: Since it's used by multiple functions, let's move it to the `metricsview.go` file, which has other metrics view related utils

runtime/server/downloads.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/rilldata/rill/runtime"
+	"github.com/rilldata/rill/runtime/queries"
+	"google.golang.org/protobuf/proto"
+)
+
+type DownloadHandler struct {
+	Runtime *runtime.Runtime
+}
+
+func (h *DownloadHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
```
begelundmuller, line 20: Can just implement it on the server like the other handlers, like `func (s *Server) downloadHandler(w http.ResponseWriter, r *http.Request)`, and use `http.HandlerFunc` to mount it on the `mux`

runtime/server/queries_downloads.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"context"
+	"encoding/base64"
+	"fmt"
+
+	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
+	"google.golang.org/protobuf/proto"
+)
+
+func (s *Server) Download(ctx context.Context, req *runtimev1.DownloadLinkRequest) (*runtimev1.DownloadLinkResponse, error) {
```
begelundmuller, line 12: Let's move this func to the `downloads.go` file so we keep the download logic together in one file

begelundmuller (changes requested):

runtime/server/server.go
```diff
@@ -177,6 +177,8 @@ func (s *Server) HTTPHandler(ctx context.Context, registerAdditionalHandlers fun
 	// Add httpMux on gRPC-gateway
 	httpMux.Handle("/v1/", gwMux)
 
+	httpMux.HandleFunc("/v1/downloads", s.downloadHandler)
```
begelundmuller, line 180: Maybe just `/v1/download`?

begelundmuller (changes requested):
1. Can you also check the performance when downloading 10k rows on a wide dataset (like `spending.parquet` in GCS)?
2. Can you also check that the model rows download returns sensible values for other types, like timestamp, structs, JSON fields, etc.?

PR #2533, opened 2023-05-31
Using name instead of column for dimensions in APIs

AdityaHegde:
## Checklist
- [x] Manual verification
- [x] Unit test coverage
- [ ] E2E test coverage
- [ ] Needs manual QA?

#### Details:
Since we are adding a unique identifiable field `name` to dimensions we need to move all APIs to use this field. If a request comes with missing dimension in catalog we throw an error.

## Steps to Verify
N/A Will be in the working branch


runtime/queries/metricsview.go
```diff
@@ -258,3 +258,52 @@ func repeatString(val string, n int) []string {
 	}
 	return res
 }
+
+func validateAndGetDimension(mv *runtimev1.MetricsView, dimName string) (string, error) {
```
begelundmuller, line 262: Call it `metricsViewDimensionToColumn` or something like that

runtime/queries/metricsview.go
```diff
@@ -258,3 +258,52 @@ func repeatString(val string, n int) []string {
...
+			}
+			// backwards compatibility for older projects that have not run reconcile on this dashboard
+			// in that case `column` will not be present
+			return dimension.Name, nil
+		}
+	}
+	return "", fmt.Errorf("dimension %s not found", dimName)
+}
+
+func convertFilterToColumn(mv *runtimev1.MetricsView, filter *runtimev1.MetricsViewFilter) error {
```
begelundmuller, line 277: Isn't this the only place that we lookup dimension columns in a filter? https://github.com/rilldata/rill-developer/blob/2efa20aa0de5694190029c68aeadf8a19be2a197/runtime/queries/metricsview.go#LL175C18-L175C18

I think it should just call `validateAndGetDimension` right there instead of doing this in-place rewriting

runtime/queries/metricsview.go
begelundmuller: Also, apply `safeName` to the result so the callers don't need a separate statement for it

begelundmuller: And the places it's called, I guess the output vars should be `colName` or something instead of `dimName`

begelundmuller (changes requested):

runtime/queries/metricsview.go
AdityaHegde: Renaming to `dimensionToSafeColumn`. Since already in metricsview package dont think it makes sense to add it to the name.

AdityaHegde: My bad. The package is not `metricsview`. Will rename appropriately.

PR #2543, opened 2023-06-01
support limited whitelisting of domains by org admins

pjain1:


admin/server/public_email_providers_list
```diff
@@ -0,0 +1,156 @@
+# List of public email providers, these email domains cannot be whitelisted
```
pjain1, line 1: Not sure if something needs to be done to bundle it in the binary.

admin/database/database.go
```diff
@@ -59,7 +59,7 @@ type DB interface {
 	UpdateOrganizationAllUsergroup(ctx context.Context, orgID, groupID string) (*Organization, error)
 
 	FindOrganizationAutoinviteDomain(ctx context.Context, orgID string, domain string) (*OrganizationAutoinviteDomain, error)
-	FindOrganizationAutoinviteDomainsForOrganization(ctx context.Context, orgID string) ([]*OrganizationAutoinviteDomain, error)
+	FindOrganizationAutoinviteDomainsForOrganization(ctx context.Context, orgID string) ([]*OrganizationWhitelistedDomain, error)
```
begelundmuller, line 62: Feels redundant to have `OrganizationAutoinviteDomain` and `OrganizationWhitelistedDomain`. Let's keep it simple and resolve role IDs in the API handler.

Also, let's change `autoinvite` to `whitelisted` everywhere for consistency

admin/server/organizations.go
```diff
@@ -527,6 +525,24 @@ func (s *Server) CreateAutoinviteDomain(ctx context.Context, req *adminv1.Create
...
 
+	if !claims.Superuser(ctx) {
+		if !claims.OrganizationPermissions(ctx, org.ID).ManageOrg {
+			return nil, status.Error(codes.PermissionDenied, "only org admins can add whitelisted domain")
+		}
+		// check if the user's domain matches the autoinvite domain
+		user, err := s.admin.DB.FindUser(ctx, claims.OwnerID())
+		if err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
```
begelundmuller, line 536: Must add check that is's a user after `auth.GetClaims` to ensure `OwnerID` is a user ID

admin/server/organizations.go
```diff
@@ -527,6 +525,24 @@ func (s *Server) CreateAutoinviteDomain(ctx context.Context, req *adminv1.Create
...
+		if !strings.HasSuffix(user.Email, req.Domain) {
+			return nil, status.Error(codes.PermissionDenied, "Domain name doesn’t match verified email domain. Please contact Rill support.")
+		}
```
begelundmuller, line 539: Probably need to parse email and check. This way, I could sign up as `benjamin@notfoo.com` and whitelist `foo.com`

admin/server/server.go
```diff
@@ -324,3 +333,34 @@ func newURLRegistry(opts *Options) *externalURLs {
...
+	}
+	defer f.Close()
+
+	scanner := bufio.NewScanner(f)
+	for scanner.Scan() {
+		domain := scanner.Text()
+		if strings.HasPrefix(domain, "#") {
+			continue
+		}
+		providers[scanner.Text()] = true
```
begelundmuller, line 358: Trim whitespace before/after to be safe

admin/server/server.go
```diff
@@ -324,3 +333,34 @@ func newURLRegistry(opts *Options) *externalURLs {
...
 	}
 }
+
+const publicEmailProvidersListFile = "public_email_providers_list"
+
+//go:embed public_email_providers_list
+var ef embed.FS
+
+// read public email providers list file line by line and load it into a map
+func loadPublicEmailProvidersList() (map[string]bool, error) {
```
begelundmuller, line 343: 1. Move the public email logic to `admin/pkg/publicemail`
2. Have a `IsPublic(email)` package-level func that parses the email does the check
3. Use globals and a `sync.Once` to lazily parse the providers into the map
4. Instead of `var ef embed.FS`, the `go:embed` directory supports loading a specific file directly into a string. So you can have something like `var domainsList string`

admin/user.go
```diff
@@ -20,6 +21,21 @@ func (s *Service) CreateOrUpdateUser(ctx context.Context, email, name, photoURL
...
+			return nil, err
+		}
+		for _, org := range orgs {
+			existingOrgMemberships[org.ID] = true
+		}
+
+		err = s.handleAutoinvites(ctx, user, existingOrgMemberships)
+		if err != nil {
+			return nil, err
+		}
```
begelundmuller, line 37: We shouldn't wait for user to log in again before adding existing users. Can just keep the logic here the same. Instead, when setting the whitelist, it should scan for and add any matching existing users to the org in the API handler.

cli/cmd/user/whitelist/list.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"github.com/rilldata/rill/cli/pkg/cmdutil"
+	"github.com/rilldata/rill/cli/pkg/config"
+	adminv1 "github.com/rilldata/rill/proto/gen/rill/admin/v1"
+	"github.com/spf13/cobra"
+)
+
+func ListCmd(cfg *config.Config) *cobra.Command {
+	removeCmd := &cobra.Command{
+		Use:   "list [--org <org>]",
```
begelundmuller, line 14: Don't put flags in the usage string, Cobra lists them automatically

cli/cmd/user/whitelist/remove.go
```diff
@@ -0,0 +1,50 @@
...
+
+	"github.com/rilldata/rill/cli/pkg/cmdutil"
+	"github.com/rilldata/rill/cli/pkg/config"
+	adminv1 "github.com/rilldata/rill/proto/gen/rill/admin/v1"
+	"github.com/spf13/cobra"
+)
+
+func RemoveCmd(cfg *config.Config) *cobra.Command {
+	removeCmd := &cobra.Command{
+		Use:   "remove <email-domain> [--org <org>]",
```
begelundmuller, line 14: Remove flag

cli/cmd/user/whitelist/remove.go
```diff
@@ -0,0 +1,50 @@
...
+			defer client.Close()
+
+			domain := args[0]
+
+			cmdutil.WarnPrinter(fmt.Sprintf("New users with email addresses ending in %q will no longer automatically be added to %q. "+
+				"Existing users previously added through this policy will keep their access. (To remove users, use `rill user remove`.)", domain, cfg.Org))
+			if !cmdutil.ConfirmPrompt("Do you want to confirm?", "", false) {
+				cmdutil.WarnPrinter("Aborted")
+				return nil
+			}
```
begelundmuller, line 33: 1. Don't need a confirm here
2. Print message after the API call

cli/cmd/user/whitelist/setup.go
```diff
@@ -0,0 +1,55 @@
...
+			if err != nil {
+				return err
+			}
+			defer client.Close()
+
+			domain := args[0]
+
+			cmdutil.WarnPrinter(fmt.Sprintf("If you confirm, new and existing users with email addresses ending in %q will automatically be added to %q with role %q."+
+				"\n\nTo whitelist another email domain than your own, reach out to support: https://rilldata.com/support", domain, cfg.Org, role))
+			if !cmdutil.ConfirmPrompt("Do you want to confirm?", "", false) {
```
begelundmuller, line 31: Just `Do you confirm?`

cli/cmd/user/whitelist/list.go
begelundmuller: Also missing (see `setup` command):

```
	listCmd.Flags().StringVar(&cfg.Org, "org", cfg.Org, "Organization")
```

cli/cmd/user/whitelist/whitelist.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"github.com/rilldata/rill/cli/pkg/cmdutil"
+	"github.com/rilldata/rill/cli/pkg/config"
+	"github.com/spf13/cobra"
+)
+
+func WhitelistCmd(cfg *config.Config) *cobra.Command {
+	whitelistCmd := &cobra.Command{
+		Use:               "whitelist",
+		Short:             "Whitelist domain",
```
begelundmuller, line 12: Make it `Whitelist access by email domain`

begelundmuller (changes requested):

admin/database/database.go
pjain1: We are doing this at other places as well returning `Member` from `FindOrganizationMemberUsers` instead of `User` and `Invite` from `FindOrganizationInvites` instead of `OrganizationInvite` etc to prevent extra lookups and get all the information in one go using joins. Just pointing it out, will change it if you still feel so.

Sure will change `autoinvite` to `whitelisted`

admin/server/organizations.go
pjain1: changing to `strings.HasSuffix(user.Email, "@" + req.Domain)`

admin/server/server.go
pjain1: 1. Okay
2. Okay
3. I thought it would be better to load file contents into memory at startup rather than at some point during runtime to have predictable memory usage.
4. Wouldn't it cause unnecessary copying into a string first and then to a map ?

admin/user.go
pjain1: This sounds expensive if there are a lot of users.

admin/database/database.go
begelundmuller: Yeah, I'm not entirely sure how we should approach joined types, but do feel our current `Invite` and `Member` types are also a bit misplaced (though at least `Member` doesn't clash with any other types, since we don't have `OrganizationMember` for example). 

In this particular case, the naming just ends up feeling too confusing (`OrganizationAutoinviteDomain` and `OrganizationWhitelistedDomain` sound like the exact same thing). Probably we should pick one of these options:

- Just resolve roles separately in the API. We could have a multi-finders like `FindOrganizationRolesByIDs(ctx context.Context, ids ...string) (map[string]*OrganizationRole, error)` to make it easier.
- Have clearly named expanded return types. So `OrganizationWhitelistedDomain` might mirror the actual table, then we could have `OrganizationWhitelistedDomainWithJoinedNames` to show the distinction. This would also have to be reflected in the `Find...` func names.

admin/server/server.go
begelundmuller: I wouldn't worry too much about parsing performance here since it's a very minor overhead. Rather, it would just be nice to keep the logic constrained to one package that doesn't have to be separately initialized and passed around as variables (since it's stateless anyway). 

You can use `embed.FS` instead of a direct `string` if you want to reduce that extra bit of memory overhead (the current file looks like ~5kb though?).

admin/user.go
begelundmuller: It's the product requirement – I guess the idea is that if you add a whitelisted domain, you expect member listings to reflect everyone with access right away. 

I don't think the performance is that bad, though. This is an API that will get hit very rarely, even at scale (will probably never be called more than a few times per hour), and we're not going to have *that* many users on a single domain anytime soon.

If performance ever becomes a problem, we could move it to a background job, but I think it's safe to do in the API handler for now.

admin/pkg/publicemail/publicemail.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+	defer f.Close()
+
+	scanner := bufio.NewScanner(f)
+	for scanner.Scan() {
+		domain := scanner.Text()
+		if strings.HasPrefix(domain, "#") {
+			continue
+		}
+		providers[strings.TrimSpace(scanner.Text())] = true
```
begelundmuller, line 36: Let's put to lower here

admin/pkg/publicemail/publicemail.go
```diff
@@ -0,0 +1,58 @@
...
+func IsPublic(domain string) bool {
+	once.Do(func() {
+		var err error
+		providers, err = loadPublicEmailProvidersList()
+		if err != nil {
+			panic(err)
+		}
+	})
+
+	_, ok := providers[domain]
```
begelundmuller, line 56: Let's use to lower for the domain also

admin/server/organizations.go
```diff
@@ -599,19 +625,51 @@ func (s *Server) CreateAutoinviteDomain(ctx context.Context, req *adminv1.Create
...
 
-	return &adminv1.CreateAutoinviteDomainResponse{}, nil
+	// add existing users belonging to the whitelisted domain to the org
+	users, err := s.admin.DB.FindUsers(ctx)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	for _, user := range users {
+		if strings.HasSuffix(user.Email, "@"+req.Domain) {
```
begelundmuller, line 635: Use `FindUsersByEmailPattern` with pattern `%@domain` instead

admin/server/organizations.go
```diff
@@ -589,7 +609,13 @@ func (s *Server) CreateAutoinviteDomain(ctx context.Context, req *adminv1.Create
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
-	_, err = s.admin.DB.InsertOrganizationAutoinviteDomain(ctx, &database.InsertOrganizationAutoinviteDomainOptions{
+	ctx, tx, err := s.admin.DB.NewTx(ctx)
+	if err != nil {
+		return nil, err
+	}
+	defer func() { _ = tx.Rollback() }()
```
begelundmuller, line 616: Find and check the users to add before starting the transaction, to keep the transaction open for less time

begelundmuller (changes requested):

begelundmuller (approved):

PR #2547, opened 2023-06-02
Logging imp

### LINKED ISSUE #2454: Runtime: Logging improvements burndown
For context and details, see [this doc](https://www.notion.so/rilldata/Logging-for-Rill-runtime-70e5cb99fe374f77bf0f6ab6300d6344).

Proposed task list:

- [ ] Include the instance ID in loggers used in reconcile and source ingestion (like passing `logger.With(zap.String("instance", instanceID))`)
- [ ] Reconcile: Log the input config at the beginning of reconcile
- [ ] Reconcile: Log the result payload at the end of reconcile
- [x] Ingestion: Log start/end log for each batch during incremental ingestion (include relevant parameters, like file names)
- [x] Ingestion: Log start/end log for each ingestion call (include relevant stats/params, like bytes ingested)
- [ ] Server: Create a util function for adding request parameters to the request trace AND request log message
  - It would be nice if the parameters were added to the trace, and the logging middleware could extract trace attributes and add them to request finish log message (we may need to use OpenTelemetry `Baggage` instead of trace attributes to facilitate this, not sure)
  - For example, the util function for adding request parameters could look something like this (quick draft, requires more research):
    ```go
    observability.SetRequestAttributes(ctx,
        otel.Attribute("dimension", req.DimensionName),
        otel.Attribute("num_filters", len(filters)),
        // ...
    )
    ```
- [ ] Server: Use the request parameter logging util to log non-sensitive and non-large payloads for all runtime RPCs, including the analytical APIs (for example, instead of logging filter values, it could log dimension names filtered on)
- [ ] Server: In the analytical APIs, add a request parameter indicating whether the query had a cache hit or not
- [x] Local CLI: When running `rill start`, configure a custom logger that:
  - Logs to stdout only on `error` level (or higher) OR when the message contains `zap.Named("console")`
  - Emits all `info` level (or higher) logs to `~/.rill/logs`, rotating/deleting old logs when the file size reaches 100MB
- [x] Add `zap.Named("console")` to the "Ingesting source" and similar logs

k-anshul:
closes #2454


admin/server/users.go
```diff
@@ -34,6 +36,10 @@ func (s *Server) ListSuperusers(ctx context.Context, req *adminv1.ListSuperusers
 }
 
 func (s *Server) SetSuperuser(ctx context.Context, req *adminv1.SetSuperuserRequest) (*adminv1.SetSuperuserResponse, error) {
+	observability.SetRequestAttributes(ctx,
+		attribute.Bool("super_user", req.Superuser),
```
begelundmuller, line 40: Just `superuser` (one word)

runtime/query.go
```diff
@@ -83,8 +83,10 @@ func (r *Runtime) Query(ctx context.Context, instanceID string, query Query, pri
 
 	// Try to get from cache
 	if val, ok := r.queryCache.cache.Get(key); ok {
+		observability.SetRequestAttributes(ctx, attribute.Bool("cache_hit", true))
 		return query.UnmarshalResult(val)
 	}
+	observability.SetRequestAttributes(ctx, attribute.Bool("cache_hit", false))
```
begelundmuller, line 89: Use key `query.cache_hit`

runtime/server/queries_timeseries.go
```diff
@@ -27,6 +36,18 @@ func (s *Server) ColumnRollupInterval(ctx context.Context, req *runtimev1.Column
 }
 
 func (s *Server) ColumnTimeSeries(ctx context.Context, req *runtimev1.ColumnTimeSeriesRequest) (*runtimev1.ColumnTimeSeriesResponse, error) {
+	observability.SetRequestAttributes(ctx,
+		attribute.String("instance_id", req.InstanceId),
+		attribute.String("table", req.TableName),
+		attribute.StringSlice("measures", marshalProtoSlice(req.Measures)),
+		attribute.String("timestamp_column", req.TimestampColumnName),
+		attribute.String("time_range", marshalProto(req.TimeRange)),
+		attribute.Int("Filters", filterCount(req.Filters)),
```
begelundmuller, line 45: lowercase `filters`

runtime/server/observability_utils.go
```diff
@@ -0,0 +1,38 @@
...
+	return res
+}
+
+func marshalProto(pb proto.Message) string {
+	b, err := protojson.Marshal(pb)
+	if err != nil {
+		return ""
+	}
+	return string(b)
+}
```
begelundmuller, line 31: It doesn't feel so clean to have encoded json fields as log fields. I'm wondering if we should either:
- Flatten messages to multiple fields with dot-separated keys, like `"parent.child.key"`
- Or just have a custom func for each pb type (there are not many). E.g. `InlineMeasures` could just return a `[]string` for `StringSlice` of the names (the other info is less important).

begelundmuller (changes requested):
Two other suggestions:
- Should we change `SetRequestAttributes` to `AddRequestAttributes`?
- Should we namespace all the arg keys, like `"args.instance_d"` to make navigating logs/traces easier?

runtime/server/observability_utils.go
k-anshul: I am not sure how we can flatten messages in a generic way. I can add a combination of both in the sense that when it is not possible to just return `[]string`. I can add multiple fields with dot-separated keys.

begelundmuller (approved):

PR #2552, opened 2023-06-02
Handle different error message fields for Auth

djbarnwal:
Auth0 is not consistent with it's error fields. 

For example adding a weak password on SignUp gives the below error as `description` is an object
Screenshot 2023-06-02 at 10 23 19 PM

This PR tries for a best effort solution looking for a string error message.

### DIFF
--- a/web-auth/src/components/Auth.svelte
+++ b/web-auth/src/components/Auth.svelte
@@ -114,11 +114,11 @@
             // Auth0 is not consistent in the naming of the error description field
             const errorText =
               typeof err?.description === "string"
-                ? err?.description
+                ? err.description
                 : typeof err?.policy === "string"
-                ? err?.policy
+                ? err.policy
                 : typeof err?.error_description === "string"
-                ? err?.error_description
+                ? err.error_description
                 : err?.message;
 
             if (err) displayError({ message: errorText });
--- a/web-auth/src/components/Auth.svelte
+++ b/web-auth/src/components/Auth.svelte
@@ -109,8 +109,9 @@
             email: email,
             password: password,
           },
-          (err) => {
-            // Auth0 is not consistent with its error message fields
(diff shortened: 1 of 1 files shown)


web-auth/src/components/Auth.svelte
```diff
@@ -109,8 +109,19 @@
...
             password: password,
           },
-          (err) => {
-            if (err) displayError({ message: err?.description });
+          // explicitly typing as any to avoid missing property TS/svelte-check error
+          (err: any) => {
+            // Auth0 is not consistent in the naming of the error description field
+            const errorText =
+              typeof err?.description === "string"
+                ? err?.description
```
ericpgreen2, line 117: Can't we remove the `?`s after we've checked for the existence of each property

ericpgreen2: e.g. `err?.description` --> `err.description`

ericpgreen2 (approved):
Looks reasonable – approving.

However, I'm seeing that the `displayError(err: any)` function could be made simpler. It takes an untyped `err` object, yet expects the object to have a `message` property. It looks like it could be `displayError(message: string)`. Feel free to make this edit or not.

djbarnwal:
> Looks reasonable – approving.
> 
> However, I'm seeing that the `displayError(err: any)` function could be made simpler. It takes an untyped `err` object, yet expects the object to have a `message` property. It looks like it could be `displayError(message: string)`. Feel free to make this edit or not.

Have made this affordance so that in future we can add more customization to how we display the error messages. For ex - via a toast or showing a `errorTitle` and `errorDescription`.

ericpgreen2:
> Have made this affordance so that in future we can add more customization to how we display the error messages. For ex - via a toast or showing a `errorTitle` and `errorDescription`.

Gotcha. In that case, at least typing the `err` object would be helpful.

PR #2538, opened 2023-05-31
Improve "Add {source, model, dashboard}" buttons

ericpgreen2:
## Checklist
- [x] Manual verification
- [ ] Unit test coverage
- [x] E2E test coverage
- [x] Needs manual QA?

#### Issue addressed: 
[PRD in Notion](https://www.notion.so/rilldata/Improving-Target-size-for-Add-Source-Models-Dashboard-Nav-c04af12473a642e8b87143b826a12038)

#### Details:
This PR improves the "Add {source, model, dashboard}" buttons. It removes the icon buttons in favor of large horizontal buttons that explicitly say "Add source", "Add model", and "Add dashboard", respectively. 

This PR does not implement the "Suggested Actions" component of the PRD. That will follow in a separate PR.

## Steps to Verify
Verify that the new buttons match the Figma mocks.

PR #2555, opened 2023-06-03
Tighten up Onboarding screen

ericpgreen2:
## Checklist
- [x] Manual verification
- [ ] Unit test coverage
- [ ] E2E test coverage
- [x] Needs manual QA?

#### Issue addressed: 
The onboarding screen (at the root page `/`) has confused some people who mistook each list item for a clickable button. Additionally, the copy is outdated now that we've changed the Add Source button (#2538).

#### Details:
Janet did a light design pass over the screen. [Here's the Figma design](https://www.figma.com/file/ADt3OivNV54xeoXkUpPxP2/RILL?type=design&node-id=1685%3A35815&t=2eEg72rm9Bo4Z88J-1). This PR implements this design.

This PR does _not_ create a new `secondary` button style. Hence the button does not match the mock exactly. Changing the `secondary` button can be done in a separate PR that considers the `secondary` buttons across the application.

## Steps to Verify
1. Open the onboarding screen
2. Check that it matches the Figma mock
3. Check that the Add Data button works correctly


djbarnwal (changes requested):
I think we can be more flexible with the width of the onboarding container. Even for a large viewports, the lines are getting wrapped. I would ideally expect the container to utilize the available space a bit more to avoid such cases. 

image

djbarnwal (approved):
Looks good

PR #2557, opened 2023-06-05
single fix permissions fix

k-anshul:
Adds support for non-glob ingestion without list permission in S3/GCS

### DIFF
--- a/runtime/connectors/blob/blobdownloader.go
+++ b/runtime/connectors/blob/blobdownloader.go
@@ -252,7 +252,7 @@ func (it *blobIterator) plan() ([]*objectWithPlan, error) {
 		attr, err := it.bucket.Attributes(it.ctx, it.opts.GlobPattern)
 		if err != nil {
 			// can fail due to permission not available
-			it.logger.Error("failed to fetch attributes of the object", zap.Error(err))
+			it.logger.Info("failed to fetch attributes of the object", zap.Error(err))
 		} else {
 			size = attr.Size
 		}
--- a/runtime/connectors/blob/blobdownloader.go
+++ b/runtime/connectors/blob/blobdownloader.go
@@ -176,16 +176,22 @@ func (it *blobIterator) NextBatch(n int) ([]string, error) {
 			if err != nil {
 				return err
 			}
-			defer file.Close()
 
 			it.localFiles[index-start] = file.Name()
 			ext := filepath.Ext(obj.obj.Key)
 			partialReader, isPartialDownloadSupported := _partialDownloadReaders[ext]
 			downloadFull := obj.full || !isPartialDownloadSupported
 
-			// Collect metrics of download size and time
 			startTime := time.Now()
+			// Collect metrics of download size and time
 			defer func() {
+				size := obj.obj.Size
+				st, err := file.Stat()
+				if err == nil {
+					size = st.Size()
+				}
+				file.Close()
+
 				duration := time.Since(startTime)
 				it.logger.Info("download complete", zap.String("object", obj.obj.Key), zap.Duration("duration", duration), observability.ZapCtx(it.ctx))
 				connectors.RecordDownloadMetrics(grpCtx, &connectors.DownloadMetrics{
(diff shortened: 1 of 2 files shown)


runtime/connectors/blob/blobdownloader.go
```diff
@@ -176,24 +176,30 @@ func (it *blobIterator) NextBatch(n int) ([]string, error) {
 			if err != nil {
 				return err
 			}
-			defer file.Close()
```
begelundmuller, line 179: `defer` calls are stacked (LIFO), so you shouldn't need to remove this

runtime/connectors/blob/blobdownloader.go
```diff
@@ -176,24 +176,30 @@ func (it *blobIterator) NextBatch(n int) ([]string, error) {
...
-			// Collect metrics of download size and time
 			startTime := time.Now()
+			// Collect metrics of download size and time
 			defer func() {
+				size := obj.obj.Size
+				st, err := file.Stat()
+				if err == nil {
+					size = st.Size()
+				}
+				file.Close()
```
begelundmuller, line 193: See other comment

runtime/connectors/blob/blobdownloader.go
```diff
@@ -239,8 +245,24 @@ func (it *blobIterator) plan() ([]*objectWithPlan, error) {
 		return nil, err
 	}
 
-	listOpts := listOptions(it.opts.GlobPattern)
+	listOpts, singleFile := listOptions(it.opts.GlobPattern)
```
begelundmuller, line 248: This would be more intuitive with a regular go-like `val, ok :=` pattern, like `listOpts, ok := `

runtime/connectors/blob/blobdownloader.go
```diff
@@ -239,8 +245,24 @@ func (it *blobIterator) plan() ([]*objectWithPlan, error) {
...
 	it.logger.Info("planner started", zap.String("glob", it.opts.GlobPattern), zap.String("prefix", listOpts.Prefix), observability.ZapCtx(it.ctx))
```
begelundmuller, line 249: Move log message to after the single file case. There will not be a corresponding completed message for single files. For single file, probably have dedicated start/completed messages

runtime/connectors/blob/blobdownloader.go
```diff
@@ -289,11 +311,12 @@ func listOptions(globPattern string) *blob.ListOptions {
 	if !fileutil.IsGlob(glob) {
 		// single file
 		listOptions.Prefix = globPattern
+		return listOptions, true
```
begelundmuller, line 314: `return nil, false`

runtime/connectors/blob/blobdownloader.go
```diff
@@ -289,11 +311,12 @@ func listOptions(globPattern string) *blob.ListOptions {
...
 	} else if prefix != "." {
 		listOptions.Prefix = prefix
 	}
 
-	return listOptions
+	return listOptions, false
```
begelundmuller, line 319: `return listOptions, true`

runtime/connectors/blob/blobdownloader.go
```diff
@@ -176,24 +176,30 @@ func (it *blobIterator) NextBatch(n int) ([]string, error) {
...
 
 			it.localFiles[index-start] = file.Name()
 			ext := filepath.Ext(obj.obj.Key)
 			partialReader, isPartialDownloadSupported := _partialDownloadReaders[ext]
 			downloadFull := obj.full || !isPartialDownloadSupported
 
-			// Collect metrics of download size and time
 			startTime := time.Now()
+			// Collect metrics of download size and time
 			defer func() {
```
begelundmuller, line 187: Why move the comment, the `startTime` is related to metrics gathering? It's nice to keep comments that relate to a block of code instead of a single line at the start of the block

begelundmuller (changes requested):

runtime/connectors/blob/blobdownloader.go
k-anshul: Makes sense. I was actually looking for a better way to do this and added this at last resort.

begelundmuller (approved):

PR #2572, opened 2023-06-06
Storybook eval 3

bcolloran:
Hey front end team, (@hamilton , @skokenes , @ericpgreen2 , @AdityaHegde , @djbarnwal ) -- In concert with some of the architecture discussions that @hamilton , @skokenes and I have been having, we've also been talking about introducing storybook as a way of accelerating and adding robustness to our front end development.

I don't know if any of you all have used it before, but I'm excited for us to experiment with this. I think it's a great tool -- it allows you to easily develop and see components in isolation, and can be incredibly useful in catching visual errors and edge cases that are much harder to detect when you are only looking at components in the context of a entire application, because you can set up a bunch of different sets of props/state (i.e. "stories") for a component and then basically do visual unit testing / fuzzing / regression QA on a component across a lot of cases without having to do a million clicks to get the app into different states -- you can just set up a story for every possible set of props a component might have, and add to that library of stories when you encounter errors and edge cases, just like you do with unit tests.

Anyway, this pull request represents a minimum viable implementation of storybook for the `web-common` folder of Rill. My advice is that we try storybook within just that slice of the app, and in particular that we pilot it alongside the architecture work that Speros has proposed (being able to mock a single data+state provider will allow us to set up stories for pretty much everything in web-common pretty easily). After we have tried working with storybook in web-common for a while, if people are finding it useful, we can then roll it out to other slices of the app.

To run storrybook, `npm install` to get it's dev deps, then `cd` into `web-common` and `npm run storybook`. For the time being, I've just put in a couple simple stories just to demo the overall concept.

No real action-items at this time, but I encourage you all to try setting up stories for components you are working on. If this tool is useful, then at some point down the road we may wish to bikeshed a bit about how we want to organize things -- for now, I think we can put all Storybook stories in the `web-common/src/stories` folder (and subfolders therein), but down the line we may want to put story folders+files closer to the component files they render, like we do with tests.


skokenes:
Looks great!

hamilton (approved):
From my perspective, I think we should be leveraging storybook very highly. The project has come pretty far since I last used it; I am finding this to be a really lightweight force-multiplier.

In the past, with sveltekit projects, I've preferred to just maintain a `/dev` route to prototype quickly, especially with data. There are some limitations of storybook here, but I think we can probably surmount those in other ways, e.g. not necessarily connecting components to data sources dynamically. We can always generate the data and statically load it into storybook.

Let's get this merged & begin utilizing it heavily.

PR #2574, opened 2023-06-07
Log creation of users, orgs, projects

begelundmuller:


himadrisingh (approved):

admin/user.go
```diff
@@ -130,6 +131,8 @@ func (s *Service) CreateOrUpdateUser(ctx context.Context, email, name, photoURL
 		return nil, err
 	}
 
+	s.logger.Info("created user", zap.String("user_id", user.ID))
```
himadrisingh, line 134: can we add org info here.

PR #2578, opened 2023-06-07
Prepare for rill repo name change

himadrisingh:


cli/pkg/telemetry/telemetry.go
```diff
@@ -18,7 +18,7 @@ const (
 	intakeURL  = "https://intake.rilldata.io/events/data-modeler-metrics"
 	intakeUser = "data-modeler"
 	intakeKey  = "lkh8T90ozWJP/KxWnQ81PexRzpdghPdzuB0ly2/86TeUU8q/bKiVug=="
-	appName    = "rill-developer"
```
begelundmuller, line 21: Maybe the analytics depends on this being fixed, so we should probably keep it the same

web-common/src/metrics/service/MetricsService.ts
```diff
@@ -52,7 +52,7 @@ export class MetricsService
   public async loadCommonFields() {
     const projectPathParts = this.localConfig.project_path.split("/");
     this.commonFields = {
-      app_name: "rill-developer",
```
begelundmuller, line 55: Same note, this might not be safe to change. And shouldn't matter keeping it at the old value.

cli/pkg/telemetry/telemetry.go
```diff
@@ -18,7 +18,7 @@ const (
...
+	appName    = "rill"
```
himadrisingh, line 21:
```suggestion
	appName    = "rill-developer"
```

web-common/src/metrics/service/MetricsService.ts
```diff
@@ -52,7 +52,7 @@ export class MetricsService
...
+      app_name: "rill",
```
himadrisingh, line 55:
```suggestion
      app_name: "rill-developer",
```

PR #2582, opened 2023-06-08
migrating to patch apis

### LINKED ISSUE #2548: Admin/Runtime: Refactor update APIs to use FieldMask
All our update APIs currently have PUT-like semantics. This has backwards compatibility problems and opens up easy-to-miss bugs, as https://github.com/rilldata/rill-developer/pull/2544 shows.

Instead, we should have PATCH-like update APIs. Some resources:

- https://cloud.google.com/apis/design/standard_methods#update
- https://pkg.go.dev/google.golang.org/protobuf/types/known/fieldmaskpb
- https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/field_mask.proto
- https://grpc-ecosystem.github.io/grpc-gateway/docs/mapping/patch_feature/
- https://netflixtechblog.com/practical-api-design-at-netflix-part-1-using-protobuf-fieldmask-35cfdc606518

Task list:

- [ ] Refactor proto API definitions
- [ ] Refactor API calls from CLI and UI
- [ ] Consider whether to refactor the DB update functions

k-anshul:
closes #2548


admin/server/projects.go
```diff
@@ -280,30 +284,29 @@ func (s *Server) UpdateProject(ctx context.Context, req *adminv1.UpdateProjectRe
...
+		Description:          valOrDefault(req.Description, proj.Description),
+		Public:               valOrDefault(req.Public, proj.Public),
+		ProdBranch:           valOrDefault(req.ProdBranch, proj.ProdBranch),
 		ProdVariables:        proj.ProdVariables,
 		GithubURL:            githubURL,
 		GithubInstallationID: proj.GithubInstallationID,
 		ProdDeploymentID:     proj.ProdDeploymentID,
-		ProdSlots:            int(req.ProdSlots),
-		Region:               req.Region,
+		ProdSlots:            int(valOrDefault(req.ProdSlots, int64(proj.ProdSlots))),
```
begelundmuller, line 308: Can reduce one level of nesting if putting `int(req.ProdSlots)` instead of `int64(proj.ProdSlots)`

admin/server/projects.go
```diff
@@ -261,13 +261,17 @@ func (s *Server) UpdateProject(ctx context.Context, req *adminv1.UpdateProjectRe
...
-		attribute.String("args.description", req.Description),
-		attribute.Bool("args.public", req.Public),
-		attribute.String("args.region", req.Region),
-		attribute.Int64("args.prod_slots", req.ProdSlots),
-		attribute.String("args.prod_branch", req.ProdBranch),
-		attribute.String("args.github_url", req.GithubUrl),
+		attribute.String("args.description", safeStr(req.Description)),
+		attribute.String("args.region", safeStr(req.Region)),
+		attribute.String("args.prod_branch", safeStr(req.ProdBranch)),
+		attribute.String("args.github_url", safeStr(req.GithubUrl)),
```
begelundmuller, line 267: If the fields are not changed, this will make the logs look like they're being set to `""`, which is confusing

begelundmuller (changes requested):
I think we should also use this occasion to fix two other issues:

1. The `UpdateOrganization` and `UpdateProject` APIs have some weird semantics where they take `id` as an identifier, and `name` is for renaming. It's confusing. Can we remove the `id` field, always use `name` as the identifier, and add a new optional field `string new_name` for renaming instead?

2. The CLI still behaviors PUT-like by passing all args by using defaults on the flags. However, the flag defaults code has a bug. For example, this code:
```go
			cmd.Flag("description").DefValue = proj.Description
			cmd.Flag("prod-branch").DefValue = proj.ProdBranch
			cmd.Flag("public").DefValue = fmt.Sprintf("%v", proj.Public)
			if cfg.Interactive {
				err = cmdutil.SetFlagsByInputPrompts(*cmd, "description", "prod-branch", "public")
				if err != nil {
					return err
				}
			}
```
Actually will *not* apply the defaults if `cfg.Interactive` is false, leading them to be set as nulls. There's a bit more context here: https://rilldata.slack.com/archives/CTZ8XBQ85/p1686168994742829

So I think it would be safer if we build the request objects incrementally, using `if cmd.Flags().Changed("...")` to conditionally set a value. It would also make the API logs more informative.

admin/server/projects.go
k-anshul: I get this error : cannot convert req.ProdSlots (variable of type *int64) to type int.

admin/server/projects.go
```diff
@@ -714,3 +717,10 @@ func safeStr(s *string) string {
 	}
 	return *s
 }
+
+func valOrDefault[T any](ptr *T, def T) T {
```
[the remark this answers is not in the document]
k-anshul: Strongly feel either `valOrDefault ` or `getOrDefault` is more suitable here.

admin/server/projects.go
k-anshul: I assumed that since empty string is not possible it would be self-explanatory that fields weren't set but yeah agreed on the confusion. I will only update when field is present like boolean fields.

admin/server/organizations.go
```diff
@@ -162,12 +162,16 @@ func (s *Server) DeleteOrganization(ctx context.Context, req *adminv1.DeleteOrga
 
 func (s *Server) UpdateOrganization(ctx context.Context, req *adminv1.UpdateOrganizationRequest) (*adminv1.UpdateOrganizationResponse, error) {
 	observability.AddRequestAttributes(ctx,
-		attribute.String("args.id", req.Id),
 		attribute.String("args.org", req.Name),
-		attribute.String("args.description", req.Description),
 	)
```
begelundmuller, line 166: Just make it one line

admin/server/projects.go
```diff
@@ -280,32 +295,31 @@ func (s *Server) UpdateProject(ctx context.Context, req *adminv1.UpdateProjectRe
...
-		}
-	}
-
 	githubURL := proj.GithubURL
-	if req.GithubUrl != "" {
-		githubURL = &req.GithubUrl
+	if req.GithubUrl != nil {
+		// If changing the Github URL, check github app is installed and caller has access on the repo
+		if safeStr(proj.GithubURL) != *req.GithubUrl {
+			_, err = s.getAndCheckGithubInstallationID(ctx, *req.GithubUrl, claims.OwnerID())
```
begelundmuller, line 302: Actually, this func doesn't have a check that `claims.OwnerID()` is a user ID. We should add that

begelundmuller (changes requested):

cli/cmd/org/edit.go
```diff
@@ -52,22 +52,18 @@ func EditCmd(cfg *config.Config) *cobra.Command {
...
+			req := &adminv1.UpdateOrganizationRequest{
+				Name: org.Name,
+			}
 			if cfg.Interactive {
-				err = cmdutil.SetFlagsByInputPrompts(*cmd, "description")
+				description, err = cmdutil.InputPrompt("Enter the description", org.Description)
 				if err != nil {
 					return err
 				}
+				req.Description = &description
```
begelundmuller, line 63: If `description` is passed via flag and `!cfg.Interactive`, description will not be updated

begelundmuller (changes requested):

begelundmuller (approved):

admin/server/server.go
```diff
@@ -34,7 +34,11 @@ import (
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
-var minCliVersion = version.Must(version.NewVersion("0.28.0"))
+var _minCliVersion = version.Must(version.NewVersion("0.20.0"))
+var methodWiseMinCliVersion = map[string]*version.Version{
+	"UpdateProject":      version.Must(version.NewVersion("0.28.0")),
+	"UpdateOrganization": version.Must(version.NewVersion("0.28.0")),
+}
```
begelundmuller, line 41: 1. Let's just use fully qualified API names (like `/rill.admin.v1.AdminService/UpdateProject`) here to avoid potential clashes in future
2. nit: call the var `_minCliVersionByMethod` to have consistent prefix

admin/server/server.go
```diff
@@ -290,13 +309,22 @@ func checkUserAgent(ctx context.Context) (context.Context, error) {
 		return nil, status.Error(codes.PermissionDenied, fmt.Sprintf("could not parse rill-cli version: %s", err.Error()))
 	}
 
-	if v.LessThan(minCliVersion) {
-		return nil, status.Error(codes.PermissionDenied, fmt.Sprintf("Rill %s is no longer supported, please upgrade to the latest version", v))
+	if v.LessThan(minVersion) {
+		return nil, status.Error(codes.PermissionDenied, fmt.Sprintf("Rill %s is no longer supported for given operation, please upgrade to the latest version", v))
 	}
```
begelundmuller, line 314: I would propose extracting the method from the `ctx` here and doing the method checks here. It keeps the version checking logic contained in one func with a clear call path (the implicit way `AuthFuncOverride` gets called is pretty weird, I had to check the docs to understand why there wasn't a reference to it)

begelundmuller (changes requested):

begelundmuller (approved):

PR #2584, opened 2023-06-08
druid timeboundary query for time range

pjain1:
This enables Druid SQL planner use optimized timeboundary query for time range. Its a limitation with Druid currently that the query will be converted to timeBoundary only if only one of the agg either min or max is set, see [here](https://github.com/apache/druid/blob/master/sql/src/main/java/org/apache/druid/sql/calcite/rel/DruidQuery.java#L972).

### DIFF
--- a/runtime/queries/column_time_range.go
+++ b/runtime/queries/column_time_range.go
@@ -127,50 +127,72 @@ func handleDuckDBInterval(interval any) (*runtimev1.TimeRangeSummary_Interval, e
 }
 
 func (q *ColumnTimeRange) resolveDruid(ctx context.Context, olap drivers.OLAPStore, priority int) error {
-	rangeSQL := fmt.Sprintf(
-		"SELECT min(%[1]s) as \"min\", max(%[1]s) as \"max\", timestampdiff(SECOND, min(%[1]s), max(%[1]s)) as \"interval\" FROM %[2]s",
+	minSQL := fmt.Sprintf(
+		"SELECT min(%[1]s) as \"min\" FROM %[2]s",
+		safeName(q.ColumnName),
+		safeName(q.TableName),
+	)
+	maxSQL := fmt.Sprintf(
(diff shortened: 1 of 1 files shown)


runtime/queries/column_time_range.go
```diff
@@ -127,50 +127,72 @@ func handleDuckDBInterval(interval any) (*runtimev1.TimeRangeSummary_Interval, e
...
+		err = minRows.MapScan(minRowMap)
 		if err != nil {
 			return err
 		}
-		if v := rowMap["min"]; v != nil {
+		maxRowMap := make(map[string]any)
+		err = maxRows.MapScan(maxRowMap)
+		if err != nil {
+			return err
+		}
```
begelundmuller, line 173: Use regular `Scan` instead of `MapScan` to scan directly into a `time.Time`. Gets rid of map and time casting.

runtime/queries/column_time_range.go
```diff
@@ -127,50 +127,72 @@ func handleDuckDBInterval(interval any) (*runtimev1.TimeRangeSummary_Interval, e
...
-		err = rows.MapScan(rowMap)
+	maxRows, err := olap.Execute(ctx, &drivers.Statement{
+		Query:            maxSQL,
+		Priority:         priority,
+		ExecutionTimeout: defaultExecutionTimeout,
+	})
+	if err != nil {
+		return err
+	}
+	defer maxRows.Close()
```
begelundmuller, line 159: Why does it need two queries? Is there any way to do just one?

If it's truly necessary, I think it should run them concurrently using `errgroup` to reduce latency.

begelundmuller (changes requested):

runtime/queries/column_time_range.go
pjain1: it needs 2 queries see the [description](https://github.com/rilldata/rill/pull/2584#issue-1747512744). Changed to run in parallel.

begelundmuller (approved):

PR #2587, opened 2023-06-08
Cloud UI: retain dashboard while reconciling

ericpgreen2:
TODO:
- [ ] Dashboard is shown during reconcile, but queries hang halfway through the reconcile 

**Problem**

While a project is reconciling, the dashboard is initially interactive. However, about halfway through the reconcile, queries hang, and a request fails with a 400 error "table does not exist".

In this case, the full reconcile takes ~30 seconds, and the queries hang for about half that time.
image

Here, a `toplist` request failed with a 400. Other times, a `timeseries` request fails with a 400.
image

Note: I'm testing by running `rill project reconcile --refresh` and interacting with the dashboard.


nishantmonu51:
This should now be unblocked by https://github.com/rilldata/rill/pull/2747 
Lets test it and see if we hit any new blockers here.

ericpgreen2:
> @ericpgreen2 : This should now be unblocked by #2747 Lets test it and see if we hit any new blockers here.

@nishantmonu51, @begelundmuller, I tried this branch again. I've updated the PR description with a new problem I'm observing.

begelundmuller:
@ericpgreen2 I tested the PR and had these findings:

- The queries did not hang for me, but did take longer to resolve during source refresh. I think this is connected to the 2 slot default on cloud, which translates to two concurrent DuckDB connections, of which background source ingestion grabs one. So all other queries become serialized and one slow query can cause everything to hang. Could this explain what you saw? I think this is acceptable, and should be less of an issue for deployments with more than 2 slots.

- I was able to reproduce the 400 error in one out of three source refreshes where I was constantly clicking filters. I think it depends on hitting the exact moment we drop the old table and rename the temporary table. I'm exploring options to do that transactionally on the backend, but it would also help if the cloud UI could show individual dashboard request errors inline (like it does on local) instead of blocking the entire page. Would that be hard to do?

- I also found that the UI is not rendering the list of dashboards when a refresh is running:

Screenshot 2023-08-10 at 16 24 12

begelundmuller:
UPDATE: This PR should fix the 400 error: https://github.com/rilldata/rill/pull/2897

ericpgreen2:
@begelundmuller, on your points:

1. Yeah, that sounds like a good explanation. 
2. Adding error boundaries to dashboard components is definitely something we need to do. Though thanks to your PR, it doesn't sound like an error boundary is necessary for this PR anymore.
3. Just fixed.

Given all this, I'll mark this PR as ready.

begelundmuller (approved):

PR #2590, opened 2023-06-09
Add dashboard YAML property for percent of total

begelundmuller:
Adds a `show_percent_of_total` property on dashboard measures. For example:

```
measures:
    - label: Measure 0
      name: measure_0
      expression: count(c0)
      description: 
      format_preset: humanize
      show_percent_of_total: true
```


skokenes (commented):
Looks good to me but it sounds like we haven't come to a decision yet on whether we want to do this or not, and how we want to name it. So let's hold on merging this

bcolloran (approved):
looks good to me

PR #2592, opened 2023-06-09
feat: date range picker

skokenes:
## Checklist
- [x] Manual verification
- [x] Unit test coverage
- [x] E2E test coverage
- [x] Needs manual QA?

#### Issue addressed: 
[Implement custom visual date picker](https://www.notion.so/rilldata/Implement-visual-custom-date-range-picker-6c4833af46de4f2f8561fdb8a3cad44a)

#### Details:
This PR extends a library called [Litepicker](https://litepicker.com/) to add a custom date range picker to our time selector.

Originally, we were going to use the successor to Litepicker called `easepick`. easepick had a much more robust plugin system for overriding behaviors. However, easepick's licensing is prohibitive so we've had to use the predecessor library Litepicker instead, since it has an MIT license.

Unfortunately, Litepicker is not nearly as OOB extensible as easepick. This means we've had to do some hacky things here to use it:
- Our `CustomPicker` class extends and overwrites many Litepicker methods. This is done in JS, not TS, because many of Litepicker's internals are private and thus throw TS errors
- We had to copy the DateTime class (`datetime.ts`) code out of Litepicker, since it is not exported from the package but the internals of Litepicker expect the class instances. We should take a glance at that code to make sure it is not doing anything strange, but I think we can skip a deep review of it. Essentially lets consider it a bundled 3rd party dependency
- The CSS in Litepicker.svelte is quite complex in order to override a lot of CSS that comes with Litepicker that is undesirable

In our implementation of CustomPicker, we require providing two input elements to bind the start and end date to. The CustomPicker manages updating the state of those elements when using the calendar, as well as responding to changes in the inputs after blur events.


ericpgreen2:
Sweet. Just gave it a quick look as is. Can we use Typescript as much as possible in the `date-picker` directory?

web-common/src/components/date-picker/Litepicker.svelte
```diff
@@ -0,0 +1,262 @@
+<script lang="ts">
+  import { onDestroy, onMount, createEventDispatcher } from "svelte";
+  import Custompicker from "./custom-picker";
+
+  export let startEl,
+    endEl,
+    defaultStart,
+    defaultEnd,
+    openOnMount = false,
+    editingDate = 0;
```
skokenes, line 10: Add type for 0 | 1

web-common/src/components/date-picker/Litepicker.svelte
```diff
@@ -0,0 +1,262 @@
...
+
+  let datepicker, picker;
```
skokenes, line 12: datepicker = container

web-common/src/components/date-picker/Litepicker.svelte
```diff
@@ -0,0 +1,262 @@
...
+  });
+
+  onDestroy(() => {
+    startEl.removeEventListener("focus", handleStartFocus);
+    endEl.removeEventListener("focus", handleEndFocus);
+    picker?.destroy();
+  });
+</script>
+
+<div bind:this={datepicker} class="w-0 h-0 absolute top-0 left-full" />
```
skokenes, line 79: Comment about placing in corner

web-common/src/components/date-picker/Litepicker.svelte
```diff
@@ -0,0 +1,262 @@
...
+      picker.show();
+      picker.setEditingDate(1);
+    }
+  };
+
+  onMount(() => {
+    picker = new Custompicker({
+      element: datepicker,
+      autoApply: false,
+      autoRefresh: true,
```
skokenes, line 34: These can probably be removed, legacy litepicker props

web-common/src/features/dashboards/time-controls/CustomTimeRangeInput.svelte
```diff
@@ -26,15 +27,20 @@
   // functions for extracting the right kind of date string out of
   // a Date object. Used in the input elements.
   export function getDateFromObject(date: Date): string {
-    return getDateFromISOString(date.toISOString());
+    return date.toLocaleDateString([], {
+      timeZone: "UTC",
+      year: "numeric",
+      month: "2-digit",
+      day: "2-digit",
```
skokenes, line 34: Double check this. Maybe we don't want to override locale settings here

web-common/src/components/date-picker/Litepicker.svelte
skokenes: We do need these settings after all

ericpgreen2:
This is great! Approving.

Though, regarding the "scroll to month" functionality, here's a scenario I ran into:
1. Edit the start date to a new month
2. Click outside to close the calendar
3. Click to edit the start date again
4. The calendar opens on the month of the original start date, not on the edited started date, where the date is currently highlighted

ericpgreen2 (approved):

skokenes:
> This is awesome! Approving.
> 
> Though, one note regarding the "scroll to month" functionality, here's a scenario I ran into:
> 
> 1. Edit the start date to a new month
> 2. Click outside to close the calendar
> 3. Click to edit the start date again
> 4. The calendar opens on the month of the original start date, not on the edited started date, where the date is currently highlighted

Let's take this item in a follow on PR.
Can you make a video showing these steps? I could not reproduce on my end so I must be doing it wrong

PR #2597, opened 2023-06-12
Cloud UI: fix `Cannot read properties of undefined`

ericpgreen2:
#2563 introduced the below bug to the Cloud UI. This PR fixes it.

We QA'd the prior PR against Rill Developer. We should have QA'd it against Rill Cloud too, and we should have basic e2e tests for Rill Cloud.

image


skokenes (approved):
Nothing easy 😅

PR #2599, opened 2023-06-12
Support default project level model materialization

begelundmuller:


runtime/services/catalog/artifacts/sql/sql.go
```diff
@@ -24,13 +24,16 @@ func init() {
...
 
-func (r *artifact) DeSerialise(ctx context.Context, filePath, blob string) (*drivers.CatalogEntry, error) {
+func (r *artifact) DeSerialise(ctx context.Context, filePath, blob string, forceMaterialize bool) (*drivers.CatalogEntry, error) {
 	name := fileutil.Stem(filePath)
 	// extract materialize option before sanitizing query as it will remove that comment
 	materialize := parseMaterializationInfo(blob)
 	if materialize == MaterializeInvalid {
 		return nil, errors.New("invalid materialize type")
 	}
+	if forceMaterialize {
```
nishantmonu51, line 34: Instead of force materialization how about changing this to defaultMaterialize and only use it when materialize = MaterializeUnspecified ? 

Essentially, you can set a default at a project level, but individual models can still override the default by explicitly setting up a materialization strategy.

begelundmuller: Good idea, updated the PR

runtime/services/catalog/artifacts/sql/sql.go
```diff
@@ -24,13 +24,20 @@ func init() {
...
 
-func (r *artifact) DeSerialise(ctx context.Context, filePath, blob string) (*drivers.CatalogEntry, error) {
+func (r *artifact) DeSerialise(ctx context.Context, filePath, blob string, materializeDefault bool) (*drivers.CatalogEntry, error) {
 	name := fileutil.Stem(filePath)
 	// extract materialize option before sanitizing query as it will remove that comment
 	materialize := parseMaterializationInfo(blob)
 	if materialize == MaterializeInvalid {
 		return nil, errors.New("invalid materialize type")
 	}
+	if materialize == MaterializeInferred {
```
nishantmonu51, line 34: this probably also need to match for MaterializeUnspecified as well..

nishantmonu51: another way could be to pass in default to parseMaterializationInfo(blob) as well and resolve it there itself.

begelundmuller: Sorry, it's a mistake, it should only be for `MaterializeUnspecified` (`inferred` is never used, but IIRC was originally added to leave room for inferring materialization based on heuristics).

nishantmonu51 (approved):
👍

PR #2602, opened 2023-06-12
Add `else` to DuckDB SQL dialect

ericpgreen2:
This PR adds `else` to our DuckDB SQL Dialect. Practically, this includes `else` in CodeMirror's syntax highlighting and autocomplete.


nishantmonu51 (approved):
👍

PR #2605, opened 2023-06-13
HACKY motherduck integration for runtime olap driver

k-anshul:


runtime/drivers/motherduck/duckdb.go
```diff
@@ -0,0 +1,223 @@
...
+		"INSTALL 'json'",
+		"LOAD 'json'",
+		"INSTALL 'icu'",
+		"LOAD 'icu'",
+		"INSTALL 'parquet'",
+		"LOAD 'parquet'",
+		"INSTALL 'httpfs'",
+		"LOAD 'httpfs'",
+		"SET max_expression_depth TO 250",
+	}
```
begelundmuller, line 41: Load motherduck extension and `SET` token from `cfg`

runtime/drivers/motherduck/duckdb.go
```diff
@@ -0,0 +1,223 @@
...
+
+	if cfg.DBFilePath != "" {
+		err = os.Remove(cfg.DBFilePath)
+		if err != nil {
+			return err
+		}
+		// Hacky approach to remove the wal file
+		_ = os.Remove(cfg.DBFilePath + ".wal")
+	}
```
begelundmuller, line 121: Will duckdb still create `.db` files locally when using motherduck? Or could it just always pass `:memory:`?

runtime/drivers/motherduck/olap.go
```diff
@@ -0,0 +1,184 @@
...
+	"github.com/rilldata/rill/runtime/drivers"
+	"github.com/rilldata/rill/runtime/pkg/observability"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/metric"
+)
+
+// Create instruments
+var (
+	meter                 = otel.Meter("github.com/rilldata/rill/runtime/drivers/duckdb")
```
begelundmuller, line 21: Update to motherduck

begelundmuller (changes requested):

runtime/drivers/motherduck/duckdb.go
k-anshul: No it doesn't. Also realised that we need to drop the instance specific database. Added a hacky impl to do that.

begelundmuller (approved):

PR #2606, opened 2023-06-13
Implement graceful handling of DuckDB fatal errors

begelundmuller:


runtime/drivers/duckdb/duckdb.go
```diff
@@ -206,18 +234,78 @@ func (c *connection) acquireOLAPConn(ctx context.Context, priority int) (*sqlx.C
...
+// checkErr marks the DB for reopening if the error is an internal DuckDB error.
+// In all other cases, it just proxies the err.
+// It should be wrapped around errors returned from DuckDB queries. **It must be called while still holding an acquired DuckDB connection.**
+func (c *connection) checkErr(err error) error {
+	if err != nil {
+		if strings.HasPrefix(err.Error(), "INTERNAL Error:") || strings.HasPrefix(err.Error(), "FATAL Error") {
+			c.dbCond.L.Lock()
+			c.dbReopen = true
+			c.logger.Error("encountered internal DuckDB error - scheduling reopen of DuckDB", zap.Error(err))
+			c.dbCond.L.Unlock()
```
k-anshul, line 307: Use `defer unlock` construct unless there is some specific reason to use this.

runtime/drivers/duckdb/duckdb.go
```diff
@@ -206,18 +234,78 @@ func (c *connection) acquireOLAPConn(ctx context.Context, priority int) (*sqlx.C
...
+
+// checkErr marks the DB for reopening if the error is an internal DuckDB error.
+// In all other cases, it just proxies the err.
+// It should be wrapped around errors returned from DuckDB queries. **It must be called while still holding an acquired DuckDB connection.**
+func (c *connection) checkErr(err error) error {
+	if err != nil {
+		if strings.HasPrefix(err.Error(), "INTERNAL Error:") || strings.HasPrefix(err.Error(), "FATAL Error") {
+			c.dbCond.L.Lock()
+			c.dbReopen = true
+			c.logger.Error("encountered internal DuckDB error - scheduling reopen of DuckDB", zap.Error(err))
```
k-anshul, line 306: Should we also add the `DBFilePath`? If some users complain that they are saying lot of intermittent errors in their dashboards, it will be handy to look for these errors for their projects and ask them to materialize model.

runtime/drivers/duckdb/duckdb.go
```diff
@@ -206,18 +234,78 @@ func (c *connection) acquireOLAPConn(ctx context.Context, priority int) (*sqlx.C
...
+		err := conn.Close()
+		c.dbCond.L.Lock()
+		c.dbConnCount--
+		if c.dbConnCount == 0 && c.dbReopen {
+			c.dbReopen = false
+			err = c.reopenDB()
+			if err == nil {
+				c.logger.Info("reopened DuckDB successfully")
+			} else {
+				c.logger.Error("reopen of DuckDB failed - the handle is now permanently locked", zap.Error(err))
```
k-anshul, line 286: Also add the `DBFilePath` here in order to be able to track back to project?

runtime/drivers/duckdb/duckdb.go
begelundmuller: The DuckDB logger has `With(zap.String("instance_id", instanceID), ...`, so that will give us the info we need

runtime/drivers/duckdb/duckdb.go
begelundmuller: Done

k-anshul (approved):

PR #2608, opened 2023-06-13
Adding user_id to cli telemetry

AdityaHegde:
When we added the user id telemetry we didnt add send it with the actual event.


nishantmonu51 (approved):
👍

PR #2614, opened 2023-06-14
Admin/Runtime: Vault for connector credentials 

rakeshsharma14317:


begelundmuller:
Had a discussion about this – decided that having a dependency on Vault in the hot path seems too risky. We will probably explore an approach where encrypted credentials are stored in the DB, and the encryption key is set as an env var using Vault in the usual way.

Thank you for prototyping this!

PR #2616, opened 2023-06-14
quarterly

egor-ryashin:


begelundmuller (changes requested):
Looks like there are some utils in `runtime/queries/sqlutil.go` that also need to be updated

begelundmuller (approved):

PR #2619, opened 2023-06-14
Feat/raw data view

skokenes:
## Checklist
- [x] Manual verification
- [ ] Unit test coverage
- [x] E2E test coverage
- [x] Needs manual QA?

#### Issue addressed: 
[Source rows viewer in dashboards](https://www.notion.so/rilldata/UI-and-UX-for-showing-raw-rows-for-a-data-model-via-for-a-given-set-of-filters-39b5950b69064e5b8f149af5b33ca6d4)

#### Details:
Adds an accordion to the dashboards that can be popped up to see a sample of up to 10k rows of the filtered dashboard data.
The accordion label includes metrics for # of filtered rows and # of total rows in a dashboard model.
I had to modify the PreviewTable styles significantly because they were buggy when changing the size configuration for the table rows.
I tried to create custom hooks for the queries in RowsViewerAccordion, but it is not possible to do so until `svelte-query` allows us to pass in a custom QueryClient rather than trying to use context. This is because the custom hook needs to call the query outside a svelte component context to work properly for filtered rows.

We could use manual QA on this to confirm that all of our tables (Sources, Models, Dimension Details, and RowViewer) look correct.


ericpgreen2:
I'm on a 27" screen, looking at the below dashboard, and I didn't realize the Source Viewer bar was there for a good 15 seconds. It was out of my field of focus. Can we make it a little bit bigger and/or use colors to make it pop a bit more?
image

ericpgreen2:
Similarly, because I have a lot of space on my screen, I'd like to drag the Source Viewer higher, so that I can see more rows - like how we can expand the Model Preview results.

I imagine these layout changes might be a bigger lift, so maybe it's more appropriate for a follow-on PR, but want to mention.

image

skokenes:
The design spec is already set for this release. 

If we want to iterate on the design further, that will have to be in another PRD for a future release. 

As far I can tell , this implementation matches the design spec provided by Janet for 0.28.

web-common/src/features/dashboards/rows-viewer-accordion/RowsViewerAccordion.svelte
```diff
@@ -0,0 +1,113 @@
...
+    on:click={toggle}
+    aria-label="Toggle rows viewer"
+  >
+    <span class="font-bold">Source Data</span>
+    {label}
+    {#if isOpen}
+      <CaretUpIcon size="14px" />
+    {:else}
+      <CaretDownIcon size="14px" />
+    {/if}
```
ericpgreen2, line 108: Nit: elsewhere we're using an animation like this to flip between up/down:

```svelte
<span class="{active ? '-rotate-180' : ''} transition-transform">
    <CaretDownIcon />
</span>
```

web-common/src/features/dashboards/rows-viewer-accordion/RowsViewerAccordion.svelte
ericpgreen2, line 1: What about moving `rows-viewer-accordion/RowsViewerAccordion.svelte` to `rows-viewer/RowsViewerAccordion.svelte` and removing the `rows-viewer-accoridon` directory?

web-local/test/ui/dashboards.spec.ts
```diff
@@ -112,6 +117,28 @@ describe("dashboards", () => {
...
     await playwrightExpect(page.getByText("275 -12 -4%")).toBeVisible();
 
+    // Check the row viewer accordion is updated
+    await playwrightExpect(
+      page.getByText("Source Data 275 of 100k rows")
+    ).toBeVisible();
+
+    // Check row viewer is collapsed
+    await playwrightExpect(
+      page.getByRole("button", { name: "7029" })
```
ericpgreen2, line 127: Nit: I'm not too sure what "7029" is. Is that a data point in a cell?

web-common/src/features/dashboards/rows-viewer-accordion/RowsViewerAccordion.svelte
```diff
@@ -0,0 +1,113 @@
...
+      inlineMeasures: [
+        {
+          name: "count",
+          expression: "count(*)",
+        },
+      ],
+    },
+    {
+      query: {
+        queryKey: ["dashboardAllRowsCt"],
```
ericpgreen2, line 30: Why the custom `queryKey`? If it's necessary, mind adding a code comment to explain?

ericpgreen2:
There's an issue in the Cloud UI:
image

ericpgreen2:
> The design spec is already set for this release.

That's fine, just wanted to leave my impressions.

Overall, this is awesome! Really nice feature & implementation. The only real issue is the Cloud UI bug. I'll approve it after that's fixed.

avaitla:
I cloned this branch then ran make. I tried it on some of our data but hit this error regarding a rate limit:

Screen Shot 2023-06-15 at 5 47 50 PM

Also the count seems incorrect?

Screen Shot 2023-06-15 at 5 57 02 PM



Maybe I didn't build it properly or something?

However if I use one of the existing sample datasets it does seem to work properly:

Screen Shot 2023-06-15 at 5 56 07 PM

web-common/src/features/dashboards/rows-viewer-accordion/RowsViewerAccordion.svelte
skokenes: I tried it but it has no UX affect because you can't eyes can't see the animation happening while the icon is moving to a different location on screen. So won't do

web-common/src/features/dashboards/rows-viewer-accordion/RowsViewerAccordion.svelte
skokenes: It should not be necessary, the queryKey generated by the custom hook should be enough to uniquely identify the query.

However, in practice I've discovered that this query frequently fails to put any data into the cache when using the autogenerated key. Instead, the query runs, the data is fetched from the server, but the query status remains at "loading" and the fetch status says "idle".

I will send you a video of this behavior. The only way I found to fix it for now is to provide a custom query key. I could use your help investigating the underlying issue; it smells like a svelte-query bug.

web-local/test/ui/dashboards.spec.ts
skokenes: Done

web-common/src/features/dashboards/rows-viewer-accordion/RowsViewerAccordion.svelte
skokenes: I added a comment about this

web-common/src/features/dashboards/rows-viewer-accordion/RowsViewerAccordion.svelte
skokenes: Done

skokenes:
I added a fallback for @avaitla 's scenario where too much data is attempted to be fetched due to a wide table or potentially very long table cell values. The fallback is that if the 10k row sample fails, we fallback to a 1k row sample.

This should work with most reasonable data. I will consult platform for an even more robust data fetching solution going forward.

avaitla:
Confirmed that does seem to resolve it for this dataset

skokenes:
@avaitla I extended the table height to show 8ish rows. We have it in our backlog to add the ability to drag to resize the table height, like in modeling.

PR #2625, opened 2023-06-15
 button group Component

bcolloran:
This PR creates a toggle button group components as requested by @jkhwu for the percent of total work. The components can be viewed in storybook ( `cd web-common` then `npm run storybook`).


bcolloran:
@skokenes -- I am confused about why adding a new component and stories for that component but not actually wiring that component up anywhere in the existing product, should cause our should cause e2e tests to fail. Any ideas?

skokenes:
I reran the tests and it passed, looks like a random failure. First one we have gotten in a long time, let's keep our eyes open for more of those.
I will review the rest of the PR tomorrow.

bcolloran:
Oh weird, i had it fail there twice in a row

attempt 1:
https://github.com/rilldata/rill/actions/runs/5284777456/jobs/9562555576

attempt 2:
https://github.com/rilldata/rill/actions/runs/5284777456/jobs/9562820417

web-common/src/components/button-toggle-group/ButtonToggleGroup.svelte
```diff
@@ -0,0 +1,126 @@
...
+  const selectedKey = writable(null);
+
+  // if a default key is provided, then select it if it is not disabled
+  // this Must be run reactively in case the default key changes or
+  //  the set of disabled keys changes
+  $: if (defaultKey !== undefined && !disabledKeys.includes(defaultKey)) {
+    selectedKey.update(() => defaultKey);
+  } else {
+    selectedKey.update(() => null);
+  }
```
skokenes, line 41: In most implementations that I've seen of generic UI components, a "default value" (defaultKey) is treated like an "initialValue" and is _not reacted_ to when it changes. I.e., it only matters what the first default value is, it would not overwrite the existing value when it changes.

At a minimum here, wouldn't you want to check that selectedKey isn't null before trying to apply a default? Isn't the default akin to a "fallback" key?

skokenes: If the user doesn't provide a default key, why set the selected key to null?

web-common/src/components/button-toggle-group/ButtonToggleGroup.svelte
```diff
@@ -0,0 +1,126 @@
...
+
+  const firstKey = writable(null);
+  const lastKey = writable(null);
```
skokenes, line 44: Can't these just be derived from subButtons? like firstKey = subButtons.at(0), lastKey = subButtons.at(-1)

web-common/src/components/button-toggle-group/ButtonToggleGroup.svelte
```diff
@@ -0,0 +1,126 @@
...
+      onDestroy(() => {
+        const i = subButtons.indexOf(subButtonKey);
+        subButtons.splice(i, 1);
+        selectedKey.update((current) =>
+          current === subButtonKey
+            ? subButtons[i] || subButtons[subButtons.length - 1]
+            : current
+        );
+      });
+    },
```
skokenes, line 91: Because you are manually maintaining firstKey/lastKey as stores rather than deriving the state from subButtons, wouldn't you need to reset the firstKey/lastKey here? For example, if the subButton being removed is the first or last button...

web-common/src/components/button-toggle-group/ButtonToggleGroup.svelte
```diff
@@ -0,0 +1,126 @@
...
+      // toggle the sub button: if it is selected, then deselect it;
+      // otherwise, select it
+      if (lastSelection === subButton) {
+        selectedKey.set(null);
+        dispatch("deselect-subbutton", subButton);
+      } else {
+        dispatch("deselect-subbutton", lastSelection);
+
+        selectedKey.set(subButton);
+        dispatch("select-subbutton", subButton);
```
skokenes, line 111: Do we need separate events for "deselect/select", or could we just have one "selectionChange" event that always emits what the new selection is, whether thats `null` or a specific subButton key?

web-common/src/components/button-toggle-group/README.md
```diff
@@ -0,0 +1,58 @@
...
+`defaultKey: number | string` -- The key of the default selection.If undefined, the first sub button will be selected by default.
+
+`disabledKeys: (number | string)[] = [];` -- An array containing the keys of any sub buttons that are disabled.
+
+
+`<ButtonToggleGroup>` Dispatches the following events:
+
+"select-subbutton" -- When a sub button is selected, this event is dispatched with the key of the sub button
+
+"deselect-subbutton" -- When a sub button is deselected this event is dispatched with the key of the sub button. Note that this event may be fired by itself if no selection is required, indicating that the only selected button has been untoggled. If another sub button has been selected, then this event will be dispatched immediately before the selection event.
```
skokenes, line 37: See my previous comment; its unclear to me why you need separate events for these as opposed to just an event indicating the latest selected value

web-common/src/components/button-toggle-group/SubButton.svelte
```diff
@@ -0,0 +1,72 @@
...
+
+  $: textStyle = isDisabled
+    ? "text-gray-400"
+    : "text-gray-700 hover:text-gray-900 ";
+
+  $: bgStyle = isDisabled
+    ? "bg-white"
+    : isSelected
+    ? "bg-gray-100 hover:bg-gray-200 "
+    : "bg-white hover:bg-gray-50 ";
```
skokenes, line 43: Not important for this PR, but a side note: We have this pattern many places in our codebase where we need to add different class names based on a variety of conditions. I wonder if its worth adding a simple `classnames()` utility for this, where you could rewrite the above (and similar scenarios) as:
```javascript
bgStyle = classnames({
  "bg-white": isDisabled,
  "bg-gray-100 hover:bg-gray-200": !isDisabled && isSelected,
  "bg-white hover:bg-gray-50": !isDisabled && !isSelected
})
```

web-common/src/components/button-toggle-group/SubButton.svelte
```diff
@@ -0,0 +1,72 @@
...
+
+  // This is needed to make sure that the left and right most child
+  // elements don't break out of the border drawn by the parent element
+  $: isFirst = key === $firstKey;
+  $: isLast = key === $lastKey;
```
skokenes, line 48: See my comments about making these manually maintained stores, as opposed to deriving them from the single source of truth which is the `subButtons` array. This would prevent potential bugs like we have in `onDestroy()`, where it seems we forgot to update firstKey/lastKey stores

Perhaps `subButtons` is what needs to be a writable store, with firstKey as derived from that.

Or alternatively, just expose a function called `getKeyIndex()` or `isFirstKey()` from the ButtonGroup context

web-common/src/components/button-toggle-group/SubButton.svelte
```diff
@@ -0,0 +1,72 @@
...
+  $: roundings = `${isFirst ? "rounded-l" : ""} ${isLast ? "rounded-r" : ""} `;
+
+  $: finalStyles = `${baseStyles} ${roundings} ${textStyle} ${bgStyle}`;
+
+  $: tooltipText = isDisabled
+    ? tootips?.disabled
+    : isSelected
+    ? tootips?.selected
+    : tootips?.unselected;
+</script>
```
skokenes, line 58: This component never calls the context's "onDestroy()" method to deregister itself. Was that on purpose?
Doesn't it need a lifecycle hook here to unregister itself from the ButtonGroup if it gets removed?

web-common/src/components/button-toggle-group/SubButton.svelte
```diff
@@ -0,0 +1,72 @@
...
+  import {
+    ButtonGroupContext,
+    buttonGroupContext,
+  } from "./ButtonToggleGroup.svelte";
+
+  import Tooltip from "../tooltip/Tooltip.svelte";
+  import TooltipContent from "../tooltip/TooltipContent.svelte";
+
+  export let key: number | string;
+  export let tootips: {
```
skokenes, line 12: Not important but while we're here, this is a typo (should say "tooltips", no?)

skokenes (approved):
Excellent work! I left a few comments. I approved anyways because I don't think we necessarily need to hold this up for those, but there may be a potential bug so it would be great if you could at least respond to the comments with whether we should make changes now, after this PR, or never.

Also, I'd love it if you could add a unit test for this component. See https://github.com/rilldata/rill/blob/main/web-common/src/components/date-picker/Litepicker.spec.ts for an example. For this component, would be nice to test:
- does it render sub buttons
- does it enforce default
- does it enforce always 1 selected
- does it call click on sub buttons that are not disabled
- does it not call click on sub buttons that are disabled
- can you add/remove sub buttons from the ButtonGroup and it still works properly

web-common/src/components/button-toggle-group/__stories__/ButtonToggleGroup.stories.svelte
```diff
@@ -0,0 +1,91 @@
...
+  const deltaTooltips = {
+    selected: "Hide percent change",
+    unselected: "Show percent change",
+    disabled: "To show percent change, select a comparison period abovec",
+  };
+</script>
+
+<Meta title="Button toggle group stories" />
+
+<Story name="Button toggle group variations (tables)">
```
skokenes, line 23: Should one (or all) of these stories show storybook actions to show the on:selected event working?

web-common/src/components/button-toggle-group/ButtonToggleGroup.svelte
skokenes: Does this component support a "controlled" approach, or just "uncontrolled"? If I want the parent component using the ButtonGroup to control the state, how do I do that? It almost feels like "defaultKey" is the way to do that, in which case its not properly named. See this example from MUI https://mui.com/material-ui/react-toggle-button/#ToggleButtons.tsx

web-common/src/components/button-toggle-group/SubButton.svelte
bcolloran: Good catch. The source I was going off of (which was recommended by none other than Rich Harris, and I suspect written by him!) also forgot to call onDestroy in the child elts, and since I was following a template from that source, I must admit I didn't look too deeply at everything. Very nice to have a second pair of eyes! :-)

bcolloran: hmm TIL... never mind it turns out that what they were doing was a bit more subtle than I thought. Also it turns out that Svelte is once again a bit mind-twisting for me.

Checked the docs at https://svelte.dev/tutorial/ondestroy -- the key line is "While it's important to call lifecycle functions during the component's initialisation, it doesn't matter _where_ you call them from." -- I think the `onDestroy` method is called during the initialization of the the subbutton component (since `registerSubbutton` is called during initialization) even though the code is located in the parent component.

There are some patterns in Svelte that I really do not like...

skokenes: Wow, that is kinda mind blowing. I did not catch that `onDestroy` was within `registerSubbutton`

web-common/src/components/button-toggle-group/ButtonToggleGroup.svelte
bcolloran: Hmm yeah this is pretty muddy -- thanks for pushing me to look more deeply into it. I guess it's probably more flexible as a controlled component. I'll rework it along those lines.

web-common/src/components/button-toggle-group/__stories__/ButtonToggleGroup.stories.svelte
bcolloran: Great idea, added storybook action to show the events working (and not working when a button is disabled)

web-common/src/components/button-toggle-group/ButtonToggleGroup.svelte
bcolloran: Yep you're right. Made that change.

web-common/src/components/button-toggle-group/SubButton.svelte
bcolloran: I'm torn on this. I can see how that might be useful, but it's also another layer of abstraction beyond plain JS string manipulation -- if I see a function like that, I'll always have to wonder "hmmm what is that function doing?" and go look at it's code to make sure it doesn't do anything surprising, vs conditionals over strings being totally explicit in-line.

web-common/src/components/button-toggle-group/__stories__/ButtonToggleGroup.stories.svelte
bcolloran: great idea, added actions

bcolloran:
Hey @skokenes , thanks for the amazing review and for pushing me to look a little more deeply at this component. I got sidetracked on all that build stuff last week, and then it took a little longer than I expected to get this out, but it was good to spend time getting spun up on how one tests svelte components -- I'm glad to see us investing in that going forward.

I ended up substantially reworking the component to be much simpler fully controlled component. I think that you were right that it was kind of muddy about whether it was controlled or not. Looking at the example from MUI and others, it seems like for things like radio buttons they often end up in that muddy middle ground of being semi controlled since the radio functionality demands a bit of internal state. I decided to just forego all of that since A more generic button group will be satisfactory for the percent of total work, and may be useful in other places. Because of that I also renamed the component from ToggleButtonGroup to plain ButtonGroup (which unfortunately means all of your comments don't point to the same files).

Anyhoo, I think this should be ready for another review. Thanks!

PR #2627, opened 2023-06-16
Put timezone to UTC

nishantmonu51:
Currently we do not explicitly support timezones. 
Until we figure out timezone support on applications side, use UTC as the timezone on runtime.


begelundmuller (approved):
LGTM

PR #2628, opened 2023-06-16
Use 24 hour format for hours

djbarnwal:
## Checklist
- [x] Manual verification
- [ ] Unit test coverage
- [ ] E2E test coverage
- [x] Needs manual QA?

#### Issue addressed: 
The timestamps returned from the runtime were not being displayed properly. 
https://rilldata.slack.com/archives/CTZ8XBQ85/p1686876647920379

#### Details:
We were using a 12 hour clock while formatting the timestamps. This PR switches it to a 24 hour clock. 
[D3 Locale format types](https://github.com/d3/d3-time-format#locale_format)

## Steps to Verify
1. Create a model
2. `SELECT cast('2022-01-01 00:00:00' as timestamp) as ts, cast('2022-01-01' as timestamptz) tsz`
3. The resulted rows should display `2022-01-01 00:00:00`

### DIFF
--- a/web-common/src/lib/formatters.ts
+++ b/web-common/src/lib/formatters.ts
@@ -73,7 +73,7 @@ export function removeTimezoneOffset(dt: Date) {
 }
 
 export const standardTimestampFormat = (v, type = "TIMESTAMP") => {
-  let fmt = timeFormat("%Y-%m-%d %H:%M:%S");
+  let fmt = timeFormat("%Y-%m-%d %H:%M:%S Z");
   if (type === "DATE") {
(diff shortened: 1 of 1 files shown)


begelundmuller:
Hey, given earlier confusion about time zones, I think you should consider including the time zone in the string (for UTC, it will just be one extra letter, `Z`). For example by using ISO 8601 format instead (`%Y-%m-%dT%H:%M:%S.%LZ`). Including the time zone in the string is the only way to eliminate any confusion about what time is being shown.

PR #2636, opened 2023-06-16
Exporting for Excel XLSX format

egor-ryashin:
Closes https://github.com/rilldata/rill/issues/2354


runtime/queries/metricsview.go
```diff
@@ -304,3 +305,48 @@ func writeCSV(meta []*runtimev1.MetricsViewColumn, data []*structpb.Struct, writ
...
+	}
+
+	if err := sw.SetRow("A1", headers, excelize.RowOpts{Height: 45, Hidden: false}); err != nil {
+		return err
+	}
+
+	row := make([]interface{}, 0, len(meta))
+	for i, s := range data {
+		for _, f := range s.Fields {
+			row = append(row, f.AsInterface())
```
begelundmuller, line 332: Just double-checking – it supports maps and lists being returned from `AsInterface()`?

runtime/server/downloads.go
```diff
@@ -76,17 +76,18 @@ func (s *Server) downloadHandler(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 
+	w.Header().Set("X-Content-Type-Options", "nosniff")
+	if request.Format == runtimev1.ExportFormat_EXPORT_FORMAT_CSV {
+		w.Header().Set("Content-Type", "text/csv")
+	} else {
+		w.Header().Set("Content-Type", "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet")
+	}
```
begelundmuller, line 84: Can we make the case explicitly check for `XLSX` and return an error if not CSV/XLSX? (e.g. in case data is corrupted)

begelundmuller (changes requested):

begelundmuller:
One other thing – it looks like the `go.mod` file only has indirect dependencies added, but there's a direct dependency from the code. Can you try to run `go mod tidy` and see if it rewrites it?

begelundmuller (approved):

PR #2637, opened 2023-06-16
Time range alias changes

djbarnwal:
## Checklist
- [x] Manual verification
- [ ] Unit test coverage
- [ ] E2E test coverage
- [x] Needs manual QA?

#### Issue addressed: 
1. Re config time ranges such that the end is exclusive
2. Pretty format the time ranges where for dates end is inclusive and for time end is exlusive
3. Enable comparison selector for all time ranges. Default to no comparison if those ranges don't have a valid default comparison range
4. Change "Last year" to "last 12 months" as it was similar to YTD. Also the alias "Last year" was confusing

**Note** : Merge after #2131


web-common/src/components/date-picker/Litepicker.svelte
```diff
@@ -8,7 +8,9 @@
     defaultStart: string,
     defaultEnd: string,
     openOnMount = false,
-    editingDate: 0 | 1 = 0;
+    editingDate: 0 | 1 = 0,
+    min: string,
+    max: string;
```
skokenes, line 13: Can you explain the purpose of min/max addition here?

skokenes (approved):
I think this is good to merge, but I did have a question about `min/max` that you added to Litepicker.svelte so I can confirm that its doing what you think its doing...

web-common/src/components/date-picker/Litepicker.svelte
djbarnwal: We have custom range selector in comparison as well. In case there is no comparison, we fallback to the all time range.

**Before**

image

**After**

image

PR #2640, opened 2023-06-18
2347: Admin/Runtime: Rate limiting (per operation, per auth/ip)

### LINKED ISSUE #2347: Admin/Runtime: Rate limiting
See details in [Notion](https://www.notion.so/rilldata/Requirements-a1fa12e88b994e5e9a63c879deda6189?p=2b5a1212303948fb8fa0f1a1e417078f&pm=s)

esevastyanov:
Closes #2347


admin/server/auth/handlers.go
```diff
@@ -24,7 +25,7 @@ const (
 // RegisterEndpoints adds HTTP endpoints for auth.
 // The mux must be served on the ExternalURL of the Authenticator since the logic in these handlers relies on knowing the full external URIs.
 // Note that these are not gRPC handlers, just regular HTTP endpoints that we mount on the gRPC-gateway mux.
-func (a *Authenticator) RegisterEndpoints(mux *http.ServeMux) {
+func (a *Authenticator) RegisterEndpoints(mux *http.ServeMux, rrl *ratelimit.RequestRateLimiter) {
```
begelundmuller, line 28: Pass `ratelimit.Limiter`

cli/cmd/admin/start.go
```diff
@@ -176,6 +178,16 @@ func StartCmd(cliCfg *config.Config) *cobra.Command {
...
+				redisAddr := conf.RedisAddr
+				if redisAddr == "" {
+					// Start a miniredis (in-memory) server (is used for API rate limiting).
+					mr, err := miniredis.Run()
+					if err != nil {
+						panic(err)
+					}
+					redisAddr = mr.Addr()
+					defer mr.Close()
+				}
```
begelundmuller, line 190: Create the Redis client here, so it can be reused in other components in the future. If `RedisAddr` is empty, it can just be uninitialized. 

Then after, it can get a limiter by calling `ratelimit.NewRedis(redis)` if the Redis client is initialized, or `ratelimit.NewNoop()` if no Redis was initialized.

The usage of `miniredis` can then be moved to the test cases, so it doesn't get included in the production binary and can't accidentally run in production.

The init code for the above will need to be copied in the `runtime` package, but that's fine – it's just a few lines, and in return the code is decoupled :)

cli/cmd/admin/start.go
```diff
@@ -61,6 +62,7 @@ type Config struct {
 	EmailSenderEmail       string                 `split_words:"true"`
 	EmailSenderName        string                 `split_words:"true"`
 	EmailBCC               string                 `split_words:"true"`
+	RedisAddr              string                 `default:"" split_words:"true"`
```
begelundmuller, line 65: I would propose renaming it to `RedisURL` and then using `redis.ParseURL` to parse it into a Redis options object. This enables setting passwords and other config as well without.

runtime/pkg/ratelimit/ratelimiter.go
```diff
@@ -0,0 +1,143 @@
...
+	"math"
+	"net/http"
+	"reflect"
+)
+
+type RequestRateLimiter struct {
+	*redis_rate.Limiter
+}
+
+func NewRequestRateLimiter(redisAddr string) *RequestRateLimiter {
```
begelundmuller: As mentioned above, introduce an interface for a `Limiter` and then have implementations `ratelimit.RedisLimiter` and `ratelimit.Noop (that just allows all calls) initially.

runtime/pkg/ratelimit/ratelimiter.go
```diff
@@ -0,0 +1,143 @@
...
+
+var Unlimited = redis_rate.PerSecond(math.MaxInt)
+
+var Forbidden = redis_rate.Limit{}
+
+func isAnonymous(ctx context.Context) bool {
+	claims := auth.GetClaims(ctx)
+	isAnonymous := claims == nil || claims.Subject() == ""
+	return isAnonymous
+}
```
begelundmuller, line 124: The `admin/server` and `runtime/server` have different auth implementations, so this will only validate runtime requests. See earlier note about implementing the checks in the respective servers instead to prevent coupling the `ratelimit` package to a specific auth implementation.

admin/server/server.go
```diff
@@ -100,11 +103,13 @@ func New(logger *zap.Logger, adm *admin.Service, issuer *runtimeauth.Issuer, opt
 		authenticator: authenticator,
 		issuer:        issuer,
 		urls:          newURLRegistry(opts),
+		limiter:       ratelimit.NewRequestRateLimiter(opts.RedisAddr),
 	}, nil
 }
 
 // ServeGRPC Starts the gRPC server.
 func (s *Server) ServeGRPC(ctx context.Context) error {
+	grpcReqRateLimiter := s.limiter.Middleware().WithAnonLimit(ratelimit.Forbidden).WithAuthLimit(ratelimit.Default)
```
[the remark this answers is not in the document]
esevastyanov: I found only one difference between `admin` and `runtime` in terms of rate limiting - the way how claims are organised and hence how requests are classified as anonymous/authenticated. Thus I decided to generalize the logic of intercepters in `ratelimit.Interceptor` but the way of how ctx/requests are classified is defined separately in `admin` and `server` packages

cli/cmd/admin/start.go
esevastyanov: Done

runtime/pkg/ratelimit/ratelimiter.go
esevastyanov: Thanks for the guide, looks better now

cli/cmd/admin/start.go
esevastyanov: good catch. I haven't noticed that `miniredis` is for testing only

runtime/pkg/ratelimit/ratelimiter.go
esevastyanov: Partly decoupled. It is possible decouple fully but this duplicates same pieces of code multiple times

admin/server/auth/handlers.go
esevastyanov: Done

admin/server/server.go
```diff
@@ -51,6 +52,7 @@ type Options struct {
 	GithubAppWebhookSecret string
 	GithubClientID         string
 	GithubClientSecret     string
+	RedisAddr              string
```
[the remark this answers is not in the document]
esevastyanov: Done

begelundmuller (changes requested):
In addition to the minor comments below, I'd like to revisit the earlier comment about using something like a `checkRateLimit` func. Some concerns:

- The rate limits are now enforced at the RPC-level, but should in most cases be enforced generally for many RPCs.
- The control flow leading to a specific enforcement is pretty complicated to trace right now. There is a non-trivial dependency between the place the middleware is added, the `CtxInspector`, and the interceptor implementation.

Here's an example of the rate limiting logic we need: the rate limit of 60 RPS should apply across all `admin` requests, perhaps with a few exceptions (when reconciling, the frontend polls on `GetProject` once per second, so it might need a higher individual limit). For `/auth`, the 10 RPS limit should apply across all endpoints (though it might need to be a little higher given callbacks), but we might need to customize a lower one for `/auth/oauth/device` specifically (recommended in the OAuth device flow spec).

A basic check function, even if repetitive, would make it easy to express this logic with a few `if` conditions, and as importantly, it would be easy to review and reason about.

admin/server/auth/handlers.go
```diff
@@ -24,17 +27,25 @@ const (
...
+		LimiterHTTPHandler("/auth/logout", limiter, ratelimit.Sensitive, http.HandlerFunc(a.authLogout))))
+	inner.Handle("/auth/logout/callback", otelhttp.WithRouteTag("/auth/logout/callback",
+		LimiterHTTPHandler("/auth/logout/callback", limiter, ratelimit.Sensitive, http.HandlerFunc(a.authLogoutCallback))))
+	inner.Handle("/auth/oauth/device_authorization", otelhttp.WithRouteTag("/auth/oauth/device_authorization",
+		LimiterHTTPHandler("/auth/oauth/device_authorization", limiter, ratelimit.Sensitive, http.HandlerFunc(a.handleDeviceCodeRequest))))
+	inner.Handle("/auth/oauth/device", otelhttp.WithRouteTag("/auth/oauth/device",
+		LimiterHTTPHandler("/auth/oauth/device", limiter, ratelimit.Sensitive, a.HTTPMiddleware(http.HandlerFunc(a.handleUserCodeConfirmation))))) // NOTE: Uses auth middleware
+	inner.Handle("/auth/oauth/token", otelhttp.WithRouteTag("/auth/oauth/token",
+		LimiterHTTPHandler("/auth/oauth/token", limiter, ratelimit.Sensitive, http.HandlerFunc(a.getAccessToken))))
 	mux.Handle("/auth/", observability.Middleware("admin", a.logger, inner))
```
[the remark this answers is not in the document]
esevastyanov: This sets a single key for all handlers. Do we want it to be one for all handlers or on per route basis?

admin/server/github.go
```diff
@@ -138,15 +139,20 @@ func (s *Server) GetGitCredentials(ctx context.Context, req *adminv1.GetGitCrede
...
+		limiterHTTPHandler("/github/connect", limiter, ratelimit.Sensitive, http.HandlerFunc(s.githubConnect)))))
+	inner.Handle("/github/connect/callback", otelhttp.WithRouteTag("/github/connect/callback", s.authenticator.HTTPMiddleware(
+		limiterHTTPHandler("/github/connect/callback", limiter, ratelimit.Sensitive, http.HandlerFunc(s.githubConnectCallback)))))
+	inner.Handle("/github/auth/login", otelhttp.WithRouteTag("github/auth/login", s.authenticator.HTTPMiddleware(
+		limiterHTTPHandler("github/auth/login", limiter, ratelimit.Sensitive, http.HandlerFunc(s.githubAuthLogin)))))
+	inner.Handle("/github/auth/callback", otelhttp.WithRouteTag("github/auth/callback", s.authenticator.HTTPMiddleware(
+		limiterHTTPHandler("github/auth/callback", limiter, ratelimit.Sensitive, http.HandlerFunc(s.githubAuthCallback)))))
+	inner.Handle("/github/post-auth-redirect", otelhttp.WithRouteTag("github/post-auth-redirect", s.authenticator.HTTPMiddleware(
+		limiterHTTPHandler("github/post-auth-redirect", limiter, ratelimit.Sensitive, http.HandlerFunc(s.githubRepoStatus)))))
 	mux.Handle("/github/", observability.Middleware("admin", s.logger, inner))
```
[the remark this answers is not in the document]
esevastyanov: Same here + `/github/webhook` is called by Github and I suppose it should be limited differently?

admin/server/auth/handlers.go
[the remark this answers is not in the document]
esevastyanov: If it is set for `/auth/` only then it is possible to look at `req.Path` in order to customize a limit or make an exclusion but in general this isn't right because path and route might be not equal

admin/server/server.go
```diff
@@ -174,10 +179,10 @@ func (s *Server) HTTPHandler(ctx context.Context) (http.Handler, error) {
 	mux.Handle("/.well-known/jwks.json", s.issuer.WellKnownHandler())
 
 	// Add auth endpoints (not gRPC handlers, just regular endpoints on /auth/*)
-	s.authenticator.RegisterEndpoints(mux)
+	s.authenticator.RegisterEndpoints(mux, s.limiter)
 
 	// Add Github-related endpoints (not gRPC handlers, just regular endpoints on /github/*)
-	s.registerGithubEndpoints(mux)
+	s.registerGithubEndpoints(mux, s.limiter)
```
begelundmuller, line 185: `s.limiter` appears unused (and can just be accessed through `s` anyway :) )

runtime/pkg/middleware/request.go
```diff
@@ -0,0 +1,105 @@
...
+		}
+
+		md := Metadata{ctx, req, method, observability.GrpcPeer(ctx)}
+		if err := fn(md); err != nil {
+			return nil, err
+		}
+
+		return handler(ctx, req)
+	}
+}
```
begelundmuller, line 52: Consider replacing these with `grpc_auth.UnaryServerInterceptor`, like here: https://github.com/rilldata/rill/blob/199453658f812528fb5c2ec7e87f599809e9ac2c/admin/server/server.go#L129C20-L129C20

Thoughts:

- You can use `grpc.Method(ctx)` to get the full method and `observability.GrpcPeer(ctx)` to get the peer.
- When rate limiting on a per-method basis, we should probably use the full method instead of parsing it, since the service prefix matters (we have multiple services, and they might use the same Redis).
- (It might still make sense to create a `grpcutil.ParseMethod(ctx)` func or something to get the individual service and method name)
- (Using the full method has another benefit, which is that we can use the generated [constants here](https://github.com/rilldata/rill/blob/199453658f812528fb5c2ec7e87f599809e9ac2c/proto/gen/rill/admin/v1/api_grpc.pb.go#L21) for the method names if we need to do manual overrides. The timeout middleware currently doesn't use it – but it probably should!)

runtime/pkg/middleware/request.go
```diff
@@ -0,0 +1,105 @@
...
+func (e *HTTPError) Error() string {
+	return e.Message
+}
+
+type Metadata struct {
+	Ctx    context.Context
+	Req    interface{}
+	Method string
+	Peer   string
+}
```
begelundmuller, line 105: Just FYI about Go: I believe there's a convention of passing `ctx` separately and as the first arg to a function (apart from certain rare cases)

begelundmuller (changes requested):
Looks very good now

Apart from the minor comments below, two questions:

- Do you think we should we apply the limits at service-level instead of method-level (maybe with a few exceptions)? My thinking is that it might better protect us from (intentional or unintentional) DoS attacks
- If we merge this, how good do you feel about people not running into these rate limits if they're good citizens? Any limits you think we should increase to be safe? (E.g. any chance someone restarts/reloads auth a bit quickly and gets a rate limit on the auth0 redirect or something? It wouldn't be good to accidentally reject people signing up)

runtime/pkg/middleware/request.go
esevastyanov: Agree, updated to use the method

esevastyanov:
>Do you think we should we apply the limits at service-level instead of method-level (maybe with a few exceptions)? My thinking is that it might better protect us from (intentional or unintentional) DoS attacks

We can switch to service-level if a method-level limit exceeded so that no multiple methods are utilized during DoS.

> If we merge this, how good do you feel about people not running into these rate limits if they're good citizens?

I've tripled the limits before we implement usage tracking and learn about outliers.

PR #2657, opened 2023-06-21
Single script to run local setup

k-anshul:


CONTRIBUTING.md
```diff
@@ -16,6 +16,8 @@ This is a full list of development dependencies:
 - [GraalVM](https://www.graalvm.org) and [Maven](https://maven.apache.org) for Java (we recommend installing both through [sdkman](https://sdkman.io))
 - [Buf](https://buf.build) (Protocol Buffers) (on macOS, install with `brew install bufbuild/buf/buf`)
 
+Run `sh scripts/devtool/dev.sh` from project root to run all services locally. Refer `scripts/devtool/README.md` for more details.
```
begelundmuller, line 19: Can we move the script itself to `./scripts/devtool.sh` (use `chmod +x` to make it executable)

scripts/devtool/main.go
```diff
@@ -0,0 +1,221 @@
...
+	waitDBUP()
+
+	c := make(chan os.Signal, 2)
+	signal.Notify(c, os.Interrupt, syscall.SIGTERM)
+	go func() {
+		<-c
+		yellow.Println("stopping services")
+		time.Sleep(2 * time.Second)
+		os.Exit(1)
+	}()
```
begelundmuller, line 106: Instead of waiting 2 seconds, I would suggest waiting indefinitely until the sub-process have been killed (e.g. to prevent leaving hanging Docker containers)

Also, I don't see how this actually sends a cancellation signal to the sub-processes? You could consider using `graceful.WithCancelOnTerminate` and `errgroup.WithContext` for coordinating errors and cancellation instead. (That should also ensure that the whole process exits if any of the sub-processes exit.)

scripts/devtool/main.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+
+	// postgres
+	cmd = exec.Command("docker-compose", "-f", "admin/docker-compose.yml", "up", "--no-recreate")
+	yellow.Println("STARTING POSTGRES")
+	err = cmd.Start()
+	if err != nil {
+		panic("could not start db")
+	}
+	waitDBUP()
```
begelundmuller, line 97: Would it make sense to create a single Docker compose file in the `devtool` folder?

Also, can we add a Redis container to the admin Docker compose? See here for context: https://rilldata.slack.com/archives/C01190F1R1D/p1688398909456289

begelundmuller (changes requested):

scripts/devtool/main.go
k-anshul: Just felt it easy to use since we already have individual docker files. Also makes it easy to pass all kind of flags like remove volumes (There should be same way using docker compose as well although).

cli/cmd/runtime/start.go
```diff
@@ -157,7 +157,7 @@ func StartCmd(cliCfg *config.Config) *cobra.Command {
 			group.Go(func() error { return s.ServeHTTP(cctx, nil) })
 			err = group.Wait()
 			if err != nil {
-				logger.Error("server crashed", zap.Error(err))
+				logger.Fatal("server crashed", zap.Error(err))
```
begelundmuller, line 160: See other note

scripts/devtool/main.go
```diff
@@ -0,0 +1,251 @@
...
+
+	// postgres
+	cmd = exec.CommandContext(ctx, "docker-compose", "-f", "admin/docker-compose.yml", "up", "--no-recreate")
+	yellow.Println("STARTING POSTGRES")
+	err = cmd.Start()
+	if err != nil {
+		panic("could not start db")
+	}
+	waitDBUP(ctx)
+	waitRedis(ctx)
```
begelundmuller, line 102: Should there also be `group.Go` calls waiting for the Docker compose, so that the process doesn't exit until the containers have been gracefully removed?

cli/cmd/admin/start.go
```diff
@@ -229,7 +229,7 @@ func StartCmd(cliCfg *config.Config) *cobra.Command {
 			// Run tasks
 			err = group.Wait()
 			if err != nil {
-				logger.Error("crashed", zap.Error(err))
+				logger.Fatal("crashed", zap.Error(err))
```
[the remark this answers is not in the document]
k-anshul: OOPS. I wanted process to exit with error so that I could get an error and close other processes in case of port mismatches. I can find some other way as well.

scripts/devtool/main.go
k-anshul: Yeah we could add for extra safety although `--no-recreate` flag should ensure that containers are not recreated if its already running.

scripts/devtool/main.go
```diff
@@ -226,12 +219,19 @@ func waitRedis(ctx context.Context) {
...
+		c := redis.NewClient(opts)
+		res, err := c.Echo(ctx, "hello").Result()
+		c.Close()
+		if errors.Is(err, context.Canceled) {
+			return
+		}
+		if err == nil && res == "hello" {
+			green.Println("REDIS STARTED")
+			return
+		}
```
begelundmuller, line 232: Missing a sleep in between retries? Otherwise, it might all happen in a few ms

scripts/devtool/main.go
```diff
@@ -203,16 +194,18 @@ func waitDBUP(ctx context.Context) {
 
 func waitAdmin(ctx context.Context) {
 	for i := 0; i < 10; i++ {
-		cmd := exec.CommandContext(ctx, "go", "run", "cli/main.go", "admin", "ping", "--url", "http://localhost:9090")
-		out, err := cmd.Output()
-		if errors.Is(err, context.Canceled) {
-			return
+		resp, err := http.Get("http://localhost:8080/v1/ping")
```
begelundmuller, line 197: Lint error

scripts/devtool/main.go
```diff
@@ -113,12 +113,8 @@ func main() {
 		}
 		group.Go(func() error {
 			err := admin.Wait()
-			if err != nil {
-				yellow.Println(err)
-			}
-			// nolint:all //required
 			red.Println("ADMIN SERVER STOPPED")
-			return err
+			return multierr.Append(err, context.Canceled)
```
begelundmuller, line 117: Why do you need to add `context.Canceled` manually? It should usually only be returned from `ctx.Err()`, so feels like something underlying might be wrong

begelundmuller (changes requested):

scripts/devtool/main.go
k-anshul: If admin server crashes due to port not available, it doesn't return an error on exit. We want downstream services to exit if admin server crashes. Returning `context.Canceled` seemed like an easy `hack` to stop all services. Changed it to use `context.WithCancel` to send cancellation signal.

begelundmuller (approved):

PR #2658, opened 2023-06-21
Fix pressing back button to default dashboard

### LINKED ISSUE #2536: [Dashboard] Pressing back button in a dashboard fails in certain cases
- Selecting a few filters in a dashboard.
- Click back button until all filters are empty.
- Back and forward do not work anymore.

This is a regression after some recent change. We should add atleast one E2E test for this.

AdityaHegde:
## Checklist
- [x] Manual verification
- [ ] Unit test coverage
- [ ] E2E test coverage
- [x] Needs manual QA?

#### Issue addressed: 
closes #2536

#### Details:
When switching to the default state we were not synching the default url state to the client side state.

## Steps to Verify
- Go to any dashboard.
- Selecting a few filters in a dashboard.
- Click back button until all filters are empty.
- Back and forward should take you through the different filters.


djbarnwal (changes requested):
https://github.com/rilldata/rill/assets/4402679/23556f8b-cd03-436f-acde-3eba75441627

Seeing some weird behavior when going back from the first selected filter. the URL is changing but the filter stays

Also, can we remove the debugging statements.

djbarnwal:
Still seeing some weird edge cases - 

The click on the back button is faint but can be seen in the video. Sometimes clicking doesn't do anything. URL gets changed but the filters remain the same. Also at once two filters got removed. 

https://github.com/rilldata/rill/assets/4402679/30e1673e-9c7f-4d84-baec-7dd1a12ef027

web-common/src/features/dashboards/proto-state/dashboard-url-state.ts
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * Code that looks at dashboard state and url state and decides which one to sync with.
+ */
+export function useDashboardUrlSync(
+  metricViewName: string,
+  metaQuery: CreateQueryResult<V1MetricsView>
+) {
+  const dashboardUrlState = useDashboardUrlState(metricViewName);
+  let desiredProto: string;
```
skokenes, line 52: It's not really the `desiredProto` is it? Its more like the `lastKnownProto`?
Because you end up overwriting it any time its not the same as either the local state or the URL state.

With the current name, it makes it hard for me to understand what happens on loading a dash for first time, where desiredProto === undefined

djbarnwal (approved):

PR #2664, opened 2023-06-21
Return time rollup values in MetricsViewRows

begelundmuller:
Adds a timestamp rollup value as the first column in the result when `time_granularity` is passed to `MetricsViewRows`. The rollup column is named `{time dimension}__{granularity}`.


runtime/queries/metricsview_rows.go
```diff
@@ -153,7 +207,21 @@ func (q *MetricsViewRows) buildMetricsRowsSQL(mv *runtimev1.MetricsView, dialect
...
+
+	if timeRollupColumnName != "" {
+		if mv.TimeDimension == "" || q.TimeGranularity == runtimev1.TimeGrain_TIME_GRAIN_UNSPECIFIED {
+			panic("timeRollupColumnName is set, but time dimension info is missing")
+		}
+
+		rollup := fmt.Sprintf("date_trunc('%s', %s) AS %s", convertToDateTruncSpecifier(q.TimeGranularity), safeName(mv.TimeDimension), safeName(timeRollupColumnName))
+
+		// Prepend the rollup column
+		selectColumns = append([]string{rollup}, selectColumns...)
```
egor-ryashin, line 220: why not just `[]string{rollup, "*"}` ?

begelundmuller: Just in case we need to conditionally inject other columns in the future, someone might miss that there's a hard reset. 

Probably a bit premature abstraction

Issue #2556, opened 2023-06-05
Application stability 


ericpgreen2:
Moved to #2288

ericpgreen2:
Moved to #2288

PR #2676, opened 2023-06-22
Parse and dispaly string logs

djbarnwal:
## Checklist
- [ ] Manual verification
- [ ] Unit test coverage
- [ ] E2E test coverage
- [x] Needs manual QA?

#### Issue addressed: 
Cloud status page doesn’t render logs that aren’t formatted as JSON #2556 

#### Details:
Parses a string log which is not a JSON as an error item.


ericpgreen2 (approved):
Nice 💪

PR #2677, opened 2023-06-22
Run profile queries after fetching columns

### LINKED ISSUE #2662: When changing a model column name, the profiling requests for that column uses the old column name
We are sending multiple calls for both old and new column name. This is because we invalidate the column profile call at the same time as other profiling calls during modelling. The old query with old column name is refetched and once column profile completes the new query with new column name is fetched. We need to add `enabled` to all profiling calls and make sure they run after column profile completes.

AdityaHegde:
## Checklist
- [x] Manual verification
- [ ] Unit test coverage
- [ ] E2E test coverage
- [x] Needs manual QA?

#### Issue addressed: 
closes #2662

## Steps to Verify
1. Create a model from any source.
2. Add `select <column1> from <source>`
3. Replace the `column1` from `column2` by pasting and not typing. This way at a single shot the column is replaced.
4. Doing this back and forth should not add any 400 errors for profiling queries.


web-common/src/runtime-client/invalidation.ts
```diff
@@ -98,11 +98,20 @@ export function invalidationForMetricsViewData(query, metricsViewName: string) {
...
+  name: string,
+  ignoreProfileColumns = false
+) {
+  const queryExtractorMatch = ProfilingQueryExtractor.exec(queryHash);
+  if (!queryExtractorMatch) return false;
+  // TODO: move the query matching to a separate module and reuse in http queue
+  const [, type, table] = queryExtractorMatch;
+  return (
+    table === name && (!ignoreProfileColumns || type !== "columns-profile")
   );
```
ericpgreen2, line 114: This is hard to follow. Could it be refactored to contain `isTableProfileQuery` and `isColumnProfileQuery`, or something along those lines?

web-common/src/runtime-client/invalidation.ts
```diff
@@ -140,8 +149,19 @@ export function invalidateProfilingQueries(
...
+  await queryClient.invalidateQueries({
+    predicate: (query) => {
+      // TODO: move the query matching to a separate module and reuse in http queue
+      const queryExtractorMatch = ProfilingQueryExtractor.exec(query.queryHash);
+      if (!queryExtractorMatch) return false;
+      const [, type, table] = queryExtractorMatch;
+      return table === name && type === "columns-profile";
+    },
+    type: "active",
+  });
```
ericpgreen2, line 161: Similar to `isProfilingQuery`, this is hard to follow. Could one function contain the logic, and you could call it here?

web-common/src/components/column-profile/ColumnProfile.svelte
```diff
@@ -113,6 +113,7 @@
         {hideRight}
         {hideNullPercentage}
         {compact}
+        enableProfiling={!$profileColumns?.isFetching}
```
ericpgreen2, line 116: Would it work with `$profileColumns?.isSuccess` or `$profileColumns?.data`? Either of those would be more semantically clear, IMO.

ericpgreen2:
I think this looks good. Besides my inline comments, a higher-level thought:
- It'd be ideal if we didn't have to do two sets of queries: one for the table columns, one for the column profiles. Do we have plans to make this one API call – something like `GetColumnProfiles(tableName)`?

web-common/src/components/column-profile/ColumnProfile.svelte
AdityaHegde: `enableProfiling` has to be false when we are refetching `profileColumns`. `isSuccess` is true even while the `profileColumns` is being refetched.

web-common/src/runtime-client/query-matcher.ts
```diff
@@ -0,0 +1,54 @@
...
+  ColumnTimeSeries = "timeseries",
+  TableCardinality = "table-cardinality",
+  TableColumns = "columns-profile",
+  TableRows = "rows",
+}
+
+const TopLevelProfilingQuery: Partial<Record<QueryRequestType, boolean>> = {
+  [QueryRequestType.TableColumns]: true,
+  [QueryRequestType.TableRows]: true,
+};
```
ericpgreen2, line 27: Does it make sense to include `TableCardinality` in this list? Then the variable could be `TableProfilingQuery` (not `TopLevelProfilingQuery`). It's not obvious to me why `TableColumns` and `TableRows` are in this unique group together.

ericpgreen2 (approved):
Left one question, else looks good 👍

web-common/src/runtime-client/query-matcher.ts
AdityaHegde: Yes updating

PR #2679, opened 2023-06-22
date formatting fix for query results

### LINKED ISSUE #2663: Runtime: Return `DATE` values as strings instead of timestamps for toplists
See context in [this Slack thread](https://rilldata.slack.com/archives/CTZ8XBQ85/p1686961827784559).

We should be able to make this change in `runtime/queries/metricsview.go:rowsToData`. The `rows` value returned there contains type info for the returned values.

Proposed task list:

- [ ] Update the `rowsToData` function to return result values of type `DATE` as a string in `YYYY-MM-DD` format instead of as a timestamp-formatted string

k-anshul:
closes #2663

### DIFF
--- a/runtime/pkg/pbutil/pbutil.go
+++ b/runtime/pkg/pbutil/pbutil.go
@@ -121,7 +121,7 @@ func ToStruct(v map[string]any, t *runtimev1.StructType) (*structpb.Struct, erro
 	} else {
 		for _, f := range t.Fields {
 			var err error
-			x.Fields[f.Name], err = ToValue(v, f.Type)
+			x.Fields[f.Name], err = ToValue(v[f.Name], f.Type)
 			if err != nil {
 				return nil, err
 			}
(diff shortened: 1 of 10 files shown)


runtime/pkg/pbutil/pbutil.go
```diff
@@ -95,32 +104,48 @@ func ToValue(v any) (*structpb.Value, error) {
...
+			if err != nil {
+				return nil, err
+			}
 		}
-		var err error
-		x.Fields[k], err = ToValue(v)
-		if err != nil {
-			return nil, err
+	} else {
+		for _, f := range t.Fields {
```
k-anshul, line 122: Is it guaranteed that the size of fields will always be equal to size of `v`. I see a check for size in `safeFieldType`

begelundmuller (approved):

runtime/pkg/pbutil/pbutil.go
begelundmuller: It's generally the case, but not if fields have been removed from the schema (like in the time series APIs in this PR). 

The size check in `safeFieldType` is just to be extra safe.

Issue #1916, opened 2023-03-13
UI: Stabilize frontend builds


begelundmuller:
Recommendations from infra for improving the stability of frontend builds:

- [x] Netlify should build web-admin from root using `npm run build -w web-admin` (should already be the case)
- [x] All dependencies not needed in production (i.e. none since it's a SPA build) should be `devDependencies`
- [x] Delete `package-lock.json` and run `npm i` from scratch
- [x] The root `package.json` should not contain any dependencies
- [x] All dependencies used by a workspace should be in its `package.json` (don't rely on transitive imports)
- [ ] Upgrade from Node 16 to LTS – add a method for enforcing a Node and NPM version when building `package-lock.json`
- [ ] Upgrade all package dependencies to latest bugfix release (not necessarily feature releases)
- [ ] Assign an NPM dependency responsible person – ensure proper code reviews for `package-lock.json` changes
- [ ] Fix frontend build warnings
- [ ] Add CI/CD notifications from Netlify for changes that impact `web-admin`
- [ ] Serve static assets in `web-admin` and `web-local` on `/assets` (or similar), not on root
- [ ] Setup build config (Vite?) to address `_redirects` temporary hack in `web-admin/package.json`
- [ ] Address [Dependabot alerts](https://github.com/rilldata/rill-developer/security/dependabot) for NPM

skokenes:
```All dependencies used by a workspace should be in its package.json (don't rely on transitive imports)```
Is there any way we can automatically audit for this one? It caused a bug today when package-lock.json was regenerated from scratch in a PR. We had a subdirectory using a transitive import, and it ended up picking up a new version of a library that broke some behavior

begelundmuller:
@skokenes I don't have any experience here, but a quick search turned up these tools:
- https://github.com/depcheck/depcheck
- https://github.com/dylang/npm-check
Not sure if they work for transitive imports in NPM workspaces, but if they do (or you find another tool), I think we should add a Github Action for it.

begelundmuller:
Closing since the builds have been stable for a long time now

PR #2686, opened 2023-06-22
fix storybook and `package-lock` brittleness by removing `shoelace` dependency

### LINKED ISSUE #2673: dependencies in `package.json` seems to be an inconsistent state with `package-lock.json`
(copied from slack)

ok, I've been thrashing around all afternoon trying to figure out what's going on with the shoelace package and why it doesn't seem to like working with storybook , and after trying a bunch of things I think i've determined that it has nothing to do with storybook, and that we actually have a long standing issue with the set of packages we're trying to use.

What I have just done most recently is

1. pull down a fresh copy of the repo, 
2. check out the commit directly before storybook mergerd
3. immediately delete package-lock.json
4. npm inatall
5. npm run dev

when I do this, I get the following error in the console:

```
Missing "./dist/internal/scroll" specifier in "@shoelace-style/shoelace" package
11:30:59 AM [vite] ✨ new dependencies optimized: @tanstack/svelte-query, yaml, ua-parser-js, @bufbuild/protobuf, crypto-js/md5, uuid, svelte-forms-lib, yup, @codemirror/autocomplete, @codemirror/commands, @codemirror/lang-sql, @codemirror/language, @codemirror/lint, @codemirror/search, @codemirror/state, @codemirror/view, d3-format, d3-array, d3-scale, d3-time-format, @tanstack/svelte-virtual, d3-interpolate-path, d3-scale-chromatic, d3-shape
11:30:59 AM [vite] ✨ optimized dependencies changed. reloading
2023-06-21T11:31:00.000-0700    INFO    Serving Rill on: http://localhost:9009
Missing "./dist/internal/scroll" specifier in "@shoelace-style/shoelace" package
11:31:08 AM [vite] Internal server error: Missing "./dist/internal/scroll" specifier in "@shoelace-style/shoelace" package
  Plugin: vite:import-analysis
  File: /home/bc/rill/rill-developer-burner/rill/web-common/src/components/modal/ModalContainer.svelte
      at e (file:///home/bc/rill/rill-developer-burner/rill/node_modules/vite/dist/node/chunks/dep-e8f070e8.js:16647:25)
      at n (file:///home/bc/rill/rill-developer-burner/rill/node_modules/vite/dist/node/chunks/dep-e8f070e8.js:16647:627)
      at o (file:///home/bc/rill/rill-developer-burner/rill/node_modules/vite/dist/node/chunks/dep-e8f070e8.js:16647:1297)
      at resolveExportsOrImports (file:///home/bc/rill/rill-developer-burner/rill/node_modules/vite/dist/node/chunks/dep-e8f070e8.js:23405:20)
      at resolveDeepImport (file:///home/bc/rill/rill-developer-burner/rill/node_modules/vite/dist/node/chunks/dep-e8f070e8.js:23424:31)
      at tryNodeResolve (file:///home/bc/rill/rill-developer-burner/rill/node_modules/vite/dist/node/chunks/dep-e8f070e8.js:23113:20)
      at Context.resolveId (file:///home/bc/rill/rill-developer-burner/rill/node_modules/vite/dist/node/chunks/dep-e8f070e8.js:22874:28)
      at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
      at async Object.resolveId (file:///home/bc/rill/rill-developer-burner/rill/node_modules/vite/dist/node/chunks/dep-e8f070e8.js:42847:32)
      at async TransformContext.resolve (file:///home/bc/rill/rill-developer-burner/rill/node_modules/vite/dist/node/chunks/dep-e8f070e8.js:42575:23)
```

At this point, I'd be curious to hear from folks who have a deeper understanding of package resolution tangles like this, but also how the shoelace package is being used in particular because:

(a) It is my understanding that if our packages are in a consistent state, then deleting package lock and doing a fresh NPM install should not cause problems. Since our packages seem to not be in a consistent state, I'm not totally sure how to proceed. I can keep digging, but if someone has experience with this they might be able to clean things up more quickly than I can.

(b) shoelace seems to be kind of a special snowflake in that we import it within the on Mount of a component. In `web-common/src/components/modal/ModalContainer.svelte` we have:
```
  onMount(async () => {
    const scroll = await import(
      "@shoelace-style/shoelace/dist/internal/scroll"
    );
    lockBodyScrolling = scroll.lockBodyScrolling;
    unlockBodyScrolling = scroll.unlockBodyScrolling;
    Modal = (await import("@shoelace-style/shoelace/dist/internal/modal"))
      .default;
  });
```

@Hamilton Ulmer, it looks like you wrote this bit above -- do you happen to recall why this is needed? Any ideas as to how it might be interacting with this other set of problems?

bcolloran:
This PR does the following:
- remove shoelace from dependencies and vendor the files needed
- update ModalContainer to import these without doing so `onMount`
- rebuild package-lock from scratch

After this PR, both storybook and `npm run dev` work for me locally.

Additionally this fixes #2673, since it rebuilds `package-lock.json` from scratch with a fresh `npm install`, which should allow us to check off that subtask in #1916. (cc @begelundmuller and @kaspersjo to make that call)


bcolloran:
Note: latest commit also adds a `@ts-ignore` to an unrelated file to suppress an a svelte-check error. My guess is that perhaps the version of svelte-check changed and made the checks more stringent or something like that. I hate adding an ignore, but I think it's worth it in this case -- this PR brings the repo to an overall more reproducible and less brittle state, and the code in question was working fine before  this ts check started failing.

bcolloran:
Wow, I'm now getting a really unbelievable set of `@typescript-eslint` errors: https://github.com/rilldata/rill/actions/runs/5360080684/jobs/9724501187?pr=2686

It looks like it's checking minified and/or build files and somehow losing it's mind all the error are in `/home/runner/work/rill/rill/web-local/build/_app/immutable/chunks/*`, which I'm pretty sure should not be targeted with prettier/eslint/typescript checks.

I think the next step would be to exclude that folder from the failing CI checks.

skokenes:
The reason you are getting eslint errors now is because something you changed is causing the build to produce .js files instead of .mjs files like they were before. eslint is picking up those .js files.
Regardless of final solution, we probably need a .eslintignore file that explicitly ignores build directories.
But also, would be good to understand what you changed that produced .js files instead of .mjs files for the build, and whether that breaks anything downstream for us

bcolloran:
@skokenes I think we're good to go on this now, should be ready for final review and merge

web-common/src/components/modal/scroll.js
```diff
@@ -0,0 +1,58 @@
...
+  }
+}
+export function unlockBodyScrolling(lockingEl) {
+  locks.delete(lockingEl);
+  if (locks.size === 0) {
+    document.body.classList.remove("sl-scroll-lock");
+    document.body.style.removeProperty("--sl-scroll-lock-size");
+  }
+}
+export function scrollIntoView(
```
skokenes, line 25: Is there a reason we use this instead of the browsers native scroll into view capability? https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView

skokenes (approved):

web-common/src/components/modal/scroll.js
bcolloran: Just copied in this file from shoelace, so I don't know if they had a good reason for this approach or not (but in my experience the browser native scroll APIs don't always do exactly what you think they should, and I've had to write wrappers around them in the past, so this doesn't really surprise me).

In either case, digging into it and rewriting+testing it is beyond what I think we should do in this PR, but I could leave a FIXME comment or something if you would like?

skokenes: Nah its cool, I approved despite this

PR #2689, opened 2023-06-23
CLI: Print suggestions when trying to deploy an existing project

begelundmuller:
Proposed updated experience – please give feedback:

Screenshot 2023-06-23 at 11 00 47

### DIFF
--- a/cli/cmd/deploy/deploy.go
+++ b/cli/cmd/deploy/deploy.go
@@ -239,7 +239,7 @@ func DeployCmd(cfg *config.Config) *cobra.Command {
 				warn.Printf("Another project %q already deploys from %q.\n", projects[0], githubURL)
 				warn.Printf("- To force the existing project to rebuild, press 'n' and run `rill project reconcile --reset`\n")
 				warn.Printf("- To delete the existing project, press 'n' and run `rill project delete`\n")
-				warn.Printf("- To deploy the repository as a new project under another name, press enter\n")
+				warn.Printf("- To deploy the repository as a new project under another name, press 'y' or enter\n")
 				if !cmdutil.ConfirmPrompt("Do you want to continue?", "", true) {
 					warn.Println("Aborted")
 					return nil
--- a/cli/cmd/deploy/deploy.go
+++ b/cli/cmd/deploy/deploy.go
@@ -236,8 +236,11 @@ func DeployCmd(cfg *config.Config) *cobra.Command {
 					}
 				}
 
-				warn.Printf("Another project %q already deploys from %q\n", projects[0], githubURL)
-				if !cmdutil.ConfirmPrompt("Do you want to continue", "", true) {
+				warn.Printf("Another project %q already deploys from %q.\n", projects[0], githubURL)
+				warn.Printf("- To force the existing project to rebuild, press 'n' and run `rill project reconcile --reset`\n")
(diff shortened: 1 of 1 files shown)


magorlick:
I know I reported being personally confused during deployment, but I would love to see 1 sentence on the scope of the user problem this change is trying to solve. Is there a PRD I could look at in Notion?

The screen shot has three options when there is a project name collision.
- Why are we asking the user to select 'n' and rerun a CLI command instead of chaining the functionality directly in this prompt for options 1 and 2?
- Why is the prompt at the end asking people to respond (Y, n) and Y doesn't do anything listed above?

What is our mental model around project names? I thought we had an opinion where the project would be the exact same name as the repository. Has that changed?

cli/cmd/deploy/deploy.go
```diff
@@ -236,8 +236,11 @@ func DeployCmd(cfg *config.Config) *cobra.Command {
 					}
 				}
 
-				warn.Printf("Another project %q already deploys from %q\n", projects[0], githubURL)
-				if !cmdutil.ConfirmPrompt("Do you want to continue", "", true) {
+				warn.Printf("Another project %q already deploys from %q.\n", projects[0], githubURL)
+				warn.Printf("- To force the existing project to rebuild, press 'n' and run `rill project reconcile --reset`\n")
+				warn.Printf("- To delete the existing project, press 'n' and run `rill project delete`\n")
+				warn.Printf("- To deploy the repository as a new project under another name, press enter\n")
```
nishantmonu51, line 242: nit: would suggest changing to ? 
```
To deploy the repository as a new project under another name, press 'y' or enter
```

nishantmonu51:
> I know I reported being personally confused during deployment, but I would love to see 1 sentence on the scope of the user problem this change is trying to solve. Is there a PRD I could look at in Notion?

Originally as requested this was a small change for improving current messaging we print on the CLI, and there is no PRD. I would consider this as a small bug fix that doesn't require PRD. 

If you originally or now want to propose a bigger scope of changes here, then I would request you to create a PRD and provide your recommendations there and we would be happy to incorporate them.

> The screen shot has three options when there is a project name collision.
> 
> * Why are we asking the user to select 'n' and rerun a CLI command instead of chaining the functionality directly in this prompt for options 1 and 2?

Each CLI command only does what is in reasonable scope of that CLI command, we intentionally don't chain unrelated command that may cause unexpected behavior e.g. someone running a `rill deploy` CLI  command but would ended up deleting a project in case we chain the options inline. 

> * Why is the prompt at the end asking people to respond (Y, n) and Y doesn't do anything listed above?

The current prompt question is whether we would want the user to continue deploying or abort which takes the boolean values Y/n. 

> What is our mental model around project names? I thought we had an opinion where the project would be the exact same name as the repository. Has that changed?

-  We take the repository name as initial default for the project name, but we do not force a 1:1 mapping there. After deployment a user is allowed to rename the project using `rill project edit`. 
- If you deploy same project second time , currently we prompt the user for a project name as well to avoid clash. Users would redeploy the same project twice in following cases - 
   - deploy from a dev/test branch other than main and get feedback from some consumers
   - deploying from a monorepo hosting multiple projects and using subpaths to deploy a project

nishantmonu51 (approved):

PR #2691, opened 2023-06-23
search and get project support for superusers

rakeshsharma14317:


admin/database/database.go
```diff
@@ -65,6 +65,7 @@ type DB interface {
 	DeleteOrganizationWhitelistedDomain(ctx context.Context, id string) error
 
 	FindProjects(ctx context.Context, afterName string, limit int) ([]*Project, error)
+	FindProjectNamesByPattern(ctx context.Context, namePattern string) ([]*ProjectNames, error)
```
begelundmuller, line 68: Make it `FindProjectPathsByPattern` and return `[]string, error` instead.

The strings will then have the format `org/project`.

This way, we'll have fewer types here and in the proto, which is nice

admin/database/database.go
```diff
@@ -231,6 +232,11 @@ type Project struct {
 	UpdatedOn            time.Time `db:"updated_on"`
 }
 
+type ProjectNames struct {
+	OrgName     string `db:"org_name"`
+	ProjectName string `db:"project_name"`
+}
```
begelundmuller, line 238: Then can remove this type

admin/database/database.go
begelundmuller: Also, it should support pagination like `FindUsersByEmailPattern`

admin/database/postgres/postgres.go
```diff
@@ -210,6 +210,17 @@ func (c *connection) FindProjects(ctx context.Context, afterName string, limit i
 	return res, nil
 }
 
+func (c *connection) FindProjectNamesByPattern(ctx context.Context, namePattern string) ([]*database.ProjectNames, error) {
+	var res []*database.ProjectNames
+	wildcard := "%"
+	pattern := fmt.Sprintf("%s%s%s", wildcard, namePattern, wildcard)
```
begelundmuller, line 216: Remove the wildcard here – it should be provided by the user at the CLI if they want it (then they can do a prefix search if they want by passing `rilldata/%`

admin/database/postgres/postgres.go
```diff
@@ -210,6 +210,17 @@ func (c *connection) FindProjects(ctx context.Context, afterName string, limit i
...
+	err := c.getDB(ctx).SelectContext(ctx, &res, "SELECT o.name as org_name, p.name as project_name FROM projects p JOIN orgs o ON p.org_id = o.id WHERE lower(concat(o.name,'/',p.name)) like lower($1) ORDER BY lower(p.name)", pattern)
```
begelundmuller, line 217: 1. Do `ORDER BY lower(o.name), lower(p.name)`
2. Can use `ILIKE` instead of `lower` on both sides (I don't think we'll get any index matches here anyway)

admin/server/projects.go
```diff
@@ -146,6 +146,54 @@ func (s *Server) GetProject(ctx context.Context, req *adminv1.GetProjectRequest)
...
+
+	dtos := make([]*adminv1.ProjectName, len(projectNames))
+	for i, p := range projectNames {
+		dtos[i] = s.projNamesToDTO(p)
+	}
+
+	return &adminv1.SearchProjectNamesResponse{Projects: dtos}, nil
+}
+
+func (s *Server) SudoGetProject(ctx context.Context, req *adminv1.SudoGetProjectRequest) (*adminv1.SudoGetProjectResponse, error) {
```
begelundmuller, line 175: I don't think this is needed – can't you just call `GetProject`?

proto/rill/admin/v1/api.proto
```diff
@@ -730,6 +756,11 @@ message Project {
   google.protobuf.Timestamp updated_on = 15;
 }
 
+message ProjectName {
+  string org_name = 1;
+  string project_name = 2;
+}
```
begelundmuller, line 762: Given earlier comment, we can also remove this

begelundmuller (changes requested):

admin/database/postgres/postgres.go
```diff
@@ -210,11 +210,12 @@ func (c *connection) FindProjects(ctx context.Context, afterName string, limit i
...
 
-func (c *connection) FindProjectNamesByPattern(ctx context.Context, namePattern string) ([]*database.ProjectNames, error) {
-	var res []*database.ProjectNames
-	wildcard := "%"
-	pattern := fmt.Sprintf("%s%s%s", wildcard, namePattern, wildcard)
-	err := c.getDB(ctx).SelectContext(ctx, &res, "SELECT o.name as org_name, p.name as project_name FROM projects p JOIN orgs o ON p.org_id = o.id WHERE lower(concat(o.name,'/',p.name)) like lower($1) ORDER BY lower(p.name)", pattern)
+func (c *connection) FindProjectPathsByPattern(ctx context.Context, namePattern, afterName string, limit int) ([]string, error) {
+	var res []string
+	err := c.getDB(ctx).SelectContext(ctx, &res, `SELECT concat(o.name,'/',p.name) as project_name FROM projects p JOIN orgs o ON p.org_id = o.id 
+	WHERE concat(o.name,'/',p.name) ilike $1 AND lower(concat(o.name,'/',p.name)) > lower($2)
```
begelundmuller, line 216: Can't use `lower` in this check here, lower strings sort differently from upper strings

cli/cmd/sudo/project/search.go
```diff
@@ -23,37 +26,28 @@ func SearchCmd(cfg *config.Config) *cobra.Command {
...
+			if res.NextPageToken != "" {
+				cmd.Println()
+				cmd.Printf("Next page token: %s\n", res.NextPageToken)
+			}
 
-			cmdutil.TablePrinter(toTable(res.Projects))
 			return nil
 		},
 	}
+	searchCmd.Flags().Uint32Var(&pageSize, "page-size", 50, "Number of users to return per page")
```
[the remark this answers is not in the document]
rakeshsharma14317: ohh typo in cpy paste :)

admin/server/projects.go
```diff
@@ -146,6 +146,41 @@ func (s *Server) GetProject(ctx context.Context, req *adminv1.GetProjectRequest)
 	}, nil
 }
 
+func (s *Server) SearchProjectNames(ctx context.Context, req *adminv1.SearchProjectNamesRequest) (*adminv1.SearchProjectNamesResponse, error) {
+	observability.AddRequestAttributes(ctx,
+		attribute.String("args.org", req.NamePattern),
```
begelundmuller, line 151:
```suggestion
		attribute.String("args.pattern", req.NamePattern),
```

admin/server/projects.go
```diff
@@ -146,6 +146,41 @@ func (s *Server) GetProject(ctx context.Context, req *adminv1.GetProjectRequest)
...
+	if err != nil {
+		return nil, err
+	}
+	pageSize := validPageSize(req.PageSize)
+
+	projectNames, err := s.admin.DB.FindProjectPathsByPattern(ctx, req.NamePattern, token.Val, pageSize)
+	if err != nil {
+		if errors.Is(err, database.ErrNotFound) {
+			return nil, status.Error(codes.NotFound, "projects not found")
+		}
```
begelundmuller, line 169:
```suggestion
```

begelundmuller (approved):

Issue #2666, opened 2023-06-21
Runtime: Set meaningful download filename for exported data


begelundmuller:
We should be able to use a response header to set the name of the downloaded file.

The filename should be `{name}[_filtered]_{current time as YYYYMMDDHHmmSS}.{extension}`, where `name` is the:
- For `MetricsViewXXX` APIs without time/value filters, the name of the underlying model
- For `MetricsViewXXX` APIs with time/value filters, the name of the underlying model with `_filtered` appended to it

PR #2697, opened 2023-06-26
Export filename

egor-ryashin:
Closes https://github.com/rilldata/rill/issues/2666

### DIFF
--- a/runtime/queries/metricsview.go
+++ b/runtime/queries/metricsview.go
@@ -18,7 +18,7 @@ import (
 	"google.golang.org/protobuf/types/known/structpb"
 )
 
-func lookupMetricsView(ctx context.Context, rt *runtime.Runtime, instanceID, name string) (*runtimev1.MetricsView, error) {
+func LookupMetricsView(ctx context.Context, rt *runtime.Runtime, instanceID, name string) (*runtimev1.MetricsView, error) {
 	obj, err := rt.GetCatalogEntry(ctx, instanceID, name)
 	if err != nil {
 		return nil, status.Error(codes.InvalidArgument, err.Error())
--- a/runtime/queries/metricsview.go
+++ b/runtime/queries/metricsview.go
@@ -18,7 +18,7 @@ import (
 	"google.golang.org/protobuf/types/known/structpb"
 )
 
-func LookupMetricsView(ctx context.Context, rt *runtime.Runtime, instanceID, name string) (*runtimev1.MetricsView, error) {
+func lookupMetricsView(ctx context.Context, rt *runtime.Runtime, instanceID, name string) (*runtimev1.MetricsView, error) {
 	obj, err := rt.GetCatalogEntry(ctx, instanceID, name)
 	if err != nil {
 		return nil, status.Error(codes.InvalidArgument, err.Error())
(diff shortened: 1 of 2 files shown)


runtime/server/downloads.go
```diff
@@ -80,8 +83,10 @@ func (s *Server) downloadHandler(w http.ResponseWriter, req *http.Request) {
 	switch request.Format {
 	case runtimev1.ExportFormat_EXPORT_FORMAT_CSV:
 		w.Header().Set("Content-Type", "text/csv")
+		w.Header().Set("Content-Disposition", fmt.Sprintf("attachment; filename=\"%s.csv\"", metricsViewName))
 	case runtimev1.ExportFormat_EXPORT_FORMAT_XLSX:
 		w.Header().Set("Content-Type", "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet")
+		w.Header().Set("Content-Disposition", fmt.Sprintf("attachment; filename=\"%s.xlsx\"", metricsViewName))
```
begelundmuller, line 89: What if the name has `"` in it?

begelundmuller (changes requested):
Please also update to reflect the filenames described here: https://github.com/rilldata/rill/issues/2666

runtime/queries/metricsview.go
```diff
@@ -18,7 +18,7 @@ import (
 	"google.golang.org/protobuf/types/known/structpb"
 )
 
-func lookupMetricsView(ctx context.Context, rt *runtime.Runtime, instanceID, name string) (*runtimev1.MetricsView, error) {
+func LookupMetricsView(ctx context.Context, rt *runtime.Runtime, instanceID, name string) (*runtimev1.MetricsView, error) {
```
begelundmuller, line 21: Instead of exporting this util func, use `rt.GetCatalogEntry` directly to reduce coupling

runtime/server/downloads.go
```diff
@@ -55,12 +57,25 @@ func (s *Server) downloadHandler(w http.ResponseWriter, req *http.Request) {
...
 		q, err = createToplistQuery(req.Context(), w, v.MetricsViewToplistRequest, request.Format)
 	case *runtimev1.ExportRequest_MetricsViewRowsRequest:
 		v.MetricsViewRowsRequest.Limit = request.Limit
+		metricsViewName = v.MetricsViewRowsRequest.MetricsViewName
+		filter = v.MetricsViewRowsRequest.Filter
+		if v.MetricsViewRowsRequest.TimeStart != nil || v.MetricsViewRowsRequest.TimeEnd != nil {
+			timeRange = true
+		}
 		q, err = createRowsQuery(req.Context(), w, v.MetricsViewRowsRequest, request.Format)
 	default:
```
begelundmuller, line 80: This will break if we add `case *runtimev1.ExportRequest_TableRows` for example (which we should do at some point, so we can remove the old Parquet export function used by the modeler) – can we move the query-specific checks into the cases and just have `var filename string` or something as output?

Alternatively, we could consider adding an `ExportFilename` function to the `Query` interface?

begelundmuller (changes requested):
Additionally, please look at the test/CI failures

begelundmuller (approved):

PR #2704, opened 2023-06-27
feat: app business model context

skokenes:
## Checklist
- [ ] Manual verification
- [ ] Unit test coverage
- [ ] E2E test coverage
- [ ] Needs manual QA?

### DIFF
--- a/web-common/src/features/dashboards/business-model/business-model.ts
+++ b/web-common/src/features/dashboards/business-model/business-model.ts
@@ -1,0 +1,50 @@
+import { writable, get, Writable, Readable, derived } from "svelte/store";
+import { getContext } from "svelte";
+import type { QueryClient } from "@tanstack/svelte-query";
+import type { Runtime } from "@rilldata/web-common/runtime-client/runtime-store";
+import {
+  MetricsExplorerEntity,
+  MetricsExplorerStoreType,
+  metricsExplorerStore,
+  useDashboardStore,
+} from "../dashboard-stores";
+import { runtime } from "@rilldata/web-common/runtime-client/runtime-store";
+
+type BusinessModel = {
+  runtime: Writable<Runtime>;
+  metricsViewName: Writable<string>;
+  metricsStore: Readable<MetricsExplorerStoreType>;
+  dashboardStore: Readable<MetricsExplorerEntity>;
+  queryClient: QueryClient;
+  setMetricsViewName: (s: string) => void;
+};
+
+export const DEFAULT_STORE_KEY = Symbol("business-model");
+
+export function getBusinessModel(): BusinessModel {
+  return getContext(DEFAULT_STORE_KEY);
+}
+
+export function createBusinessModel({
+  queryClient,
+  metricsViewName,
+}: {
+  queryClient: QueryClient;
+  metricsViewName: string;
+}): BusinessModel {
+  const metricsViewNameStore = writable(metricsViewName);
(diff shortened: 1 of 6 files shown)


ericpgreen2 (commented):
I like where this is going. A few thoughts:
1. The "business model" seems to encapsulate the client-state managers (the Svelte Stores) and the server-state managers (the TanStack client). The one piece of state it doesn't seem to manage is the URL state, which is currently handled by `<DashboardStateProvider/>` (which has an imprecise name IMO). If we want the business model to handle all state management, should we include a URL State Manager as well, and remove the `<DashboardStateProvider />` layer? Or would that not make sense when we have multiple runtimes on the same page?

2. Related to 1, how do you envision the `metricsViewName` syncing with the URL?

3. The "business model" name isn't so intuitive to me. Thoughts on `<StateManagersProvider/>`, `const stateManagers = getStateManagers()`? Or if it's just scoped to the dashboard, `<DashboardStateManagersProvider />`?

skokenes:
1. I've avoided touching the state URL syncing piece yet because @AdityaHegde has been refactoring that in a different branch. But yes, the goal would be to merge that into one provider (IMO syncing should be part of the business model, either as a plugin or just the way Aditya already has it) so we will only have 1 provider. I didn't do that here because I didn't want to conflict with Aditya's work
2. Aditya's sync manager work handles this well already, I believe
3. Open to other names, idk that its just state though. Also don't think we need to bikeshed on what we name something

ericpgreen2:
> 1. I've avoided touching the state URL syncing piece yet because @AdityaHegde has been refactoring that in a different branch. But yes, the goal would be to merge that into one provider (IMO syncing should be part of the business model, either as a plugin or just the way Aditya already has it) so we will only have 1 provider. I didn't do that here because I didn't want to conflict with Aditya's work
> 2. Aditya's sync manager work handles this well already, I believe
> 3. Open to other names, idk that its just state though. Also don't think we need to bikeshed on what we name something

1 & 2. Cool, sounds good.
3. What does it include besides state management? I need to mentally translate "business model" into something else, and we'll be seeing this name everywhere so I do think it should be clear.

skokenes:
There may be other side effect stuff for the cloud unrelated to state management, like making an API call to deploy something or create an alert or scheduled report.

Still, if you guys are most comfortable calling it StateManager, that's fine with me. Let's just come to a quick consensus on what it is

web-common/src/features/dashboards/business-model/business-model.ts
```diff
@@ -0,0 +1,50 @@
...
+  });
+  return {
+    runtime: runtime,
+    metricsViewName: metricsViewNameStore,
+    metricsStore: metricsExplorerStore,
+    queryClient,
+    dashboardStore,
+    setMetricsViewName: (name) => {
+      metricsViewNameStore.set(name);
+    },
```
AdityaHegde, line 48: Would it make sense to add additional "derived" stores here? That way multiple components using the derived store will get a single copy and we wont need to implement our own caching by `metricsViewName`.

AdityaHegde (approved):
Looks good. Perhaps makes sense to get this in before we all start working on individual refactors.

djbarnwal (approved):

web-common/src/features/dashboards/business-model/business-model.ts
skokenes: It might! Let's discuss as a team how we want to manage derived stores and sharing them

PR #2710, opened 2023-06-28
Deprecate app store in favor of derived store

djbarnwal:
## Checklist
- [x] Manual verification
- [ ] Unit test coverage
- [ ] E2E test coverage
- [x] Needs manual QA?

#### Issue addressed: 
Uses `page` store from sveltekit to derive active entity
Adds missing navigation events

### DIFF
--- a/web-common/src/features/models/workspace/inspector/EmbeddedSourceReference.svelte
+++ b/web-common/src/features/models/workspace/inspector/EmbeddedSourceReference.svelte
@@ -7,7 +7,6 @@
   import * as classes from "@rilldata/web-local/lib/util/component-classes";
   import { getContext } from "svelte";
   import WithModelResultTooltip from "./WithModelResultTooltip.svelte";
-  import { goto } from "$app/navigation";
   import { behaviourEvent } from "@rilldata/web-common/metrics/initMetrics";
   import {
     MetricsEventScreenName,
@@ -33,8 +32,7 @@
     queryHighlight.set(undefined);
   }
 
-  function navigateToSource() {
-    goto(`/source/${entry.name}`);
+  function emitNavigationToSourceTelemetry() {
     behaviourEvent.fireNavigationEvent(
       entry.name,
       BehaviourEventMedium.Menu,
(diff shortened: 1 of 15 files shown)


web-common/src/layout/app-store.ts
```diff
@@ -14,54 +17,86 @@ export interface ActiveEntity {
...
+        type: MetricsEventScreenName.Source,
+      };
+      break;
+    case "/(application)/model/[name]":
+      activeEntity = {
+        name: $page?.params?.name,
+        type: MetricsEventScreenName.Model,
+      };
+      break;
+    case "/(application)/dashboard/[name]":
```
AdityaHegde, line 46: Add cloud routes as well.

web-common/src/features/models/workspace/inspector/EmbeddedSourceReference.svelte
```diff
@@ -25,17 +32,31 @@
...
+      MetricsEventScreenName.Model,
+      MetricsEventScreenName.Source
+    );
+  }
 </script>
 
 <WithModelResultTooltip {modelHasError}>
-  <a
-    href="/source/{entry.name}"
+  <div
```
AdityaHegde, line 49: Perhaps leave the a tag with href as is. Just add the firing event handler.

web-common/src/layout/app-store.ts
```diff
@@ -14,54 +17,86 @@ export interface ActiveEntity {
...
+  }
+
+  appStore.setActiveEntity(
+    activeEntity.name,
+    ScreenToEntityMap[activeEntity.type]
+  );
+  return activeEntity;
+});
+
+// TODO: Do we still need this?
```
AdityaHegde, line 76: You can add the previous entity name to appScreen and get rid of this. Call `httpRequestQueue.inactiveByName` directly in the body of `appScreen`.

AdityaHegde (changes requested):

web-common/src/layout/app-store.ts
djbarnwal: I want to keep `appScreen` clean as it's being heavily utilized across the application for telemetry. Better it just contains the current active entity. Will leave this as is for now.

web-common/src/layout/app-store.ts
```diff
@@ -14,54 +17,91 @@ export interface ActiveEntity {
...
+// App store is being utilized for making previous entity inactive in the HTTP request queue
+const { update } = writable({
   activeEntity: undefined,
   previousActiveEntity: undefined,
 } as AppStore);
 
 const appStoreReducers = {
   setActiveEntity(name: string, type: EntityType) {
     update((state) => {
+      state.previousActiveEntity = state.activeEntity;
```
AdityaHegde, line 91: Why is this moved to the top? Wouldnt this make the current entity inactive? Either move this to the bottom or move the call to `setActiveEntity` to the top in `appScreen` body.

AdityaHegde (approved):
Just address the small comment. Good otherwise.

web-common/src/layout/app-store.ts
djbarnwal: The previous implementation had a bug. Corrected it with these changes. You can verify by console logging `state.previousActiveEntity.name` when it's being made inactive. 

Now when you move from asset A to asset B, asset A is being made inactive now.

`name` provided through props is the current active entity being set. `state.activeEntity` just before being updated is now the previous active entity which has to be made inactive.

Issue #2633, opened 2023-06-16
Runtime: Implement validation so we never run SQL queries with empty select lists


begelundmuller:
Due to a frontend error, requests were made without specifying any parameters, and we saw some errors like `Parser Error: SELECT clause without selection list`.

We should add validation before running the query that prevents this from happening (e.g. using proto validation annotations or `if` checks in the handlers.)

begelundmuller:
Closed by https://github.com/rilldata/rill/pull/2712

PR #2712, opened 2023-06-28
Additional queries validation

egor-ryashin:
see https://github.com/rilldata/rill/issues/2633


proto/rill/runtime/v1/queries.proto
```diff
@@ -557,17 +557,17 @@ message ColumnCardinalityResponse {
...
-  string table_name = 2;
-  repeated BasicMeasure measures = 3;
-  string timestamp_column_name = 4;
+  string table_name = 2 [(validate.rules).string.min_len = 1];
+  repeated BasicMeasure measures = 3 [(validate.rules).repeated.min_items = 1];
+  string timestamp_column_name = 4 [(validate.rules).string.min_len = 1];
   TimeSeriesTimeRange time_range = 5;
-  int32 pixels = 7;
-  int32 sample_size = 8;
+  int32 pixels = 7 [(validate.rules).int32.gt = 0];
```
begelundmuller, line 564: Doesn't the API support 0 for this?

runtime/server/queries_metrics.go
```diff
@@ -224,3 +244,10 @@ func validateInlineMeasures(ms []*runtimev1.InlineMeasure) error {
 	}
 	return nil
 }
+
+func validateMeasures(ms []*runtimev1.InlineMeasure, measures []string) error {
+	if len(ms)+len(measures) == 0 {
+		return fmt.Errorf("no measures specified")
+	}
+	return nil
+}
```
begelundmuller, line 253: I think you can just use proto validations – only the `measures` slice defines what measures are returned, the inline measures I believe just inject extra definitions

begelundmuller (changes requested):

proto/rill/runtime/v1/queries.proto
egor-ryashin: yes, good catch

begelundmuller (changes requested):
1. Needs to re-generate the proto files for changes to take effect
2. Please do `make cli` and do an experience pass on `./rill start` with a simple project?Just to make sure the UI doesn’t depend on making funny queries

begelundmuller (approved):

PR #2716, opened 2023-06-29
Adding batching API

AdityaHegde:
This adds an API to batch profiling and metrics view queries. Using streaming response to send data in chunks.

- [x] Add query endpoint to proto with request and responses.
- [x] Add the implementation to forward the query to the correct method.
- [x] Add tests.
- [ ] ~~Disable in cloud (?)~~ Not needed


runtime/server/batch_query.go
```diff
@@ -0,0 +1,177 @@
...
+	"sync"
+
+	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
+	"github.com/rilldata/rill/runtime/server/auth"
+)
+
+func (s *Server) QueryBatch(req *runtimev1.QueryBatchRequest, srv runtimev1.QueryService_QueryBatchServer) error {
+	ctx := srv.Context()
+
+	if !auth.GetClaims(ctx).CanInstance(req.InstanceId, auth.ReadOLAP) {
```
begelundmuller, line 15: `ReadOLAP` actually gives access to run any SQL query, so it's not available in cloud. We would need to check for `ReadMetrics` (for metrics view APIs) and `ReadProfiling` (for modeler APIs) individually for each query. (I guess this already happens?)

runtime/server/batch_query.go
```diff
@@ -0,0 +1,177 @@
...
+		return ErrForbidden
+	}
+	var wg sync.WaitGroup
+
+	for _, query := range req.Queries {
+		wg.Add(1)
+		go func(query *runtimev1.QueryBatchSingleRequest) {
```
begelundmuller, line 22: I feel like I more often see `query := query` on the line before (instead of passing it as an arg), but both should work

runtime/server/batch_query.go
```diff
@@ -0,0 +1,177 @@
...
+			defer wg.Done()
+
+			resp := s.forwardQuery(ctx, query)
+
+			if err := srv.Send(resp); err != nil {
+				s.logger.Debug(fmt.Sprintf("Profiling Query Error: %v", err))
+			}
```
begelundmuller, line 29: Are there any cases where `Send` will error where we shouldn't stop processing queries and terminate the batch? If not, consider using `errgroup` instead

runtime/server/batch_query.go
```diff
@@ -0,0 +1,177 @@
...
+
+func (s *Server) forwardQuery(ctx context.Context, query *runtimev1.QueryBatchSingleRequest) *runtimev1.QueryBatchResponse {
+	resp := &runtimev1.QueryBatchResponse{
+		Id:   query.Id,
+		Type: query.Type,
+	}
+
+	var err error
+	switch query.Type {
+	case runtimev1.QueryBatchType_QUERY_BATCH_TYPE_METRICS_VIEW_TOPLIST:
```
begelundmuller, line 45: Consider a type switch instead, so the enum isn't needed

begelundmuller: In all the cases, override the individual query's `instanceId` and `priority` to the value from the batched query. It paves the way for removing those fields from the individual queries. The repetition is okay for now

begelundmuller (changes requested):

runtime/server/batch_query.go
```diff
@@ -0,0 +1,177 @@
...
+
+			resp := s.forwardQuery(ctx, query)
+
+			if err := srv.Send(resp); err != nil {
+				s.logger.Debug(fmt.Sprintf("Profiling Query Error: %v", err))
+			}
+		}(query)
+	}
+
+	wg.Wait()
```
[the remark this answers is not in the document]
AdityaHegde: Sounds good. Lemme create a follow up once we have this implemented in UI as well.

runtime/server/batch_query.go
AdityaHegde: Yes. Have added a TODO to check for access upfront.

runtime/server/batch_query.go
AdityaHegde: Have updated to override `instanceId`. For `priority` does it make sense to only override if the priority is not passed?

runtime/server/batch_query.go
```diff
@@ -0,0 +1,251 @@
...
+		Key: queryEntry.Key,
+	}
+
+	var err error
+	switch typedQueryEntry := queryEntry.Query.(type) {
+	case *runtimev1.QueryBatchEntry_MetricsViewToplistRequest:
+		var r *runtimev1.MetricsViewToplistResponse
+		if typedQueryEntry.MetricsViewToplistRequest.Priority == 0 {
+			typedQueryEntry.MetricsViewToplistRequest.Priority = query.Priority
+		}
```
begelundmuller, line 51: See earlier comment about moving priority to the entry. If the frontend doesn't need the conditional, I think it's okay to just always use the value from the entry

begelundmuller (changes requested):

runtime/server/batch_query.go
AdityaHegde: Ya i will remove these. Frontend can send the same priority to all requests if needed. Let that control be on the sender.

PR #2725, opened 2023-06-30
Using the correct field to show toplist delta

AdityaHegde:
## Checklist
- [x] Manual verification
- [ ] Unit test coverage
- [ ] E2E test coverage
- [x] Needs manual QA?

#### Issue addressed: 
We were using `dimensionName` to show toplist delta values in dimension detail table.

#### Details:
Using `dimensionColumn` to show the delta values. The values should now match the toplist from full dashboard.

## Steps to Verify
1. Go to a dashboard with comparison.
2. Compare delta values for a dimension in full dashboard and detail table. They should now match.

### DIFF
--- a/web-common/src/features/dashboards/dimension-table/DimensionDisplay.svelte
+++ b/web-common/src/features/dashboards/dimension-table/DimensionDisplay.svelte
@@ -335,6 +335,7 @@
       $comparisonTopListQuery?.data,
       values,
       dimensionName,
+      dimensionColumn,
       leaderboardMeasureName
     );
   }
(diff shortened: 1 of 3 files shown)


ericpgreen2 (approved):
Good catch. Confirming once more – any other places where we need `dimensionColumn`? And is this backwards-compatible with the old dimension keys? Feel free to merge when you want.

AdityaHegde:
I couldnt find any more instance needing to use `dimensionColumn`. `dimensionColumn` is filled with `name` if `column` is null. So it is backwards compatible.

PR #2667, opened 2023-06-21
feat: export button on dashboard

skokenes:
## Checklist
- [x] Manual verification
- [ ] Unit test coverage
- [x] E2E test coverage
- [ ] Needs manual QA?

#### Issue addressed: 
[Export filtered dashboard data to CSV or other file type](https://www.notion.so/rilldata/UI-to-support-export-for-the-filtered-OBT-a3ac1e4dfafd435fbcfb06716cab27f5?pvs=4)

#### Details:
This feature is local only. When looking at a dashboard, a user can press an export button on the row viewer to download the filtered rows in either CSV or XLSX format. The Export icon was updated per @jkhwu's feedback.

This PR also fixes a minor bug that I was hitting with the date picker.

## Steps to Verify
Open a dashboard, filter some data, press the Export button in the bottom right corner.

PR #2735, opened 2023-07-03
fixes the e2e dashboard test where the number doesn't match the dash

hamilton:
`dashboards.spec.ts` in our e2e tests are actually failing. It appears tht the number after setting a custom range has not been updated to potentially match the actual value; either that, or the test is showing that our custom time range logic is wrong. I think whoever worked on that feature should check this test.

If the PR is right, and the number needed to be updated, we can merge this. If the PR is wrong & in fact our implementation is wrong, that too should be addressed and this PR should be closed.

## Steps to Verify
1. run the e2e tests. They should pass.


skokenes:
Should be fixed by https://github.com/rilldata/rill/pull/2667

PR #2739, opened 2023-07-04
Motherduck connector

### LINKED ISSUE #2734: Runtime: Implement Motherduck source connector

k-anshul:
closes #2734 

Example source:

```
type: motherduck
query: 'select * from bids_data'
```

TODOs:

- [x] UI support for motherduck connector/hide motherduck tab from UI
- [ ] Disable browser open on md_connect 
- [ ] Connect to md in ready only mode.


runtime/connectors/motherduck/motherduck.go
```diff
@@ -0,0 +1,49 @@
...
+	DisplayName: "Motherduck",
+	Description: "Import data from Motherduck.",
+	Properties: []connectors.PropertySchema{
+		{
+			Key:         "query",
+			Type:        connectors.StringPropertyType,
+			Required:    true,
+			DisplayName: "Query",
+			Description: "Query to extract data from Motherduck.",
+			Placeholder: "select * from my_db.my_table;", // todo :: check this
```
begelundmuller, line 25: Could it be problematic if people try to reference the DB name in the query? Vs. just showing something like `SELECT * FROM my_table`

runtime/drivers/duckdb/connectors.go
```diff
@@ -118,6 +121,86 @@ func (c *connection) ingest(ctx context.Context, env *connectors.Env, source *co
...
+		defer res.Close()
+
+		res.Next()
+		var localDB string
+		if err := res.Scan(&localDB); err != nil {
+			return c.checkErr(err)
+		}
+
+		// get token
+		token := env.Variables["TOKEN"]
```
begelundmuller, line 139: Is this lookup case insensitive?

runtime/drivers/duckdb/connectors.go
```diff
@@ -118,6 +121,86 @@ func (c *connection) ingest(ctx context.Context, env *connectors.Env, source *co
...
+		if token == "" && env.AllowHostAccess {
+			token = os.Getenv("motherduck_token")
+		}
+		// load motherduck extension; connect to motherduck service
+		err = c.Exec(ctx, &drivers.Statement{Query: "LOAD 'motherduck';", Priority: 1})
```
begelundmuller, line 144: Does this need to be `INSTALL 'motherduck'; LOAD 'motherduck';`?

begelundmuller: Also, you can omit `Priority` from all these queries – when using `WithConnection`, a conn is already acquired, so priority doesn't matter

runtime/drivers/duckdb/connectors.go
```diff
@@ -118,6 +121,86 @@ func (c *connection) ingest(ctx context.Context, env *connectors.Env, source *co
...
+		for res.Next() {
+			var name string
+			if res.Scan(&name) != nil {
+				return c.checkErr(err)
+			}
+
+			names = append(names, name)
+		}
+		if len(names) == 1 { // single motherduck db, use db to allow user to run query without specifying db name
+			err = c.Exec(ctx, &drivers.Statement{Query: fmt.Sprintf("USE %q;", names[0]), Priority: 1})
```
begelundmuller, line 171: Does it not support using `?` and passing the name as an arg?

I think using %q might break if the DB has a `"` in it's name (since SQL escapes quotes using double quotes instead of backslashes, IIRC). It's very unlikely, though, so otherwise this is also fine

runtime/drivers/duckdb/connectors.go
```diff
@@ -118,6 +121,86 @@ func (c *connection) ingest(ctx context.Context, env *connectors.Env, source *co
...
+				if err != nil {
+					err = c.checkErr(err)
+					c.logger.Error("failed to switch to local database", zap.Error(err))
+				}
+			}(ensuredCtx)
+		}
+
+		qryProp, ok := source.Properties["query"]
+		if !ok {
+			return fmt.Errorf("query field is mandatory for `Motherduck` source")
```
begelundmuller, line 187: nit: `property "query" is mandatory for connector "motherduck"`

runtime/drivers/duckdb/connectors.go
```diff
@@ -118,6 +121,86 @@ func (c *connection) ingest(ctx context.Context, env *connectors.Env, source *co
...
+		names := make([]string, 0)
+		for res.Next() {
+			var name string
+			if res.Scan(&name) != nil {
+				return c.checkErr(err)
+			}
+
+			names = append(names, name)
+		}
+		if len(names) == 1 { // single motherduck db, use db to allow user to run query without specifying db name
```
begelundmuller, line 170: Should we consider allowing people to pass `use: name` in the YAML for cases where there are multiple DBs?

runtime/drivers/duckdb/connectors.go
```diff
@@ -118,6 +121,86 @@ func (c *connection) ingest(ctx context.Context, env *connectors.Env, source *co
...
+		}
+
+		qryProp, ok := source.Properties["query"]
+		if !ok {
+			return fmt.Errorf("query field is mandatory for `Motherduck` source")
+		}
+
+		userQuery := strings.TrimSpace(qryProp.(string))
+		userQuery, _ = strings.CutSuffix(userQuery, ";") // trim trailing semi colon
+		query := fmt.Sprintf("CREATE OR REPLACE TABLE %s.%s AS (%s);", localDB, source.Name, userQuery)
```
begelundmuller, line 192: Best to use quotes around the names `"mydb"."mytable"` to be safe (in case we allow spaces in the source name in the future)

runtime/drivers/duckdb/connectors.go
```diff
@@ -118,6 +121,86 @@ func (c *connection) ingest(ctx context.Context, env *connectors.Env, source *co
 	return summary, nil
 }
 
+func (c *connection) ingestMotherduckData(ctx context.Context, source *connectors.Source, env *connectors.Env) (*drivers.IngestionSummary, error) {
+	err := c.WithConnection(ctx, 1, func(ctx, ensuredCtx context.Context) error {
+		res, err := c.Execute(ctx, &drivers.Statement{Query: "SELECT current_database();"})
+		if err != nil {
+			return c.checkErr(err)
```
begelundmuller, line 128: Since this only calls `Execute` and `Exec`, which already calls `c.checkErr`, you can just return errors directly from this func

begelundmuller (changes requested):

runtime/drivers/duckdb/connectors.go
k-anshul: In a sense yes. Before setting this we convert every variable to uppercase.

runtime/drivers/duckdb/connectors.go
k-anshul: > Does this need to be INSTALL 'motherduck'; LOAD 'motherduck';?

Yes it worked for me since I already had installed `motherduck`. Wouldn't work on new machine. Will make the change.

runtime/connectors/motherduck/motherduck.go
k-anshul: I felt we should encourage people to use db name with query so added db name in placeholder query. Shouldn't be a problem if db name is used, I verified by running query with and without db name.

runtime/drivers/duckdb/connectors.go
k-anshul: Yeah can be done. Should we name the field `db` instead. 
If there are multiple db we can parse error to give nice instructions to use this field for table not found kind of errors.

runtime/drivers/duckdb/connectors.go
k-anshul: Yeah seems like we can't use `?` with use statments.

runtime/drivers/duckdb/connectors.go
k-anshul: Although duckdb's error is self explanatory as well. 
```
/sources/md.yaml: Catalog Error: Table with name bids_data does not exist!
Did you mean "my_db.bids_data"?
```

begelundmuller: Both `use` or `db` or `database` seem fine to me, you decide

DuckDB's error seems adequate to me

runtime/drivers/duckdb/connectors.go
begelundmuller: Okay, take a look at `safeName` in `runtime/queries/sqlutil.go`. Might make sense to copy it into this package to get safe names

runtime/connectors/motherduck/motherduck.go
begelundmuller: Okay, that makes sense. Let's keep it

nishantmonu51 (approved):

begelundmuller (approved):

PR #2741, opened 2023-07-04
check if provided sources exists for reconcile refresh sources

rakeshsharma14317:


admin/projects.go
```diff
@@ -331,6 +332,32 @@ func (s *Service) triggerReconcile(ctx context.Context, depl *database.Deploymen
 
 // TriggerRefreshSource triggers refresh of a deployment's sources. If the sources slice is nil, it will refresh all sources.f
 func (s *Service) TriggerRefreshSources(ctx context.Context, depl *database.Deployment, sources []string) error {
+	// check if provided sources are exists in catalog
+	if len(sources) > 0 {
+		rt, err := s.openRuntimeClientForDeployment(depl)
+		if err != nil {
+			return s.endReconcile(ctx, depl, nil, err)
```
begelundmuller, line 339: Return `err`, don't call `endReconcile` in this func

admin/projects.go
```diff
@@ -331,6 +332,32 @@ func (s *Service) triggerReconcile(ctx context.Context, depl *database.Deploymen
...
+		}
+		defer rt.Close()
+
+		// Get paths of sources
+		res, err := rt.ListCatalogEntries(ctx, &runtimev1.ListCatalogEntriesRequest{InstanceId: depl.RuntimeInstanceID, Type: runtimev1.ObjectType_OBJECT_TYPE_SOURCE})
+		if err != nil {
+			return s.endReconcile(ctx, depl, nil, err)
```
begelundmuller, line 346: Remove call to `endReconcile`

admin/projects.go
```diff
@@ -331,6 +332,32 @@ func (s *Service) triggerReconcile(ctx context.Context, depl *database.Deploymen
...
+		catalogSources := []string{}
+		for _, entry := range res.Entries {
+			catalogSources = append(catalogSources, entry.Name)
+		}
+
+		for _, source := range sources {
+			if !slices.Contains(catalogSources, source) {
+				return fmt.Errorf("source %q is not found", source)
+			}
+		}
```
begelundmuller, line 358:
```suggestion
		for _, source := range sources {
			found := false
			for _, entry := range res.Entries {
				if strings.EqualFold(source, entry.Name) {
					found = true
					break
				}
			}
			if !found {
				return fmt.Errorf("source %q not found", source)
			}
		}
```

begelundmuller (approved):

begelundmuller (changes requested):
See comments above

admin/projects.go
rakeshsharma14317: ohh, copy pst :)

begelundmuller (approved):

PR #2742, opened 2023-07-05
Implement DB size estimation for DuckDB

rakeshsharma14317:


runtime/drivers/duckdb/connectors.go
```diff
@@ -43,6 +43,20 @@ func (c *connection) Ingest(ctx context.Context, env *connectors.Env, source *co
...
+	var paths []string
+	path := c.config.DSN
+	if c.config.DSN == "" {
+		return 0, false
+	}
+
+	// Add .wal file path (e.g final size will be sum of *.db and *.db.wal)
+	dbWalPath := fmt.Sprintf("%s.wal", path)
+	paths = append(paths, path, dbWalPath)
+	c.logger.Info("duckdb file paths", zap.Strings("paths", paths), zap.Int64("size", fileSize(paths)))
```
begelundmuller, line 56: Remove log

runtime/drivers/duckdb/connectors.go
```diff
@@ -43,6 +43,20 @@ func (c *connection) Ingest(ctx context.Context, env *connectors.Env, source *co
 	return summary, err
 }
 
+func (c *connection) EstimateSize() (int64, bool) {
+	var paths []string
+	path := c.config.DSN
+	if c.config.DSN == "" {
+		return 0, false
+	}
```
begelundmuller, line 51: If it's in-memory (no path), return `0, true`

runtime/drivers/duckdb/connectors.go
```diff
@@ -43,6 +43,20 @@ func (c *connection) Ingest(ctx context.Context, env *connectors.Env, source *co
 	return summary, err
 }
 
+func (c *connection) EstimateSize() (int64, bool) {
+	var paths []string
+	path := c.config.DSN
```
begelundmuller, line 48: The DSN may have options, like `xxx?opt1=val1`, so this will not work reliably. Need to use `c.config.DBFilePath`

begelundmuller (changes requested):

begelundmuller (approved):

PR #2745, opened 2023-07-06
Duckdb sql parser

### LINKED ISSUE #2732: Runtime: Implement DuckDB SQL utils
We need some DuckDB SQL utils for async reconcile. It would be nice to leverage the new DuckDB SQL to JSON parser. This PR contains proposed stubs for the library: https://github.com/rilldata/rill/pull/2731

Proposed task list:

- [ ] Support extracting/rewriting table references (for SQL sources)
- [ ] Support sanitizing a query (to not re-create models when only a comment is changed)
- [ ] Support extracting comment annotations (for `@materialize` syntax)
- [ ] Support extracting column references (for generating a metrics view)

AdityaHegde:
closes #2732

Fills in the APIs that will be used by the new reconcile. Adds an AST struct with methods to rewrite table refs, limits and to get annotations and column refs.

Currently rewriting to a base table is supported.

Column ref extraction is not a 100% right now. But dashboard as SQL is not well defined so we can create follow up PRs to fix it.

In general there might be a lot more edge cases. We should use follow up PRs and make this available for new reconcile ASAP.

### DIFF
--- a/runtime/pkg/duckdbsql/ast_values.go
+++ b/runtime/pkg/duckdbsql/ast_values.go
@@ -1,0 +1,92 @@
+package duckdbsql
+
+type astNode map[string]interface{}
+
+const (
+	astKeyError            string = "error"
+	astKeyStatements       string = "statements"
+	astKeyNode             string = "node"
+	astKeyType             string = "type"
+	astKeyKey              string = "key"
+	astKeyFromTable        string = "from_table"
+	astKeySelectColumnList string = "select_list"
+	astKeyTableName        string = "table_name"
+	astKeyFunction         string = "function"
+	astKeyFunctionName     string = "function_name"
+	astKeyChildren         string = "children"
+	astKeyValue            string = "value"
+	astKeyLeft             string = "left"
+	astKeyRight            string = "right"
+	astKeyColumnNames      string = "column_names"
+	astKeyAlias            string = "alias"
+	astKeyID               string = "id"
+	astKeySample           string = "sample"
+	astKeyColumnNameAlias  string = "column_name_alias"
+	astKeyModifiers        string = "modifiers"
+	astKeyLimit            string = "limit"
+	astKeyClass            string = "class"
+	astKeyCTE              string = "cte_map"
+	astKeyMap              string = "map"
+	astKeyQuery            string = "query"
+)
+
+func toBoolean(a astNode, k string) bool {
+	v, ok := a[k]
+	if !ok {
+		return false
+	}
+	switch vt := v.(type) {
+	case bool:
+		return vt
+	default:
+		return false
+	}
+}
+
+func toString(a astNode, k string) string {
+	v, ok := a[k]
+	if !ok {
+		return ""
+	}
+	switch vt := v.(type) {
+	case string:
+		return vt
+	default:
+		return ""
+	}
+}
+
+func toNode(a astNode, k string) astNode {
+	v, ok := a[k]
+	if !ok {
+		return nil
+	}
+	switch vt := v.(type) {
+	case map[string]interface{}:
+		return vt
+	default:
+		return nil
+	}
+}
+
+func toArray(a astNode, k string) []interface{} {
+	v, ok := a[k]
+	if !ok {
+		return make([]interface{}, 0)
+	}
+	switch v.(type) {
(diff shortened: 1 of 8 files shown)


runtime/pkg/duckdbsql/ast_values.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+func toArray(a astNode, k string) []interface{} {
+	v, ok := a[k]
+	if !ok {
+		return make([]interface{}, 0)
+	}
+	switch v.(type) {
+	case interface{}:
+		return v.([]interface{})
```
begelundmuller, line 85: Why not `case []interface{}`?

runtime/pkg/duckdbsql/duckdbsql.go
```diff
@@ -4,24 +4,163 @@ import (
...
 
-// Sanitize strips comments and normalizes a DuckDB SQL statement
-func Sanitize(sql string) (string, error) {
-	panic("not implemented")
+func (a *AST) GetTableRef() (*TableRef, bool) {
+	if len(a.fromNodes) == 1 {
+		return a.fromNodes[0].ref, true
+	}
+	return nil, false
+}
```
begelundmuller, line 84: What is the point of this func? (Looks like it will fail if multiple tables are referenced?)

runtime/pkg/duckdbsql/duckdbsql.go
```diff
@@ -4,24 +4,163 @@ import (
...
+		if !shouldReplace {
+			continue
+		}
+
+		if node.ref.Name == "" && newRef.Name != "" {
+			err := node.rewriteToBaseTable(newRef.Name)
+			if err != nil {
+				return err
+			}
+		}
```
begelundmuller, line 99: It looks like the node is not replaced in all cases.

If it's easiest to only support rewriting to base tables, that's okay, just put it in the docstring and return an error if attempting to rewrite to something else

runtime/pkg/duckdbsql/duckdbsql.go
```diff
@@ -48,50 +178,41 @@ type Annotation struct {
...
 	if err != nil {
-		return "", err
+		return nil, err
 	}
 
 	return res, nil
 }
 
 // Use a global in-memory DuckDB connection for invoking DuckDB's json_serialize_sql and json_deserialize_sql
+// TODO: Why not get driver connection?
```
begelundmuller, line 215: Since AST parsing doesn't depend on any DB state, just seemed nicer to decouple things

runtime/pkg/duckdbsql/duckdbsql.go
begelundmuller: Also, will this preserve aliases in all cases? E.g. if rewriting `SELECT * FROM "path/to/file.csv" foo`, the output should be `SELECT * FROM new_table_name foo`

begelundmuller (commented):
Left a few comments, but high level this approach looks good.

It's hard not to worry about the fragility of the SQL -> JSON mapping without proper type support, but I guess there's nothing we can do about that, except to have good testing.

We should also make sure to test for some "sad" cases like PIVOT and INSERT statements – we do not expect support for those, but should make sure we get nice errors and not panics in those cases.

runtime/pkg/duckdbsql/duckdbsql.go
AdityaHegde: This was some prototyping sources as sql. Will remove all related code.

runtime/pkg/duckdbsql/ast_values.go
AdityaHegde: I was getting issues around casting. We actually get into `case interface{}` instead of `case []interface{}` here.

runtime/pkg/duckdbsql/duckdbsql.go
AdityaHegde: Sounds good. I will remove the 1st check, we need to support base table to base table as well since plain urls like `from "s3:/path/to/data.csv"` needs to be replaced.

I have a test for making sure aliases are maintained. I get the alias from the existing node.

begelundmuller (approved):

PR #2746, opened 2023-07-06
Motherduck connector preflight APIs

k-anshul:
Preflight API to list all available tables across all databases


begelundmuller (approved):
Approving this for now, but as part of the connectors/drivers refactor, I think we should refactor this into a generic `OLAPListTables` API that takes `instance_id` and `connector` as args.

PR #2749, opened 2023-07-06
feat: unified actions selectors

skokenes:
## Checklist
- [x] Manual verification
- [ ] Unit test coverage
- [x] E2E test coverage
- [ ] Needs manual QA?


web-common/src/features/dashboards/actions/index.ts
```diff
@@ -0,0 +1,41 @@
...
+import { metricsExplorerStore } from "../dashboard-stores";
+
+export function clearFilterForDimension(
+  ctx: BusinessModel,
+  dimensionId,
+  include: boolean
+) {
+  const metricViewName = get(ctx.metricsViewName);
+  cancelDashboardQueries(ctx.queryClient, metricViewName);
+  metricsExplorerStore.clearFilterForDimension(
```
ericpgreen2, line 13: Should the `metricsExplorerStore` come from the business model context?

web-common/src/features/dashboards/filters/Filters.svelte
```diff
@@ -70,54 +45,15 @@ The main feature-set component for dashboard filters
...
-      // We prune the measure values and use the dimension labels for the filter
-      topListQuery = createQueryServiceMetricsViewToplist(
-        $runtime.instanceId,
-        metricViewName,
-        topListParams
-      );
+  let activeDimensionName: string;
+
+  $: {
+    if (activeDimensionName) {
```
ericpgreen2, line 51: If the query is conditional on `activeDimensionName`, should it go in the query's `enabled` option?

web-common/src/features/dashboards/selectors/index.ts
```diff
@@ -0,0 +1,95 @@
...
+              selector
+                ? selector(data?.entry?.metricsView)
+                : data?.entry?.metricsView,
+            queryClient: ctx.queryClient,
+          },
+        }
+      ).subscribe(set);
+    }
+  );
+};
```
ericpgreen2, line 32: Should we use the [v5 reactivity pattern](https://tanstack.com/query/v5/docs/svelte/reactivity) and put the derived store _within_ the `createQuery` (or `createOrvalGeneratedQuery`) call?

web-common/src/features/dashboards/actions/index.ts
ericpgreen2, line 1: I like this!

web-common/src/features/dashboards/filters/Filters.svelte
skokenes: Yea totally, good call. Let me take a stab at adding that and see how it looks

web-common/src/features/dashboards/actions/index.ts
skokenes: Yea actually all it really contains at this point is actions, and I think we should just split those out into separate action functions in the actions file.

I didn't do that yet here to limit the scope of changes, but can try adding that as well

web-common/src/features/dashboards/selectors/index.ts
skokenes: Really interesting idea that I hadn't considered. Let's discuss as a team, I see pros and cons:
- pro: will make moving to sq5 easier later
- cons: will require more code changes as we refactor here. Not clear yet how exactly we want to use sq5 API so we may be guessing at this point

web-common/src/features/dashboards/actions/index.ts
AdityaHegde: If we do split it, it would perhaps make sense to move the function body here wrapped with an update.

web-common/src/features/dashboards/selectors/index.ts
AdityaHegde: Would we need to make tweaks to orval generator? Does it even allow us to do this?

web-common/src/features/dashboards/filters/Filters.svelte
skokenes: Done

web-common/src/features/dashboards/actions/index.ts
skokenes: Done

web-common/src/features/dashboards/state-managers/state-managers.ts
```diff
@@ -33,10 +38,22 @@ export function createStateManagers({
...
+      return store.subscribe(set);
+    }
+  );
+
+  const updateDashboard = (
+    callback: (metricsExplorer: MetricsExplorerEntity) => void,
+    absenceCallback?: () => MetricsExplorerEntity
+  ) => {
+    const name = get(dashboardStore).name;
+    updateMetricsExplorerByName(name, callback, absenceCallback);
```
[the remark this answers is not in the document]
skokenes: Yes I agree, I will leave a TODO here for now.
The reason I didn't outright do it yet is that it requires access to the `update` method from the MetricsExplorerStore, and that method is currently not publicly exposed outside of the singleton file where MetricsExplorerStore is created.

AdityaHegde (approved):

ericpgreen2 (approved):

Issue #2730, opened 2023-07-03
Runtime: Implement file watching


begelundmuller:
For async reconcile, we need the ability to watch for for file changes on disk. The capability should be implemented on `drivers.RepoStore` (only for the `drivers/file` driver) and exposed over a streaming gRPC API. 

Proposed API spec:

```proto
service RuntimeService {
  // WatchFiles streams repo file update events. It is not supported on all backends.
  rpc WatchFiles(WatchFilesRequest) returns (stream WatchFilesResponse) {
    option (google.api.http) = {
      get: "/v1/instances/{instance_id}/files/watch"
    };
  }
}

// FileEvent describes a file change.
enum FileEvent {
  FILE_EVENT_UNSPECIFIED = 0;
  FILE_EVENT_WRITE = 1;
  FILE_EVENT_DELETE = 2;
  FILE_EVENT_RENAME = 3;
}

// Request message for RuntimeService.WatchFiles
message WatchFilesRequest {
  string instance_id = 1 [(validate.rules).string = {pattern: "^[_\\-a-zA-Z0-9]+$"}];
  bool replay = 2;
}

// Response message for RuntimeService.WatchFiles
message WatchFilesResponse {
  FileEvent event = 1;
  string path = 2;
}
```

For the file watching logic, consider using https://github.com/fsnotify/fsnotify (potentially override it to use https://github.com/fsnotify/fsevents on macOS instead).

egor-ryashin:
Could you elaborate `replay`?

begelundmuller:
> Could you elaborate `replay`?

If `replay` is true, it should emit `WRITE` events for all the existing files at the start of the request

begelundmuller:
Closed by https://github.com/rilldata/rill/pull/2753

PR #2753, opened 2023-07-07
Runtime: Implement file watching 

egor-ryashin:
https://github.com/rilldata/rill/issues/2730


proto/rill/runtime/v1/api.proto
```diff
@@ -190,6 +190,11 @@ service RuntimeService {
   rpc ListConnectors(ListConnectorsRequest) returns (ListConnectorsResponse) {
     option (google.api.http) = {get: "/v1/connectors/meta"};
   }
+
+  // WatchFiles streams repo file update events. It is not supported on all backends.
```
begelundmuller, line 194: Can we move it up to after `ListFiles`? (and same with Request/Response messages for easy navigation

runtime/server/server.go
```diff
@@ -243,9 +243,15 @@ func timeoutSelector(service, method string) time.Duration {
 	if service == "rill.runtime.v1.RuntimeService" && (strings.Contains(method, "Trigger") || strings.Contains(method, "Reconcile")) {
 		return time.Minute * 30
 	}
+
 	if service == "rill.runtime.v1.QueryService" {
 		return time.Minute * 5
 	}
+
+	if method == WatchFilesMethod {
```
begelundmuller, line 251: Should we refactor this func to pass in the full method name, and then use `runtimev1.RuntimeService_WatchFiles_FullMethodName` (and related constants) instead?

runtime/server/watchfiles.go
```diff
@@ -0,0 +1,115 @@
...
+
+func (s *Server) WatchFiles(req *runtimev1.WatchFilesRequest, stream runtimev1.RuntimeService_WatchFilesServer) error {
+	repo, err := s.runtime.Repo(stream.Context(), req.InstanceId)
+	if err != nil {
+		return err
+	}
+
+	if repo.Driver() != "file" {
+		return fmt.Errorf("%s repository is not supported", repo.Driver())
+	}
```
begelundmuller, line 26: Instead of checking the repo is a `file` type in the server, could we add something like a `Watch(ctx, callback func(event)) error` function on `drivers.RepoStore` instead? And have it return an error from any repo driver that doesn't support it (like `github`, which is readonly).

Imagine we wanted to add a DB-backed repo store in the future (to support the data modeler in the cloud). In that implementation, we could then use something like Postgres' `LISTEN/NOTIFY` to watch changes inside the driver, without refactoring the server implementation.

begelundmuller (changes requested):

runtime/drivers/file/repo.go
```diff
@@ -135,3 +137,109 @@ func (c *connection) Delete(ctx context.Context, instID, filePath string) error
 func (c *connection) Sync(ctx context.Context, instID string) error {
 	return nil
 }
+
+func (c *connection) Watch(ctx context.Context, replay bool, callback drivers.WatchCallback) error {
+	if c.Driver() != "file" {
+		return fmt.Errorf("%s repository is not supported", c.Driver())
+	}
```
begelundmuller, line 144: Since this is in the `runtime/drivers/file` driver, is this redundant?

runtime/drivers/file/repo.go
```diff
@@ -135,3 +137,109 @@ func (c *connection) Delete(ctx context.Context, instID, filePath string) error
...
+			return err
+		}
+	}
+
+	for {
+		select {
+		case event, ok := <-watcher.Events:
+			if !ok {
+				return nil
+			}
```
begelundmuller, line 195: Should this return an error?

runtime/drivers/file/repo.go
```diff
@@ -135,3 +137,109 @@ func (c *connection) Delete(ctx context.Context, instID, filePath string) error
...
+
+			e := drivers.WatchEvent{
+				Path: filepath.Join("/", relativeName),
+			}
+			if event.Op&fsnotify.Create != 0 {
+				e.Type = runtimev1.FileEvent_FILE_EVENT_WRITE
+				fi, err := os.Stat(event.Name)
+				if err != nil {
+					return err
+				} else if fi.IsDir() {
```
begelundmuller, line 210: Would it make sense to ignore the error? In case the file was a tmp file that was rapidly moved

runtime/drivers/github/repo.go
```diff
@@ -115,3 +115,7 @@ func (c *connection) Delete(ctx context.Context, instID, filePath string) error
 func (c *connection) Sync(ctx context.Context, instID string) error {
 	return c.cloneOrPull(ctx, false)
 }
+
+func (c *connection) Watch(ctx context.Context, replay bool, callback drivers.WatchCallback) error {
+	return fmt.Errorf("%s repository is not supported", c.Driver())
```
begelundmuller, line 120: Mention "watch" in the err message

runtime/server/server.go
```diff
@@ -239,13 +239,19 @@ func HTTPErrorHandler(ctx context.Context, mux *gateway.ServeMux, marshaler gate
 	gateway.DefaultHTTPErrorHandler(ctx, mux, marshaler, w, r, err)
 }
 
-func timeoutSelector(service, method string) time.Duration {
-	if service == "rill.runtime.v1.RuntimeService" && (strings.Contains(method, "Trigger") || strings.Contains(method, "Reconcile")) {
+func timeoutSelector(fullMethodName string) time.Duration {
+	if strings.HasPrefix(fullMethodName, "rill.runtime.v1.RuntimeService") && (strings.Contains(fullMethodName, "/Trigger") || strings.Contains(fullMethodName, "/Reconcile")) {
```
begelundmuller, line 243: 1. Missing `/` prefix for service name check
2. I think some of the reconcile functions current aren't prefixed with `Reconcile`, so probably shouldn't have `/` prefix

runtime/server/watchfiles.go
```diff
@@ -0,0 +1,29 @@
+package server
+
+import (
+	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
+	"github.com/rilldata/rill/runtime/drivers"
+)
+
+const (
+	WatchFilesMethod = "WatchFiles"
+)
```
begelundmuller, line 10: Not used

begelundmuller (changes requested):

runtime/drivers/file/repo.go
egor-ryashin: The channel is closed, it can be a shutdown process, should we return an error during a planned shut down?

begelundmuller (approved):

PR #2754, opened 2023-07-08
2344: Usage tracking

### LINKED ISSUE #2344: Runtime: Usage tracking
See details in [Notion](https://www.notion.so/rilldata/PRDs-a1fa12e88b994e5e9a63c879deda6189?p=54921cb3597045b89de12b1e2f54c0a5&pm=s)

esevastyanov:
Closes #2344 

Activity Client (Buffered) is an accumulator of usage tracking events, it periodically sinks the accumulated events.
Sink receives events from Client and sinks them to console/~file~/kakfa/~gcs~.

Types of Sink:
1. Noop: does nothing (is used by locally running app only)
2. Console: logs events to a console (might be used by locally running app)
3. ~File: appends events to a file (might be used by locally running app so that stored events are able to be ingested by rill)~
4. Kakfa: produces events to a Kafka cluster
5. ~GCS: stores events to a GCS bucket partitioning them by organization_id/project_id (might be used to partition events by organization and project so that each organization can access only its projects' usage events)~

Each event is converted to a JSON


admin/deployments.go
```diff
@@ -99,6 +100,8 @@ func (s *Service) createDeployment(ctx context.Context, opts *createDeploymentOp
 		EmbedCatalog:        embedCatalog,
 		Variables:           opts.ProdVariables,
 		IngestionLimitBytes: ingestionLimit,
+		OrganizationId:      opts.OrganizationID,
+		ProjectId:           opts.ProjectID,
```
nishantmonu51, line 104: instead of explicitly defined types for org_id and project_id, would recommend making it a bit generic by adding a Map of labels that are associated with a project. Also, these labels should *only* be used for the purpose of emitting metrics.   
For initial version we can start with a list of pre-defined labels that runtime adds like org_id, org_name, project_id, project_name. 

In future, we can also support users defining their own labels to projects. some labels could be - 

1. `poc:true` - to attribute cost of projects internally for poc's vs paid accounts 
2. `team: sales/eng/CS` - again handy for cost and usage attribution 
4. `orgName` - name of org
5. `projectName` - name of project

cli/cmd/runtime/start.go
```diff
@@ -56,6 +57,18 @@ type Config struct {
...
+	// Usage sink period in millis
+	UsageSinkPeriod int `default:"1000" split_words:"true"`
+	// Max queue size of a usage client
+	UsageMaxQueueSize int `default:"1000" split_words:"true"`
+	// Kafka brokers of a usage sink
+	UsageKafkaBrokers string `default:"" split_words:"true"`
+	// Kafka topic of a usage sink
+	UsageKafkaTopic string `default:"" split_words:"true"`
+	// GCS bucket of a usage sink
+	UsageGCSBucket string `default:"" split_words:"true"`
```
nishantmonu51, line 71: These emitters can be used to emit metrics that are not usage related as well, like health events. How about renaming to generic Emitters ?

runtime/pkg/middleware/usage.go
```diff
@@ -0,0 +1,61 @@
...
+		}
+
+		newCtx := usage.ContextWithUsageDims(ss.Context(), *usage.String("user_id", subject))
+		wss := grpc_middleware.WrapServerStream(ss)
+		wss.WrappedContext = newCtx
+
+		start := time.Now()
+		defer func() {
+			// Emit usage metric
+			usage.GetClient().Emit(newCtx, "request/time", float64(time.Since(start).Milliseconds()))
```
nishantmonu51, line 32: does it also contain the request name and any other dimensions ?

runtime/pkg/middleware/usage.go
```diff
@@ -0,0 +1,67 @@
...
+	) error {
+		subject := auth.GetClaims(ss.Context()).Subject()
+		if subject == "" {
+			subject = "anonymous"
+		}
+
+		newCtx := emitter.ContextWithUsageDims(ss.Context(),
+			*emitter.String("user_id", subject),
+			*emitter.String("request_method", info.FullMethod),
+		)
```
begelundmuller, line 28: 1. Override the existing `ctx` instead of having `newCtx` – ensures no bugs will happen by accidentally using the old `ctx`
2. Feels like the pointer dereferencing here should not be necessary – can any of the funcs be refactored to make them work better together

runtime/pkg/middleware/usage.go
```diff
@@ -0,0 +1,67 @@
...
+	) (interface{}, error) {
+		subject := auth.GetClaims(ctx).Subject()
+		if subject == "" {
+			subject = "anonymous"
+		}
+
+		newCtx := emitter.ContextWithUsageDims(ctx,
+			*emitter.String("user_id", subject),
+			*emitter.String("request_method", info.FullMethod),
+		)
```
begelundmuller, line 57: See earlier comments

runtime/registry.go
```diff
@@ -16,7 +17,17 @@ func (r *Runtime) FindInstances(ctx context.Context) ([]*drivers.Instance, error
...
+
+	if err != nil && instance != nil {
+		var usageDims []emitter.Dim
+		for k, v := range instance.Labels {
+			usageDims = append(usageDims, *emitter.String(k, v))
+		}
+		emitter.AddDimsToContext(ctx, usageDims...)
+	}
+	return instance, err
 }
```
begelundmuller, line 31: `FindInstance` may be called for a variety of reasons – even if it's *likely* to only be called for a single instance on instance-scoped API calls, that dependency is too implicit for us to rely on it for tracking.

I would consider a more explicit approach, similar to logging. Perhaps having a `usageClient.WithDims(...)` wrapper could help (not sure).

begelundmuller (changes requested):
In addition to the comments below, I'm not a fan of the package name `emitter`. Unfortunately, we have the issue of emitting a lot of different things from the application, and this is only one of them. We need a name that uniquely identifies the usage/activity emitting use case, as opposed to:

- Emitting observability (logging/metrics/traces)
- Emitting product analytics telemetry

runtime/registry.go
esevastyanov: Seems like every request with an instance id restores the instance from a DB thus I updated the logic so that it explicitly stores the instance id into the context and at the time FindInstance is called, the context is populated by the instance labels. This keeps the same number of requests to the metastore.

runtime/pkg/emitter/global.go
```diff
@@ -0,0 +1,16 @@
...
+
+var globalClient unsafe.Pointer
+
+func Get() *Client {
+	return (*Client)(atomic.LoadPointer(&globalClient))
+}
+
+func Set(client *Client) {
+	atomic.StorePointer(&globalClient, unsafe.Pointer(client))
+}
```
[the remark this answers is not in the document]
esevastyanov: Moved to Server

runtime/pkg/emitter/sink.go
```diff
@@ -0,0 +1,103 @@
...
+	"fmt"
+	"os"
+	"sync"
+
+	"go.uber.org/zap"
+)
+
+// Sink is used by a Client to sink collected Event-s.
+type Sink interface {
+	Sink(events []Event)
```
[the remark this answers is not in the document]
esevastyanov: Added a context into Sink but it is not used in existing implementations because there is no way to cancel the event producing into kafka, I can only stop event serialization

runtime/pkg/middleware/usage.go
esevastyanov: request method + other dimensions created based on request params

cli/cmd/runtime/start.go
esevastyanov: Renamed to Activity

runtime/pkg/activity/client.go
```diff
@@ -0,0 +1,171 @@
...
+	c.buffer = append(c.buffer, event)
+
+	if len(c.buffer) >= c.bufferSize {
+		go func() {
+			err := c.flush()
+			if err != nil {
+				c.logger.Error("could not flush activity events", zap.Error(err))
+			}
+		}()
+	}
```
begelundmuller, line 77: I think this can cause buffer overflows – the goroutine may not get scheduled before another call to Emit acquires the lock, causing it to append to the already full buffer

begelundmuller (changes requested):

go.mod
```diff
@@ -134,7 +135,7 @@ require (
 	github.com/dustin/go-humanize v1.0.1 // indirect
 	github.com/emirpasic/gods v1.18.1 // indirect
 	github.com/felixge/httpsnoop v1.0.3 // indirect
-	github.com/fsnotify/fsnotify v1.6.0 // indirect
+	github.com/fsnotify/fsnotify v1.6.0
```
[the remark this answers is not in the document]
esevastyanov: This is a result of running `go mod tidy`

runtime/pkg/middleware/activity.go
```diff
@@ -0,0 +1,63 @@
...
+
+		start := time.Now()
+		defer func() {
+			// Emit usage metric
+			activityClient.Emit(ctx, "request/time", float64(time.Since(start).Milliseconds()))
+		}()
+
+		return handler(ctx, req)
+	}
+}
```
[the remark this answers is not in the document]
esevastyanov: Sure, If we already started following the Prometheus guidlines then lets continue doing this. The name `request/time` follows [Druid](https://druid.apache.org/docs/latest/operations/metrics.html) naming convention

runtime/pkg/observability/middleware.go
```diff
@@ -322,4 +323,7 @@ func AddRequestAttributes(ctx context.Context, attrs ...attribute.KeyValue) {
 			*fields = append(*fields, zap.Any(string(attr.Key), attr.Value.AsInterface()))
 		}
 	}
+
+	// Add attributes as activity dimensions
+	activity.WithDims(ctx, attrs...)
```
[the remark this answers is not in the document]
esevastyanov: How large might the payloads be? The default limit of Kafka message size is 1MB, it is a quarter of grpc message limit

runtime/pkg/activity/client.go
```diff
@@ -0,0 +1,171 @@
...
+		key := string(dim.Key)
+		if key != "time" && key != "name" && key != "value" {
+			flattened[key] = dim.Value.AsInterface()
+		} else {
+			return nil, errors.New("time/name/value are not allowed as dimension names")
+		}
+	}
+
+	return json.Marshal(flattened)
+}
```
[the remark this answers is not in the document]
esevastyanov: + Removed a check of time/name/value dimension names in favour of post-override
+ I believe a custom encoder might be implemented separately if the current approach would affect the performance

begelundmuller (changes requested):
Everything else looks good!

begelundmuller (approved):

PR #2759, opened 2023-07-10
fix: Filters bar not rendering search results properly

skokenes:
## Checklist
- [x] Manual verification
- [x] Unit test coverage
- [ ] E2E test coverage
- [ ] Needs manual QA?

#### Issue addressed: 
Fixes bugs:
- Filter bar search results were not rendering properly. Showing "undefined" instead
- Filter bar search would flash an empty state on first search
- Search component had a race condition with autofocus that could cause a TypeError to throw

#### Details:
**Filter bar search results were not rendering properly. Showing "undefined" instead**: The Filters component tries to look up search result data based on the dimension id, but the API returns the data keyed by dimension column name. This PR adds a lookup of the column name and uses that to access the search result display value instead of the dimension id.

**Filter bar search would flash an empty state on first search**: By default, we pass an empty array of search results to the RemovableMenuList component. This component has no way of distinguishing between what are actually empty search results, and what is just missing data (say when a search has not yet happened). This causes the component to flash a "No results found" message when a search is started for the first time but hasn't actually finished. This PR adds the ability to pass `null` for the searchedValues of the RemovableMenuList, indicating to the component that we do not have any search results yet.

**- Search component had a race condition with autofocus that could cause a TypeError to throw**: The Search component tries to focus the input on the next animation frame after mounting. However, its possible that within the first frame the component could be unmounted again. In this case, the input no longer would exist and can't be focused. We add an optional check so that a TypeError isn't thrown.

## Steps to Verify
1. Open a dashboard
2. Apply some filters
3. Click on a filter pill and type a search term in
4. Observe the search results appearing after fetching, with proper display names and no intermediate state

### DIFF
--- a/web-common/src/features/dashboards/filters/Filters.svelte
+++ b/web-common/src/features/dashboards/filters/Filters.svelte
@@ -69,7 +69,7 @@ The main feature-set component for dashboard filters
 
   let topListQuery;
   let searchText = "";
-  let searchedValues = [];
+  let searchedValues: string[] | null = null;
   let activeDimensionName;
 
   $: metricTimeSeries = useModelHasTimeSeries(
@@ -82,7 +82,7 @@ The main feature-set component for dashboard filters
 
   $: if (activeDimensionName) {
     if (searchText == "") {
-      searchedValues = [];
+      searchedValues = null;
     } else {
       let topListParams = {
         dimensionName: activeDimensionName,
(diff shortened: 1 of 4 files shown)


web-common/src/features/dashboards/filters/Filters.svelte
```diff
@@ -128,8 +128,10 @@ The main feature-set component for dashboard filters
 
   $: if (!$topListQuery?.isFetching && searchText != "") {
     const topListData = $topListQuery?.data?.data ?? [];
-    searchedValues =
-      topListData.map((datum) => datum[activeDimensionName]) ?? [];
+    const activeColumn =
```
AdityaHegde, line 131: We should move the calculation of the active column outside of this since this runs every keystroke.

AdityaHegde (changes requested):

web-common/src/features/dashboards/filters/Filters.svelte
skokenes: Done

AdityaHegde (approved):

PR #2760, opened 2023-07-10
Unified Time Range data

AdityaHegde:
## Checklist
- [x] Manual verification
- [ ] Unit test coverage
- [ ] E2E test coverage
- [x] Needs manual QA?

#### Issue addressed: 
We have a lot of null checks in our components, mainly around time range fields. This can lead to issues in some edge cases where unexpected value is used and can lead to errors.

#### Details:
Replacing all usage of time related fields,
- Start and end for selected time range
- Comparison time ranges
- Scrubbed time range
- Timezone

Adding a store that always maintains expected values for all the above fields.

## Steps to Verify
Hard to give the exact steps. Playing around with dashboard is needed.


web-common/src/features/dashboards/time-controls/TimeControlsStoreTest.svelte
```diff
@@ -0,0 +1,14 @@
...
+   * This component is used to test the time control store. Without a render the requests are not resolving.
+   * TODO: get rid of it once testing is figured out.
+   */
+
+  export let timeControlsStore: TimeControlStore;
+</script>
+
+<div>
+  {$timeControlsStore.timeStart}
+</div>
```
djbarnwal, line 14: Is this file still used somewhere?

djbarnwal (changes requested):
Looks good for most parts. Found some issues around time
- Default time range of `P14D` does not work. Not sure if this is specific to this branch
- The comparison value before zooming and after zooming changes

**Before zoom**

image

**After zoom**

image

(-52 vs -64) The comparison value fetched are different

- Are we still using `useFetchTimeRange` and `useComparisonRange`. Can we remove those and their references from the codebase?

AdityaHegde:
Removed the unused methods. The 2 issues you mentioned is present on main as well. Lets merge this and fix them later.

web-common/src/features/dashboards/time-series/MetricsTimeSeriesCharts.svelte
```diff
@@ -215,13 +215,7 @@
           start,
           end,
         });
-
-        // hack to prevent race condition in URL proto changes
-        setTimeout(() => {
-          metricsExplorerStore.setSelectedScrubRange(metricViewName, undefined);
-        }, 50);
-      }
-      if (
```
djbarnwal, line 272: Do we have to remove this? I can still see scrub being shown even after we zoom in

djbarnwal (approved):
If I zoom into a scrub, I still see a scrub on top of it. Any reason we removed the timeout?
Approving so that you can merge at your discretion.

web-common/src/features/dashboards/time-series/MetricsTimeSeriesCharts.svelte
AdityaHegde: Lets fix this on main and unblock you. We should not need the timeout.

PR #2761, opened 2023-07-10
Async reconcile: resource definitions and code file parser

begelundmuller:
The key files in this PR are the revamped resource definitions in `proto/rill/runtime/v1/resources.proto` and the new compiler for parsing projects into resources in `runtime/compilers/rillv1`.

Some implementation notes (in no particular order):

- Supports merging config from multiple files, e.g. `model.yaml` and `model.sql` can set config on the same resource
- Supports emitting multiple resources from one file, e.g. for embedded sources
- Supports a flexible directory structure (files outside the known directories must specify a `kind` in the YAML or SQL)
- Introduces a `Migration` resource, to support `init.sql` and other ad-hoc SQL statements
- Supports specifying a `connector` for any SQL-based resource, allowing e.g. models to be orchestrated on other DBs. If `connector == ""`, the new reconcile will use the default connector (which is DuckDB on `stage.db`).
- Doesn't rely on file timestamps
- Supports incremental parsing of a subset of files (for keystroke-by-keystroke)
- Adds a templating engine for resolving templated queries
- If templating is used in a SQL file, DuckDB inference (like detecting referenced tables or rewriting embedded sources) is disabled. Instead, you have to use `{{ ref "other-model" }}` for models-referencing-models
- Supports setting config in SQL files using `-- @annotation: ` or `{{ configure "key" "value" }}` syntax. The comment-based annotations are only supported for DuckDB SQL files.
- The parser only parses files in isolation and doesn't check references. Any DAG-based validation is expected to be handled by reconcile.

Notes about catalog changes:

- The word "resource" replaces "catalog entry/object"
- Resources have separate "spec" (desired state) and "state" (actual state)
- The new catalog will support multiple resources of different kinds with the same name
  - For cases where uniqueness is required across kinds (specifically source and model), the reconciler will be responsible for validating uniqueness


runtime/compilers/rillv1/parser.go
```diff
@@ -0,0 +1,603 @@
...
+			checkPaths = append(checkPaths, stem+".yaml", stem+".yml")
+		} else if isYAML {
+			checkPaths = append(checkPaths, stem+".sql")
+		}
+
+		// Remove all resources derived from this path, and add any related paths to the check list
+		for _, resource := range p.resourcesForPath[path] {
+			// Multiple entries in resourcesForPath may point to the same resource.
+			// By adding resource.Paths to checkPaths, the outer loop will eventually clear those (maybe it already has).
+			checkPaths = append(checkPaths, resource.Paths...)
```
AdityaHegde, line 236: Should we do the deduplication here? Embedded sources seem to explode checkPaths quite a bit. The test with 2 models referencing once source leads to about 14 entries in this when one of the model is deleted.

runtime/compilers/rillv1/parse_metrics_view.go
```diff
@@ -0,0 +1,171 @@
...
+
+	spec.Title = tmp.Title
+	spec.Description = tmp.Description
+	spec.Model = tmp.Model
+	spec.TimeDimension = tmp.TimeDimension
+	spec.SmallestTimeGrain = smallestTimeGrain
+	spec.DefaultTimeRange = tmp.DefaultTimeRange
+	spec.AvailableTimeZones = tmp.AvailableTimeZones
+
+	for i, dim := range tmp.Dimensions {
```
AdityaHegde, line 95: Where will the duplicate dimension name and measure name go if not here?

runtime/compilers/rillv1/parser.go
begelundmuller: Good observation. I tried investigating some simple deduplication checks / other logic changes, but wasn't able to come up with something that benchmarked better. For small list sizes, it seems appending to checkPaths is about the same cost as doing extra deduplication checks.

Given the small number of files that will usually be reparsed, I think it's okay to keep it slightly unoptimized (the seenPaths will still guard against duplicating any expensive ops, like checking file stat), and it's nice to keep the logic simpler

AdityaHegde (approved):
Code looks good. Any tweaks can go in new PRs

Issue #1519, opened 2022-12-22
Improve infra drivers abstractions


begelundmuller:
The `runtime/drivers` package contains the runtime's logic for interfacing with external infra for data storage and processing (like duckdb, druid, postgres, etc.). It's an important package, but has grown organically and is in need of a few refactors and better architectural docs.

As background, each driver in the package can implement one or more of four interfaces:
- `OLAPStore` for storing data and running analytical queries
- `CatalogStore` for storing sources, models and metrics views, including metadata like last refresh time
- `RepoStore` for storing code artifacts (this is essentially a file system shim)
- `RegistryStore` for tracking instances (DSNs for OLAPs and repos)

Much of the complexity in the package derives from the fact that each driver may implement more than one of these interfaces. For example:
- In local single-instance mode (what you get when running `rill start`), the configuration is:
  - Registry: in-memory
  - Repo: file system
  - OLAP: DuckDB
  - Catalog: DuckDB
- In a managed cloud offering, we may have:
  - Registry: Postgres
  - Repo: Git (readonly)
  - OLAP: DuckDB, Druid, and maybe others
  - Catalog: Postgres

A particularly confusing problem right now is that some drivers are single-instance while others are multi-instance. This is currently not enforced in the interfaces, but it should be. For example, in the examples above, the local catalog is instance-specific (catalog info is embedded with the data in a DuckDB file), while the cloud catalog is stored in a Postgres database that's shared across all instances. 

Tasks:

- [x] Add a README in the package explaining the OLAP, catalog, repo, and registry concepts
- [x] Enforce difference between single-instance and multi-instance drivers
  - Add `shared bool` parameter to `Open` – the option indicates multi-instance vs single-instance usage
  - Refactor functions to accept an instance ID on instantiation instead of in sub-functions – i.e. make them `c.CatalogStore(instanceID)`, `c.OLAPStore(instanceID)` and `c.RepoStore(instanceID)` and remove the `instanceID` parameter from their interface functions. The implementations should check `shared` before resolving and return an error for configurations they don't support. If `shared` is true and they support multi-instance mode, they should return a wrapped object containing the instance ID.
- [x] Rename the `Connection` type to `Handle` to avoid ambiguity with DB connections
- [x] Rename the `Connection.XXXStore` functions to `Connection.AsXXX` 
- [x] Update the runtime instantiation logic to reflect the clearer abstractions:

 ```go
// NOTE: Just a draft

 type runtime.Options struct {
 	SharedInfra: {
[...]

PR #2763, opened 2023-07-11
Connector refactor

k-anshul:
- Unifies connectors and drivers into a single package by adding new interfaces to drivers to encapsulate the logic currently found in connectors.
- Introduces a concept of transporters that move data between different connectors. A Transporter can be registered for any pair of connector types.
- Introduces a Progress interface that transporters can use to report progress info.

TODO :: 

- [ ] Use the connectors declared in rill.yaml for all driver/connector configuration (including OLAP). By default, we will automatically add a duckdb connector with dsn: stage.db to every project.
- [ ] Implement the drivers refactors proposed in https://github.com/rilldata/rill/issues/1519
- [ ] Comment on https://github.com/rilldata/rill/pull/2746#pullrequestreview-1518313358


runtime/drivers/s3/s3.go
```diff
@@ -234,7 +249,7 @@ func getAwsSessionConfig(ctx context.Context, conf *Config, bucket string, creds
 	}
 
 	sharedConfigState := session.SharedConfigDisable
-	if env.AllowHostAccess {
+	if val, ok := c.config["allow_host_access"]; ok && val.(bool) {
```
begelundmuller, line 252: type cast panic issue

runtime/drivers/s3/api.go
```diff
@@ -27,7 +26,7 @@ func (c Connector) ListBuckets(ctx context.Context, env *connectors.Env) ([]stri
 	}
 
 	sharedConfigState := session.SharedConfigDisable
-	if env.AllowHostAccess {
+	if val, ok := c.config["allow_host_access"]; ok && val.(bool) {
```
begelundmuller, line 29: 1. type cast panic issue 
2. should we pass `allow_host_access` as an explicit param in `Open` (maybe have `opts *OpenOptions` for future params)? Maybe not worth it, don't know

runtime/drivers/s3/s3.go
```diff
@@ -83,8 +124,8 @@ type Config struct {
 	url                   *globutil.URL
 }
 
-func ParseConfig(props map[string]any) (*Config, error) {
-	conf := &Config{}
+func parseConfig(props map[string]any) (*config, error) {
```
begelundmuller, line 127: Maybe rename to `parseSourceProperties` to disambiguate from the connector config?

runtime/services/catalog/migrator/sources/sources.go
```diff
@@ -176,50 +176,168 @@ func (m *sourceMigrator) ExistsInOlap(ctx context.Context, olap drivers.OLAPStor
...
+		case <-ticker.C:
+			olap, _ := olapConnection.AsOLAPStore()
+			if size, ok := olap.EstimateSize(); ok && size > ingestionLimit {
+				limitExceeded = true
+				cancel()
+			}
+		}
+	}()
+	err = t.Transfer(ctxWithTimeout, src, sink, drivers.NewTransferOpts(drivers.WithLimitInBytes(ingestionLimit)), p)
+	done <- true
```
begelundmuller, line 244: I think you can remove `done` and just have a `case <-ctxWithTimeout.Done()`

runtime/services/catalog/migrator/sources/sources.go
```diff
@@ -176,50 +176,168 @@ func (m *sourceMigrator) ExistsInOlap(ctx context.Context, olap drivers.OLAPStor
...
+			Table: tableName,
+		}
+	default:
+		return nil
+	}
+}
+
+// TODO :: may be pass all envs ?
+// It is slightly confusing right now that what gets passed to driver.Open vs source property
+func connectorVariables(src *runtimev1.Source, env map[string]string, repoRoot string) map[string]any {
```
begelundmuller, line 319: Agreed on the confusion. Hopefully some of the other refactors for connector config will help alleviate it

begelundmuller (changes requested):

runtime/drivers/drivers.go
```diff
@@ -72,19 +96,222 @@ type Connection interface {
...
+	// AsRepoStore returns a AsRepoStore if the driver can serve as such, otherwise returns false.
 	// A repo stores file artifacts (either in a folder or virtualized in a database).
-	RepoStore() (RepoStore, bool)
+	AsRepoStore() (RepoStore, bool)
 
-	// OLAPStore returns an OLAPStore if the driver can serve as such, otherwise returns false.
+	// AsOLAPStore returns an AsOLAPStore if the driver can serve as such, otherwise returns false.
 	// OLAP stores are where we actually store, transform, and query users' data.
-	OLAPStore() (OLAPStore, bool)
+	AsOLAPStore() (OLAPStore, bool)
```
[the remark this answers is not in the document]
k-anshul: We can change to `AsRegistry` and `AsOLAP`. For others yeah I think it makes sense to keep `Store` suffix.

runtime/drivers/duckdb/duckdb.go
```diff
@@ -103,31 +121,70 @@ type connection struct {
 	dbErr       error
 }
 
+// Driver implements drivers.Connection.
+func (c *connection) Driver() string {
+	if val, ok := c.driverConfig["driver"]; ok {
+		return val.(string)
+	}
+	return "duckdb"
+}
```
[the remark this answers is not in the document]
k-anshul: 2. I felt even driver name is also a config. May be it will be more explicit once we move to parsing it via `mapstructure`. But we could pass name as parameter as well. What do you think ?

runtime/drivers/s3/api.go
k-anshul: 2. We can parse config to a structure and that should make it more explicit.

runtime/drivers/duckdb/duckdb.go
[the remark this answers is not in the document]
begelundmuller: Since the caller passes it as a separate parameter, not as part of the config, it feels like it should also be a separate parameter elsewhere (otherwise, it's easy to forget passing it in the config also).

An alternative would be to have separate drivers passed to `drivers.Register`, like `drivers.Register("motherduck", motherduckDriver)`, which then shares the same implementation, but set different driver names.

Probably, just passing the driver name to `Open` is easiest

runtime/drivers/duckdb/motherduck/motherduck.go
```diff
@@ -0,0 +1,42 @@
...
+
+type connector struct{}
+
+func (c connector) Spec() drivers.Spec {
+	return spec
+}
+
+func (c connector) HasAnonymousAccess(ctx context.Context, props map[string]any) (bool, error) {
+	return false, nil
+}
```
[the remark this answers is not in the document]
begelundmuller: I think what throws me off is the implicit coupling between the connector name and the driver name, and the implicit relationship between the connector spec and the driver config properties.

I'm not really able to put into words what the difference is between a "connector" and a "driver" anymore, so that's why I thought we should unify.

Any other ideas for how we can reduce the confusion?

runtime/drivers/s3/api.go
begelundmuller: I'm good either way. Parsing to a config also seems good

runtime/drivers/duckdb/duckdb.go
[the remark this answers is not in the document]
k-anshul: I liked the alternative better since name in all cases would be the same. `name` can just be a field in a driver which can then be passed to connection as driver name.

runtime/drivers/file/file.go
```diff
@@ -28,20 +61,54 @@ func (d driver) Open(dsn string, logger *zap.Logger) (drivers.Connection, error)
...
+}
+
+type sourceProperties struct {
+	Path   string `mapstructure:"path"`
+	Format string `mapstructure:"format"`
+}
+
+func parseSourceProperties(props map[string]any) (*sourceProperties, error) {
+	conf := &sourceProperties{}
+	err := mapstructure.Decode(props, &conf)
```
begelundmuller, line 94: Not sure it's a problem here, but I have previously seen issues in some libraries when parsing into a `**` (it uses `&` twice). Other places you use `mapstructure` in this PR only takes one level of pointer. Maybe fix to be safe?

begelundmuller (approved):

PR #2764, opened 2023-07-11
Runtime: Add available_time_zones to dashboard YAML

begelundmuller:
Uses `time.LoadLocation` to validate IANA time zone identifiers


runtime/services/catalog/artifacts/artifacts_test.go
```diff
@@ -633,6 +633,61 @@ region: {{.env.region}}
...
 }
 
+func TestMetricsViewAvailableTimeZones(t *testing.T) {
+	repoStore := repoStore(t)
+	registryStore := registryStore(t)
+	tests := []struct {
+		name     string
+		filePath string
+		content  string
+		want     *drivers.CatalogEntry
```
egor-ryashin, line 643: I would name `expected`

egor-ryashin (approved):

runtime/services/catalog/artifacts/artifacts_test.go
begelundmuller: Just copied it from the other test cases :)

PR #2751, opened 2023-07-06
Percent of total in leaderboards

bcolloran:
add percent of total to leaderboards

PR #2768, opened 2023-07-12
Percent of total dimension table

bcolloran:
Add percent of total in dimension table.

Follow up to https://github.com/rilldata/rill/pull/2751 -- recommend merging that before this, and then reviewing this, since this builds off changes in that PR it will make the review much simpler.

remember that per the spec, you will need to add valid_percent_of_total: true to any measures for which you want to see the percent of total

PRD: https://www.notion.so/rilldata/Display-percent-of-total-metrics-in-dashboards-3764999f8d0d41c7832484d452aef56a


djbarnwal (approved):

jkhwu:
@djbarnwal , the idea of the dimension table that we discussed during design reviews is that all columns would be shown, regardless of toggle status on the leaderboard view. The toggle is only for the leaderboards because there's only one context column and they can choose what they want to populate it with.

I agree that the tooltip wording is not super clear, do you have an alternate proposal that might make more sense to the user? Would it make sense to direct them to change the dashboard yaml and add `valid_percent_of_total: true`?

Issue #2758, opened 2023-07-10
Add timezone support to Dashboard


djbarnwal:
Detailed discussion - https://www.notion.so/rilldata/Support-timezone-detection-selection-and-display-in-dashboards-e30fdbc8593149be8684449b5d272336?d=e3b14f8e6df64f63bb1400245efe4d1e#901409b0829e483a9f6bb2801f0c8512

Tasks - 

-  Add timezone selector to dashboard menu
    - [x] Add UI selector as per mocks
    - [x] Sort based on user locale and offsets
    - [x] Add abbreviation mappings for timezones
- [x] Add timezone state and set defaults based on user locale
- [x] Integrate `dashboard.yaml`/ timezone yaml config to UI timezone state
- [x] Update time range preset transforms based on the zone
- [x] Update `prettyFormatTimeRange` to use proper dates
- [x] Update TopK and Totals analytical API calls
- [x] Enable support for timezone offsets in chart elements
- [x] Rewire Timeseries API call
- [x] Update chart bounds in accordance to the timezone
- [x] Enable support for timezone in custom calendar selector
- [x] Update chart labels (axes and mouseovers)
- [x] Add timezone iana string to URL proto
- [x] Save user preference for Rill Developer
- [ ] Save user preference for Cloud dashboards
- [x] Write and update tests for each of the above methods
- [ ] Update DateTime formatter to support timezone for Rows viewer (_To be updated later based on new table architecture_)

PR #2775, opened 2023-07-14
Add timezone support to dashboard

djbarnwal:
Add timezone support to Dashboard

List of tasks -  #2758


web-common/src/features/metrics-views/metrics-internal-store.ts
```diff
@@ -47,7 +48,15 @@ timeseries: ""
 measures:
   - label: "Total Records"
     expression: "count(*)"
-dimensions: []
+dimensions:
```
AdityaHegde, line 51: Whats the reason for adding a dummy dimension?

AdityaHegde (approved):
Good to merge after fixing the small nits.

web-common/src/features/metrics-views/metrics-internal-store.ts
djbarnwal: Added this because the current semantic of having `[]` resulted in an error in the YAML editor. Also you don't know what are the keys you need to put in the `dimensions`. Now the first error the user sees is for wrong model name, followed by wrong column name which should give a path to the user to have the right spec.

web-common/src/features/dashboards/time-controls/TimeControls.svelte
```diff
@@ -40,15 +46,24 @@
   import TimeComparisonSelector from "./TimeComparisonSelector.svelte";
   import TimeGrainSelector from "./TimeGrainSelector.svelte";
   import TimeRangeSelector from "./TimeRangeSelector.svelte";
+  import TimeZoneSelector from "./TimeZoneSelector.svelte";
 
   export let metricViewName: string;
 
+  const localUserPreferences = localStorageStore<LocalUserPreferences>(
+    `${metricViewName}-userPreference`,
```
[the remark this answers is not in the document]
djbarnwal: Have moved out `userPreference` into it's own file. We need to create a `dashboard-store.ts` like store for local user preference as well. This is out of scope for this PR but we should come back on this later.

PR #2776, opened 2023-07-14
Leaderboard code cleanups

bcolloran:
see PRD in notion https://www.notion.so/rilldata/Leaderboard-refactors-313ed03e84644519b97fbf9f57612adf


bcolloran:
@nishantmonu51 this PR is ready for code review (requesting from @AdityaHegde ), though I will want to go through some more time on QA tests vs main before merging to make sure there are no subtle style regressions, and I may do a tiny bit more merging of styles since i think there is still some confusing redundancy.

Notion has [the PRD](https://www.notion.so/rilldata/Leaderboard-refactors-313ed03e84644519b97fbf9f57612adf), which describes most of what I set out to do. As planned, the leaderboard component code is dramatically simpler. Additional bits not mentioned in the PRD:
- I split out a component for the TopKListItem. A lot of the complexity in the leader board components boiled down to the fact that we were attempting to use a generic component for two very different use cases. This lead to a tremendous amount of extra work to get this one component to serve the needs of both the top-k lists and the leaderboard lists.



@skokenes -- I've been hitting tons of random test failures as you can see from the red Xs and green checks above. I cannot reproduce these failures locally, and have always been able to clear them by rerunning the tests (though I have not bothered to do so in all cases, hence the red Xs). Seems to be a timeout issue. I don't think my changes in this PR should have caused any queries to run more slowly or anything like that, and I don't notice any performance degradation locally, but I still only have a basic familiarity with the e2e tests, so I wanted to flag this for you in case you have any ideas or insight.

AdityaHegde (approved):
Looks good code wise. Didnt find any regressions in my limited QA.

PR #2779, opened 2023-07-17
Fix race condition in query batch test

AdityaHegde:
The code that collects the responses in query batch test was not thread safe. Adding a mutex to make sure when responses are returned at the same doesnt result in a panic.


begelundmuller (approved):
Looks good. Let's merge and see if it works :)

PR #2780, opened 2023-07-17
adding prod ttl to project details

rakeshsharma14317:


cli/cmd/sudo/project/get.go
```diff
@@ -44,6 +44,7 @@ func GetCmd(cfg *config.Config) *cobra.Command {
 			fmt.Printf("Prod OLAP DSN: %s\n", project.ProdOlapDsn)
 			fmt.Printf("Prod slots: %d\n", project.ProdSlots)
 			fmt.Printf("Prod deployment ID: %s\n", project.ProdDeploymentId)
+			fmt.Printf("Prod TTL: %d\n", project.ProdTtlSeconds)
```
begelundmuller, line 47: Update to `Prod hibernation TTL: %s` and then use the `time` package to format it as a human-readably number (instead of seconds)

begelundmuller (changes requested):

admin/server/projects.go
```diff
@@ -765,6 +765,7 @@ func (s *Server) projToDTO(p *database.Project, orgName string) *adminv1.Project
 		Subpath:          p.Subpath,
 		GithubUrl:        safeStr(p.GithubURL),
 		ProdDeploymentId: safeStr(p.ProdDeploymentID),
+		ProdTtlSeconds:   *p.ProdTTLSeconds,
```
begelundmuller, line 768: This pointer dereference will panic if the pointer is nil

begelundmuller (changes requested):

begelundmuller (approved):

PR #2782, opened 2023-07-17
Async reconcile controller APIs

begelundmuller:
This PR contains stubs for the proposed new APIs for async reconcile. They should replace all the existing catalog and reconcile APIs (which will be deprecated, but kept for backwards compatibility until the migration is complete).

The basic idea of these APIs is a) to decouple file saving from reconcile and errors, b) to support editing in another app (like VS Code) and instantly see changes and errors in Rill Developer. This will happen through long-running RPCs with streaming responses that deliver updates to the UI.

Key notes:

- The PR builds on the resource refactors in https://github.com/rilldata/rill/pull/2761. Here's a brief recap:
  - The word "resource" replaces the old words "catalog entry" and "catalog object"
  - A resource can be a source, model, metrics view, etc.
  - A resource has separate `spec` (desired state) and `state` (current state) fields
- Introduces a `WatchResources` RPC which will stream a snapshot of a resource after its spec and/or state is updated
- Introduces a `WatchLogs` RPC which will stream logs from reconcile (source ingestion, etc.) to the frontend
- Introduces a generic `CreateTrigger` API for creating any kind of trigger (such as reconcile, source refresh, and in the future, things like manual run of scheduled reports, etc.)
  - However, the frontend will have less need for this, since reconcile will directly watch the file system for file changes
- In addition to the new APIs in this PR, we are also adding an API for watching for file changes in this PR: https://github.com/rilldata/rill/pull/2753

Proposed new frontend implementation of the modeller:

- Globally, call `WatchFiles` and `WatchResources` and build a cached state of the files and resources in the project
- When creating/deleting a file, just call the relevant file RPC, and wait for WatchFiles to invalidate the global file cache
- When editing a file, just call `PutFile` – then rely on the runtime to propagate updates over `WatchResources`
  - The runtime will add errors, progress information, etc. in a resource's `state`
  - The runtime adds source file path metadata to resources, so they can be coupled to specific files in the UI


AdityaHegde:
This new approach feels cleaner. All mutations would one do one thing, parsing reconcile response is moved to a single place. Especially the modeller code, a lot of code is simplified.

Some drawbacks I can think ok,
1. There is some additional global context needed to be added to support something like `Create item => Navigate to created item`. Perhaps we can add a timer to navigate once the file shows up in the `WatchFiles`. But still adds some ugliness.
2. There is a possibility that the `Resource` will be out of sync with `File`. We need to be careful to handle this properly in our editors. Like what happens when quickly typing something and errors from intermediate reconciles show up in part of sql/yaml that no longer exists? This might be for a short period (milliseconds perhaps) but we need to figure out UX for this.
3. None of the streaming APIs are supported by tanstack-query out of the box. We can use a custom query function and hide away the list of files and list of resources. New updates can use `setQueryData` to update the list.

begelundmuller:
> 1. There is some additional global context needed to be added to support something like `Create item => Navigate to created item`. Perhaps we can add a timer to navigate once the file shows up in the `WatchFiles`. But still adds some ugliness.

`PutFile` would not return until the file has actually been created (even if the event hasn't been sent over `WatchFiles` yet). Would that be sufficient?

Otherwise, for `PutFile` specifically, we could have it only return once any relevant `WatchFiles` events have been emitted (this is slightly weird on the backend, but not very difficult).

One thing to note – the `Watch` RPCs would be very fast and stable on `localhost`. So perhaps just waiting for the event would not be noticeable to the user. I think only in a hosted environment would there be odd latencies, but the streaming RPCs wouldn't be needed on cloud.

> 2. There is a possibility that the `Resource` will be out of sync with `File`. We need to be careful to handle this properly in our editors. Like what happens when quickly typing something and errors from intermediate reconciles show up in part of sql/yaml that no longer exists? This might be for a short period (milliseconds perhaps) but we need to figure out UX for this.

This is a good point. I think in practice, this could probably be solved with tiny batching delays (maybe ~100ms) in either triggering reconcile (backend), sending update events (backend), or rendering events (frontend). 

> 3. None of the streaming APIs are supported by tanstack-query out of the box. We can use a custom query function and hide away the list of files and list of resources. New updates can use `setQueryData` to update the list.

That sounds good. Just a question – what if you just had a Svelte store containing a map, and an async function reading data from the streaming RPC and updating the map in the Svelte store? And then any component using the map would be re-rendered on update?

BTW, all resources will have a `version` number on them that gets incremented whenever there's a change to it. You could incorporate the version into cache keys to automatically have queries get invalidated when a resource is updated.

begelundmuller:
Closing in favor of https://github.com/rilldata/rill/pull/2786

Issue #2750, opened 2023-07-06
Docs: Reference docs for dashboard YAML don't include `default_time_range` or `smallest_time_grain`


ericpgreen2:
See https://docs.rilldata.com/reference/project-files/dashboards

PR #2783, opened 2023-07-17
Update dashboard YAML reference docs

begelundmuller:
Updates the dashboard YAML reference docs to reflect the YAML properties currently supported by the backend

Closes https://github.com/rilldata/rill/issues/2750


hamilton (approved):
thanks for the update here!

PR #2786, opened 2023-07-18
Async reconcilers for sources, models, metrics, migrations

begelundmuller:
This PR implements the async reconcilers, but the functions they rely on from the controller have not yet been implemented. The idea was to let the async reconciler implementations guide the functionality required from the controller.

This means that while the code compiles, it can't yet be tested. The final step before testing is to implement the `Controller`, which will contain the DAG-based scheduler (event loop).

In terms of implementation details, two noteworthy details:

- In `runtime/reconcilers/README.md` I described some principles for async reconciler development
- The `ProjectParser` resource and accompanying reconciler in `runtime/reconcilers/project_parser.go` is a "meta"-resource that will be created at project creation time (not from a code artifact). It's purpose is to continuously parse the project and create/update/delete resources based on the parser output.


AdityaHegde (approved):
Looks good at a high level. We can address possible edge cases before completely switching over to this reconcile.

PR #2795, opened 2023-07-20
2736 admin implement service accounts

rakeshsharma14317:


admin/database/database.go
```diff
@@ -363,6 +376,27 @@ type UpdateUserOptions struct {
 	PreferenceTimeZone  string
 }
 
+// Service represents a service account.
+// Service accounts may belong to single organization
+type Service struct {
+	ID        string
+	OrgName   string    `db:"org_name"`
```
begelundmuller, line 383: Should be `OrgID`

admin/server/auth/claims.go
```diff
@@ -148,7 +148,7 @@ func (c *authTokenClaims) ProjectPermissions(ctx context.Context, orgID, project
 	case authtoken.TypeUser:
 		// continue
 	case authtoken.TypeService:
-		panic(errors.New("service tokens not supported"))
+		// continue
```
begelundmuller, line 151: Same issue as above

admin/server/service.go
```diff
@@ -0,0 +1,280 @@
...
+	"go.opentelemetry.io/otel/attribute"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// CreateService creates a new service account.
+func (s *Server) CreateService(ctx context.Context, req *adminv1.CreateServiceRequest) (*adminv1.CreateServiceResponse, error) {
+	observability.AddRequestAttributes(ctx,
+		attribute.String("args.name", req.Name),
+		attribute.String("args.OrganizationName", req.OrganizationName),
```
begelundmuller, line 19: The attribute should be snake case

admin/server/service.go
```diff
@@ -0,0 +1,280 @@
...
+	)
+
+	// Check the request is made by an authenticated user
+	claims := auth.GetClaims(ctx)
+	if claims.OwnerType() != auth.OwnerTypeUser {
+		return nil, status.Error(codes.Unauthenticated, "not authenticated as a user")
+	}
```
begelundmuller, line 26: No need to check the call is made by a user – that's only needed when `claims.OwnerID()` is assumed to be a user ID

admin/server/service.go
```diff
@@ -0,0 +1,280 @@
...
+	return &adminv1.CreateServiceResponse{
+		Service: serviceToPB(service),
+	}, nil
+}
+
+// ListServices lists all service accounts in an organization.
+func (s *Server) ListServices(ctx context.Context, req *adminv1.ListServicesRequest) (*adminv1.ListServicesResponse, error) {
+	observability.AddRequestAttributes(ctx,
+		attribute.String("args.org", req.OrganizationName),
+		attribute.String("args.OrganizationName", req.OrganizationName),
```
begelundmuller, line 51: snake case

admin/server/service.go
```diff
@@ -0,0 +1,280 @@
...
+	)
+
+	// Check the request is made by an authenticated user
+	claims := auth.GetClaims(ctx)
+	if claims.OwnerType() != auth.OwnerTypeUser {
+		return nil, status.Error(codes.Unauthenticated, "not authenticated as a user")
+	}
```
begelundmuller, line 58: See above

admin/server/service.go
```diff
@@ -0,0 +1,280 @@
...
+		Services: servicesToPB(services),
+	}, nil
+}
+
+// UpdateService updates a service account.
+func (s *Server) UpdateService(ctx context.Context, req *adminv1.UpdateServiceRequest) (*adminv1.UpdateServiceResponse, error) {
+	observability.AddRequestAttributes(ctx,
+		attribute.String("args.name", req.Name),
+		attribute.String("args.OrganizationName", req.OrganizationName),
+	)
```
begelundmuller, line 85: snake case

admin/server/service.go
```diff
@@ -0,0 +1,280 @@
...
+
+	if req.NewName != nil {
+		observability.AddRequestAttributes(ctx, attribute.String("args.new_name", *req.NewName))
+	}
+
+	// Check the request is made by an authenticated user
+	claims := auth.GetClaims(ctx)
+	if claims.OwnerType() != auth.OwnerTypeUser {
+		return nil, status.Error(codes.Unauthenticated, "not authenticated as a user")
+	}
```
begelundmuller, line 95: See above

admin/server/service.go
```diff
@@ -0,0 +1,280 @@
...
+		Service: serviceToPB(updatedService),
+	}, nil
+}
+
+// DeleteService deletes a service account.
+func (s *Server) DeleteService(ctx context.Context, req *adminv1.DeleteServiceRequest) (*adminv1.DeleteServiceResponse, error) {
+	observability.AddRequestAttributes(ctx,
+		attribute.String("args.name", req.Name),
+		attribute.String("args.OrganizationName", req.OrganizationName),
+	)
```
begelundmuller, line 129: snake case

admin/server/service.go
```diff
@@ -0,0 +1,280 @@
...
+	service, err := s.admin.DB.FindServiceByName(ctx, req.OrganizationName, req.Name)
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+
+	// Check the request is made by an authenticated user
+	claims := auth.GetClaims(ctx)
+	if claims.OwnerType() != auth.OwnerTypeUser {
+		return nil, status.Error(codes.Unauthenticated, "not authenticated as a user")
+	}
```
begelundmuller, line 140: See above

admin/server/service.go
```diff
@@ -0,0 +1,280 @@
...
+	observability.AddRequestAttributes(ctx,
+		attribute.String("args.service_name", req.ServiceName),
+		attribute.String("args.organization_name", req.OrganizationName),
+	)
+
+	// Check the request is made by an authenticated user
+	claims := auth.GetClaims(ctx)
+	if claims.OwnerType() != auth.OwnerTypeUser {
+		return nil, status.Error(codes.Unauthenticated, "not authenticated as a user")
+	}
```
begelundmuller, line 171: See above

admin/server/service.go
```diff
@@ -0,0 +1,280 @@
...
```
begelundmuller, line 214: See above

admin/server/service.go
```diff
@@ -0,0 +1,280 @@
...
+func (s *Server) RevokeServiceAuthToken(ctx context.Context, req *adminv1.RevokeServiceAuthTokenRequest) (*adminv1.RevokeServiceAuthTokenResponse, error) {
+	observability.AddRequestAttributes(ctx,
+		attribute.String("args.token_id", req.TokenId),
+	)
+
+	// Check the request is made by an authenticated user
+	claims := auth.GetClaims(ctx)
+	if claims.OwnerType() != auth.OwnerTypeUser {
+		return nil, status.Error(codes.Unauthenticated, "not authenticated as a user")
+	}
```
begelundmuller, line 251: See above

admin/server/service.go
```diff
@@ -0,0 +1,280 @@
...
+
+	// Need a check for service permission to revoke a token
+	// if !claims.OrganizationPermissions(ctx, org.ID).ManageOrg {
+	// 	return nil, status.Error(codes.PermissionDenied, "not allowed to update org")
+	// }
```
begelundmuller, line 256: Checking `ManageOrg` is good for now

admin/service.go
```diff
@@ -0,0 +1,34 @@
...
+
+	err = tx.Commit()
+	if err != nil {
+		return nil, err
+	}
+
+	s.logger.Info("created service", zap.String("name", serviceName), zap.String("org_name", orgName))
+
+	return service, nil
+}
```
begelundmuller, line 34: The transaction is not needed when only one DB call is made.

This func can probably be omitted.

The call to `s.DB.InsertService` can be inlined in the API handler.

cli/cmd/service/token/issue.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/spf13/cobra"
+)
+
+func IssueCmd(cfg *config.Config) *cobra.Command {
+	var name string
+	issueCmd := &cobra.Command{
+		Use:   "issue [service-name]",
+		Args:  cobra.NoArgs,
+		Short: "Issue service token",
+		RunE: func(cmd *cobra.Command, args []string) error {
```
begelundmuller, line 19: Probably we should use the positional arg, like in the `rill service` commands? Or allow both flag and positional?

cli/cmd/service/token/issue.go
```diff
@@ -0,0 +1,53 @@
...
+			// Set new access token
+			err = dotrill.SetAccessToken(res.Token)
+			if err != nil {
+				return err
+			}
+
+			// set the default token to the one we just got
+			cfg.AdminTokenDefault = res.Token
+			cmdutil.PrintlnSuccess("Issued token")
+			fmt.Println("token is", res.Token)
```
begelundmuller, line 43: It should not set service tokens as the user's token. It should just print the token in the CLI

proto/rill/admin/v1/api.proto
```diff
@@ -202,6 +202,24 @@ service AdminService {
...
+    option (google.api.http) = {
+      post: "/v1/tokens/service",
+      body: "*"
+    };
+  }
+
+  // RevokeServiceAuthToken revoke the service auth token
+  rpc RevokeServiceAuthToken(RevokeServiceAuthTokenRequest) returns (RevokeServiceAuthTokenResponse) {
+    option (google.api.http) = {delete: "/v1/tokens/service"};
+  }
```
begelundmuller, line 221: 1. Move them to after the service-related APIs
2. Use the same paths as the services – like `/v1/organizations/{organization_name}/services/{name}/tokens`

proto/rill/admin/v1/api.proto
```diff
@@ -733,6 +812,30 @@ message IssueRepresentativeAuthTokenResponse {
...
+}
+
+message ListServiceAuthTokensRequest {
+  string organization_name = 1 [(validate.rules).string.min_len = 1];
+  string service_name = 2;
+}
+
+message ListServiceAuthTokensResponse {
+  repeated string tokens = 1;
+}
```
begelundmuller, line 837: Should return a `ServiceToken` type containing id, created_on, and expires_on. Then we can print these values

proto/rill/admin/v1/api.proto
```diff
@@ -789,6 +892,15 @@ message User {
   google.protobuf.Timestamp updated_on = 7;
 }
 
+message Service {
+  string id = 1;
+  string service_name = 2;
```
begelundmuller, line 897: Just use `name` for `service_name` (like in `Organization` and `Project`)

proto/rill/admin/v1/api.proto
```diff
@@ -789,6 +892,15 @@ message User {
...
+  string org_name = 3;
```
begelundmuller, line 898: Return both `org_id` and `org_name`, like in `Project`

proto/rill/admin/v1/api.proto
```diff
@@ -789,6 +892,15 @@ message User {
...
+  // string role = 4;
```
begelundmuller, line 899: Can remove comment

admin/server/auth/claims.go
```diff
@@ -117,7 +117,7 @@ func (c *authTokenClaims) OrganizationPermissions(ctx context.Context, orgID str
...
```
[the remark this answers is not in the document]
begelundmuller: Instead, the implementation of `OrganizationPermissions` probably needs to be refactored to forward the call into two internal functions `organizationPermissionsUser` and `organizationPermissionsService` where the individual implementations are placed.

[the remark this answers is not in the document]
begelundmuller: For the initial implementation of the service permissions, I would propose a service should just have full permissions for the organization it belongs to, as well as full permissions for any project that belongs to the organization.

cli/cmd/service/token/issue.go
rakeshsharma14317: If we keep both, then if nothing is provided then it will take the default value of flag and result accordingly, keeping both for will update as per user experience

admin/database/database.go
```diff
@@ -399,6 +433,23 @@ type InsertUserAuthTokenOptions struct {
...
+	ServiceID  string     `db:"service_id"`
+	CreatedOn  time.Time  `db:"created_on"`
+	ExpiresOn  *time.Time `db:"expires_on"`
+}
+
+// InsertServiceAuthTokenOptions defines options for creating a ServiceAuthToken.
+type InsertServiceAuthTokenOptions struct {
+	ID         string
+	SecretHash []byte
+	ServieID   string
```
begelundmuller, line 449: Typo

admin/database/postgres/migrations/0011.sql
```diff
@@ -0,0 +1,17 @@
+CREATE TABLE service (
+	id UUID DEFAULT uuid_generate_v4() PRIMARY KEY,
+	org_id UUID NOT NULL REFERENCES orgs (id) ON DELETE RESTRICT,
```
begelundmuller, line 3: Should be `ON DELETE CASCADE`

admin/database/postgres/migrations/0011.sql
```diff
@@ -0,0 +1,17 @@
...
+	name TEXT NOT NULL,
```
begelundmuller, line 4: Need to add a unique index `(org_id, lower(name))`, see migration for `projects` table for example

admin/pkg/authtoken/authtoken.go
```diff
@@ -58,6 +58,20 @@ func NewRandom(t Type) *Token {
...
+		Type:   t,
+		ID:     id,
+		Secret: secret,
+	}
+}
+
+// TokenString returns the token as a string.
+func (t *Token) TokenString() string {
+	return fmt.Sprintf("%s_%s_%s", Prefix, t.Type, marshalBase62(append(t.ID[:], t.Secret[:]...)))
+}
```
begelundmuller, line 73: Seems like it's not used anywhere?

admin/server/auth/claims.go
```diff
@@ -247,3 +190,112 @@ func unionProjectRoles(a *adminv1.ProjectPermissions, b *database.ProjectRole) *
...
 }
+
+// organizationPermissionsUser is a Claims implementation for resolve the organization permissions for user
+func (c *authTokenClaims) organizationPermissionsUser(ctx context.Context, orgID string) *adminv1.OrganizationPermissions {
+	c.Lock()
+	defer c.Unlock()
+
+	if perm, ok := c.orgPermissionsCache[orgID]; ok {
+		return perm
+	}
```
begelundmuller, line 201: If you move this back into `OrganizationPermissions`, the caching will work for both service and user tokens

admin/server/auth/claims.go
```diff
@@ -247,3 +190,112 @@ func unionProjectRoles(a *adminv1.ProjectPermissions, b *database.ProjectRole) *
...
+			ManageProjectMembers: true,
+		}
+	}
+
+	c.Lock()
+	defer c.Unlock()
+
+	if perm, ok := c.projectPermissionsCache[projectID]; ok {
+		return perm
+	}
```
begelundmuller, line 264: Same note, can be moved to common function so it works for services and users tokens

admin/server/auth/claims.go
```diff
@@ -247,3 +190,112 @@ func unionProjectRoles(a *adminv1.ProjectPermissions, b *database.ProjectRole) *
...
+			ReadOrg:          true,
+			ManageOrg:        true,
+			ReadProjects:     true,
+			CreateProjects:   true,
+			ManageProjects:   true,
+			ReadOrgMembers:   true,
+			ManageOrgMembers: true,
+		}
+	}
+	return orgPermissions
```
begelundmuller, line 237: Do a `return` on line 227 and then on line 237 do `return &adminv1.OrganizationPermissions{}` so that the return value is never `nil`

cli/cmd/service/rename.go
```diff
@@ -0,0 +1,43 @@
...
+			if err != nil {
+				return err
+			}
+			defer client.Close()
+
+			res, err := client.UpdateService(cmd.Context(), &adminv1.UpdateServiceRequest{
+				Name:             args[0],
+				OrganizationName: cfg.Org,
+				NewName:          &newName,
+			})
```
begelundmuller, line 28: Must check that `newName` is not empty

begelundmuller (changes requested):

admin/pkg/authtoken/authtoken.go
rakeshsharma14317: Yeah just added and testing it out for ListTokens API, not needed now, removing

begelundmuller (approved):

PR #2807, opened 2023-07-21
Sql sources prototype

AdityaHegde:
Prototype to support sql source. Supports any free form query that references a single url source.
EG:
```
type: "duckdb"
query: |
  select * exclude(publisher),
  (case when publisher = 'Yahoo' then 0 when publisher = 'Google' then 1 else 2 end) as pub,
  from read_csv_auto('/path/to/AdBids.csv') where publisher in ('Yahoo', 'Google')
```


begelundmuller (commented):
Thanks for the prototype here! 

Instead of the complexity of parsing the SQL into source parameters, and need of changing the schema relaxation logic (which is already pretty complex), what about just rewriting the path value for S3/GCS sources to the local downloaded paths, but otherwise just run the raw SQL provided?

I'm picturing something like this:

1. During source parsing, use `duckdbsql` to detect if it references S3 or GCS paths
    a. If yes, set the connector to "s3" or "gcs", rewrite the path in the SQL to a `?`, and then add the resulting "sql" as a source property. 
    b. If no, set the connector as "duckdb" and add "sql" as a property.
2. Add a "duckdb_to_duckdb" transporter that just gets the "sql" property and creates it as a table (`CREATE TABLE xxx AS (%s)`)
3. In the S3 and GCS transporters, if the "sql" property is present, also create the table using the raw SQL (`CREATE TABLE xxx AS (%s)`) and pass the downloaded paths as a query arg.

runtime/drivers/duckdb/transporter/objectStore_to_duckDB.go
```diff
@@ -260,3 +267,53 @@ func (a *appender) scanSchemaFromQuery(ctx context.Context, qry string) (map[str
...
+	p drivers.Progress,
+) error {
+	allFiles := make([]string, 0)
+	for iterator.HasNext() {
+		files, err := iterator.NextBatch(opts.IteratorBatch)
+		if err != nil {
+			return err
+		}
+		allFiles = append(allFiles, files...)
+	}
```
begelundmuller, line 286: I think the iterator deletes all the files from the previous iteration: https://github.com/rilldata/rill/blob/7e19861b2790db8d7ace6e6b4ebeaf1eb7f09ff8/runtime/drivers/blob/blobdownloader.go#L167C33-L167C33

Might need to add an iterator option to retain files until the iterator is closed.

runtime/drivers/duckdb/transporter/objectStore_to_duckDB.go
```diff
@@ -260,3 +267,53 @@ func (a *appender) scanSchemaFromQuery(ctx context.Context, qry string) (map[str
...
+
+	ast, err := duckdbsql.Parse(originalSQL)
+	if err != nil {
+		return err
+	}
+
+	err = ast.RewriteTableRefs(func(table *duckdbsql.TableRef) (*duckdbsql.TableRef, bool) {
```
begelundmuller, line 293: Maybe put a comment that upstream logic is expected to guarantee that there's only one table ref in the SQL

begelundmuller: Or safer – add a check here that returns an error if more than one table reference was matched

runtime/services/catalog/artifacts/yaml/objects.go
```diff
@@ -14,9 +16,6 @@ import (
 	"github.com/rilldata/rill/runtime/pkg/duration"
 	"github.com/rilldata/rill/runtime/pkg/fileutil"
 	"google.golang.org/protobuf/types/known/structpb"
-
-	// Load IANA time zone data
-	_ "time/tzdata"
```
begelundmuller, line 19: Restore this formatting – not sure why it didn't trigger, but there's supposed to be a CI rule to have blank imports in the third group

begelundmuller (changes requested):

runtime/services/catalog/artifacts/yaml/objects.go
begelundmuller: Haha, looks like it did trigger

runtime/drivers/duckdb/transporter/objectStore_to_duckDB.go
```diff
@@ -260,3 +268,69 @@ func (a *appender) scanSchemaFromQuery(ctx context.Context, qry string) (map[str
...
+		return err
+	}
+
+	// Validate the sql is supported for sources
+	// TODO: find a good common place for this validation and avoid code duplication here and in sources packages as well
+	refs := ast.GetTableRefs()
+	if len(refs) > 1 {
+		return errors.New("sql source can have only one table reference")
+	}
+	ref := refs[0]
```
begelundmuller, line 301: For safety, also check it's not empty (which would panic on the next line)

begelundmuller (approved):

PR #2813, opened 2023-07-24
Batch file events and replay added directories

begelundmuller:


runtime/server/repos.go
```diff
@@ -57,16 +58,34 @@ func (s *Server) WatchFiles(req *runtimev1.WatchFilesRequest, ss runtimev1.Runti
...
+				return err
+			}
+		}
+	}
 
-			return err
+	return repo.Watch(ss.Context(), "", func(events []drivers.WatchEvent) {
+		for _, event := range events {
+			if !event.Dir {
+				err := ss.Send(&runtimev1.WatchFilesResponse{
```
egor-ryashin, line 80: should we add batches to the response too?

runtime/drivers/file/watcher.go
```diff
@@ -0,0 +1,224 @@
...
+
+	select {
+	case <-w.done:
+		// Already closed
+		return
+	default:
+	}
+
+	closeErr := w.watcher.Close()
+	if err != nil {
```
egor-ryashin, line 69: should we do `errors.Join` here?

egor-ryashin (changes requested):

runtime/drivers/file/watcher.go
```diff
@@ -0,0 +1,224 @@
...
+}
+
+func (w *watcher) run() {
+	err := w.runInner()
+	w.closeWithErr(err)
+}
+
+func (w *watcher) runInner() error {
+	timer := time.NewTimer(batchInterval)
+	timerActive := true
```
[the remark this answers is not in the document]
begelundmuller: Good catch, yeah it seems like it – made the change

runtime/drivers/file/watcher.go
```diff
@@ -0,0 +1,224 @@
...
+	for {
+		select {
+		case <-timer.C:
+			timerActive = false
+			w.flush()
+		case err, ok := <-w.watcher.Errors:
+			if !ok {
```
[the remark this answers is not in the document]
begelundmuller: Decided not to flush in error scenarios since something with the underlying file access might have gone wrong, so we don't want to trigger actions against the buffered paths

runtime/drivers/file/watcher.go
```diff
@@ -0,0 +1,224 @@
...
+				return nil
+			}
+			return err
+		case e, ok := <-w.watcher.Events:
+			if !ok {
+				return nil
```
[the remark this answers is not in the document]
begelundmuller: This case should only happen if the watcher is closed – so it would be weird to trigger the callback in such cases. If the watcher is closed, the caller should not expect more callbacks.

runtime/server/repos.go
begelundmuller: It might make sense, depends what the frontend team would prefer. Will ask them and then we can make the change in a follow up PR if needed

runtime/drivers/file/watcher.go
```diff
@@ -0,0 +1,218 @@
...
+
+			// Calling addDir after appending to w.buffer, to sequence events correctly
+			if we.Dir && e.Has(fsnotify.Create) {
+				err = w.addDir(e.Name, true)
+				if err != nil {
+					return err
+				}
+			}
+
+			ticker.Reset(batchInterval)
```
egor-ryashin, line 173: The buffer can grow indefinitely in case there's a constant stream of changes with events coming within 250ms.

runtime/drivers/file/watcher.go
```diff
@@ -0,0 +1,218 @@
...
+	err := w.runInner()
+	w.closeWithErr(err)
+}
+
+func (w *watcher) runInner() error {
+	ticker := time.NewTicker(batchInterval)
+	for {
+		select {
+		case <-ticker.C:
+			ticker.Stop()
```
egor-ryashin, line 130: I cannot find where ticker is restarted again?

egor-ryashin (changes requested):

egor-ryashin (approved):

PR #2814, opened 2023-07-24
Compiler: check refs and unique source/model names

begelundmuller:
This PR contains some minor improvements to the project parser for async reconcile

- Make sure a resource's `Refs` only contains valid refs
  - If a ref does not explicitly provide a kind (like "source" or "model"), we try to infer it
  - If no kind can be inferred, the ref is discarded (see comments in PR for details)
- Add check that a source and a model does not use the same name
  - The new catalog supports different resource kinds having the same name, but we need this exception to apply for sources/models since they are both mapped to tables)
- Track modifications to `rill.yaml` in `Diff`

### DIFF
--- a/runtime/compilers/rillv1/parser.go
+++ b/runtime/compilers/rillv1/parser.go
@@ -27,9 +27,10 @@ var ErrInvalidProject = errors.New("parser: not a valid project (rill.yaml not f
 // One file may output multiple resources and multiple files may contribute config to one resource.
 type Resource struct {
 	// Metadata
-	Name  ResourceName
-	Refs  []ResourceName
-	Paths []string
+	Name    ResourceName
+	Paths   []string
+	Refs    []ResourceName // Derived from rawRefs after parsing (can't contain ResourceKindUnspecified)
+	rawRefs []ResourceName // Populated during parsing (may contain ResourceKindUnspecified)
 
 	// Only one of these will be non-nil
 	SourceSpec      *runtimev1.SourceSpec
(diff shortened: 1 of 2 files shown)


runtime/compilers/rillv1/parser.go
```diff
@@ -259,14 +273,52 @@ func (p *Parser) Reparse(ctx context.Context, paths []string) (*Diff, error) {
...
+	modifiedRillYAML := p.RillYAML == nil
+
 	// Phase 2: Parse (or reparse) the related paths, adding back resources
 	err := p.parsePaths(ctx, parsePaths)
 	if err != nil {
 		return nil, err
 	}
 
+	// Infer or re-infer refs for...
+	inferRefsSeen := make(map[ResourceName]bool)
```
AdityaHegde, line 286: nit: perhaps move inference this to a separate method?

runtime/compilers/rillv1/parser.go
```diff
@@ -345,6 +393,27 @@ func (p *Parser) parsePaths(ctx context.Context, paths []string) error {
 		return ErrInvalidProject
 	}
 
+	// As a special case, we need to check that there aren't any sources and models with the same name.
+	// NOTE 1: We always attach the error to the model when there's a collision.
+	// NOTE 2: Using a map since the two-way check (necessary for reparses) may match the same resource twice.
+	modelsWithNameErrs := make(map[ResourceName]string)
```
AdityaHegde, line 399: I cant find the place where this check is happening for the same `kind/name` with different stem. Or do we not allow different stems to have same name be different resource?

AdityaHegde (approved):
minor nit and an unrelated question. LGTM

runtime/compilers/rillv1/parser.go
begelundmuller: Thanks, it's only used in Reparse and also depends on a variable in its scope, so keeping it like this for now

runtime/compilers/rillv1/parser.go
begelundmuller: Discussed offline, will probably want to make some refactors to avoid this in a separate PR

PR #2817, opened 2023-07-24
`range` doesn't take into account time zone DST

egor-ryashin:
For example:
```
D select time_bucket(interval '1 week', timestamptz '2022-10-25', 'Europe/Copenhagen') as week_start_copenhagen;
┌──────────────────────────┐
│  week_start_copenhagen   │
│ timestamp with time zone │
├──────────────────────────┤
│ 2022-10-23 22:00:00+00   │
└──────────────────────────┘
D select time_bucket(interval '1 week', timestamptz '2022-11-01', 'Europe/Copenhagen') as week_start_copenhagen;
┌──────────────────────────┐
│  week_start_copenhagen   │
│ timestamp with time zone │
├──────────────────────────┤
│ 2022-10-30 23:00:00+00   │
└──────────────────────────┘
D select * from range(time_bucket(interval '1 week', timestamptz '2022-10-25', 'Europe/Copenhagen'), time_bucket(interval '1 week', timestamptz '2022-11-01', 'Europe/Copenhagen'), interval '1 week') ;
┌──────────────────────────┐
│          range           │
│ timestamp with time zone │
├──────────────────────────┤
│ 2022-10-23 22:00:00+00   │
│ 2022-10-30 22:00:00+00   │
└──────────────────────────┘
```
1st and 2nd queries show different week starts in terms of the hour shift, 3rd shows the same for both weeks.

### DIFF
--- a/runtime/queries/column_timeseries.go
+++ b/runtime/queries/column_timeseries.go
@@ -127,13 +127,13 @@ func (q *ColumnTimeseries) Resolve(ctx context.Context, rt *runtime.Runtime, ins
 		querySQL := `CREATE TEMPORARY TABLE ` + temporaryTableName + ` AS (
 			-- generate a time series column that has the intended range
 			WITH template as (
-			SELECT
-				range as ` + tsAlias + `
-			FROM
-				range(
-				date_trunc('` + dateTruncSpecifier + `', timezone(?, ?::TIMESTAMPTZ)),
-				date_trunc('` + dateTruncSpecifier + `', timezone(?, ?::TIMESTAMPTZ)),
-				INTERVAL '1 ` + dateTruncSpecifier + `')
+				SELECT
+					range as ` + tsAlias + `
+				FROM
+					range(
+					date_trunc('` + dateTruncSpecifier + `', timezone(?, ?::TIMESTAMPTZ)),
+					date_trunc('` + dateTruncSpecifier + `', timezone(?, ?::TIMESTAMPTZ)),
+					INTERVAL '1 ` + dateTruncSpecifier + `')
 			),
 			-- transform the original data, and optionally sample it.
 			series AS (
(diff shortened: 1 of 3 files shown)


runtime/server/queries_timeseries_test.go
```diff
@@ -457,16 +471,19 @@ func TestServer_Timeseries_timezone_dst_forward(t *testing.T) {
...
+	require.Equal(t, 5, len(results))
 
 	require.Equal(t, "2023-03-26 00:00:00", results[0].Ts.AsTime().Format(time.DateTime))
 	require.Equal(t, 1.0, results[0].Records.Fields["count"].GetNumberValue())
 	require.Equal(t, "2023-03-26 01:00:00", results[1].Ts.AsTime().Format(time.DateTime))
-	require.Equal(t, 1.0, results[1].Records.Fields["count"].GetNumberValue())
-	require.Equal(t, "2023-03-26 02:00:00", results[2].Ts.AsTime().Format(time.DateTime))
+	require.Equal(t, 0.0, results[1].Records.Fields["count"].GetNumberValue())
+	require.Equal(t, "2023-03-26 01:00:00", results[2].Ts.AsTime().Format(time.DateTime))
 	require.Equal(t, 1.0, results[2].Records.Fields["count"].GetNumberValue())
```
begelundmuller, line 481: It looks like the same time series value occurred twice in the output?

egor-ryashin: Yes, during backward shift one hour is double in the local time, should we merge them?

begelundmuller: Yes, the line chart is expecting one value per time bin

runtime/queries/column_timeseries.go
```diff
@@ -121,34 +121,38 @@ func (q *ColumnTimeseries) Resolve(ctx context.Context, rt *runtime.Runtime, ins
...
-				timezone(?, date_trunc('` + dateTruncSpecifier + `', timezone(?, ?::TIMESTAMPTZ))),
-				timezone(?, date_trunc('` + dateTruncSpecifier + `', timezone(?, ?::TIMESTAMPTZ))),
-				INTERVAL '1 ` + dateTruncSpecifier + `')
+				SELECT
+					range as ` + tsAlias + `
+				FROM
+					range(
+					date_trunc('` + dateTruncSpecifier + `', timezone(?, ?::TIMESTAMPTZ)),
+					date_trunc('` + dateTruncSpecifier + `', timezone(?, ?::TIMESTAMPTZ)),
+					INTERVAL '1 ` + dateTruncSpecifier + `')
```
begelundmuller, line 136: Could it just do `SELECT DISTINCT` here instead of aggregating and using `LAST`?

egor-ryashin: We can do it after final timezone conversion only.

begelundmuller (approved):

PR #2826, opened 2023-07-26
Leaderboard context column e2e tests

bcolloran:
Add e2e test for context columns, as well as a couple aria tweaks that make testing easier.

Meant as a follow up to #2804; only 3 files are changed in this PR relative to that one


bcolloran:
@skokenes here are the e2e tests I added for the context column toggles. I did decide to group them in a function just for organizational purposes, following the pattern you recommended that Hamilton had used previously.

web-common/src/components/button-group/SubButton.svelte
```diff
@@ -22,17 +23,18 @@
 
   registerSubButton(value);
 
-  $: isDisabled = $disabledKeys.includes(value);
+  $: disabled = $disabledKeys.includes(value);
+  $: console.log("disabled", disabled);
```
skokenes, line 27: Assuming this wasn't meant to be left behind?

skokenes (approved):
Awesome, lets merge. I left a comment about a console.log statement that I think you forgot to review

PR #2829, opened 2023-07-27
Misc. runtime review

begelundmuller:


runtime/services/catalog/migrator/sources/sources.go
```diff
@@ -226,9 +226,10 @@ func ingestSource(ctx context.Context, olap drivers.OLAPStore, repo drivers.Repo
 
 	ingestionLimit := opts.IngestStorageLimitInBytes
 	p := &progress{}
-	ticker := time.NewTicker(5 * time.Second)
 	limitExceeded := false
 	go func() {
+		ticker := time.NewTicker(5 * time.Second)
+		defer ticker.Stop()
 		select {
```
k-anshul, line 233: This should also be inside a for loop.

k-anshul (approved):

PR #2845, opened 2023-07-31
Adding sql source for object store types

AdityaHegde:
## Checklist
- [x] Manual verification
- [ ] Unit test coverage
- [x] E2E test coverage
- [x] Needs manual QA?

#### Issue addressed: 
We added sql sources support in #2807 . This PR makes it so that the sources modal generates a yaml that uses this feature.

#### Details:
Adding a source through the modal for gcs, s3 and http(s) will create a source with a yaml with a query,
```
type: duckdb
sql: select * from read_csv_auto('s3://path/to/data.csv')
```
Here the function is dependant on the file extension.

Note: Local files are not handled this way just yet. A follow up PR will address this with some additional backend changes.

## Steps to Verify
1. Create a source with gcs, s3 or https.
2. A yaml should be generated with an sql statement.


web-common/src/features/sources/sourceUtils.ts
```diff
@@ -37,6 +41,25 @@ export function compileCreateSourceYAML(
...
+function buildDuckDbQuery(path: string): string {
+  const extension = extractFileExtension(path as string);
+  if (containsAny(extension, [".csv", ".tsv", ".txt"])) {
+    return `select * from read_csv_auto('${path}')`;
+  } else if (containsAny(extension, [".parquet"])) {
+    return `select * from read_parquet('${path}')`;
+  } else if (containsAny(extension, [".json", ".ndjson"])) {
+    return `select * from read_json('${path}',auto_detect=true,format='auto')`;
+  }
+  return `select * from '${path}'`;
```
ericpgreen2, line 53: This will error, right? Mind adding a comment to explain this case?

web-common/src/features/sources/extract-table-name.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+        );
+      });
+    }
+  });
+});
+
+function getVariations(
+  fileName,
+  expectedFileName,
+  expectedExtension = ".parquet"
```
ericpgreen2, line 33: It'd help readability of the code below if you explicitly set a value in all cases and didn't have a default

web-common/src/features/sources/extract-table-name.spec.ts
ericpgreen2, line 1: The `Variations` object is hard to make sense of. A few ideas:
- Rename `Variations` to `TestCases`, and `getVariations` to `generateTestCases`
- Move to `TestCases` to top of file
- Add a type
- Use the `title` attribute for the entire (not partial) `it` message

web-common/src/features/sources/sourceUtils.ts
```diff
@@ -37,6 +41,25 @@ export function compileCreateSourceYAML(
...
+}
+
+function containsAny(fullExtensions: string, extensions: Array<string>) {
```
ericpgreen2, line 56: I don't follow what "fullExtensions" means. It looks like it's just an "extension"

ericpgreen2, line 56: `containsAny` might be better named `isExtensionInList`

AdityaHegde: This stems from the go APIs to get extension. It only extracts the last part.

web-common/src/features/sources/sourceUtils.ts
```diff
@@ -37,6 +41,32 @@ export function compileCreateSourceYAML(
   return `${topLineComment}\n\ntype: "${connectorName}"\n` + compiledKeyValues;
 }
 
+function buildDuckDbQuery(path: string): string {
+  const extension = extractFileExtension(path as string);
+  if (extensionContainsParts(extension, [".csv", ".tsv", ".txt"])) {
+    return `select * from read_csv_auto('${path}')`;
```
ericpgreen2, line 47: Let's be consistent with `read_json` and use `read_csv` with `auto_detect=true`? Nishant also [cites this spec in Notion](https://www.notion.so/rilldata/infer-types-for-untyped-sources-and-enable-correction-via-YAML-editor-851a7555d056497ba945dcb930348f6e?pvs=4#df6a1bebbf574045a7cd1be5d3ef0d80).

Nishant also has `ignore_errors=1, header=true`. Add those too?

web-common/src/features/sources/extract-table-name.spec.ts
```diff
@@ -0,0 +1,57 @@
+import {
+  extractFileExtension,
+  extractTableName,
+  sanitizeEntityName,
+} from "@rilldata/web-common/features/sources/extract-table-name";
+import { describe, it, expect } from "vitest";
+
+function getVariations(
```
ericpgreen2, line 8: This is still called "variations" which isn't very descriptive. Rename to "TestCases" and "generateTestCases"?

web-common/src/features/sources/sourceUtils.ts
```diff
@@ -37,6 +41,32 @@ export function compileCreateSourceYAML(
...
+
+/**
+ * Checks if a file extension '.v1.parquet.gz' contains parts like '.parquet'
+ */
+function extensionContainsParts(
+  fileExtension: string,
+  extensionParts: Array<string>
+) {
+  for (const extension of extensionParts) {
+    if (fileExtension.indexOf(extension) >= 0) return true;
```
ericpgreen2, line 65: Any reason to not use `.includes()`? It'd be more readable.

ericpgreen2 (approved):
I left a few more small comments, then it looks good 👍

PR #2846, opened 2023-08-01
granular access policies 

### LINKED ISSUE #2346: Admin/Runtime: Granular access policies

pjain1:
Fixes #2346

1. What should be the behaviour when policy is defined but `has_access` is not ? then should we assume true as default.
2. Front end calls files api to get list of files in dashboard folder and uses that to list out dashboards on the project, however we currently do not check policy in this api call as we will have to get catalog entry if we want to do that. So even if user does not have access to few dashboards they are all shown, however the one with no access are greyed out with hover msg like "This dashboard isn't working, check project logs" as while getting the metricsview definition it will get error. May be front end should check the error code to hide these dashboards.


runtime/server/catalog.go
```diff
@@ -33,6 +33,58 @@ func (s *Server) ListCatalogEntries(ctx context.Context, req *runtimev1.ListCata
...
+					for _, exclude := range policy.Exclude {
+						restricted[exclude] = true
+					}
+					for _, dim := range mv.Dimensions {
+						if !restricted[dim.Name] {
+							allowedDims = append(allowedDims, dim)
+						}
+					}
+					mv.Dimensions = allowedDims
+				}
```
begelundmuller, line 80: 1. Maybe pull it into a separate function the same logic is in `GetCatalogEntry`
2. I'm not sure it's safe to mutate the `mv` returned from a catalog lookup, might be cached now or in the future

begelundmuller (changes requested):

admin/server/projects.go
```diff
@@ -122,6 +123,21 @@ func (s *Server) GetProject(ctx context.Context, req *adminv1.GetProjectRequest)
...
+	}
+	// Find User groups
+	groups, err := s.admin.DB.FindUserGroups(ctx, user.ID, proj.OrganizationID)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	groupNames := make([]string, len(groups))
+	for i, group := range groups {
+		groupNames[i] = group.Name
+	}
```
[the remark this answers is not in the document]
pjain1: Should I do `claims.OwnerType()` check and add user info only if the claims is of type `OwnerTypeUser` ?

runtime/compilers/rillv1/template_test.go
```diff
@@ -70,3 +70,15 @@ func TestAnalyze(t *testing.T) {
...
+	template := "SELECT partner_id FROM domain_partner_mapping WHERE domain = {{ .claims.domain }} and groups IN ({{ .claims.groups | join \",\" }})"
+	resolved, err := ResolveTemplate(template, TemplateData{
+		Claims: map[string]any{
+			"domain": "'rilldata.com'",
+			"groups": []string{"'admin'", "'user'"},
+		},
+	})
+	require.NoError(t, err)
+	require.Equal(t, "SELECT partner_id FROM domain_partner_mapping WHERE domain = 'rilldata.com' and groups IN ('admin','user')", resolved)
+}
```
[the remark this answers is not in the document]
pjain1: the main reason for doing this is to handle list of values, for example there is no templating syntax to achieve a valid list syntax when user quotes the variable. Here, for the part `groups IN ({{ .claims.groups | join \",\" }})` if user does `groups IN ('{{ .claims.groups | join \",\" }}')` then the resolved expression will be `groups IN ('admin,user')` which is incorrect. Any better way that you can think of to handle lists. 

I was also thinking if the user should just specify `{{ .claims.groups }}` and we should figure out if its a list so add `( )` and join the group values with a `,` but went with this to start with.

runtime/queries/metricsview.go
```diff
@@ -24,11 +26,14 @@ func lookupMetricsView(ctx context.Context, rt *runtime.Runtime, instanceID, nam
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 
-	if obj.GetMetricsView() == nil {
+	if !obj.IsMetricsView() {
```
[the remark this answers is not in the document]
pjain1: `GetMetricsView()` panics if its not a metrics view so the check `obj.GetMetricsView() == nil` does not make sense as it will never go inside `if` condition. So added this method as its used during query cache key computation - if entry is metricsview then add user attr to the key, it was also being used in `column_time_range` where we don't want to panic if its not a metricsview but just move on.
```
func (e *CatalogEntry) GetMetricsView() *runtimev1.MetricsView {
	obj, ok := e.Object.(*runtimev1.MetricsView)
	if !ok {
		panic(fmt.Errorf("entry '%s' is not a metrics view", e.Name))
	}
	return obj
}
```

runtime/server/auth/claims.go
```diff
@@ -43,9 +54,16 @@ func (c *jwtClaims) CanInstance(instanceID string, p Permission) bool {
...
+	return &c.User
+}
+
 // openClaims implements Claims and allows all actions.
 // It is used for servers with auth disabled.
-type openClaims struct{}
+type openClaims struct {
+	jwt.RegisteredClaims
+	User UserClaims `json:"usr,omitempty"`
+}
```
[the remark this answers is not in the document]
pjain1: actually thought of that but was not sure, will create a new one

runtime/server/auth/metricsview_policy.go
```diff
@@ -0,0 +1,114 @@
...
+	// take hash of mv and userClaims
+	key := fmt.Sprintf("%v:%v:%v:%v:%v:%v", userClaims.Name, userClaims.Email, userClaims.Domain, strings.Join(userClaims.Groups, ","), mv.Name, mv.CatalogUpdatedOn)
+	cached, ok := lru.Get(key)
+	if ok {
+		return cached.(*ResolvedMetricsViewPolicy), nil
+	}
+
+	if userClaims.IsOpen {
+		return openPolicy, nil
+	}
```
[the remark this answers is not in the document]
pjain1: I think when aud in nil (local mode) and if attributes are set we apply policy otherwise we don't. So of the attribute values are set in the UI policies are applied otherwise it has the same behaviour as current, would be useful during development.

runtime/server/auth/metricsview_policy.go
```diff
@@ -0,0 +1,114 @@
...
+	resolved := &ResolvedMetricsViewPolicy{}
+
+	userClaimsMaps := make(map[string]any)
+	userClaimsMaps["name"] = fmt.Sprintf("'%v'", userClaims.Name)
+	userClaimsMaps["email"] = fmt.Sprintf("'%v'", userClaims.Email)
+	userClaimsMaps["domain"] = fmt.Sprintf("'%v'", userClaims.Domain)
+	for i, g := range userClaims.Groups {
+		userClaims.Groups[i] = fmt.Sprintf("'%v'", g)
+	}
+	userClaimsMaps["groups"] = strings.Join(userClaims.Groups, ",")
```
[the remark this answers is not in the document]
pjain1: I looked at the functions but did not find any function that can add quotes on individual group values also the list functions would add `[]` causing further issues so I went ahead with this approach.

runtime/server/catalog.go
pjain1: 2. This is just a GET API call and seriliazed entry is returned to the caller so I think it should be safe, caching may happen at DB layer but we are mutating before returning the entry. Any other case that you can think of ?

pjain1:
Few pending things (needs discussion on some items as well) -
1. Apply include/exclude to measures
2. [UI] support to ask for dev JWT in local env and design around it
3. Adding support for roles in user attributes and semantics around it
4. [UI] Fix use of files api to display list of dashboards

runtime/server/auth/metricsview_policy.go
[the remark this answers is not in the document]
begelundmuller: See response to other comment

runtime/server/catalog.go
begelundmuller: We are adding a catalog object cache in async reconcile, so it won't be safe to mutate the returned values without copying it

runtime/compilers/rillv1/template_test.go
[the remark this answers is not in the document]
pjain1: I think the main thing to decide here is whether we should quote or make the attributes sql compatible for the user or we leave it on the user to create sql compatible syntax and we just replace the template with values as it is. If we leave on user then they might need to understand how sprig templating works and functions they can use in which case the 1st option above looks good and we don't quote single attribute values as well. Otherwise the 3rd option looks good. I did not get how 2nd option is applicable here.

begelundmuller:
> 1. Apply include/exclude to measures

Can you suggest what this could look like? We can maybe leverage that dimension/measure names are unique to avoid the need for separate sections?

> 3. Adding support for roles in user attributes and semantics around it

We could pass a list of all the user's resolved roles, but it seems weird because it would mix org and project level roles. Passing the permissions also doesn't seem great since they may change.

Ideally, people would handle these kinds of permissions in their mappings, but in practice, people will probably want an easy way to know about roles, especially if someone is admin or not.

What about we just introduce a `.claims.admin` bool field that is true if the resolved permissions have `manage_project`? That should address must use cases.

runtime/queries/metricsview.go
```diff
@@ -25,17 +26,20 @@ import (
...
 	obj, err := rt.GetCatalogEntry(ctx, instanceID, name)
 	if err != nil {
-		return nil, status.Error(codes.InvalidArgument, err.Error())
+		return nil, time.Now(), status.Error(codes.InvalidArgument, err.Error())
 	}
 
-	if obj.GetMetricsView() == nil {
-		return nil, status.Errorf(codes.NotFound, "object named '%s' is not a metrics view", name)
+	if !obj.IsMetricsView() {
+		return nil, time.Now(), status.Errorf(codes.NotFound, "object named '%s' is not a metrics view", name)
```
begelundmuller, line 37: Use `time.Time{}` for the zero time

runtime/queries/column_time_range.go
```diff
@@ -19,6 +19,8 @@ type ColumnTimeRange struct {
 	TableName  string
 	ColumnName string
 	Result     *runtimev1.TimeRangeSummary
+	// will be empty for profiling queries
+	PolicyFilter string
```
begelundmuller, line 23: It would be great not to mix up the `ColumnTimeRange` (for modeling) and `MetricsViewTimeRange` (dashboard) APIs. There's not so much code here – can we just duplicate it in `MetricsViewTimeRange` so we keep them separated?

runtime/queries/column_timeseries.go
```diff
@@ -43,8 +43,9 @@ type ColumnTimeseries struct {
 	Result              *ColumnTimeseriesResult                           `json:"-"`
 
 	// MetricsView-related fields. These can be removed when MetricsViewTimeSeries is refactored to a standalone implementation.
-	MetricsView       *runtimev1.MetricsView       `json:"-"`
-	MetricsViewFilter *runtimev1.MetricsViewFilter `json:"filters"`
+	MetricsView       *runtimev1.MetricsView             `json:"-"`
+	MetricsViewFilter *runtimev1.MetricsViewFilter       `json:"filters"`
+	MetricsViewPolicy *runtime.ResolvedMetricsViewPolicy `json:"-"`
```
begelundmuller, line 48: Keep an eye on this PR: https://github.com/rilldata/rill/pull/2925

runtime/queries/metricsview_comparison_toplist.go
```diff
@@ -69,24 +72,60 @@ func (q *MetricsViewComparisonToplist) Resolve(ctx context.Context, rt *runtime.
...
+			}
+		} else {
+			// if no include/exclude is specified, then all dimensions are allowed
+			allowed = true
+		}
+
+		if !allowed {
+			return status.Error(codes.Unauthenticated, "action not allowed")
+		}
+	}
```
begelundmuller, line 114: Can it use `checkFieldAccess` here?

runtime/registry_test.go
```diff
@@ -535,11 +535,12 @@ func NewTestRunTime(t *testing.T) *Runtime {
...
-		MetastoreConnector:  "metastore",
-		QueryCacheSizeBytes: int64(datasize.MB) * 100,
-		AllowHostAccess:     true,
-		SystemConnectors:    globalConnectors,
+		ConnectionCacheSize:   100,
+		MetastoreConnector:    "metastore",
+		QueryCacheSizeBytes:   int64(datasize.MB) * 100,
+		AllowHostAccess:       true,
+		SystemConnectors:      globalConnectors,
+		PolicyEngineCacheSize: 100,
```
begelundmuller, line 543: Put right after QueryCacheSizeBytes

runtime/runtime.go
```diff
@@ -19,7 +20,8 @@ type Options struct {
 	AllowHostAccess     bool
 	SafeSourceRefresh   bool
 	// SystemConnectors are drivers whose handles are shared with all instances
-	SystemConnectors []*runtimev1.Connector
+	SystemConnectors      []*runtimev1.Connector
+	PolicyEngineCacheSize int
```
begelundmuller, line 24: Put right after QueryCacheSizeBytes

runtime/server/auth/claims.go
```diff
@@ -74,3 +85,30 @@ func (c anonClaims) Can(p Permission) bool {
 func (c anonClaims) CanInstance(instanceID string, p Permission) bool {
 	return false
 }
+
+func (c anonClaims) Attributes() map[string]any {
+	return nil
+}
+
+// devJWTClaims implements Claims and allows all actions.
+// It is used for servers with auth disabled.
```
begelundmuller, line 94: Suggestion: "It is used for mimicking user attributes on local when auth is disabled"

(because `openClaims` is used in regular cases when auth is disabled)

runtime/server/server.go
```diff
@@ -319,3 +319,20 @@ func (s *Server) addInstanceRequestAttributes(ctx context.Context, instanceID st
...
 	observability.AddRequestAttributes(ctx, attrs...)
 }
+
+func (s *Server) IssueDevJWT(ctx context.Context, req *runtimev1.IssueDevJWTRequest) (*runtimev1.IssueDevJWTResponse, error) {
+	attr := map[string]any{
+		"Name":   req.Name,
+		"Email":  req.Email,
+		"Domain": req.Email[strings.LastIndex(req.Email, "@")+1:],
+		"Groups": req.Groups,
+	}
```
begelundmuller, line 329: 1. Shouldn't the keys be lowercase, like in admin?
2. Missing key for `admin`

runtime/server/catalog.go
```diff
@@ -5,7 +5,9 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/mohae/deepcopy"
```
begelundmuller, line 8: This does not seem like a reliable dependency

begelundmuller: You can leverage `proto.Clone(m)` on the actual `runtimev1.MetricsView` object instead

runtime/server/catalog.go
```diff
@@ -33,6 +35,29 @@ func (s *Server) ListCatalogEntries(ctx context.Context, req *runtimev1.ListCata
...
+
+			if policy != nil && !policy.HasAccess {
+				continue
+			}
+			// deep copy the catalog entry, so we can filter dimensions which mutates the metricsview
+			newEntry := deepcopy.Copy(entry).(*drivers.CatalogEntry)
+			mv := newEntry.GetMetricsView()
+
+			filterDimensions(policy, mv)
+			accessibleEntries = append(accessibleEntries, newEntry)
```
begelundmuller, line 56: It would be preferable only to clone the metrics view if it has dimension filters (most will probably only have row filters)

runtime/server/catalog.go
```diff
@@ -63,14 +88,62 @@ func (s *Server) GetCatalogEntry(ctx context.Context, req *runtimev1.GetCatalogE
...
+		}
+
+		if policy != nil && !policy.HasAccess {
+			return nil, ErrForbidden
+		}
+		// deep copy the entry, so we can filter dimensions which mutates the metricsview
+		newEntry = deepcopy.Copy(entry).(*drivers.CatalogEntry)
+		mv := newEntry.GetMetricsView()
+
+		filterDimensions(policy, mv)
```
begelundmuller, line 105: Same comment as above

runtime/services/catalog/artifacts/artifacts_test.go
```diff
@@ -545,7 +545,7 @@ func TestSanitizedName(t *testing.T) {
 	}
 }
 
-func TestReadWithEnvVariables(t *testing.T) {
+/*func TestReadWithEnvVariables(t *testing.T) {
```
begelundmuller, line 548: Can add back for sources now?

runtime/policies.go
```diff
@@ -0,0 +1,124 @@
...
+	key := fmt.Sprintf("%v:%v:%v", instanceID, mv.Name, lastUpdatedOn)
+	for k, v := range attributes {
+		key += fmt.Sprintf(":%v:%v", k, v)
+	}
+	var h maphash.Hash
+	_, err := h.WriteString(key)
+	if err != nil {
+		return nil, err
+	}
+	cacheKey := h.Sum64()
```
begelundmuller, line 63: 1. I don't believe map iteration order is deterministic
2. The maphash implementation is not a secure hash (so someone could get another user's access policy)
3. Should be able to write the individual components of the key directly to the hash (removing need for building a string first)

runtime/policies.go
```diff
@@ -0,0 +1,124 @@
...
+	p.lock.Lock()
+	defer p.lock.Unlock()
+
+	cached, ok := p.cache.Get(cacheKey)
+	if ok {
+		return cached.(*ResolvedMetricsViewPolicy), nil
+	}
+
+	resolved := &ResolvedMetricsViewPolicy{}
+	templateData := &rillv1.TemplateData{User: attributes}
```
begelundmuller, line 74: Can remove `&`?

runtime/policies.go
```diff
@@ -0,0 +1,124 @@
...
+			return nil, err
+		}
+		resolved.Filter = filter
+	}
+
+	for _, inc := range mv.Policy.Include {
+		inc.Condition, err = rillv1.ResolveTemplate(inc.Condition, *templateData)
+		if err != nil {
+			return nil, err
+		}
```
begelundmuller, line 99: Overwrites the `mv` object

runtime/query.go
```diff
@@ -87,6 +88,18 @@ func (r *Runtime) Query(ctx context.Context, instanceID string, query Query, pri
...
+			policy := entry.GetMetricsView().Policy
+			if policy != nil {
+				attr := auth.GetClaims(ctx).Attributes()
+				attrKey := ""
+				for k, v := range attr {
+					attrKey += fmt.Sprintf(":%v:%v", k, v)
+				}
+				depKeys[i] += attrKey
+			}
+		}
```
begelundmuller, line 102: Doesn't seem quite right, some issues:

1. Iterating over the attributes has non-deterministic order
2. This section is for dependency names, not for the policy (if something had a dependency on 2 metrics views, the attributes would be added twice, though nothing does that now)
3. It will calculate an attribute-based cache key redundantly since the same already happens in the policy engine. Should such a hash calculation be moved to the claims object and cached there?
4. Since the policy filter relates to how the query is executed, would it make more sense to cache based on the resolved filter instead of the input attributes? The resolved filter could maybe be incorporated into `query.Key()`? 
  - This would enable cache sharing among people who resolve to the same filter, for example if the policy only checks domain.

begelundmuller (changes requested):

runtime/queries/metricsview.go
```diff
@@ -578,3 +579,20 @@ func (q *MetricsViewRows) generateFilename(mv *runtimev1.MetricsView) string {
...
+		return nil, nil, err
+	}
+
+	resolvedPolicy, err := rt.ResolveMetricsViewPolicy(auth.GetClaims(ctx).Attributes(), instanceID, mv, lastUpdatedOn)
+	if err != nil {
+		return nil, nil, err
+	}
+	if resolvedPolicy != nil && !resolvedPolicy.HasAccess {
+		return nil, nil, status.Error(codes.Unauthenticated, "action not allowed")
+	}
```
begelundmuller, line 595: What about:

- Computing the resolved policy in the server API handler
- Then setting it as a property on the query before resolving it

That should make it available for use in `Key()`, avoiding the hacks around casting to `MetricViewQuery`. It would also avoid mixing authentication context (which is scoped to `runtime/server`) into the `runtime/queries` package (which may be called from outside server scope, such is in the upcoming scheduled reports feature).

Is that possible or does anything require the policy to be resolved inside the `queries` package?

begelundmuller (changes requested):

pjain1:
lint job is failing with 
```
  Error: G602: Potentially accessing slice out of bounds (gosec)
  Error: G602: Potentially accessing slice out of bounds (gosec)
```
It passes locally.

runtime/server/catalog.go
```diff
@@ -33,12 +35,33 @@ func (s *Server) ListCatalogEntries(ctx context.Context, req *runtimev1.ListCata
...
+	} else {
+		pbs = make([]*runtimev1.CatalogEntry, len(entries))
+		for i, obj := range entries {
+			var err error
+			pbs[i], err = catalogObjectToPB(obj)
+			if err != nil {
+				return nil, status.Error(codes.Unknown, err.Error())
+			}
 		}
 	}
```
begelundmuller, line 66: I think it needs to push the conditional inside the for loop. Because if `req.Type` is empty, `ListCatalogEntries` will return all objects (including metrics views)

begelundmuller (approved):

PR #2855, opened 2023-08-01
Adding a name safeguard for url state

AdityaHegde:
## Checklist
- [x] Manual verification
- [ ] Unit test coverage
- [ ] E2E test coverage
- [ ] Needs manual QA?

#### Issue addressed: 
Switching between dashboard is inconsistent with retaining all the selections.

#### Details:
There was a race condition in the url sync component. It would run with state of older dashboard for newer dashboard.

## Steps to Verify
1. Create 2 dashboards.
2. Go to one and select a few filters and change the time settings.
3. Go to the other and do the same.
4. Switching between these should retain the filters and time selections.


skokenes (approved):
Great!

PR #2857, opened 2023-08-01
Adding url utils for partial replace of search params

AdityaHegde:
## Checklist
- [ ] Manual verification
- [ ] Unit test coverage
- [ ] E2E test coverage
- [ ] Needs manual QA?


web-common/src/features/feature-flags.ts
```diff
@@ -6,3 +7,14 @@ export type FeatureFlags = {
...
 });
+
+export function retainFeaturesFlags(navigation: BeforeNavigate) {
+  if (!navigation.from.url.searchParams.has("features")) {
+    return;
+  }
+
+  navigation.to.url.searchParams.set(
+    "features",
+    navigation.from.url.searchParams.get("features")
```
skokenes, line 18: In your code you use the search param `features`, but in your tests you are using the param `feature` (no 's').
Sounds like we need to pick with one and stick with it. Doesn't really matter to me but I'm used to it being just `feature` in other systems I've used

web-common/src/lib/url-utils.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+  beforeAll(() => {
+    createPageMock();
+  });
+
+  describe("getFullUrlForPath", () => {
+    const Variations: Array<[string, string, string]> = [
+      ["/path/to/dashboard", "/new/path/dashboard", "/new/path/dashboard"],
+      ["/path/to/dashboard", "new/path/dashboard", "/new/path/dashboard"],
+      [
+        "/path/to/dashboard/?feature=all",
```
skokenes, line 29: These say `feature` but the feature flags url code says `features`, we need to pick one

web-common/src/lib/url-utils.ts
```diff
@@ -0,0 +1,26 @@
+import { page } from "$app/stores";
+import { get } from "svelte/store";
+
+export function getUrlForPath(path: string, retainParams = ["features"]): URL {
```
skokenes, line 4: `features` or `feature`?

skokenes (changes requested):
The code says `features`, but the tests say `feature` (no 's'). 
Otherwise, this looks great!

skokenes (approved):

PR #2861, opened 2023-08-01
Parquet exporting

egor-ryashin:


runtime/queries/metricsview.go
```diff
@@ -404,3 +409,107 @@ func writeXLSX(meta []*runtimev1.MetricsViewColumn, data []*structpb.Struct, wri
...
 	return err
 }
+
+func writeParquet(meta []*runtimev1.MetricsViewColumn, data []*structpb.Struct, ioWriter io.Writer) error {
+	fields := make([]arrow.Field, 0, len(meta))
+	for _, f := range meta {
+		arrowField := arrow.Field{}
+		arrowField.Name = f.Name
+		switch f.Type {
+		case "CODE_BOOL":
```
begelundmuller, line 419: Can use the generated constants instead, like `runtimev1.Type_CODE_BOOL`

runtime/queries/metricsview.go
```diff
@@ -404,3 +409,107 @@ func writeXLSX(meta []*runtimev1.MetricsViewColumn, data []*structpb.Struct, wri
...
+		case "CODE_UINT64":
+			arrowField.Type = arrow.PrimitiveTypes.Uint64
+		case "CODE_DECIMAL":
+			arrowField.Type = arrow.PrimitiveTypes.Float64
+		case "CODE_FLOAT32":
+			arrowField.Type = arrow.PrimitiveTypes.Float32
+		case "CODE_FLOAT64":
+			arrowField.Type = arrow.PrimitiveTypes.Float64
+		case "CODE_STRUCT", "CODE_UUID", "CODE_ARRAY", "CODE_TIMESTAMP", "CODE_TIME", "CODE_DATE", "CODE_STRING", "CODE_MAP":
+			arrowField.Type = arrow.BinaryTypes.String
```
begelundmuller, line 446: Timestamps are very common, so we should use the arrow type for that (it might be an annotated type, don't remember)

runtime/queries/metricsview.go
egor-ryashin: we can have it like this:
```go
		typeCode := runtimev1.Type_Code(runtimev1.Type_Code_value[f.Type])
		switch typeCode {
		case runtimev1.Type_CODE_BOOL:
			arrowField.Type = arrow.FixedWidthTypes.Boolean
```

runtime/queries/metricsview.go
egor-ryashin: That depends on this https://github.com/xitongsys/parquet-go/pull/551

begelundmuller (approved):

PR #2868, opened 2023-08-02
used_on field to user and service auth tokens

rakeshsharma14317:


admin/database/postgres/migrations/0012.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE user_auth_tokens ADD COLUMN used_on TIMESTAMPTZ DEFAULT now() NOT NULL;
+ALTER TABLE service_auth_tokens ADD COLUMN used_on TIMESTAMPTZ DEFAULT now() NOT NULL;
+ALTER TABLE users ADD COLUMN active_on TIMESTAMPTZ DEFAULT now() NOT NULL;
```
begelundmuller, line 3: Can we also add `UPDATE` statements setting these fields to the `updated_on` value for existing rows? Just to make existing rows have more conservative values

admin/used_flusher.go
```diff
@@ -47,6 +53,27 @@ func (u *usedFlusher) Deployment(id string) {
...
+
+	u.userTokens[id] = true
+}
+
+func (u *usedFlusher) ServiceTokens(id string) {
+	u.mu.Lock()
+	defer u.mu.Unlock()
+
+	u.serviceTokens[id] = true
+}
```
begelundmuller, line 75: Make the function names singular like `Deployment` and `User` (i.e. `UserToken` instead of `UserTokens`)

admin/used_flusher.go
```diff
@@ -73,24 +100,48 @@ func (u *usedFlusher) flush() {
...
+	u.userTokens = make(map[string]bool)
+
+	serviceTokens := u.serviceTokens
+	u.serviceTokens = make(map[string]bool)
 	u.mu.Unlock()
 
-	if len(deployments) > 0 {
-		u.logger.Info("flushing used_on to db", zap.Int("deployments", len(deployments)))
+	// Helper function to perform the flushing of used_on to the database.
+	flushToDB := func(data map[string]bool, updateFn func(ctx context.Context, ids []string) error, logMsg string) {
```
begelundmuller, line 115: Instead of an inline function, move to a separate function like `func (u *usedFlusher) flushSingle(...` or something like that

admin/used_flusher.go
```diff
@@ -73,24 +100,48 @@ func (u *usedFlusher) flush() {
...
+
+	users := u.users
+	u.users = make(map[string]bool)
+
+	userTokens := u.userTokens
+	u.userTokens = make(map[string]bool)
+
+	serviceTokens := u.serviceTokens
+	u.serviceTokens = make(map[string]bool)
 	u.mu.Unlock()
```
begelundmuller, line 112: Slight optimization suggestion: only call `make` if the previous one was not empty. I.e. wrap the `u.xxx = ...` calls with `if len(u.xxx) > 0 { ...`

begelundmuller (changes requested):

begelundmuller (approved):

PR #2869, opened 2023-08-02
feat: pivot skeleton

skokenes:
## Checklist
- [x] Manual verification
- [ ] Unit test coverage
- [ ] E2E test coverage
- [ ] Needs manual QA?

#### Issue addressed: 
Foundation for the time dimensions detail pivot table https://www.notion.so/rilldata/Time-Dimension-Detail-Design-Mocks-and-Specification-2dd22773d6e8479ca8d297f95877c669

#### Details:
This is the first in a long series of incremental PRs to satisfy the Time Dimensions Detail Table feature.
This first PR implements a very basic, configurable table with
- vertical and horizontal virtualization
- fixed column support

Due to the complexity and unpredictability of future pivot table requirements, this PR tries to avoid over-abstraction and configuration. We will gradually introduce more configurability and abstractions as we add features. In this PR, mock state and data is used to control the table, since the scope of this PR is only the table scrolling and basic rendering. Future PRs will show how to leverage real state and async data sources.

CleanShot 2023-08-02 at 12 44 23

## Steps to Verify
For example, given a new button "Foo" in the dashboard:
1. Run `npm run dev`
2. Open http://localhost:3000/dev/table/<YOUR_DASHBOARD_NAME>
3. Play with the mock data in table by scrolling vertically and horizontally


web-common/src/features/dashboards/time-dimension-details/Pivot.svelte
```diff
@@ -0,0 +1,216 @@
...
+        },
+      ];
+    }, []);
+
+    // If current virtual column set has fixed columns, remove them since we will use our measurements
+    nonfixedVirtualColumns = virtualColumns.filter(
+      (c) => c.index >= fixedColCt
+    );
+
+    paddingLeft =
```
AdityaHegde, line 80: Feels weird to set paddingLeft and set it again in the next line. Perhaps use a local variable for this line?

web-common/src/features/dashboards/time-dimension-details/Pivot.svelte
```diff
@@ -0,0 +1,216 @@
...
+  bind:this={container}
+  style={`height: ${height}px; overflow: auto;`}
+  class="border"
+>
+  <table
+    class="border-collapse"
+    style={`height: ${totalVerticalSize}px; max-height: ${totalVerticalSize}px; width: ${totalHorizontalSize}px; max-width: ${totalHorizontalSize}px; overflow: none; table-layout: fixed`}
+  >
+    <thead class="sticky top-0 bg-gray-100 z-10">
+      <PivotVirtualRow element="th" {paddingLeft} {paddingRight}>
```
AdityaHegde, line 103: Would it make sense to move the header row, padding row and the data rows in to their own component?

web-common/src/features/dashboards/time-dimension-details/TimeDimensionDetailsTable.svelte
```diff
@@ -0,0 +1,45 @@
+<script lang="ts">
+  import { flexRender } from "@tanstack/svelte-table";
+  import Pivot from "./Pivot.svelte";
+  import TimeDimensionDetailsTableCell from "./TimeDimensionDetailsTableCell.svelte";
+
+  // Mock state for now
+  let state = {
+    getRowSize: () => 35,
```
AdityaHegde, line 8: We should consider passing all this as context and follow the line chart design. Where a top level wrapper component allows for configuration and sets a context. All the components down the tree uses this. Perhaps when we actually use this and start generalising it a bit.

web-common/src/features/dashboards/time-dimension-details/Pivot.svelte
skokenes: No, I don't think adding a layer of abstraction gains us anything at this stage. This file is pretty comprehensible and maintainable as is IMO.
If complexity expands as we add features to the point that separating these segments into their own files would be beneficial, we can do that then.

web-common/src/features/dashboards/time-dimension-details/TimeDimensionDetailsTable.svelte
skokenes: For now, this component is just a mock of how we will set up state.
A Pivot context could work, but at this stage feels like overkill. We only prop drill this property 1 level. We definitely can revisit it later if we find a real need for it.

ericpgreen2 (approved):
Looks good to me

Issue #2853, opened 2023-08-01
Runtime: Support DuckDB exports using `COPY TO`


begelundmuller:

begelundmuller:
Closed by https://github.com/rilldata/rill/pull/2870

PR #2870, opened 2023-08-02
Support COPY TO exporting

egor-ryashin:
https://github.com/rilldata/rill/issues/2853


begelundmuller (commented):
This approach makes sense, a few suggestions:

- Since the exporters are specific to `MetricsViewRows`, consider putting them on that struct (like `q.exportDefault` and `q.exportDuckDB`). I wouldn't worry much about the argument list to these functions getting long since they're internal to this file.

- Since this will be needed for multiple queries, you could add two common util functions similar to `writeXLSX` and others for a) running the `COPY` statement, and b) reading the file into `w`

runtime/queries/metricsview_rows.go
```diff
@@ -100,21 +102,47 @@ func (q *MetricsViewRows) Resolve(ctx context.Context, rt *runtime.Runtime, inst
...
+	case runtimev1.ExportFormat_EXPORT_FORMAT_CSV:
+		return writeCSV(query.Result.Meta, query.Result.Data, w)
+	case runtimev1.ExportFormat_EXPORT_FORMAT_XLSX:
+		return writeXLSX(query.Result.Meta, query.Result.Data, w)
+	}
+
+	return nil
+}
+
+func duckDBCopyExport(sql string, args []any, filename string, olap drivers.OLAPStore, mv *runtimev1.MetricsView, opts *runtime.ExportOptions, w io.Writer, rt *runtime.Runtime, instanceID string, ctx context.Context) error {
```
begelundmuller, line 130: As the linter warning also indicates, can we move the args to `generalExport` and to `duckDBCopyExport` that are in common with `Export` to the start of the arg list for readability?

runtime/queries/metricsview_rows.go
```diff
@@ -123,18 +151,73 @@ func (q *MetricsViewRows) Export(ctx context.Context, rt *runtime.Runtime, insta
...
+		return err
+	}
+
+	defer f.Close()
+
+	_, err = io.Copy(w, f)
+	return err
+}
+
+func (q *MetricsViewRows) Export(ctx context.Context, rt *runtime.Runtime, instanceID string, w io.Writer, opts *runtime.ExportOptions) error {
```
begelundmuller, line 165: Please move the external `Export` func up above the `generalExport` and `duckDBCopyExport` functions

runtime/queries/metricsview_rows.go
```diff
@@ -123,18 +151,73 @@ func (q *MetricsViewRows) Export(ctx context.Context, rt *runtime.Runtime, insta
...
+	}
+
+	mv, err := lookupMetricsView(ctx, rt, instanceID, q.MetricsViewName)
+	if err != nil {
+		return err
+	}
+
+	switch olap.Dialect() {
+	case drivers.DialectDuckDB:
+		if opts.Format == runtimev1.ExportFormat_EXPORT_FORMAT_CSV {
```
begelundmuller, line 178: DuckDB `COPY` also supports Parquet, so we should use it for that also

runtime/queries/metricsview_rows.go
begelundmuller: Actually, since this function is also used for toplists, we should move it to a generic place similar to `writeCSV` funcs and others

begelundmuller (changes requested):
There are also some linter errors to address

PR #2871, opened 2023-08-02
Leaderboard interval formatting

bcolloran:
Implements leaderboard part of [interval formatting PRD](https://www.notion.so/rilldata/Support-display-of-intervals-and-formatting-of-intervals-e-g-25-days-in-dashboards-8720522eded648f58f35421ebc28ee2f), which allows compact representation of numeric measures that have `format_preset: interval_ms`. 

Per the spec, it is up to the user to cast intervals to numerics (or rather, to use the `datediff` function to find the difference between two dates, as apparently it's not possible to cast interval to numeric directly?), and to do the arithmetic required to convert to milliseconds (if needed).


djbarnwal (commented):
Looks good code wise, but I haven't verified it on a live dashboard.

I tried creating a measure with the expression `to_days(2323)` to get an INTERVAL return type,  but got `[Object object]` on the dashboard big number display. Not sure if this is related.

bcolloran:
> Looks good code wise, but I haven't been able to verify it on a live dashboard.
> 
> I tried creating a measure with the expression `to_days(2323)` to get an INTERVAL return type, but got `[Object object]` on the dashboard big number display. Not sure if this is related.

@djbarnwal -- I think it's unrelated, but it sounds like a genuine upstream bug in the measure expression validation: `to_days()` should not be considered a valid aggregation function in a measure expression. If you still have your YAML example and can reproduce this behavior on `main`, I think it would be worth filing a new issue about this.

bcolloran:
@djbarnwal @nishantmonu51 -- this is ready for experience pass and final code review+approval. Thanks!!

bcolloran:
Hmm, I'm getting CI failures that I can't explain. I can't think of any reason why my [latest commit](https://github.com/rilldata/rill/pull/2871/commits/7f2add6f17165de5207b4bf5cc3c6dca467b725f) would cause the timeout errors I'm seeing compared to the previous commit, which passed CI.

I've rerun CI several times but no luck. Just tried merging main in case there have been an CI stabilty updates

nishantmonu51 (approved):
did an experience pass, looks good to me. Thanks @bcolloran for getting this through.

PR #2875, opened 2023-08-03
Driver abstractions

k-anshul:


cli/cmd/runtime/start.go
```diff
@@ -197,3 +199,13 @@ func StartCmd(cliCfg *config.Config) *cobra.Command {
 	}
 	return startCmd
 }
+
+func parseConnectorDefs(s string) []*rillv1.ConnectorDef {
```
begelundmuller, line 203: The `package runtime` should probably have its own type for this (maybe something like `runtime.Connector`, and call the field `Config` instead of `Defaults`.

cli/cmd/runtime/start.go
```diff
@@ -68,6 +68,8 @@ type Config struct {
 	ActivitySinkKafkaBrokers string `default:"" split_words:"true"`
 	// Kafka topic of an activity client's sink
 	ActivitySinkKafkaTopic string `default:"" split_words:"true"`
+	GlobalDrivers          string `split_words:"true"`
+	PrivateDrivers         string `split_words:"true"`
```
begelundmuller, line 72: 1. On the runtime-level, should there not only be one map? And then a separate one can be provided on the instance-level (when calling `CreateInstance`)

runtime/connections.go
```diff
@@ -17,93 +28,166 @@ func (r *Runtime) Registry() drivers.RegistryStore {
...
+	if err != nil {
+		return nil, nil, err
+	}
+	defer release()
+
+	// TODO :: parse it and keep a cache in instance to avoid reparsing
+	yaml, err := rillv1.ParseRillYAML(ctx, repo, instanceID)
+	if err != nil {
+		return nil, nil, err
+	}
```
begelundmuller, line 42: You could call `EditInstance` from reconcile when it reads the `rill.yaml`

runtime/runtime.go
```diff
@@ -6,53 +6,103 @@ import (
...
 
+	"github.com/rilldata/rill/runtime/compilers/rillv1"
 	"github.com/rilldata/rill/runtime/drivers"
 	"go.uber.org/zap"
+	"golang.org/x/exp/maps"
+)
+
+const (
+	// may be use __system__ prefix ?
+	_metastoreDriverName = "metastore"
```
begelundmuller, line 17: Can just be a parameter in `Options`

runtime/registry.go
```diff
@@ -187,15 +191,17 @@ func (r *Runtime) EditInstance(ctx context.Context, inst *drivers.Instance) erro
 // TODO :: this is a rudimentary solution and ideally should be done in some producer/consumer pattern
 func (r *Runtime) evictCaches(ctx context.Context, inst *drivers.Instance) {
 	// evict and close exisiting connection
-	r.connCache.evict(ctx, inst.ID, inst.OLAPDriver, inst.OLAPDSN)
-	r.connCache.evict(ctx, inst.ID, inst.RepoDriver, inst.RepoDSN)
+	c, _, _ := r.opts.OLAPDef(inst.OLAPDSN)
+	r.connCache.evict(ctx, inst.ID, inst.OLAPDriver, variables(_olapDriverName, c.Defaults, inst.ResolveVariables()))
+	c, _, _ = r.opts.RepoDef(inst.RepoDSN)
+	r.connCache.evict(ctx, inst.ID, inst.RepoDriver, variables(_repoDriverName, c.Defaults, inst.ResolveVariables()))
```
begelundmuller, line 197: We can keep `inst.OLAPDSN` and `inst.RepoDSN` for backwards compatibility, but ideally that would just be handled in the APIs.

Elsewhere, could we have `inst.OLAP` and `inst.Repo` be names of connectors, and then just have `inst.Connectors` contain all the declared connectors for it?

cli/cmd/runtime/start.go
k-anshul: I thought since this is going to be fixed/mandatory for every instance we can keep them as part of runtime config itself to group them. 
I can add them as part of instance as well.

runtime/drivers/bigquery/bigquery.go
```diff
@@ -95,7 +95,7 @@ type configProperties struct {
 	AllowHostAccess bool   `mapstructure:"allow_host_access"`
 }
 
-func (d driver) Open(config map[string]any, logger *zap.Logger) (drivers.Connection, error) {
```
begelundmuller, line 98: I think it should add a check and return an error if `shared` is `true` (since it doesn't support that). All the drivers that can't be shared should do that (only `sqlite` can be shared right now, I think?)

runtime/drivers/registry.go
```diff
@@ -44,6 +42,17 @@ type Instance struct {
 	IngestionLimitBytes int64 `db:"ingestion_limit_bytes"`
 	// Annotations to enrich activity events (like usage tracking)
 	Annotations map[string]string
+	// Instance specific connectors
+	Connectors []*runtimev1.ConnectorDef
+	// RillYAML is the project's parsed rill yaml
+	// Its just a cache and can be reproduced from rill.yaml
+	// TODO :: deprecate ProjectVariables once completely moved to v1 parsing
+	RillYAML *RillYAML `db:"rill_yaml"`
```
begelundmuller, line 50: Instead of deprecating `ProjectVariables`, I think we should add `ProjectConnectors` here as well, but not include a `RillYAML` fields (which is just a parser concern, not part of the internal representation)

runtime/drivers/sqlite/migrations/0006.sql
```diff
@@ -0,0 +1,20 @@
...
+UPDATE
+	instances
+SET
+	connectors = format('[%s, %s]',
+	json_replace('{"type":"%s","name":"repo","configs":{"dsn":"%s"}}',
+	'$.type',
+	repo_driver,
+	'$.configs.dsn',
+	repo_dsn),
+	json_replace('{"type":"%s","name":"olap","configs":{"dsn":"%s"}}',
```
begelundmuller, line 11: Per earlier suggestion, remember to change `configs` to `config` here

runtime/query.go
```diff
@@ -65,7 +65,8 @@ func (r *Runtime) Query(ctx context.Context, instanceID string, query Query, pri
 	if err != nil {
 		return err
 	}
-	if inst.OLAPDriver == "druid" {
+	c, _, _ := r.OLAPDef(inst)
+	if c.Type == "druid" {
```
begelundmuller, line 69: `OLAPDef` feels too specific. Maybe do `handle = r.OLAP(instanceID)` and check `handle.Driver() == "druid"`?

Alternatively, it could have something like `r.ConnectorDef(instanceID, name)`?

runtime/registry.go
```diff
@@ -92,8 +95,11 @@ func (r *Runtime) DeleteInstance(ctx context.Context, instanceID string, dropDB
 
 	// Drop the underlying data store
 	if dropDB {
-		conn, err := r.connCache.get(ctx, instanceID, inst.OLAPDriver, inst.OLAPDSN)
+		c, shared, _ := r.OLAPDef(inst)
+		vars := r.variables(inst.OLAPDriver, c.Configs, inst.ResolveVariables())
+		conn, release, err := r.connCache.get(ctx, instanceID, c.Type, vars, shared)
```
begelundmuller, line 100: This should call into a generic function like `AcquireHandle` or something similar. We don't want to replicate the connection cache accessing logic too many places.

Maybe some more generic functions might be needed in there, like `TryAcquire` (only if open) or something like that.

Or a function like `r.ConnectorDef(inst.ID, inst.RepoDriver)` for resolving the connector definition without opening it

runtime/registry.go
```diff
@@ -187,35 +195,88 @@ func (r *Runtime) EditInstance(ctx context.Context, inst *drivers.Instance) erro
...
 	if err != nil {
 		return nil, nil, err
 	}
-	olapStore, ok := olap.AsOLAP()
+	olapStore, ok := olap.AsOLAP(inst.ID)
 	if !ok {
 		return nil, nil, fmt.Errorf("not a valid OLAP driver: '%s'", inst.OLAPDriver)
 	}
 	return olap, olapStore, nil
 }
```
begelundmuller, line 239: Could we leverage some generic functions here? It doesn't feel great to be crafting these repo and OLAP `Open` payloads multiple places

runtime/registry.go
```diff
@@ -187,35 +195,88 @@ func (r *Runtime) EditInstance(ctx context.Context, inst *drivers.Instance) erro
...
+	for _, c := range inst.Connectors {
+		if c.Name == inst.RepoDriver {
+			return c, false, nil
+		}
+	}
+	if c, err := r.ConnectorDefByName(inst.RepoDriver); err == nil {
+		return &runtimev1.ConnectorDef{Name: c.Name, Type: c.Type, Configs: c.Configs}, true, nil
+	}
+	return nil, false, fmt.Errorf("dev error, repo connector doesn't exist")
+}
```
begelundmuller, line 263: As hinted elsewhere – could these be replaced by something like a generic `r.Connector(instID, name)` function that generates a *resolved* connector definition (including merging relevant variables to the config payload). Which in turn could be used everywhere we connect to an instance connector

runtime/runtime.go
```diff
@@ -13,46 +13,50 @@ import (
...
 	ConnectionCacheSize int
 	MetastoreDriver     string
-	MetastoreDSN        string
 	QueryCacheSizeBytes int64
 	AllowHostAccess     bool
 	SafeSourceRefresh   bool
+	// GlobalDrivers are drivers whose handles are shared with all instances
+	GlobalDrivers []*Connector
+	// PrivateDrivers are drivers whose handles are private to an instance
+	PrivateDrivers []*Connector
```
begelundmuller, line 22: Not used

runtime/runtime.go
```diff
@@ -13,46 +13,50 @@ import (
 type Options struct {
 	ConnectionCacheSize int
 	MetastoreDriver     string
-	MetastoreDSN        string
 	QueryCacheSizeBytes int64
 	AllowHostAccess     bool
 	SafeSourceRefresh   bool
+	// GlobalDrivers are drivers whose handles are shared with all instances
+	GlobalDrivers []*Connector
```
begelundmuller, line 20: Can we call it `SystemConnectors` or just  `Connectors`? (Since it's on `Runtime` it's clear these are not instance-specific)

begelundmuller: An unrelated note that I'm putting here – as hinted in another comment, we'd need a `bool` on `Connector` to mark whether a runtime-level connector can be accessed by an instance. But since there are no such cases right now, I would propose we don't add that support and just never allow an instance to access these for now.

runtime/runtime.go
```diff
@@ -13,46 +13,50 @@ import (
...
+	// PrivateDrivers are drivers whose handles are private to an instance
+	PrivateDrivers []*Connector
+}
+
+type Connector struct {
+	Type    string
+	Name    string
+	Configs map[string]string
 }
```
begelundmuller, line 29: Since we anyway have the `runtimev1.Connector` type now, should we just use that and remove this one?

runtime/server/queries_columns_test.go
```diff
@@ -3,10 +3,11 @@ package server
 import (
 	"context"
 	"fmt"
-	"github.com/rilldata/rill/runtime/pkg/ratelimit"
 	"testing"
 	"time"
 
+	"github.com/rilldata/rill/runtime/pkg/ratelimit"
+
 	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
```
begelundmuller, line 11: Remove newline

begelundmuller (changes requested):

proto/rill/runtime/v1/api.proto
```diff
@@ -243,6 +234,17 @@ message Instance {
   map<string, string> project_variables = 8;
   // ingestion limit across all sources, 0 means no limit
   int64 ingestion_limit_bytes = 9;
+  // bare minimum connectors required by the instance. 
+  repeated ConnectorDef connectors = 10;
+}
+ 
+message ConnectorDef {
```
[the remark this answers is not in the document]
k-anshul: There is already a `Connector` in api.proto : https://github.com/rilldata/rill/blob/27edbd045e568af6dc9ea41aa6cfcebf86beeb9a/proto/rill/runtime/v1/api.proto#L715

runtime/caches.go
```diff
@@ -63,61 +76,102 @@ func (c *connectionCache) Close() error {
...
+		// not using instanceID to ensure all instances share the same handle
+		key = driver + generateKey(config)
+	} else {
+		key = instanceID + driver + generateKey(config)
+	}
+	conn, ok := c.cache[key]
+	if !ok { // not in use
+		value, found := c.lruCache.Get(key)
+		if !found { // not opened
+			handle, err := c.openAndMigrate(ctx, instanceID, driver, shared, config)
```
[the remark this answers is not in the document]
k-anshul: Sure. I also think we can redesign this entire cache to not use a global lock. We can take it up as a separate PR.

runtime/compilers/rillv1/parse_rillyaml.go
```diff
@@ -60,6 +63,9 @@ func (p *Parser) parseRillYAML(ctx context.Context, path string) error {
 	}
 
 	for i, c := range tmp.Connectors {
+		if _reservedConnectorNames[c.Name] {
+			return fmt.Errorf("%s are reserved connector names", maps.Keys(_reservedConnectorNames))
```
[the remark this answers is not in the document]
k-anshul: I thought it is easier for user if we return all the reserved names in one go.

runtime/connections.go
```diff
@@ -17,93 +27,168 @@ func (r *Runtime) Registry() drivers.RegistryStore {
...
+			}
+		}
+	}
+	if c, err := r.ConnectorDefByName(connector); err == nil { // connector found
+		// defined in runtime options
+		return r.connCache.get(ctx, instanceID, c.Type, r.variables(connector, c.Configs, instance.ResolveVariables()), true)
+	}
+	// neither defined in rill.yaml nor in runtime options, directly used in source
+	return r.connCache.get(ctx, instanceID, connector, r.variables(connector, nil, instance.ResolveVariables()), false)
+}
```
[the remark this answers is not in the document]
k-anshul: 1. Sure. We can add that support when a need to use global connector in subsequent PRs. For now I will remove the access to global connectors.

runtime/connections.go
```diff
@@ -17,93 +27,168 @@ func (r *Runtime) Registry() drivers.RegistryStore {
...
+	for key, value := range def {
+		vars[strings.ToLower(key)] = value
+	}
+
+	// connector variables are of format connector.name.var
+	// there could also be other variables like allow_host_access, region etc which are global for all connectors
+	prefix := fmt.Sprintf("connector.%s.", name)
+	for key, value := range variables {
+		if !strings.HasPrefix(key, "connector.") { // global variable
+			vars[strings.ToLower(key)] = value
```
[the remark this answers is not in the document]
k-anshul: Currently we don't have any other way to pass global configs like `allow_host_access`

runtime/runtime.go
k-anshul: Let's call it `SystemConnectors`. Will be explicit that these are not instance specific.

runtime/drivers/bigquery/bigquery.go
k-anshul: Postgres can't be shared either but its not being used in runtime now I guess.

proto/rill/runtime/v1/api.proto
```diff
@@ -221,19 +221,10 @@ message Instance {
...
-  }];
-  // DSN for connection to OLAP
-  string olap_dsn = 3;
+  string olap_driver = 2;
   // Driver for reading/editing code artifacts (options: file, metastore, github).
   // This enables virtualizing a file system in a cloud setting.
-  string repo_driver = 4 [(validate.rules).string = {in: ["file", "metastore", "github"]}];
-  // DSN for connecting to repo
-  string repo_dsn = 5;
+  string repo_driver = 4;
```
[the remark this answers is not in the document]
k-anshul: Yeah we could do that in a separate PR. I will raise a follow up PR.

runtime/caches.go
```diff
@@ -63,61 +76,102 @@ func (c *connectionCache) Close() error {
...
-		defer cancel()
-
-		err = conn.Migrate(ctx)
-		if err != nil {
-			return nil, err
+	// increase reference
+	conn.ref++
+	// transfer from lru to in-use cache
+	c.cache[key] = conn
+	c.lruCache.Remove(key)
```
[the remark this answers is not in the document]
k-anshul: Moved it to the case where connection is found in lru

runtime/drivers/registry.go
k-anshul: Instead of adding separate `ProjectVariables` and `ProjectConnectors` field it seems better to just add a single field with all project details as cache. I can rename `RillYAML` to something more generic like `ProjectDetails`/`ProjectInfo` or even `Project`.
Although I think `RillYAML` itself isn't too bad either since we are calling it just a cache of parsed rill yaml.

runtime/registry.go
k-anshul: The `registry.go` is the only place where we expect to directly construct such payloads since instance hasn't been created. Everywhere else we don't expect clients to call `r.Connector` so I didn't create such functions. There are some nuances as well like connector config is `map[string]string` and driver expects `map[string]any` so returning a connector def with all resolved variables is not straightforward as well. 
Since this is the only place we can take a call to construct such payloads(which also uses the same function that AcquireHandle internally does). Alternatively we could save instance in db first and then do all the check like valid repo, valid olap etc and delete instance if required so that we can always call `AcquireHandle`

runtime/registry.go
k-anshul: Responded here : https://github.com/rilldata/rill/pull/2875#discussion_r1291019218

proto/rill/runtime/v1/api.proto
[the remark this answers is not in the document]
begelundmuller: I see – the existing `Connector` maps internally to a `drivers.Spec`. We could change the existing `Connector` type to `ConnectorSpec`.

runtime/compilers/rillv1/parse_rillyaml.go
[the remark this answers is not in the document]
begelundmuller: I think the error message should also specifically identify the problem. Like if you call a table "from" in SQL, it will say something like "unexpected keyword 'from'", not "the reserved keywords are [.......]".

runtime/connections.go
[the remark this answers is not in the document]
begelundmuller: Can we then make a specific exception here for `allow_host_access`?

runtime/drivers/bigquery/bigquery.go
begelundmuller: Yeah we should probably just remove Postgres

runtime/drivers/registry.go
begelundmuller: I understand your point, I think though it feels simpler to just store it flattened (especially since it's not removing `ProjectVariables` now). It's also weird to call it `RillYAML` if it only has the variables and connectors, not the other fields in `rill.yaml`

runtime/registry.go
begelundmuller: What if just create the instance without validating the olap and repo? It should be a rare case that they fail, and I think the error would just be returned from the first reconcile instead. Then people could update the credentials to make it work/

runtime/drivers/registry.go
k-anshul: I think this is the classical discussion of `flatten` vs `nested`. Flatten is simple, nested is extensible. 
I can flatten it for now but once we completely move to v1 parsing and extend rill yaml further, it may be a good opportunity to nest all these fields.

runtime/registry.go
k-anshul: Yeah that should simplify the instance API workflow as well which are slightly complicated right now.

cli/pkg/local/app.go
```diff
@@ -78,20 +78,20 @@ func NewApp(ctx context.Context, ver config.Version, verbose bool, olapDriver, o
...
 	}
 
 	rtOpts := &runtime.Options{
 		ConnectionCacheSize: 100,
-		MetastoreDriver:     "metastore",
+		MetastoreDriver:     "sqlite",
 		QueryCacheSizeBytes: int64(datasize.MB * 100),
 		AllowHostAccess:     true,
-		GlobalDrivers:       globalDrivers,
+		SystemConnectors:    systemConnectors,
```
begelundmuller, line 94: The name is "metastore", but the `MetastoreDriver` is "sqlite" – does this work?

runtime/connections.go
```diff
@@ -5,48 +5,65 @@ import (
...
+		return r.connCache.get(ctx, instanceID, c.Type, r.connectorConfig(connector, c.Config, instance.ResolveVariables()), false)
+	}
+
+	// neither defined in rill.yaml nor set in instance, directly used in source
+	return r.connCache.get(ctx, instanceID, connector, r.connectorConfig(connector, nil, instance.ResolveVariables()), false)
 }
 
-func (r *Runtime) AcquireHandle(ctx context.Context, instanceID, connector string) (drivers.Handle, func(), error) {
+// FlushHandle flushes the db handle for the specific connector from the cache
+func (r *Runtime) FlushHandle(ctx context.Context, instanceID, connector string, drop bool) error {
```
begelundmuller, line 38: Call it `EvictHandle` for consistency with cache terminology

runtime/connections.go
```diff
@@ -5,48 +5,65 @@ import (
...
-		}
+	var driverType string
+	var connectorConfig map[string]any
+	if c, err := r.connectorDef(instance, connector); err == nil {
+		driverType = c.Type
+		connectorConfig = r.connectorConfig(connector, c.Config, instance.ResolveVariables())
+	} else {
+		driverType = connector
+		connectorConfig = r.connectorConfig(connector, nil, instance.ResolveVariables())
 	}
```
begelundmuller, line 52: Could it push the handling of implicit connectors into `connectorDef` also?

runtime/connections.go
```diff
@@ -163,18 +172,34 @@ func (r *Runtime) NewCatalogService(ctx context.Context, instanceID string) (*ca
...
+	}
+
+	// defined in rill.yaml
+	for _, c := range inst.ProjectConnectors {
+		if c.Name == name {
+			return c, nil
+		}
+	}
+	return nil, fmt.Errorf("connector %s doesn't exist", name)
+}
```
begelundmuller, line 190: Can we make this resolve the connector config correctly (calling `connectorConfig`) and set it as the `config` value on the returned object? So the callers don't need to call `connectorConfig` separately

runtime/registry.go
```diff
@@ -22,36 +22,14 @@ func (r *Runtime) FindInstance(ctx context.Context, instanceID string) (*drivers
...
-		return fmt.Errorf("failed to prepare instance: %w", err)
-	}
-	err = repo.Migrate(ctx)
+	_, err = r.connectorDef(inst, inst.RepoDriver)
 	if err != nil {
-		return fmt.Errorf("failed to prepare instance: %w", err)
+		return fmt.Errorf("invalid repo driver")
 	}
 
 	// this is a hack to set variables and pass to connectors
```
begelundmuller, line 35: Can this be removed now and populated at runtime based on `r.opts.AllowHostAccess`?

runtime/runtime.go
```diff
@@ -45,7 +37,7 @@ func New(opts *Options, logger *zap.Logger) (*Runtime, error) {
 		migrationMetaCache: newMigrationMetaCache(math.MaxInt),
 		queryCache:         newQueryCache(opts.QueryCacheSizeBytes),
 	}
-	store, _, err := rt.newMetaStore(context.Background(), "")
+	store, _, err := rt.AcquireGlobalHandle(context.Background(), "metastore")
```
begelundmuller, line 40: Should it be `AcquireSystemHandle` for consistency with `SystemConnectors`?

runtime/testruntime/testruntime.go
```diff
@@ -31,21 +31,21 @@ type TestingT interface {
...
 			// Setting a test-specific name ensures a unique connection when "cache=shared" is enabled.
 			// "cache=shared" is needed to prevent threading problems.
-			Configs: map[string]string{"dsn": fmt.Sprintf("file:%s?mode=memory&cache=shared", t.Name())},
+			Config: map[string]string{"dsn": fmt.Sprintf("file:%s?mode=memory&cache=shared", t.Name())},
 		},
 	}
 	opts := &runtime.Options{
 		ConnectionCacheSize: 100,
-		MetastoreDriver:     "metastore",
+		MetastoreDriver:     "sqlite",
```
begelundmuller, line 45: Shouldn't `MetastoreDriver` be `"metastore"`?

cli/cmd/runtime/start.go
```diff
@@ -125,7 +125,13 @@ func StartCmd(cliCfg *config.Config) *cobra.Command {
...
 				AllowHostAccess:     conf.AllowHostAccess,
 				SafeSourceRefresh:   conf.SafeSourceRefresh,
-				GlobalDrivers:       parse(conf.GlobalDrivers),
+				SystemConnectors: []*runtimev1.Connector{
+					{
+						Type:   conf.MetastoreDriver,
+						Name:   "metastore",
+						Config: map[string]string{"dsn": conf.MetastoreURL},
+					},
+				},
```
begelundmuller, line 134: Should update `runtime.Options.MetastoreDriver` to be `"metastore"` now, right? And probably rename `runtime.Options.MetastoreDriver` to `runtime.Options.MetastoreConnector`?

begelundmuller (changes requested):

runtime/registry.go
k-anshul: The older reconcile still doesn't use `AcquireHandle` and opens connections directly. Once we move to new reconcile it can be removed.

runtime/connections.go
k-anshul: The `drivers.Open` take `map[string]any` but the connector configs are `map[string]string`. One tricky case is handling of `allow_host_access` which is passed directly as a bool and parsed as bool in drivers. At some point I imagined we the drivers should be able to access runtime options directly or use `mapstructure.WeakDecode` to handle such cases so that `drivers.Open` can take `map[string]string`. At that point we should be able to remove `connectorConfig` usages as well. Should we take this separately ?

runtime/connections.go
k-anshul: Answered in the comment below

admin/deployments.go
```diff
@@ -146,10 +156,25 @@ func (s *Service) updateDeployment(ctx context.Context, depl *database.Deploymen
...
+				c.Config = make(map[string]string)
+			}
+			c.Config["dsn"] = repoDSN
+			c.Type = repoDriver
+		}
+	}
+
 	_, err = rt.EditInstance(ctx, &runtimev1.EditInstanceRequest{
 		InstanceId: depl.RuntimeInstanceID,
 		RepoDriver: &repoDriver,
```
begelundmuller, line 176: This will change it from "repo" to "github"

proto/rill/runtime/v1/api.proto
```diff
@@ -347,23 +343,11 @@ message DeleteInstanceResponse {}
...
-      "metastore",
-      "github"
-    ]
-  }];
-  optional string repo_dsn = 5;
+  optional string olap_driver = 2;
+  optional string repo_driver = 4;
   optional bool embed_catalog = 6;
   optional int64 ingestion_limit_bytes = 8;
+  repeated Connector connectors = 9;
```
begelundmuller, line 350: For `EditInstance`, how will it know if this field is `optional` or nil? We might need a separate API like for variables

begelundmuller (approved):

PR #2876, opened 2023-08-03
check if local port is blocked

rakeshsharma14317:


begelundmuller (commented):
This seems good, but I wonder if there are some edge cases where this will break, such as in a Docker container. Can you provide some resources on best practices for the `net.Listen` address?

rakeshsharma14317:
Hey, Most of the places i found that using `net.DialTimeout` for checking the port availability  ref: [[](url)](https://github.com/projectdiscovery/naabu/blob/main/v2/pkg/scan/connect.go#L12) 
Or else using localIP with `net.Listen` for check with error msg for address already in use, ref: https://github.com/golang/go/issues/22269#issuecomment-336787071

Also there are some place where doing a request call to url as https://github.com/nknorg/go-portscanner/blob/8493ef01db79ae14a973300a38600c7400e265d8/portscanner.go#L69C1-L69C1

There is one more solution which i found to check using `cmd` if we wanted it inside docker container as well - example https://stackoverflow.com/questions/66576706/how-to-get-details-of-service-listening-on-a-port-in-go

But majorly using DialTimeout which used most (But it will be generic msg we should return saying `either port is in use of connection timeout`)

begelundmuller:
With help from @kaspersjo, the issue turned out to be that `net.Listen` will succeed if it acquires the port on IPv6 but not on IPv4. 

A side effect of hardcoding the localhost IP is that it only tries to get the port on IPv4, causing this PR to work. But it is not desirable to hardcode the IP.

Closing in favor of this PR which directly checks that the IPv4 port is available (without hardcoding the host address): https://github.com/rilldata/rill/pull/2893/files

PR #2877, opened 2023-08-03
Add scrub and zoom to timeseries charts

djbarnwal:
Add scrub and zoom functionality to the charts -

https://www.notion.so/rilldata/Enable-scrub-to-filter-on-measure-line-charts-5a662647d43e4e23937a709145bbdd9c#9f1481a1cc224d73877b7ea4f0b1606e

Items remaining (non blocker for functionality of this PR) - 

- [ ] Add zoom to scrub menu on right click
_The `WithTogglableFloatingElement` component needs a parent to attach to_  
- [ ] Change cursor on hover over edges and inside scrub
_Use the bisected mouseover value over the label accessor to find out if the cursor is inside scrub/over edges/outside_
- [ ] Handle right extremum snapping and bounds
_The last time grain in the charts is truncated to 55% of it's size to remove whitespace from the charts at the end. This results in a bad UX when scrubbed is snapped to the last grain_
- [ ] Use better colors for scrubbed area and lines
- [ ] Scrub across multiple chart


web-common/src/features/dashboards/time-series/MeasureScrub.svelte
```diff
@@ -0,0 +1,102 @@
...
+  import type { Writable } from "svelte/store";
+
+  export let start;
+  export let stop;
+  export let isScrubbing = false;
+  export let showLabels = false;
+  export let mouseoverTimeFormat;
+
+  const plotConfig: Writable<PlotConfig> = getContext(
+    "rill:data-graphic:plot-config"
```
skokenes, line 14: We should consider exporting this as a const somewhere

skokenes (changes requested):
I'm not sure that the solution conforms to the design spec?
1. The design says the interaction to start scrubbing is mousedown. However, in testing this branch I see that scrubbing only works with a shift+mousedown.
2. There appears to be a bug when changing an existing scrub. While moving/adjusting an existing scrub, the entire dashboard re-calculates to the overall, non-scrubbed time range. I believe the desired behavior is that while adjusting the scrub area, the numbers all stay the same until the new scrub area is complete. See video

https://github.com/rilldata/rill/assets/5554373/58c6de9b-338e-4111-ac43-571748cfb2b6

djbarnwal:
> I'm not sure that the solution conforms to the design spec?
> 
>     1. The design says the interaction to start scrubbing is mousedown. However, in testing this branch I see that scrubbing only works with a shift+mousedown.
> 
>     2. There appears to be a bug when changing an existing scrub. While moving/adjusting an existing scrub, the entire dashboard re-calculates to the overall, non-scrubbed time range. I believe the desired behavior is that while adjusting the scrub area, the numbers all stay the same until the new scrub area is complete. See video
> 

1. Scrubbing with mousedown (without shift) works but there are some edge cases which results in a bad UX. Any mousedown click event on the chart results in a `start-scrub` event (without dragging). This would probably need some changes in the scrub action factory. Have opted for shift click for reducing scope and getting the feature across. Will come back on this later.

2. Changed to have the desired behaviour.

web-common/src/features/dashboards/dashboard-stores.ts
```diff
@@ -518,4 +531,98 @@ export function useDashboardStore(
...
 }
 
+/***
+ * Dervied stores to get time range and comparison range to be
+ * used for fetching data. If we have a scrub range and
+ * isScrubbing is false, use that, otherwise use the selected
+ * time range
+ */
+
+let lastScrubRange = { start: null, end: null };
```
skokenes, line 541: Why the global state here? Is that going to cause weird behavior when switching between dashboards?

skokenes (changes requested):
If you create a scrub range, and then try to change the time range to some window that doesn't cover the scrub range, the time controls error out and stop responding.

https://github.com/rilldata/rill/assets/5554373/6f5bb5a6-7873-40f8-bbd5-cfad1499d768

web-common/src/features/dashboards/dashboard-stores.ts
djbarnwal: Fixed

web-common/src/features/dashboards/dashboard-stores.ts
```diff
@@ -518,4 +544,69 @@ export function useDashboardStore(
...
+
+      const comparisonRange = getComparionRangeForScrub(
+        entity.selectedTimeRange?.start,
+        entity.selectedTimeRange?.end,
+        entity.selectedComparisonTimeRange?.start,
+        entity.selectedComparisonTimeRange?.end,
+        start,
+        end
+      );
+      console.log(start.toISOString(), end.toISOString(), comparisonRange);
```
skokenes, line 597: Remove?

skokenes (approved):

bcolloran:
Hey Dhiraj, @nishantmonu51 asked me to do an experience pass over this. I think it looks great, only a couple things:

1) small edge case where setting/adjusting the filter area on a chart with missing data on some dates causes the chart line to flicker:
 https://github.com/rilldata/rill/assets/2380975/cf27efeb-39a3-4592-8c8f-45c5b475f4f6

2) it would be great to use the horizontal resize mouse cursor when hovering on the edge of a filter range, to make it a bit more discoverable that the range can be resized

I don't think either of those are blockers (I think I'd try to tackle (2) quickly in this PR since you're spun up on this now, but leave (1) until later), and I didn't spot any regressions. Nice work!

PR #2878, opened 2023-08-04
Runtime: Validate dimension and measure columns are unique

begelundmuller:
See [this bug report](https://rilldata.slack.com/archives/CTZ8XBQ85/p1691166232980139) in Slack for context

### DIFF
--- a/runtime/services/catalog/migrator/metricsviews/metrics_views.go
+++ b/runtime/services/catalog/migrator/metricsviews/metrics_views.go
@@ -75,8 +75,20 @@ func (m *metricsViewMigrator) Validate(ctx context.Context, olap drivers.OLAPSto
 
 	var validationErrors []*runtimev1.ReconcileError
 
+	columnNames := make(map[string]bool)
 	dimensionNames := make(map[string]bool)
 	for i, dimension := range mv.Dimensions {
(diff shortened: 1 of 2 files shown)


nishantmonu51 (approved):

runtime/services/catalog/migrator/metricsviews/metrics_views.go
```diff
@@ -75,8 +75,20 @@ func (m *metricsViewMigrator) Validate(ctx context.Context, olap drivers.OLAPSto
 
 	var validationErrors []*runtimev1.ReconcileError
 
+	columnNames := make(map[string]bool)
 	dimensionNames := make(map[string]bool)
 	for i, dimension := range mv.Dimensions {
+		if _, ok := columnNames[strings.ToLower(dimension.Column)]; ok {
```
AdityaHegde, line 81: Is this needed because of backwards compatibility or do we not allow for duplicate columns anymore?

AdityaHegde: Ah the API response is still returning column names instead of dimension name. We should perhaps change that?

begelundmuller: Yeah, we should definitely change that at some point :)

PR #2888, opened 2023-08-08
Adding local_file sql source support

AdityaHegde:
When we added sql sources we didnt fully support local files. Local files with path relative to repo root is not replaced.

This PR updates the local file code path to rewrite relative paths.

### DIFF
--- a/runtime/services/catalog/migrator/sources/sources.go
+++ b/runtime/services/catalog/migrator/sources/sources.go
@@ -145,13 +145,23 @@ func (m *sourceMigrator) Validate(ctx context.Context, olap drivers.OLAPStore, c
 }
 
 func (m *sourceMigrator) IsEqual(ctx context.Context, cat1, cat2 *drivers.CatalogEntry) bool {
-	if cat1.GetSource().Connector != cat2.GetSource().Connector {
+	// TODO: This is hopefully not needed in the new reconcile where parse and changing of connector happens before equals is called
+	isSQLSource := cat1.GetSource().Connector == "duckdb" && (cat2.GetSource().Connector == "s3" || cat2.GetSource().Connector == "gcs")
+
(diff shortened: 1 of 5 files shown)


runtime/services/catalog/migrator/sources/sources.go
```diff
@@ -317,6 +325,39 @@ func mergeFromParsedQuery(apiSource *runtimev1.Source) error {
...
+			if strings.Contains(p, "/") {
+				p, err := fileutil.ResolveLocalPath(p, repoRoot, allowRootAccess)
+				if err != nil {
+					resolveErr = err
+					return nil, false
+				}
+				newPaths = append(newPaths, p)
+			} else {
+				newPaths = append(newPaths, p)
+			}
```
begelundmuller, line 342: Why only when there's a `/` in the path? Wouldn't doing something like `select * from "xxx.csv"` load from process root instead of repo root in this case?

runtime/services/catalog/migrator/sources/sources.go
```diff
@@ -301,12 +302,19 @@ func mergeFromParsedQuery(apiSource *runtimev1.Source) error {
...
+	case "https":
 		return nil
+	case "local_file":
+		queryStr, err = rewriteLocalRelativePath(ast, repoRoot, strings.EqualFold(env["allow_host_access"], "true"))
+		if err != nil {
+			return err
+		}
+	default:
+		apiSource.Connector = c
+		props["path"] = p
```
begelundmuller, line 315: Maybe only do this for s3 and gcs specifically? And push to DuckDB in all unknown cases

begelundmuller (changes requested):
Can you also look into if DuckDB has some parameters we can set when `!allow_host_access` to limit reads outside of the repo directory? To improve security in cloud.

PR #2889, opened 2023-08-08
Support for setting local credentials in .env file

k-anshul:
High level working discussed offline: 

1. We add support for setting variables in .env to reconcile
2. Any connectors with missing credentials will fail in source ingestion. We could make sure to format the errors somewhat consistently, like Missing credentials for connector 'xxxx'.
3. The UI could expose editing of .env directly to keep things simple. We could provide an RPC ScanConnectors that returns hints the UI can show the user (it would be very similar to that used in rill env configure):
4. All explicit connectors (defined in rill.yaml) and implicit connectors
5. For each connector, whether anonymous access appears to be available.
6. For each connector, the variables that are required/optional for it (can be set in .env)


runtime/compilers/rillv1/parser.go
```diff
@@ -121,9 +121,10 @@ type Parser struct {
 	DuckDBConnectors []string
 
 	// Output
-	RillYAML  *RillYAML
-	Resources map[ResourceName]*Resource
-	Errors    []*runtimev1.ParseError
+	RillYAML     *RillYAML
+	Resources    map[ResourceName]*Resource
+	Errors       []*runtimev1.ParseError
+	EnvVariables map[string]string
```
begelundmuller, line 127:
```suggestion
	RillYAML  *RillYAML
	DotEnv    map[string]string
	Resources map[ResourceName]*Resource
	Errors    []*runtimev1.ParseError
```

runtime/compilers/rillv1/parse_dotenv.go
```diff
@@ -0,0 +1,26 @@
...
+			return nil
+		}
+		return err
+	}
+	envMap, err := godotenv.Unmarshal(data)
+	if err != nil {
+		return err
+	}
+
+	p.EnvVariables = envMap
```
begelundmuller, line 24:
```suggestion
	p.DotEnv = envMap
```

runtime/compilers/rillv1/parser.go
```diff
@@ -172,6 +173,11 @@ func Parse(ctx context.Context, repo drivers.RepoStore, instanceID string, duckD
 		return nil, err
 	}
 
+	// parse .env file separately
+	if err := p.parseDotEnv(ctx); err != nil {
+		return nil, err
+	}
```
begelundmuller, line 179: This is not sufficient – see how `rill.yaml` is handled:

1. Need to clear it in `Reparse` if path was changed 
2. Need to add bool for tracking modification to it in `Diff`
3. Should be parsed in `parsePaths` so it's parsed both in `Parse` and `Reparse`
4. Unlike `rill.yaml` (which returns an error), the returned error should be added to parse errors instead of being returned (like all other errors than `rill.yaml` errors)

proto/rill/runtime/v1/connectors.proto
```diff
@@ -197,3 +204,19 @@ message BigQueryListTablesResponse {
...
+
+message ScanConnectorsResponse {
+  repeated ScannedConnector connectors = 1;
+  map<string, string> existing_credentials = 2; // credentials already present in env file
+}
+
+message ScannedConnector {
+  string name = 1;
+  string type = 2;
+  repeated string variables = 3;
```
begelundmuller, line 220: 1. Can't this be extracted from the connector spec already? Feels like we should either return the spec or have a separate API for getting the specs (don't we have that already?)
2. We should not conflate connector config with variables. An instance's *variables* are one way to set connector *config* (through the `connector.name.xxx` convention), but can also be used for other things (like templating into code files). And connector config can be set in other ways than variables, e.g. in `rill.yaml`.

proto/rill/runtime/v1/connectors.proto
```diff
@@ -197,3 +204,19 @@ message BigQueryListTablesResponse {
...
   repeated string names = 2;
 }
+
+message ScanConnectorsRequest {
+  string instance_id = 1;
+}
+
+message ScanConnectorsResponse {
+  repeated ScannedConnector connectors = 1;
+  map<string, string> existing_credentials = 2; // credentials already present in env file
```
begelundmuller, line 214: I don't think this belongs here. See other note about connector config and variables. We should either return *resolved* connector config. Or have an instance-level API to lookup all variables.

begelundmuller (changes requested):

proto/rill/runtime/v1/connectors.proto
k-anshul: 1. By spec do you mean only the keys and not values. We have one to list all keys, other to list all credentials for a connector. 
2. Sure we have some APIs a combination of which can give this entire information. I can remove these fields.

PR #2890, opened 2023-08-09
Periodically emit DuckDB database metrics

### LINKED ISSUE #2848: Runtime: Periodically emit DuckDB database metrics

esevastyanov:
Closes #2848 

The PR updates a cached DuckDB connection so that it can emit DuckDB stats.
Each cached connection is cached per instance id and dsn and hence can autonomously collect and emit required stats per instance id and dsn.

### DIFF
--- a/runtime/caches.go
+++ b/runtime/caches.go
@@ -11,6 +11,7 @@ import (
 	"github.com/rilldata/rill/runtime/pkg/activity"
 	"github.com/rilldata/rill/runtime/pkg/observability"
 	"github.com/rilldata/rill/runtime/services/catalog"
+	"go.opentelemetry.io/otel/attribute"
 	"go.uber.org/zap"
 )
 
@@ -65,7 +66,7 @@ func (c *connectionCache) Close() error {
 	return firstErr
 }
 
-func (c *connectionCache) get(ctx context.Context, instanceID, driver, dsn string) (drivers.Connection, error) {
+func (c *connectionCache) get(ctx context.Context, instanceID, driver, dsn string, activityDims []attribute.KeyValue) (drivers.Connection, error) {
 	// TODO: This locks for all instances for the duration of Open and Migrate.
 	// Adapt to lock only on the lookup, and then on the individual instance's Open and Migrate.
 
@@ -86,7 +87,7 @@ func (c *connectionCache) get(ctx context.Context, instanceID, driver, dsn strin
 		config := map[string]any{
 			"dsn":          dsn,
 			"activity":     c.activity,
-			"activityDims": activity.GetDimsFromContext(ctx),
+			"activityDims": activityDims,
 		}
 		conn, err := drivers.Open(driver, config, logger)
 		if err != nil {
--- a/runtime/caches.go
+++ b/runtime/caches.go
@@ -84,10 +84,16 @@ func (c *connectionCache) get(ctx context.Context, instanceID, driver, dsn strin
 		if instanceID != "default" {
 			logger = c.logger.With(zap.String("instance_id", instanceID), zap.String("driver", driver))
 		}
+
+		activityClient := c.activity
+		if activityClient != nil {
+			activityClient = activityClient.With(activityDims...)
+		}
+
 		config := map[string]any{
-			"dsn":          dsn,
-			"activity":     c.activity,
-			"activityDims": activityDims,
+			"dsn": dsn,
+			// TODO: remove activity from the config and pass it as a func parameter, similar to the logger
+			"activity": activityClient,
 		}
 		conn, err := drivers.Open(driver, config, logger)
 		if err != nil {
--- a/runtime/caches.go
+++ b/runtime/caches.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/hashicorp/golang-lru/simplelru"
 	"github.com/rilldata/rill/runtime/drivers"
+	"github.com/rilldata/rill/runtime/pkg/activity"
 	"github.com/rilldata/rill/runtime/pkg/observability"
 	"github.com/rilldata/rill/runtime/services/catalog"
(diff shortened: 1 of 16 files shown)


runtime/caches.go
```diff
@@ -81,7 +84,12 @@ func (c *connectionCache) get(ctx context.Context, instanceID, driver, dsn strin
 		if instanceID != "default" {
 			logger = c.logger.With(zap.String("instance_id", instanceID), zap.String("driver", driver))
 		}
-		conn, err := drivers.Open(driver, map[string]any{"dsn": dsn}, logger)
+		config := map[string]any{
+			"dsn":          dsn,
+			"activity":     c.activity,
+			"activityDims": activityDims,
```
begelundmuller, line 90: 1. Similar to logging, could we add something lightweight like `c.activity.With(dims)` to avoid propagating the dims alongside?
2. Instead of passing it in the `config`, it would be cleaner to pass it along with the logger to `Open`. But let's do it later to avoid too many merge conflicts with this PR: https://github.com/rilldata/rill/pull/2875. Will add an issue for it.

runtime/drivers/duckdb/config.go
```diff
@@ -50,9 +58,11 @@ func newConfig(dsn string) (*config, error) {
...
 	// Return config
 	cfg := &config{
-		DSN:        dsn,
-		PoolSize:   poolSize,
-		DBFilePath: uri.Path,
+		DSN:          dsn,
+		PoolSize:     poolSize,
+		DBFilePath:   uri.Path,
+		Activity:     client,
+		ActivityDims: activityDims,
```
begelundmuller, line 65: Feels like this doesn't belong on the `config` struct but on the main handle, alongside the `logger`

runtime/drivers/duckdb/duckdb.go
```diff
@@ -56,7 +61,23 @@ func (d Driver) Open(config map[string]any, logger *zap.Logger) (drivers.Connect
...
+		}
+	}
+
+	var activityDims []attribute.KeyValue
+	activityDimsAny := config["activityDims"]
+	if activityDimsAny != nil {
+		if activityDims, ok = activityDimsAny.([]attribute.KeyValue); !ok {
+			return nil, fmt.Errorf("couldn't cast activity dimensions")
+		}
+	}
```
begelundmuller, line 78: Add comment that this is temporary until we add it to `drivers.Open`

runtime/drivers/duckdb/duckdb.go
```diff
@@ -409,3 +439,126 @@ func (c *connection) checkErr(err error) error {
...
+			if err != nil {
+				c.logger.Error("couldn't convert duckdb memory limit to bytes", zap.Error(err))
+			} else {
+				c.config.Activity.Emit(c.ctx, "olap_memory_limit_bytes", memoryLimit, commonDims...)
+			}
+
+			c.config.Activity.Emit(c.ctx, "olap_block_size_bytes", float64(stat.BlockSize), commonDims...)
+			c.config.Activity.Emit(c.ctx, "olap_total_blocks", float64(stat.TotalBlocks), commonDims...)
+			c.config.Activity.Emit(c.ctx, "olap_free_blocks", float64(stat.FreeBlocks), commonDims...)
+			c.config.Activity.Emit(c.ctx, "olap_used_blocks", float64(stat.UsedBlocks), commonDims...)
```
begelundmuller, line 507: Since these metrics are very duckdb-specific, consider changing the `olap_` prefix to `duckdb_`

runtime/drivers/duckdb/duckdb.go
```diff
@@ -409,3 +439,126 @@ func (c *connection) checkErr(err error) error {
...
+			statTicker.Stop()
+			return
+		}
+	}
+}
+
+// Reversed logic of StringUtil::BytesToHumanReadableString
+// see https://github.com/cran/duckdb/blob/master/src/duckdb/src/common/string_util.cpp#L157
+// Examples: 1 bytes, 2 bytes, 1KB, 1MB, 1TB, 1PB
+func humanReadableSizeToBytes(sizeStr string) (float64, error) {
```
begelundmuller, line 519: Is there any way to make duckdb return the values unformatted? Maybe a setting?

runtime/drivers/duckdb/duckdb.go
```diff
@@ -409,3 +439,126 @@ func (c *connection) checkErr(err error) error {
...
+	var multiplier float64
+
+	re := regexp.MustCompile(`^([\d.]+)\s*(\S+)$`)
```
begelundmuller, line 522: Move regex compilation to a global before the function to avoid recompiling it every time

runtime/drivers/duckdb/duckdb.go
```diff
@@ -409,3 +439,126 @@ func (c *connection) checkErr(err error) error {
...
+func (c *connection) periodicallyEmitStats(d time.Duration) {
+	if c.config.Activity == nil {
+		// Activity client isn't set, there is no need to report stats
+		return
+	}
+
+	statTicker := time.NewTicker(d)
+	for {
+		select {
+		case <-statTicker.C:
```
begelundmuller, line 453: Can you throw in a measure for the size observed on the file system using `c.EstimateSize()`? Just in case DuckDB's metadata isn't completely honest about things

begelundmuller (changes requested):

runtime/caches.go
esevastyanov: Added a lightweight `wrappedClient` to hold dimension

runtime/drivers/duckdb/duckdb.go
esevastyanov: Added `duckdb_estimated_size_bytes` metric for this purpose

begelundmuller (approved):

PR #2891, opened 2023-08-09
Adding support for http sql sources

AdityaHegde:
Http sources go into `fileStore_to_duckDB`. SQL sources handling is missing for this code path. This PR adds it by checking for `sql`.


AdityaHegde:
Changes made in #2888

PR #2895, opened 2023-08-10
feat: TDT async data 

skokenes:
## Checklist
- [x] Manual verification
- [ ] Unit test coverage
- [ ] E2E test coverage
- [ ] Needs manual QA?

#### Issue addressed: 
Incremental step in adding the [time dimension details table](https://www.notion.so/rilldata/Time-Dimension-Detail-Design-Mocks-and-Specification-2dd22773d6e8479ca8d297f95877c669)

This PR adds async data fetching

#### Details:
Each cell of the table now has the ability to fetch its data via svelte-query. The cells look up which "block" of data they are in, then request that block. In this initial PR, we set the block size to 100, so data will be fetched in pages of 0-99, 100-199, etc. We can play with this number over time to find the best performing result. Svelte-query handles deduping these requests and serving from cache, so when many cells request the same block, they all get the same page.

The data fetching in this PR is just mocking a network call, with a configurable delay set to 1s. Once we have real backend APIs for TDT, we can swap this out.

### DIFF
--- a/web-common/src/features/dashboards/time-dimension-details/util.ts
+++ b/web-common/src/features/dashboards/time-dimension-details/util.ts
@@ -1,6 +1,6 @@
 // given area, return block to fetch
 export const getBlock = (blockSize: number, start: number, end: number) => {
-  // If distance is bigger than possible block, error?
+  // If distance is bigger than possible block, throw an error
   if (end - start > blockSize) {
     throw new Error("Range is too big");
   }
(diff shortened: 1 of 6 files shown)


web-common/src/features/dashboards/time-dimension-details/util.ts
```diff
@@ -0,0 +1,18 @@
+// given area, return block to fetch
+export const getBlock = (blockSize: number, start: number, end: number) => {
```
ericpgreen2, line 2: Adding a return type would both help the reader better understand what a "block" is and help IntelliSense where this function is consumed.

ericpgreen2 (approved):
👍

PR #2901, opened 2023-08-11
Execute `git reset` in case of `fast-forward` is unavailable

egor-ryashin:


runtime/drivers/github/github.go
```diff
@@ -232,6 +232,24 @@ func (c *connection) pullUnsafe(ctx context.Context) error {
 	})
 	if errors.Is(err, git.NoErrAlreadyUpToDate) {
 		return nil
+	} else if errors.Is(err, git.ErrNonFastForwardUpdate) {
+		head, err := repo.Head()
+		if err != nil {
+			return err
+		}
+
+		rev, err := repo.ResolveRevision(plumbing.Revision("remotes/origin/" + head.Name().Short()))
```
begelundmuller, line 241: Is the remote guaranteed to be called `origin`? (I think so, just to be sure)

begelundmuller (approved):

PR #2908, opened 2023-08-11
2849: org and project name annotations for usage tracking

### LINKED ISSUE #2849: Admin/Runtime: Add org and project name annotations for usage tracking

esevastyanov:
Closes #2849 

Creates a method `EditInstanceAnnotations` of `RuntimeService` that is called if annotations were updated (organization or project names were updated). This doesn't trigger the reconcile.
If a name of an organization is updated then `EditInstanceAnnotations` is called for every deployment related to the organization.


admin/projects.go
```diff
@@ -247,12 +256,37 @@ func (s *Service) UpdateProjectVariables(ctx context.Context, proj *database.Pro
 	return s.DB.UpdateProjectVariables(ctx, proj.ID, variables)
 }
 
+// UpdateProjectAnnotations pushes the updated annotations to deployments.
+// NOTE : this does not trigger reconcile.
+func (s *Service) UpdateProjectAnnotations(ctx context.Context, proj *database.Project, annotations map[string]string) error {
```
begelundmuller, line 261: 1. I was confused about what "project annotations are – maybe `UpdateDeploymentAnnotations` would be a better name?
2. Instead of accepting a free-form `map` here, I think it should take an org and project object, and then the function should internally construct the payload we expect

admin/server/organizations.go
```diff
@@ -186,6 +186,33 @@ func (s *Server) UpdateOrganization(ctx context.Context, req *adminv1.UpdateOrga
...
+			break
+		}
+
+		for _, project := range projects {
+			err := s.admin.UpdateProjectAnnotations(ctx, project, map[string]string{
+				"organization_id":   org.ID,
+				"organization_name": org.Name,
+				"project_id":        project.ID,
+				"project_name":      project.Name,
+			})
```
begelundmuller, line 207: We should only have one place anywhere that we construct this payload, so it doesn't go out of sync (right now, it's also created in `createDeployment`)

begelundmuller:
Another issue – I think all open connections for the instance would need to re-created to propagate the updated annotations in the runtime when this is merged: https://github.com/rilldata/rill/pull/2890.

This PR should have a connection cache eviction function that could be used: https://github.com/rilldata/rill/pull/2875

esevastyanov:
>This PR should have a connection cache eviction function that could be used: https://github.com/rilldata/rill/pull/2875

@begelundmuller Should I wait till #2875 is merged? I can utilize existing `evict` function
```func (c *connectionCache) evict(ctx context.Context, instanceID, driver, dsn string) bool {```

**Comment by esevastyanov** [comment]
[c3f1857](https://github.com/rilldata/rill/pull/2908/commits/c3f1857d0cb3cf6c45c7af8c4ae984fa0760ba3b) evicts caches

File: admin/projects.go [hunk 3]
```diff
@@ -218,11 +225,25 @@ func (s *Service) UpdateProject(ctx context.Context, proj *database.Project, opt
...
+		org, err := s.DB.FindOrganization(ctx, proj.OrganizationID)
+		if err != nil {
+			return nil, err
+		}
+
+		err = s.UpdateDeploymentAnnotations(ctx, org, proj)
+		if err != nil {
+			return nil, err
+		}
+	}
```
begelundmuller, line 245: 1. `requireNewDeployments` can never be true (it returns early if that's the case)
2. What about the case where a project's branch and name is updated in the same request? I'm not sure `impactsDeployments` and `projNameUpdated` are mutually exclusive

admin/server/organizations.go
```diff
@@ -186,6 +187,13 @@ func (s *Server) UpdateOrganization(ctx context.Context, req *adminv1.UpdateOrga
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 
+	if orgName != org.Name {
```
begelundmuller, line 190: This will never be true – it's comparing on the result value (after the update)

begelundmuller (changes requested):

admin/server/organizations.go
esevastyanov: I moved this check and annotations update up so that the organization is updated in db at the end of the flow. This fixes the issue and the base concept

admin/projects.go
esevastyanov: I introduced annotations into `EditInstanceRequest` so that annotations are updated (if needed) if `impactsDeployments == true`
`UpdateDeploymentAnnotations` is called only if annotations changed and `impactsDeployments == false`

begelundmuller (approved):

PR #2915, opened 2023-08-14
Adding sql sources parsing to new reconcile

AdityaHegde:
Adding sql sources parsing to the new reconcile. Actual ingestion changes is not added yet since some parts are not coded in yet.


begelundmuller (commented):
Thanks for this PR – instead of doing the path rewriting in the parser, which seems to lack some execution context, such as `allow_root_access`, is there a way we could push this rewriting into the DuckDB driver itself, or alternatively into the source reconciler as a DuckDB-specific rewrite at execution time?

begelundmuller:
Thanks for this PR – used as inspiration, but pushed the entire rewrite into the DuckDB transporter for now. I think it might make sense to bring some object store planning awareness up into the reconciler/parser at some point, but that'll be a separate project. 

Closing in favor of https://github.com/rilldata/rill/pull/3082

PR #2920, opened 2023-08-14
Batching Modeller Profiling Queries

AdityaHegde:
## Checklist
- [x] Manual verification
- [ ] Unit test coverage
- [x] E2E test coverage
- [x] Needs manual QA?

#### Issue addressed: 
With large profiles the UI slows down significantly.

#### Details:
Integrating the batching API for profiling queries.

## Steps to Verify
1. Import a source with significantly large columns.
2. Play around with the modeller based on this source.
3. The UX should be smoother than on main.


ericpgreen2:
QA finding: when adding/removing a `WHERE` clause to filter a dataset, the row count in the top-right-hand-corner sometimes doesn't update.

web-common/src/lib/create-throttler.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Run callback once every `time`ms
+ */
+export function createThrottler(time: number) {
+  let timeout: ReturnType<typeof setTimeout>;
+  return (callback: () => void) => {
+    if (timeout) return;
+    timeout = setTimeout(() => {
+      console.log(new Date().toISOString(), "sort");
```
ericpgreen2, line 9: Leftover console log

web-common/src/runtime-client/streaming-fetch-wrapper.ts
ericpgreen2, line 1: Isn't this on `main`? It looks like you might be renaming the file, but need to delete the original `fetch-streaming-wrapper.ts`

web-common/src/lib/create-throttler.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Run callback once every `time`ms
+ */
+export function createThrottler(time: number) {
```
ericpgreen2, line 4: Might be better if this function also accepted the `callback` as a parameter, so the component code could be a one-liner like:
```
throttle(myfunc(), 500)
```

web-common/src/components/column-profile/columns-profile-data.ts
```diff
@@ -0,0 +1,261 @@
...
+};
+export type ColumnsProfileDataMethods = {
+  load: (
+    instanceId: string,
+    tableName: string,
+    profileColumnResponse: QueryObserverResult<V1TableColumnsResponse>
+  ) => Promise<void>;
+};
+export type ColumnsProfileDataStore = Readable<ColumnsProfileData> &
+  ColumnsProfileDataMethods;
```
ericpgreen2, line 73: All this data/state looks like server-side data, which should be managed by TanStack Query. Can we use the TanStack stores rather than syncing the data into our own store?

ericpgreen2:
Curious – did you look at [TanStack's documentation for batching queries](https://tanstack.com/query/v4/docs/react/community/batching-requests-using-bathshit)? Thoughts on that approach?

web-common/src/components/column-profile/columns-profile-data.ts
AdityaHegde: We would still need derived stores, check the files that were deleted. This is a simpler approach than that. Besides it is simpler for the fetcher code to update these stores.

AdityaHegde:
The batching API they have doesnt support streaming response. One of the main goals is to still have incremental UI updates but the requests are batched.

web-common/src/lib/create-throttler.ts
AdityaHegde: Check the usage in `ColumnProfile.svelte`, The entire call is within a reactive statement. If we had the callback during instantiation we would need to manually add all dependants in the reactive statement. The way it is right now, it is simpler to use in a component.

web-common/src/runtime-client/batched-request.ts
```diff
@@ -0,0 +1,96 @@
...
+import { DefaultQueryPriority } from "@rilldata/web-common/runtime-client/http-request-queue/priorities";
+import { streamingFetchWrapper } from "@rilldata/web-common/runtime-client/streaming-fetch-wrapper";
+
+export type BatchRequest = {
+  request: V1QueryBatchEntry;
+  priority: number;
+  resolve: (data: V1QueryBatchResponse) => void;
+  reject: (err: Error) => void;
+};
+export class BatchedRequest {
```
ericpgreen2, line 15: I'd expect a Batcher to self-manage when to fire HTTP requests. Like, users of this class should only be calling `batcher.add(myRequest)`, not `batcher.send()`.

Internal to this class, I'd expect to see logic for when a batch is full or when a time window has passed – after which the request would be issued. `send()` would be a private method, not a public one (unless it's a public `forceSend()`).

web-common/src/runtime-client/batched-request.ts
```diff
@@ -0,0 +1,96 @@
...
+  private expectedRequests = 0;
+  private controller: AbortController;
+
+  public get ready() {
+    return this.requests.length >= this.expectedRequests;
+  }
+
+  public register() {
+    this.expectedRequests++;
+  }
```
ericpgreen2, line 26: I don't see these methods used anywhere

web-common/src/components/column-profile/ColumnProfileProvider.svelte
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  $: if ($profileColumns?.data && !$profileColumns?.isFetching) {
+    columnsProfile.load(
+      $runtime?.instanceId,
+      objectName,
+      $profileColumns,
+      referencedThings
+    );
+  }
```
ericpgreen2, line 63: This sort of logic for when to sync the TanStack store with the `columnsProfile` store is pretty sneaky/fragile. I've liked how you've used `derived` stores in the past – I feel like that approach would be more robust?

web-common/src/features/sources/workspace/SourceWorkspace.svelte
```diff
@@ -13,7 +14,9 @@
     <WorkspaceContainer assetID={sourceName}>
       <SourceWorkspaceHeader {sourceName} slot="header" />
       <SourceWorkspaceBody {sourceName} slot="body" />
-      <SourceInspector {sourceName} slot="inspector" />
+      <ColumnProfileProvider objectName={sourceName} slot="inspector">
```
ericpgreen2, line 17: `ColumnProfileProvider` should go within the `SourceInspector`, not outside it.

web-common/src/components/column-profile/ColumnProfileProvider.svelte
```diff
@@ -0,0 +1,66 @@
...
+    createQueryServiceTableColumns,
+    createRuntimeServiceListCatalogEntries,
+    V1TableColumnsResponse,
+  } from "@rilldata/web-common/runtime-client";
+  import type { V1CatalogEntry } from "@rilldata/web-common/runtime-client/index.js";
+  import { runtime } from "@rilldata/web-common/runtime-client/runtime-store";
+  import type { CreateQueryResult } from "@tanstack/svelte-query";
+
+  export let objectName: string;
+  export let sql = "";
```
ericpgreen2, line 18: It'd be a cleaner API if the SQL were fetched within the component, not passed in

AdityaHegde:
Will revisit this. Too many changes are made to merge main.

PR #2927, opened 2023-08-16
added pagination for listing user/orgs/projects

rakeshsharma14317:
Kept the default limit for user/org and project listing as `50` and can be listed by adding flag `--page-size`

### DIFF
--- a/cli/cmd/user/list.go
+++ b/cli/cmd/user/list.go
@@ -38,12 +38,13 @@ func ListCmd(cfg *config.Config) *cobra.Command {
 						return err
 					}
 				} else {
-					err = listProjectMembers(cmd, client, cfg.Org, projectName, strings.TrimPrefix(pageToken, "usr"), pageSize)
+					err = listProjectMembers(cmd, client, cfg.Org, projectName, pageToken, pageSize)
 					if err != nil {
 						return err
 					}
 
-					err = listProjectInvites(cmd, client, cfg.Org, projectName, strings.TrimPrefix(pageToken, "inv"), pageSize)
+					cmd.Println()
+					err = listProjectInvites(cmd, client, cfg.Org, projectName, pageToken, pageSize)
 					if err != nil {
 						return err
 					}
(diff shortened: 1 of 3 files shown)


cli/cmd/user/list.go
```diff
@@ -68,6 +97,8 @@ func ListCmd(cfg *config.Config) *cobra.Command {
 
 	listCmd.Flags().StringVar(&cfg.Org, "org", cfg.Org, "Organization")
 	listCmd.Flags().StringVar(&projectName, "project", "", "Project")
+	listCmd.Flags().Uint32Var(&pageSize, "page-size", 50, "Number of users to return per page")
+	listCmd.Flags().StringVar(&pageToken, "page-token", "", "Pagination token")
```
begelundmuller, line 101: This would need separate page tokens for members and invites.

To prevent placing that burden on the user, I would propose prepending the token for each with `usr` or `inv`. Then when a page token is passed, you can look at that prefix to determine whether to paginate on users or invites.

When paginating on either users or invites, it should not load the other (only the one being paginated on).

If you use some well-named util functions, it should be possible to make this pretty clean.

begelundmuller (changes requested):

begelundmuller (changes requested):
Also fix CI errors

cli/cmd/user/list.go
```diff
@@ -23,42 +27,54 @@ func ListCmd(cfg *config.Config) *cobra.Command {
...
+						return err
+					}
 
-				invites, err := client.ListProjectInvites(context.Background(), &adminv1.ListProjectInvitesRequest{
-					Organization: cfg.Org,
-					Project:      projectName,
-				})
-				if err != nil {
-					return err
+					cmd.Println()
```
begelundmuller, line 46: This will cause a dangling newline in cases where there are no invites or no members

begelundmuller (changes requested):

begelundmuller (approved):

PR #2932, opened 2023-08-16
Add leaderboard sort direction toggling

bcolloran:
Adds sort direction toggling to the leaderboards. As of this PR:
- sort order and sorting measure are maintained between leaderboard and detail table
- sort indication icon added in leaderboard
- sort order can be toggled by clicking leaderboard sort icon

Not yet present:
- sort by percent of total
- sort by absolute change over time
- sort by percentage change over time
- sort dimension table by dimension
- updates to leaderboard context column menu styling


AdityaHegde:
@bcolloran I have added the fields to the url. I also updated the PR to use the generated enums rather than the custom ones. This avoids unnecessary conversions.

AdityaHegde (approved):

proto/rill/ui/v1/dashboard.proto
```diff
@@ -7,11 +7,48 @@ import "rill/runtime/v1/queries.proto";
 
 // DashboardState represents the dashboard as seen by the user
 message DashboardState {
-  enum DashboardLeaderboardContextColumn {
-    DASHBOARD_LEADERBOARD_CONTEXT_COLUMN_UNSPECIFIED=0;
-    DASHBOARD_LEADERBOARD_CONTEXT_COLUMN_PERCENT = 1;
-    DASHBOARD_LEADERBOARD_CONTEXT_COLUMN_DELTA_CHANGE = 2;
-    DASHBOARD_LEADERBOARD_CONTEXT_COLUMN_HIDDEN = 3;
+  enum LeaderboardContextColumn {
+    LEADERBOARD_CONTEXT_COLUMN_UNSPECIFIED=0;
```
bcolloran, line 11: @AdityaHegde -- can you figure out a way to remove the `_UNSPECIFIED` option on this and the other enums? This new options change the semantics of these enums -- in each case, the enums had previously enumerated all the valid states of the dashboard, but now an invalid state is included.

web-common/src/features/dashboards/dashboard-stores.ts
```diff
@@ -107,10 +74,10 @@ export interface MetricsExplorerEntity {
 
   // This is the sort type that will be used for the leaderboard
   // and dimension detail table. See SortType for more details.
-  dashboardSortType: SortType;
+  dashboardSortType: DashboardState_LeaderboardSortType;
```
bcolloran, line 77: @AdityaHegde -- for readability, can you change the imports for these renamed enums to `import {DashboardState_LeaderboardSortType as SortType} from ...` so we don't have these long ugly strings all over the place?

(and likewise of course for `DashboardState_LeaderboardSortDirection`)

Please update this in all files, thanks!

bcolloran:
@AdityaHegde Thanks for making those changes! I noticed a couple things that could be improved, can you make those updates before we merge this? Thanks so much!

proto/rill/ui/v1/dashboard.proto
bcolloran: Either that, or we might need separate enums to track the valid states in the URL data and in the actual dashboard, and/or a more sophisticated translation layer. I just don't want the `*_UNSPECIFIED` semantics to have any possibility of leaking into the actual product state

web-common/src/features/dashboards/dashboard-stores.ts
bcolloran: that would also reduce the merge conflicts for #2957 and #2951

bcolloran: I went ahead and made this change in this commit, https://github.com/rilldata/rill/pull/2932/commits/a420c865d4424f29e2554f85dfffe26455223c73, please double check it to make sure that looks ok to you :-)

AdityaHegde: Thanks!

PR #2933, opened 2023-08-17
Fixing init dashboard with undefined comparision

AdityaHegde:
## Checklist
- [x] Manual verification
- [ ] Unit test coverage
- [ ] E2E test coverage
- [x] Needs manual QA?

#### Issue addressed: 
When a dashboard is 1st opened comparison menu shows undefined.

#### Details:
Explicitly setting `showComparison` to false. Ideally we also need some checks in the time comparison component.

## Steps to Verify
1. Refresh when in a dashboard with timestamp
2. Comparison option should show `no comparision`


web-common/src/features/dashboards/dashboard-stores.ts
```diff
@@ -366,6 +370,7 @@ const metricViewReducers = {
 
   displayComparison(name: string, showComparison: boolean) {
     updateMetricsExplorerByName(name, (metricsExplorer) => {
+      console.log("displayComparison", showComparison);
```
djbarnwal, line 373: Remove this stale log

djbarnwal (approved):
There's a stale console log which should be removed. Approving so that post the fix, the PR can be merged.

PR #2940, opened 2023-08-17
feat: TDD rendering 2

skokenes:
## Checklist
- [x] Manual verification
- [ ] Unit test coverage
- [ ] E2E test coverage
- [ ] Needs manual QA?

#### Issue addressed: 
Adds more rendering changes for the TDD

#### Details:
- Vertical scroll is now contained within the body of the table; the header stays above the scroll
- Table rows handle basic filter state. Filter checks are shown outside of the dimension cell to the left. Padding is added to the table contents left side to make room for these overflowing filter checks
- Some refactoring of how components and constants are defined
- "tdt" references renamed to "tdd" (Time Dimension Details) for consistency

https://github.com/rilldata/rill/assets/5554373/fdc40514-e55a-462f-a683-f261f7f3f529

## Steps to Verify
Go to localhost:3000/dev/table/a


ericpgreen2 (approved):
👍

skokenes:
> Looks good. I noticed a lot of warnings in the browser console though.

Yea, thats an annoying Svelte dev thing where it warns if you have a component called with a property that it doesn't expect as input. In this case, we are passing in user-defined components for what to render the cells, and sometimes our components don't care about every property that the Pivot wrapper may offer that cell to consume.

I don't think it will have any effect in production, but if we want to get rid of these dev warnings, I believe we could take each possible custom cell component and add in `export const <pivotProp>;`, even though the prop is never used.

PR #2946, opened 2023-08-17
Async reconcile event loop

begelundmuller:
- Implements an *instance controller* in `runtime/controller.go` which schedules reconcile calls based on the resource DAG using an event loop. See `runtime/reconcilers/README.md` for details about the scheduling semantics, and see the docstring for `Controller.Run` for a starting point for information about how the scheduling is implemented.
- Implements a catalog cache in `runtime/catalog_cache.go` that caches resources and buffers writes in-memory, enabling fast and frequent catalog reads and writes. Writes are periodically flushed to disk (and can be manually triggered by calling `flush`).
- Defines and implements new driver functions for persisting async reconcile resources in the DuckDB catalog. When we have fully migrated to async reconcile, we can remove the old catalog functions.


AdityaHegde (approved):
Looks good at a high level. We can revisit once we add tests and find any edge cases.

PR #2951, opened 2023-08-18
Change context column toggle buttons to dropdown

bcolloran:
implement requirement described here: https://www.notion.so/rilldata/Expand-sorting-options-for-leaderboards-dimension-detail-tables-b4ec4ac635fd4634afff0c21f9779da4?pvs=4#c001bec5bbf94e6889e82933b4c8f709


web-common/src/components/menu/compositions/SelectMenu.svelte
```diff
@@ -7,22 +7,15 @@ A slot is provided to change the text within the button.
...
 
-  export let options;
-  export let selection;
-  export let tailwindClasses = undefined;
-  export let activeTailwindClasses = undefined;
-  /** When true, will make the trigger element a block-level element.
-   * This is most useful when embedding a select menu in a table or wherever
-   * a block-level treatment is needed.
-   */
-  export let block = false;
```
djbarnwal, line 22: Any reason why we are removing these props?

djbarnwal (dismissed):
Couple of UI related feedback - 

image

The check mark looks a bit too small. Probably we can have similar styling as that of the dimension selector -

image



When the sidebar is open, for some dashboards the selector is partly rendered outside of my screen (I am on a 14inch Mac) -
image
We probably need to make the selector bar responsive.

web-common/src/components/menu/compositions/SelectMenu.svelte
bcolloran: They are not used anywhere, so simplifying the component -- it's needless over-engineering and confusing to consumers of an API to have a bunch of extraneous options (most of which are undocumented and untyped).

bcolloran:
@djbarnwal -- good catch regarding the check mark size. But that's actually a pre-existing issue (see e.g the timezone selector and the "metrics trends by ..." selector). @jkhwu and I had already been discussing revamping all of the selector menus, so I'm going to hold off on unifying all of these until I get her input on which style we prefer -- [this is tracked in notion here](https://www.notion.so/rilldata/Standardize-UI-component-library-02c66f1361d64333a23fd50bdb353e36?pvs=4#11ac4728539b4c48b7e7c3dbee4c5dfe), so not ignoring it, I just don't have the information to make the right choice at this point, so I don't think this is a blocker

ericpgreen2 (approved):

PR #2956, opened 2023-08-22
Re-structure scrub methods

djbarnwal:
Restructure scrub and zoom so that it can be reused at other places such as Time Detailed View.

This PR moves scrub methods from `MeasureChart` to `MeasureScrub` and fixes - 
1. Glitch animation when scrubbing
2. Removes scrub when mouse is released outside the charts

### DIFF
--- a/web-common/src/components/data-graphic/utils.ts
+++ b/web-common/src/components/data-graphic/utils.ts
@@ -58,6 +58,24 @@ export function pathDoesNotDropToZero(yAccessor: string) {
   };
 }
 
+export interface PlotConfig {
+  top: number;
+  bottom: number;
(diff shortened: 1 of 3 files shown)


ericpgreen2:
Huh, the scrub isn't working for me. When I select a sub-range and release the mouse, the sub-range instantly gets cleared. I'm on Chrome.

djbarnwal:
> Huh, the scrub isn't working for me. When I select a sub-range and release the mouse, the sub-range instantly gets cleared. I'm on Chrome.

Thanks for the review. Turns out some of the DOM APIs I used were specific to Gecko browsers. Fixed now.

PR #2957, opened 2023-08-22
Add absolute change option in context column

bcolloran:
adds an "absolute change" option to the context menu.

requirement here: https://www.notion.so/rilldata/Expand-sorting-options-for-leaderboards-dimension-detail-tables-b4ec4ac635fd4634afff0c21f9779da4?pvs=4#116b22b05ac54bb5b02a5c3b61641290

NOTE: this should merge after  #2951


web-common/src/features/dashboards/dashboard-stores.ts
```diff
@@ -492,14 +492,24 @@ const metricViewReducers = {
...
-      // NOTE: only show delta change if comparison is enabled
+      // NOTE: only show delta change if time comparison is enabled
       if (metricsExplorer.showComparison === false) return;
 
       metricsExplorer.leaderboardContextColumn =
         LeaderboardContextColumn.DELTA_PERCENT;
     });
   },
 
+  displayDeltaAbsolute(name: string) {
```
AdityaHegde, line 503: I am not sure if having a dedicated method for every option makes sense. Could we perhaps take the LeaderboardContextColumn value as argument?

web-common/src/features/dashboards/leaderboard/LeaderboardContextColumnMenu.svelte
```diff
@@ -26,6 +26,8 @@
       metricsExplorerStore.displayDeltaChange(metricViewName);
     } else if (key === LeaderboardContextColumn.PERCENT) {
       metricsExplorerStore.displayPercentOfTotal(metricViewName);
+    } else if (key === LeaderboardContextColumn.DELTA_ABSOLUTE) {
```
AdityaHegde, line 29: This if-else chain can be moved over to switch-case. Easy to add more options. Applies to code in LeaderboardHeader and leaderboard-utils as well.

AdityaHegde (approved):
Looks good apart from a couple of small nits.

bcolloran:
Thanks for the review @AdityaHegde , I made the changes you requested and will merge once CI passes

PR #2964, opened 2023-08-23
added --clean flag for rill start

rakeshsharma14317:


cli/cmd/start/start.go
```diff
@@ -36,6 +37,25 @@ func StartCmd(cfg *config.Config) *cobra.Command {
...
+					if err != nil {
+						return err
+					}
+
+					err = deleteFilesWithExtensions(currentDir, ".db", ".wal")
+					if err != nil {
+						return err
+					}
+				}
+			}
```
begelundmuller, line 57: 1. You need to push the reset down into the implementation of `local.NewApp` because the correct DB path won't be resolved until then (it becomes relative to `projectPath` in certain cases).
2. Use `drivers.Drop(...)` for dropping the DB

begelundmuller (changes requested):

cli/cmd/start/start.go
```diff
@@ -124,6 +125,7 @@ func StartCmd(cfg *config.Config) *cobra.Command {
 	startCmd.Flags().BoolVar(&noUI, "no-ui", false, "Serve only the backend")
 	startCmd.Flags().BoolVar(&verbose, "verbose", false, "Sets the log level to debug")
 	startCmd.Flags().BoolVar(&strict, "strict", false, "Exit if project has build errors")
+	startCmd.Flags().BoolVar(&reset, "reset", false, "Reset project before build")
```
begelundmuller, line 128: Change the description to `"Clear and re-ingest source data"`

cli/pkg/local/app.go
```diff
@@ -146,6 +146,13 @@ func NewApp(ctx context.Context, ver config.Version, verbose bool, olapDriver, o
 		return nil, err
 	}
 
+	if reset {
+		err := drivers.Drop(olapDriver, map[string]any{"dsn": olapDSN}, logger)
+		if err != nil {
+			return nil, fmt.Errorf("failed to clean OLAP: %w", err)
+		}
+	}
```
begelundmuller, line 154: 1. This should run before creating the instance (around line 122)
2. Put the olapConfig into a new variable and use it here and on line 140 to avoid duplication

runtime/drivers/duckdb/duckdb.go
```diff
@@ -126,6 +126,8 @@ func (d Driver) Drop(config map[string]any, logger *zap.Logger) error {
 		return fmt.Errorf("require dsn to drop duckdb connection")
 	}
 
+	fmt.Println("Dropping DuckDB database at", dsn)
```
begelundmuller, line 129: Remove print statement

begelundmuller (changes requested):

begelundmuller (approved):

PR #2966, opened 2023-08-23
docs for granular access policies

pjain1:


docs/docs/reference/project-files/dashboards.md
```diff
@@ -43,3 +43,9 @@ _**`measures`**_ — numeric [aggregates](/develop/metrics-dashboard#measures) o
...
+  - _**`has_access`**_ - Condition to check if user has access to the dashboard. If not defined, it will resolve to `false` meaning dashboard won't be accessible to anyone. _(optional)_
+  - _**`filter`**_ - Condition to filter the data as per user/custom attributes. It needs to be a valid SQL `WHERE` clause. It will be injected in the queries after resolving the user attributes templating. _(optional)_
+  - _**`exclude`**_ - List of columns to exclude from the dashboard. Each column can have an `if` condition to check if the column should be excluded or not. If the condition is not defined, it's an error. If the condition is defined, it will be resolved after resolving the user attributes templating. If `exclude` is defined all other columns are included. _(optional)_
+  - _**`include`**_ - List of columns to include in the dashboard. Each column can have an `if` condition to check if the column should be included or not. If the condition is not defined, it's an error. If the condition is defined, it will be resolved after resolving the user attributes templating. If `include` is defined all other columns are excluded. _(optional)_
```
begelundmuller, line 51: 1. It would be nice to have sub-bullets here for each allowed nested key, like for `measures` and others
2. Also link to the syntax for the `if` fields

docs/docs/reference/project-files/dashboards.md
```diff
@@ -43,3 +43,9 @@ _**`measures`**_ — numeric [aggregates](/develop/metrics-dashboard#measures) o
       - _`currency_usd`_ —  output rounded to 2 decimal points prepended with a dollar sign
       - _`percentage`_ — output transformed from a rate to a percentage appended with a percentage sign
       - _`comma_separators`_ — output transformed to decimal formal with commas every 3 digits
+
+_**`policy`**_ - Optionally define granular access policies for the dashboard. See [Secure: Dashboards](../../develop/policies) for more details.
+  - _**`has_access`**_ - Condition to check if user has access to the dashboard. If not defined, it will resolve to `false` meaning dashboard won't be accessible to anyone. _(optional)_
```
begelundmuller, line 48: Add link to syntax

begelundmuller (commented):
I would propose adding a separate reference page for "Templating and expression syntax" that describes the difference between the two, has a few examples, and links to the source docs for each.

It's okay for that page to be quite short, would just be nice to have something to link to.

pjain1:
> I would propose adding a separate reference page for "Templating and expression syntax" that describes the difference between the two, has a few examples, and links to the source docs for each.


This section is just that - https://github.com/rilldata/rill/pull/2966/files#diff-c16b1a7d0b355c9c5c507f75e353162075743fe93cdbcdc04c67830850969dfcR75-R94. I explains the two phases for policy resolution - templating and expression evaluation. Then links to function/operators supported for both and the some examples. Should I elaborate more on it or should just put it in a different file ?

docs/docs/develop/policies.md
```diff
@@ -0,0 +1,118 @@
...
+---
+
+:::caution
+
+This is an experimental feature.
+
+:::
+
+Rill supports granular access policies for dashboards. You can define access policies for dashboards in the respective `dashboard_name.yaml` file under `policies` section.
+See policies configuration reference for all the possible options here.
```
begelundmuller, line 15: Add link

docs/docs/develop/policies.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Dashboard Policies
+Currently, Rill supports the following use cases for dashboards:
+- Dashboard level access - Based on this condition you can restrict access to a dashboard for specific users based on user attributes.
+- Row level filtering - Based on this condition you can restrict access to certain rows for a specific users based on user attributes.
+- Column level access - Based on this condition you can restrict access to certain columns in a dashboard for a specific users based on user attributes.
```
begelundmuller, line 21: Instead of "Column level access", can we make it "Dimension/measure level access"

docs/docs/develop/policies.md
```diff
@@ -0,0 +1,118 @@
...
+- Dashboard level access - Based on this condition you can restrict access to a dashboard for specific users based on user attributes.
+- Row level filtering - Based on this condition you can restrict access to certain rows for a specific users based on user attributes.
+- Column level access - Based on this condition you can restrict access to certain columns in a dashboard for a specific users based on user attributes. 
+Currently only dimensions defined in the dashboard definition can be used for column level access.
+
+### User Attributes
+Following user attributes are available for defining policies:
+- `name` - name of the user. Example - `John Doe`    
+- `email` - email of the user. Example - `john.doe@test.com`
+- `domain` - domain of the email. Example - `test.com`
```
begelundmuller, line 28: Should use `example.com` to avoid clashes with real domains: https://en.wikipedia.org/wiki/Example.com

docs/docs/develop/policies.md
```diff
@@ -0,0 +1,118 @@
...
+    ```
+   > **_Note:_** If include is defined all other dimensions will be excluded. Only one of the
+     `include`/`exclude` can be defined at a time.
+
+> **_Note:_** `has_access` and `if` conditions on `include`/`exclude` fields needs to be valid expressions that evaluates to true/false.
+
+## Advanced Examples
+Policies are resolved in two phases - 
+1. The templating engine replaces placeholder like this `{{ }}` with actual values
+2. For `has_access` and `if` conditions, the resultant expression is evaluated to true/false, whereas for `filter` condition the resultant expression is injected in the query.
```
begelundmuller, line 77: Clarify that it is injected as a SQL expression in the WHERE clause or the underlying query

docs/docs/develop/policies.md
```diff
@@ -0,0 +1,118 @@
...
+
+> **_Note:_** We use sprig templating engine for templating and govaluate for expression evaluation. So any valid operator/expression they support will work here. You can refer to their documentation for more details.
+[Here](http://masterminds.github.io/sprig/) are the supported sprig functions and [here](https://github.com/Knetic/govaluate/blob/master/MANUAL.md#operators) are the supported expression evaluation operators.
```
begelundmuller, line 80: We actually use the Go [`text/template`](https://pkg.go.dev/text/template) templating engine with Sprig functions added to the context.

docs/docs/develop/policies.md
```diff
@@ -0,0 +1,118 @@
...
+
+> **_Note:_** We use sprig templating engine for templating and govaluate for expression evaluation. So any valid operator/expression they support will work here. You can refer to their documentation for more details.
+[Here](http://masterminds.github.io/sprig/) are the supported sprig functions and [here](https://github.com/Knetic/govaluate/blob/master/MANUAL.md#operators) are the supported expression evaluation operators. 
+
+Now lets look at some advanced examples.
+
+1. Let's say you want to restrict access to a dashboard only to admin users or users belonging to `test.com` domain.
+    ```yaml 
+    policy:
+      has_access: "'{{ .user.admin }}' == true || '{{ .user.domain }}' == 'test.com'"
```
begelundmuller, line 87: 1. Are double quotes around the expression needed here?
2. Why does `{{ .user.admin }}` need to have single quotes around it? Since it's a bool, I would have thought it just resolved to `true == true`, not `'true' == true`

docs/docs/develop/policies.md
```diff
@@ -0,0 +1,118 @@
...
+    ```
+2. Let's say additionally we want to filter queries based on user's groups and there exist a `group` dimension in the model.
+    ```yaml 
+    policy:
+      has_access: "'{{ .user.admin }}' == true || '{{ .user.domain }}' == 'test.com'"
+      filter: "groups IN ('{{ .user.groups | join "', '" }}')"
+    ```
```
begelundmuller, line 94: Let's remove this for now since we don't support creating user groups on admin-side yet

docs/docs/develop/policies.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Custom attributes
+Sometimes you may want to use custom attributes for defining policies other than existing user attributes. For example, you may have a `department` dimension and want to filter the dashboard view as per the users department.
```
begelundmuller, line 97: Can we make the example `tenant_id` instead? I think that would show a more powerful use case (partner filtering)

docs/docs/develop/policies.md
```diff
@@ -0,0 +1,118 @@
...
+You need to define the mapping of users to department as a separate [source](../reference/project-files/sources) in Rill. 
+Then you can refer to the source mapping in the policy. 
+
+For example, lets say we create a file named `mappings.csv` having the following data and put it in `data/` directory.
+```csv
+email,department
+john.doe@test.com,marketing
+jane.doe@test.com,tech
```
begelundmuller, line 105: Also use example.com

docs/docs/develop/policies.md
```diff
@@ -0,0 +1,118 @@
...
+type: "local_file"
+path: "data/mappings.csv"
+``` 
+
+Now we can refer to this source in the policy as follows:
+```yaml
+policy:
+has_access: "'{{ .user.admin }}' == true || '{{ .user.domain }}' == 'test.com'"
+filter: "department IN (SELECT department FROM mappings WHERE email = '{{ .user.email }}' )"
+```
```
begelundmuller, line 118: 1. Missing indentation under policy
2. Can we make `has_access` be just `true` here to focus just on the sub-query complexity?

begelundmuller (changes requested):
I think most of the content about templating and expressions here can stay, but I would still additionally propose a separate reference page (as described above) with more explicit documentation about how templating and expressions work (not specific to policies – templating is already available for models and sources also, and expressions will soon be available in other places too)

docs/docs/develop/policies.md
```diff
@@ -0,0 +1,131 @@
...
+
+```yaml
+policy:
+  has_access: true
+  exclude:
+    - name: ssn
+      if: "'{{ .user.domain }}' != 'example.com'"
+```
+
+Alternatively, you can explicitly define the dimensions and measures to include using the `include` key. It uses the same syntax as `exclude` and automatically exclude all names not explicitly defined in the list. See the [Dashboard YAML](../reference/project-files/dashboards) reference for details.
```
begelundmuller, line 92:
```suggestion
Alternatively, you can explicitly define the dimensions and measures to include using the `include` key. It uses the same syntax as `exclude` and automatically excludes all names not explicitly defined in the list. See the [Dashboard YAML](../reference/project-files/dashboards) reference for details.
```

docs/docs/develop/policies.md
```diff
@@ -0,0 +1,131 @@
...
+```yaml 
+policy:
+  has_access: "{{ .user.admin }} == true || '{{ .user.domain }}' == 'example.com'"
+```
+
+> **_Note:_** If the `policy` section is defined and `has_access` is not, then `has_access` will default to `false`, meaning that it won't be accessible to anyone.
+
+### Show only data from the user's own domain
+
+You can force the dashboard to filter all dimension and measure values by the user's domain. Assuming your dashboard's model has a `domain` column, add the following clause to your dashboard's YAML:
```
begelundmuller, line 70:
```suggestion
You can limit the data available to the dashboard by applying a filter on the underlying data. Assuming the dashboard's underlying model has a `domain` column, adding the following clause to the dashboard's YAML will only show dimension and measure values for the current user's email domain:
```

docs/docs/develop/policies.md
```diff
@@ -0,0 +1,131 @@
...
+```
+-->
+
+### Advanced Example: Custom attributes
+
+For some use cases, the built-in user attributes do not provide sufficient context to correctly restrict access. For example, a dashboard for a multi-tenant SaaS application might have a `tenant_id` column and external users should only be able to see data for the tenant they belong to.
+
+To support this, we can ingest a separate data [source](./import-data) containing mappings of user email addresses to tenant IDs and reference it in the row-level filter.
+
+For example, if we have a `mappings.csv` file in the `data` directory of our Rill project:
```
begelundmuller, line 112:
```suggestion
For example, if we have a `mappings.csv` file in the `data` directory of our Rill project with the following contents:
```

begelundmuller (approved):

PR #2968, opened 2023-08-24
dashboard policy validations

pjain1:


pjain1:
@ericpgreen2 In `GetCatalogEntry`, we actually resolve the template using user attributes and evaluate the resultant expression that's why you will see errors if it fail during either resolution or expression evaluation. Currently we do not do that in `PutAndReconcile` as I thought user context might not be available always during local development as there might not be any user logged in. How do we solve this ? should we make sure if user is writing a policy, they have to provide user attributes for testing during local development or should we assume some test user attributes to validate the expressions ?
cc @begelundmuller

runtime/services/catalog/artifacts/yaml/objects.go
```diff
@@ -425,6 +446,21 @@ func fromMetricsViewArtifact(metrics *MetricsView, path string) (*drivers.Catalo
 	}, nil
 }
 
+func validatedPolicyFieldList(fieldConditions []*ConditionalColumn, dimensions map[string]bool, property string) error {
+	if len(fieldConditions) > 0 {
+		for _, field := range fieldConditions {
+			if field.Name == "" || field.Condition == "" {
+				return fmt.Errorf("invalid 'policy': '%q' fields must have a valid name and condition", property)
```
begelundmuller, line 453: 1. In the YAML, the fields are "name" and "if" (not "condition")
2. It should put single quotes around `name` and `condition` since they're YAML fields (like for `'policy'`)

runtime/services/catalog/artifacts/yaml/objects.go
```diff
@@ -425,6 +446,21 @@ func fromMetricsViewArtifact(metrics *MetricsView, path string) (*drivers.Catalo
...
+			}
+			// check the name is a valid dimension that exists in the metrics view
+			if !dimensions[field.Name] {
+				return fmt.Errorf("invalid 'policy': '%q' property [%q] does not exists in dimensions list", property, field.Name)
```
begelundmuller, line 457: The quoting is weird here – `'%q'` will output as `'"xxx"'`, and `[%q]` is inconsistent with other formatting. Maybe just use `%q` everywhere and nothing else around it?

runtime/compilers/rillv1/parse_metrics_view.go
```diff
@@ -186,10 +186,14 @@ func (p *Parser) parseMetricsView(ctx context.Context, node *Node) error {
 	spec.DefaultTimeRange = tmp.DefaultTimeRange
 	spec.AvailableTimeZones = tmp.AvailableTimeZones
 
+	// create a map of dimensions to be used for policy validation
+	dimensions := map[string]bool{}
```
begelundmuller, line 190: There are already `names` and `columns` maps available in scope – can we find a way to just use one of those?

runtime/compilers/rillv1/parse_metrics_view.go
```diff
@@ -223,6 +227,12 @@ func (p *Parser) parseMetricsView(ctx context.Context, node *Node) error {
 		// validation has been done above, only one of these will be set
 		if tmp.Policy.Include != nil {
 			for _, include := range tmp.Policy.Include {
+				if include.Name == "" || include.Condition == "" {
+					return fmt.Errorf("invalid 'policy': 'include' fields must have a valid name and condition")
+				}
+				if !dimensions[include.Name] {
+					return fmt.Errorf("invalid 'policy': 'include' property [%q] does not exists in dimensions list", include.Name)
+				}
```
begelundmuller, line 235: Validation can't happen here – see note earlier in the function:
```
// NOTE: After calling upsertResource, an error must not be returned. Any validation should be done before calling it.
```

begelundmuller (changes requested):

begelundmuller:
@pjain1 @ericpgreen2 While we don't have any user context when reconciling/validating a policy, we should be able to use a dummy payload to at least validate that the expressions and templating parses correctly. We could use a payload like this:
```
email: mock@example.org
domain: example.org
name: Mock User
groups: []
admin: false
```

runtime/compilers/rillv1/parse_metrics_view.go
pjain1: ok missed that, will use

runtime/compilers/rillv1/parse_metrics_view.go
pjain1: yeah

runtime/services/catalog/artifacts/yaml/objects.go
pjain1: What about -
```
return fmt.Errorf("invalid 'policy': '%s' property [%s] does not exists in dimensions list", property, field.Name)
```

pjain1: so the key will be single quoted and value will be double quoted

pjain1:
@ericpgreen2 added validations with mock user data

runtime/compilers/rillv1/parse_metrics_view.go
```diff
@@ -171,6 +165,74 @@ func (p *Parser) parseMetricsView(ctx context.Context, node *Node) error {
...
+		if len(tmp.Policy.Include) > 0 && len(tmp.Policy.Exclude) > 0 {
+			return errors.New("invalid 'policy': only one of 'include' and 'exclude' can be specified")
+		}
+		if tmp.Policy.Include != nil {
+			for _, include := range tmp.Policy.Include {
+				if include.Name == "" || include.Condition == "" {
+					return fmt.Errorf("invalid 'policy': 'include' fields must have a valid 'name' and 'if' condition")
+				}
+				if !names[include.Name] {
+					return fmt.Errorf("invalid 'policy': 'include' property %q does not exists in dimensions list", include.Name)
```
begelundmuller, line 204: 1. `names` covers both dimension and measure names – so I guess the error should be "does not match a dimension or measure name"
2. Since the validation now checks both dimension and measure names, we should update the validation in the API handlers to check measure names as well (and also update the `GetCatalogEntry` call to filter out inaccessible measures)

ericpgreen2:
> @ericpgreen2 added validations with mock user data

So much better!

begelundmuller (approved):

PR #2975, opened 2023-08-24
Sort leaderboards by context columns

bcolloran:
Implements sorting by context columns *in the leaderboards* (adding dimension tables will be a separate PR to reduce the footprint).

@nishantmonu51 and @jkhwu, PR is ready for UX/QA pass. Please note that this PR switches to a new API for retrieving sorted data from the backend, and during development of this functionality I encountered some performance regressions. Those seem to have been fixed on the data sets that I normally use for QA, but it will be worth stress testing leaderboards using a range of datasets -- note that both the rows in the data set and the number of available leaderboards can affect performance.

@AdityaHegde , this continues to build on the last couple PRs of mine that you have reviewed, so I'd love for you to review this as well. I believe Nishant has been eager to get this in (I'll let him speak to the details) so if you can look at it soon that would be great.


bcolloran:
@nishantmonu51 and @jkhwu, I've tried to do a thorough QA on this, but I'm probably at the point of hitting diminishing returns-- I think it's looking pretty good, but I might be too close to it at this point and missing something that would be obvious to someone with fresh eyes and/or a different QA pattern. Also, as mentioned before, trying some other data would be great. I think we're in pretty good shape, but looking forward to your feedback in case I've missed anything! :-)

bcolloran:
(Also @nishantmonu51: if you are able to do QA and don't find any problems and there are no code issues after Aditya's review, feel free to merge during IST)

ericpgreen2 (approved):
Approving from a code perspective. Nice to see the new API make the `Leaderboard` component a little bit simpler.

nishantmonu51:
I tested this out, it seems to be working fine for leaderboards but not working when i expand the dimension tables..
In expanded tables i am unable to sort by the contextual column. 
@bcolloran : Is that expected ?

nishantmonu51:
Closed in favor of #3089

PR #2982, opened 2023-08-25
Syft PoC

naagas-syft:
## About
This PR is a PoC to integrate Syft into Rill local web.  Please see inline comments on the PR.

## Overview
- syft/events.ts includes all the event modeling.
- `npx syft generate ts` generates the library. which can be used in the app.
- fixed a bug where, we were sending the appScreen (which is an object) instead of sending an enum. 

@nishantmonu51 @ericpgreen2 fyi!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3669,6 +3669,20 @@
         "react": ">=16.8.0"
       }
     },
+    "node_modules/@es-joy/jsdoccomment": {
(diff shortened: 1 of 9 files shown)


nishantmonu51:
Thanks @naagas-syft for the PR, I am closing this for now as part of cleaning up old PRs. Will pickup changes from the branch when getting back to this.

PR #2986, opened 2023-08-28
APIs to mimic another dashboard user with granular access policies

rakeshsharma14317:
This PR introduces two new APIs that enable admins to test access policies in cloud:

- `SearchProjectUsers` – use to search all users who have access to the project
- `GetDeploymentCredentials` – use to get a JWT that mimics another user


admin/database/postgres/postgres.go
```diff
@@ -521,6 +521,21 @@ func (c *connection) FindUsersByEmailPattern(ctx context.Context, emailPattern,
...
+// SearchProjectUsers searches for users that have access to the project.
+func (c *connection) SearchProjectUsers(ctx context.Context, projectID, emailQuery, afterEmail string, limit int) ([]*database.User, error) {
+	var res []*database.User
+	err := c.getDB(ctx).SelectContext(ctx, &res, `
+		SELECT u.* FROM users u
+		JOIN users_projects_roles upr ON u.id = upr.user_id
+		JOIN project_roles r ON r.id = upr.project_role_id 
+		WHERE upr.project_id=$1 AND lower(u.email) ILIKE lower($2) AND lower(u.email) > lower($3)
+		ORDER BY lower(u.email) LIMIT $3`, projectID, emailQuery, afterEmail, limit)
+	if err != nil {
```
begelundmuller, line 533: It also needs to search org members with transitive access to the project. I think you need something like this:
```sql
        SELECT u.* FROM users u
        WHERE u.id IN (
            SELECT upr.user_id FROM users_projects_roles upr WHERE upr.project_id=$1
            UNION
            SELECT ugu.user_id FROM usergroups_projects_roles ugpr JOIN usergroups_users ugu ON ugpr.usergroup_id = ugu.usergroup_id WHERE ugpr.project_id=$1
        )
        AND lower(u.email) LIKE lower($2)
        AND lower(u.email) > lower($3)
        ORDER BY lower(u.email) ASC LIMIT $4
```

begelundmuller (changes requested):

admin/database/postgres/postgres.go
```diff
@@ -521,6 +521,25 @@ func (c *connection) FindUsersByEmailPattern(ctx context.Context, emailPattern,
...
+		SELECT u.* FROM users u
+        WHERE u.id IN (
+            SELECT upr.user_id FROM users_projects_roles upr WHERE upr.project_id=$1
+            UNION
+            SELECT ugu.user_id FROM usergroups_projects_roles ugpr JOIN usergroups_users ugu ON ugpr.usergroup_id = ugu.usergroup_id WHERE ugpr.project_id=$1
+        )
+        AND lower(u.email) LIKE lower($2)
+        AND lower(u.email) > lower($3)
+        ORDER BY lower(u.email) ASC LIMIT $4`, projectID, emailQuery, afterEmail, limit)
+	if err != nil {
```
begelundmuller, line 537: Indent with tabs instead of spaces for alignment

admin/server/deployment.go
```diff
@@ -150,3 +152,79 @@ func (s *Server) TriggerRedeploy(ctx context.Context, req *adminv1.TriggerRedepl
...
+	case *adminv1.GetDeploymentCredentialsRequest_UserId:
+		attr, err = s.jwtAttributesForUser(ctx, permissions, claims.OwnerID(), proj.OrganizationID)
+		if err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+	case *adminv1.GetDeploymentCredentialsRequest_Attrs:
+		attr = id.Attrs.AsMap()
+	default:
+		return nil, status.Error(codes.InvalidArgument, "id must be either user or attrs")
+	}
```
begelundmuller, line 201: 1. Call the variable `for` instead of `id` to match the `req.For` name
2. Should pass `for.UserId` instead of `claims.OwnerID()` to `jwtAttributesForUser`
3. In `default` case, just return `invalid 'for' type`

begelundmuller (changes requested):

begelundmuller (approved):

begelundmuller (approved):

PR #2988, opened 2023-08-28
Fix zoom bugs

djbarnwal:
## Checklist
- [x] Manual verification
- [ ] Unit test coverage
- [ ] E2E test coverage
- [x] Needs manual QA?

#### Issue addressed: 
1. Zoom was broken on dashboards with high number of charts when keybinding `Z` was used
2. The zoomed in dates were not accurate due to timezone changes

#### Details:
1. Moves `<svelte:window>` to a component single component so to avoid multiple dispatches/state changes for the same keybinding.
2. Handles timezone offset when zoomed in

## Steps to Verify
Play around with zoom on a dashboard with 5 or more charts.

### DIFF
--- a/web-common/src/features/dashboards/time-series/MetricsTimeSeriesCharts.svelte
+++ b/web-common/src/features/dashboards/time-series/MetricsTimeSeriesCharts.svelte
@@ -7,6 +7,7 @@
     useDashboardStore,
     useFetchTimeRange,
     useComparisonRange,
+    metricsExplorerStore,
   } from "@rilldata/web-common/features/dashboards/dashboard-stores";
   import {
     humanizeDataType,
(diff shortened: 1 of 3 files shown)


web-common/src/features/dashboards/time-series/MetricsTimeSeriesCharts.svelte
```diff
@@ -247,6 +249,34 @@
...
+
+  function onKeyDown(e) {
+    if (scrubStart && scrubEnd) {
+      // if key Z is pressed, zoom the scrub
+      if (e.key === "z") {
+        const { start, end } = getOrderedStartEnd(
+          $dashboardStore?.selectedScrubRange?.start,
+          $dashboardStore?.selectedScrubRange?.end
+        );
+        metricsExplorerStore.setSelectedTimeRange(metricViewName, {
```
AdityaHegde, line 261: Why not merge these actions instead of using setTimeout?

djbarnwal: I tried merging it into a single dashboard reducer call but the two values (`timeRange` and `scrubRange`) are serialized independently on the protos. Successive calls without any delay leads to race condition.

AdityaHegde (approved):

PR #2990, opened 2023-08-29
Adding code for watching files and resources

AdityaHegde:
This is an incomplete feature. Depends on #2946 before a complete integration.


web-local/src/routes/(application)/+layout.svelte
```diff
@@ -1,15 +1,30 @@
...
+  $: {
+    if (filesUnsubscribe) filesUnsubscribe();
+    filesUnsubscribe = filesRequests(queryClient);
+  }
+
+  let resourcesUnsubscribe: Unsubscriber;
+  $: {
+    if (resourcesUnsubscribe) resourcesUnsubscribe();
+    resourcesUnsubscribe = resourcesRequests(queryClient);
+  }
```
ericpgreen2, line 27: Can we push the unsubscribe logic into the functions so they're one liners? And maybe with better names, it'd be:
```typescript
watchAndHandleFiles(queryClient)
watchAndHandleResources(queryClient)
```

web-common/src/features/reconcile/WatchRequestClient.ts
ericpgreen2, line 1: Thoughts on moving the code in the `features/reconcile` directory into `features/entity-management`? “Entity management” is the meta feature for general CRUD. Fetching files & resources falls into this bucket, I believe.

web-common/src/features/reconcile/files-requests.ts
```diff
@@ -0,0 +1,86 @@
...
+    client = new WatchRequestClient<
+      RuntimeServiceWatchFilesParams,
+      V1WatchFilesResponse
+    >(
+      `${runtime.host}/v1/instances/${runtime.instanceId}/resources/-/watch`,
+      // TODO: filters
+      {}
+    );
+
+    handleFileResponses(client, queryClient, runtime.instanceId);
```
ericpgreen2, line 32: Instead of injecting the `client` here, I'd expect a `handleFileResponse` callback to be applied to each message in the stream along these lines:
```
for await (const res of stream) {
    handleFileResponse(res)
}
```

You could open the stream in this `fileRequests` function rather than opening it in the `handleFileResponses` function.

web-common/src/features/reconcile/WatchRequestClient.ts
```diff
@@ -0,0 +1,77 @@
...
+
+export class WatchRequestClient<
+  Req extends WatchRequest,
+  Res extends WatchResponse
+> {
+  private controller: AbortController;
+  private readonly url: string;
+  private shouldCancel = false;
+
+  public constructor(url: string, req: Req) {
```
ericpgreen2, line 34: Consider passing into the client a callback function that can be applied to each message in the stream.

Below, rather than a generator emitting messages, you'd handle them directly, like:
```typescript
for await (const res of stream) {
    handleStreamResponse(res)
}
```

web-common/src/features/reconcile/WatchRequestClient.ts
```diff
@@ -0,0 +1,77 @@
...
+    const urlObj = new URL(url);
+    for (const p in req) {
+      urlObj.searchParams.set(p, String(req[p]));
+    }
+    this.url = urlObj.toString();
+  }
+
+  public async *send(reconnected: () => void) {
```
ericpgreen2, line 42: `reconnected` sounds like a boolean, so `onReconnect` would be better

ericpgreen2:
Looks like some good functionality in here.

At a high-level, it feels like there's a lot of layers to the code, e.g.: `filesRequest` -> `handleFileResponses` -> `WatchClient` -> `streamingFetchWrapper` -> `fetch`. I think it'd make the code more approachable if we could remove/condense at least 1 of these layers.

AdityaHegde:
`handleFileResponses` and `WatchClient` should be merged i agree. The rest are needed IMO. `streamingFetchWrapper` is used in batching (another PR)

web-local/src/routes/(application)/+layout.svelte
AdityaHegde: We need to handle the unsubscribe in the component. I think it is fine to have this.

web-common/src/features/reconcile/WatchRequestClient.ts
AdityaHegde: I have moved the files and resource clients there. The base class can be used for logs as well, so moved to runtime-client.

ericpgreen2: Great

web-local/src/routes/(application)/+layout.svelte
```diff
@@ -1,15 +1,27 @@
...
   import { QueryClientProvider } from "@tanstack/svelte-query";
+  import { onMount } from "svelte";
   import { createQueryClient } from "../../lib/svelte-query/globalQueryClient";
 
   const queryClient = createQueryClient();
 
   beforeNavigate(retainFeaturesFlags);
+
+  onMount(() => {
+    const watchFilesUnsubscribe = watchFilesClient(queryClient);
```
ericpgreen2, line 18: Nice, this is better. Now how about naming like this:
```
const stopWatchingFiles = startWatchingFiles(queryClient)
...
```

web-common/src/runtime-client/watch-request-client.ts
ericpgreen2, line 1: Name suggestion: `WatchRequestClient` —> `Watcher`  or `WatchClient`

web-common/src/runtime-client/watch-request-client.ts
```diff
@@ -0,0 +1,82 @@
...
+  | V1WatchFilesResponse
+  | V1WatchResourcesResponse
+  | V1WatchLogsResponse;
+
+type StreamingFetchResponse<Res extends WatchResponse> = {
+  result?: Res;
+  error?: { code: number; message: string };
+};
+
+export class WatchRequestClient<Res extends WatchResponse> {
```
ericpgreen2, line 22: Do we need `Res extends WatchResponse` or can we just use `WatchResponse`?

web-common/src/runtime-client/watch-request-client.ts
```diff
@@ -0,0 +1,82 @@
...
+  private controller: AbortController;
+
+  public constructor(
+    private readonly getUrl: (runtime: Runtime) => string,
```
ericpgreen2, line 26: Instead of passing in a `getURL` function, it looks like you could just pass a `slug` (`/files/watch` or `/resources/watch`) and use that to compute the URL within the `WatchRequestClient`.

web-common/src/features/entity-management/watch-files-client.ts
```diff
@@ -0,0 +1,62 @@
...
+}
+
+function invalidateFile(queryClient: QueryClient, res: V1WatchFilesResponse) {
+  // Watch file returns events for all files under the project. Ignore everything except .sql, .yaml & .yml
+  if (
+    !res.path.endsWith(".sql") &&
+    !res.path.endsWith(".yaml") &&
+    !res.path.endsWith(".yml")
+  )
+    return;
```
ericpgreen2, line 27: I think we should remove this condition. This work will enable us to show the whole project directory in the sidebar, including `.md`, `.env`, and `.py` files.

Until we get there, I think the filtering for specific file types should be logic in the sidebar component, not in this client.

web-common/src/features/entity-management/watch-resources-client.ts
```diff
@@ -0,0 +1,92 @@
...
+export function watchResourcesClient(queryClient: QueryClient) {
+  return new WatchRequestClient<V1WatchResourcesResponse>(
+    (runtime) =>
+      `${runtime.host}/v1/instances/${runtime.instanceId}/resources/-/watch`,
+    (res) => invalidateResourceResponse(queryClient, res),
+    () => invalidateAllResources(queryClient)
+  ).start();
+}
+
+function invalidateResourceResponse(
```
ericpgreen2, line 21: This is more `handleResourceResponse`, the invalidation comes later in your utility functions below

web-common/src/features/entity-management/watch-files-client.ts
```diff
@@ -0,0 +1,62 @@
...
+export function watchFilesClient(queryClient: QueryClient) {
+  return new WatchRequestClient<V1WatchFilesResponse>(
+    (runtime) =>
+      `${runtime.host}/v1/instances/${runtime.instanceId}/files/watch`,
+    (res) => invalidateFile(queryClient, res),
+    () => invalidateAllFiles(queryClient)
+  ).start();
+}
+
+function invalidateFile(queryClient: QueryClient, res: V1WatchFilesResponse) {
```
ericpgreen2, line 20: `handleFileResponse`

web-common/src/features/entity-management/watch-resources-client.ts
```diff
@@ -0,0 +1,92 @@
...
+    case "RESOURCE_EVENT_DELETED":
+      invalidateRemovedResource(queryClient, instanceId, res.resource);
+      queryClient.refetchQueries(
+        getRuntimeServiceListResourcesQueryKey(instanceId)
+      );
+      break;
+  }
+}
+
+async function invalidateResource(
```
ericpgreen2, line 50: Consider removing the `invalidateResource` and `invalidateRemovedResource` wrappers, given they each just call one `queryClient.XQueries()` function.

web-common/src/runtime-client/watch-request-client.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  private async maintainConnection() {
+    let firstRun = true;
+    // Maintain the controller here to make sure we check `aborted` for the correct one.
+    // Checking for `this.controller` might lead to edge cases where it has a newer controller after `runtime` changed.
+    let controller = new AbortController();
+
+    const url = this.getUrl(get(runtime));
+
+    while (!controller.signal.aborted) {
```
ericpgreen2, line 53: This looks like it'll run until aborted, so won't it stop once line 57 gets hit `controller.abort()`?

web-common/src/runtime-client/fetch-streaming-wrapper.ts
```diff
@@ -0,0 +1,50 @@
...
+  method: string,
+  body?: Record<string, unknown>,
+  signal?: AbortSignal
+): AsyncGenerator<T> {
+  let response: Response;
+  try {
+    response = await fetch(url, {
+      method,
+      ...(body ? { body: JSON.stringify(body) } : {}),
+      headers: { "Content-Type": "application/json" },
```
ericpgreen2, line 16: We'll need to provide our own headers at some point (JWT tokens)

web-common/src/runtime-client/watch-request-client.ts
AdityaHegde: This is needed to get typescript styling for the caller.

web-common/src/runtime-client/watch-request-client.ts
AdityaHegde: Let this be as is. The way the url is built should be the responsibility of the caller. This way a single class can handle requests to runtime and admin name spaces.

web-common/src/features/entity-management/watch-files-client.ts
AdityaHegde: I disagree. Lets add to this filter once we integrate the new editor properly. Right now we get 3+ events of updated to `.db.wal` for every file write.

web-common/src/features/entity-management/watch-resources-client.ts
AdityaHegde: It is just a single call right now. But once we move all invalidations from existing implementation this will grow. Better to keep them in a function make the caller smaller and easy to understand.

web-common/src/runtime-client/watch-request-client.ts
AdityaHegde: Yes the idea is that for every subscribe callback call, only one instance of `maintainConnection` should be active.

web-common/src/runtime-client/watch-request-client.ts
ericpgreen2: But this is part of the `runtime-client` and won't be used to connect to the admin server, right?

web-common/src/features/entity-management/watch-files-client.ts
ericpgreen2: Sure, that works

ericpgreen2:
Every 30 seconds I get this in my browser console:
image

ericpgreen2:
Looks pretty good to me. Though maybe you want @begelundmuller to take a quick look too – I'll leave that call to you.

ericpgreen2 (approved):

web-common/src/runtime-client/watch-request-client.ts
AdityaHegde: Hmmm you are right, even the logs API is on the runtime. We can refactor to have just the final url path once we know of all end points.

web-common/src/features/entity-management/watch-resources-client.ts
```diff
@@ -0,0 +1,92 @@
...
+      queryClient.refetchQueries(
+        getRuntimeServiceListResourcesQueryKey(instanceId)
+      );
+    // eslint-disable-next-line no-fallthrough
+    case "RESOURCE_EVENT_UPDATED_SPEC":
+    case "RESOURCE_EVENT_UPDATED_STATE":
+      invalidateResource(queryClient, instanceId, res.resource);
+      break;
+
+    case "RESOURCE_EVENT_DELETED":
```
begelundmuller, line 41: Note that these event names changed slightly here: https://github.com/rilldata/rill/pull/3010

begelundmuller (approved):
LGTM

PR #2991, opened 2023-08-29
Fix default time range adding comparision

AdityaHegde:
## Checklist
- [x] Manual verification
- [x] Unit test coverage
- [ ] E2E test coverage
- [x] Needs manual QA?

#### Issue addressed: 
When a default time range is selected in metrics config, comparison should be enabled by default if available.

## Steps to Verify
1. Create a dashboard with timestamp.
2. Set `default_time_range` to `P1D`.
3. Go to dashboard.
4. Time range should be last 24 hrs and comparison should be enabled to last day.


djbarnwal (approved):
Looks good!

PR #3001, opened 2023-08-30
Arrow

k-anshul:


k-anshul:
Sample performance results:
Query : `SELECT * FROM bigquery-public-data.covid19_open_data.compatibility_view LIMIT 10000000`
Old Approach : 135 seconds
New Approach: 60 seconds

runtime/drivers/bigquery/arrow.go
```diff
@@ -0,0 +1,155 @@
...
+	return rec, nil
+}
+
+// some impl details are copied from array.simpleRecords
+type arrowRecordReader struct {
+	bqIter      bigquery.ArrowIterator
+	records     []arrow.Record
+	cur         arrow.Record
+	arrowSchema *arrow.Schema
+	refCount    int64
```
begelundmuller, line 52: Did you consider using an `atomic.Int64` instead? The docs seems to recommend that

runtime/drivers/bigquery/arrow.go
```diff
@@ -0,0 +1,155 @@
...
+		return nil, err
+	}
+	defer rdr.Release()
+	records := make([]arrow.Record, 0)
+	for rdr.Next() {
+		rec := rdr.Record()
+		rec.Retain()
+		records = append(records, rec)
+	}
+	return records, rdr.Err()
```
begelundmuller, line 154: Why is the extra layer of buffering in `rs.records` needed versus directly buffering `rdr` and proxying to it in `Next()`?

begelundmuller (changes requested):

runtime/drivers/bigquery/sql_store.go
```diff
@@ -57,157 +68,201 @@ func (c *Connection) Query(ctx context.Context, props map[string]any, sql string
...
+			if err := writer.WriteBuffered(rec); err != nil {
+				return nil, err
+			}
+		}
+	}
+	if rdr.Err() != nil {
+		return nil, fmt.Errorf("file write failed with error: %w", rdr.Err())
+	}
+	writer.Close()
+	fw.Close()
```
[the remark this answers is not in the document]
k-anshul: Yes. `writer.Close()` will return an error which we are ignoring and `fw.Close()` will not do anything. There are many return paths in this code so I used a defer as well to close there in error cases.

runtime/drivers/bigquery/arrow.go
k-anshul: No I just copied from the `array.simpleRecords` code. But yeah thanks `atomic.Int64` is a better choice.

runtime/drivers/bigquery/arrow.go
k-anshul: I tried proxying `rdr` to next but overall it feels lot more complicated than current implementation. There are multiple cases to consider. `rdr` will be nil in first next call, `rdr` can return next as false if is are no more data or if there is some genuine error(which needs to be handled separately as well). Overall it felt lot easier to just look at the size of the array.
Also as of now `ArrowRecordBatch` from bigquery will only return one arrow record but it's better to keep array for any change in underlying impls in future.

begelundmuller (approved):

PR #3006, opened 2023-08-31
Use DuckDB SQL for access policy expressions

begelundmuller:


runtime/pkg/duckdbsql/expression.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"errors"
+	"fmt"
+)
+
+// EvaluateBool uses DuckDB to evaluate the given expression as a boolean value.
+func EvaluateBool(expr string) (bool, error) {
+	// NOTE: Doing type check in SQL instead of in Go to protect against expressions that return huge values.
+	res, isBool, err := queryBoolPair(fmt.Sprintf("SELECT v::BOOLEAN AS b, TYPEOF(v) = 'BOOLEAN' AS is_bool FROM (SELECT (%s) AS v)", expr))
```
pjain1, line 11: `v::BOOLEAN` will try to cast v as boolean so if its not boolean, query will fail so whats the need of `is_bool` expression. We can just rely on err from query to check if its boolean or not ?

begelundmuller:
> also we should add validation for row_filter now in objects.go and parse_metrics_view.go using ducksql ?

`EvaluateBool` uses a global, read-only, in-memory DuckDB handle, so it wouldn't for `row_filter` (since it may reference mapping tables, which wouldn't exist).

We could use the project's OLAP DB for validating `row_filter`, but not sure that's easily available (and then we would also need to support Druid validation), so thinking we shouldn't tackle that now.

Did I understand that correctly?

runtime/pkg/duckdbsql/expression.go
pjain1: ok was not aware of this, so this looks good.

pjain1 (approved):

PR #3010, opened 2023-08-31
Watch catalog resources API

begelundmuller:


runtime/server/controller.go
```diff
@@ -18,17 +24,106 @@ func (s *Server) WatchLogs(req *runtimev1.WatchLogsRequest, srv runtimev1.Runtim
...
+	if req.Replay {
+		rs, err := ctrl.List(ss.Context(), req.Kind)
+		if err != nil {
+			return status.Error(codes.InvalidArgument, err.Error())
+		}
+
+		// TODO: Enforce security policy.
+
+		for _, r := range rs {
+			err := ss.Send(&runtimev1.WatchResourcesResponse{
```
AdityaHegde, line 77: We should add a lock around send.

runtime/controller.go
```diff
@@ -673,22 +726,23 @@ func (c *Controller) safeRename(from, to *runtimev1.ResourceName) error {
 		return c.catalog.rename(from, to)
 	}
 
-	// Collision, do a delete+create instead of a rename
+	// Collision, do a create+delete instead of a rename
```
AdityaHegde, line 729: Whats the idea behind doing it in this order?

AdityaHegde (approved):
Looks good other than a couple of small comments.

runtime/controller.go
```diff
@@ -290,6 +300,38 @@ func (c *Controller) List(ctx context.Context, kind string) ([]*runtimev1.Resour
 	return c.catalog.list(kind, false)
 }
 
+// SubscribeCallback is the callback type passed to Subscribe.
+type SubscribeCallback func(e runtimev1.ResourceEvent, n *runtimev1.ResourceName, r *runtimev1.Resource)
+
+// Subscribe registers a callback that will receive resource update events.
```
begelundmuller, line 306:
```suggestion
// Subscribe registers a callback that will receive resource update events.
// The callback is invoked in the same goroutine as the call to Subscribe (so multiple callbacks can't be invoked concurrently).
```

runtime/controller.go
begelundmuller:
```suggestion
	// Collision, do a create+delete instead of a rename
	// (since creation might fail if the name is taken, whereas the delete is almost certain to succeed)
```

begelundmuller: Added a comment to explain – creation may uncover name conflicts, whereas deletion is basically sure to succeed

runtime/server/controller.go
begelundmuller: All calls to `ss.Send` happen in the same goroutine, so no lock is needed (added a comment on `Subscribe` to clarify this behavior)

PR #3012, opened 2023-09-01
connector mgmt changes

### LINKED ISSUE #2904: Runtime: Misc. connector management changes
- [ ] Rename `olap_driver` and `repo_driver` to `xxx_connector`
- [ ] Should we remove the `embed_catalog` option?
- [ ] Don't call `openAndMigrate` under the global connection cache lock
- [ ] Pass the `activity.Client` to `drivers.Open` instead of as a config
- [ ] Pass a reference to `runtime` to the connection cache to avoid passing activity attributes on every call

k-anshul:
closes #2904


runtime/drivers/sqlite/migrations/0008.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE instances ADD COLUMN olap_connector TEXT NOT NULL DEFAULT '';
+UPDATE instances SET olap_connector = olap_driver;
+ALTER TABLE instances DROP COLUMN olap_driver; 
+
+ALTER TABLE instances ADD COLUMN repo_connector TEXT NOT NULL DEFAULT '';
+UPDATE instances SET repo_connector = repo_driver;
+ALTER TABLE instances DROP COLUMN repo_driver;
```
begelundmuller, line 7: Why not just `RENAME COLUMN`?

k-anshul: OOPS. I forgot `RENAME COLUMN` is also available.

runtime/caches.go
```diff
@@ -82,85 +86,123 @@ func (c *connectionCache) Close() error {
...
+	if dedupe {
+		c.lock.Lock()
+		defer c.lock.Unlock()
+		conn.ref++
+		// remove from lru-cache
+		c.lruCache.Remove(key)
+		// set in in-use cache
+		c.cache[key] = conn
+	}
+	return conn.Handle, c.releaseFn(key, conn), nil
```
begelundmuller, line 183: Is there an unlikely but possible race condition here where the connection is evicted and closed again before the ref is bumped for a deduped case?

PR #3014, opened 2023-09-01
athena-driver

egor-ryashin:
## Checklist
- [x] Manual verification
- [ ] Unit test coverage
- [ ] E2E test coverage
- [x] Needs manual QA?

#### Issue addressed: 

https://github.com/rilldata/rill/pull/3014

#### Details:

Adds Athena data source for data ingestion, see screenshots.

## Steps to Verify

image

image


runtime/drivers/athena/athena.go
```diff
@@ -0,0 +1,294 @@
...
+
+// AsSQLStore implements drivers.Connection.
+func (c *Connection) AsSQLStore() (drivers.SQLStore, bool) {
+	return nil, false
+}
+
+// DownloadFiles returns a file iterator over objects stored in gcs.
+// The credential json is read from config google_application_credentials.
+// Additionally in case `allow_host_credentials` is true it looks for "Application Default Credentials" as well
+func (c *Connection) DownloadFiles(ctx context.Context, source *drivers.BucketSource) (drivers.FileIterator, error) {
```
begelundmuller, line 193: Incorrect docstring

runtime/drivers/athena/athena.go
```diff
@@ -0,0 +1,294 @@
...
+		return nil, fmt.Errorf("failed to parse config: %w", err)
+	}
+
+	prefix := "parquet_output_" + uuid.New().String()
+	bucketName := strings.TrimPrefix(strings.TrimRight(conf.OutputLocation, "/"), "s3://")
+	unloadPath := bucketName + "/" + prefix
+	err = c.unload(ctx, conf, "s3://"+unloadPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to unload: %w", err)
+	}
```
begelundmuller, line 205: 1. Why can't we use the output location outright? The Athena code samples don't appear to do any rewriting of the output path: https://docs.aws.amazon.com/athena/latest/ug/code-samples.html
2. If we need to rewrite the output location, use `url.Parse` and associated functions to safely edit the URL

runtime/drivers/athena/athena.go
```diff
@@ -0,0 +1,294 @@
...
+
+	bucketObj, err := c.openBucket(ctx, conf, bucketName)
+	if err != nil {
+		return nil, fmt.Errorf("cannot open bucket %q: %w", unloadPath, err)
+	}
+
+	opts := rillblob.Options{
+		ExtractPolicy: &runtimev1.Source_ExtractPolicy{
+			// FilesStrategy: runtimev1.Source_ExtractPolicy_STRATEGY_HEAD,
+		},
```
begelundmuller, line 215: Return an error if there's an extract policy. Not sure it makes sense to support for Athena (folks can just put a `LIMIT` clause instead)

runtime/drivers/athena/athena.go
```diff
@@ -0,0 +1,294 @@
...
+		var failureErr awserr.RequestFailure
+		if !errors.As(err, &failureErr) {
+			return nil, fmt.Errorf("failed to create the iterator %q %w", unloadPath, err)
+		}
+
+		// check again
+		if errors.As(err, &failureErr) && (failureErr.StatusCode() == http.StatusForbidden || failureErr.StatusCode() == http.StatusBadRequest) {
+			return nil, drivers.NewPermissionDeniedError(fmt.Sprintf("can't access remote err: %v", failureErr))
+		}
+	}
```
begelundmuller, line 231: Are these conditions needed here? IIRC, they are for handling public bucket access, which is not needed for Athena reads. I would suggest checking with @k-anshul.

runtime/drivers/athena/athena.go
```diff
@@ -0,0 +1,294 @@
...
+
+func (c *Connection) unload(ctx context.Context, conf *sourceProperties, path string) error {
+	finalSQL := fmt.Sprintf("UNLOAD (%s) TO '%s' WITH (format = 'PARQUET')", conf.SQL, path)
+
+	cfg, err := awsconfig.LoadDefaultConfig(context.TODO(), awsconfig.WithSharedConfigProfile(conf.ProfileName))
+	if err != nil {
+		return err
+	}
+
+	client := athena.NewFromConfig(cfg)
```
begelundmuller, line 254: Since this uses the AWS v2 SDK, I don't think it can re-use the same credentials config as the `s3` connector, but please pull it out to a separate function and make sure it supports the same behavior as the one for `s3` (`getAwsSessionConfig`)

runtime/drivers/athena/athena.go
```diff
@@ -0,0 +1,294 @@
...
+
+	resultConfig := &types.ResultConfiguration{
+		OutputLocation: aws.String("s3://" + strings.TrimPrefix(strings.TrimRight(conf.OutputLocation, "/"), "s3://") + "/output/"),
+	}
```
begelundmuller, line 258: Why is it necessary to specify the output location both here and in `UNLOAD`? Also, see note above about using URL parsing libraries if needing to manipulate URLs

runtime/drivers/athena/athena.go
```diff
@@ -0,0 +1,294 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Get Query execution and check for the Query state constantly every 2 second
+	executionID := *athenaExecution.QueryExecutionId
+
+	r := retrier.New(retrier.LimitedExponentialBackoff(20, 100*time.Millisecond, 1*time.Second), nil) // 100 200 400 800 1000 1000 1000 1000 1000 1000 ... < 20 sec
+
+	return r.Run(func() error {
```
begelundmuller, line 276: Given the Athena latencies, I don't think a mixed exponential/linear strategy is needed. Can just keep it simple and do a loop with a sleep (will also make it easier to reason about how cancellation gets enforced). Or maybe a loop with a `select` that checks both the timer and `ctx.Done()` (to support faster cancellation)

runtime/drivers/athena/athena.go
```diff
@@ -0,0 +1,294 @@
...
+		status, stateErr := client.GetQueryExecution(ctx, &athena.GetQueryExecutionInput{
+			QueryExecutionId: &executionID,
+		})
+
+		if stateErr != nil {
+			return stateErr
+		}
```
begelundmuller, line 283: Can just use the normal var name `err` – no apparent conflict in scope

runtime/drivers/athena/athena.go
```diff
@@ -0,0 +1,294 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	s3client := s3v2.NewFromConfig(cfg)
+	return s3blob.OpenBucketV2(ctx, s3client, bucket, nil)
+}
+
+func (c *Connection) unload(ctx context.Context, conf *sourceProperties, path string) error {
+	finalSQL := fmt.Sprintf("UNLOAD (%s) TO '%s' WITH (format = 'PARQUET')", conf.SQL, path)
```
begelundmuller, line 247: When using `UNLOAD`, I'm guessing Athena won't garbage collect the files? So we should ensure they are removed in both success and error cases.

runtime/drivers/athena/athena.go
```diff
@@ -0,0 +1,294 @@
...
+			Placeholder: "select * from catalog.table;",
+		},
+		{
+			Key:         "output.location",
+			DisplayName: "Output location",
+			Description: "Oputut location for query results in S3.",
+			Placeholder: "bucket-name",
+			Type:        drivers.StringPropertyType,
+			Required:    true,
+		},
```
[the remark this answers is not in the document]
egor-ryashin: 2. Nope, just a backet name.

runtime/drivers/athena/athena.go
```diff
@@ -0,0 +1,294 @@
...
+		{
+			Key:         "profile.name",
+			DisplayName: "AWS profile",
+			Description: "AWS profile for credentials.",
+			Type:        drivers.StringPropertyType,
+			Required:    true,
+		},
```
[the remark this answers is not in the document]
egor-ryashin: > Also, should this be in ConfigProperties and not SourceProperties?

There can be sources with different profiles (meanwhile profiles can be for the same AWS account).

runtime/drivers/athena/athena.go
egor-ryashin: 1.  Athena Go API and "gocloud.dev/blob/s3blob" demand different S3 location format.
s3blob needs only the bucket name:
```
	s3bucket, err := s3blob.OpenBucketV2(context.TODO(), s3client, "athena-output-2820", nil)
```
and throws error `InvalidBucketName: The specified bucket is not valid.` when `s3://athena-output-2820` passed.
While Athena SDK complains `InvalidRequestException: Invalid location athena-output-20287` if `s3` prefix is not specified:
```
	executeParams := &athena.StartQueryExecutionInput{
		QueryString:         aws.String("UNLOAD (SELECT * FROM egordb.ptable limit 10) TO '%s' WITH (format = 'PARQUET')", "athena-output-20287"),
		ResultConfiguration: resultConfig,
	}
```
And we need to decide what the user should pass: "s3://bucket-name" or "bucket-name", or the user has the luxury to specify both and expect the application to figure it out.
Right now the implementation allows all approaches but it requires additional parameter transformations.

runtime/drivers/athena/athena.go
egor-ryashin: It's a required parameter for any query.
```
	// The location in Amazon S3 where your query and calculation results are stored,
	// such as s3://path/to/query/bucket/ . To run the query, you must specify the
	// query results location using one of the ways: either for individual queries
	// using either this setting (client-side), or in the workgroup, using
	// WorkGroupConfiguration . If none of them is set, Athena issues an error that no
	// output location is provided.
```

runtime/drivers/athena/athena.go
```diff
@@ -0,0 +1,314 @@
...
+
+// MigrationStatus implements drivers.Connection.
+func (c *Connection) MigrationStatus(ctx context.Context) (current, desired int, err error) {
+	return 0, 0, nil
+}
+
+// AsObjectStore implements drivers.Connection.
+func (c *Connection) AsObjectStore() (drivers.ObjectStore, bool) {
+	return c, true
+}
```
begelundmuller, line 184: It would be preferable to implement this as a `SQLStore` instead of an `ObjectStore`, and having the source be a `DatabaseSource` instead of a `BucketSource`. The DuckDB `sqlstore_to_duckdb.go` transporter is equally capable of ingesting downloaded files, so I don't think it would be a big change. And then the `fromAthena` change can be reverted in the `objectstore_to_duckdb.go` transporter.

See note about implementing as a

runtime/drivers/athena/athena.go
```diff
@@ -0,0 +1,314 @@
...
+	conf, err := parseSourceProperties(source.Properties)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse config: %w", err)
+	}
+
+	cfg, err := awsconfig.LoadDefaultConfig(
+		ctx,
+		awsconfig.WithRegion(conf.Region),
+		awsconfig.WithCredentialsProvider(credentials.NewStaticCredentialsProvider(c.config.AccessKeyID, c.config.SecretAccessKey, c.config.SessionToken)),
+	)
```
begelundmuller, line 234: Does this work for environment credentials (in `~/.aws`)? See the S3 connector – the expected behavior is: use access key if provided, else fallback to environment credentials unless `AllowHostAccess` is false.

runtime/drivers/athena/athena.go
```diff
@@ -0,0 +1,314 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	prefix := "parquet_output_" + uuid.New().String()
+	bucketName := strings.TrimPrefix(strings.TrimRight(conf.OutputLocation, "/"), "s3://")
+	unloadPath := bucketName + "/" + prefix
+	err = c.unload(ctx, cfg, conf, "s3://"+unloadPath)
```
begelundmuller, line 242: See earlier note – it should take `OutputLocation` in the `s3://bucket/path` format, and then we can use `url.Parse` to parse it and obtain the bucket name for `openBucket`.

We should also make sure it supports nested output locations, as far as I can tell, it can't be assumed the bucket is dedicated to only Athena?

runtime/drivers/athena/athena.go
```diff
@@ -0,0 +1,314 @@
...
+		awsconfig.WithRegion(conf.Region),
+		awsconfig.WithCredentialsProvider(credentials.NewStaticCredentialsProvider(c.config.AccessKeyID, c.config.SecretAccessKey, c.config.SessionToken)),
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	s3client := s3v2.NewFromConfig(cfg)
+	return s3blob.OpenBucketV2(ctx, s3client, bucket, nil)
+}
```
begelundmuller, line 276: This duplicates the `awsconfig` from `DownloadFiles` and may have the same credentials issues

runtime/drivers/athena/athena.go
```diff
@@ -0,0 +1,314 @@
...
+
+		state := status.QueryExecution.Status.State
+
+		if state == types.QueryExecutionStateSucceeded || state == types.QueryExecutionStateCancelled {
+			return nil
+		} else if state == types.QueryExecutionStateFailed {
+			return fmt.Errorf("Athena query execution failed %s", *status.QueryExecution.Status.AthenaError.ErrorMessage)
+		}
+		return fmt.Errorf("Execution is not completed yet, current state: %s", state)
+	})
```
begelundmuller, line 313: 1. This seems to translate to a 20 second query execution timeout? It's too little, it should probably continue to check until the `ctx` is cancelled. (It would also return a weird retry error message when hitting the 20s timeout.)
2. It seems it will continue to retry even if query execution failed?
3. I don't see why this warrants a third party library dependency instead of a for loop and simple `select` that checks a timer and context cancellation

runtime/drivers/athena/athena.go
```diff
@@ -0,0 +1,314 @@
...
+	}
+	_, err = s3client.DeleteObjects(ctx, &s3v2.DeleteObjectsInput{
+		Delete: &s3v2types.Delete{
+			Objects: ids,
+		},
+	})
+	return err
+}
+
+func (c *Connection) DownloadFiles(ctx context.Context, source *drivers.BucketSource) (drivers.FileIterator, error) {
```
begelundmuller, line 224: It seems it doesn't call `cleanPath` on success cases? It's probably easiest to wrap the file iterator with an object that calls `cleanPath` on close

begelundmuller:
Another comment – please also add docs changes for the Athena connector. And @nishantmonu51 requested to let people know that having an S3 file retention rule for the `output_location` is a good idea, in case data is orphaned there (e.g. if Rill crashes).

runtime/drivers/athena/athena.go
egor-ryashin: 3. We've already imported that lib and it provides an opinionated approach that one doesn't need to reinvent and it's presents itself right away `retrier`, `ConstantBackoff` - the intention is more recognisable from those words. If I see a low level Timer then I need read more scrupulously to figure out if it's just a simple retry pattern or something else and more complex.

begelundmuller (changes requested):
Two other questions:

1. It's missing the ability to configure the AWS region now?
2. What region will be used if the workgroup is nil? Is it guaranteed to be the output_location one?

runtime/drivers/athena/athena.go
```diff
@@ -0,0 +1,332 @@
...
+		return conf.OutputLocation, nil
+	} else if conf.WorkGroup != "" {
+		wo, err := client.GetWorkGroup(ctx, &athena.GetWorkGroupInput{
+			WorkGroup: aws.String(conf.WorkGroup),
+		})
+		if err != nil {
+			return "", err
+		}
+		return *wo.WorkGroup.Configuration.ResultConfiguration.OutputLocation, nil
+	}
```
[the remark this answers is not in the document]
egor-ryashin: It's not guaranteed, but specifying workgroup for the purpose of not specifying output location and having this workgroup without a location asks for the fail-fast approach.

runtime/drivers/athena/athena.go
```diff
@@ -0,0 +1,332 @@
...
+
+		state := status.QueryExecution.Status.State
+
+		if state == types.QueryExecutionStateSucceeded || state == types.QueryExecutionStateCancelled {
+			return nil
+		} else if state == types.QueryExecutionStateFailed {
+			return fmt.Errorf("Athena query execution failed %s", *status.QueryExecution.Status.AthenaError.ErrorMessage)
+		}
+		return fmt.Errorf("Athena ingestion timeout")
+	})
```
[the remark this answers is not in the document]
egor-ryashin: Fixed.

runtime/drivers/duckdb/transporter/utils.go
```diff
@@ -98,12 +98,12 @@ func parseFileSourceProperties(props map[string]any) (*fileSourceProperties, err
...
 }
 
-func sourceReader(paths []string, format string, ingestionProps map[string]any) (string, error) {
+func sourceReader(paths []string, format string, ingestionProps map[string]any, fromAthena bool) (string, error) {
 	// Generate a "read" statement
 	if containsAny(format, []string{".csv", ".tsv", ".txt"}) {
 		// CSV reader
 		return generateReadCsvStatement(paths, ingestionProps)
-	} else if strings.Contains(format, ".parquet") {
+	} else if strings.Contains(format, ".parquet") || fromAthena {
```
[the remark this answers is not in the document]
egor-ryashin: Athena outputs parquet files without extension, and there's no configuration to change that. So the source reader doesn't have anything to detect `parquet` format here.

egor-ryashin:
> It's missing the ability to configure the AWS region now?

Yes, the region should be returned back. Right now, the region is resolved from the default aws configuration profile.

runtime/drivers/athena/athena.go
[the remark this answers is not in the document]
egor-ryashin: > What's the recommended polling interval?

There's no recommendation.

runtime/drivers/duckdb/transporter/utils.go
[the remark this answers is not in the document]
esevastyanov: Partly changed this so that there is no `fromAthena` arg but `.parquet` extension is added to `format`

runtime/drivers/athena/athena.go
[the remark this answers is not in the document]
esevastyanov: It is 1 sec in [code samples](https://docs.aws.amazon.com/athena/latest/ug/code-samples.html)

esevastyanov:
> Is it guaranteed to be the output_location one?

`output_location` is optional, `workgroup` is optional.
Default value for a workgroup is `primary` (this is a default Athena workgroup that cannot be deleted).
A workgroup may have no output location set.
A workgroup may also have a feature that overrides a client-side output location by its own so the logic ignores `workgroup` property if `output_location` is specified.

runtime/drivers/athena/sql_store.go
```diff
@@ -0,0 +1,270 @@
...
+		loadOptions = append(loadOptions, config.WithCredentialsProvider(p))
+	} else if !c.config.AllowHostAccess {
+		return aws.Config{}, fmt.Errorf("static creds are not provided, and host access is not allowed")
+	}
+
+	return config.LoadDefaultConfig(ctx, loadOptions...)
+}
+
+func (c *Connection) unload(ctx context.Context, client *athena.Client, conf *sourceProperties, unloadLocation string) error {
+	finalSQL := fmt.Sprintf("UNLOAD (%s) TO '%s' WITH (format = 'PARQUET')", conf.SQL, unloadLocation)
```
begelundmuller, line 111: We should put a newline after the injected SQL (`UNLOAD (%s\n) ...`) in case there's a comment at the end of it

runtime/drivers/athena/sql_store.go
```diff
@@ -0,0 +1,270 @@
...
+
+			switch status.QueryExecution.Status.State {
+			case types2.QueryExecutionStateSucceeded, types2.QueryExecutionStateCancelled:
+				return nil
+			case types2.QueryExecutionStateFailed:
+				return fmt.Errorf("Athena query execution failed %s", *status.QueryExecution.Status.AthenaError.ErrorMessage)
+			}
+		}
+		time.Sleep(time.Second)
+	}
```
begelundmuller, line 160: 1. Instead of using a hard-coded timer, it can use the `ctx` (check `<-ctx.Done()`), which has a timeout that's configurable through the `timeout:` YAML property.
2. If there's a `ctx` cancellation/timeout, it would be nice to cancel the running query.
3. For `case ..., types2.QueryExecutionStateCancelled` – shouldn't this case return an error? Otherwise, it will try to consume the results of the cancelled query.

runtime/drivers/athena/sql_store.go
```diff
@@ -0,0 +1,270 @@
...
+		}
+
+		if len(ids) > 0 {
+			if err := deleteBatch(ids); err != nil {
+				return err
+			}
+		}
+
+		if out.IsTruncated {
+			continuationToken = out.NextContinuationToken
```
begelundmuller, line 242: Maybe also check `out.NextContinuationToken != nil` just to be extra cautious that we don't loop forever?

runtime/drivers/duckdb/transporter/sqlstore_to_duckDB.go
```diff
@@ -76,13 +78,20 @@ func (s *sqlStoreToDuckDB) Transfer(ctx context.Context, srcProps, sinkProps map
...
 		if err != nil {
 			return err
 		}
 
 		format := fileutil.FullExt(files[0])
+		if fromAthena {
+			// Athena doesn't specify ".parquet" extension in output file names
+			// Append ".parquet" extension to the extension generated by Athena
+			format += ".parquet"
+		}
```
begelundmuller, line 93: This reflection really hurts... Is there not a way to have Athena give output files the right file extension with `UNLOAD`?

If not, we should add a way to propagate the format, maybe by adding it in `rillblob.Options` and exposing it as `iter.Format()` or something like that.

begelundmuller (changes requested):

runtime/drivers/athena/sql_store.go
```diff
@@ -0,0 +1,270 @@
...
+	}
+
+	opts := rillblob.Options{
+		GlobPattern: unloadPath + "/**",
+	}
+
+	it, err := rillblob.NewIterator(ctx, bucketObj, opts, c.logger)
+	if err != nil {
+		return nil, errors.Join(fmt.Errorf("cannot download parquet output %q %w", opts.GlobPattern, err), cleanupFn())
+	}
```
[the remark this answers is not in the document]
esevastyanov: Reduced the number of calls by moving the call into `defer`

runtime/drivers/athena/sql_store.go
```diff
@@ -0,0 +1,270 @@
...
+
+	outputURL, err := url.Parse(outputLocation)
+	if err != nil {
+		return nil, err
+	}
+
+	// outputLocation s3://bucket/path
+	// unloadLocation s3://bucket/path/rill_tmp_<uuid>
+	// unloadPath path/rill_tmp_<uuid>
+	unloadFolderName := "rill_tmp_" + uuid.New().String()
```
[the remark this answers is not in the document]
esevastyanov: Used `rill-tmp- `

begelundmuller (approved):
Looks solid!

PR #3018, opened 2023-09-04
Runtime: duckdb sqlite extensions based connector for sqlite

### LINKED ISSUE #2993: Runtime: Postgres connector

k-anshul:
- Adds a sqlite connector


begelundmuller (commented):
Hey, I took a look at this PR and added some comments in Notion [here](https://www.notion.so/rilldata/Extend-our-existing-DuckDB-connector-to-support-Postgres-and-SQLite-343844b9024a4051bdc6f971ccdd4d59?pvs=4#261768be31ff4afeb5969df34093daf5)

runtime/drivers/duckdb/transporter/sqlextension_to_duckDB.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	extensionName := t.from.Driver()
+	if err := t.to.Exec(ctx, &drivers.Statement{Query: fmt.Sprintf("INSTALL '%s'; LOAD '%s';", extensionName, extensionName)}); err != nil {
+		return fmt.Errorf("failed to load %s extension %w", extensionName, err)
+	}
+
+	userQuery := strings.TrimSpace(src.SQL)
+	userQuery, _ = strings.CutSuffix(userQuery, ";") // trim trailing semi colon
+	query := fmt.Sprintf("CREATE OR REPLACE TABLE %s AS (%s);", safeName(fSink.Table), userQuery)
```
begelundmuller, line 47: Extensions are loaded at the connection-level, not DB-level, so the ingestion query is not guaranteed to have the extension loaded. Probably need to use `WithConnection`

begelundmuller (changes requested):
Can you also merge main and update for changed driver interfaces?

runtime/drivers/sqlite/sqlite.go
```diff
@@ -55,7 +56,21 @@ func (d driver) Drop(config map[string]any, logger *zap.Logger) error {
...
+		DisplayName: "SQLite",
+		Description: "Import data from SQLite table to DuckDB.",
+		SourceProperties: []drivers.PropertySchema{
+			{
+				Key:         "sql",
+				Type:        drivers.StringPropertyType,
+				Required:    true,
+				DisplayName: "SQL",
+				Description: "Query to extract data from SQLite",
+				Placeholder: "SELECT * FROM sqlite_scan('sqlite.db', 'film');",
```
begelundmuller, line 69: I just realised this is still problematic – it has the problem described in bullet 2 [here](https://www.notion.so/rilldata/Extend-our-existing-DuckDB-connector-to-support-Postgres-and-SQLite-343844b9024a4051bdc6f971ccdd4d59?pvs=4#261768be31ff4afeb5969df34093daf5).

If it has `connector: sqlite`, you would expect `sql: <sqlite query>`, and should at least have a path to generic support for other OLAPs.

If this is a DuckDB-only thing, then it should probably be part of the `duckdb` connector (could just check if the source contains `sqlite_scan` and opportunistically load the extension if it has.

begelundmuller (changes requested):

runtime/drivers/sqlite/sqlite.go
k-anshul: Good point.

> If this is a DuckDB-only thing, then it should probably be part of the duckdb connector (could just check if the source contains sqlite_scan and opportunistically load the extension if it has.

I think doing this also just fulfils half the criteria listed in the point. We would have a `sqlite` connector but one that only works with `duckdb`. For a user it will also be inconsistent to use connector type as `postgres` for `postgres` connector and `duckdb` for `sqlite` connector.
I would suggest we take approach similar to postgres then for sqlite as well.

begelundmuller: Sorry, what I meant by "be part of the duckdb connector" is that it would be in a source with `connector: duckdb`. So it would fulfill the criteria (I think), but not be a standalone connector.

runtime/services/catalog/migrator/sources/sources.go
```diff
@@ -385,6 +385,8 @@ func connectorVariables(src *runtimev1.Source, env map[string]string, repoRoot s
 	case "motherduck":
 		vars["token"] = env["token"]
 		vars["dsn"] = ""
+	case "sqlite":
+		vars["dsn"] = ""
```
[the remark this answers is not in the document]
k-anshul: Yes not needed now.

begelundmuller (approved):

PR #3022, opened 2023-09-05
Fix toplist comparison deltas

begelundmuller:
As pointed about by @bcolloran, it's also worthwhile revisiting the base/comparison terminology used here since the comparison value is actually used as the base for the comparison. Let's revisit that in a separate PR when we also rename `MetricsViewComparisonToplist` to `MetricsViewToplist` (or maybe it's time we just make it `MetricsToplist`?).

See context in [Slack](https://rilldata.slack.com/archives/C02T907FEUB/p1692996522697639?thread_ts=1692829203.034269&cid=C02T907FEUB).


bcolloran (approved):
I can't really comment on the code other than to say it looks in line with what's already there, so I assume it's fine. :-)

But I can confirm that the API is returning the expected results, so I think we're good to go!

Thanks @begelundmuller! I'll go ahead and merge.

PR #3030, opened 2023-09-06
Fix JWT attributes bug

begelundmuller:
Introduced in https://github.com/rilldata/rill/pull/2986

### DIFF
--- a/admin/server/server.go
+++ b/admin/server/server.go
@@ -364,7 +364,7 @@ func (s *Server) checkRateLimit(ctx context.Context) (context.Context, error) {
 	return ctx, nil
 }
 
-func (s *Server) jwtAttributesForUser(ctx context.Context, permissions *adminv1.ProjectPermissions, orgID, userID string) (map[string]any, error) {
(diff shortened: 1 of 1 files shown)


admin/server/server.go
```diff
@@ -364,7 +364,7 @@ func (s *Server) checkRateLimit(ctx context.Context) (context.Context, error) {
 	return ctx, nil
 }
 
-func (s *Server) jwtAttributesForUser(ctx context.Context, permissions *adminv1.ProjectPermissions, orgID, userID string) (map[string]any, error) {
+func (s *Server) jwtAttributesForUser(ctx context.Context, permissions *adminv1.ProjectPermissions, userID, orgID string) (map[string]any, error) {
```
k-anshul, line 367: We seem to be using `orgID`, `userID` in other similar functions so wondering if we should instead just change the callers here ?

begelundmuller: Hmm, good point. A search shows multiple cases of both in the codebase. I think there's a pattern to have args related to the function name (in this case `...ForUser`) come before other args, so that fits here. Might be worth doing a general check of the code base for it, but keeping like this in the meantime.

PR #3032, opened 2023-09-06
Fixing scrub zoom and leaderboard comparison

AdityaHegde:
## Checklist
- [x] Manual verification
- [x] Unit test coverage
- [x] E2E test coverage
- [ ] Needs manual QA?

#### Issue addressed: 
After the latest PR for time range refactors a couple things were broken,
1. The race condition during scrub to zoom reappeared.
2. Switching to time range without valid comparison does not switch the comparison for leaderboard.

#### Details:
1. Fixing the race condition by making sure a single dashboard state update happens for a single action. Also updates the url only once.
2. Moving the default selection to dashboard store action.

## Steps to Verify
Issue 1
1. Create a dashboard with 5+ measures.
2. Use scrub to zoom a few times.
3. Scrub should not appear after zooming.

Issue 2
1. Create a dashboard.
2. Switch to a time range with comparison. Leaderboard context should select to comparison.
3. Switch to a time range without comparison. Leaderboard context should deselect from comparison.

### DIFF
--- a/web-common/src/features/dashboards/time-controls/time-control-store.spec.ts
+++ b/web-common/src/features/dashboards/time-controls/time-control-store.spec.ts
@@ -283,8 +283,7 @@ describe("time-control-store", () => {
     unmount();
   });
 
-  // TODO: fix this by using zone independent tests
-  it.skip("Scrubbing to zoom", async () => {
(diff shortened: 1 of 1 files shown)


djbarnwal (approved):
Looks good!

PR #3035, opened 2023-09-07
Options to improve ingest performance

begelundmuller:
- [x] Add `duckdb-vip` driver that doesn't enforce any limits
- [x] Increase reconcile timeout to 59 minutes (not 60, to avoid force termination at ingress)
- [x] Increase concurrent file downloads to 32 (from 8)
- [ ] Consider: Configure insert batches by GBs downloaded instead of number of files downloaded


k-anshul (approved):

k-anshul:
@begelundmuller Just realised we need to update this config as well : 

```
		Connectors: []*runtimev1.Connector{
			{
				Name:   "olap",
				Type:   olapDriver,
				Config: map[string]string{"dsn": olapDSN},
			},
			{
				Name:   "repo",
				Type:   repoDriver,
				Config: map[string]string{"dsn": repoDSN},
			},
		},
```

PR #3036, opened 2023-09-07
Runtime: Azure storage connector

rakeshsharma14317:


runtime/drivers/azure/azure.go
```diff
@@ -0,0 +1,257 @@
...
+			DisplayName: "Azure Storage Account",
+			Description: "Azure Storage Account inferred from your local environment.",
+			Type:        drivers.InformationalPropertyType,
+			Hint:        "Set your local credentials: <code>az login</code> Click to learn more.",
+			Href:        "https://docs.rilldata.com/develop/import-data#configure-credentials-for-azure",
+		},
+	},
+	ConfigProperties: []drivers.PropertySchema{
+		{
+			Key:  "azure.storage.account",
```
k-anshul, line 52: This should be `azure_storage_account`. 
Also need the other two keys: `azure_storage_sas_token` and `azure_storage_key`

runtime/drivers/azure/azure.go
```diff
@@ -0,0 +1,257 @@
...
+			Hint: "Enter path of file to load from.",
```
k-anshul, line 53: Hint and ValidateFunc not required here. Check `s3` ConfigProperties.

runtime/drivers/azure/azure.go
```diff
@@ -0,0 +1,257 @@
...
+				return err
+			},
+		},
+	},
+}
+
+type driver struct{}
+
+type configProperties struct {
+	Account string `mapstructure:"azure.storage.account"`
```
k-anshul, line 76: This need to be changed according to the other comment.

runtime/drivers/azure/azure.go
```diff
@@ -0,0 +1,257 @@
...
+
+func (d driver) Drop(config map[string]any, logger *zap.Logger) error {
+	return drivers.ErrDropNotSupported
+}
+
+func (d driver) Spec() drivers.Spec {
+	return spec
+}
+
+func (d driver) HasAnonymousSourceAccess(ctx context.Context, src drivers.Source, logger *zap.Logger) (bool, error) {
```
k-anshul, line 104: Need to adapt this for public sources.

runtime/drivers/azure/azure.go
k-anshul: Also needs `allow_host_access`. See other drivers.

runtime/drivers/azure/azure.go
```diff
@@ -0,0 +1,257 @@
...
+}
+
+// DownloadFiles returns a file iterator over objects stored in azure blob storage.
+func (c *Connection) DownloadFiles(ctx context.Context, source *drivers.BucketSource) (drivers.FileIterator, error) {
+	conf, err := parseSourceProperties(source.Properties)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse config: %w", err)
+	}
+
+	name := os.Getenv("AZURE_STORAGE_ACCOUNT")
```
k-anshul, line 188: We need to check `allow_host_access` before looking at the env variables. If `allow_host_access` is true we should look at the env variable else we should read from `configProperties`. 
Check s3 for reference.

runtime/drivers/azure/azure.go
```diff
@@ -0,0 +1,257 @@
...
+
+func parseSourceProperties(props map[string]any) (*sourceProperties, error) {
+	conf := &sourceProperties{}
+	err := mapstructure.Decode(props, conf)
+	if err != nil {
+		return nil, err
+	}
+	if !doublestar.ValidatePattern(conf.Path) {
+		// ideally this should be validated at much earlier stage
+		// keeping it here to have gcs specific validations
```
k-anshul, line 244: invalid comment

runtime/drivers/azure/azure.go
```diff
@@ -0,0 +1,257 @@
...
+		return nil, fmt.Errorf("glob pattern %s is invalid", conf.Path)
+	}
+	url, err := globutil.ParseBucketURL(conf.Path)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse path %q, %w", conf.Path, err)
+	}
+	if url.Scheme != "azblob" {
```
k-anshul, line 251: Is `az` also an alias for `azblob`. Check if paths can be configured as `az://container/key`

runtime/services/catalog/migrator/sources/sources_test.go
```diff
@@ -88,6 +88,9 @@ func TestConnectorWithSourceVariations(t *testing.T) {
 		//{"duckdb", "", map[string]any{
 		//	"sql": `select * from read_csv_auto('gs://scratch.rilldata.com/rill-developer/AdBids.csv.gz')`,
 		//}},
+		//{"duckdb", "", map[string]any{
```
k-anshul, line 91: Why this comment ?

web-common/src/features/sources/modal/yupSchemas.ts
```diff
@@ -69,6 +69,20 @@ export function getYupSchema(connector: V1ConnectorSpec) {
           .required("Source name is required"),
         project_id: yup.string().required("project_id is required"),
       });
+    case "azure":
+      return yup.object().shape({
+        path: yup
+          .string()
+          // .matches(/^https?:\/\//, 'Path must start with "http(s)://"')
+          .required("Path is required"),
```
k-anshul, line 77: This check should be similar to gcs or s3 ?

runtime/drivers/azure/azure.go
```diff
@@ -0,0 +1,257 @@
...
+
+func init() {
+	drivers.Register("azure", driver{})
+	drivers.RegisterAsConnector("azure", driver{})
+}
+
+var spec = drivers.Spec{
+	DisplayName:        "Azure Blob Storage",
+	Description:        "Connect to Azure Blob Storage.",
+	ServiceAccountDocs: "https://docs.rilldata.com/deploy/credentials/azure",
```
k-anshul, line 30: The docs also needs to be added.

runtime/services/catalog/migrator/sources/sources_test.go
rakeshsharma14317: Just added to test if this works or not, commented as Suggested by Aditya as it only when we want to test locally its working for such url or not)

runtime/drivers/azure/azure.go
rakeshsharma14317: checked, Even in documentation mentioned as `azblob`

runtime/drivers/azure/azure.go
rakeshsharma14317: Will take it in the end

runtime/drivers/azure/azure.go
```diff
@@ -0,0 +1,310 @@
...
+}
+
+// getClient returns a new azure blob client.
+func (c *Connection) getClient(ctx context.Context, conf *sourceProperties) (*container.Client, error) {
+	name := c.config.Account
+	key := c.config.Key
+
+	if c.config.AllowHostAccess {
+		name = os.Getenv("AZURE_STORAGE_ACCOUNT")
+		key = os.Getenv("AZURE_STORAGE_KEY")
```
k-anshul, line 271: What about `SAS_TOKEN` ?

runtime/drivers/azure/azure.go
```diff
@@ -0,0 +1,310 @@
...
+	}
+
+	credential, err := azblob.NewSharedKeyCredential(name, key)
+	if err != nil {
+		return nil, err
+	}
+
+	containerURL := fmt.Sprintf("https://%s.blob.core.windows.net/%s", name, conf.url.Host)
```
k-anshul, line 279: If else can be used instead of creating anonymous client by default.
```
var client *container.Client
if key == "" {
    client, err = container.NewClientWithNoCredential(containerURL, nil)
} else {
    client, err = container.NewClientWithSharedKeyCredential(containerURL, credential, nil)
}
if err != nil {
    return nil, err
}
```

runtime/drivers/azure/azure.go
```diff
@@ -0,0 +1,310 @@
...
+		client, err = container.NewClientWithSharedKeyCredential(containerURL, credential, nil)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return client, nil
+}
+
+func (c *Connection) openBucket(ctx context.Context, conf *sourceProperties) (*blob.Bucket, error) {
```
k-anshul, line 295: The name should signify that the bucker is opened with no credentials. Or can inline this with `HasAnonymousAccess` as well.

k-anshul (changes requested):
Additionally we need to add connector preflight APIs as well. Check `connectors.proto`. But I guess those can be added in a separate PR as well.

runtime/drivers/azure/azure.go
```diff
@@ -0,0 +1,332 @@
...
+			Key:         "path",
+			DisplayName: "Blob URI",
+			Description: "Path to file on the disk.",
+			Placeholder: "azblob://container-name/path/to/file.csv",
+			Type:        drivers.StringPropertyType,
+			Required:    true,
+			Hint:        "Glob patterns are supported",
+		},
+		{
+			Key:         "azure.storage.account",
```
k-anshul, line 41: Make it `azure.credentials`. Check `s3.go`.

runtime/drivers/azure/azure.go
```diff
@@ -0,0 +1,332 @@
...
+
+type driver struct{}
+
+type configProperties struct {
+	Account         string `mapstructure:"azure_storage_account"`
+	Key             string `mapstructure:"azure_storage_key"`
+	SASToken        string `mapstructure:"azure_storage_sas_token"`
+	StorageDomain   string `mapstructure:"azure_storage_domain,default:blob.core.windows.net"`
+	IsCDN           bool   `mapstructure:"azure_storage_is_cdn,default:false"`
+	IsLocalEmulator bool   `mapstructure:"azure_storage_is_local_emulator,default:false"`
```
k-anshul, line 73: As I understood it is only for developmental purposes so we can ignore `azure_storage_is_local_emulator`.
`azure_storage_domain` and `azure_storage_is_cdn` seems to be optional configs. Let's ignore them and only add if some user asks for them otherwise we need to take the values every time we do `rill env configure`.

runtime/drivers/azure/azure.go
```diff
@@ -0,0 +1,332 @@
...
+	}
+
+	var client *container.Client
+	if opts.SASToken != "" {
+		serviceURL, err := azureblob.NewServiceURL(opts)
+		if err != nil {
+			return nil, err
+		}
+
+		client, err = azureblob.NewDefaultClient(serviceURL, azureblob.ContainerName(conf.url.Host))
```
k-anshul, line 295: This internally still looks at the env variables. We need to make sure that if `c.config.AllowHostAccess` we should not access env variables.

runtime/drivers/azure/azure.go
```diff
@@ -0,0 +1,332 @@
...
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		// Create container url of the Azure Storage account.
+		url := fmt.Sprintf("https://%s.blob.core.windows.net", opts.AccountName)
+		credential, err := azidentity.NewDefaultAzureCredential(nil)
```
k-anshul, line 302: This needs to be done only when `c.config.AllowHostAccess` is true.

k-anshul (changes requested):

runtime/drivers/azure/azure.go
```diff
@@ -0,0 +1,363 @@
...
+			if err != nil {
+				return nil, err
+			}
+
+			containerURL, err := url.JoinPath(string(serviceURL), conf.url.Host)
+			if err != nil {
+				return nil, err
+			}
+
+			client, err = container.NewClientWithNoCredential(containerURL, azClientOpts)
```
k-anshul, line 317: Same as above. Pass `nil` instead of default `azClientOpts`

runtime/drivers/azure/azure.go
```diff
@@ -0,0 +1,363 @@
...
+			if err != nil {
+				return nil, fmt.Errorf("failed container.NewClientWithNoCredential: %w", err)
+			}
+		} else {
```
k-anshul, line 321: 1. This case should be outside of the parent if case with a condition that `allow_host_access` is true since if user has done az login they don't need to set `account_name`.
2. As discussed verify if we need to enable `DisableInstanceDiscovery` during `azidentity.NewDefaultAzureCredential`
3. Pass nil instead of `azClientOpts`

runtime/drivers/azure/azure.go
```diff
@@ -0,0 +1,363 @@
...
+			}
+		}
+	} else if connectionString != "" {
+		var err error
+		client, err = container.NewClientFromConnectionString(connectionString, conf.url.Host, azClientOpts)
+		if err != nil {
+			return nil, fmt.Errorf("failed container.NewClientFromConnectionString: %w", err)
+		}
+	} else {
+		return nil, errors.New("internal error, no account name")
```
k-anshul, line 338: This is not an internal error. Need to throw `drivers.NewPermissionDeniedError`.

k-anshul (changes requested):

runtime/drivers/azure/azure.go
```diff
@@ -0,0 +1,363 @@
...
+	opts := rillblob.Options{
+		GlobMaxTotalSize:      conf.GlobMaxTotalSize,
+		GlobMaxObjectsMatched: conf.GlobMaxObjectsMatched,
+		GlobMaxObjectsListed:  conf.GlobMaxObjectsListed,
+		GlobPageSize:          conf.GlobPageSize,
+		GlobPattern:           conf.url.Path,
+		ExtractPolicy:         conf.extractPolicy,
+	}
+
+	// Permission error will be thrown while iterating the bucket object, default error will be for az login, considering that as primary auth
```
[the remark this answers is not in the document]
rakeshsharma14317: added previously, will explain over call

runtime/services/catalog/migrator/sources/embedded_sources.go
```diff
@@ -55,6 +55,8 @@ func parseEmbeddedSourceConnector(path string) (string, string, bool) {
 			connector = "s3"
 		case "gs":
 			connector = "gcs"
+		case "azblob":
+			connector = "azure"
```
begelundmuller, line 59: Should be `case "azure"`

begelundmuller (changes requested):

runtime/drivers/azure/azure.go
```diff
@@ -0,0 +1,423 @@
...
+	}
+	if c.config.ConnectionString != "" {
+		connectionString = c.config.ConnectionString
+	}
+
+	if conf.AccountName != "" {
+		accountName = conf.AccountName
+	} else {
+		conf.AccountName = accountName
+	}
```
begelundmuller, line 324: This doesn't immediately appear necessary, but then I realized there's a very hard to see dependency here for propagating the correct account name to `createAnonymousClient`.

I would suggest refactoring the handling of the account name to make this simple and clear (even duplicating the logic would be better than this hidden dependency)

runtime/drivers/azure/azure.go
```diff
@@ -0,0 +1,423 @@
...
+			return nil, fmt.Errorf("failed azidentity.NewDefaultAzureCredential: %w", err)
+		}
+		client, err := container.NewClient(containerURL, cred, nil)
+		if err != nil {
+			return nil, fmt.Errorf("failed container.NewClient: %w", err)
+		}
+		return client, nil
+	}
+
+	return nil, drivers.NewPermissionDeniedError(fmt.Sprintf("can't access remote host without credentials, err:%v", errors.New("no credentials provided")))
```
begelundmuller, line 390: Can just make the error `... without credentials: no credentials provided`

runtime/services/catalog/migrator/sources/sources.go
```diff
@@ -300,7 +300,7 @@ func mergeFromParsedQuery(apiSource *runtimev1.Source, env map[string]string, re
 		if err != nil {
 			return err
 		}
-	case "s3", "gcs":
+	case "s3", "gcs", "azure":
 		apiSource.Connector = c
 		props["path"] = p
```
begelundmuller, line 305: I think similar case handling for `azure` is needed these places:

- Line 71 in `runtime/drivers/duckdb/transporter/duckdb_to_duckdb.go`
- Line 170 in `runtime/connections.go`

begelundmuller (changes requested):

begelundmuller (approved):

k-anshul (approved):

PR #3051, opened 2023-09-08
Fix up `MenuItem` component

ericpgreen2:
Our `MenuItem` component had a few shortcomings that I had to address in https://github.com/rilldata/rill/pull/3047. To ensure there aren't any side effects to these changes, I've created this PR to test/accept it in isolation.

Specifically:
- The background color on hover didn't fill the full width of some menus
- Focus was applied on hover
- Clicks weren't propagating to parent components
- Content placed in the "right" slot wasn't centered

### DIFF
--- a/web-common/src/components/menu/core/MenuItem.svelte
+++ b/web-common/src/components/menu/core/MenuItem.svelte
@@ -71,6 +71,12 @@
   }
   let hovered = false;
 
+  function onMouseOver() {
+    if (!disabled) {
+      hovered = true;
+    }
+  }
+
   function onFocus() {
     if (!disabled) {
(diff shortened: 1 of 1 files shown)


djbarnwal (approved):
LGTM. 
A small concern is that we are not breaking any existing menus by propagating clicks to the parents. Briefly checked the dashboard and it seems to work fine.

ericpgreen2:
> A small concern is that we are not breaking any existing menus by propagating clicks to the parents. Briefly checked the dashboard and it seems to work fine.

Agreed. I checked as well and didn't find anything, but I think we should merge and keep an eye out in our QA for this release.

PR #3052, opened 2023-09-08
Cloud UI: don't truncate dashboard titles

ericpgreen2:
This PR switches titles in dashboard listings from using a fixed width container + truncation to using a variable width container + overflow scroll. This ensures you can always see the full dashboard title.


djbarnwal (approved):
LGTM

PR #3057, opened 2023-09-11
Simplify connector source and sink props

begelundmuller:
- Removes source and sink types in favour of `map[string]any`
- Enables moving property parsing out of reconcile into the connectors (needed for async reconcile)


runtime/reconcilers/source.go
```diff
@@ -363,70 +363,12 @@ func (r *SourceReconciler) ingestSource(ctx context.Context, src *runtimev1.Sour
...
-			return nil, fmt.Errorf("property \"sql\" is mandatory for connector \"bigquery\"")
-		}
-		return &drivers.DatabaseSource{
-			SQL:   query,
-			Props: props,
-		}, nil
-	default:
-		return nil, fmt.Errorf("source connector %q not supported", conn.Driver())
-	}
+	// props["extract"] = src.Policy // TODO: add
```
k-anshul, line 368: I think this todo can be removed as well.

k-anshul (approved):

PR #3065, opened 2023-09-12
Postgres connector

k-anshul:


runtime/drivers/postgres/postgres.go
```diff
@@ -15,6 +15,37 @@ import (
...
+			Placeholder: "postgresql://user:password@localhost/mydb",
+			Hint:        "Either set this or configure PG_DATABASE_URL env variable.",
+		},
+	},
+	ConfigProperties: []drivers.PropertySchema{
+		{
+			Key:    "pg_database_url",
+			Secret: true,
+		},
+	},
```
begelundmuller, line 48: The `pg_` prefix feels redundant here, since it's in the scope of a `type: postgres`. Would suggest just `database_url`

runtime/drivers/postgres/postgres.go
```diff
@@ -15,6 +15,37 @@ import (
...
+			DisplayName: "SQL",
+			Description: "Query to extract data from Postgres.",
+			Placeholder: "select * from table;",
+		},
+		{
+			Key:         "pg_database_url",
+			DisplayName: "Postgress Connection String",
+			Type:        drivers.StringPropertyType,
+			Required:    false,
+			Href:        "https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING",
```
begelundmuller, line 38: Since it's a "connection URI", not "connection string", should it link to `https://www.postgresql.org/docs/current/libpq-connect.html#id-1.7.3.8.3.6` instead?

runtime/drivers/postgres/postgres.go
```diff
@@ -15,6 +15,37 @@ import (
...
+			Placeholder: "postgresql://user:password@localhost/mydb",
```
begelundmuller, line 39: Use the default localhost connection string instead – IIRC it's `postgresql://postgres@localhost:5432/postgres` (but please double check)

runtime/drivers/postgres/postgres.go
```diff
@@ -15,6 +15,37 @@ import (
...
+			Hint:        "Either set this or configure PG_DATABASE_URL env variable.",
```
begelundmuller, line 40: Suggestion: "Either set this or pass `--env connectors.postgres.database_url=...` to `rill start`"

runtime/services/catalog/migrator/sources/sources.go
```diff
@@ -361,6 +362,13 @@ func connectorVariables(src *runtimev1.Source, env map[string]string, repoRoot s
 		vars["dsn"] = repoRoot
 	case "bigquery":
 		vars["google_application_credentials"] = env["google_application_credentials"]
+	case "postgres":
+		props := src.Properties.AsMap()
+		if dsn, ok := props["pg_database_url"]; ok {
+			vars["dsn"] = dsn
+		} else if allowHostAccess, _ := vars["allow_host_access"].(bool); allowHostAccess {
+			vars["dsn"] = os.Getenv("PG_DATABASE_URL")
+		}
```
begelundmuller, line 371: 1. Can we remove this and instead rely on only scoped variables (`connector.name.database_url`) – we don't have backwards compatibility to worry about here
2. Instead of `"dsn"`, can we refactor the driver to consume `database_url` instead (for consistency to avoid hard-coding property name mappings between code files and internals)

runtime/drivers/postgres/sql_store.go
```diff
@@ -0,0 +1,190 @@
...
+	// load pgx driver
+	_ "github.com/jackc/pgx/v4/stdlib"
+)
+
+// Query implements drivers.SQLStore
+func (c *connection) Query(ctx context.Context, props map[string]any) (drivers.RowIterator, error) {
+	sql, ok := props["sql"].(string)
+	if !ok {
+		return nil, fmt.Errorf("property \"sql\" is mandatory for connector \"motherduck\"")
+	}
```
begelundmuller, line 23: 1. Not motherduck
2. Consider adding `sourceProperties` struct like other connectors to make it easier to add new properties in the future

runtime/drivers/postgres/sql_store.go
```diff
@@ -0,0 +1,190 @@
...
+
+	schema, err := rowsToSchema(res)
+	if err != nil {
+		return nil, err
+	}
+
+	slices.SortFunc(schema.Fields, func(a, b *runtimev1.StructType_Field) bool {
+		return a.Name < b.Name
+	})
```
begelundmuller, line 37: Why is this needed? I think it will be unexpected for sources to have a different column ordering than the SQL statement

runtime/drivers/postgres/sql_store.go
```diff
@@ -0,0 +1,190 @@
...
+	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
+	"github.com/rilldata/rill/runtime/drivers"
+	"golang.org/x/exp/slices"
+
+	// load pgx driver
+	_ "github.com/jackc/pgx/v4/stdlib"
+)
+
+// Query implements drivers.SQLStore
+func (c *connection) Query(ctx context.Context, props map[string]any) (drivers.RowIterator, error) {
```
begelundmuller, line 19: The `SourceProperties` make it seem like the `database_url` can be overriden in source props, but it's not consumed here (only in `Open`, which gets `ConfigProperties`)

runtime/drivers/postgres/sql_store.go
```diff
@@ -0,0 +1,190 @@
...
+	case "DOUBLE PRECISION", "FLOAT8":
+		t.Code = runtimev1.Type_CODE_FLOAT64
+	case "INTEGER", "INT", "INT4":
+		t.Code = runtimev1.Type_CODE_INT32
+	case "INTERVAL":
+		t.Code = runtimev1.Type_CODE_STRING // TODO - Consider adding interval type
+	case "JSON":
+		t.Code = runtimev1.Type_CODE_JSON
+	case "JSONB":
+		t.Code = runtimev1.Type_CODE_BYTES
```
begelundmuller, line 156: I think `JSON` would be more correct here

begelundmuller (changes requested):
Another question –

runtime/drivers/postgres/postgres.go
k-anshul: I had prefixed it with `pg` to keep it separate from `database_url` for other sql drivers that we may support in future. But given we plan to expose `connectors.postgres.database_url` to users as you indicated in other comments we can change it to `database_url`.

runtime/drivers/postgres/postgres.go
k-anshul: I think there is a password as well so I can use `postgresql://postgres:postgres@localhost:5432/postgres`

runtime/drivers/postgres/sql_store.go
k-anshul: Yeah that is one different implementation from other connectors. Everywhere else we are not opening any connection in `drivers.Open` at least for source connectors. May be I can keep the implementation same in postgres as well and only open connection to underlying infra in `Query` method. This would also make it easy to handle this comment : https://github.com/rilldata/rill/pull/3065#discussion_r1325080034
What do you think ?

runtime/drivers/postgres/postgres.go
begelundmuller: Okay, I think prefixing is better – leaning into named connectors also enable e.g. pulling data from two different Postgres databases into the same project

runtime/drivers/postgres/sql_store.go
begelundmuller: Seems okay to open it in `Query`. I wonder if we should remove the `database_url` from the source properties though, and require people to configure it as connector properties? That's how we do it for other connectors.

runtime/drivers/postgres/sql_store.go
```diff
@@ -0,0 +1,190 @@
...
+		t.Code = runtimev1.Type_CODE_TIMESTAMP
+	case "UUID":
+		t.Code = runtimev1.Type_CODE_UUID
+	case "VARCHAR":
+		t.Code = runtimev1.Type_CODE_STRING
+	case "POINT", "LINE", "LSEG", "BOX", "PATH", "POLYGON", "CIRCLE":
+		t.Code = runtimev1.Type_CODE_JSON // postgres predefined struct types, move to struct once appender supports structs
+	default:
+		// There are many datatypes in postgres, convert all to string
+		t.Code = runtimev1.Type_CODE_STRING
```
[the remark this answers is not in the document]
k-anshul: Arrays are already handled in the start of this function:
```
// type of array of base types being with _ like _FLOAT8
if strings.HasPrefix(dbt, "_") {
  // TODO :: use lists once appender supports it
  t.Code = runtimev1.Type_CODE_JSON
  return t
}
```

The `sql` interface of `pgx` driver already takes care of converting these other datatypes to string: https://github.com/jackc/pgx/blob/b301530a5fd50fe8ee46149fec58162a52bed8c6/stdlib/sql.go#L791

runtime/drivers/postgres/postgres.go
k-anshul: It accepts both types. Even string format `host=localhost port=5432 dbname=postgres user=postgres password=postgres` works.

runtime/drivers/postgres/sql_store.go
k-anshul: I was previously relying on a different approach to scan data. Agreed its no longer required now with current approach.

runtime/drivers/postgres/sql_store.go
```diff
@@ -0,0 +1,179 @@
...
+		return nil, err
+	}
+
+	var dsn string
+	if srcProps.DatabaseURL != "" { // get from src properties
+		dsn = srcProps.DatabaseURL
+	} else if url, ok := c.config["database_url"].(string); ok && url != "" { // get from driver configs
+		dsn = url
+	} else {
+		return nil, fmt.Errorf("require database_url to open postgres connection. Either set database_url in source yaml or pass --env connectors.postgres.database_url=... to rill start")
```
begelundmuller, line 31:
```suggestion
		return nil, fmt.Errorf("the property 'database_url' is required Postgres. Provide 'database_url' in the YAML properties or pass '--env connectors.postgres.database_url=...' to 'rill start'")
```

begelundmuller (approved):

PR #3068, opened 2023-09-12
Use mutex instead of tx for DuckDB drop and replace table

begelundmuller:
- We're seeing the DuckDB `.wal` explode in size when using a transaction to drop and replace a table. For a ~25GB dataset, we've observed a `.wal` size grow to ~250GB and freezing for almost an hour.
- This PR replaces the DB-level transaction with a `RWMutex` in the DuckDB driver. Regular queries take a read lock, "transactions" take a write lock.
- This loses the fault tolerance of transactions, but keeps consistency.


nishantmonu51 (approved):
👍

PR #3071, opened 2023-09-13
Emit ingestion metrics

esevastyanov:
This PR adds two metrics:
1. Source ingestion time - `ingestion_ms`
2. Number of bytes ingested - `ingestion_bytes`


runtime/reconcilers/source.go
```diff
@@ -392,10 +393,27 @@ func (r *SourceReconciler) ingestSource(ctx context.Context, src *runtimev1.Sour
...
+			attribute.String("destination", sinkConn.Driver()),
+			attribute.Bool("cancelled", errors.Is(outErr, context.Canceled)),
+			attribute.Bool("failed", outErr != nil),
+			attribute.Bool("limit_exceeded", limitExceeded),
+			attribute.Int64("limit_bytes", opts.LimitInBytes),
+		}
+		r.C.Activity.Emit(ctx, "ingestion_ms", float64(transferLatency), commonDims...)
+
+		// TODO: emit the number of bytes ingested (this might be extracted from a progress)
+	}()
```
begelundmuller, line 415: The `if limitExceeded` is checked before the `defer`. It would be best to move the `defer` to right after `transferStart` is declared (before calling `Transfer`)

runtime/services/catalog/migrator/sources/sources.go
```diff
@@ -220,10 +223,26 @@ func ingestSource(ctx context.Context, olap drivers.OLAPStore, repo drivers.Repo
...
+			attribute.Bool("cancelled", errors.Is(outErr, context.Canceled)),
+			attribute.Bool("failed", outErr != nil),
+			attribute.Bool("limit_exceeded", limitExceeded),
+			attribute.Int64("limit_bytes", ingestionLimit),
+		}
+		ac.Emit(ctx, "ingestion_ms", float64(transferLatency), commonDims...)
+		if p.unit == drivers.ProgressUnitByte {
+			ac.Emit(ctx, "ingestion_bytes", float64(p.catalogObj.BytesIngested), commonDims...)
+		}
+	}()
```
begelundmuller, line 245: Same comment as before

begelundmuller (changes requested):
The rest looks good!

begelundmuller (approved):

begelundmuller (approved):

PR #3072, opened 2023-09-13
Move source properties backwards compatibility rules to drivers

begelundmuller:
- Moves backwards compatibility rules for source properties into the drivers
- Enables the backwards compatibility rules to also work for the new reconcile

Note: The parsing of `fileSourceProperties` in the DuckDB transporters isn't the cleanest, since a) ideally they would be parsed in the source driver, not sink driver, b) some of them are actually sink properties. But that was also an issue before, so we can clean that up later.

### DIFF
--- a/runtime/services/catalog/artifacts/yaml/objects.go
+++ b/runtime/services/catalog/artifacts/yaml/objects.go
@@ -23,26 +23,27 @@ import (
  * This file contains the mapping from CatalogObject to Yaml files
  */
 type Source struct {
-	Type                  string
-	Path                  string         `yaml:"path,omitempty"`
-	CsvDelimiter          string         `yaml:"csv.delimiter,omitempty" mapstructure:"csv.delimiter,omitempty"`
-	URI                   string         `yaml:"uri,omitempty"`
-	Region                string         `yaml:"region,omitempty" mapstructure:"region,omitempty"`
-	S3Endpoint            string         `yaml:"endpoint,omitempty" mapstructure:"endpoint,omitempty"`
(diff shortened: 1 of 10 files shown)


runtime/services/catalog/artifacts/yaml/objects.go
```diff
@@ -23,27 +23,28 @@ import (
...
+	GlobPageSize                int            `yaml:"glob.page_size,omitempty" mapstructure:"glob.page_size,omitempty"`
+	BatchSize                   string         `yaml:"batch_size,omitempty" mapstructure:"batch_size,omitempty"`
+	HivePartition               *bool          `yaml:"hive_partitioning,omitempty" mapstructure:"hive_partitioning,omitempty"`
+	Timeout                     int32          `yaml:"timeout,omitempty"`
+	Format                      string         `yaml:"format,omitempty" mapstructure:"format,omitempty"`
+	Extract                     map[string]any `yaml:"extract,omitempty" mapstructure:"extract,omitempty"`
+	DuckDBProps                 map[string]any `yaml:"duckdb,omitempty" mapstructure:"duckdb,omitempty"`
+	Headers                     map[string]any `yaml:"headers,omitempty" mapstructure:"headers,omitempty"`
+	AllowSchemaRelaxation       *bool          `yaml:"allow_schema_relaxation,omitempty" mapstructure:"allow_schema_relaxation,omitempty"`
+	IngestAllowSchemaRelaxation *bool          `yaml:"ingest.allow_schema_relaxation,omitempty" mapstructure:"allow_schema_relaxation,omitempty"`
```
k-anshul, line 44: `mapstructure` also needs to be changed to `ingest.allow_schema_relaxation`. Seems like it is incorrect in main branch.

k-anshul (approved):

PR #3076, opened 2023-09-13
Make AcquireHandle work for any source connector

begelundmuller:
This PR makes the changes necessary to enable async reconcile to generically use `AcquireHandle` to open any source connector (like `s3`, `bigquery`, `duckdb`, etc.).

- Incorporates connector variables for implicit source connectors (like `connector: s3`) into `AcquireHandle`, enabling generic use of `AcquireHandle` for source ingestion in async reconcile.
- Renames default "olap" driver to the driver name. This makes our connector semantics more consistent, since we can generically call `AcquireHandle` for source connectors, even when the source has `connector: duckdb`.
- To simplify the code, will always evict cached connections when an instance is edited.
- Makes the new parser set the OLAP connector as the default connector for models when no other connector is explicitly set (async reconcile enables setting a connector not just for sources, but also for models).


runtime/registry.go
```diff
@@ -81,29 +68,14 @@ func (r *Runtime) DeleteInstance(ctx context.Context, instanceID string, dropDB
 	return r.Registry().DeleteInstance(ctx, instanceID)
 }
 
-// EditInstance edits exisiting instance.
-// The API compares and only evicts caches if drivers or dsn is changed.
-// This is done to ensure that db handlers are not unnecessarily closed
+// EditInstance edits exisiting instance. Calling it will evict all connection handles associated with the instance.
 func (r *Runtime) EditInstance(ctx context.Context, inst *drivers.Instance) error {
-	_, err := r.connectorDef(inst, inst.OLAPConnector)
```
k-anshul, line 88: Although we agreed to keep such checks in reconcile but it can still be useful to keep rudimentary API parameter checks.

begelundmuller: I thought about it, but there's very little to check here – in theory, neither are required, and also, the OLAP connector could be defined in `rill.yaml` (so checks would fail here). In practice of course, we always set them to more or less hard-coded values.

k-anshul: I actually had this issue in mind: https://github.com/rilldata/rill/pull/3066
Without this check it would have broken an existing running project. 

>the OLAP connector could be defined in rill.yaml

But while creating instance we would always define a project connector right ?

begelundmuller: It's a valid point, but worst case is that people see a "connector not found" error message on the project logs (at least won't fail silently). 

Yes in practice we currently hard-code them, but in theory we can support setting the OLAP connector to `myduckdb` and then only defining it in `rill.yaml`, like:

```
# rill.yaml
...
connectors:
- name: myduckdb
  type: duckdb
  dsn: path/to/my/duck.db
```

Additionally, there's a fair amount of functionality that can work with no OLAP or repo configured, like connector preflight, and ad-hoc reconcile resources (like the upcoming scheduled reports, which will be added to the catalog through API requests instead of from code files). 

Not needing to configure them can also be useful for testing.

k-anshul (approved):

PR #3083, opened 2023-09-14
Add access policy checks for new resource API

begelundmuller:
Didn't initially port over access policy checks in the new resource APIs. Doing that here.

The actual `ResolveMetricsViewSecurityV2` is not yet implemented for the `runtimev1.MetricsViewSpec`. To avoid too much duplicate code, will do that in the PR that will replace usage of `runtimev1.MetricsView` with `runtimev1.MetricsViewSpec` everywhere.

### DIFF
--- a/runtime/server/controller.go
+++ b/runtime/server/controller.go
@@ -2,14 +2,18 @@ package server
 
 import (
 	"context"
+	"errors"
 
 	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
+	"github.com/rilldata/rill/runtime"
+	"github.com/rilldata/rill/runtime/drivers"
 	"github.com/rilldata/rill/runtime/pkg/observability"
 	"github.com/rilldata/rill/runtime/server/auth"
 	"go.opentelemetry.io/otel/attribute"
 	"go.uber.org/zap"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/proto"
 )
 
 // GetLogs implements runtimev1.RuntimeServiceServer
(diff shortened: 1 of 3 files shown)


runtime/server/controller.go
```diff
@@ -44,7 +48,20 @@ func (s *Server) ListResources(ctx context.Context, req *runtimev1.ListResources
...
+		if err != nil {
+			return nil, status.Error(codes.InvalidArgument, err.Error())
+		}
+		if !access {
+			// Remove from the slice
+			rs[i] = rs[len(rs)-1]
+			rs[len(rs)-1] = nil
+			rs = rs[:len(rs)-1]
+		}
+		rs[i] = r
```
pjain1, line 63: This looks odd, as it puts back the resource in it was removed because of no access.

begelundmuller: Good cathc, it's a bug, fixed it

runtime/server/controller.go
```diff
@@ -44,7 +48,20 @@ func (s *Server) ListResources(ctx context.Context, req *runtimev1.ListResources
...
+		r := rs[i]
+		r, access, err := s.applySecurityPolicy(ctx, req.InstanceId, r)
+		if err != nil {
+			return nil, status.Error(codes.InvalidArgument, err.Error())
+		}
+		if !access {
+			// Remove from the slice
+			rs[i] = rs[len(rs)-1]
+			rs[len(rs)-1] = nil
+			rs = rs[:len(rs)-1]
```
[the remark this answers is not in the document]
begelundmuller: Not safe when using `for _, r := range rs`, that's why it uses a regular `for i := 0; i < len(rs); i++` instead. Then it's safe in this case (where items are only removed from the back).

pjain1 (approved):

PR #3088, opened 2023-09-14
Forward Kafka log events to a logger

esevastyanov:


cli/cmd/runtime/start.go
```diff
@@ -126,7 +126,7 @@ func StartCmd(cliCfg *config.Config) *cobra.Command {
 			case "", "noop":
 				sink = activity.NewNoopSink()
 			case "kafka":
-				sink, err = activity.NewKafkaSink(conf.ActivitySinkKafkaBrokers, conf.ActivitySinkKafkaTopic)
+				sink, err = activity.NewKafkaSink(conf.ActivitySinkKafkaBrokers, conf.ActivitySinkKafkaTopic, *logger)
```
begelundmuller, line 129: The Zap logger is supposed to be passed as `*zap.Logger` (that's how it's returned from the library). So should update `KafkaSink` to take a `*zap.Logger` instead.

begelundmuller (changes requested):

begelundmuller (approved):

PR #3095, opened 2023-09-18
create a view on ingested table

k-anshul:
Rename source leaks data.

### DIFF
--- a/runtime/drivers/duckdb/transporter/utils.go
+++ b/runtime/drivers/duckdb/transporter/utils.go
@@ -11,7 +11,9 @@ import (
 )
 
 type sinkProperties struct {
-	Table string `mapstructure:"table"`
+	Table         string `mapstructure:"table"`
+	Schema        string `mapstructure:"schema"`
+	FullTableName string `mapstructure:"_"`
 }
 
 func parseSinkProperties(props map[string]any) (*sinkProperties, error) {
@@ -19,6 +21,7 @@ func parseSinkProperties(props map[string]any) (*sinkProperties, error) {
 	if err := mapstructure.Decode(props, cfg); err != nil {
 		return nil, fmt.Errorf("failed to parse sink properties: %w", err)
 	}
+	cfg.FullTableName = fullTableName(cfg)
 	return cfg, nil
 }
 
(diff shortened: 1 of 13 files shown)


runtime/drivers/duckdb/transporter/utils.go
```diff
@@ -11,14 +11,17 @@ import (
 )
 
 type sinkProperties struct {
-	Table string `mapstructure:"table"`
+	Table         string `mapstructure:"table"`
+	Schema        string `mapstructure:"schema"`
+	FullTableName string `mapstructure:"_"`
```
begelundmuller, line 16: 1. Should use `-` instead of `_` for the mapstructure

runtime/reconcilers/util.go
```diff
@@ -99,17 +100,39 @@ func olapDropTableIfExists(ctx context.Context, c *runtime.Controller, connector
...
+				if _, err = conn.ExecContext(ensuredCtx, fmt.Sprintf("DROP TABLE IF EXISTS rill_sources.%s", safeSQLName(tableName))); err != nil {
+					lastErr = err
+				}
+			}
+			return lastErr
+		}
+		// table
+		_, err = conn.ExecContext(ctx, fmt.Sprintf("DROP TABLE IF EXISTS %s", safeSQLName(table)))
+		return err
 	})
```
begelundmuller, line 136: 1. Needs to pass `true, false` to `WithConnection` (not `false, false`)
2. Reduce nesting by handling the table case early and exiting (see [reduce nesting](https://github.com/uber-go/guide/blob/master/style.md#reduce-nesting) in the style guide)
3. Also, you can't use the `conn` to execute a new query until the `rows` from the previous case has been closed. Dropping the shadow table should happen after the find step, not inside the loop.
4. Regex compiles are kind of expensive, and I'm a bit worried about edge cases of `sprintf` + regex compile. Wouldn't it be sufficient to do a prefix check? I would also recommend doing a double underscore `__` between the base name and numbers.

begelundmuller (changes requested):
It's also missing view+table creation in the new reconcile. Were you thinking of doing it in a followup? We are hopefully merging it this week, so it will be short-lived otherwise.

runtime/reconcilers/util.go
k-anshul: 1. I expected these queries to run fast so that's why marked long running as false. Is it always expected to be true for `WithConnection` ?
4. Without regex there can be some edge cases if user has two sources named `test` and `test__1` so with a prefix check underlying tables of both will pass the check.

k-anshul: May be we can add some other string between table name and timestamp but that will still not be 100% reliable.

begelundmuller: 1. Dropping a table won't necessarily run very fast, so best to mark it long-running
2. I don't think 100% reliable is possible in any case. But yeah something in between would increase safety, something like `[name]__shadow[ID]` should be pretty safe

PR #3097, opened 2023-09-19
list copying fix

k-anshul:


begelundmuller (changes requested):
I noticed this happens inside the goroutine:
```
it.localFiles = append(it.localFiles, file.Name())
```
Could that have a race condition where two goroutines append simultaneously, so one of the appends get overwritten? That would mean `len(it.localFiles)` would be one less than `it.index-start`.

If yes, I think an easy solution is to just append `filename` instead of `file.Name()`, and then moving the filename generation and appending out of the goroutine.

begelundmuller (approved):

PR #3098, opened 2023-09-19
Updating dashboards to use the new API

AdityaHegde:


web-admin/src/features/projects/dashboards.ts
```diff
@@ -13,6 +16,7 @@ export interface DashboardListItem {
   isValid: boolean;
 }
 
+// TODO: use the creator pattern to get rid of the raw call to http endpoint
```
ericpgreen2, line 19: Yeah, this raw HTTP call isn't ideal. Ideally, we'd have scoped `RuntimeProviders`, rather than 1 global one, and we could use a `QueryClient`.

Anyways, from a design perspective, I think we'll want to remove the dashboard lists from the home & org pages. That'll remove the need for this raw HTTP call. So, it's fine to leave this as a one-off raw HTTP call temporarily.

ericpgreen2:
Looking good 👍

AdityaHegde:

PR #3105, opened 2023-09-20
Runtime: Download files from object stores in the background

k-anshul:


runtime/drivers/blob/blobdownloader.go
```diff
@@ -82,6 +92,10 @@ func (opts *Options) validate() {
 	if opts.GlobPageSize == 0 {
 		opts.GlobPageSize = 1000
 	}
+	if opts.BatchSizeBytes == 0 {
+		// 5 GB
+		opts.BatchSizeBytes = 5 * 1024 * 1024 * 1024
```
nishantmonu51, line 97: given that we are defaulting to 2 slots, how about reducing this to 2G ?

runtime/drivers/blob/blobdownloader.go
```diff
@@ -138,6 +155,10 @@ func NewIterator(ctx context.Context, bucket *blob.Bucket, opts Options, l *zap.
 
 // Close frees the resources
 func (it *blobIterator) Close() error {
+	it.cancel()
+	if it.grp != nil {
+		_ = it.grp.Wait() // wait for background calls to complete
+	}
 	// remove temp dir since recursive paths created have to be removed as well
 	err := os.RemoveAll(it.tempDir)
```
begelundmuller, line 163: Since `Close` may be called on line 133, would suggest:
```suggestion
	// remove temp dir since recursive paths created have to be removed as well
	var err error
	if it.tempDir != "" {
		err = os.RemoveAll(it.tempDir)
	}
```

runtime/drivers/blob/blobdownloader.go
```diff
@@ -148,138 +169,86 @@ func (it *blobIterator) Close() error {
...
-			default:
-				// should not reach here
-				panic(fmt.Errorf("partial download not supported for extension %q", ext))
+	var err error
+	for totalSizeInBytes < it.opts.BatchSizeBytes {
+		fileName, ok := <-it.downloadedFiles
+		if !ok { // channel is closed
+			err = it.err
+			if errors.Is(err, it.err) {
+				err = nil
```
begelundmuller, line 207: I don't understand this

runtime/drivers/blob/blobdownloader.go
```diff
@@ -148,138 +169,86 @@ func (it *blobIterator) Close() error {
...
 		paths := it.nextPaths
-		it.index += len(paths)
 		it.nextPaths = nil
+		it.err = io.EOF
 		return paths, nil
 	}
 
+	if it.downloadedFiles == nil {
+		// start downloading files
+		it.downloadedFiles = make(chan downloadedFile, it.batchCount)
```
begelundmuller, line 189: The impact of a dynamic channel buffer size is really hard to reason about here. Would recommend refactoring to follow the style guide here: https://github.com/uber-go/guide/blob/master/style.md#channel-size-is-one-or-none

runtime/drivers/blob/blobdownloader.go
```diff
@@ -293,84 +262,60 @@ func (it *blobIterator) NextBatch(n int) ([]string, error) {
...
 			}
+			it.logger.Info("download complete", zap.String("object", obj.obj.Key), zap.Duration("duration", duration), observability.ZapCtx(it.ctx))
+			drivers.RecordDownloadMetrics(grpCtx, &drivers.DownloadMetrics{
+				Connector: "blob",
+				Ext:       ext,
+				Partial:   !downloadFull,
+				Duration:  duration,
+				Size:      size,
+			})
+			it.downloadedFiles <- downloadedFile{fileName: filename, sizeInBytes: size}
```
begelundmuller, line 310: If something goes wrong during insertion and the client calls `Close`, and the channel buffer is full, the background downloaders will hang forever while attempting to send to `downloadedFiles` here, because it is not consumed anymore.

runtime/drivers/blob/blobdownloader.go
```diff
@@ -426,12 +372,21 @@ func (it *blobIterator) plan() ([]*objectWithPlan, error) {
...
 	if len(items) == 0 {
 		return nil, fmt.Errorf("no files found for glob pattern %q", it.opts.GlobPattern)
 	}
+
+	if size < it.opts.BatchSizeBytes { // need to ingest complete batch in one go
+		it.batchCount = _concurrentBlobDownloadLimit
+	} else {
+		// may be just keep it some factor of _concurrentBlobDownloadLimit ?
+		it.batchCount = int(it.opts.BatchSizeBytes/(size/int64(matchCount))) + 1
+	}
```
begelundmuller, line 385: Not fully sure what this is trying to accomplish – can we avoid the `batchCount` entirely? It would be easier to reason about if we just rely on batch size in bytes and the concurrent download limit.

begelundmuller (changes requested):

runtime/drivers/blob/blobdownloader.go
k-anshul: We don't want to return `io.EOF` from `Next`. Clients check `hasNext` to check if more elements are present and any error returned from next is a genuine error.

runtime/drivers/blob/blobdownloader.go
begelundmuller: I don't think you can check a ctx while sending to a channel. You may need to drain the channel instead during close.

runtime/drivers/blob/blobdownloader.go
begelundmuller: I still don't understand – the code reads like:
```go
a := 10
if a == 10 {
  a = 0
}
```

runtime/drivers/blob/blobdownloader.go
k-anshul: Yeah I also think we should do some benchmarks for this and set it slightly defensively.

begelundmuller: Seems like a good idea

k-anshul: Done. Changed to 2G.

runtime/drivers/blob/blobdownloader.go
```diff
@@ -138,176 +154,164 @@ func NewIterator(ctx context.Context, bucket *blob.Bucket, opts Options, l *zap.
...
+			sizeInBytes += obj.obj.Size
 			// need to create file by maintaining same dir path as in glob for hivepartition support
 			filename := filepath.Join(it.tempDir, obj.obj.Key)
-			if err := os.MkdirAll(filepath.Dir(filename), os.ModePerm); err != nil {
-				return err
-			}
+			files = append(files, filename)
+			wg.Add(1)
+			g.Go(func() error {
+				defer wg.Done()
```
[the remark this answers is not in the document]
k-anshul: I didn't understand calls to both `g.Go` and `wg.Wait()` happens in the same goroutine, that goroutine can either be waiting on `g.Go` when it can't spawn new goroutines or `wg.Wait()` when enough data has been downloaded (with download in progress for some).

runtime/drivers/blob/blobdownloader.go
```diff
@@ -316,61 +320,28 @@ func (it *blobIterator) NextBatch(n int) ([]string, error) {
...
+			// check if we have downloaded enough data and can be pushed to the channel
+			if sizeInBytes >= it.opts.BatchSizeBytes {
+				wg.Wait()
+				if stop { // download failed, exit
+					it.err = g.Wait() // get the failure error
+					close(it.files)
+					return
+				}
+				it.files <- files
+				files = make([]string, 0)
```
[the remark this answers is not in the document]
k-anshul: Not with the current implementation since we want to download the complete batch in background.

runtime/drivers/blob/blobdownloader.go
```diff
@@ -138,176 +154,164 @@ func NewIterator(ctx context.Context, bucket *blob.Bucket, opts Options, l *zap.
...
 	g.SetLimit(_concurrentBlobDownloadLimit)
-	for i, obj := range it.objects[start:end] {
-		obj := obj
-		index := start + i // with repect to object slice
-		g.Go(func() error {
+	var stop bool
+	files := make([]string, 0)
+	var sizeInBytes int64
+	var wg sync.WaitGroup
+	for i := 0; i < len(it.objects) && !stop; i++ {
```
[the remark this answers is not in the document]
k-anshul: An alternate solution can be to just use `waitgroup` and do wait whenever we reach `_concurrentBlobDownloadLimit` or have downloaded (triggered download) of batch size.

begelundmuller (approved):

PR #3106, opened 2023-09-20
Async reconcile UI global listener

AdityaHegde:


web-common/src/features/entity-management/resources-store.ts
```diff
@@ -0,0 +1,128 @@
...
+import { derived, get, Readable, writable } from "svelte/store";
+
+/**
+ * Global resources store that maps file name to a resource and parse errors.
+ */
+export type ResourcesState = {
+  // this is just a mapping of file path to resource name
+  // storing the entire resource is not necessary since tanstack query will do that for the get resource api
+  resources: Record<string, V1ResourceName>;
+  errors: Record<string, Array<V1ParseError>>;
```
ericpgreen2, line 20: I’m thinking that tracking a file’s parse errors in this store is overloading the store’s purpose. A file with a parse error isn't yet associated with a resource.

Would it make sense to instead have a `useParseErrors(instanceId: string, fileName: string)` selector that operates on top of a TanStack Query call to the `GetResource` API for the `ProjectParser` resource?

web-common/src/features/entity-management/resource-invalidations.ts
```diff
@@ -0,0 +1,136 @@
...
+  if (MainEntities[res.name.kind]) {
+    resourcesStore.setResource(res.resource);
+    queryClient.refetchQueries(
+      // we only use individual kind's queries
+      getRuntimeServiceListResourcesQueryKey(instanceId, {
+        kind: res.name.kind,
+      })
+    );
+  } else if (
+    res.name.kind === ResourceKind.ProjectParser &&
```
ericpgreen2, line 59: On line 34, you're returning early if `ResourceKind.ProjectParser`

web-common/src/features/entity-management/resource-invalidations.ts
```diff
@@ -0,0 +1,136 @@
...
+    case ResourceKind.MetricsView:
+      return invalidateMetricsViewData(
+        queryClient,
+        resource.meta.name.name,
+        failed
+      );
+  }
+}
+
+async function invalidateRemovedResource(
```
ericpgreen2, line 102: `invalidateRemovedResource` --> `removeQueriesForResource`, given there's no invalidation happening

web-common/src/features/entity-management/resource-selectors.ts
```diff
@@ -6,11 +6,12 @@ import {
...
 export enum ResourceKind {
-  Source = "source",
-  Model = "model",
-  MetricsView = "metricsview",
-  // TODO: do a correct map based on backend code
+  ProjectParser = "rill.runtime.v1.ProjectParser",
+  Source = "rill.runtime.v1.SourceV2",
+  Model = "rill.runtime.v1.ModelV2",
+  MetricsView = "rill.runtime.v1.MetricsViewV2",
 }
```
ericpgreen2, line 13: It'd be nice if Orval generated these types. Any chance you know if there's something missing from the backend API definition?

web-common/src/features/entity-management/file-selectors.ts
```diff
@@ -0,0 +1,33 @@
+import { createRuntimeServiceListFiles } from "@rilldata/web-common/runtime-client";
+
+/**
+ * In dev mode we still need to get the files for left nav.
+ * This is because parse errors will not have a resource.
+ */
+export function useMainEntityFiles(
+  instanceId: string,
+  prefix: "sources" | "models" | "dashboards",
+  extension: ".yaml" | ".sql"
```
ericpgreen2, line 10: Since `extension` can be deduced from the `prefix`, it shouldn't be a parameter of the function, rather it should be computed within the function body

ericpgreen2 (commented):
Overall, looks pretty good 👍

web-common/src/features/entity-management/resources-store.ts
AdityaHegde: This is more performant. Instead of filtering through project parse errors for each instance of get errors we have a map at a single place. Lets think about performance from the get go.

web-common/src/features/entity-management/resource-invalidations.ts
AdityaHegde: I am checking if the kind is present in the map not its boolean value in 34. Lemme know if the comment doesnt make it clear.

web-common/src/features/entity-management/resource-selectors.ts
AdityaHegde: The enum sits in go rather than the proto definition. @begelundmuller Any reason to not add it to the proto?

web-common/src/features/entity-management/resources-store.ts
AdityaHegde: It also allows for getting errors from different files that map to a single resource easily.

ericpgreen2: Yeah, I see the performance optimization. But we're filtering a list of files/errors – wouldn't a project have to be crazy big for this to be human-perceptible?

My concern was with the semantics – a file with a parse error isn't associated with a resource (source/model/metrics view), so it feels weird that it'd be in a `resources-store`. If keeping this approach, maybe just a better name would be the `files-store`?

web-common/src/features/entity-management/resource-invalidations.ts
ericpgreen2: Ah whoops I misread that. If I understand this right, maybe this would be more clear:
```
const MainEntities = [ResourceKind.Source, ResourceKind.Model, ResourceKind.MetricsView]

...

if (res.name.kind in MainEntities || res.name.kind === ResourceKind.ProjectParser)
```

AdityaHegde:

PR #3109, opened 2023-09-20
Add CTA to sign up for a user test

ericpgreen2:
This PR adds to the `/welcome` page a CTA to sign up for a user test. [Notion PRD](https://www.notion.so/rilldata/Participant-Payments-200-AMZ-gift-cardt-ecf64f6d06524321b472a654ba63ac6b?d=22a2e42909844192993652638afb7eba)


djbarnwal (changes requested):
The figma mocks show that the CTA container should be the same width as the projects container but in the application it seems, its a bit smaller 

image

ericpgreen2:
> The figma mocks show that the CTA container should be the same width as the projects container but in the application it seems, its a bit smaller

Yep, fixed 👍

AdityaHegde (approved):

PR #3110, opened 2023-09-21
Async reconcile UI sources integration

AdityaHegde:


web-common/src/features/entity-management/resource-selectors.ts
```diff
@@ -30,6 +32,7 @@ export function useResource<T = V1Resource>(
         select: (data) =>
           selector ? selector(data?.resource) : data?.resource,
         enabled: !!name && !!kind,
+        queryClient,
```
ericpgreen2, line 35: Why pass in the `QueryClient`? We only have one global query client, so won’t that be picked up automatically?

AdityaHegde: Tanstack query tries to get the global one using context API, which fails in certain cases. Passing this in avoid it. We should solve this in a better way in a future PR.

AdityaHegde:

PR #3114, opened 2023-09-21
sort by context column in dimension table

bcolloran:
Adds the ability to sort by context columns in the dimension table


web-common/src/components/virtualized-table/core/ColumnHeader.svelte
```diff
@@ -25,8 +26,11 @@
   export let header;
   export let position: HeaderPosition = "top";
   export let enableResize = true;
+  // FIXME: pretty sure isSelected can be deprecated in favor of highlight
   export let isSelected = false;
   export let bgClass = "";
+  export let highlight: boolean;
```
AdityaHegde, line 32: This seems to be unused. Expected?

web-common/src/features/dashboards/dimension-table/dimension-table-utils.ts
```diff
@@ -271,3 +215,229 @@ export function estimateColumnSizes(
 
   return estimateColumnSize;
 }
+
+export function prepareVirtualizedDimTableColumns(
```
AdityaHegde, line 219: Feels weird having so many arguments. Since most of these are from `MetricsExplorerEntity` we could perhaps get that as an argument?

web-common/src/features/dashboards/dimension-table/dimension-table-utils.ts
```diff
@@ -271,3 +215,229 @@ export function estimateColumnSizes(
...
+  }
+}
+
+/**
+ * This function prepares the data for the dimension table
+ * from data returned by the createQueryServiceMetricsViewComparisonToplist
+ * API.
+ *
+ */
+export function prepareDimensionTableRows(
```
AdityaHegde, line 359: Same as above

web-common/src/features/dashboards/dashboard-utils.ts
```diff
@@ -0,0 +1,70 @@
...
+}
+
+export function prepareSortedQueryBody(
+  dimensionName: string,
+  measureNames: string[],
+  timeControls: TimeControlState,
+  sortMeasureName: string,
+  sortType: DashboardState_LeaderboardSortType,
+  sortAscending: boolean,
+  filterForDimension: {
```
AdityaHegde, line 35: You can use `V1MetricsViewFilter` as a type here.

web-common/src/features/dashboards/dimension-table/DimensionDisplay.svelte
```diff
@@ -178,129 +113,95 @@
...
-            description: comparison.description,
-            enableResize: false,
-            format: comparison.format,
-          };
-        }
-        return undefined;
-      })
-      .filter((column) => !!column);
-  }
+  $: tableRows = prepareDimensionTableRows(
```
AdityaHegde, line 160: I feel like we should merge `sortedQueryBody` `sortedQuery` and `tableRows` into a single thing since they are only used at the end in `tableRows`.

Not a 100% sure how exactly, but we could perhaps do a derived store from `createQueryServiceMetricsViewComparisonToplist`. Ideally matching the new pattern using StateManagers, but that can be a further iteration.

This is not a blocker for the immediate release.

AdityaHegde (commented):
Switching the sorting order for measure moves the comparison columns around. Feels a bit jarring. Is that part of the design?

web-common/src/components/virtualized-table/core/ColumnHeader.svelte
bcolloran: thanks, fixed this -- that was supposed to be used but apparently I didn't finish that cleanup!!

web-common/src/features/dashboards/dashboard-utils.ts
bcolloran: ah thanks! fixed that

web-common/src/features/dashboards/dimension-table/dimension-table-utils.ts
bcolloran: Hey @AdityaHegde, regarding this comment and the other two: I'm doing more work on this stuff to enable sorting by dimension values (have a PR underway right now). Would it be ok to accept this PR for now and I can clean this up as you suggest in the follow-up PR, since I'm making changes to this code anyway?

web-common/src/features/dashboards/dimension-table/dimension-table-utils.ts
bcolloran: Hey @AdityaHegde, regarding this comment and the other two: I'm doing more work on this stuff to enable sorting by dimension values (have a PR underway right now). Would it be ok to accept this PR for now and I can clean this up as you suggest in the follow-up PR, since I'm making changes to this code anyway?

web-common/src/features/dashboards/dimension-table/DimensionDisplay.svelte
bcolloran: (same as above)

Hey @AdityaHegde, regarding this comment and the other two: I'm doing more work on this stuff to enable sorting by dimension values (have a PR underway right now). Would it be ok to accept this PR for now and I can clean this up as you suggest in the follow-up PR, since I'm making changes to this code anyway?

AdityaHegde (approved):
We can take up the improvements in a follow up

web-common/src/features/dashboards/dimension-table/dimension-table-utils.ts
bcolloran: addressed in https://github.com/rilldata/rill/pull/3173/commits/10614a6fefc22766d3941464f9a5dbd829a922b5

PR #3115, opened 2023-09-21
Add button to export data from dimension detail table

ericpgreen2:
This PR adds a button to export data from the dimension detail table.

Along the way, this PR also:
- Adjusts the dimension detail table's `DimensionHeader` component to match the new Figma mocks. Specifically, the `Switch` and `Search` components needed slight tweaking.
- Updates the "Export Model Data" button to have the same aesthetics as the new "Export Dimension Detail Table" button

Note: The exports do not currently include the "Percent of Total" column because that column is computed client-side, not server-side. In the future, we can move this calculation server-side and include it in the export.


begelundmuller:
Hey, the API for exporting the comparison table is available now: https://github.com/rilldata/rill/pull/3119

web-common/src/features/dashboards/dimension-table/export-toplist.ts
```diff
@@ -0,0 +1,63 @@
...
+  metricViewName,
+  format,
+  timeControlStore,
+}: {
+  query: ReturnType<typeof createQueryServiceExport>;
+  metricViewName: string;
+  format: V1ExportFormat;
+  // we need this from argument since getContext is called to get the state managers
+  // which cannot run outside of component initialisation
+  timeControlStore: TimeControlStore;
```
djbarnwal, line 23: Can we call `getStateManagers()` inside the Export button component and pass that to the query here so that we have `ctx` for all the required parameters apart from `format`?

djbarnwal (approved):

web-common/src/features/dashboards/dimension-table/export-toplist.ts
ericpgreen2: Good callout. Admittedly I just replicated the approach taken in `rows-viewer/export-metrics.ts`.

PR #3120, opened 2023-09-22
Cloud UI: Add repo subpath to Github block

ericpgreen2:
This PR adds the repo subpath (if any) to the Github block.

It looks like this:
image


djbarnwal (approved):
Looks good code wise, but the subpath text looks very confusing without any description around it.

Approving so you can merge at your convenience.

ericpgreen2:
Very good call. Just added context 👍

PR #3122, opened 2023-09-23
Don't show comparison circles on exclude mode

djbarnwal:
Fix bug when comparison circles were being shown on exclude mode without the dimension being compared.


nishantmonu51 (approved):
👍

PR #3123, opened 2023-09-25
Blob downloader iterator refactor

begelundmuller:


k-anshul (changes requested):

runtime/drivers/blob/blobdownloader.go
```diff
@@ -111,110 +86,105 @@ func (opts *Options) validateLimits(size int64, matchCount int, fetched int64) e
...
-	// drain the channel
-	if it.files != nil {
-		select {
-		case <-it.files:
-		default:
+
+	// Drain downloadsCh until it's closed (to avoid the background goroutine hanging forever)
+	var stop bool
+	for !stop {
+		_, ok := <-it.downloadsCh
```
k-anshul, line 165: Won't this block till no new data is present ?

k-anshul: Sorry my bad, I see the channel is closed in all case in `downloadFiles`

k-anshul (approved):

PR #3138, opened 2023-09-26
chore: upgrade to Svelte 4

bcolloran:
upgrade to svelte 4 in the wake of a couple super frustrating svelte bugs!


bcolloran:
In case anyone ever looks at this again: this turned out to be a huge amount of work, in part because of the way we have split up our Javascript code into multiple folders with separate sveltekit configs.

PR #3142, opened 2023-09-27
duckdb upgrade to v0.9.0

k-anshul:
Changes identified so far:
1. extract('epoch' from  timestamp_column) returns nanoseconds as well so return type is double.

### DIFF
--- a/cli/pkg/local/app.go
+++ b/cli/pkg/local/app.go
@@ -113,7 +113,6 @@ func NewApp(ctx context.Context, ver config.Version, verbose, strict, reset bool
 		return nil, err
 	}
 
-	isDefault := false
 	// If the OLAP is the default OLAP (DuckDB in stage.db), we make it relative to the project directory (not the working directory)
 	isDefault := false
 	if olapDriver == DefaultOLAPDriver && olapDSN == DefaultOLAPDSN {
--- a/cli/pkg/local/app.go
+++ b/cli/pkg/local/app.go
@@ -113,6 +113,7 @@ func NewApp(ctx context.Context, ver config.Version, verbose, strict, reset bool
 		return nil, err
 	}
 
+	isDefault := false
 	// If the OLAP is the default OLAP (DuckDB in stage.db), we make it relative to the project directory (not the working directory)
 	isDefault := false
 	if olapDriver == DefaultOLAPDriver && olapDSN == DefaultOLAPDSN {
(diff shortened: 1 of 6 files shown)


runtime/drivers/duckdb/duckdb.go
```diff
@@ -130,7 +130,20 @@ func (d Driver) Open(cfgMap map[string]any, shared bool, ac activity.Client, log
...
-		return nil, err
+		if !strings.Contains(err.Error(), "created with an older, incompatible version of Rill") {
+			return nil, err
+		}
+		c.logger.Named("console").Info("removing db file created using old version")
+		tmpPath := cfg.DBFilePath + ".tmp"
+		_ = os.RemoveAll(tmpPath)
+		walPath := cfg.DBFilePath + ".wal"
+		_ = os.Remove(walPath)
+		_ = os.Remove(cfg.DBFilePath)
```
begelundmuller, line 141: As previously discussed, we should have a configProperty like `error_on_incompatible_version` and always return the error if it's true.

And the CLI should set that property to true in `rill start` if the user manually specified a DB path.

begelundmuller (changes requested):

runtime/drivers/duckdb/duckdb.go
k-anshul: I misunderstood our another slack conversation and thought we need to always delete the file. Will make these changes.

runtime/drivers/duckdb/duckdb.go
```diff
@@ -130,7 +130,20 @@ func (d Driver) Open(cfgMap map[string]any, shared bool, ac activity.Client, log
 	// Open the DB
 	err = c.reopenDB()
 	if err != nil {
-		return nil, err
+		if c.config.ErrorOnIncompatibleVersion || !strings.Contains(err.Error(), "created with an older, incompatible version of Rill") {
+			return nil, err
+		}
+		c.logger.Named("console").Info("removing db file created using old version")
+		tmpPath := cfg.DBFilePath + ".tmp"
```
begelundmuller, line 137:
```suggestion
		}

		c.logger.Named("console").Info("Resetting .db file because it was created with an older, incompatible version of Rill")

		tmpPath := cfg.DBFilePath + ".tmp"
```

begelundmuller (approved):

PR #3149, opened 2023-09-28
remove unused editable table components

bcolloran:
These are the components that were used in the old metrics editor. They are no longer used anywhere, and unlikely to be in the future. If they are ever needed we can copy them from an old branch, but they are just noise in the repo right now


ericpgreen2 (approved):
Nice, thanks for the cleanup

PR #3150, opened 2023-09-28
expands leaderboard clickable area

bcolloran:
Fixes a bug noticed by Mike -- on `main`, only clicking on the text in a leaderboard title opens the dimension table; this PR expands the clickable are to fill the space to the right of that.


ericpgreen2 (approved):
Works 👍

PR #3160, opened 2023-10-02
Runtime: Support extracting templating variables referenced in code files

rakeshsharma14317:


runtime/compilers/rillv1/template.go
```diff
@@ -144,13 +146,20 @@ func AnalyzeTemplate(tmpl string) (*TemplateMetadata, error) {
 		return nil, err
 	}
 
+	variables, err := ExtractVariablesFromTemplate(t.Tree)
+	if err != nil {
+		fmt.Printf("failed to extract variables from template: %v\n", err)
```
begelundmuller, line 151: Remove log

runtime/compilers/rillv1/template.go
```diff
@@ -278,3 +287,56 @@ func EvaluateBoolExpression(expr string) (bool, error) {
...
+		walkNodes(n.Pipe, fn)
+	case *parse.PipeNode:
+		for _, cmd := range n.Cmds {
+			walkNodes(cmd, fn)
+		}
+	case *parse.CommandNode:
+		for _, arg := range n.Args {
+			walkNodes(arg, fn)
+		}
+	}
```
begelundmuller, line 330: Just to be clear, add a `default:` case that just does `return nil`

runtime/compilers/rillv1/template_test.go
```diff
@@ -46,10 +49,22 @@ func TestAnalyze(t *testing.T) {
...
 				ResolvedWithPlaceholders: `SELECT * FROM <no value> WHERE hello='<no value>' AND world='<no value>'`,
 			},
 		},
+		{
+			name:     "variables",
+			template: `SELECT * FROM {{.env.partner_table_name}} WITH SAMPLING {{.env.partner_table_name}} .... {{.user.domain}}`,
+			want: &TemplateMetadata{
+				Refs:                     []ResourceName{},
+				Config:                   map[string]any{},
+				Variables:                map[string][]string{"env": {"partner_table_name", "partner_table_name"}, "user": {"domain"}},
```
begelundmuller, line 63: Even if referenced multiple times, it should only be returned once

runtime/compilers/rillv1/template.go
```diff
@@ -278,3 +287,56 @@ func EvaluateBoolExpression(expr string) (bool, error) {
...
+	variablesMap := make(map[string][]string)
+	// Iterate through the variables and split them into key and value and add them to the map
+	for _, variable := range variables {
+		// Split the variable into parts using dot (.) as the separator
+		parts := strings.Split(variable, ".")
+		key := parts[0]
+		value := parts[1]
+
+		variablesMap[key] = append(variablesMap[key], value)
+	}
```
begelundmuller, line 308: Instead of building a one-level map (anyway doesn't work for deeper nesting), just return the full variable name minus the initial `.`.

So something like:
```go
if strings.HasPrefix(var, ".") {
  var = var[1:]
  variablesMap[var] = true
}

return maps.Keys(variablesMap), nil
```

begelundmuller: Also, this can be done directly during the walk, no need for allocating the slice `variables`

runtime/compilers/rillv1/template.go
```diff
@@ -61,6 +62,7 @@ type TemplateResource struct {
 type TemplateMetadata struct {
 	Refs                     []ResourceName
 	Config                   map[string]any
+	Variables                map[string][]string
```
begelundmuller, line 65: Just `[]string`, not a map here. See other comments.

begelundmuller (changes requested):

runtime/compilers/rillv1/template.go
rakeshsharma14317: j`oinIdentifiers(vn.Ident)` will return variables after removing the first `.`
so just need to append that and return right?

Then I guess this should work

`func ExtractVariablesFromTemplate(tree *parse.Tree) ([]string, error) {
	variablesMap := make(map[string]bool)
	walkNodes(tree.Root, func(n parse.Node) {
		if vn, ok := n.(*parse.FieldNode); ok {
			v := joinIdentifiers(vn.Ident)
			variablesMap[v] = true
		}
	})

	return maps.Keys(variablesMap), nil
}`

For template: `SELECT * FROM {{.env.partner_table_name}} WITH SAMPLING {{.env.partner_table_name}} .... {{.user.domain}}` we output should be `[]string{"env.partner_table_name", "user.domain"}`

begelundmuller (approved):

mindspank:
Is this good to merge?

PR #3170, opened 2023-10-04
streaming logs API

pjain1:


runtime/controller.go
```diff
@@ -118,6 +126,165 @@ func NewController(ctx context.Context, rt *Runtime, instanceID string, logger *
...
+	for b.bufferSize > b.maxBufferSize && len(b.messages) > 0 {
+		removedMsg := b.messages[0]
+		b.messages = b.messages[1:]
+		b.bufferSize -= int64(removedMsg.Size)
+	}
+
+	for client := range b.clients {
+		client <- message
+	}
+}
```
begelundmuller, line 239: I'm wondering if there's a way to avoid continuously re-allocating the messages buffer? Some ideas are a) also having a limit on the number of messages and using a circular buffer, or b) serializing logs to a bytes format and having one circular bytes buffer of length `maxBufferSize` (that would also require changing the RPCs).

What do you think?

runtime/controller.go
```diff
@@ -118,6 +126,165 @@ func NewController(ctx context.Context, rt *Runtime, instanceID string, logger *
...
+	for i := startIdx; i != endIdx; {
+		logs = append(logs, messages[i])
+		if asc {
+			i++
+		} else {
+			i--
+		}
+	}
+	return logs
+}
```
begelundmuller, line 286: Some general comments:
1. It would be good to avoid passing the API signature types into the controller – instead, it could take a normal callback function, which the server wraps the API handler with. Take a look at the `ListResources` and `WatchResources` APIs in the async reconcile PR: https://github.com/rilldata/rill/pull/3092
2. Also consider implementing the `replay` in the API handler by calling `GetLogs` before starting a subscription, instead of propagating it here (perfect consistency for replay isn't needed). Again, can look at `WatchResources` for inspiration.
3. Should we also support a `page_size` and `page_token` for `GetLogs`? And maybe limiting replay? I'm thinking about the UX for showing a logs screen on the status page – might be a bit heavy to fetch 16MB logs at the start each time.

begelundmuller (changes requested):

runtime/runtime.go
```diff
@@ -32,6 +33,7 @@ type Runtime struct {
 	queryCache         *queryCache
 	securityEngine     *securityEngine
 	activity           activity.Client
+	controllerCache    map[string]*Controller
```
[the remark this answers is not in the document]
pjain1: this is a test code that I left in here by mistake.

runtime/controller.go
```diff
@@ -118,6 +126,165 @@ func NewController(ctx context.Context, rt *Runtime, instanceID string, logger *
...
+		attrs[attr.Key] = attr.Value.Any()
+		size += len(attr.Key) + len(attr.Value.String()) // approx size
+		return true
+	}
+	// hacky way of collecting attributes
+	record.Attrs(gatherAttr)
+	payload, err := structpb.NewStruct(attrs)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
pjain1: you mean for the payload ?

runtime/runtime.go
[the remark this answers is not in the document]
pjain1: didn't #3092 already has too many changes, would be easier to review here ? I can do it if you want.

runtime/controller.go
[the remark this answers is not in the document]
begelundmuller: Yeah

runtime/runtime.go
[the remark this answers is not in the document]
begelundmuller: Okay, if you remove the test code such it doesn't have merge conflicts with 3092 then having it separately is also fine

runtime/controller.go
pjain1: a. how much limit would be put, we will have to come up with some upper bound which will depend on the size of logs buffer size b. If we serialize and send the serilized stream, wouldn't client need to handle that ? What about dynamic linked list not backed by a slice so memory is reclaimed when log is truncated but it will have memory overhead of pointers.

runtime/controller.go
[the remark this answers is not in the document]
pjain1: you mean return a string having serialized json ?

cli/cmd/project/logs.go
```diff
@@ -0,0 +1,109 @@
...
+				Name:             name,
+			})
+			if err != nil {
+				return err
+			}
+
+			depl := proj.ProdDeployment
+			if depl == nil {
+				return nil
+			}
```
begelundmuller, line 53: Maybe return an error instead, something like "Project %q is not currently deployed"

cli/cmd/project/logs.go
```diff
@@ -0,0 +1,109 @@
...
+
+				go func() {
+					for {
+						res, err := logClient.Recv()
+						if err != nil {
+							fmt.Println("failed to receive logs: %w", err)
+							cancel()
+							break
+						}
+						fmt.Println(res.Log)
```
begelundmuller, line 81: Maybe format the log nicely using a similar format as `rill start` (I think it's just a Zap development logger)

cli/cmd/project/logs.go
```diff
@@ -0,0 +1,109 @@
...
+			}
+			return nil
+		},
+	}
+
+	logsCmd.Flags().SortFlags = false
+	logsCmd.Flags().StringVar(&name, "project", "", "Project Name")
+	logsCmd.Flags().StringVar(&path, "path", ".", "Project directory")
+	logsCmd.Flags().BoolVarP(&follow, "follow", "f", false, "Follow logs")
+	logsCmd.Flags().IntVarP(&tail, "tail", "t", -1, "Number of lines to show from the end of the logs, use -1 for all logs")
```
begelundmuller, line 106: What about defaulting to something less, maybe 10 if `follow` is true and 50 if it's false?

go.mod
```diff
@@ -60,6 +60,7 @@ require (
 	github.com/stretchr/testify v1.8.3
 	github.com/testcontainers/testcontainers-go v0.19.0
 	github.com/xuri/excelize/v2 v2.7.1
+	github.com/zeebo/assert v1.3.0
```
begelundmuller, line 63: Is this needed – we use `github.com/stretchr/testify` for testing assertions

runtime/controller.go
```diff
@@ -115,11 +117,41 @@ func NewController(rt *Runtime, instanceID string, logger *zap.Logger, ac activi
...
+	return d.logs.Add(record)
+}
+
+func (d *duplicatingHandler) WithAttrs(attrs []slog.Attr) slog.Handler {
+	panic("not implemented")
+}
+
+func (d *duplicatingHandler) WithGroup(name string) slog.Handler {
+	panic("not implemented")
+}
```
begelundmuller, line 153: 1. We should implement `WithAttrs` and `WithGroup` (it should be quite simple, something like  `return duplicatingHandler{zapHandler: d.zapHandler.With...}`)
2. Should we move the duplicatingHandler into `logbuffer` or another util package? Or alternatively move to the bottom of this file, since it's a util (but this file is already so big...).

begelundmuller (changes requested):
Looks good! 

Another question –have you tested this with lower limits while using the application for a while, to make sure it handles overflow cases and doesn't have too big performance overhead?

cli/cmd/project/logs.go
pjain1: you mean log this as json using logger created using `zap.NewDevelopment()`

cli/cmd/project/logs.go
pjain1: this is how kubectl logs work so just emulated it, but can make change if required

go.mod
pjain1: go land added that will remove

runtime/controller.go
pjain1: 1. `zapHandler.WithAttrs(attrs)` returns a `slog.Hanlder` not `zapslog.Handler` which is needed here. Also if we implement `WithAttrs` wouldn't we need to store attributes in duplicating handler so that they can be added to the payload of the `Log` in the buffer.
2. Don't think it belongs to the `logbuffer` package, probably we can rename the package to `logutils` and add it there.

pjain1: 1. It seems like one can simply typecast it to `slog.Handler`

runtime/controller.go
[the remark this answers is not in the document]
pjain1: using `string payload` instead of `google.protobuf.Struct payload` in the `Log` proto message ?

cli/cmd/project/logs.go
begelundmuller: I mean to make it look like the format in `rill start`:

```
❯ rill start dev-project
2023-10-23T10:06:33.736	INFO	Serving Rill on: http://localhost:9009
2023-10-23T10:07:06.080	INFO	Reconciling resource	{"name": "metrics_margin_monitoring", "kind": "Source"}
2023-10-23T10:07:06.101	INFO	Ingesting source data	{"name": "metrics_margin_monitoring", "connector": "duckdb"}
2023-10-23T10:07:06.877	INFO	Reconciled resource	{"name": "metrics_margin_monitoring", "kind": "Source", "elapsed": "798ms"}
2023-10-23T10:07:06.877	INFO	Reconciling resource	{"name": "metrics_margin_model", "kind": "Model"}
2023-10-23T10:07:06.882	INFO	Reconciled resource	{"name": "metrics_margin_model", "kind": "Model", "elapsed": "5ms"}
2023-10-23T10:07:06.882	INFO	Reconciling resource	{"name": "customer_margin_dash", "kind": "MetricsView"}
2023-10-23T10:07:06.888	INFO	Reconciled resource	{"name": "customer_margin_dash", "kind": "MetricsView", "elapsed": "6ms"}
```

cli/cmd/project/logs.go
begelundmuller: Okay, maybe it's not so bad for the non-follow case. For follow, I think it would be a nicer experience to get quite few logs by default (since most of the time, you would use it to know what happens after running the command, not before). 

What do you think?

runtime/controller.go
begelundmuller: Okay – I think we should still keep the `logbuffer` package for the log buffer since it seems like a self-contained thing. We can add a `logutil` package for the duplicating handler.

A thought: do you think all this would be easier if we didn't use `slog` and just used `zap` with a custom backend?

begelundmuller:
Something to do in a follow up PR as well: instead of creating the logger in the controller, I think we should move it to `runtime/registry.go` where the controller is created. We could store the logger on the `instanceWithController` struct – this would have some advantages:

- Allow us to propagate controller start/restart/error messages (including errors)
- Allows us to retain old logs even when the controller is restarted (which happens e.g. when connector config is changed)
- Makes it easier to use the same logger for driver handles (which are opened in `runtime/connection_cache.go`), so we can propagate e.g. source download logs

cli/cmd/project/logs.go
pjain1: no strong opinions here I am in habit of using kubectl logs, in any case lets keep the default value for tail consistent for follow and non-follow case

runtime/controller.go
pjain1: I think that will involve implementing `zapcore.Core` and use `zapcore.Tee`, will have to implement and see how easy it is

begelundmuller (approved):

PR #3171, opened 2023-10-04
DuckDB tables as external DB files

k-anshul:
Adds a `duckdb-ext-storage` driver thats adds each source in a separate duckdb file.

TODOS:
- [ ] `AddTableColumn` and `AlterTableColumn` should take runtime.Types as inputs.
- [x] Add a `version.txt` file in source directory for fetching current version instead of relying on latest file.
- [ ] Pass `priority`,`longrunning`,`tx` as CRUD API options. Fix model creation queries running with high priority.
- [ ] Rethink `WithConnection` given CRUD APIs operate one level above driver APIs
- [ ] Test Motherduck and SQLite
- [ ] Avoid use of `txMu` (closes db and reopens again) during renames

Open questions:
- [ ] Keep same directory structure even if using single db file on cloud (i.e. `<instance_id>/main.db`)? 
- [x] Should we add `table_as_view` as query param to make it work for locally running application as well ? (**NO**)


runtime/drivers/duckdb/olap.go
```diff
@@ -181,6 +186,344 @@ func (c *connection) EstimateSize() (int64, bool) {
...
+		Query:    fmt.Sprintf("ALTER TABLE %s.default ADD COLUMN %s %s", safeSQLName(dbName), safeSQLName(columnName), typ),
+		Priority: 1,
+	})
+	if err != nil {
+		return err
+	}
+
+	// recreate view to propagate schema changes
+	return c.Exec(ctx, &drivers.Statement{
+		Query:    fmt.Sprintf("CREATE OR REPLACE VIEW %s AS SELECT * FROM %s.default", safeSQLName(tableName), safeSQLName(dbName)),
```
begelundmuller, line 219: Why does it need to re-create the view if only adding a column? It would still be doing `SELECT * FROM` the same underlying table.

runtime/drivers/duckdb/olap.go
```diff
@@ -181,6 +186,344 @@ func (c *connection) EstimateSize() (int64, bool) {
...
+		Query:       fmt.Sprintf("DETACH %s", safeSQLName(oldDB)),
+		Priority:    100,
+		LongRunning: true,
+	})
+	if err != nil {
+		return err
+	}
+
+	// delete source directory
+	return os.RemoveAll(sourceDir)
```
begelundmuller, line 377: 1. Only necessary to ignore query cancellations after the first query
2. Is DETACH safe to call multiple times? If not, it should ignore the error to ensure idempotency

runtime/drivers/duckdb/olap.go
```diff
@@ -181,6 +186,344 @@ func (c *connection) EstimateSize() (int64, bool) {
...
+		Query:       fmt.Sprintf("INSERT INTO %s.default %s (%s)", safeSQLName(dbName(name, version)), insertByNameClause, sql),
+		Priority:    1,
+		LongRunning: true,
+	})
+}
+
+// RenameTable implements drivers.OLAPStore.
+func (c *connection) RenameTable(ctx context.Context, oldName, newName string, view bool) error {
+	if oldName == newName {
+		// todo check case sensitivity
```
begelundmuller, line 413: Must not check case sensitivity – those actually need to be renamed

runtime/drivers/duckdb/olap.go
```diff
@@ -181,6 +186,344 @@ func (c *connection) EstimateSize() (int64, bool) {
...
+	}
+
+	// ignore query cancellations
+	ctx = context.WithValue(context.Background(), connCtxKey{}, connFromContext(ctx))
+	newSrcDir := filepath.Join(filepath.Dir(c.config.DBFilePath), newName)
+	// new name should not exist
+	exist := dirExist(newSrcDir)
+	if exist {
+		return fmt.Errorf("table %q already exist", newName)
+	}
```
begelundmuller, line 438: I don't think it should fail here – don't renames need to be implemented as force renames? Otherwise, you would need to drop the previous table before renaming, causing a period of failing queries

runtime/drivers/duckdb/olap.go
```diff
@@ -181,6 +186,344 @@ func (c *connection) EstimateSize() (int64, bool) {
...
+
+	// rename filename
+	newVersion := fmt.Sprint(time.Now().UnixMilli())
+	newDB := dbName(newName, newVersion)
+	newFilePath := filepath.Join(newSrcDir, fmt.Sprintf("%s.db", newVersion))
+	err = os.Rename(filepath.Join(newSrcDir, fmt.Sprintf("%s.db", oldVersion)), newFilePath)
+	if err != nil {
+		return err
+	}
```
begelundmuller, line 482: Take a closer look at the consistent renaming logic in [Notion](https://www.notion.so/rilldata/Ingest-sources-in-a-new-DB-d35afe9eb0484c68a1ba6bcb4d5f1096?pvs=4#35c05c6633074c8c8f5eb02cb6f4da0a).

It needs to move the `.db` file itself in order to keep serving queries while the renaming is happening

runtime/drivers/duckdb/olap.go
```diff
@@ -221,3 +564,28 @@ func fileSize(paths []string) int64 {
 	}
 	return size
 }
+
+func dirExist(dir string) bool {
+	_, err := os.Stat(dir)
+	// error is always a path error
+	return err == nil
+}
```
begelundmuller, line 572: Check using `errors.Is` – could also be malformed path error for example

runtime/drivers/duckdb/olap.go
```diff
@@ -181,6 +186,344 @@ func (c *connection) EstimateSize() (int64, bool) {
...
+	for i := len(files) - 1; i >= 0; i-- {
+		file := files[i]
+		name, _, found := strings.Cut(file.Name(), ".db")
+		if !found { // .wal or .tmp file
+			continue
+		}
+		return name, true, nil
+	}
+	return "", false, nil
+}
```
begelundmuller, line 525: I think using `version.txt` would be much simpler and avoid various consistency issues around renames (like if they fail halfway through moving files)

runtime/drivers/duckdb/olap.go
```diff
@@ -181,6 +186,344 @@ func (c *connection) EstimateSize() (int64, bool) {
...
+	}
+
+	// check if some older version existed previously to detach it later
+	oldVersion, oldVersionExists, _ := c.tableVersion(name)
+
+	newVersion := fmt.Sprint(time.Now().UnixMilli())
+	dbFile := filepath.Join(sourceDir, fmt.Sprintf("%s.db", newVersion))
+	db := dbName(name, newVersion)
+	// attach new db
+	err := c.Exec(ctx, &drivers.Statement{Query: fmt.Sprintf("ATTACH '%s' AS %s", dbFile, safeSQLName(db))})
```
begelundmuller, line 288: What if the DB file name contains single quotes?

begelundmuller (changes requested):
It is also missing changes in `runtime/reconcilers/util.go`

runtime/drivers/duckdb/config.go
```diff
@@ -16,6 +16,8 @@ type config struct {
 	PoolSize int `mapstructure:"pool_size"`
 	// AllowHostAccess denotes whether to limit access to the local environment and file system
 	AllowHostAccess bool `mapstructure:"allow_host_access"`
+	// TableAsView controls if table is modelled as view
+	TableAsView bool `mapstructure:"table_as_view"`
```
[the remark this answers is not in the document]
k-anshul: Can keep `db_per_table `

runtime/drivers/duckdb/olap.go
k-anshul: This puzzled me as well but it seems view is no longer queryable if the schema of the underlying table changes.
image

runtime/drivers/duckdb/olap.go
```diff
@@ -181,6 +186,344 @@ func (c *connection) EstimateSize() (int64, bool) {
...
+		return err
+	}
+
+	if !exist {
+		return fmt.Errorf("table %q does not exist", tableName)
+	}
+	dbName := dbName(tableName, version)
+	// ignore query cancellations
+	ctx = context.WithValue(context.Background(), connCtxKey{}, connFromContext(ctx))
+	err = c.Exec(ctx, &drivers.Statement{
```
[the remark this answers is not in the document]
k-anshul: If the first query(Alter Table) succeeds then the view is no longer queryable so we need second query to succeed as well. Although the query can still fail due to any number of reasons but most probable because of query cancellations.

runtime/drivers/duckdb/olap.go
k-anshul: 2. No. I want the drop to fail if detach fails to release any resources with `attach`. I can add checks for db not found errors.

runtime/drivers/duckdb/olap.go
k-anshul: There is no force rename in duckdb SQL as well. Do we want this functionality to be different than duckdb ?

runtime/drivers/duckdb/olap.go
k-anshul: As per duckdb's implementation we can't rename `test` to `Test` so wondering if we should have the same impl here.

runtime/drivers/duckdb/olap.go
k-anshul: Will handle that. 
In general I think instead of handling all such special cases we should evaluate if we can make our lives easier and only allow alphanumeric characters and spaces in source, model, dashboard names. What difference is it going to make if a user must make their source `customer data` instead of `customer's data`

runtime/drivers/duckdb/duckdb.go
```diff
@@ -417,7 +438,31 @@ func (c *connection) reopenDB() error {
...
+		name := filepath.Dir(path)
+		version, found := strings.CutSuffix(filepath.Base(path), ".db")
+		if !found {
+			return nil
+		}
+		path = filepath.Join(dir, path)
+		db := dbName(name, version)
+		_, err := conn.ExecContext(context.Background(), fmt.Sprintf("ATTACH '%s' AS %s", path, safeSQLName(db)))
+		return err
+	})
```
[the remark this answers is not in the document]
k-anshul: 4. Should we explicitly check for version mismatch error ?

runtime/drivers/duckdb/olap.go
k-anshul: Also even in the current logic we are closing all running queries then dropping original table and renaming temporary table to the new table.

runtime/drivers/duckdb/config.go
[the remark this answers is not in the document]
begelundmuller: I like both, you decide :)

runtime/drivers/duckdb/olap.go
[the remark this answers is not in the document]
begelundmuller: Okay, then it should only use a non-cancellable ctx after the first query

runtime/drivers/duckdb/olap.go
begelundmuller: Weird...

runtime/drivers/duckdb/olap.go
begelundmuller: Why is that? The reconcile logic assumes that table drops are idempotent, i.e. act like `DROP TABLE IF EXISTS`, not `DROP TABLE`. 

Imagine pod shutdown halfway through a table drop – when it restarts, reconcile will re-run incomplete operations.

runtime/drivers/duckdb/olap.go
begelundmuller: That case is already handled in reconcile, try searching for `__rill_tmp_rename_`.

I'm indifferent about whether it's implemented here or in reconcile, it's a pretty rare case (maybe just keep it in reconcile to keep the changes here minimal?).

runtime/drivers/duckdb/olap.go
begelundmuller: There is a force rename in DuckDB:
```sql
BEGIN:
DROP TABLE IF EXISTS newName;
ALTER TABLE name RENAME TO newName;
COMMIT;
```
Which is how the current logic in reconcile works (though using a simulated transaction by passing `tx=true` to `WithConnection`).

runtime/drivers/duckdb/olap.go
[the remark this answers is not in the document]
k-anshul: yeah that should be done.

runtime/drivers/duckdb/olap.go
begelundmuller: Have seen some issues about this stuff in the past (mostly for column names though). It can sneak in – e.g. someone's tool exports a file named `Benjamin's export.csv` and then they drag+drop it into Rill. Doing sanitization of input names everywhere easily becomes as complex as handling complex names everywhere.

Even if we add upstream restrictions, I find it better to just handle it at every layer instead of needing to think through how those assumptions propagate. Kind of like layers of security.

runtime/drivers/duckdb/duckdb.go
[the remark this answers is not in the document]
k-anshul: If we have successfully updated `version.txt` then whatever `version.txt` says is the source of truth (which would also point to the right `db` file). If we drop the directory even though there is a correct `version.txt` we also need to remove the `view`.

runtime/drivers/duckdb/olap.go
k-anshul: Yeah agreed. I am saying we should ignore `db not found error` to ensure `idempotency` but fail on other errors.

runtime/drivers/duckdb/olap.go
k-anshul: Sure then I guess we need to add case insensitive check.

runtime/drivers/duckdb/olap.go
```diff
@@ -181,6 +186,374 @@ func (c *connection) EstimateSize() (int64, bool) {
...
+			return err
+		}
+
+		if oldVersionExists {
+			oldDB := dbName(name, oldVersion)
+			// ignore these errors since source has been correctly ingested and attached
+			c.detachAndRemoveFile(ctx, oldDB, filepath.Join(sourceDir, fmt.Sprintf("%s.db", oldVersion)))
+		}
+		return nil
+	})
```
begelundmuller, line 327: This doesn't need `WithConnection` if you move `updateVersion` up before the `CREATE OR REPLACE VIEW`.

runtime/drivers/duckdb/olap.go
```diff
@@ -181,6 +186,374 @@ func (c *connection) EstimateSize() (int64, bool) {
...
+func (c *connection) tableVersion(name string) (string, bool, error) {
+	pathToFile := filepath.Join(c.config.ExtStoragePath, name, "version.txt")
+	contents, err := os.ReadFile(pathToFile)
+	if err != nil {
+		if errors.Is(err, fs.ErrNotExist) {
+			return "", false, nil
+		}
+		return "", false, err
+	}
+	return string(contents), true, nil
```
begelundmuller, line 542: Just to be safe, add a `string.TrimSpace` here

runtime/drivers/duckdb/olap.go
k-anshul: Actually the `detachAndRemoveFile` need to be done with `ensuredCtx`. Missed it while doing refactor.

PR #3173, opened 2023-10-04
Sort dim table by dim value

bcolloran:
enables sorting dimension table by the dimension value


web-common/src/features/dashboards/stores/metrics-explorer-entity.ts
```diff
@@ -11,75 +11,112 @@ import type { V1MetricsViewFilter } from "@rilldata/web-common/runtime-client";
 
 export interface MetricsExplorerEntity {
   name: string;
-  // selected measure names to be shown
+  /**
+   * selected measure names to be shown
+   */
```
bcolloran, line 16: Note: the only changes in this file are converting `//` comments to `/** */` comments so that these descriptions flow through to intellisense

web-common/src/features/dashboards/dimension-table/DimensionHeader.svelte
```diff
@@ -18,12 +18,24 @@
...
 
+  const stateManagers = getStateManagers();
+  const {
+    selectors: {
+      sorting: { sortedByDimensionValue },
+    },
+    actions: {
+      sorting: { toggleSort },
+    },
+  } = stateManagers;
```
bcolloran, line 37: Note that this pattern allows you to easily search through existing actions and selectors using autocomplete, with full type-hinting and doc comments.

bcolloran:
Hey @AdityaHegde , this PR enables sorting by dimension values in the dimension table. I've also implemented some of the clean up you suggested in the last PR, though I still need to think about some of it.

I've also started migrating parts of the state to the new store mechanism as you suggested, though I've done it following the pattern that Hamilton and Speros and I discussed at length a couple months ago. Happy to go into the details, but the overview is:
- actions and selectors are attached to the store so that they can be easily found and navigated via autocomplete, and to reduce the number of imports needed
- actions and selectors are created by way of simple functions, and then at store creation time these functions are closed over a reference to the actual live data store.
- Doing things this way also allows us to make sure we can expose nice type guidance and intellisense docs

Happy discuss and tweak details, but I'm really happy with this direction and looking forward to migrating more of the dimension table and leaderboard code to this pattern to eliminate a bunch of the props drilling and repetition.

web-common/src/features/dashboards/state-managers/actions/index.ts
```diff
@@ -0,0 +1,99 @@
...
+ */
+type DashboardUpdaters2<T extends DashboardMutatorFns> = Expand<{
+  [P in keyof T]: (...params: DropFirst<Parameters<T[P]>>) => void;
+}>;
+
+/**
+ * Takes an object containing `DashboardMutatorFn`s,
+ * and returns an object of functions that directly update the dashboard.
+ */
+function createDashboardUpdaters<T extends DashboardMutatorFns>(
```
AdityaHegde, line 89: This is a lot better that what i had originally when backend was in nodejs. You can still see the code in ActionBase.ts. Having implicit types based on argument is a better approach!

AdityaHegde (approved):
Just a small refactor. But otherwise this is so much better than the massive file in dashboard-store!

web-common/src/features/dashboards/state-managers/actions/index.ts
```diff
@@ -0,0 +1,99 @@
...
+  updateDashboard: DashboardCallbackExecutor,
+  mutator: DashboardMutatorFn<T>
+): (...params: T) => void {
+  return (...x) => {
+    const callback = (dash: MetricsExplorerEntity) => mutator(dash, ...x);
+    updateDashboard(callback);
+  };
+}
+
+type DashboardMutatorFns = {
```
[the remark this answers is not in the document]
bcolloran: Ah, good catch, thanks! Fixed in latest commit

PR #3175, opened 2023-10-05
Free form ISO duration support for default_time_range

### LINKED ISSUE #3156: `default_time_range` isn't working
After adding this to my dashboard YAML...
```yaml
default_time_range: "P4W"
```
... I'd expect to open the dashboard and see the Last 4 Weeks. Instead I see the "All time" range.

AdityaHegde:
## Checklist
- [x] Manual verification
- [x] Unit test coverage
- [x] E2E test coverage
- [x] Needs manual QA?

#### Issue addressed: 
closes #3156

#### Details:
1. Adds support for any valid ISO duration for `default_time_range`.
2. Adds a new item to time controls when the default time range is not one of the allowed ranges. (needs design pass)

## Steps to Verify
1. Create a dashboard with a timestamp.
2. Change the `default_time_range` to something not supported like `P2W`.
4. Go to dashboard, the time range should be `Last 2 Weeks`.
5. Change the `default_time_range` to something that is supported like `P4W`.
6. Go to dashboard, time range should be `Last 4 Weeks`.


web-common/package.json
```diff
@@ -40,6 +40,7 @@
     "axios": "^0.27.2",
     "codemirror": "^6.3.2",
     "date-fns": "^2.30.0",
+    "duration-fns": "^3.0.2",
```
djbarnwal, line 43: I am a bit apprehensive on introducing a new library for handling time apart from luxon

web-common/src/lib/time/ranges/iso-ranges.ts
```diff
@@ -0,0 +1,189 @@
...
+
+  const { startTime, endTime } = isoDurationToTimeRange(isoDuration, end, zone);
+  return {
+    name: TimeRangePreset.DEFAULT,
+    start: startTime,
+    end: endTime,
+  };
+}
+
+export function humaniseISODuration(isoDuration: string): string {
```
djbarnwal, line 77: I suppose we can directly utilize  Luxon methods to convert an ISO string to human readable duration? https://moment.github.io/luxon/api-docs/index.html#durationtohuman

djbarnwal (changes requested):
Briefly went over the PR and I saw that we are introducing a new package for durations. I think we should be able to use luxon's duration object for getting the same results.

web-common/package.json
AdityaHegde: Good point. I will replace this to use luxon.

web-common/src/lib/time/ranges/iso-ranges.ts
```diff
@@ -0,0 +1,170 @@
...
+    }
+  }
+
+  return undefined;
+}
+
+function subtractFromPeriod(duration: Duration, period: Period) {
+  if (!PeriodToUnitsMap[period]) return duration;
+  return duration.minus({ [PeriodToUnitsMap[period]]: 1 });
+}
```
djbarnwal, line 170: Since this is a transformation function independent of ISOs, I would move this to `transforms/index`

web-common/src/lib/time/ranges/iso-ranges.ts
```diff
@@ -0,0 +1,170 @@
...
+      period,
+      truncationType: TimeTruncationType.START_OF_PERIOD,
+    },
+  ];
+}
+
+const PeriodAndUnits: Array<{
+  period: Period;
+  unit: keyof Duration;
+}> = [
```
djbarnwal, line 128: `config.ts` or `types.ts` seems like a more appropriate place for this

web-common/src/features/dashboards/time-controls/TimeRangeSelector.svelte
```diff
@@ -238,6 +252,18 @@
         {allTime.label}
       </span>
     </MenuItem>
+    {#if showDefaultItem && $timeControlsStore.defaultTimeRange}
+      <DefaultTimeRangeMenuItem
+        on:before-select={setIntermediateSelection(TimeRangePreset.DEFAULT)}
```
djbarnwal, line 257: The value of intermediate selection should be the humanize value of the ISO duration. If I have the default duration selected in the menu and after that if I try to change to a different preset, I see a value of `Default` being shown in the menu. In some slower systems and networks, this could be an issue.

web-common/src/features/dashboards/time-controls/DefaultTimeRangeMenuItem.svelte
```diff
@@ -0,0 +1,15 @@
+<script lang="ts">
+  import { humaniseISODuration } from "@rilldata/web-common/lib/time/ranges/iso-ranges";
+  import MenuItem from "../../../components/menu/core/MenuItem.svelte";
+
+  export let selected = false;
+  export let isoDuration: string;
+</script>
+
+<MenuItem animateSelect={false} on:select>
```
djbarnwal, line 9: If we have the right intermediate value, `animateSelect` being `true` makes sense and is consistent with the other menu items.

djbarnwal (changes requested):
UX reviews -

1. I can't view the latest data point in some cases. Probably need to update `getAdjustedChartTime` ? - 

image

2. Menu has some animation/transition issues, mentioned below.

web-common/src/features/dashboards/time-controls/TimeRangeSelector.svelte
AdityaHegde: Good catch. Dont think it makes sense to set it to the humanised ISO. The issue was in the value shown for the select label.

djbarnwal (approved):
Looks good, works as expected.

PR #3182, opened 2023-10-06
Fixing issues around dashboard state after an edit to the dashboard

AdityaHegde:
## Checklist
- [x] Manual verification
- [x] Unit test coverage
- [x] E2E test coverage
- [x] Needs manual QA?

#### Issue addressed: 
Just after editing a dashboard the experience of using it was inconsistent.
1. The default state didnt match the new config.
2. The selections would not reflect changes unless the url state was cleared. This lead to user thinking the changes didnt apply.
3. Changing time controls and refreshing the page doesnt persist the url state.

#### Details:
1. Deriving the default proto state directly from the meta and the time range summary query to keep it in sync with dashboard config.
2. Clearing the dashboard state after an edit to the config.
3. Removed the forced reset of time range stuff.

## Steps to Verify
Issue 1 and 2
1. Create a dashboard with timestamp.
2. Navigate to the dashboard and change the selected time range and other settings.
4. Set the default_time_range to something that is valid within the dashboard.
5. Navigate to the dashboard and the new time range should be reflected as well as other settings should be reverted.

Issue 3
1. Change time range and other settings in the dashboard.
2. Refresh the page.
3. Time range selections should persist.

### DIFF
--- a/web-common/src/features/dashboards/stores/dashboard-stores-test-data.ts
+++ b/web-common/src/features/dashboards/stores/dashboard-stores-test-data.ts
@@ -1,8 +1,8 @@
-import {
-  MetricsExplorerEntity,
-  metricsExplorerStore,
-} from "@rilldata/web-common/features/dashboards/dashboard-stores";
+import type { DashboardFetchMocks } from "@rilldata/web-common/features/dashboards/dashboard-fetch-mocks";
+import { createStateManagers } from "@rilldata/web-common/features/dashboards/state-managers/state-managers";
+import { metricsExplorerStore } from "@rilldata/web-common/features/dashboards/stores/dashboard-stores";
 import { LeaderboardContextColumn } from "@rilldata/web-common/features/dashboards/leaderboard-context-column";
+import type { MetricsExplorerEntity } from "@rilldata/web-common/features/dashboards/stores/metrics-explorer-entity";
 import { getLocalIANA } from "@rilldata/web-common/lib/time/timezone";
 import {
   getOffset,
@@ -23,6 +23,7 @@ import {
   V1TimeGrain,
 } from "@rilldata/web-common/runtime-client";
 import type { QueryObserverResult } from "@tanstack/query-core";
+import { QueryClient } from "@tanstack/svelte-query";
 import type { CreateQueryResult } from "@tanstack/svelte-query";
 import { get, writable } from "svelte/store";
 import { expect } from "vitest";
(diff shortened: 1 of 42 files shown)


ericpgreen2:
I found a UX issue:
- As I navigate between different dashboards in Rill Developer, without changing any dashboard's client-side state, sometimes I see a `state` parameter in the URL.
- In these cases, again clicking the same dashboard in the navigation sidebar, the `state` parameter gets removed.

web-common/src/features/dashboards/stores/dashboard-stores.ts
```diff
@@ -375,21 +196,14 @@ const metricViewReducers = {
     });
   },
 
-  sync(
-    name: string,
-    metricsView: V1MetricsView,
-    fullTimeRange: V1ColumnTimeRangeResponse | undefined
-  ) {
+  sync(name: string, metricsView: V1MetricsView) {
```
ericpgreen2, line 199: I'm unclear about the separation of concerns for this `sync` function. It syncs measures and dimensions, but not time controls. Why?

Also, is this function still necessary now that you've added [this code](https://github.com/rilldata/rill/pull/3182/files#diff-e65c925c7fdd33020a3635868ebecd92ccc541aacec0a9100ae8b88b16b99aeaR41-R42)?

I might just be getting confused by the generic "sync" function name – maybe a more specific name will clear this up.

web-common/src/features/dashboards/stores/dashboard-stores-test-data.ts
```diff
@@ -322,6 +327,32 @@ export function getOffsetByHour(time: Date) {
   );
 }
 
+export function initStateManagers(
```
ericpgreen2, line 330: IMO this extra layer makes the code harder to follow – adds indirection.

web-common/src/features/dashboards/stores/dashboard-stores.ts
AdityaHegde: For measures and dimensions there is a clear idea on how we can merge them with current state. EG: if a selected leaderboard measure is deleted we should select the next valid one.

There is no defined merge for time controls. We definitely need one. But just replacing everything to default will cause issues when sync is called and there is URL state.

AdityaHegde: Lets reduce the scope of this PR to only deal with default time range issue and not the sync issue.

web-common/src/features/dashboards/stores/dashboard-stores-test-data.ts
AdityaHegde: It is a subset of test initialisation used in 2 places. Once we add tests for everything that uses `StateManagers` this would be used there as well.

AdityaHegde:
That is the expected behaviour. The same happens on the current stable release.

ericpgreen2 (approved):
Approving so you can merge when you wish. Discussed the `state` issue over Zoom.

PR #3185, opened 2023-10-06
enable debugging logs, errors and metrics

k-anshul:


runtime/drivers/duckdb/duckdb.go
```diff
@@ -135,6 +138,11 @@ func (d Driver) Open(cfgMap map[string]any, shared bool, ac activity.Client, log
 		ctx:            ctx,
 		cancel:         cancel,
 	}
+	attrs := []attribute.KeyValue{attribute.String("db", c.config.DBFilePath)}
+	c.metrics = observability.Must(meter.RegisterCallback(func(ctx context.Context, observer metric.Observer) error {
+		observer.ObserveInt64(runningOLAPQueriesGauge, int64(c.dbConnCount), metric.WithAttributes(attrs...))
+		return nil
+	}, runningOLAPQueriesGauge))
```
begelundmuller, line 145: 1. `dbConnCount` also covers meta queries, not just OLAP queries. So instead of `running_olap_queries`, this should probably be `connections_in_use` or something like that.
2. nit: add a newline before and a little comment :)

begelundmuller (changes requested):

begelundmuller (approved):

PR #3207, opened 2023-10-10
Stop the watch requests when the page loses focus

AdityaHegde:
When too many tabs are open the watch request get blocked because of the 6 connection limit per domain. We need to make sure that pages out of focus stop the watch requests.

### DIFF
--- a/web-common/src/lib/throttler.ts
+++ b/web-common/src/lib/throttler.ts
@@ -1,0 +1,27 @@
+export class Throttler {
+  private timer: ReturnType<typeof setTimeout> | undefined;
+  private callback: () => void | Promise<void>;
+
+  public constructor(private readonly timeout: number) {}
+
+  public throttle(callback: () => void | Promise<void>) {
+    this.cancel();
+    this.callback = callback;
+
+    this.timer = setTimeout(() => {
+      this.timer = undefined;
+      this.callback();
+    }, this.timeout);
+  }
+
+  public throttling() {
+    return this.timer !== undefined;
+  }
+
+  public cancel() {
+    if (!this.timer) return;
+
+    clearTimeout(this.timer);
+    this.timer = undefined;
+  }
+}
--- a/web-common/src/lib/throttler.ts
+++ b/web-common/src/lib/throttler.ts
@@ -14,7 +14,7 @@ export class Throttler {
     }, this.timeout);
   }
 
-  public throttling() {
+  public isThrottling() {
     return this.timer !== undefined;
   }
 
--- a/web-common/src/runtime-client/watch-request-client.ts
+++ b/web-common/src/runtime-client/watch-request-client.ts
@@ -1,3 +1,4 @@
+import { Throttler } from "@rilldata/web-common/lib/throttler";
 import { pageInFocus } from "@rilldata/web-common/lib/viewport-utils";
 import { ExponentialBackoffTracker } from "@rilldata/web-common/runtime-client/exponential-backoff-tracker";
 import type {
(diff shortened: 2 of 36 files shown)


ericpgreen2:
To test this, I opened up 7 tabs with Rill Developer open, and I refreshed a source. The refresh finished in one tab, but not in the others. In the other tabs, on the source page, I just see the reconcile spinners running continuously. Here's what I see in my browser console: 
image

web-common/src/lib/throttler.ts
```diff
@@ -0,0 +1,27 @@
...
+    this.cancel();
+    this.callback = callback;
+
+    this.timer = setTimeout(() => {
+      this.timer = undefined;
+      this.callback();
+    }, this.timeout);
+  }
+
+  public throttling() {
```
ericpgreen2, line 17: `isThrottling` might be clearer

web-common/src/runtime-client/watch-request-client.ts
```diff
@@ -33,17 +37,38 @@ export class WatchRequestClient<Res extends WatchResponse> {
...
-        !runtimeState ||
-        (runtimeState.instanceId === this.prevInstanceId &&
-          runtimeState.host === this.prevHost)
-      ) {
+    const store = derived([runtime, pageInFocus], (state) => state);
+    const unsubscribe = store.subscribe(([runtimeState, pageInFocus]) => {
+      const runtimeUnchanged =
+        runtimeState.instanceId === this.prevInstanceId &&
+        runtimeState.host === this.prevHost;
+      // const focusUnchanged = pageInFocus === this.prevFocus;
```
ericpgreen2, line 45: old comment?

web-common/src/runtime-client/watch-request-client.ts
```diff
@@ -33,17 +37,38 @@ export class WatchRequestClient<Res extends WatchResponse> {
...
+        this.onReconnect();
+      }
       this.prevInstanceId = runtimeState.instanceId;
       this.prevHost = runtimeState.host;
+      this.prevFocus = true;
 
+      if (this.outOfFocusThrottler.throttling()) {
+        // Cancel any callbacks for out of focus
+        this.outOfFocusThrottler.cancel();
+        // The client is already running. Do not cancel
```
ericpgreen2, line 69: I don't understand this comment given the `cancel()` method above

web-common/src/runtime-client/watch-request-client.ts
```diff
@@ -33,17 +37,38 @@ export class WatchRequestClient<Res extends WatchResponse> {
   ) {}
 
   public start(): Unsubscriber {
```
ericpgreen2, line 39: Feels to me like the `start()` method shouldn't additionally bake in logic to pause/cancel the operation. Could a better structure be to have a `pause()` or `cancel()` method that you can call on the `WatchRequestClient` when the page goes out-of-focus?

web-common/src/runtime-client/watch-request-client.ts
AdityaHegde: The comment is for the watcher client.

web-common/src/runtime-client/watch-request-client.ts
AdityaHegde: It might be better to have everything in one place. The logic of calling all onReconnect should be here as well. Lets spend more time once the release is done.

ericpgreen2:
Leaving our Slack notes here, so we have them:
- On every page focus, it looks like a bunch of things are getting invalidated, most noticeably the source preview and profiling results. I guess invalidating everything every time is the naive approach and ensures correctness. It just doesn’t feel so smooth. We can call this an optimization and take it at a later point.
- Sometimes I see spinners without the "ingesting source" helper text. We think it's because I'm hitting the browser's 6 connection limit before old connections expire.
- Can we clean up the browser console?

ericpgreen2 (approved):

AdityaHegde:
Users will not hit this most of the times. So i think this is not that big a concern.

Will cleanup the console just before release. It helps quite a bit with debugging for now.

PR #3211, opened 2023-10-10
Removing watcher clients from cloud

AdityaHegde:
Adding the watcher client in the cloud would mean that all the users with a dashboard open would have an open connection with the server.

Removing the watcher client and going back to periodic refresh. The list all resources has a periodic refresh and any updated dashboards are invalidated.

### DIFF
--- a/web-admin/src/features/projects/ProjectDashboardsListener.svelte
+++ b/web-admin/src/features/projects/ProjectDashboardsListener.svelte
@@ -1,0 +1,19 @@
+<script lang="ts">
+  import { listenAndInvalidateDashboards } from "@rilldata/web-admin/features/projects/dashboards";
+  import { runtime } from "@rilldata/web-common/runtime-client/runtime-store";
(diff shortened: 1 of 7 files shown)


ericpgreen2:
UXQA finding (though I'm not sure if it entirely relates to this PR): I’m seeing weird behavior on the Home and Org pages. Every project status is the same. It shows the status of whatever project I last visited.

It doesn't seem like this PR accounts for the Home and Org pages. However... one thing I’m working on with Janet is the removal of the dashboard lists from the Home and Org pages in favor of project cards. Would that make this PR simpler? If needed, we can get this Project Card PR in tomorrow.

AdityaHegde:
Good catch. I have updated the PR to make sure the correct message is shown. I have not updated the list to use this method. We should address it in the rewrite that you have planned.

ericpgreen2:
Cool, looks good. Here's the [PR for project cards](https://github.com/rilldata/rill/pull/3224).

ericpgreen2 (approved):

Issue #3198, opened 2023-10-09
Ability to expand columns in dimension leaderboards to see the full text


cohenscottr:
**Is your feature request related to a problem? Please describe.**
Some dimension names contain long/large strings that get cut off on display. When drilling into the leaderboard, users would like the ability to expand the column width to see more of the text of the dimension name.

**Describe the solution you'd like**
Ability to expand dimension tables - specifically in the leaderboard drilldown (vs. Explore)

**Describe alternatives you've considered**
Tooltips are helpful but only one dimension at a time

ericpgreen2:
Closed by #3213

PR #3213, opened 2023-10-10
Resize handle dim table

bcolloran:
add the ability to resize the dimension value column in the dimension table


bcolloran:
This will fix #3198. Ping @ericpgreen2 for review so that we can spin a release today

PR #3218, opened 2023-10-11
Making improvements to the reconcile spinner

AdityaHegde:
Improves the look and feel of the spinners along with fixing a few issues with the spinner.

### DIFF
--- a/web-common/src/components/preview-table/ConnectedPreviewTable.svelte
+++ b/web-common/src/components/preview-table/ConnectedPreviewTable.svelte
@@ -17,7 +17,8 @@
     objectName,
     {}
   );
-  $: profileColumns = $profileColumnsQuery?.data?.profileColumns;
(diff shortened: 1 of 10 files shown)


ericpgreen2:
The reconcile spinner is looking good. I conferred with Janet and made a couple small adjustments: tweaked the spacing; emphasized the font on the resource name (per your good suggestion).

Errors are still not working properly:
- When I edit a valid source to have invalid YAML, then "Save and refresh" the source, the old source rows are displayed, rather than showing me the error.
- When editing a model to have an error, the profiling results freeze in a weird state

I'll approve so you can merge when you want.

ericpgreen2 (approved):

AdityaHegde:
Fixed the issue with invalid yaml. Lets take the modelling issue in a separate PR, have added it to the issues list.

PR #3219, opened 2023-10-11
handle controller init errors

k-anshul:


runtime/controller.go
```diff
@@ -138,6 +141,9 @@ func (c *Controller) Run(ctx context.Context) error {
 	// Check we are still the leader
 	err = c.catalog.checkLeader(ctx)
 	if err != nil {
+		c.initErr = err
+		close(c.closed)
+		c.catalog.release()
```
begelundmuller, line 146: Should use `c.catalog.close()`

begelundmuller (changes requested):

PR #3220, opened 2023-10-11
For anchored time ranges use the min of latest data and now

### LINKED ISSUE #3199: Restrict YTD in dashboard to current year
**Describe the bug**
If a dataset has future data (e.g. forecasting data), Year to Date filter can potentially go beyond current year. 

**To Reproduce**
1. Add dataset with future data
2. Select YTD filter
3. Will include future year vs. current year

**Expected behavior**
Expect current year to date

**Screenshots**
image

djbarnwal:
Closes #3199 
Fixes the case when anchored time ranges such as Today, Month to Date, Year to Date etc. would use the latest data as reference point instead of the current time in cases when latest data is in the future.

More - https://rilldata.slack.com/archives/CTZ8XBQ85/p1696875814646139


ericpgreen2 (approved):
Great!

PR #3228, opened 2023-10-12
add trace to reconcile

k-anshul:


runtime/controller.go
```diff
@@ -1268,6 +1274,18 @@ func (c *Controller) invoke(r *runtimev1.Resource) error {
...
+	tracerAttrs := []attribute.KeyValue{
+		attribute.String("name", n.Name),
+		attribute.String("kind", unqualifiedKind(n.Kind)),
+		attribute.Bool("deleted", inv.isDelete),
+	}
+	if inv.isRename {
+		tracerAttrs = append(tracerAttrs, attribute.String("renamed_from", r.Meta.RenamedFrom.Name))
+	}
+	ctx, span := tracer.Start(ctx, "reconcile", trace.WithAttributes(tracerAttrs...))
+	defer span.End()
```
begelundmuller, line 1288: It needs to go inside the goroutine

runtime/controller.go
```diff
@@ -1268,6 +1274,18 @@ func (c *Controller) invoke(r *runtimev1.Resource) error {
 		c.Logger.Info("Reconciling resource", logArgs...)
 	}
 
+	// Start tracing span
+	tracerAttrs := []attribute.KeyValue{
+		attribute.String("name", n.Name),
+		attribute.String("kind", unqualifiedKind(n.Kind)),
+		attribute.Bool("deleted", inv.isDelete),
```
begelundmuller, line 1282: Only add deleted if it's true

begelundmuller (changes requested):
If you want to couple it to logs, you also need to edit the logging logic. The reconcilers emit logs using `slog` (`r.C.Logger` in the reconcilers), which currently sends them directly to Zap, but with [this PR](https://github.com/rilldata/rill/pull/3170) it will start multi-plexing them both to Zap and to an in-memory buffer that we'll serve over an API (for seeing reconcile logs in the frontend).

runtime/controller.go
k-anshul: Any particular reasons for that ? Won't it just add `deleted:false`

begelundmuller (approved):

PR #3235, opened 2023-10-13
Runtime: Change IngestionLimitBytes to StorageLimitBytes and make it a DuckDB-specific parameter

### LINKED ISSUE #3130: Runtime: Change `IngestionLimitBytes` to `StorageLimitBytes` and make it a DuckDB-specific parameter
Also move the DB size polling into the DuckDB driver instead of having it in the source reconciler

k-anshul:
closes #3130


runtime/drivers/duckdb/olap.go
```diff
@@ -565,6 +584,48 @@ func (c *connection) updateVersion(name, version string) error {
...
+					limitExceeded.Store(true)
+					cancel()
+					return
+				}
+			}
+		}
+	}()
+
+	err := c.Exec(ctx, stmt)
+	cancel()
```
begelundmuller, line 622: nit: move to a `defer cancel()` right after call to `WithCancel`. It's safe to call multiple times and ensures it will always be called even if someone adds other conditions in the future.

runtime/drivers/duckdb/transporter/sqlstore_to_duckDB.go
```diff
@@ -54,8 +54,7 @@ func (s *sqlStoreToDuckDB) Transfer(ctx context.Context, srcProps, sinkProps map
 		defer rowIter.Close()
 		return s.transferFromRowIterator(ctx, rowIter, sinkCfg.Table, opts.Progress)
 	}
-
-	limitInBytes := opts.LimitInBytes
+	limitInBytes, _ := s.to.(drivers.Handle).Config()["storage_limit_bytes"].(int64)
```
begelundmuller, line 57: Consider making this safe:
1. Turn `duckdb.connection.config` into `duckdb.Connection.Config`
2. `sqlStoreToDuckDB.to` can be made a `*duckdb.Connection` instead (since it's always a duckdb handle)
3. Then could do `s.to.Config.StorageLimitBytes`

Even if probably not required here, I think this is anyway a good change because we'll probably have more cases coming where the DuckDB transporters need access to DuckDB config values.

begelundmuller (changes requested):

runtime/drivers/duckdb/transporter/sqlstore_to_duckDB.go
k-anshul: Sure. We will need `duckdb.Connection.Configs` though since there is always a function `Config`

k-anshul: Okay this is not possible since duckDB pkg imports transporter pkg.

begelundmuller (approved):

PR #3239, opened 2023-10-14
Cloud UI: show dashboards in a table

ericpgreen2:
This PR switches the display of a project's dashboards from a list to a table. This is a virtualized table that can be searched and, in the future, filtered and sorted. Additionally, in the display, we now include the time when a dashboard was last refreshed.

Note: the table is currently left-aligned. The PR to add project tabs (https://github.com/rilldata/rill/pull/3240) will center the table.

### Summary by CodeRabbit

- New Feature: Added a comprehensive table view for dashboards, enhancing data organization and accessibility.
- New Feature: Implemented search functionality in the dashboard table, allowing users to quickly find specific dashboards.
- New Feature: Introduced a 'time ago' function, providing users with a human-readable timestamp for dashboard updates.
- Improvement: Enhanced error handling with specific messages when dashboards fail to load, improving user experience during troubleshooting.
- Improvement: Updated the dashboard page layout with better padding and spacing, enhancing the visual appearance.
- Improvement: Added a message and link to documentation when no dashboards are found, guiding users on how to deploy a dashboard.
- Style: Updated various components with improved styling for better user interface consistency.


web-admin/src/features/dashboards/listing/utils.ts
```diff
@@ -0,0 +1,35 @@
+export function timeAgo(date: Date): string {
+  const now: number = new Date().getTime();
+  const diff: number = now - date.getTime();
+  const minute: number = 60 * 1000;
+  const hour: number = 60 * minute;
+  const day: number = 24 * hour;
+  const week: number = 7 * day;
```
djbarnwal, line 7: Nit: Why don't we use our existing methods and framework around time for this? (Luxon and Durations)

web-common/src/components/search/Search.svelte
```diff
@@ -25,17 +25,15 @@
...
     <Search />
   </div>
   <input
     bind:this={ref}
     type="text"
     autocomplete="off"
-    class="outline-none bg-gray-100 surface-impression border border-gray-200 dark:border-gray-400
-        rounded-sm focus:border-gray-300
-        block w-full pl-8 p-1"
+    class="bg-white border border-gray-200 focus:border-blue-400 outline-none rounded-sm block w-full pl-8 p-1"
```
djbarnwal, line 36: These changes are affecting search components at other places

djbarnwal (changes requested):
The search bar for filter pills, comparison selector, visible measure selector etc. looks misaligned and has an outline - 

image

image

web-common/src/components/search/Search.svelte
ericpgreen2: Ah, good catch, just fixed the padding in 3 other Search instances. Janet created this [new Search component](https://www.figma.com/file/nqqazRo1ckU9ooC9ym9weI/Rill-Design-System?type=design&node-id=78-203&mode=design&t=sGyyC9hgxQtE13gy-0) that this code moves towards.

web-admin/src/features/dashboards/listing/utils.ts
ericpgreen2: Good idea to use `luxon` – just switched. But is this code redundant of something already in our codebase? I looked a little bit but didn't see a method that could be dropped in here.

web-common/src/components/search/Search.svelte
ericpgreen2: Note that the blue outline is from the focus state, and these search instances are currently set to focus-on-mount.

djbarnwal (approved):
Looks good.

One thing which still bugs me is the blue border on the search bar. It's too in the eye if I haven't clicked on the search bar itself. Maybe we can have a prop for if we want to enable the blue border or not.
image

Anyway that's a separate issue. I would leave it up to you. Approving so that you can merge at your convenience.

ericpgreen2:
> One thing which still bugs me is the blue border on the search bar. It's too in the eye if I haven't clicked on the search bar itself. Maybe we can have a prop for if we want to enable the blue border or not.

Totally agree. We do have that prop in `autofocus`. Just turned off `autofocus` and aesthetics look better.

PR #3252, opened 2023-10-17
Using tanstack query for all lists in cloud

AdityaHegde:
## Checklist
- [x] Manual verification
- [ ] Unit test coverage
- [ ] E2E test coverage
- [x] Needs manual QA?

#### Issue addressed: 
We were using direct calls to get dashboard list in project page.

#### Details:
Replacing the direct call to instead use tanstack query. This way there shouldnt be any issue with stale jwt.

### DIFF
--- a/web-admin/src/features/projects/ProjectDeploymentStatusChip.svelte
+++ b/web-admin/src/features/projects/ProjectDeploymentStatusChip.svelte
@@ -17,7 +17,7 @@
   import Spinner from "@rilldata/web-common/features/entity-management/Spinner.svelte";
   import { EntityStatus } from "@rilldata/web-common/features/entity-management/types";
   import { getRuntimeServiceListResourcesQueryKey } from "@rilldata/web-common/runtime-client";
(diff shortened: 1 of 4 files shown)


web-admin/src/features/projects/ProjectDeploymentStatusChip.svelte
```diff
@@ -36,10 +36,7 @@
 
   let deploymentStatusFromDashboards: CreateQueryResult<V1DeploymentStatus>;
   $: if ($proj?.data)
```
ericpgreen2, line 38: Do we still need this conditional?

ericpgreen2 (approved):
Nice, I couldn't replicate the 401 anymore.

web-admin/src/features/projects/ProjectDeploymentStatusChip.svelte
AdityaHegde: Nope, removed.

PR #3254, opened 2023-10-17
configurable threads limit per cpu

k-anshul:


runtime/drivers/duckdb/config.go
```diff
@@ -64,12 +70,26 @@ func newConfig(cfgMap map[string]any) (*config, error) {
...
+		threads := int(cpuThreadRatio * float64(cfg.CPU))
+		if threads <= 0 {
+			threads = 1
+		}
+		qry.Add("threads", strconv.Itoa(threads))
+		if cfg.CPU <= threads {
+			cfg.PoolSize = cfg.CPU
+		} else {
+			cfg.PoolSize = threads
+		}
```
begelundmuller, line 87: Can we also make a change so the pool size is always at least 2? Having it at 1 is not great with async reconciles. Would need to change line 40, 94, and then here it should be `max(2, min(cfg.CPU, threads))`.

begelundmuller (approved):

PR #3263, opened 2023-10-18
Fixing histogram query on duckdb 0.9

AdityaHegde:
Histogram queries throw an error when run on duckdb 0.9.
`Error: INTERNAL Error: Failed to bind column reference "SUBQUERY" [476.0] (bindings: [401.0])`

Found a workaround where calculating min, max and range outside of the CTE fixes this.


begelundmuller (commented):
Looks good, two questions:
1. Did you test on a few different datasets? Maybe especially an empty one and one with a column with only nulls
2. Why does it need to uncomment frontend code?

AdityaHegde:
I am in the process of adding more tests to make sure any edge cases are not missed. Uncommented code was for testing will revert.

AdityaHegde:
Seems like there are tests around nulls already. We should be good there.

begelundmuller (approved):

PR #3266, opened 2023-10-18
Controller: use a separate lock for resource event subscribers

begelundmuller:


begelundmuller:
This could cause a resource to be mutated while being serialized over the network. Not worth it for now.

PR #3271, opened 2023-10-19
Run controller initialization in the background

begelundmuller:
Previously, the registry cache would load all existing instances from the metastore on startup, start their controllers, call `WaitUntilReady`, then create the global project parser if it didn't already exist.

Since controller initialization needs to load the catalog, and we embed the catalog in DuckDB, if opening any DuckDB handle is slow (e.g. has a large wal), this would could cause the entire runtime initialization to hang. This means the runtime server isn't started, so health checks fail, and we got a crash loop in Kubernetes.

This PR moves controller initialization into the background. Instead, any call to `runtime.Controller(ctx, instanceID)` will hang until the controller has finished initializing ready or has errored.

### DIFF
--- a/runtime/connection_cache.go
+++ b/runtime/connection_cache.go
@@ -19,7 +19,7 @@ import (
 
 var errConnectionCacheClosed = errors.New("connectionCache: closed")
 
-const migrateTimeout = 30 * time.Second
+const migrateTimeout = 2 * time.Minute
 
 // connectionCache is a thread-safe cache for open connections.
 // Connections should preferably be opened only via the connection cache.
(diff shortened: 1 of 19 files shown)


himadrisingh (approved):

runtime/connection_cache.go
```diff
@@ -257,13 +257,19 @@ func (c *connectionCache) openAndMigrate(ctx context.Context, instanceID, driver
 	}
 
 	handle, err := drivers.Open(driver, config, shared, activityClient, logger)
+	if err == nil && ctx.Err() != nil {
```
k-anshul, line 260: This doesn't handle timeout fully. Is that expected ?

begelundmuller: Yeah, it's not ideal, but `drivers.Open` doesn't currently accept a `ctx`, and anyway, DuckDB (which is the main issue for slow opens) doesn't support cancellation during opening as far as I know. So not much more we can do :(

PR #3279, opened 2023-10-19
Bigquery: Allow extracting results more than 10GB in size

esevastyanov:
## Summary by CodeRabbit

New Features:
- Enhanced the `QueryAsFiles` function to support BigQuery SELECT ALL statements, improving data retrieval capabilities.
- Introduced a new helper function `createClient` for creating a BigQuery client, simplifying the client creation process.

Refactor:
- Updated the `QueryAsFiles` function to use the storage API for reading table data when handling SELECT ALL statements, optimizing data access.

Style:
- Improved error handling and logging in the `QueryAsFiles` function, enhancing debugging and issue tracking.

### DIFF
--- a/runtime/drivers/bigquery/bigquery.go
+++ b/runtime/drivers/bigquery/bigquery.go
@@ -38,7 +38,7 @@ var spec = drivers.Spec{
 		{
 			Key:         "project_id",
 			Type:        drivers.StringPropertyType,
-			Required:    false,
+			Required:    true,
 			DisplayName: "Project ID",
 			Description: "Google project ID.",
 			Placeholder: "my-project",
--- a/runtime/drivers/bigquery/bigquery.go
+++ b/runtime/drivers/bigquery/bigquery.go
@@ -38,7 +38,7 @@ var spec = drivers.Spec{
 		{
 			Key:         "project_id",
 			Type:        drivers.StringPropertyType,
-			Required:    true,
+			Required:    false,
 			DisplayName: "Project ID",
 			Description: "Google project ID.",
 			Placeholder: "my-project",
--- a/runtime/drivers/bigquery/sql_store.go
+++ b/runtime/drivers/bigquery/sql_store.go
@@ -20,6 +20,7 @@ import (
 	"github.com/rilldata/rill/runtime/pkg/observability"
 	"go.uber.org/zap"
 	"google.golang.org/api/iterator"
+	"google.golang.org/api/option"
 )
 
 // recommended size is 512MB - 1GB, entire data is buffered in memory before its written to disk
@@ -44,21 +45,8 @@ func (c *Connection) QueryAsFiles(ctx context.Context, props map[string]any, opt
 		return nil, err
 	}
 
-	now := time.Now()
-
 	var client *bigquery.Client
 	var it *bigquery.RowIterator
-	bqClientFn := func(projectID string) (*bigquery.Client, error) {
-		client, err := bigquery.NewClient(ctx, projectID, opts...)
-		if err != nil {
-			if strings.Contains(err.Error(), "unable to detect projectID") {
-				return nil, fmt.Errorf("projectID not detected in credentials. Please set `project_id` in source yaml")
-			}
-			return nil, fmt.Errorf("failed to create bigquery client: %w", err)
-		}
-		return client, nil
-	}
-
 	r := regexp.MustCompile("(?i)^\\s*SELECT\\s+\\*\\s+FROM\\s+(`?[a-zA-Z0-9_.-]+`?)\\s*$")
 	match := r.FindStringSubmatch(srcProps.SQL)
 	if match != nil {
@@ -73,20 +61,13 @@ func (c *Connection) QueryAsFiles(ctx context.Context, props map[string]any, opt
 		switch len(parts) {
 		case 1:
 			tableID = parts[0]
-		case 2:
-			dataset, tableID = parts[0], parts[1]
 		case 3:
 			projectID, dataset, tableID = parts[0], parts[1], parts[2]
 		default:
 			return nil, fmt.Errorf("invalid table format, `project_id.dataset.table` is expected")
 		}
 
-		// Override projectID if a user-defined is set
-		if srcProps.ProjectID != "" && srcProps.ProjectID != bigquery.DetectProjectID {
-			projectID = srcProps.ProjectID
-		}
-
-		client, err = bqClientFn(projectID)
+		client, err = createClient(ctx, srcProps.ProjectID, opts)
 		if err != nil {
 			return nil, err
 		}
@@ -96,9 +77,11 @@ func (c *Connection) QueryAsFiles(ctx context.Context, props map[string]any, opt
 			return nil, err
 		}
 
-		it = client.Dataset(dataset).Table(tableID).Read(ctx)
+		it = client.DatasetInProject(projectID, dataset).Table(tableID).Read(ctx)
 	} else {
-		client, err = bqClientFn(srcProps.ProjectID)
+		now := time.Now()
+
+		client, err = createClient(ctx, srcProps.ProjectID, opts)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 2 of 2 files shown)


runtime/drivers/bigquery/sql_store.go
```diff
@@ -43,41 +44,93 @@ func (c *Connection) QueryAsFiles(ctx context.Context, props map[string]any, opt
...
-	if err != nil && !strings.Contains(err.Error(), "Syntax error") {
-		// close the read storage API client
-		client.Close()
-		c.logger.Info("query failed, retrying without storage api", zap.Error(err))
-		// the query results are always cached in a temporary table that storage api can use
-		// there are some exceptions when results aren't cached
-		// so we also try without storage api
-		client, err = bigquery.NewClient(ctx, srcProps.ProjectID, opts...)
+		// Override projectID if a user-defined is set
+		if srcProps.ProjectID != "" && srcProps.ProjectID != bigquery.DetectProjectID {
```
k-anshul, line 85: I think we can't rely on the `project_id` passed in query for all cases. User may be trying to read data from another project where they don't have access to compute resources.
The project should always be either detected from credentials or passed explicitly in `project_id` field.

runtime/drivers/bigquery/sql_store.go
```diff
@@ -43,41 +44,93 @@ func (c *Connection) QueryAsFiles(ctx context.Context, props map[string]any, opt
...
-			return nil, fmt.Errorf("failed to create bigquery client: %w", err)
+			return nil, err
+		}
+
+		if err = client.EnableStorageReadClient(ctx, opts...); err != nil {
+			client.Close()
+			return nil, err
+		}
+
+		it = client.Dataset(dataset).Table(tableID).Read(ctx)
```
k-anshul, line 99: This should be `DatasetInProject` instead of `Dataset` since user may be trying to read data from another project.

runtime/drivers/bigquery/sql_store.go
```diff
@@ -43,41 +44,93 @@ func (c *Connection) QueryAsFiles(ctx context.Context, props map[string]any, opt
...
+		}
+
+		if err := client.EnableStorageReadClient(ctx, opts...); err != nil {
+			client.Close()
+			return nil, err
 		}
 
 		q := client.Query(srcProps.SQL)
 		it, err = q.Read(ctx)
+		if err != nil && !strings.Contains(err.Error(), "Syntax error") {
```
k-anshul, line 113: We should also look for the result too big [error](https://cloud.google.com/knowledge/kb/bigquery-response-too-large-to-return-consider-setting-allowlargeresults-to-true-in-your-job-configuration-000004266) and provide hint to the user to use `select * from project.dataset.table` or set `dataset` and `table` whichever approach we finalise.

runtime/drivers/bigquery/sql_store.go
```diff
@@ -43,41 +44,93 @@ func (c *Connection) QueryAsFiles(ctx context.Context, props map[string]any, opt
...
 
-	client, err := bigquery.NewClient(ctx, srcProps.ProjectID, opts...)
-	if err != nil {
-		if strings.Contains(err.Error(), "unable to detect projectID") {
-			return nil, fmt.Errorf("projectID not detected in credentials. Please set `project_id` in source yaml")
+	now := time.Now()
+
+	var client *bigquery.Client
+	var it *bigquery.RowIterator
+	bqClientFn := func(projectID string) (*bigquery.Client, error) {
```
k-anshul, line 51: This can be extracted to a function `NewClient`/`CreateClient`/`GetClient` ?

runtime/drivers/bigquery/sql_store.go
```diff
@@ -43,41 +44,93 @@ func (c *Connection) QueryAsFiles(ctx context.Context, props map[string]any, opt
...
+			client.Close()
+			return nil, err
+		}
 	}
-	if err != nil {
-		client.Close()
-		return nil, err
-	}
+
 	c.logger.Info("query took", zap.Duration("duration", time.Since(now)), observability.ZapCtx(ctx))
```
k-anshul, line 134: This log is only applicable for `else` case. May be move it to else and add corresponding log to if case as well.

k-anshul (changes requested):

runtime/drivers/bigquery/bigquery.go
```diff
@@ -38,7 +38,7 @@ var spec = drivers.Spec{
 		{
 			Key:         "project_id",
 			Type:        drivers.StringPropertyType,
-			Required:    true,
+			Required:    false,
```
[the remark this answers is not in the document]
esevastyanov: This just makes the form easier for a user to fill if the project id from a query is a billing project id.
But if there is an SDK issue then the change might be reverted so that client project is mandatory for all cases

runtime/drivers/bigquery/sql_store.go
```diff
@@ -43,41 +44,93 @@ func (c *Connection) QueryAsFiles(ctx context.Context, props map[string]any, opt
...
 		}
-		return nil, fmt.Errorf("failed to create bigquery client: %w", err)
+		return client, nil
 	}
 
-	if err := client.EnableStorageReadClient(ctx, opts...); err != nil {
-		client.Close()
-		return nil, err
-	}
+	r := regexp.MustCompile("(?i)^\\s*SELECT\\s+\\*\\s+FROM\\s+(`?[a-zA-Z0-9_.-]+`?)\\s*$")
```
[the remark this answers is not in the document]
esevastyanov: This is an extension of the current approach. If there were two forms for loading data from BQ, we would be anyway needed to somehow direct the user to choose the right one for their needs.

> This regex variable can be global ?

It might be unified to be global but as of now, backticks are specific for BQ

runtime/drivers/bigquery/sql_store.go
```diff
@@ -43,41 +44,93 @@ func (c *Connection) QueryAsFiles(ctx context.Context, props map[string]any, opt
...
+		// project_id, dataset, backticks are optional
+		fullTableName := match[1]
+		fullTableName = strings.Trim(fullTableName, "`")
+
+		var projectID, dataset, tableID string
+
+		parts := strings.Split(fullTableName, ".")
+		switch len(parts) {
+		case 1:
+			tableID = parts[0]
```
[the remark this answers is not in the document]
esevastyanov: Technically, this isn't possible with the golang driver. But it is a common practice while using "bq" command.
Excluded this option.

runtime/drivers/bigquery/sql_store.go
esevastyanov: Moved the log to `else` and left `if` with no timing as the iterator is instantiated with no `compute` there

runtime/drivers/bigquery/bigquery.go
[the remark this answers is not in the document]
k-anshul: Yeah initially we kept is as optional but some users complained that they find it weird that its marked as optional but later complains that we need to add `project_id` so we marked this as mandatory.

runtime/drivers/bigquery/sql_store.go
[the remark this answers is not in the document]
k-anshul: > It might be unified to be global but as of now, backticks are specific for BQ

Sorry I meant global within this pkg. Since this is a fixed regex and regex compilation can be computationally expensive we can avoid doing this for every query.

runtime/drivers/bigquery/sql_store.go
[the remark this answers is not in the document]
k-anshul: Sorry I didn't understand. You removed the case when user has given `select * from dataset.table` which seems acceptable(we can rely on detected project id) but kept `select * from table` which doesn't give us dataset.

runtime/drivers/bigquery/sql_store.go
k-anshul: It should be pretty common to read data from public datasets.

runtime/drivers/bigquery/sql_store.go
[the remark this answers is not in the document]
esevastyanov: Sorry, my bad, removed the wrong case. Dataset is mandatory now.

runtime/drivers/bigquery/sql_store.go
esevastyanov: Yes, these don't depend on a query and take almost identical time on every run

runtime/drivers/bigquery/sql_store.go
esevastyanov: Updated the logic so that If source project id isn't passed via a query then a corresponding source property is used

runtime/drivers/bigquery/sql_store.go
```diff
@@ -43,42 +48,85 @@ func (c *Connection) QueryAsFiles(ctx context.Context, props map[string]any, opt
...
-	if err != nil {
-		if strings.Contains(err.Error(), "unable to detect projectID") {
-			return nil, fmt.Errorf("projectID not detected in credentials. Please set `project_id` in source yaml")
+	var client *bigquery.Client
+	var it *bigquery.RowIterator
+
+	match := selectQueryRegex.FindStringSubmatch(srcProps.SQL)
+	if match != nil {
+		// "SELECT * FROM `project_id.dataset.table`" statement so storage api might be used
+		// project_id, dataset, backticks are optional
```
k-anshul, line 57: dataset is not optional

k-anshul (approved):

PR #3287, opened 2023-10-20
duckdb sources have anonymous access

k-anshul:


begelundmuller:
The `parser.AnalyzeConnectors` function will ignore the project's "default connector", which is currently hard-coded to DuckDB, but that wasn't configured for the `rill env configure` command.

begelundmuller (approved):

PR #3295, opened 2023-10-23
better error msg in dev script

k-anshul:


scripts/devtool/main.go
```diff
@@ -91,11 +92,11 @@ func main() {
 	yellow.Println("STARTING DOCKER SERVICES")
 	err = pgCmd.Start()
 	if err != nil {
-		panic("could not start db")
+		panic(fmt.Sprintf("could not start db %s", err.Error()))
 	}
 	group.Go(func() error {
 		err := pgCmd.Wait()
-		red.Println("STOPPING DOCKER SERVICES")
+		red.Printf("STOPPING DOCKER SERVICES %s\n", err.Error())
```
begelundmuller, line 99: What if `err == nil`? And could it look weird during a normal shutdown (assuming that returns context cancelled?)

scripts/devtool/main.go
```diff
@@ -110,11 +111,11 @@ func main() {
 		yellow.Println("STARTING ADMIN")
 		err = admin.Start()
 		if err != nil {
-			panic("unable to start admin server")
+			panic(fmt.Sprintf("unable to start admin server: %s", err.Error()))
 		}
 		group.Go(func() error {
 			err := admin.Wait()
-			red.Println("ADMIN SERVER STOPPED")
+			red.Printf("ADMIN SERVER STOPPED %s\n", err.Error())
```
begelundmuller, line 118: Same

scripts/devtool/main.go
```diff
@@ -129,11 +130,11 @@ func main() {
 		yellow.Println("STARTING RUNTIME")
 		err = rt.Start()
 		if err != nil {
-			panic("unable to start runtime server")
+			panic(fmt.Sprintf("unable to start runtime server: %s", err.Error()))
 		}
 		group.Go(func() error {
 			err := rt.Wait()
-			red.Println("RUNTIME SERVER STOPPED")
+			red.Printf("RUNTIME SERVER STOPPED %s\n", err.Error())
```
begelundmuller, line 137: Same

PR #3306, opened 2023-10-24
Retry DuckDB extension installation with different mirror

begelundmuller:
See this Slack [thread](https://rilldata.slack.com/archives/CTZ8XBQ85/p1698092603383519) for context

### DIFF
--- a/runtime/drivers/duckdb/duckdb.go
+++ b/runtime/drivers/duckdb/duckdb.go
@@ -427,6 +427,10 @@ func (c *connection) reopenDB() error {
 	connector, err := duckdb.NewConnector(c.config.DSN, func(execer driver.ExecerContext) error {
 		for _, qry := range bootQueries {
 			_, err := execer.ExecContext(context.Background(), qry, nil)
+			if err != nil && strings.Contains(err.Error(), "Failed to download extension") {
+				// Retry using another mirror. Based on: https://github.com/duckdb/duckdb/issues/9378
+				_, err = execer.ExecContext(context.Background(), qry+" FROM 'http://nightly-extensions.duckdb.org'", nil)
+			}
 			if err != nil {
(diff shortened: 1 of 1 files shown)


runtime/drivers/duckdb/duckdb.go
```diff
@@ -427,6 +427,10 @@ func (c *connection) reopenDB() error {
 	connector, err := duckdb.NewConnector(c.config.DSN, func(execer driver.ExecerContext) error {
 		for _, qry := range bootQueries {
 			_, err := execer.ExecContext(context.Background(), qry, nil)
+			if err != nil && strings.Contains(err.Error(), "Failed to download extension") {
+				// Retry using another mirror. Based on: https://github.com/duckdb/duckdb/issues/9378
+				_, err = execer.ExecContext(context.Background(), qry+" FROM 'http://nightly-extensions.duckdb.org'", nil)
```
k-anshul, line 432: Is there a possibility that nightly can be incompatible with release version sometimes ?
We can also consider @himadrisingh's suggestion to also keep our CDN as backup.

begelundmuller: They suggested this fix themselves in the issue linked through above. Anyway, I think it should be good, it incorporates the version number in the generated URL, even when using another mirror ([source](https://github.com/duckdb/duckdb/blob/66be885660aab4d4bde16874ee6be1ed447bc19d/src/main/extension/extension_install.cpp#L182)).

k-anshul: Yeah I looked into the issue. Was wondering if they suggesting nightly URL was a temporary workaround only and may not work in future. 
But if there is a version number in URL then we should be good.

k-anshul (approved):

PR #3315, opened 2023-10-24
enable strictNullChecks in web-common/src/components, and fix all errors

bcolloran:
incremental adoption of `strictNullChecks`


web-common/src/tsconfig.json
ericpgreen2, line 1: This file doesn't look necessary?

ericpgreen2 (approved):
Nice, I like the approach to adopt this incrementally 👍

PR #3319, opened 2023-10-25
use sqlite as catalog for local

### LINKED ISSUE #3297: Runtime: Investigate using sqlite instead of DuckDB for the catalog in cloud

k-anshul:
closes #3297,#3341


cli/pkg/local/app.go
```diff
@@ -47,6 +47,7 @@ const (
 	DefaultInstanceID = "default"
 	DefaultOLAPDriver = "duckdb"
 	DefaultOLAPDSN    = "stage.db"
+	DefaultDBDir      = "rill_db"
```
begelundmuller, line 50: Discussed with Nishant to call it `tmp`. What do you think?

cli/pkg/local/app.go
```diff
@@ -47,6 +47,7 @@ const (
 	DefaultInstanceID = "default"
 	DefaultOLAPDriver = "duckdb"
 	DefaultOLAPDSN    = "stage.db"
```
begelundmuller, line 49: Now that we're anyway moving it, can we change it to `main.db`? That would also fit with the default DuckDB schema name

cli/pkg/local/app.go
```diff
@@ -82,12 +83,28 @@ func NewApp(ctx context.Context, ver config.Version, verbose, strict, reset bool
...
+		return nil, err
+	}
+
 	// Create a local runtime with an in-memory metastore
 	systemConnectors := []*runtimev1.Connector{
 		{
 			Type:   "sqlite",
 			Name:   "metastore",
-			Config: map[string]string{"dsn": "file:rill?mode=memory&cache=shared"},
+			Config: map[string]string{"dsn": fmt.Sprintf("file:%s?cache=shared", filepath.Join(dbDirPath, "metastore.db"))},
```
begelundmuller, line 107: Can we just call it `meta.db`? Also, let's pull the name up into const as well

begelundmuller (changes requested):
Also need to update the `--reset` flag to drop the whole directory

runtime/runtime.go
```diff
@@ -35,6 +36,7 @@ type Runtime struct {
 	logger         *zap.Logger
 	activity       activity.Client
 	metastore      drivers.Handle
+	catalog        drivers.Handle
```
begelundmuller, line 39: It feels like this should be at the instance-level instead, since the catalog is instance-specific, i.e. passing a `CatalogConnector` to `CreateInstance`

cli/pkg/local/app.go
```diff
@@ -82,18 +84,58 @@ func NewApp(ctx context.Context, ver config.Version, verbose, strict, reset bool
...
 		{
 			Type:   "sqlite",
 			Name:   "metastore",
 			Config: map[string]string{"dsn": "file:rill?mode=memory&cache=shared"},
 		},
+		{
+			Type:   "sqlite",
+			Name:   "catalog",
+			Config: map[string]string{"dsn": fmt.Sprintf("file:%s?cache=shared", filepath.Join(dbDirPath, DefaultMetaStore))},
+		},
```
begelundmuller, line 132: And passing this to the instance's connectors instead of `systemConnectors`

cli/pkg/local/app.go
```diff
@@ -82,18 +84,58 @@ func NewApp(ctx context.Context, ver config.Version, verbose, strict, reset bool
...
+		// rename to main.db and move to db directory
+		err := os.Rename(filepath.Join(projectPath, "stage.db"), filepath.Join(dbDirPath, DefaultOLAPDSN))
+		if err != nil {
+			return nil, err
+		}
+
+		err = os.Rename(filepath.Join(projectPath, "stage.db.wal"), filepath.Join(dbDirPath, fmt.Sprintf("%s.wal", DefaultOLAPDSN)))
+		if err != nil {
+			return nil, err
+		}
```
begelundmuller, line 111: There can be a `.db` file without a `.wal` file, then this will error

cli/pkg/local/app.go
```diff
@@ -82,18 +84,58 @@ func NewApp(ctx context.Context, ver config.Version, verbose, strict, reset bool
...
+		// catalog is embedded for older flow
+		embedCatalog = true
```
begelundmuller, line 113: This doesn't really work – on the next restart, `embedCatalog` will be false (since `stage.db` was moved), and it will rebuild everything. Should either not move `stage.db` and use it directly (probably best) or just drop it and rebuild.

runtime/compilers/rillv1/init.go
```diff
@@ -29,7 +29,7 @@ func InitEmpty(ctx context.Context, repo drivers.RepoStore, instanceID, title st
 	if gitignore != "" {
 		gitignore += "\n"
 	}
-	gitignore += ".DS_Store\n\n# Rill\n*.db\n*.db.tmp\n*.db.wal\n.env\n"
+	gitignore += ".DS_Store\n\n# Rill\n*.db\n*.db.tmp\n*.db.wal\n.env\ntmp\n"
```
begelundmuller, line 32: Guess we don't need `*.db` anymore if we put it all in `tmp`?

begelundmuller (changes requested):

runtime/compilers/rillv1/init.go
k-anshul: We will need it for backward compatibility ?

runtime/runtime.go
k-anshul: For cloud `metastore` acts as `catalog` as well (when not using `embedCatalog`) which is not instance specific. If I keep this at instance level then we need to do some changes in `connnection.go` to make sure all instances share the same handle or use instance specific `catalog` for cloud as well (which seems unnecessary).

cli/pkg/local/app.go
k-anshul: This seems to be getting somewhat complicated now. How about we keep it clean and just drop and rebuild which may anyways be required if we merge this in a release with duckdb upgrade ?

runtime/runtime.go
begelundmuller: We can keep the current behavior where it falls back to the metastore if `catalogConnector == ""`. That should avoid any migrations or significant refactors.

Eventually we should probably just set `catalogConnector = "metastore"` on the instances, but yes that requires a refactor enabling them to access system connectors, we can skip on that now.

cli/pkg/local/app.go
begelundmuller: Maybe I'm missing something, seems quite simple to say "if `stage.db` exists, set DSN to it and `embedCatalog = true`? But if I'm missing something or it gets ugly, sure let's just drop and rebuild

runtime/compilers/rillv1/init.go
begelundmuller: Yeah this is only wwhen initializing a new project

cli/pkg/local/app.go
k-anshul: Yeah I mean we will always have to carry these if else conditions. Data will be split between tmp directory (catalog) and stage.db. 
In this case it seems better to just drop and rebuild which we anyways do from time to time.

cli/pkg/local/app.go
```diff
@@ -46,7 +46,9 @@ const (
 const (
 	DefaultInstanceID = "default"
 	DefaultOLAPDriver = "duckdb"
-	DefaultOLAPDSN    = "stage.db"
+	DefaultOLAPDSN    = "main.db"
+	DefaultMetaStore  = "meta.db"
```
begelundmuller, line 50: We should call this var `DefaultCatalogStore` since that's what it is (the metastore will still be in-memory – the name `meta.db` is mainly for user-understanding if they peek in the directory)

cli/pkg/local/app.go
```diff
@@ -82,6 +84,31 @@ func NewApp(ctx context.Context, ver config.Version, verbose, strict, reset bool
...
+		return nil, err
+	}
+
+	// old behaviour when data was stored in a stage.db file in the project directory.
+	// drop old file, remove this code after some time
+	_, err = os.Stat(filepath.Join(projectPath, "stage.db"))
+	if err == nil { // a old stage.db file exists
+		_ = os.Remove(filepath.Join(projectPath, "stage.db"))
+		_ = os.Remove(filepath.Join(projectPath, "stage.db.wal"))
+		logger.Named("console").Info("dropping old stage.db file and rebuilding project")
```
begelundmuller, line 104: Capitalize first word in user-facing logs, as `Dropping ...`

cli/pkg/local/app.go
```diff
@@ -94,6 +121,7 @@ func NewApp(ctx context.Context, ver config.Version, verbose, strict, reset bool
 	rtOpts := &runtime.Options{
 		ConnectionCacheSize:          100,
 		MetastoreConnector:           "metastore",
+		CatalogConnector:             "catalog",
```
begelundmuller, line 124: This is weird – there is no connector named `"catalog"` in `systemConnectors`. Would propose just removing `CatalogConnector` altogether from `runtime.Options`

runtime/connections.go
```diff
@@ -125,11 +125,26 @@ func (r *Runtime) Catalog(ctx context.Context, instanceID string) (drivers.Catal
 		return store, release, nil
 	}
 
-	store, ok := r.metastore.AsCatalogStore(instanceID)
+	if r.opts.CatalogConnector == "" {
```
begelundmuller, line 128: Should be `inst.CatalogConnector` – otherwise, metastore will always take precedence over `inst.CatalogConnector` (instance-level config should always have precedence). Removing `r.opts.CatalogConnector` would avoid some complexity that will otherwise be needed here to handle all cases.

runtime/drivers/sqlite/registry.go
```diff
@@ -242,6 +246,7 @@ func (c *connection) EditInstance(_ context.Context, inst *drivers.Instance) err
 			olap_connector = $2,
 			repo_connector = $3,
 			admin_connector = $4,
+			catalog_connector = $5,
 			updated_on = $5,
```
begelundmuller, line 250: Numbers are off – doesn't this breaks `EditInstance`?

begelundmuller (changes requested):

runtime/drivers/sqlite/registry.go
k-anshul: Weirdly `sqlite` doesn't throw an error when doing `update` but throws error on `select` after doing this operation. Updated the unit tests to directly use select.
I also happen to notice that while doing `EditInstance` we are updating the cache with the passed instance object and not with object fetched from db : https://github.com/rilldata/rill/blob/2f7e89c2bff10e647cfb967feb92fec6a837cf69/runtime/registry.go#L278
This doesn't seem correct to me even though everything in instance right now is updated from the object only but we should select the updated object and then update the cache for following reasons:
1. DB can have defaults which may not be captured in the passed instance object. 
2. For this particular case it is possible that cache and db can end up not being in sync in case on concurrent edits.

runtime/connection_cache_test.go
```diff
@@ -386,10 +391,11 @@ func newTestRuntimeWithInst(t *testing.T) *Runtime {
...
 	inst := &drivers.Instance{
-		ID:            instanceId,
-		OLAPConnector: "duckdb",
-		RepoConnector: "repo",
-		EmbedCatalog:  true,
+		ID:               instanceId,
+		OLAPConnector:    "duckdb",
+		RepoConnector:    "repo",
+		CatalogConnector: "catalog",
+		EmbedCatalog:     true,
```
begelundmuller, line 398: Redundant to set a catalog connector when embedCatalog is true?

begelundmuller (approved):

runtime/drivers/sqlite/registry.go
begelundmuller: Sounds very weird. Yeah, seems like there's a race condition there (though the `EditInstance` DB call updates all fields, so chance is quite low). Ideally I guess `EditInstance` would return the new one, or alternatively, `FindInstance` should be called under the lock. Will you add an issue for it?

k-anshul: Yeah calling a `FindInstance` under the lock should be enough ?

PR #3317, opened 2023-10-24
fix null strictness in dashboard folder

bcolloran:
fixes all strictNullCheck errors in dashboard folder

### Summary by CodeRabbit

- New Feature: Enhanced the dashboard's robustness by improving data handling and error handling. This includes better handling of undefined values and nullability checks, ensuring a smoother user experience.
- Refactor: Improved code clarity and maintainability by making variable types more explicit across various components. This change enhances the software's intuitiveness and reduces potential errors.
- Bug Fix: Fixed potential issues with the handling of dashboard filters and the display of labels and data in tables. This ensures that the dashboard displays accurate and relevant information.
- Style: Removed unused constants and refactored code for better readability and efficiency. This change improves the overall code quality and maintainability.

PR #3327, opened 2023-10-26
Add types to Pivot table

djbarnwal:
## Checklist
- [x] Manual verification
- [ ] Unit test coverage
- [ ] E2E test coverage
- [x] Needs manual QA?

#### Issue addressed: 
Continuation of #3317

#### Details:
Apparently `regular-table` has examples and support for APIs which are not present in their type signatures. To address this have modified the APIs so that they are in line with their types

### Summary by CodeRabbit

- Refactor: Enhanced the Pivot table component in the dashboard feature. This includes improved handling of table styling and data representation, providing a more consistent and reliable user experience.
- Bug Fix: Fixed issues with the `scrollToCell` function, which now accurately navigates to the correct cell in the Pivot table, improving navigation within the table.
- Style: Updated the visual representation of loading and null cells in the Pivot table, making it easier for users to understand the table's state.
- Refactor: Improved the type definitions used in the Pivot table component, increasing the reliability of the code and reducing potential errors.

### DIFF
--- a/web-common/src/features/dashboards/pivot/Pivot.svelte
+++ b/web-common/src/features/dashboards/pivot/Pivot.svelte
@@ -1,19 +1,13 @@
 <script lang="ts">
   import "regular-table";
   import "regular-table/dist/css/material.css";
+  import type { RegularTableElement } from "regular-table";
   import { createEventDispatcher, onMount } from "svelte";
   import type { PivotPos, PivotRenderCallback } from "./types";
   import { isEmptyPos, range } from "./util";
 
-  const LOADING_CELL = {
-    isLoader: true,
-    value: `<div class="h-4 bg-gray-50 rounded loading-cell" style="width: 100%; min-width: 32px;"/>`,
-  };
-
-  const NULL_CELL = {
-    isNull: true,
-    value: `<div style="font-size:0.925em" class="opacity-50 italic">no data</div>`,
-  };
+  const LOADING_CELL = `<div load class="loading-cell h-4 bg-gray-50 rounded" style="width: 100%; min-width: 32px;"/>`;
+  const NULL_CELL = `<div style="font-size:0.925em" class="null-cell opacity-50 italic ">no data</div>`;
 
   export let getColumnHeaderData: (pos: PivotPos) => any = () => [];
   export let getRowHeaderData: (pos: PivotPos) => any = () => [];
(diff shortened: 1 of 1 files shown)


bcolloran:
> The more I go into it, the more I see that the types are very badly written for the library, marking a lot of essential fields as optional as well. Instead of adding null checks at every variable I am more in favour of just ignoring strictNullChecks for this file and adding it to ignore list. What do you think @bcolloran ?

Hmm that's too bad about that library.

In my opinion, `Pivot.svelte` represents a very important piece of functionality for Rill, and it's a big file with a lot of surface area, so I think it's risky to disable all ts-checks for the entire file, which would  happen if we add the file to the ignore list.

My recommendation would be to try something like wrapping the functionality of the library in a small, lighweight shim file that can either (1) encapsulate whatever null checks are needed or (2) be a smaller target for ts-ignore if we really need to go that way.

But more generally: we keep encountering bugs -- and more importantly, our _customers_ keep encountering bugs -- that  are directly attributable to invalid values being passed down through the component tree. As our code base continues to grow in size and complexity, accurate types are really an essential way of documenting the contract that our code should adhere to, and checking this in CI helps us to do this, so I am not in favor of adding any more files to the ignore list. I think we need to just do the work of making sure types are valid. If it's hard work, it indicates a "code smell" -- a shaky library, poor encapsulation, overloaded or underspecified data being passed around -- and specifying the types is all the more important, because it will be even harder when someone else has to work on that code or when you yourself have to revisit it in 6 months after you have forgotten the details. 

I'll let @ericpgreen2 @AdityaHegde and @nishantmonu51  weigh in, but my *strong* recommendation is that we spend a little more time getting this stuff right when it's fresh than waiting for it to bite us later.

ericpgreen2:
I see both points-of-view. I really like adding `strictNullChecks` across the codebase, and I don't like adding more files to our lint ignore list. But we shouldn't let perfect be the enemy of the good. An e2e test for pivot table interactions can help cover us as well. I'd defer to @djbarnwal as maintainer of this feature to make the right call.

djbarnwal:
Have added null checks as well. Will also follow up with a E2E test for the table and the chart.

bcolloran (approved):

bcolloran:
This is fantastic @djbarnwal , thank you!  :-D

Approving since this logic is already merged, e2e tests can be separate

PR #3329, opened 2023-10-26
Support ISO 8601 time ranges in comparison toplist API

AdityaHegde:
For scheduled reports we need iso 8601 time ranges. This PR adds the support only for comparison toplist API. In future we can add this to all API and only use these from frontend.


AdityaHegde:
Not needed

PR #3340, opened 2023-10-27
support unnesting array dimensions

pjain1:
## Summary by CodeRabbit

New Feature:
- Introduced the ability to "unnest" dimensions in metrics views. This allows for more granular data analysis and improves the flexibility of data queries.

Improvement:
- Enhanced the handling of dimensions in metrics views, including better construction of filter clauses and conversion of dimension names to safe column names.

Compatibility:
- Included handling for older projects and different SQL dialects, ensuring smooth operation across various setups.

Performance:
- Optimized SQL queries for metrics view comparisons and toplists, particularly when dealing with unnested dimensions, leading to more efficient data retrieval.


runtime/queries/metricsview.go
```diff
@@ -322,17 +334,29 @@ func convertToXLSXValue(pbvalue *structpb.Value) (interface{}, error) {
...
 func metricsViewDimensionToSafeColumn(mv *runtimev1.MetricsViewSpec, dimName string) (string, error) {
 	dimName = strings.ToLower(dimName)
+	dimension, err := findDimension(mv, dimName)
+	if err != nil {
+		return "", err
+	}
+	return safeName(getMetricsViewDimensionName(dimension)), nil
+}
+
+func findDimension(mv *runtimev1.MetricsViewSpec, dimName string) (*runtimev1.MetricsViewSpec_DimensionV2, error) {
```
begelundmuller, line 344: For consistency with other functions here, consider calling the func `metricsViewDimension(mv, name)` (it's a big package with many util funcs)

runtime/queries/metricsview.go
```diff
@@ -322,17 +334,29 @@ func convertToXLSXValue(pbvalue *structpb.Value) (interface{}, error) {
...
-			// in that case `column` will not be present
-			return safeName(dimension.Name), nil
+			return dimension, nil
 		}
 	}
-	return "", fmt.Errorf("dimension %s not found", dimName)
+	return nil, fmt.Errorf("dimension %s not found", dimName)
+}
+
+func getMetricsViewDimensionName(dimension *runtimev1.MetricsViewSpec_DimensionV2) string {
```
begelundmuller, line 353: This resolves the column name, not the dimension name – should something like `metricsViewDimensionColumn(dim)`

runtime/queries/metricsview_aggregation.go
```diff
@@ -139,18 +139,28 @@ func (q *MetricsViewAggregation) buildMetricsAggregationSQL(mv *runtimev1.Metric
...
+			dim, err := findDimension(mv, d.Name)
 			if err != nil {
 				return "", nil, err
 			}
-
-			selectCols = append(selectCols, fmt.Sprintf("%s as %s", col, safeName(d.Name)))
-			groupCols = append(groupCols, col)
+			rawColName := getMetricsViewDimensionName(dim)
+			col := safeName(rawColName)
+			unnestColName := safeName(fmt.Sprintf("%s_%s", "unnested", rawColName))
```
begelundmuller, line 154: 1. There might be another column named `unnested_...` already :) Consider using the `tempName` util func instead
2. This var is not used in the `else` clause below, so can be moved into the `if` clause

runtime/queries/metricsview_aggregation.go
```diff
@@ -139,18 +139,28 @@ func (q *MetricsViewAggregation) buildMetricsAggregationSQL(mv *runtimev1.Metric
...
+
+			if dim.Unnest && dialect != drivers.DialectDruid {
+				// select "unnested_colName" as "colName" ... FROM "mv_table", LATERAL UNNEST("mv_table"."colName") tbl("unnested_colName") ...
+				selectCols = append(selectCols, fmt.Sprintf(`%s as %s`, unnestColName, col))
+				unnestClauses = append(unnestClauses, fmt.Sprintf(`, LATERAL UNNEST(%q.%s) tbl(%s)`, mv.Table, col, unnestColName))
```
begelundmuller, line 159: I think we need to use `safeName` on `mv.Table` instead of `%q` (since Go escaping is different from SQL escaping) – but might be wrong, please double check

runtime/queries/metricsview_comparison_toplist.go
```diff
@@ -236,12 +236,23 @@ func timeRangeClause(timeRange *runtimev1.TimeRange, td string, args *[]any) str
...
+	colName := safeName(rawColName)
+	unnestColName := safeName(fmt.Sprintf("%s_%s", "unnested", rawColName))
 
-	selectCols := []string{colName}
+	var selectCols []string
+	unnestClause := ""
+	if dim.Unnest && dialect != drivers.DialectDruid {
+		// select "unnested_colName" as "colName" ... FROM "mv_table", LATERAL UNNEST("mv_table"."colName") tbl("unnested_colName") ...
+		selectCols = append(selectCols, fmt.Sprintf(`%s as %s`, unnestColName, colName))
+		unnestClause = fmt.Sprintf(`, LATERAL UNNEST(%q.%s) tbl(%s)`, mv.Table, colName, unnestColName)
```
begelundmuller, line 252: Same comments as above

runtime/queries/metricsview_comparison_toplist.go
```diff
@@ -311,27 +322,44 @@ func (q *MetricsViewComparison) buildMetricsTopListSQL(mv *runtimev1.MetricsView
...
+	} else {
```
begelundmuller, line 360: Same

runtime/queries/metricsview_toplist.go
```diff
@@ -178,12 +178,24 @@ func (q *MetricsViewToplist) buildMetricsTopListSQL(mv *runtimev1.MetricsViewSpe
...
+	unnestColName := safeName(fmt.Sprintf("%s_%s", "unnested", rawColName))
+
+	var selectCols []string
+	unnestClause := ""
+	if dim.Unnest && dialect != drivers.DialectDruid {
+		// select "unnested_colName" as "colName" ... FROM "mv_table", LATERAL UNNEST("mv_table"."colName") tbl("unnested_colName") ...
+		selectCols = append(selectCols, fmt.Sprintf(`%s as %s`, unnestColName, colName))
+		unnestClause = fmt.Sprintf(`, LATERAL UNNEST(%q.%s) tbl(%s)`, mv.Table, colName, unnestColName)
+	} else {
+		selectCols = append(selectCols, colName)
```
begelundmuller, line 196: Same

begelundmuller (changes requested):

runtime/queries/metricsview_aggregation.go
pjain1: `safeName`  escapes the double quotes using `\` and `%q` should do the same, not sure whats the difference. Also we are using `%q` for escaping table names in comparison toplist sql and possibly other places so will need to fix that as well in case this is wrong.

pjain1: ok changed these and older references as well to `mv.Table` to `safeName(mv.Table)`

begelundmuller (approved):

PR #3346, opened 2023-10-30
Adding support of default_comparison

AdityaHegde:
## Checklist
- [x] Manual verification
- [x] Unit test coverage
- [x] E2E test coverage
- [ ] Needs manual QA?

#### Issue addressed: 
Add support for changing default comparison selections.

#### Details:
Adding `default_comparison` to dashbaord.
```
default_comparison:  
    mode:  none | time | dimension
    dimension:  <name of the dimension>
```

## Summary by CodeRabbit

New Features:
- Introduced a new feature, `DefaultComparison`, in the `MetricsViewSpec` that allows users to set a default comparison mode and dimension.
- Updated the dashboard to handle the new `DefaultComparison` feature, ensuring that the selected dimension is only updated when time comparison is not shown.

Refactor:
- Updated various functions to accept `V1MetricsViewSpec` instead of `V1MetricsView`, aligning with the new `DefaultComparison` feature.

Tests:
- Added a new function `runThroughDefaultComparisons` to verify that the dashboard updates correctly with different comparison modes.

Chores:
- Ensured that the `metricsExplorer` object has the expected default values for `showTimeComparison` and `selectedComparisonDimension` properties.


proto/rill/runtime/v1/resources.proto
```diff
@@ -153,6 +153,17 @@ message MetricsViewSpec {
...
+    DEFAULT_COMPARISON_MODE_UNSPECIFIED = 0;
+    DEFAULT_COMPARISON_MODE_NONE = 1;
+    DEFAULT_COMPARISON_MODE_TIME = 2;
+    DEFAULT_COMPARISON_MODE_DIMENSION = 3;
+  }
+  // DefaultComparison settings for comparison controls.
+  message DefaultComparison {
+    DefaultComparisonMode mode = 1;
+    string dimension = 2;
+  }
```
begelundmuller, line 166: 1. Can we inline this to `default_comparison_mode` and `default_comparison_dimension` in the spec (not the YAML). Generally the protobuf definitions are fine to have more flattened than the YAML (keeps the types easier to navigate)
2. What do you think about dropping the `mode` and just setting `dimension` to the time dimension name if you want to compare by time?

runtime/compilers/rillv1/parse_metrics_view.go
```diff
@@ -6,29 +6,29 @@ import (
 	"fmt"
 	"strings"
 	"time"
+	// Load IANA time zone data
+	_ "time/tzdata"
```
begelundmuller, line 10: Lint error – anonymous imports should be in the third section

runtime/compilers/rillv1/parse_metrics_view.go
```diff
@@ -59,6 +59,16 @@ type MetricsViewYAML struct {
 			Condition string `yaml:"if"`
 		}
 	}
+	DefaultComparison *struct {
```
begelundmuller, line 62: Remove the pointer, just check the values for `!= ""` (reduces chance of nil pointer errors, we've had those for YAML parsing edge cases before)

runtime/compilers/rillv1/parse_metrics_view.go
```diff
@@ -180,6 +190,24 @@ func (p *Parser) parseMetricsView(ctx context.Context, node *Node) error {
 		return fmt.Errorf("must define at least one measure")
 	}
 
+	if tmp.DefaultComparison != nil {
+		if tmp.DefaultComparison.Mode == "" {
+			tmp.DefaultComparison.Mode = "none"
+		} else {
```
begelundmuller, line 196: Just add `""` to the `validComparisonModes` as well

runtime/compilers/rillv1/parse_metrics_view.go
```diff
@@ -180,6 +190,24 @@ func (p *Parser) parseMetricsView(ctx context.Context, node *Node) error {
...
+			tmp.DefaultComparison.Mode = strings.ToLower(tmp.DefaultComparison.Mode)
+			if _, ok := validComparisonModes[tmp.DefaultComparison.Mode]; !ok {
+				return fmt.Errorf("invalid mode: %s. allowed values: %s",
```
begelundmuller, line 199: Use `%` for quoting user input

runtime/compilers/rillv1/parse_metrics_view.go
```diff
@@ -180,6 +190,24 @@ func (p *Parser) parseMetricsView(ctx context.Context, node *Node) error {
...
+					tmp.DefaultComparison.Mode, strings.Join(maps.Keys(validComparisonModes), ","))
+			}
+		}
+
+		if tmp.DefaultComparison.Dimension != "" {
+			if ok := names[tmp.DefaultComparison.Dimension]; !ok {
+				return fmt.Errorf("default comparison dimension %s doesnt exist", tmp.DefaultComparison.Dimension)
```
begelundmuller, line 206: `%q`, and `doesnt` -> `doesn't`

begelundmuller (changes requested):
Also check linter CI

proto/rill/runtime/v1/resources.proto
AdityaHegde: 1. Makes sense.
2. There might be more of these modes. Also the time dimension should match the `time_dimension` otherwise the UI wont work as expected, so it is better to have a single source for the `time_dimension` and comparison only mention that it is a time comparison.

runtime/compilers/rillv1/parse_metrics_view.go
AdityaHegde: Got it. My editor has different defaults. Will fix these.

web-common/src/features/dashboards/stores/dashboard-store-defaults.ts
```diff
@@ -42,11 +43,34 @@ export function setDefaultTimeRange(
...
+function setDefaultComparison(
+  metricsView: V1MetricsViewSpec,
+  metricsExplorer: MetricsExplorerEntity,
+  fullTimeRange: V1ColumnTimeRangeResponse | undefined
+) {
+  switch (metricsView.defaultComparisonMode) {
+    case MetricsViewSpecDefaultComparisonMode.DEFAULT_COMPARISON_MODE_DIMENSION:
+      metricsExplorer.selectedComparisonDimension =
+        metricsView.defaultComparisonDimension ??
+        metricsView.dimensions?.[0]?.name;
```
djbarnwal, line 57: If not `dimension` key is provided along with mode "dimension", no comparison is shown.
This is because the value of `metricsView.defaultComparisonDimension` is an empty string in such cases.

djbarnwal (changes requested):

web-common/src/features/dashboards/stores/dashboard-store-defaults.ts
AdityaHegde: Weird. I have a test for this that passes.

djbarnwal (approved):

proto/rill/runtime/v1/resources.proto
```diff
@@ -153,6 +153,17 @@ message MetricsViewSpec {
     repeated FieldConditionV2 include = 3;
     repeated FieldConditionV2 exclude = 4;
   }
+  enum DefaultComparisonMode {
+    DEFAULT_COMPARISON_MODE_UNSPECIFIED = 0;
+    DEFAULT_COMPARISON_MODE_NONE = 1;
+    DEFAULT_COMPARISON_MODE_TIME = 2;
+    DEFAULT_COMPARISON_MODE_DIMENSION = 3;
+  }
```
begelundmuller, line 161: Just realized we can change this to `ComparisonMode` (since "default" is more like a variable than a type). So it will be `ComparisonMode default_comparison_mode = 14;`

proto/rill/runtime/v1/resources.proto
begelundmuller, line 166: This type is no longer used

runtime/compilers/rillv1/parse_metrics_view.go
```diff
@@ -180,6 +192,17 @@ func (p *Parser) parseMetricsView(ctx context.Context, node *Node) error {
 		return fmt.Errorf("must define at least one measure")
 	}
 
+	tmp.DefaultComparison.Mode = strings.ToLower(tmp.DefaultComparison.Mode)
+	if _, ok := comparisonModesMap[tmp.DefaultComparison.Mode]; !ok {
+		return fmt.Errorf("invalid mode: %q. allowed values: %s",
+			tmp.DefaultComparison.Mode, strings.Join(validComparisonModes, ","))
```
begelundmuller, line 198: nit: just make this one line (we've generally been cool with wide lines on the backend, we have wide displays and space is more precious vertically; when we do break lines for readability, do it at every breakpoint for consistency (e.g. at every `,` in the scope))

begelundmuller (changes requested):

begelundmuller (approved):

PR #3362, opened 2023-11-01
Adding support for round_to_grain for APIs

AdityaHegde:
UI rounds time ranges based on selected time grain. We need the same support for reports.

This PR adds,
- [x] Support for round_to_grain.
- [x] Verify the new method `CeilTime`
- [x] First day and first month support.
- [x] Update report generation to use this.
- [x] Add tests for day light savings.

### DIFF
--- a/runtime/queries/timeutil.go
+++ b/runtime/queries/timeutil.go
@@ -1,0 +1,96 @@
+package queries
+
+import (
+	"time"
+	// Load IANA time zone data
+	_ "time/tzdata"
+
+	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
+)
+
+func TruncateTime(start time.Time, tg runtimev1.TimeGrain, tz *time.Location, firstDay, firstMonth int) time.Time {
+	switch tg {
+	case runtimev1.TimeGrain_TIME_GRAIN_MILLISECOND:
+		return start.Truncate(time.Millisecond)
+	case runtimev1.TimeGrain_TIME_GRAIN_SECOND:
+		return start.Truncate(time.Second)
+	case runtimev1.TimeGrain_TIME_GRAIN_MINUTE:
+		return start.Truncate(time.Minute)
+	case runtimev1.TimeGrain_TIME_GRAIN_HOUR:
+		start = start.In(tz)
+		start = time.Date(start.Year(), start.Month(), start.Day(), start.Hour(), 0, 0, 0, tz)
+		return start.In(time.UTC)
+	case runtimev1.TimeGrain_TIME_GRAIN_DAY:
+		start = start.In(tz)
+		start = time.Date(start.Year(), start.Month(), start.Day(), 0, 0, 0, 0, tz)
+		return start.In(time.UTC)
+	case runtimev1.TimeGrain_TIME_GRAIN_WEEK:
+		start = start.In(tz)
+		weekday := int(start.Weekday())
+		if weekday == 0 {
+			weekday = 7
+		}
+		if firstDay < 1 {
+			firstDay = 1
+		}
+		if firstDay > 7 {
+			firstDay = 7
+		}
+
+		daysToSubtract := -(weekday - firstDay)
+		if weekday < firstDay {
+			daysToSubtract = -7 + daysToSubtract
+		}
+		start = time.Date(start.Year(), start.Month(), start.Day(), 0, 0, 0, 0, tz)
+		start = start.AddDate(0, 0, daysToSubtract)
+		return start.In(time.UTC)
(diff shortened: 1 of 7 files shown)


runtime/reconcilers/report.go
```diff
@@ -367,3 +369,15 @@ func formatExportFormat(f runtimev1.ExportFormat) string {
...
 }
+
+func fillTimeRange(tr *runtimev1.TimeRange, t time.Time) *runtimev1.TimeRange {
+	if tr == nil {
+		tr = &runtimev1.TimeRange{}
+	}
+
+	// TODO: does it make sense to override start/end here? Or should we keep the user defined start/end?
+	tr.End = timestamppb.New(t)
+	tr.Start = nil
```
begelundmuller, line 380: We should override end, but not touch start. Long term, I think we'll want to pass `reportTime` into the queries, and let the query implementation decide how to let it override the time range. But for now, just overriding the end gets us the behavior we need.

runtime/reconcilers/report.go
```diff
@@ -367,3 +369,15 @@ func formatExportFormat(f runtimev1.ExportFormat) string {
 		return f.String()
 	}
 }
+
+func fillTimeRange(tr *runtimev1.TimeRange, t time.Time) *runtimev1.TimeRange {
```
begelundmuller, line 373: Maybe call it `overrideTimeRange`

runtime/queries/timeutil.go
```diff
@@ -0,0 +1,180 @@
...
+		start = start.AddDate(0, 1, 0)
+	case runtimev1.TimeGrain_TIME_GRAIN_QUARTER:
+		start = start.AddDate(0, 3, 0)
+	case runtimev1.TimeGrain_TIME_GRAIN_YEAR:
+		start = start.AddDate(1, 0, 0)
+	}
+	return TruncateTime(start, tg, tz, firstDay, firstMonth)
+}
+
+func StartTimeForRange(tr *runtimev1.TimeRange, mv *runtimev1.MetricsViewSpec) (time.Time, time.Time, error) {
```
begelundmuller, line 101: Is this more like `ResolveTimeRange`? It also returns an end time

runtime/queries/timeutil.go
```diff
@@ -0,0 +1,180 @@
...
+		}
+
+		start = time.Date(year, time.Month(firstMonth), 1, 0, 0, 0, 0, tz)
+		return start.In(time.UTC)
+	}
+
+	return start
+}
+
+func CeilTime(start time.Time, tg runtimev1.TimeGrain, tz *time.Location, firstDay, firstMonth int) time.Time {
```
begelundmuller, line 75: Not used, can remove?

runtime/queries/timeutil.go
```diff
@@ -0,0 +1,180 @@
...
+		start = time.Date(start.Year(), start.Month(), start.Day(), 0, 0, 0, 0, tz)
+		start = start.AddDate(0, 0, daysToSubtract)
+		return start.In(time.UTC)
+	case runtimev1.TimeGrain_TIME_GRAIN_MONTH:
+		start = start.In(tz)
+		start = time.Date(start.Year(), start.Month(), 1, 0, 0, 0, 0, tz)
+		start = start.In(time.UTC)
+		return start
+	case runtimev1.TimeGrain_TIME_GRAIN_QUARTER:
+		monthsToSubtract := 1 - int(start.Month())%3 // todo first month of year
```
begelundmuller, line 56: I guess we need to address that `todo`?

runtime/queries/timeutil.go
```diff
@@ -0,0 +1,180 @@
...
+		start = start.AddDate(0, 3, 0)
+	case runtimev1.TimeGrain_TIME_GRAIN_YEAR:
+		start = start.AddDate(1, 0, 0)
+	}
+	return TruncateTime(start, tg, tz, firstDay, firstMonth)
+}
+
+func StartTimeForRange(tr *runtimev1.TimeRange, mv *runtimev1.MetricsViewSpec) (time.Time, time.Time, error) {
+	tz := time.UTC
+	tr.TimeZone = "America/Los_Angeles"
```
begelundmuller, line 103: Should have been removed?

begelundmuller (changes requested):

runtime/queries/timeutil.go
AdityaHegde: Yes that name makes more sense.

runtime/reconcilers/report.go
AdityaHegde: Ya makes sense.

begelundmuller (approved):
Looks good!

PR #3365, opened 2023-11-02
Support available_time_ranges in dashboard config

AdityaHegde:
## Checklist
- [ ] Manual verification
- [ ] Unit test coverage
- [ ] E2E test coverage
- [ ] Needs manual QA?


runtime/compilers/rillv1/parser_test.go
```diff
@@ -94,6 +94,13 @@ measures:
     expression: count(*)
 first_day_of_week: 7
 first_month_of_year: 3
+available_time_ranges:
+  - P2W
+  - range: P4W
+  - range: P2M
+    comparison_ranges:
+      - P1M
+      - P1Y
```
begelundmuller, line 103: Are these comparison ranges or comparison offsets?

runtime/compilers/rillv1/parse_metrics_view.go
```diff
@@ -6,29 +6,28 @@ import (
 	"fmt"
 	"strings"
 	"time"
+	// Load IANA time zone data
+	_ "time/tzdata"
```
begelundmuller, line 10: Formatting err

proto/rill/runtime/v1/resources.proto
```diff
@@ -180,6 +184,8 @@ message MetricsViewSpec {
   uint32 first_day_of_week = 12;
   // Month number to use as the base for time aggregations by year. Defaults to 1 (January).
   uint32 first_month_of_year = 13;
+  // List of available time ranges with comparison ranges that would replace the default list.
+  repeated AvailableTimeRange available_time_ranges = 14;
```
begelundmuller, line 188: Need to merge main; conflicts with `default_comparison_mode`

begelundmuller (changes requested):

runtime/compilers/rillv1/parser_test.go
AdityaHegde: These are just for testing. Actual values are pending a decision on the format.

runtime/compilers/rillv1/parse_metrics_view.go
AdityaHegde: Ya will fix these just before merging.

runtime/compilers/rillv1/parser_test.go
begelundmuller: I meant that if they are offsets, then I think we should make the key `comparison_offsets` instead :) This sounds like you're comparing e.g. 2 months of revenue vs. 1 month or 1 year of revenue – I'm missing context, but I don't think that's what we're doing?

AdityaHegde:
Closing in favour of #3418 for backend changes. UI changes will come in a later PR after #3418 and #3417 are done.

PR #3368, opened 2023-11-03
Support explicitly overriding DuckDB threads and max_memory configs

k-anshul:
1. Adds option to set custom duckdb configurations via `rill env set connector.duckdb.boot_queries "SET max_memory='12GB';SET threads=30" --project=rill-examples`
2. `pool_size` can be capped by `rill env set connector.duckdb.pool_size 2 --project=rill-examples`
3. Limits default pool size to a max of 5.


begelundmuller:
Pushed the following changes:
- Separate each config into separate code blocks so it’s clearer how they’re resolved
- Use specific config keys for the overrides instead of boot queries. The boot queries are run for each conn that is created, and it feels unsafe to e.g. edit the memory size with multiple conns open (who knows what side effects that might have internally).

begelundmuller (approved):

PR #3371, opened 2023-11-03
Add e2e tests for TDD view

djbarnwal:
Add UI tests for TDD


nishantmonu51:
Reran the CI/CD few times and it seems to be passing consistently, LGTM. 👍

djbarnwal:
Merging so that we have tests running for TDD but as a quick follow up would create a workflow of its own for TDD tests as @AdityaHegde suggested offline.

PR #3383, opened 2023-11-03
Add pinning to table

djbarnwal:


ericpgreen2:
100% agree on your proposal to move the pin icon one row up into the header row.

ericpgreen2:
I'm not sure if it's related to this PR, but one UX problem I'm seeing for the first time is that when I move my browser focus to a different tab, then return focus to the Rill (Developer) tab where I have the TDD view open, I hit a loading state where both the chart and the time series table have spinners.

I'd expect any data refetch to happen in the background, rather than tear down the current data.

ericpgreen2 (approved):
Approving from a code standpoint 👍

bcolloran:
hey @ericpgreen2 and @djbarnwal, I know this is merged, but I want to flag an issue for you both in the future.

@djbarnwal, I have seen several places in the app where you use -1 as a sentinel value for invalid data. Sentinel values are not a best practice in languages where nullable values are available, because they obscure the semantics of the value for other users of the code (especially when undocumented, as is the case for `pinIndex` and the other places where this pattern is used). The only way that another member of the team will become aware that this invalid value may show up is if they happen to notice the "-1"s floating around, which may not be present in all the cases where this value is referenced.

The preferred approach would be to use the type  `pinIndex: number | null` or `pinIndex: number | undefined` rather than the current `pinIndex: number`. This makes it clearer to other teammates (and your future self) that not everything that shows up in a pinIndex value can necessarily be used as a valid index.

(also fyi @AdityaHegde in case you see this pattern in a review)

PR #3390, opened 2023-11-06
PIVOT ON support

egor-ryashin:


runtime/queries/metricsview_aggregation.go
```diff
@@ -76,14 +91,138 @@ func (q *MetricsViewAggregation) Resolve(ctx context.Context, rt *runtime.Runtim
...
+	})
+	if connErr != nil {
+		return connErr
+	}
+
+	conn, err := connector.Connect(ctx)
+	if err != nil {
+		return err
+	}
+	defer conn.Close()
```
begelundmuller, line 132: The `db` variable will be `nil` on every invocation after the first (which triggers `dbOnce`). Generally, this code mixes up `db` and `conn` multiple places.

I would propose moving the entire olapQuery -> temporary table in a temporary DuckDB handle logic into a util function. Something like:
```
schema, data, err := olapQuery(...)
if err != nil {...}

conn, tableName, release, err := ephemeralDuckDBForResult(schema, data)
// ...
```

If you use that for Druid, combined with a call to `WithConnection` for the DuckDB case (which provides access to the underlying `conn *sql.Conn` in the callback), you might be able to have common logic for doing the pivot (and export) for both. But it's also worth considering treating both cases completely separately, even if it's more verbose.

Regardless, I think the various bugs around connection management here show that it's worth putting some time into thinking about how to make the logic flows here as simple to understand/review as possible.

runtime/queries/metricsview_aggregation.go
```diff
@@ -11,6 +11,20 @@ import (
...
+	"sync"
+
+	"github.com/marcboeker/go-duckdb"
+	"github.com/rilldata/rill/runtime/pkg/pbutil"
+	"google.golang.org/protobuf/types/known/structpb"
+)
+
+var (
+	dbOnce    sync.Once
+	connector driver.Connector
```
begelundmuller, line 27: It's not safe to reuse a connector for multiple DBs. So it should be created at the same time as the `OpenDB` call, every time.

begelundmuller (changes requested):

runtime/drivers/duckdb/olap.go
```diff
@@ -696,3 +696,34 @@ func safeSQLString(name string) string {
...
+		}
+
+		fields[i] = &runtimev1.StructType_Field{
+			Name: ct.Name(),
+			Type: t,
+		}
+	}
+
+	return &runtimev1.StructType{Fields: fields}, nil
+}
```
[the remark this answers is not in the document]
egor-ryashin: Almost, the parameter type is different though 
image

runtime/queries/metricsview_aggregation.go
```diff
@@ -76,14 +91,138 @@ func (q *MetricsViewAggregation) Resolve(ctx context.Context, rt *runtime.Runtim
...
+	if err != nil {
+		return err
+	}
+
+	_, err = db.ExecContext(ctx, qry)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		_, _ = db.ExecContext(ctx, `DROP TABLE "`+temporaryTableName+`"`)
```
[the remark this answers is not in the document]
egor-ryashin: DROP doesn't have TEMPORARY clause

runtime/queries/metricsview_aggregation.go
```diff
@@ -25,6 +43,7 @@ type MetricsViewAggregation struct {
 	Offset             int64                                        `json:"offset,omitempty"`
 	MetricsView        *runtimev1.MetricsViewSpec                   `json:"-"`
 	ResolvedMVSecurity *runtime.ResolvedMetricsViewSecurity         `json:"security"`
+	PivotOn            []string                                     `json:"pviot_on,omitempty"`
```
mindspank, line 46: `pviot_on` > `pivot_on`

runtime/drivers/duckdb/olap.go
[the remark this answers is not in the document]
egor-ryashin: Removed.

runtime/queries/metricsview_aggregation.go
```diff
@@ -276,3 +457,163 @@ func (q *MetricsViewAggregation) buildTimestampExpr(dim *runtimev1.MetricsViewAg
...
+		query += fmt.Sprintf("%s %s", safeName(s.Name), duckDBType)
+		if i != len(schema.Fields) {
+			query += ","
+		}
+	}
+	query += ")"
+	return query, nil
+}
+
+func pbTypeToDuckDB(t *runtimev1.Type) (string, error) {
```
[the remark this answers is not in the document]
egor-ryashin: Removed.

egor-ryashin:
First benchmarks:
```
Running tool: /opt/homebrew/opt/go/libexec/bin/go test -benchmem -run=^$ -bench ^(BenchmarkMetricsViewsAggregation|BenchmarkMetricsViewsAggregation_pivot|BenchmarkMetricsViewsAggregation_spending|BenchmarkMetricsViewsAggregation_spending_pivot)$ github.com/rilldata/rill/runtime/queries -count=1

goos: darwin
goarch: arm64
pkg: github.com/rilldata/rill/runtime/queries
BenchmarkMetricsViewsAggregation-10                   	     268	   4406848 ns/op	   15070 B/op	     383 allocs/op
BenchmarkMetricsViewsAggregation_pivot-10             	     218	   5671571 ns/op	   19537 B/op	     432 allocs/op
BenchmarkMetricsViewsAggregation_spending-10          	      64	  16482316 ns/op	   15826 B/op	     373 allocs/op
BenchmarkMetricsViewsAggregation_spending_pivot-10    	      38	  32526113 ns/op	   77312 B/op	    1238 allocs/op
```
The growth is non-linear (may be due to temporary table - as the result is 10 rows).

egor-ryashin:
With the limit on subquery the benchmarks are:
```
Running tool: /opt/homebrew/opt/go/libexec/bin/go test -benchmem -run=^$ -bench ^(BenchmarkMetricsViewsAggregation|BenchmarkMetricsViewsAggregation_pivot|BenchmarkMetricsViewsAggregation_spending|BenchmarkMetricsViewsAggregation_spending_100|BenchmarkMetricsViewsAggregation_spending_pivot|BenchmarkMetricsViewsAggregation_spending_pivot_100)$ github.com/rilldata/rill/runtime/queries -count=1

goos: darwin
goarch: arm64
pkg: github.com/rilldata/rill/runtime/queries
BenchmarkMetricsViewsAggregation-10                       	     280	   4274759 ns/op	   15055 B/op	     383 allocs/op
BenchmarkMetricsViewsAggregation_pivot-10                 	     216	   5512721 ns/op	   18784 B/op	     407 allocs/op
BenchmarkMetricsViewsAggregation_spending-10              	      67	  16572626 ns/op	   15875 B/op	     373 allocs/op
BenchmarkMetricsViewsAggregation_spending_100-10          	      61	  16938342 ns/op	   87936 B/op	    2356 allocs/op
BenchmarkMetricsViewsAggregation_spending_pivot-10        	      66	  18002439 ns/op	   21155 B/op	     440 allocs/op
BenchmarkMetricsViewsAggregation_spending_pivot_100-10    	      51	  20305096 ns/op	  195780 B/op	    3670 allocs/op
```

runtime/drivers/duckdb/utils.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 // rawConn is similar to *sql.Conn.Raw, but additionally unwraps otelsql (which we use for instrumentation).
-func rawConn(conn *sql.Conn, f func(driver.Conn) error) error {
+func RawConn(conn *sql.Conn, f func(driver.Conn) error) error {
 	return conn.Raw(func(raw any) error {
 		// For details, see: https://github.com/XSAM/otelsql/issues/98
```
begelundmuller, line 18: This is an implementation detail we should try to avoid leaking from the driver

runtime/queries/metricsview_aggregation.go
```diff
@@ -76,14 +90,159 @@ func (q *MetricsViewAggregation) Resolve(ctx context.Context, rt *runtime.Runtim
...
+
+	schema, data, err := olapQuery(ctx, olap, priority, sqlString, args)
+	if err != nil {
+		return err
+	}
+
+	duckDBOLAP, _ := handle.AsOLAP("")
+	err = duckDBOLAP.WithConnection(ctx, priority, false, false, func(ctx context.Context, ensuredCtx context.Context, conn *databasesql.Conn) error {
+		temporaryTableName := tempName("_for_pivot_")
+		createTableSQL, err := duckdbolap.CreateTableQuery(schema, temporaryTableName)
```
begelundmuller, line 148: We should avoid using our DuckDB OLAP driver for this kind of ephemeral in-memory transformation. A few reasons:
- Our driver has a lot of extra overhead, like loading a bunch of extensions, migrations for catalog tables, emitting DB stats telemetry, etc.
- It leaks some pretty internal details of the driver into the `runtime/queries` package (like the `RawConn` func should not be public)
- Lifecycle management of the DuckDB handle becomes tricky with a global here, since DuckDB doesn't release memory it has taken. Since this is just for exports, it's probably better to open/close a handle on each call initially.

As discussed previously, I would suggest creating a generic `runtime/pkg/duckdbpivot` package with a `Pivot` function that takes a `sqlx.Rows` iterator as an argument, and returns a pivoted `sqlx.Rows`. Internally, it would use `database/sql` to directly open an in-memory DuckDB handle to just create the temporary table and do the pivot. 

That would also encapsulate the appender logic in a scoped package, and allow easier testing of the code.

===

An implication for the above is that here, the DuckDB and Druid cases become pretty different (since we want to avoid the overhead of converting from our DuckDB driver to `duckdbpivot`), but since the pivot SQL is pretty simple, I think some duplication is the best way to keep it simple.

runtime/queries/metricsview_aggregation.go
```diff
@@ -173,13 +333,14 @@ func (q *MetricsViewAggregation) buildMetricsAggregationSQL(mv *runtimev1.Metric
...
 	for _, m := range q.Measures {
+		sn := safeName(m.Name)
 		switch m.BuiltinMeasure {
 		case runtimev1.BuiltinMeasure_BUILTIN_MEASURE_UNSPECIFIED:
 			expr, err := metricsViewMeasureExpression(mv, m.Name)
 			if err != nil {
 				return "", nil, err
 			}
-			selectCols = append(selectCols, fmt.Sprintf("%s as %s", expr, safeName(m.Name)))
+			selectCols = append(selectCols, fmt.Sprintf("%s as %s", expr, sn))
```
begelundmuller, line 343: `sn` only referenced once in the scope – probably simpler to keep as before

runtime/queries/metricsview_aggregation.go
```diff
@@ -246,16 +407,30 @@ func (q *MetricsViewAggregation) buildMetricsAggregationSQL(mv *runtimev1.Metric
...
+			strings.Join(selectCols, ", "),
+			safeName(mv.Table),
+			strings.Join(unnestClauses, ""),
+			whereClause,
+			groupClause,
+			orderClause,
+			limitClause,
+			q.Offset,
+		)
+	}
```
begelundmuller, line 433: They seem equivalent apart from the offset clause? I would suggest we return an error if passing `q.Offset` together with `q.PivotOn`, since that's not easy to support right now

begelundmuller (changes requested):
Lastly, it's not currently adding any limits on the source Druid data. I would suggest we put a cap at e.g. 100k rows, and return an error if it's exceeded

runtime/queries/metricsview_aggregation.go
egor-ryashin: > Pivot function that takes a sqlx.Rows iterator as an argument, and returns a pivoted sqlx.Rows. Internally, it would use database/sql to directly open an in-memory DuckDB handle to just create the temporary table and do the pivot.

won't work. The function creates a handle but doesn't return it and it's not closed

egor-ryashin: btw, I cannot pass `sqlx.Rows` because I have `drivers.Result` from the `OLAPStore`

egor-ryashin: > Our driver has a lot of extra overhead, like loading a bunch of extensions, migrations for catalog tables, emitting DB stats telemetry, etc.
It leaks some pretty internal details of the driver into the runtime/queries package (like the RawConn func should not be public)
Lifecycle management of the DuckDB handle becomes tricky with a global here, since DuckDB doesn't release memory it has taken. Since this is just for exports, it's probably better to open/close a handle on each call initially.

Addressed those.

egor-ryashin:
Benchmarks:
```
goos: darwin
goarch: arm64
pkg: github.com/rilldata/rill/runtime/queries
BenchmarkMetricsViewsAggregation-10                           	     278	   4322011 ns/op	   15070 B/op	     383 allocs/op
BenchmarkMetricsViewsAggregation_pivot_2_measures-10          	     130	   8498407 ns/op	   81369 B/op	     592 allocs/op
BenchmarkMetricsViewsAggregation_pivot-10                     	     213	   5566473 ns/op	   18764 B/op	     407 allocs/op
BenchmarkMetricsViewsAggregation_spending-10                  	      62	  17000323 ns/op	   15923 B/op	     373 allocs/op
BenchmarkMetricsViewsAggregation_spending_100-10              	      64	  17758497 ns/op	   87879 B/op	    2355 allocs/op
BenchmarkMetricsViewsAggregation_spending_pivot-10            	      64	  18638850 ns/op	   21165 B/op	     440 allocs/op
BenchmarkMetricsViewsAggregation_spending_pivot_100-10        	      52	  19814269 ns/op	  195862 B/op	    3671 allocs/op
BenchmarkMetricsViewsAggregation_Druid_2_measures-10          	       6	 213449556 ns/op	   38426 B/op	     457 allocs/op
BenchmarkMetricsViewsAggregation_Druid_pivot-10               	      12	 103349292 ns/op	   21792 B/op	     256 allocs/op
BenchmarkMetricsViewsAggregation_Druid_pivot_2_measures-10    	       9	 125543921 ns/op	   28650 B/op	     343 allocs/op
```

runtime/queries/metricsview_aggregation.go
```diff
@@ -471,6 +471,9 @@ func (q *MetricsViewAggregation) buildMetricsAggregationSQL(mv *runtimev1.Metric
 			*q.Limit = 100
 		}
 		limitClause = fmt.Sprintf("LIMIT %d", *q.Limit)
+	} else if q.PivotOn != nil {
+		l := 1_000_000 / cols
+		limitClause = fmt.Sprintf("LIMIT %d", l)
 	}
```
begelundmuller, line 477: 1. Pull the number into a constant at the top of the file, as suggested in the previous comment
2. We need to return an error if the limit is exceeded (for both DuckDB and Druid). Since otherwise people would get partial/wrong results. So the limit probably needs to be `maxPivotCells + 1` and then after querying, it count the number of rows and check it's `<= maxPivotCells`.

begelundmuller (changes requested):
There are also some merge conflicts to address

begelundmuller (approved):

PR #3407, opened 2023-11-07
Adding safeguard for creation of dashboard while model is being ingested

AdityaHegde:
## Checklist
- [x] Manual verification
- [ ] Unit test coverage
- [ ] E2E test coverage
- [x] Needs manual QA?

#### Issue addressed: 
When model ingestion is slow create dashboard button can trigger for older data.

#### Details:
Adding a safeguard to make sure we wait before getting the schema before creating a dashboard. Note that pressing the button immediately after changing the model will still break this, but a user wont be that quick.

### DIFF
--- a/web-common/src/features/models/createDashboardFromModel.ts
+++ b/web-common/src/features/models/createDashboardFromModel.ts
@@ -1,0 +1,213 @@
+import { goto } from "$app/navigation";
+import { notifications } from "@rilldata/web-common/components/notifications";
+import { useDashboardFileNames } from "@rilldata/web-common/features/dashboards/selectors";
+import {
+  getFileAPIPathFromNameAndType,
+  getFilePathFromNameAndType,
+} from "@rilldata/web-common/features/entity-management/entity-mappers";
+import { getName } from "@rilldata/web-common/features/entity-management/name-utils";
+import {
+  createSchemaForTable,
+  ResourceKind,
+  useResource,
+} from "@rilldata/web-common/features/entity-management/resource-selectors";
+import { waitForResource } from "@rilldata/web-common/features/entity-management/resource-status-utils";
+import { EntityType } from "@rilldata/web-common/features/entity-management/types";
+import { generateDashboardYAMLForModel } from "@rilldata/web-common/features/metrics-views/metrics-internal-store";
+import { appScreen } from "@rilldata/web-common/layout/app-store";
+import { overlay } from "@rilldata/web-common/layout/overlay-store";
+import { behaviourEvent } from "@rilldata/web-common/metrics/initMetrics";
+import type { BehaviourEventMedium } from "@rilldata/web-common/metrics/service/BehaviourEventTypes";
+import {
+  MetricsEventScreenName,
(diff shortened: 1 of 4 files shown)


ericpgreen2:
The selectors have a lot of layers: `modelHasDataStore` → `createSchemaForTable` → `useSchemaForTable` → `createConnectorServiceOLAPGetTable`

Ideally, every selector sits just one level above a Orval-generated TanStack Query hook. So, could we make it:
- `modelHasDataStore` → `createConnectorServiceOLAPGetTable`
- `useSchemaForTable` → `createConnectorServiceOLAPGetTable`

web-common/src/features/models/createDashboardFromModel.ts
```diff
@@ -0,0 +1,213 @@
...
+
+  return createMutation<
+    Awaited<Promise<void>>,
+    TError,
+    { data: CreateDashboardFromModelRequest },
+    TContext
+  >(mutationFn, mutationOptions);
+};
+
+function modelHasDataStore(
```
ericpgreen2, line 106: Is there a better name than `modelHasData`? I’m not 100% sure what it means. Is `isModelSchemaReady` more accurate/precise? Or something else?

web-common/src/features/models/createDashboardFromModel.ts
```diff
@@ -0,0 +1,213 @@
...
+  const { mutation: mutationOptions } = options ?? {};
+  const queryClient = mutationOptions?.queryClient ?? useQueryClient();
+
+  const mutationFn: MutationFunction<
+    Awaited<Promise<void>>,
+    { data: CreateDashboardFromModelRequest }
+  > = async (props) => {
+    const { data } = props ?? {};
+    if (!data.schema)
+      throw new Error("Failed to create dashboard: Model is not ready");
```
ericpgreen2, line 67: Instead of adding this safeguard in the action, can we instead add the safeguard at the UI-level by disabling the buttons until the “model has data”?

web-common/src/features/models/createDashboardFromModel.ts
```diff
@@ -0,0 +1,213 @@
...
+      unsub();
+    });
+  });
+}
+
+/**
+ * Wrapper function that takes care of UI side effects on top of creating a dashboard from model.
+ * TODO: where would this go?
+ */
+export function useCreateDashboardFromModelUIAction(
```
ericpgreen2, line 160: Nice, this is great decoupling the UI logic from the action logic. I think this is a reasonable place to put the code.

web-common/src/features/models/createDashboardFromModel.ts
AdityaHegde: Ya makes sense. I have refactored to use this instead.

ericpgreen2 (approved):
Approving 👍, just it'd still be nice to consolidate the `useSchemaForTable` and  `createSchemaForTable` layers

AdityaHegde:
Thanks. I missed that while refactoring.

PR #3412, opened 2023-11-07
 Bugbash: make shift+click alert the same in dimension table and leaderboard

bcolloran:
fixes bugbash item here: https://www.notion.so/rilldata/Bug-Bash-Rill-GA-6867ccc289494508ae5f8d966d8902c5?pvs=4#b67e3c98591249679788a9b275588780


djbarnwal (approved):
Looks good!

PR #3417, opened 2023-11-08
Adding parsing of rill iso extensions

AdityaHegde:
To be able to support all time settings supported in UI in reports we need to extend the ISO8601 format. Using `rill-<DAX>` to add support for things like Year-to-Date, Previous-Month etc.


runtime/pkg/duration/duration.go
```diff
@@ -36,15 +57,23 @@ var (
 func ParseISO8601(from string) (Duration, error) {
```
begelundmuller, line 57: Need to update the docstring here to describe the new extension supported.

runtime/pkg/duration/duration.go
```diff
@@ -5,12 +5,20 @@ import (
...
-// Duration represents an ISO8601 duration with Rill-specific extensions.
+type Duration interface {
+	Add(t time.Time) time.Time
+	Sub(t time.Time) time.Time
+}
+
+// StandardDuration represents an ISO8601 duration with Rill-specific extensions.
 // See ParseISO8601 for details.
-type Duration struct {
+type StandardDuration struct {
```
begelundmuller, line 21: The word "standard" kind of implies that it doesn't have Rill-specific extensions. The only Rill-specific extension it has is support for `inf`. With the new interface, would suggest moving that out into a separate type `InfDuration` (then this type becomes standard-only).

runtime/pkg/duration/duration.go
```diff
@@ -36,15 +57,23 @@ var (
...
+		return StandardDuration{Inf: true}, nil
+	}
+
+	rillDur := strings.Replace(from, "rill-", "", 1)
+	if a, ok := daxToDateNotations[rillDur]; ok {
+		return ToDateDuration{anchor: a}, nil
+	}
+	if o, ok := daxOffsetNotations[rillDur]; ok {
+		return o, nil
 	}
```
begelundmuller, line 69: A brief comment explaining the `rill-` prefix here being used for dax would be helpful (future folks will lack that context)

runtime/pkg/duration/duration.go
```diff
@@ -104,3 +133,15 @@ func (d Duration) Sub(t time.Time) time.Time {
 	td := time.Duration(d.Second)*time.Second + time.Duration(d.Minute)*time.Minute + time.Duration(d.Hour)*time.Hour
 	return t.Add(-td)
 }
+
+type ToDateDuration struct {
+	anchor runtimev1.TimeGrain
+}
```
begelundmuller, line 139: Would propose changing it to `TruncToDate` instead, `ToDate` sounds like "convert to date".

Also, a small docstring would be helpful.

runtime/pkg/duration/duration.go
```diff
@@ -104,3 +133,15 @@ func (d Duration) Sub(t time.Time) time.Time {
...
+
+func (d ToDateDuration) Add(t time.Time) time.Time {
+	return time.Time{}
+}
+
+func (d ToDateDuration) Sub(t time.Time) time.Time {
+	return TruncateTime(t, d.anchor, t.Location(), 1, 1) // TODO: get first day and month
```
begelundmuller, line 146: Does this TODO need to be addressed now?

runtime/pkg/duration/duration.go
```diff
@@ -5,12 +5,20 @@ import (
 	"fmt"
 	"regexp"
 	"strconv"
+	"strings"
 	"time"
+
+	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
```
begelundmuller, line 11: We should avoid importing API types into generic `pkg` packages when possible. In this case, would consider seeing if you can find a relevant stdlib type or otherwise define an enum in this package (and have a simple `switch`-convert function on the `queries` package side).

runtime/pkg/duration/timeutil.go
```diff
@@ -0,0 +1,71 @@
+package duration
+
+import (
+	"time"
+
+	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
```
begelundmuller, line 6: Same comment about protos

runtime/pkg/duration/timeutil.go
```diff
@@ -0,0 +1,71 @@
...
+
+	// Load IANA time zone data
+	_ "time/tzdata"
+)
+
+func TruncateTime(start time.Time, tg runtimev1.TimeGrain, tz *time.Location, firstDay, firstMonth int) time.Time {
```
begelundmuller, line 12: Since this doesn't use the `Duration` type at all, should it maybe be moved to a separate package `timeutil`? (Could define the generic non-proto grain type we need in there and use it in the `duration` package)

begelundmuller (changes requested):

runtime/pkg/duration/duration.go
AdityaHegde: It is not really supported in the frontend. We can revisit this once we have scheduled reports released.

begelundmuller (approved):
Looks good!

PR #3420, opened 2023-11-08
Gray out unselected rows in table

djbarnwal:
In TDD gray out rows which are not included or excluded.

This PR also moves the `regular-table` styles inside the repo so that they can be managed easily.

### DIFF
--- a/web-common/src/features/dashboards/pivot/Pivot.svelte
+++ b/web-common/src/features/dashboards/pivot/Pivot.svelte
@@ -330,4 +330,7 @@
   :global(regular-table thead th) {
     height: var(--row-height);
   }
+  :global(regular-table tr:hover) {
+    color: none;
+  }
 </style>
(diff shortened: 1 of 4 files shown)


ericpgreen2:
I like the direction, but I think the gray is too light. See the "2.53" and "2.3" scores for color contrast:  
image
image

The numbers are close to invisible under the "low contrast" emulation:
image

Any way we can achieve the same "disabled" effect, but get the color contrast score a bit higher?

ericpgreen2 (approved):
Outside of the contrast issue, code looks good, so I'll approve for you to merge at your discretion.

djbarnwal:
> Any way we can achieve the same "disabled" effect, but get the color contrast score a bit higher?

I switched to using `ui-copy-disabled-faint` class which we use throughout the codebase. Still seeing that the contrast score is low. Since this is an issue for the whole application, will be good to address it separately.

PR #3442, opened 2023-11-09
enable strictNullChecks at root of repo, exempting targeted folders

bcolloran:
Expands strict null checks in the repo and limits exemptions to certain folders.

This PR only cleans up a few additional files, but will prevent future null check violations outside of the specified folders


ericpgreen2 (approved):
💪

PR #3481, opened 2023-11-14
Adding telemetry endpoint to admin service

AdityaHegde:
For cloud UI we need an endpoint on admin to fire telemetry.


admin/server/server.go
```diff
@@ -295,6 +300,15 @@ func (s *Server) Ping(ctx context.Context, req *adminv1.PingRequest) (*adminv1.P
 	return resp, nil
 }
 
+func (s *Server) Telemetry(ctx context.Context, req *adminv1.TelemetryRequest) (*adminv1.TelemetryResponse, error) {
+	dims := make([]attribute.KeyValue, 0)
+	for k, v := range req.Event {
+		dims = append(dims, attribute.String(k, v))
+	}
+	s.uiActivity.Emit(ctx, "cloud-ui-telemetry", 0, dims...)
```
esevastyanov, line 308: Lets use 1 as a value so that values might be aggregated and visualized
```suggestion
	s.uiActivity.Emit(ctx, "cloud-ui-telemetry", 1, dims...)
```

esevastyanov (approved):

PR #3488, opened 2023-11-14
Fix: Use column name instead of dimension name

djbarnwal:
Use column name for fetching values from API in the dimension table


nishantmonu51 (approved):
👍 , Tested locally

AdityaHegde (approved):

PR #3491, opened 2023-11-15
Fix "no data", "prev_value", and a tooltip issue

bcolloran:
This should fix the issue described here: https://rilldata.slack.com/archives/CTZ8XBQ85/p1699976127305919

before:
image

after:
image

### DIFF
--- a/web-common/src/components/data-types/PercentageChange.svelte
+++ b/web-common/src/components/data-types/PercentageChange.svelte
@@ -19,9 +19,10 @@
 
   $: if (
     !isNoData &&
+    // expanding this out in full provides type narrowing
     !isPercDiff(value) &&
     value !== null &&
-    value !== undefined && // this is a NumberParts object
+    value !== undefined &&
     typeof value !== "number"
   ) {
     // in this case, we have a NumberParts object.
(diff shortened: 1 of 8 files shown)


web-common/src/components/data-types/PercentageChange.svelte
```diff
@@ -45,7 +57,7 @@
   {dark}
 >
   <slot name="value">
-    {#if isNoData}
+    {#if isNoData || isPercDiff(value)}
```
djbarnwal, line 60: Usage of `|| isPercDiff(value)` is redundant as `isNoData` already factors it. (Also applicable for line 22)

web-common/src/components/data-types/type-utils.ts
```diff
@@ -1,6 +1,17 @@
+/**
+ * Describes the king of missing data that is being displayed.
```
djbarnwal, line 2: Nit: typo

djbarnwal (approved):
Looks good, left some nit comments. Approving so that you can merge at your convenience.

bcolloran:
Thank for the review and for catching those @djbarnwal! I've cleaned it up!

PR #3494, opened 2023-11-15
Adding support for daylight savings time in timeseries queries

AdityaHegde:
- [x] Fix extra null row around DST
- [x] Fix `first_day_of_week` giving incorrect results in the duckdb query. (Off by one day every other row) (@AdityaHegde )
- [x] Analyze performance regression (Egor)


runtime/queries/metricsview_timeseries_test.go
```diff
@@ -33,32 +33,20 @@ func TestMetricsViewsTimeseries_month_grain(t *testing.T) {
...
-	require.Equal(t, parseTime(t, "2023-08-01T00:00:00Z").AsTime(), rows[i].Ts.AsTime())
-	i++
-	require.Equal(t, parseTime(t, "2023-09-01T00:00:00Z").AsTime(), rows[i].Ts.AsTime())
-	i++
-	require.Equal(t, parseTime(t, "2023-10-01T00:00:00Z").AsTime(), rows[i].Ts.AsTime())
-	i++
-	require.Equal(t, parseTime(t, "2023-11-01T00:00:00Z").AsTime(), rows[i].Ts.AsTime())
-	i++
-	require.Equal(t, parseTime(t, "2023-12-01T00:00:00Z").AsTime(), rows[i].Ts.AsTime())
+	assertTimeSeriesResponse(t, q.Result, []string{
```
egor-ryashin, line 36: I think it will be slightly harder to find a timestamp that differs, the error line now always points to the first line.

egor-ryashin (approved):

runtime/queries/metricsview_timeseries_test.go
AdityaHegde: Ah that was the reason to do something like this. Perhaps it would be better to convert both to strings and compare lists?

begelundmuller (commented):
The changes in this PR look good to me. 

For context for others, we found that just using `time_bucket` and passing a time zone to it corrects for offsets, but still treats DST changes as separate bins (with the same labels), which is the behavior we decided we want (see [Notion](https://www.notion.so/rilldata/How-to-display-and-calculate-time-ranges-with-time-zones-97bfd537433a42e69de9413590639f4b)).

Two questions before this is ready to merge:
1. @AdityaHegde, did you confirm this does not null-fill at DST changes (i.e. just skips the value instead)?
2. @egor-ryashin, I did some testing and the Druid implementation (not changed in this PR) appears to already have the desired behavior. Is that also your impression?

egor-ryashin:
Yes, it works for both.

begelundmuller:
Update from @AdityaHegde:
> Null filling is failing with first_of_week=7. The issue is duration.Add is not honouring DST for week grain. But seem to do it for day and hour. Checking in depth

AdityaHegde:
Note that #3560 is fixed from the API side in a previous commit. Had a debug session with @AndrewRTsao , the issue is in the formatter we use in the UI. Formatting seem to convert the UTC into another UTC with an offset leading to issues based on user's time zone. We should instead use `luxon` to format it and keep the underlying UTC time the same.

Lets tackle the labelling issue separately. I have reverted the labelling changes in charts from this PR so that we can fix it together.

runtime/queries/metricsview_timeseries.go
```diff
@@ -340,16 +342,62 @@ func (q *MetricsViewTimeSeries) buildDuckDBSQL(args []any, mv *runtimev1.Metrics
...
-		safeName(mv.TimeDimension),     // 2
-		tsAlias,                        // 3
-		strings.Join(selectCols, ", "), // 4
-		safeName(mv.Table),             // 5
-		whereClause,                    // 6
-		shift,                          // 7
-	)
+	sql := ""
+	if shift == "" {
+		if q.TimeGranularity == runtimev1.TimeGrain_TIME_GRAIN_HOUR {
```
begelundmuller, line 347: I think the use of `time_bucket` might be needed for all grains of hour or less (i.e. also minutes, seconds). Otherwise, it will emit just one spiked bin instead of two at DST:

See this example:

```
D select timezone('Europe/Copenhagen', date_trunc('MINUTE', timezone('Europe/Copenhagen', '2023-10-29T00:01:10Z'::TIMESTAMPTZ))) as t;
┌──────────────────────────┐
│            t             │
│ timestamp with time zone │
├──────────────────────────┤
│ 2023-10-29 01:01:00+00   │
└──────────────────────────┘
D select timezone('Europe/Copenhagen', date_trunc('MINUTE', timezone('Europe/Copenhagen', '2023-10-29T01:01:10Z'::TIMESTAMPTZ))) as t;
┌──────────────────────────┐
│            t             │
│ timestamp with time zone │
├──────────────────────────┤
│ 2023-10-29 01:01:00+00   │
└──────────────────────────┘
```

(notice the input UTC timestamps are one hour apart, but the output timestamp is the same)

begelundmuller (changes requested):

begelundmuller:
Hey @egor-ryashin, some more questions about this PR:

1. The issue says "Fix first_day_of_week giving incorrect results" is still unresolved. Is that correct? If yes, we should fix it before merging.
2. I'm still concerned that the `timezone(date_trunc(timezone...` trick has some other issues that we're not aware of – otherwise, `time_bucket` should be able to match its performance for daily granularities. Did you investigate/ask about potential issues with this trick?
3. There's an unresolved comment about the benchmarks referencing `spending` project that's not committed. If you can't find another approach, can you at least add comments to the benchmark code about how to reproduce that so others can run the benchmarks?

egor-ryashin:
2. Looking into that.
3. Yes, but I have no definite confirmation/negation about correctness right now.

egor-ryashin:
1. It's resolved.

begelundmuller (approved):

PR #3497, opened 2023-11-15
Handling a project reset when the UI is open

AdityaHegde:
When a project is reset the instance all queries run on is deleted. If a dashboard or the project page is open it will lead to weird error message.

Handling the instance not found error. In project page it just shows the message for absence of deployment. For dashboard page we now show a better error.


web-admin/src/features/errors/error-utils.ts
```diff
@@ -1,57 +1,74 @@
...
+      (error.response.data as RpcStatus).message === "driver: not found" &&
+      (isProjectPage || isDashboardPage)
+    ) {
+      const [, org, proj] = get(page).url.pathname.split("/");
+      // goto(`/${org}/${proj}/-/status`);
+      goto(`/${org}/${proj}`).then(() =>
+        queryClient.resetQueries(getAdminServiceGetProjectQueryKey(org, proj))
+      );
+      return;
+    }
```
ericpgreen2, line 65: Can you add a comment above this block to explain why it's needed?

web-admin/src/features/errors/error-utils.ts
```diff
@@ -1,57 +1,74 @@
...
-    goto(`${ADMIN_URL}/auth/login?redirect=${window.origin}`);
-    return;
-  }
+    if (
+      error.response?.status === 400 &&
+      (error.response.data as RpcStatus).message === "driver: not found" &&
+      (isProjectPage || isDashboardPage)
+    ) {
+      const [, org, proj] = get(page).url.pathname.split("/");
+      // goto(`/${org}/${proj}/-/status`);
```
ericpgreen2, line 60: Old code?

web-admin/src/routes/+layout.svelte
```diff
@@ -11,18 +11,15 @@
     QueryClient,
     QueryClientProvider,
   } from "@tanstack/svelte-query";
-  import { globalErrorCallback } from "../features/errors/error-utils";
+  import {
+    createErrorCallback,
+    globalErrorCallback,
```
ericpgreen2, line 16: `globalErrorCallback` is defined but not used

web-admin/src/features/errors/error-utils.ts
```diff
@@ -1,57 +1,74 @@
...
+      goto(`/${org}/${proj}`).then(() =>
+        queryClient.resetQueries(getAdminServiceGetProjectQueryKey(org, proj))
```
ericpgreen2, line 62: Is the `GetProject` query the one that will hit the "driver: not found" error? If so, I wonder if we could:
1. Add the `query` parameter to the global error callback. E.g.: `(error: AxiosError, query: Query) => ...`
2. Here, use `query.reset()` instead of the `queryClient`

Then we can avoid making the changes in `routes/+layout.svelte`

AdityaHegde: All queries can have this error. Especially when the user is in the dashboard page it needs to catch everything.

web-admin/src/features/errors/error-utils.ts
AdityaHegde: I guess i left it out as a question on where we should be redirecting to. The status page or the dashboard list page.

ericpgreen2 (approved):
Looks good. I just edited the error page copy slightly.

One thing we should also address is that the project page's dashboard links should be disabled when the resources are pending (like after a project reset), but that can be done in a separate PR.

Before merging, would you mind filling out the PR description so we can have more detail if we ever look back at this PR in the future?

PR #3507, opened 2023-11-16
Use correct column property in table

djbarnwal:
Use correct column property in dimension table. 
Fixes 
- Duplicate columns
- Filtering


nishantmonu51 (approved):
👍, LGTM

PR #3508, opened 2023-11-16
Updating a resource delete to send a delete event

AdityaHegde:
When a resource is deleted end up getting a WRITE event instead. This breaks the tracking for what is reconciling.

### DIFF
--- a/runtime/catalog_cache.go
+++ b/runtime/catalog_cache.go
@@ -391,7 +391,7 @@ func (c *catalogCache) updateDeleted(name *runtimev1.ResourceName) error {
 	r.Meta.SpecUpdatedOn = timestamppb.Now()
 	c.link(r)
 	c.dirty[nameStr(r.Meta.Name)] = r.Meta.Name
-	c.addEvent(r.Meta.Name, r, runtimev1.ResourceEvent_RESOURCE_EVENT_WRITE)
+	c.addEvent(r.Meta.Name, r, runtimev1.ResourceEvent_RESOURCE_EVENT_DELETE)
 	return nil
 }
 


ericpgreen2:
In a perfect world, would this be fixed on the backend? I'm a bit nervous that this is tech debt on the frontend side. The `switch` statement for the different resource events was super clear, so it'd be a shame to get rid of it.

AdityaHegde:
The event with resource object is a soft delete. When hard delete happens there will not be a resource object. Another thing to note, if we treat soft delete as an update we will end up refreshing the queries which will fail. This change basically solves these 2 issues.

web-common/src/features/entity-management/resource-invalidations.ts
```diff
@@ -52,14 +51,10 @@ export function invalidateResourceResponse(
...
   // so, we should not `await` here
-  switch (res.event) {
-    case V1ResourceEvent.RESOURCE_EVENT_WRITE:
-      invalidateResource(queryClient, instanceId, res.resource);
-      break;
-
-    case V1ResourceEvent.RESOURCE_EVENT_DELETE:
-      invalidateRemovedResource(queryClient, instanceId, res.resource);
-      break;
+  if (res.resource?.meta?.deletedOn) {
```
ericpgreen2, line 54: To make this code a little more self-documenting, what about something like:
```typescript
// Soft deletes precede hard deletes in the reconcile process
const isSoftDelete = !!res.resource?.meta?.deletedOn

if (isSoftDelete) ...
```

ericpgreen2 (approved):

Issue #3490, opened 2023-11-15
Chore: clear out unused parts of /dev folder


bcolloran:
the /dev folder is no longer being used, so we should remove it so that it doesn't generate false positives when doing code search.

bcolloran:
Update: @djbarnwal  is still using the `dev` route for some work, so it's worth keeping it around. On slack he said he'd be willing to clear out everything other than what he is using, so reassigning this issue to him. Thanks @djbarnwal !

PR #3510, opened 2023-11-16
remove dev route

bcolloran:
just a chore PR to reduce false positives when doing full text search in the repo. @ericpgreen2 @AdityaHegde @djbarnwal -- I haven't seen any recent usage of this folder, please confirm that it can be removed.


bcolloran:
will fix #3490

ericpgreen2:
Thank you! This folder always bugged me :) A nice follow-on would to be remove the now-needless `(application)` route group.

ericpgreen2 (approved):

bcolloran:
This is still being used by @djbarnwal , so he is going to follow up with a PR that removes everything he doesn't need

PR #3515, opened 2023-11-17
Runtime: Ability to cast low cardinality varchar columns to enum

k-anshul:
1. Ability to cast low cardinality varchar columns to enum
2. Merged transporter pkg with duckdb pkg. All transporter related files now start with transporter_xx.
3. Handling of enums is flaky, shown to user as varchar.
4. Fixed an unrelated miss related to older PRs in azure driver.


runtime/drivers/duckdb/olap.go
```diff
@@ -630,6 +630,106 @@ func (c *connection) execWithLimits(parentCtx context.Context, stmt *drivers.Sta
...
+		if err != nil {
+			return fmt.Errorf("failed switch db %q: %w", dbName, err)
+		}
+		defer func() {
+			// switch to original db, notice `db.schema` just doing USE db switches context to `main` schema in the current db if doing `USE main`
+			// we want to switch to original db and schema
+			switchErr = c.Exec(ensuredCtx, &drivers.Statement{Query: fmt.Sprintf("USE %s.%s", safeSQLName(currentDB), safeSQLName(currentSchema))})
+		}()
+
+		err = c.Exec(ctx, &drivers.Statement{Query: fmt.Sprintf("DROP TYPE %s", safeSQLName(enum))})
```
nishantmonu51, line 701: NIT: Use DROP TYPE IF EXISTS. 
Also, since we are creating a new versioned DB on every type, would we ever hit the case when enum already exists?

nishantmonu51: You can also use CREATE OR REPLACE TYPE in the create command to avoid running DROP TYPE command.

runtime/drivers/duckdb/olap.go
```diff
@@ -630,6 +630,106 @@ func (c *connection) execWithLimits(parentCtx context.Context, stmt *drivers.Sta
...
+				return err
+			}
+		}
+		_ = res.Close()
+		if !exists {
+			return fmt.Errorf("column %q can't be converted to enum, has zero non null values", col)
+		}
+
+		// scan current db and current schema
+		res, err = c.Execute(ctx, &drivers.Statement{Query: "SELECT current_database(), current_schema()"})
```
nishantmonu51, line 674: wondering, should this and all other statements in this method use ensuredCtx instead ?

go.mod
```diff
@@ -273,4 +273,6 @@ replace github.com/apache/calcite-avatica-go/v5 v5.2.0 => github.com/rilldata/ca
 // revert to original SDK once this PR is merged : https://github.com/googleapis/google-cloud-go/pull/8506
 replace cloud.google.com/go/bigquery v1.53.0 => github.com/k-anshul/google-cloud-go/bigquery v0.0.0-20230830110740-57fe623cde03
 
+replace github.com/marcboeker/go-duckdb v1.5.2 => github.com/k-anshul/go-duckdb v0.0.0-20231117093648-132fd3a6b235
```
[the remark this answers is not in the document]
k-anshul: Yes. The driver currently doesn't have support for enums so I will be raising a PR in the original driver to add support for enums. Will revert then.

runtime/drivers/duckdb/olap.go
k-anshul: I was hoping to use `ctx` till we drop the enum so that query can be cancelled if required. Post that we are better off with ignoring query cancellation and completing everything to not end up in some failed state like table exists but type doesn't.

runtime/drivers/duckdb/olap.go
k-anshul: `CREATE OR REPLACE` doesn't exist for `TYPE` so need to drop first and then recreate. 

> Also, since we are creating a new versioned DB on every type, would we ever hit the case when enum already exists?

Yeah I was thinking there may be conflicts during refresh but I think even during refresh we create new db. Will check this and remove it if required.

runtime/drivers/duckdb/olap.go
nishantmonu51: ok

nishantmonu51 (approved):

PR #3516, opened 2023-11-17
Error telemetry in cloud

AdityaHegde:
## Checklist
- [x] Manual verification
- [ ] Unit test coverage
- [ ] E2E test coverage
- [x] Needs manual QA?

#### Issue addressed: 
We do not have any feedback from cloud UI.

#### Details:
As a first step adding an event for error faced by the users in cloud.

Data is directly pulled from the kafka topic basically skipping the intake client. So the end store is the same as the dev pipeline.


web-common/src/metrics/ErrorEventHandler.ts
```diff
@@ -36,4 +38,29 @@ export class ErrorEventHandler {
...
+    return this.metricsService.dispatch("errorBoundaryEvent", [
+      this.commonUserMetrics,
+      this.getScreenNameFromPage(),
+      api,
+      status,
+      message,
+    ]);
+  }
+
+  private getScreenNameFromPage() {
```
ericpgreen2, line 52: One option is to leverage the functions in `web-admin/features/navigation/nav-utils.ts`. Particularly for the project page, we should make sure we're capturing all the sub-routes.

web-common/src/metrics/service/MetricsService.ts
```diff
@@ -49,45 +54,58 @@ export class MetricsService
...
+      user_id: localConfig.user_id,
+      analytics_enabled: localConfig.analytics_enabled,
+      mode: localConfig.readonly ? "read-only" : "edit",
+    };
+  }
+
+  public loadCloudFields(fields: CloudMetricsFields) {
+    this.commonFields = {
+      // TODO: build_id and version
+      app_name: "rill-developer",
```
ericpgreen2, line 76: `rill-cloud`?

web-common/src/metrics/service/MetricsService.ts
```diff
@@ -49,45 +54,58 @@ export class MetricsService
...
+      client_id: this.getOrSetClientID(),
+      is_dev: fields.isDev,
+      project_id: fields.projectId,
+      user_id: fields.userId,
+      analytics_enabled: true,
+      mode: "read-only",
```
ericpgreen2, line 82: Given Rill Cloud doesn't have modes, can we omit `mode`?

ericpgreen2:
Can you add handling for client-side errors at runtime?

I think something like this:
```typescript
window.onerror = (msg, url, line, col, error) => {
    // capture error & emit telemetry
    // show a full error page
};
window.onunhandledrejection = (event) => {
    // capture error & emit telemetry
    // show a full error page
};
```

ericpgreen2:
Can you add details in the PR description about where the telemetry gets sent?

admin/server/telemetry.go
```diff
@@ -0,0 +1,37 @@
...
+
+func (s *Server) Telemetry(ctx context.Context, req *adminv1.TelemetryRequest) (*adminv1.TelemetryResponse, error) {
+	dims := make([]attribute.KeyValue, 0)
+	for k, v := range req.Event.AsMap() {
+		a, ok := toKeyValue(k, v)
+		if ok {
+			dims = append(dims, a)
+		}
+	}
+	s.uiActivity.Emit(ctx, "cloud-ui-telemetry", 1, dims...)
```
esevastyanov, line 18: What do you think about passing `name` and `value` as part of `TelemetryRequest` so that it will be easier to pass various metrics later?

esevastyanov (approved):

web-common/src/metrics/service/ErrorEventFactory.ts
```diff
@@ -26,6 +27,14 @@ export interface ErrorEvent extends MetricsEvent {
   glob: boolean;
 }
 
+export interface ErrorBoundaryEvent extends MetricsEvent {
+  action: ErrorEventAction;
+  screen_name: MetricsEventScreenName;
+  api: string;
+  status: string;
+  message: string;
+}
```
ericpgreen2, line 36: Think it'd make sense to split this into something like the following two types?
1. `HTTPErrorEvent` – to be used in our TanStack Query error handler, where the `api` field is relevant.
2. `JavascriptErrorEvent` – to be used in the Javascript error listeners, where a `stacktrace` field would be relevant.

web-admin/src/features/navigation/nav-utils.ts
```diff
@@ -18,3 +19,26 @@ export function isDashboardPage(page: Page): boolean {
...
+    case isProjectPage(page):
+      return MetricsEventScreenName.Project;
+    case isDashboardPage(page):
+      return MetricsEventScreenName.Dashboard;
+    case isReportPage(page):
+      return MetricsEventScreenName.Report;
+    case isReportExportPage(page):
+      return MetricsEventScreenName.ReportExport;
+  }
+  return MetricsEventScreenName.Home;
```
ericpgreen2, line 43: What about `MetricsEventScreenName.Other`? E.g., when we add the Alerts pages, if we forget to add a mapping, then we'd want the telemetry to say "other" not "home".

web-admin/src/features/errors/error-utils.ts
```diff
@@ -130,3 +150,39 @@ export function createErrorPagePropsFromRoutingError(
     body: "Try refreshing the page, and reach out to us if that doesn't fix the error.",
   };
 }
+
+export function addJavascriptErrorListeners() {
+  const errorHandler = (errorEvt: ErrorEvent) => {
+    console.log("error", errorEvt);
```
ericpgreen2, line 156: Leftover console log?

web-admin/src/features/errors/error-utils.ts
```diff
@@ -9,12 +12,29 @@ import { ADMIN_URL } from "../../client/http-client";
...
 
 export function createGlobalErrorCallback(queryClient: QueryClient) {
-  return (error: AxiosError) => {
+  return (error: AxiosError, query: Query) => {
     const isProjectPage = get(page).route.id === "/[organization]/[project]";
     const isDashboardPage =
       get(page).route.id === "/[organization]/[project]/[dashboard]";
 
-    if (!error.response) return;
+    console.log("createGlobalErrorCallback");
```
ericpgreen2, line 20: Leftover console log

web-admin/src/routes/+layout.svelte
```diff
@@ -37,6 +42,9 @@
 
   beforeNavigate(retainFeaturesFlags);
   clearViewedAsUserAfterNavigate(queryClient);
+  initCloudMetrics();
+
+  onMount(() => addJavascriptErrorListeners());
```
ericpgreen2, line 47: As a follow-up, can you add these to Rill Developer?

ericpgreen2:
This is looking good! I haven't QA'd it though. Can add a screenshot of the various telemetry in action in the downstream viz tool? Rill Cloud or Rill Legacy? It'd be very cool if this telemetry were included as a dashboard in the `rill-cloud-metrics` project.

web-admin/src/features/navigation/nav-utils.ts
AdityaHegde: I think unknown might be a better fit.

web-admin/src/routes/[organization]/[project]/+layout.svelte
```diff
@@ -21,6 +24,14 @@
...
 
   $: onProjectPage = isProjectPage($page);
+
+  $: if ($page.params.project && $user.data?.user?.id) {
+    metricsService.loadCloudFields({
+      isDev: window.location.host.startsWith("localhost"),
+      projectId: $page.params.project,
+      userId: $user.data?.user?.id,
+    });
+  }
```
Comment (ericpgreen2) [inline web-admin/src/routes/[organization]/[project]/+layout.svelte:34]: 1. What about tracking the `organizationId` too?
2. Can we load the `userId` from the `AvatarButton.svelte` component? Because the userID will be present if the user is logged in on the Home and Org pages.

ericpgreen2 (approved):
Left a comment on `metricsService.loadCloudFields()`, and there's still a leftover console log, else looks good!

Looking forward to seeing the Rill dashboard!

PR #3520, opened 2023-11-19
Fix: No apply in style tags

djbarnwal:
`@apply` can break dev server when making changes to css files. 
More at https://tailwindcss.com/docs/functions-and-directives#using-apply-with-per-component-css

Use the styles directly instead of using `@apply`


ericpgreen2 (approved):
Since it looks like this is the only place we use the `ui-card` class, can we also remove it from `app.css`?

PR #3532, opened 2023-11-21
fix: auto indent yaml

mindspank:
## Checklist
- [x] Manual verification
- [ ] Unit test coverage
- [ ] E2E test coverage
- [ ] Needs manual QA?

Removing the enter hook will make auto indent work again for yaml files. I have tested for both models and sources and can't find anything that would break due to the change. Auto complete etc still works.


ericpgreen2 (approved):
Nice!

PR #3533, opened 2023-11-21
fix formatter warnings

### LINKED ISSUE #3527: Browser console is full of `humanizeDataTypeUnabridged` warnings
On some dashboards (like [this one](https://ui.rilldata.com/demo/rill-github-analytics/duckdb_commits)), the browser console is flooded by these warnings:
image

bcolloran:
Fixes #3527. Tagging @djbarnwal, because I made the formatters accept `undefined` as you requested last week, and therefore removed the need for the `safeFormatter` function, so I took that out. Please confirm that this meets your needs! :-)

### DIFF
--- a/web-common/src/features/dashboards/time-dimension-details/util.ts
+++ b/web-common/src/features/dashboards/time-dimension-details/util.ts
@@ -12,22 +12,6 @@ export function transposeArray(arr, rowCount, columnCount) {
   return columnarBody;
 }
 
-/**
- * This formatter is less strict and
- * returns undefined if the input is undefined
- * */
-export function safeFormatter(formatter) {
-  return (value) => {
-    if (value === undefined) {
-      return undefined;
-    } else if (value === null) {
-      return null;
-    } else {
-      return formatter(value);
-    }
-  };
-}
-
 export function getClassForCell(
   palette: "fixed" | "scrubbed" | "default",
   highlightedRow,
--- a/web-common/src/lib/number-formatting/format-measure-value.ts
+++ b/web-common/src/lib/number-formatting/format-measure-value.ts
@@ -70,13 +70,20 @@ function humanizeDataTypeUnabridged(value: number, type: FormatPreset): string {
  * This higher-order function takes a measure spec and returns
  * a function appropriate for formatting values from that measure.
  *
- * As of Nov 2023, all measure values supplied to the client
- * are in the form of a number or undefined for missing values.
+ * You may optionally add type paramaters to allow non-numeric null
+ * undefined values to be passed through unmodified.
+ * - `createMeasureValueFormatter<null | undefined>(measureSpec)` will pass through null and undefined values unchanged
+ * - `createMeasureValueFormatter<null>(measureSpec)` will pass through null values unchanged
+ * - `createMeasureValueFormatter<undefined>(measureSpec)` will pass through undefined values unchanged
+ *
+ *
+ * FIXME: we want to remove the need for this to *ever* accept undefined values,
+ * as we switch to always using `null` to represent missing values.
  */
-export const createMeasureValueFormatter = (
+export function createMeasureValueFormatter<T extends null | undefined = never>(
   measureSpec: MetricsViewSpecMeasureV2,
   useUnabridged = false
-): ((value: number | undefined) => string | undefined) => {
+): (value: number | T) => string | T {
   const humanizer = useUnabridged
     ? humanizeDataTypeUnabridged
     : humanizeDataType;
@@ -84,7 +91,9 @@ export const createMeasureValueFormatter = (
   // Return and empty string if measureSpec is not provided.
   // This may e.g. be the case during the initial render of a dashboard,
   // when a measureSpec has not yet loaded from a metadata query.
-  if (measureSpec === undefined) return (_value: number) => "";
+  if (measureSpec === undefined) {
+    return (value: number | T) => (typeof value === "number" ? "" : value);
+  }
 
   // Use the d3 formatter if it is provided and valid
   // (d3 throws an error if the format is invalid).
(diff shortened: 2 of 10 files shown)


web-common/src/features/dashboards/time-dimension-details/util.ts
```diff
@@ -12,22 +12,6 @@ export function transposeArray(arr, rowCount, columnCount) {
...
-/**
- * This formatter is less strict and
- * returns undefined if the input is undefined
- * */
-export function safeFormatter(formatter) {
-  return (value) => {
-    if (value === undefined) {
-      return undefined;
-    } else if (value === null) {
-      return null;
```
djbarnwal, line 24: We still need the ability to return `null` as output when the input is `null`. The core formatter returns "null" (a string) for `null` input.

With string passed down, the data table shows the string as is 
image

djbarnwal (changes requested):

web-common/src/features/dashboards/time-dimension-details/util.ts
bcolloran: ah ok, thanks @djbarnwal, I'll update that!

web-common/src/lib/number-formatting/format-measure-value.ts
```diff
@@ -70,31 +70,44 @@ function humanizeDataTypeUnabridged(value: number, type: FormatPreset): string {
...
+ * - `createMeasureValueFormatter<null | undefined>(measureSpec)` will pass through null and undefined values unchanged
+ * - `createMeasureValueFormatter<null>(measureSpec)` will pass through null values unchanged
+ * - `createMeasureValueFormatter<undefined>(measureSpec)` will pass through undefined values unchanged
+ *
+ *
+ * FIXME: we want to remove the need for this to *ever* accept undefined values,
+ * as we switch to always using `null` to represent missing values.
  */
-export const createMeasureValueFormatter = (
+export function createMeasureValueFormatter<T extends null | undefined = never>(
```
bcolloran, line 83: @djbarnwal I added the ability for the formatter to accept and pass through nulls and/or undefined values. I'm not sure we actually use undefined values anywhere, but I've left that in just to be safe. However in the future we'll want to deprecate that ability as we tighten up our types everywhere and make sure that we aren't using undefined to represent missing data

web-common/src/features/dashboards/time-dimension-details/util.ts
bcolloran: fixed!

bcolloran:
Hey @djbarnwal thanks for the first review. I made some changes to the way this works, and its ready for you to take another look. Thanks!

web-common/src/lib/number-formatting/format-measure-value.ts
djbarnwal: We use `null` as an identifier for missing data (API query completed and data is missing from source) and `undefined` as an identifier for unknown data (API query in progress and undefined is the transient value). For most cases we wait for the values to be available before rendering their components but in some cases (such as TDD table), we might have a mix of actual data, missing data and unknown data at the same time (since we are running 100s of API calls and don't want to wait for every value to be available before rendering). `undefined` is useful in those cases. I would be open to have an alternate system for this if you have any suggestions.

djbarnwal (approved):
Looks good, thanks for the changes!

web-common/src/lib/number-formatting/format-measure-value.ts
bcolloran: Ok, gotcha.

I suspect that we can probably in all cases push the `undefined` boundary out to a higher level of the component tree -- in general for value formatters in particular, the value formatter should really only be invoked in the case of a value that is actually ready. If there is no value ready, we would want to show a loading spinner or something in a higher level component, but in any case, that should be the decision and responsibility of component requesting the formatted value, and not of the value formatter itself. And as it stands now (after this PR and with the `safeFormatter` idea), in the case of an `undefined` value, the formatter just passes that value back to the caller unchanged, so the caller still needs to decide what to do with an `undefined` -- the formatter does not actually provide any _formatting_ for the undefined (and the same goes for nulls).

Additionally, there are some places where the formatters are used where they _do not_ have the possibility of encountering nulls and undefined. But having the possibility that those might be allowed can make the code at those call sites more complex, and/or the cognitive load on the engineer greater to think about the different possibilities. (In this PR I have done some optional type stuff to partly get around this in terms of the code complexity (the `null` and `undefined` cases don't have to be handled in code unless the calling code opts in to that), but the cognitive load of having to know about those optional types still exists)

So this is currently not a great separation of concerns and responsibilities -- it makes the types looser on the formatting function, and the formattter doesn't operate on it's inputs in a way that narrows the output types or provides it's caller with guarantees about what it will get back -- the caller will still have to handle the undef/null case. If the formatter actually formatted e.g. `null`-> "no data" and `undefined` -> "loading" or something like that, it would be a different matter, but as it stands we add complexity and uncertainty about return types etc to a function whose scope we could limit to simply taking numbers and returning strings.

I think that in the long run, and across the whole code base, a cleaner, simpler, less cognitively burdensome arrangement would be:
- the formatter function that always takes input of one type and returns outputs of one type
- don't make it the responsibility of the formatter to accept null and undef since: (1) calling code may or may not have nulls and undefs, and (2) passing those values back unchanged means the calling code still needs to handle those cases anyway

And this is a general architectural principle: having simple types makes things easier, and having certainty about the types you are dealing with makes things easier. You want union types in as few places as possible, and when you encounter one, it is preferable to narrow the type and dispatch the narrowed type to code designed for handling that type. And `any` types are terrible -- they are the opposite of this. JS does not encourage these patterns, so that's the hole we have to dig ourselves out of, but we can get there, we'll just need to learn some new habits and put a little more thought into things like this that can slip through the cracks in a JS world :-)

This is not urgent, so we can move in that direction gradually, but architecturally, this is the direction I think we want to head.

(fyi @ericpgreen2 and @AdityaHegde since this conversation has zoomed out a bit -- just for a bit of background on how I'm thinking about this kind of thing)

PR #3537, opened 2023-11-22
Docs: Adding project CLI commands and flexible project structure callouts 

AndrewRTsao:
Adding new pages for:
- rill project describe
- rill project logs
- rill project refresh
- rill project reset

Deprecating (redirecting the end user to the refresh and reset commands instead):
- rill project reconcile

Updating:
- Source YAML page (to better call out flexible project structure)
- Model YAML page (to better call out flexible project structure)
- Dashboard YAML page (to bette call out flexible project structure)
- Better linking the above reference pages from the develop section

This is mostly mentioning files that are nested under the default resource directories will still be treated as the same resource kind and documenting the **kind** property, plus making better callouts about these reference pages. These particular changes / docs updates are also informed by the following [Notion page](https://www.notion.so/rilldata/Flexible-Project-structure-platform-support-1890ea64caf8460a8b465f12d6654d0f).

### DIFF
--- a/docs/docs/develop/import-data.md
+++ b/docs/docs/develop/import-data.md
@@ -33,7 +33,7 @@ path: /path/to/data.csv
 
 Rill will ingest the data next time you run `rill start`.
 
-Note that if you provide a relative path, *the path should be relative to your Rill project root* (where your `rill.yaml` file is located), **not** relative to the `sources` directory.
+Note that if you provide a relative path, _the path should be relative to your Rill project root_ (where your `rill.yaml` file is located), **not** relative to the `sources` directory.
 
 :::tip Source Properties
 
--- a/docs/docs/develop/import-data.md
+++ b/docs/docs/develop/import-data.md
@@ -33,7 +33,13 @@ path: /path/to/data.csv
 
 Rill will ingest the data next time you run `rill start`.
 
-Note that if you provide a relative path, the path should be relative to your Rill project root (where your `rill.yaml` file is located), not relative to the `sources` directory.
+Note that if you provide a relative path, *the path should be relative to your Rill project root* (where your `rill.yaml` file is located), **not** relative to the `sources` directory.
+
+:::tip Source Properties
+
+For more details about available configurations and properties, check our [Source YAML](../reference/project-files/sources) reference page.
+
+:::
 
 ## Adding a remote source
 
@@ -57,7 +63,11 @@ You can also push filters to your source definition using inline editing. Common
 - Push transformations for key fields to source (particularly casting time fields, data types)
 - Resolve ingestion issues by declaring types (examples: STRUCT with different values to VARCHAR, fields mixed with INT and VARCHAR values)
 
-For details about all available properties for all remote connectors, see the syntax [reference](../reference/project-files/sources).
+:::tip Source Properties
+
+For more details about available configurations and properties, check our [Source YAML](../reference/project-files/sources) reference page.
+
+:::
 
 ## Authenticating remote sources
 
--- a/docs/docs/reference/cli/project/reconcile.md
+++ b/docs/docs/reference/cli/project/reconcile.md
@@ -11,7 +11,7 @@ rill project reconcile [<project-name>] [flags]
 
 :::warning Deprecation Notice 
 
-This is a **legacy** command that's been preserved for backwards compatibility. *Starting in Rill v0.37.0 and newer*, we recommend using the [rill project refresh](refresh.md) and [rill project reset](reset.md) commands to refresh and reset / redeploy your projects respectively.
+This is a **legacy** command that's been preserved for backwards compatibility. *Starting in Rill v0.37.0 and newer*, we recommend using the [rill project refresh](refresh.md) and [rill project reset](reset.md) commands to refresh and reset/redeploy your projects respectively.
 
 :::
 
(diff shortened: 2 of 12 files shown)


docs/docs/develop/import-data.md
```diff
@@ -33,7 +33,13 @@ path: /path/to/data.csv
 
 Rill will ingest the data next time you run `rill start`.
 
-Note that if you provide a relative path, the path should be relative to your Rill project root (where your `rill.yaml` file is located), not relative to the `sources` directory.
+Note that if you provide a relative path, *the path should be relative to your Rill project root* (where your `rill.yaml` file is located), **not** relative to the `sources` directory.
```
k-anshul, line 36: `*the path should be relative to your Rill project root*` This shows up as italic right now, shouldn't this be bold?

docs/docs/reference/cli/project/reconcile.md
```diff
@@ -9,6 +9,12 @@ Send trigger to deployment
 rill project reconcile [<project-name>] [flags]
 ```
 
+:::warning Deprecation Notice 
+
+This is a **legacy** command that's been preserved for backwards compatibility. *Starting in Rill v0.37.0 and newer*, we recommend using the [rill project refresh](refresh.md) and [rill project reset](reset.md) commands to refresh and reset / redeploy your projects respectively.
```
k-anshul, line 14: Nit: reset/redeploy instead of reset / redeploy

docs/docs/reference/project-files/dashboards.md
```diff
@@ -6,6 +6,12 @@ sidebar_position: 30
 
 In your Rill project directory, create a `<dashboard_name>.yaml` file in the `dashboards` directory. Rill will ingest the dashboard definition next time you run `rill start`.
 
+:::info Did you know?
+
+Files that are *nested at any level* under your `dashboards` directory will be assumed to be metric definitions.
```
k-anshul, line 11: I think either this is not working yet or there is a bug. I couldn't get it working.
cc : @ericpgreen2

docs/docs/reference/project-files/dashboards.md
```diff
@@ -57,3 +63,6 @@ _**`security`**_ - define a [security policy](../../develop/security) for the da
   - _**`include`**_ - List of dimension or measure names to include in the dashboard. If `include` is defined all other dimensions and measures are excluded. _(optional)_
     - **`if`** - Expression to decide if the column should be included or not. It can leverage templated user attributes. Needs to be a valid SQL expression that evaluates to a boolean. _(required)_
     - **`names`** - List of fields to include. Should match the `name` of one of the dashboard's dimensions or measures. _(required)_
+
+**`kind`**
+ — Can be one of `{source,model,dashboard}` and used by Rill to identify the resource kind. This property is optional and only necessary when the YAML file exists outside of the default location (e.g. a dashboard YAML outside of the `dashboards` directory).
```
k-anshul, line 68: > Can be one of `{source,model,dashboard}`

Since this page is specific to dashboards may be we should specify `kind` as `dashboard` only and similarly on sources page specify `kind` as `source` only ?

docs/docs/reference/project-files/dashboards.md
ericpgreen2: Correct, we haven't yet added support for a flexible project structure in Rill Developer. The backend reconciliation should work, but the frontend doesn't yet account for it.

docs/docs/develop/import-data.md
AndrewRTsao: I had intentionally italicized this part, since I bolded a part later, to avoid overusing the bolding. Open to formatting changes though.

docs/docs/reference/project-files/dashboards.md
AndrewRTsao: Okay cool I preempted this part then because it resulted in a [customer impacting issue](https://rilldata.slack.com/archives/C05693FLY7Q/p1700495656176499) (they had archived legacy models in a different subfolder but didn't include the "kind" -> resulting in backend error).

I think we can just wait to merge these changes once the frontend support is added

docs/docs/reference/project-files/dashboards.md
AndrewRTsao: Good callout. Since we don't currently exactly have an equivalent page for models, I wanted to at least point out what the available options are. I've rewritten this section to make it more specific though to the resource.

ericpgreen2:
There's a bunch of great docs here! Given full support for the flexible project structure isn't immediate, what about putting those callouts in a different PR that we can merge when ready? Then we won't block on the docs for `project reset`, `project reconcile`, `project describe`, etc.

AndrewRTsao:
@ericpgreen2 Great call out! I've gone ahead and split out the initial mentions to the flexible project structure into a separate commit, which we can use and further develop once the feature is fully implemented. This PR should now only contain the new CLI commands along with some formatting improvements (adding tool tip to link to our resource reference pages).

New PR (related to flexible project structure pointers): https://github.com/rilldata/rill/pull/3561

docs/docs/reference/project-files/dashboards.md
k-anshul: Sure. By the way there is one for both source and model as well :`docs/docs/reference/project-files/models.md`

k-anshul (approved):

PR #3540, opened 2023-11-22
Add report "Run now" button

ericpgreen2:
This PR adds a "Run now" button to the report page. This button is used to trigger an ad-hoc run of the report.

[Figma mock](https://www.figma.com/file/Qt6EyotCBS3V6O31jVhMQ7/RILL?type=design&node-id=5744%3A592897&mode=design&t=cyloPlXpIACsksrK-1)


ericpgreen2:
> Saw that the button mock has a different color than the one preset here. Is that intended?

Yeah, Janet is playing around with different primary colors, but her explorations are still WIP.

PR #3541, opened 2023-11-22
Adding the theme resource to support custom themes in the UI

### LINKED ISSUE #3408: Platform support for themes
See details in [Notion](https://www.notion.so/rilldata/Implement-support-for-custom-themes-7d2e56f0ce58411b8685ca5954108772)

AdityaHegde:
closes #3408

This PR adds the `Theme` resource. We are using https://github.com/mazznoer/csscolorparser to parse and store the RGBA values of the primary and secondary colours.


mindspank:
Implements #3408

proto/rill/runtime/v1/resources.proto
```diff
@@ -301,6 +302,27 @@ message BucketExtractPolicy {
...
+  ThemeState state = 2;
+}
+
+message ThemeSpec {
+  message Color {
+    float red = 1;
+    float green = 2;
+    float blue = 3;
+    float alpha = 4;
+  }
```
begelundmuller, line 318: Feels like a generic type, so maybe pull it out into a standalone generic type in `runtime/v1/color.proto`?

proto/rill/runtime/v1/resources.proto
```diff
@@ -301,6 +302,27 @@ message BucketExtractPolicy {
...
+  message Colors {
+    optional Color primary = 1;
+    optional Color secondary = 2;
+  }
+  Colors colors = 1;
```
begelundmuller, line 323: Would recommend flattening here (not in the YAML though), to just `primary_color` and `secondary_color`. Flattening is nice in the protobufs since it reduces the number of types.

begelundmuller (changes requested):

begelundmuller (approved):

PR #3543, opened 2023-11-22
Upgrade to Svelte 4

### LINKED ISSUE #2684: Upgrade to Svelte 4
https://svelte.dev/docs/v4-migration-guide

ericpgreen2:
Closes #2684

This PR upgrades Svelte 3 to Svelte 4. A handful of changes were required:
- I ran the Svelte 4 [migration script](https://svelte.dev/docs/v4-migration-guide) on each of our `web` workspaces
- I upgraded our `web` dependencies to versions that support Svelte 4
    - One package `@rgossiaux/svelte-headlessui` has not yet added official support for Svelte 4. Given community members have reported that [the package does work with Svelte 4 in practice](https://github.com/rgossiaux/svelte-headlessui/issues/152), I've included an [npm override](https://docs.npmjs.com/cli/v9/configuring-npm/package-json#overrides) in the root `package.json`.
- I fixed a few cases where Svelte 4 is more strict than Svelte 3
    - ARIA roles are required for interactive elements
    - Default slot bindings are no longer exposed to named slots (see the `floating-element` changes)
    - The `derived()` store function no longer accepts falsy values (see the `timeseries-data-store` refactor)
- I fixed failing unit tests
    - Some were failing due to our protobuf-base64 encoding/decoding functions. Fixed by migrating to `@bufbuild/protobuf`'s base64 functions.
    - Some were failing due to components' `onMount` not getting called by `vitest`. Fixed [by following this StackOverflow post](https://stackoverflow.com/questions/76577665/vitest-and-svelte-components-onmount).


web-auth/src/components/EmailPassForm.svelte
```diff
@@ -68,7 +68,7 @@
 
 <div>
   {#if showForm}
-    <div class="mt-6 mb-4 flex flex-col gap-y-4" transition:slide>
+    <div class="mt-6 mb-4 flex flex-col gap-y-4" transition:slide|global>
```
AdityaHegde, line 71: We have seen issues with global transitions. Has it been fixed in svelte4? Applies to all places where this change has been made.

web-common/src/components/chip/removable-list-chip/RemovableListChip.svelte
```diff
@@ -77,7 +77,7 @@ are details left to the consumer of the component; this component should remain
         {active}
       />
     </Chip>
-    <div slot="tooltip-content" transition:fly|local={{ duration: 100, y: 4 }}>
+    <div slot="tooltip-content" transition:fly={{ duration: 100, y: 4 }}>
```
AdityaHegde, line 80: `|local` was added to fix issues with elements lingering. Has it been fixed in svelte 4? Applies to all instances of this change.

web-common/src/components/chip/removable-list-chip/RemovableListChip.svelte
```diff
@@ -93,6 +93,7 @@ are details left to the consumer of the component; this component should remain
   <RemovableListMenu
     {excludeStore}
     slot="floating-element"
+    let:toggleFloatingElement
```
AdityaHegde, line 96: Why is this needed?

web-common/src/components/menu/wrappers/SimpleActionMenu.svelte
```diff
@@ -41,6 +41,7 @@ and the menu closes.
   <slot {handleClose} toggleMenu={toggleFloatingElement} {active} />
   <Menu
     slot="floating-element"
+    let:handleClose
```
AdityaHegde, line 44: Why is this needed? Wont it mask the variable defined above on `WithTogglableFloatingElement`?

AdityaHegde (commented):
Looks good mostly, just a few concerns.

web-auth/src/components/EmailPassForm.svelte
ericpgreen2: In Svelte 4, [transitions are local by default](https://svelte.dev/docs/v4-migration-guide#transitions-are-local-by-default). In Svelte 3, they were global by default. So this code ensures our application behaves the same as it did with Svelte 3.

All the changes to transitions were made automatically when I ran the Svelte 4 migration script: `npx svelte-migrate@latest svelte-4`.

I agree we should be wary of global transitions, but let's consider application behavior changes in a separate PR.

web-common/src/components/chip/removable-list-chip/RemovableListChip.svelte
ericpgreen2: See above – transitions are now local by default

web-common/src/components/chip/removable-list-chip/RemovableListChip.svelte
ericpgreen2: See [the migration guide's section on default slot bindings](https://svelte.dev/docs/v4-migration-guide#default-slot-bindings).

Previously, we were using _default_ slot props (like `toggleFloatingElement`) in _named_ slots (like `slot="floating-element"`). But that's no longer allowed. Now, to migrate, I've added a _named_ slot prop `toggleFloatingElement`.

AdityaHegde (approved):

PR #3550, opened 2023-11-24
Runtime: external storage enabled by default

k-anshul:
Needs https://github.com/rilldata/rill/pull/3546 before this can be merged.

- `duckdb` driver now enables `duckdb-ext-storage` support by default.
- Removed `duckdb-ext-storage` so db needs to updated after these changes are merged.
- Added `duckdb-int-storage` which allows to disable external db storage in case its required.


begelundmuller (commented):
Not going to go ahead with this until we've gotten to the bottom of this issue, which appears to be related to ATTACH: https://github.com/rilldata/rill-private-issues/issues/17

begelundmuller:
Awaiting resolution of this issue: https://github.com/rilldata/rill-private-issues/issues/62

admin/deployments.go
```diff
@@ -52,25 +52,25 @@ func (s *Service) createDeployment(ctx context.Context, opts *createDeploymentOp
...
+			return nil, fmt.Errorf("slot count can't be 0 for driver 'duckdb-int-storage'")
 		}
 
 		olapConfig["dsn"] = fmt.Sprintf("%s.db", path.Join(alloc.DataDir, instanceID))
 		olapConfig["cpu"] = strconv.Itoa(alloc.CPU)
 		olapConfig["memory_limit_gb"] = strconv.Itoa(alloc.MemoryGB)
 		olapConfig["storage_limit_bytes"] = strconv.FormatInt(alloc.StorageBytes, 10)
 		embedCatalog = false
-	case "duckdb-ext-storage": // duckdb driver having capability to store table as view
+	case "duckdb":
```
begelundmuller, line 68: There are still old deployments with `duckdb-ext-storage` in the DB, so I think we should support that case here as well (for resets)

cli/pkg/local/app.go
```diff
@@ -516,3 +525,29 @@ func initLogger(isVerbose bool, logFormat LogFormat) (logger *zap.Logger, cleanu
 		luLogger.Close()
 	}
 }
+
+func isExternalStorageEnabled(dbPath string, variables map[string]string) (bool, error) {
```
begelundmuller, line 529: Add a comment that this is for temporary backwards compatibility

begelundmuller (changes requested):
Also notice merge conflicts

cli/pkg/local/app.go
k-anshul: Theoretically this is not temporary, a project which was created before these changes will have to always operate with the older logic. Moving the comment below this line here.

admin/deployments.go
k-anshul: I was planning to change all of these to `duckdb` before release but yeah handling both is better.

begelundmuller (approved):

PR #3558, opened 2023-11-27
Improve toast notification that shows on scheduled report creation

ericpgreen2:
Addresses [this UXQA item in Notion](https://www.notion.so/rilldata/Match-design-of-success-notification-36c678ff28bb44eb8bda58cb7e76f285).

Note: The notifications/toasts code is quite hacky. At some point, we should review all our notifications, consider leveraging a third party library (like [Melt UI](https://www.melt-ui.com/docs/builders/toast)), and refactor accordingly.

### DIFF
--- a/web-common/src/components/icons/Check.svelte
+++ b/web-common/src/components/icons/Check.svelte
@@ -1,6 +1,7 @@
 <script>
   export let size = "1em";
   export let color = "currentColor";
+  export let className = "";
 </script>
 
 <svg
@@ -8,6 +9,7 @@
   viewBox="0 0 24 24"
   fill="none"
   xmlns="http://www.w3.org/2000/svg"
+  class={className}
 >
   <path
     fill-rule="evenodd"
--- a/web-common/src/components/notifications/NotificationCenter.svelte
+++ b/web-common/src/components/notifications/NotificationCenter.svelte
@@ -1,7 +1,8 @@
 <script lang="ts">
   import Notification from "./Notification.svelte";
-  import notificationStore from "./notificationStore";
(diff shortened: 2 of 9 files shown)


web-common/src/components/icons/Check.svelte
```diff
@@ -1,18 +1,20 @@
...
   viewBox="0 0 24 24"
   fill="none"
   xmlns="http://www.w3.org/2000/svg"
+  class={className}
 >
   <path
     fill-rule="evenodd"
     clip-rule="evenodd"
-    d="M18.5873 8.89285L10.4977 18.1532L5.40894 12.9881L7.01172 11.409L10.3986 14.8467L16.8928 7.4126L18.5873 8.89285Z"
+    d="M18.8161 6.57646C19.0184 6.75315 19.0449 7.05808 18.8763 7.26707L10.0888 18.1532L5.35091 13.3443C5.15711 13.1476 5.15947 12.831 5.35618 12.6372L6.24661 11.7599C6.44332 11.5661 6.7599 11.5685 6.9537 11.7652L9.98966 14.8467L17.1579 5.82284C17.3354 5.59943 17.6635 5.56958 17.8784 5.75728L18.8161 6.57646Z"
```
djbarnwal, line 17: Why do we need this change?

djbarnwal (approved):
Looks good!

web-common/src/components/icons/Check.svelte
ericpgreen2: Janet updated the Check icon. The new version is slightly larger (within its frame) and more closely mirrors the size of the Close icon, which is on the opposite side of the toast component.

PR #3559, opened 2023-11-27
add default tooltip max width to fix flicker noticed by Nishant

bcolloran:
Fixes flicker mentioned here https://rilldata.slack.com/archives/C02T907FEUB/p1701116172362659, add adds a default (overridable) max width to all tooltips


ericpgreen2 (approved):
Nice, makes sense 👍

PR #3561, opened 2023-11-28
Docs: Adding primer about flexible project structure in our reference docs

AndrewRTsao:
Splitting out the docs changes about the flexible project structure so we can merge once full support has been added:
- Calls out the flexible project structure in each resource reference page
- Documents the **kind** property that is needed for YAML files that exist outside their default directory
- Minor cleanup to align the formatting of our dashboard properties page with our source properties page

### DIFF
--- a/docs/docs/reference/project-files/dashboards.md
+++ b/docs/docs/reference/project-files/dashboards.md
@@ -6,32 +6,38 @@ sidebar_position: 30
 
 In your Rill project directory, create a `<dashboard_name>.yaml` file in the `dashboards` directory. Rill will ingest the dashboard definition next time you run `rill start`.
 
+:::info Did you know?
+
+Files that are *nested at any level* under your `dashboards` directory will be assumed to be metric definitions.
+
+:::
+
 ## Properties
 
-_**`model`**_ — the model name powering the dashboard with no path _(required)_
+**`model`** — the model name powering the dashboard with no path _(required)_
 
-_**`title`**_ — the display name for the dashboard _(required)_
+**`title`** — the display name for the dashboard _(required)_
 
-_**`timeseries`**_ — the timestamp column from your model that will underlie x-axis data in the line charts _(optional)_. If not specified, the line charts will not appear.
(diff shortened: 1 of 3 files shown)


docs/docs/reference/project-files/dashboards.md
```diff
@@ -6,32 +6,38 @@ sidebar_position: 30
...
+:::info Did you know?
+
+Files that are *nested at any level* under your `dashboards` directory will be assumed to be metric definitions.
+
+:::
+
 ## Properties
 
-_**`model`**_ — the model name powering the dashboard with no path _(required)_
+**`model`** — the model name powering the dashboard with no path _(required)_
```
begelundmuller, line 17: We also support `table:` instead of `model:` (they are aliases under the hood). The motivation for `table:` is that a dashboard can also run a) directly on a source, b) on a pre-existing table (as is the case with our current Druid-backed projects).

This PR might not be the place to discuss it, but we might also consider standardizing on `table:` for all cases?

begelundmuller (approved):
LGTM

docs/docs/reference/project-files/dashboards.md
AndrewRTsao: That's a great callout and at least I'll mention it in the docs until we've decided whether we want to standardize on one over the other

AndrewRTsao:
FYI @begelundmuller should we wait to add these primers in our docs until frontend support has been added for the flexible project structure? From speaking with @ericpgreen2, this still may take some time before fully implemented.

AndrewRTsao:
> @AndrewRTsao I'm fine either way. I guess it's a question of whether we want to document something that works, but requires using another editor than Rill Developer for modeling?

I think we'll wait until frontend support has been added. Originally, I wanted to avoid potential issues coming up in case of arbitrary yaml files in the project directory but I think it'll introduce more confusion preemptively mentioning in docs.

AndrewRTsao:
Will close for now

PR #3570, opened 2023-11-28
add e2e flow to smoketest formatters

bcolloran:
also adds a couple aria roles to make testing a couple of these things easier


bcolloran:
fixes #3440

web-local/test/ui/utils/commonHelpers.ts
```diff
@@ -85,3 +85,11 @@ export async function waitForValidResource(
     }
   });
 }
+
+export async function waitForDashboard(page: Page) {
```
ericpgreen2, line 89: `waitForAdbidsDashboard` to be more precise?

ericpgreen2 (approved):
Looks good!

web-local/test/ui/utils/commonHelpers.ts
bcolloran: this line actually fell out of this review when I updated to align with @AdityaHegde's PR. 

agree in principle with this renaming, we can adjust in subsequent PRs. (though i also thought that at the moment the adbids dashboard is the only dashboard we e2e test, to there is no ambiguity?)

Issue #3424, opened 2023-11-08
clear out the `--ignore` section of our typescript checks for svelte components


bcolloran:
Currently, our CI basically doesn't type check a huge number of our svelte components. We need to start pulling things out of this list. (And of course we should never add new items!)

```
npx svelte-check --workspace web-common --no-tsconfig --ignore "src/components/(data-graphic|notifications|column-profile|virtualized-table|menu|modal|chip|overlay|floating-element|data-types|BarAndLabel.svelte|button|preview-table/ConnectedPreviewTable.svelte),src/features/models/workspace/inspector,src/**/workspace,src/features/dashboards/(time-series|leaderboard|time-controls),src/layout/navigation/NavigationEntry.svelte"
```

# checklists
(updated on 2023-11-29 to only show the current state, removing checkboxes already checked)
(updated again on 2023-12-14)
(updated 2024-02-12)
(updated on 2024-02-20)

here are PRs for the last 4 files:
https://github.com/rilldata/rill/pull/4053
https://github.com/rilldata/rill/pull/4092
https://github.com/rilldata/rill/pull/4093
https://github.com/rilldata/rill/pull/4110
**once these merge this issue can finally be closed!!!**

## web-common
- [ ] src/features/dashboards/time-series
    - [ ] MetricsTimeSeriesCharts.svelte

- [ ] src/features/dashboards/time-controls/
    - [ ] TimeRangeSelector.svelte
    - [ ] TimeControls.svelte
- [ ] src/components/data-graphic/elements/GraphicContext.svelte

## web-local
done!

PR #3571, opened 2023-11-28
more svelte check --ignore cleanup

bcolloran:
part of  https://github.com/rilldata/rill/issues/3424


web-common/src/features/metrics-views/workspace/editor/Placeholder.svelte
```diff
@@ -53,21 +53,20 @@
...
         }),
     });
 
-    const dashboardYAML = generateDashboardYAMLForModel(
-      modelName,
-      schemaResp.schema
-    );
+    const dashboardYAML = schemaResp?.schema
+      ? generateDashboardYAMLForModel(modelName, schemaResp?.schema)
+      : "";
```
bcolloran, line 69: @ericpgreen2 @AdityaHegde -- please double check me here. Pretty sure that this would just fails silently previously if the schema was not defined (`generateDashboardYAMLForModel` does not handle that case),  but please verify that returning an empty string in case of no schema is ok

web-common/src/components/virtualized-table/sections/PinnedColumns.svelte
```diff
@@ -1,10 +1,10 @@
 <script lang="ts">
-  import type { VirtualizedTableColumns } from "../../../types";
   import { createEventDispatcher } from "svelte";
   import Cell from "../core/Cell.svelte";
   import ColumnHeader from "../core/ColumnHeader.svelte";
   import Row from "../core/Row.svelte";
   import type { PinnedColumnSide } from "../types";
+  import type { VirtualizedTableColumns } from "@rilldata/web-local/lib/types";
```
ericpgreen2, line 7: Can we move this type into `web-common`? `web-common` shouldn't be importing from `web-local`

.github/workflows/web-test.yml
```diff
@@ -75,7 +75,7 @@ jobs:
         run: |-
           npx prettier --check "web-common/**/*"
           npx eslint web-common --quiet
-          npx svelte-check --threshold error --workspace web-common --no-tsconfig --ignore "src/components/(data-graphic|notifications|column-profile|virtualized-table|menu),src/features/models/workspace/inspector,src/**/workspace,src/features/dashboards/(time-series|time-controls),src/layout/navigation/NavigationEntry.svelte"
+          npx svelte-check --threshold error --workspace web-common --no-tsconfig --ignore "src/components/(data-graphic|column-profile),src/features/dashboards/(time-series|time-controls),src/features/sources/workspace/SourceWorkspaceHeader.svelte,src/features/models/workspace/ModelWorkspaceHeader.svelte,src/features/dashboards/workspace/Dashboard.svelte"
```
ericpgreen2, line 78: Love that this list is getting smaller!

ericpgreen2 (approved):

PR #3577, opened 2023-11-29
Runtime: Fix failures in models if sql has comment at the end

### LINKED ISSUE #3464: Commented code at end of model causes error
**Describe the bug**
Adding comment to a sql model at end of model causes issue. At beginning or middle of model does not seem to cause the same.

**To Reproduce**
1. Comment a line at end of sql model
2. Produces build error below

image

Additional reference:
https://discord.com/channels/959796421832343562/959801632151969792/1173429640866562099

**Expected behavior**
No error

**Screenshots**
See above

k-anshul:
closes #3464


begelundmuller (approved):
Seems like a good solution, and I think we already do this some other places

PR #3582, opened 2023-11-29
Add back Storybook smoketest dependencies

ericpgreen2:
#3552 removed some Storybook dependencies that weren't obviously used. It turns out they're used by the Storybook smoketest, which runs in CI/CD: `npm run storybook:smoketest -w web-common`

While we don't actively have any stories in the smoketest, this PR adds back the dependencies to give us the capability going forward.

image


bcolloran (approved):
makes sense 👍

PR #3587, opened 2023-11-30
Runtime: Adds `decimal256` support

### LINKED ISSUE #3528: Bigquery: Add support for autocast for BigNumeric columns
Add a source config for auto cast from BigNumeric to double while ingesting data.

1. Default behaviour would be to allow for implicit typecasting and add warnings
2. Developers can override the default property and disable auto-cast by setting the config at source or project level

k-anshul:
closes #3528


runtime/drivers/bigquery/sql_store.go
```diff
@@ -334,8 +334,24 @@ func (f *fileIterator) downloadAsJSONFile() error {
...
+			r, ok := row[f].(*big.Rat)
+			if !ok {
+				continue
+			}
+			num, _ := r.Float64()
+			if math.IsInf(num, 0) {
+				row[f] = r.FloatString(38)
+			} else {
+				row[f] = num
 			}
```
begelundmuller, line 355: This will only work for numbers fitting in a float64. Is there any chance we can improve precision here?

One idea (haven't tried it, just based on looking at the docs) is:
```go
n := new(big.Float)
n.SetRat(r)
row[f] = n.String()
```

begelundmuller (changes requested):

runtime/drivers/bigquery/sql_store.go
k-anshul: No. The if part takes car of outputting numbers of arbitrary precision and width. I chose 38 which is the max precision for `decimal256`. 
I am eagerly trying to see if the number can fit in `float64` so that those numbers are output as float instead of varchar. I am assuming if users have problem casting columns in submitted sql they would have problem casting in model as well. Also I am suspecting most columns marked as BigNumeric would rarely need such high precisions.

```
r := big.NewRat(1, 2)
_, ok := r.SetString("5.7896044618658097711785492504343953926634992332820282019728792003956564819967E+38")
if !ok {
	panic("yellow")
}

fmt.Println(r.FloatString(38)) // prints 578960446186580977117854925043439539266.34992332820282019728792003956564819967

n := new(big.Float)
n.SetRat(r)
fmt.Println(n.SetPrec(38).String()) // prints 5.789604462e+38
```

begelundmuller (approved):

PR #3590, opened 2023-11-30
Embedded dashboards 

pjain1:
Still needs UI polish to remove the top header bar


docs/docs/integration/embedding.md
```diff
@@ -0,0 +1,137 @@
...
+
+```bash
+
+rill service create <service_name> [--org "My Organization"]
+```
+
+To issue service token for an existing service account, use the following command:
+
+```bash
+rill service token issue <service_name> [--org "My Organization"]
```
nishantmonu51, line 47: Can we also mention about  
1. How to revoke the service account tokens ? 
2. How to delete a service account
3. Does issuing a new token invalidates any old tokens ?

pjain1: unhid service account cmds, added docs and a reference to them here.

docs/docs/integration/embedding.md
```diff
@@ -0,0 +1,139 @@
...
+```html
+<iframe src="<iframeSrc>" width="100%" height="1000" frameborder="0" ></iframe>
+```
+
+### Example
+Here's an example of how to create a dashboard component in a React application:
+
+```jsx
+import React, { useEffect, useState } from 'react';
+import { useParams } from 'react-router-dom';
```
ericpgreen2, line 100: This import isn't used in the example anywhere

docs/docs/integration/embedding.md
```diff
@@ -0,0 +1,139 @@
...
+
+const Dashboard = () => {
```
ericpgreen2, line 102: Let's call it `RillDashboard`, given this is from the perspective of another app

docs/docs/integration/embedding.md
```diff
@@ -0,0 +1,139 @@
...
+    };
+    getIframeUrl();
+  }, []);
+
+  return (
+    <iframe
+      src={iframeUrl}
+      width="100%"
+      height="1000"
+      frameborder="0"
```
ericpgreen2, line 132: `frameborder` is a deprecated attribute, so let's remove it

docs/docs/integration/embedding.md
```diff
@@ -0,0 +1,139 @@
...
+      allowfullscreen
```
ericpgreen2, line 133: should be camel cased: `allowFullScreen`

docs/docs/integration/embedding.md
```diff
@@ -0,0 +1,139 @@
...
+        }
+      );
+      const resp = await response.json();
+      setIframeUrl(resp.iframeSrc);
+    };
+    getIframeUrl();
+  }, []);
+
+  return (
+    <iframe
```
ericpgreen2, line 128: It'll avoid an accessibility warning if we include a `title` property, e.g. `title="rill-dashboard"`

docs/docs/integration/embedding.md
```diff
@@ -0,0 +1,139 @@
...
+        `https://admin.rilldata.com/v1/organizations/<org-name>/projects/<project-name>/iframe`,
+        {
+          method: 'POST',
+          headers: {
+            'Content-Type': 'application/json',
+            Authorization: `Bearer <rill-svc-token>`,
+          },
+          body: JSON.stringify({
+            resource: '<dashboard-name>',
+            user_email: '<user-email>',
```
ericpgreen2, line 117: Since the `user_email` is highly likely to be dynamic, it'd be nice to demonstrate that in this example with a hook like `const { userEmail } = useUser();` and by adding `userEmail` to the `useEffect` dependency array.

docs/docs/integration/embedding.md
ericpgreen2: This works instead: `style={{ border: 0 }}`

docs/docs/integration/embedding.md
```diff
@@ -0,0 +1,139 @@
...
+  }, []);
+
+  return (
+    <iframe
+      src={iframeUrl}
+      width="100%"
+      height="1000"
+      frameborder="0"
+      allowfullscreen
+    ></iframe>
```
ericpgreen2, line 134: It'd be a little cleaner to use `<iframe />` rather than `<iframe></iframe>`

docs/docs/integration/embedding.md
ericpgreen2: Actually, I think it'd be cleaner if we removed this attribute altogether – not sure it's necessary for a basic example

ericpgreen2:
Looking pretty good from a frontend perspective! I added a couple commits: 
- We needed an explicit app-wide `white` background color, else it looked broken when I embedded it into an app with a colored background
- I hid the top navigational bar for embeds

docs/docs/integration/embedding.md
pjain1: sure thanks

pjain1:
@ericpgreen2 thanks for the review and fixes. I think we should add theme support as well here once its in the main. It will be passed in as query parameter to the `embed` page. Also I still see the header, am I missing something ?
Screenshot 2023-12-04 at 11 53 52 AM

docs/docs/integration/embedding.md
```diff
@@ -89,21 +89,21 @@ The response will contain an iframe URL that can be used to embed the dashboard
...
 import React, { useEffect, useState } from 'react';
-import { useParams } from 'react-router-dom';
 
-const Dashboard = () => {
+const RillDashboard = () => {
   const [iframeUrl, setIframeUrl] = useState('');
 
   useEffect(() => {
-    const getIframeUrl = async () => {
+      const { userEmail } = useUser();
```
ericpgreen2, line 105: The `useUser()` hook should go at the top-level above `useEffect`. Also, it'd be good to show its import, like:
```jsx
import { useUser } from '../hooks/useUser';
```

ericpgreen2:
> @ericpgreen2 thanks for the review and fixes. I think we should add theme support as well here once its in the main. It will be passed in as query parameter to the `embed` page.

Yeah, agreed. Theme docs can go in a separate PR. I left one last comment, but I'll approve from the frontend perspective.

ericpgreen2 (approved):

docs/docs/integration/embedding.md
pjain1: sure

docs/docs/integration/embedding.md
```diff
@@ -0,0 +1,139 @@
...
+          headers: {
+            'Content-Type': 'application/json',
+            Authorization: `Bearer <rill-svc-token>`,
+          },
+          body: JSON.stringify({
+            resource: '<dashboard-name>',
+            user_email: userEmail,
+          }),
+        }
+      );
```
begelundmuller, line 121: I feel ambivalent about this – on the one hand, it's an easy way to get a POC in place, but on the other hand, it's very dangerous to suggest folks to put their service tokens in the frontend.

The alternative is to put a templated snippet for calling the embedder's own backend, which should do the handshake of getting the iframe URL.

begelundmuller (changes requested):

admin/server/embed.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+		attr, err = s.getAttributesFor(ctx, user.ID, proj.OrganizationID, proj.ID)
+		if err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+	case *adminv1.GetIFrameRequest_Attributes:
+		attr = forVal.Attributes.AsMap()
+	default:
+		return nil, status.Error(codes.InvalidArgument, "invalid 'for' type")
+	}
```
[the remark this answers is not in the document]
pjain1: moved to `deployment.go`

docs/docs/integration/embedding.md
```diff
@@ -0,0 +1,139 @@
...
+    "user_email":"<user-email>"
+}'
+```
+
+The API accepts the following parameters:
+
+| Parameter | Description                                      | Required           |
+| --- |--------------------------------------------------|--------------------|
+| resource | The name of the dashboard to embed               | Yes                |
+| user_email | The email of the user to embed the dashboard for | Yes                |
```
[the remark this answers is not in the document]
pjain1: can we suggest to send empty attributes in case no user is present ?

docs/docs/integration/embedding.md
pjain1: changed to use backend api

docs/docs/integration/embedding.md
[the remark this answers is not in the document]
pjain1: added a tip that `"attributes":{}` can be used for no user context, this way no change required in backend.

docs/docs/integration/embedding.md
```diff
@@ -0,0 +1,139 @@
...
+```bash
+
+curl -X POST --location 'https://admin.rilldata.com/v1/organizations/<org-name>/projects/<project-name>/iframe' \
+--header 'Content-Type: application/json' \
+--header 'Authorization: Bearer <rill-svc-token>' \
+--data-raw '{
+    "resource": "<dashboard-name>",
+    "user_email":"<user-email>"
+}'
+```
```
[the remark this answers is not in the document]
pjain1: added examples using chatgpt

docs/docs/integration/embedding.md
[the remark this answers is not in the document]
begelundmuller: Not exactly sure about `oneof`, but generally in protobuf, all fields are optional, so you might just be able to check for the case where `.For == nil` and add empty attributes in that case.

docs/docs/integration/embedding.md
```diff
@@ -246,12 +246,12 @@ public class DashboardController {
...
+| --- |------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|---------------------------------|
+| resource | The name of the dashboard to embed                                                                                                                                                             | Yes                             |
+| user_email | The email of the user to embed the dashboard for                                                                                                                                               | Yes (either this or `attributes`) |
+| attributes | Json payload to be put in the access token, used to pass attributes to the dashboard for enforcing policies. When using this make sure to pass all the attributes used in your security policy | No (either this or `user_email`)  |
```
begelundmuller, line 253: I guess `user_email` should not be required also, and the `(either this or ...` comment no longer applies? Since they're not needed for non-security policy projects

begelundmuller (changes requested):

docs/docs/integration/embedding.md
pjain1: yes made user_email optional. but the comment would be needed to indicate that only one can be set, right ?

begelundmuller (approved):

docs/docs/integration/embedding.md
```diff
@@ -0,0 +1,328 @@
...
+      const getIframeUrl = async () => {
+          try {
+              const response = await fetch(`/api/getIframeUrl`, {
+                  method: 'POST',
+                  headers: {
+                      'Content-Type': 'application/json',
+                  },
+                  body: JSON.stringify({
+                      resource: dashboardName,
+                      user_email: '<user-email>',
```
begelundmuller, line 302: The email should not be passed from the frontend (that would be a security loophole that's easy to exploit), it should be resolved on the backend based on the embedder's auth logic for resolving the current user

pjain1: yeah I think this was left by mistake, backend api already has a template for email. So I can just remove it from here.

begelundmuller (approved):

PR #3592, opened 2023-11-30
Refactor: Format tootltip value for percentages

djbarnwal:
Big number tooltips for measures with percentage as format will be formatted as well.

https://www.notion.so/rilldata/Bignum-is-66-0-but-the-tooltip-value-is-in-decimal-form-196ab772a8eb48fdbcf45ac563bc6cdb


ericpgreen2 (approved):

ericpgreen2:
Approved, but looks like an e2e test needs addressing

PR #3605, opened 2023-12-01
User-friendly error instead of "Out of buffer" on ingesting TIME column from Snowflake

### LINKED ISSUE #3604: Out of buffer on ingesting TIME column from Snowflake
time32/64 types are [not supported](https://github.com/duckdb/duckdb/issues/1742) by DuckDB
Fail fast with a user-friendly error instead of "Out of buffer"

esevastyanov:
Closes #3604

### DIFF
--- a/runtime/drivers/snowflake/sql_store.go
+++ b/runtime/drivers/snowflake/sql_store.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"time"
 
+	"github.com/apache/arrow/go/v14/arrow"
 	"github.com/apache/arrow/go/v14/arrow/memory"
 	"github.com/apache/arrow/go/v14/parquet"
 	"github.com/apache/arrow/go/v14/parquet/compress"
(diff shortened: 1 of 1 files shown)


runtime/drivers/snowflake/sql_store.go
```diff
@@ -159,6 +160,11 @@ func (f *fileIterator) Next() ([]string, error) {
 
 	// common schema
 	schema := (*firstBatch)[0].Schema()
+	for _, f := range schema.Fields() {
+		if f.Type.ID() == arrow.TIME32 || f.Type.ID() == arrow.TIME64 {
+			return nil, fmt.Errorf("TIME type isn't supported, consider excluding it")
```
begelundmuller, line 165: Consider adding the column name as well, e.g. `"TIME data type (column %q) is not currently supported, consider excluding it or casting it to another data type"`

begelundmuller (approved):

PR #3606, opened 2023-12-01
add limit to comparison query in comparison toplist

pjain1:
Druid still has a calcite bug, without this limit clause calcite query planner fails. Will reopen the issue in OSS Druid. I guess limit does not hurt as anyways its applied to base query and dimension is already filtered to values returned from the base query so its basically harmless.


egor-ryashin:
Could you add the exception stacktrace here?

pjain1:
you mean druid exception ?

pjain1:
This is only needed for calcite planner to work, adding the limit is not needed for query correctness.

egor-ryashin (approved):

PR #3623, opened 2023-12-05
Fixing model table not updating after fixing the model sql

AdityaHegde:
## Checklist
- [x] Manual verification
- [ ] Unit test coverage
- [ ] E2E test coverage
- [x] Needs manual QA?

#### Issue addressed: 
Model's table is not being update after fixing the query

## Steps to Verify
1. Start typing the query,
```
select 1 as n
union
select 2 as n
```
2. While the 1st part is complete the table should display the value `1`
3. While typing the rest the model goes to error state.
4. After finishing typing the final part, the table should display both `1` and `2`

### DIFF
--- a/web-common/src/features/entity-management/resource-invalidations.ts
+++ b/web-common/src/features/entity-management/resource-invalidations.ts
@@ -103,13 +103,24 @@ async function invalidateResource(
 
   switch (resource.meta.name.kind) {
     case ResourceKind.Source:
-    // eslint-disable-next-line no-fallthrough
+      if (resource.source?.state?.table)
+        // make sure table is populated
+        return invalidateProfilingQueries(
+          queryClient,
+          resource.meta.name.name,
+          failed
+        );
+      break;
+
     case ResourceKind.Model:
-      return invalidateProfilingQueries(
-        queryClient,
-        resource.meta.name.name,
-        failed
-      );
+      if (resource.model?.state?.table)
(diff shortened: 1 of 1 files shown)


web-common/src/features/entity-management/resource-invalidations.ts
```diff
@@ -103,13 +103,24 @@ async function invalidateResource(
 
   switch (resource.meta.name.kind) {
     case ResourceKind.Source:
-    // eslint-disable-next-line no-fallthrough
+      if (resource.source?.state?.table)
+        // make sure table is populated
+        return invalidateProfilingQueries(
+          queryClient,
+          resource.meta.name.name,
```
ericpgreen2, line 110: I'm seeing some type issues in my editor: 
image

ericpgreen2 (approved):
QA'd and works 👍 

It feels like this isn't the first time an errored model has broken the app, so it'd be nice to extend the "Create and edit model" e2e test to catch this bug to prevent future regressions.

web-common/src/features/entity-management/resource-invalidations.ts
AdityaHegde: There are a bunch of errors like this across the app. These are because a setting was enabled to add strict mode and orval adds `?` to all fields even the ones that are not optional in the proto definition. This is one of those that can be ignored.

PR #3624, opened 2023-12-05
Adding support for defining filters on measures

AdityaHegde:
This is a backend only PR to support filters on measures

- [x] Define a proto structure for measure filter
- [x] Implement a generic buildHavingClause
- [x] Add to MetricsViewAggregation
- [x] Add to MetricsViewComparison
- [x] Add to MetricsViewToplist
- [x] Add to MetricsViewTimeSeries
- [x] Simplify proto structure to reuse column type from measure sort type
- [x] Look into using alias instead of measure name + column type
- [x] Performance consideration and query tweaks
- [x] Verify for druid


proto/rill/runtime/v1/expressions.proto
```diff
@@ -0,0 +1,35 @@
+syntax = "proto3";
+package rill.runtime.v1;
+
+import "google/protobuf/struct.proto";
+
+message Expression {
+  oneof expression {
+    google.protobuf.Value value = 1;
+    string column = 2;
```
begelundmuller, line 9: Name suggestion: `identifier` and make it 1 instead of 2

proto/rill/runtime/v1/expressions.proto
```diff
@@ -0,0 +1,35 @@
...
+    ConditionExpression condition_expression = 3;
+  }
+}
+
+message ConditionExpression {
```
begelundmuller, line 14: Suggestion: Just name it `Condition`

proto/rill/runtime/v1/queries.proto
```diff
@@ -8,6 +8,7 @@ import "rill/runtime/v1/export_format.proto";
 import "rill/runtime/v1/schema.proto";
 import "rill/runtime/v1/time_grain.proto";
 import "validate/validate.proto";
+import "expressions.proto";
```
begelundmuller, line 11: 1. Full path like above, i.e. `rill/runtime/v1/expressions.proto`
2. Singular, i.e. `expression.proto`

proto/rill/runtime/v1/expressions.proto
```diff
@@ -0,0 +1,35 @@
...
+}
+
+enum Operation {
+  OPERATION_UNSPECIFIED = 0;
+  OPERATION_EQUALS = 1;
+  OPERATION_NOT_EQUALS = 2;
+  OPERATION_LESSER = 3;
+  OPERATION_LESSER_OR_EQUALS = 4;
+  OPERATION_GREATER = 5;
+  OPERATION_GREATER_OR_EQUALS = 6;
```
begelundmuller, line 26: Is it more common to use "equals" or "equal"? 

Also, should we consider the usual shorthands, like `EQ`, `NEQ`, `LT`, `LTE`, etc.?

proto/rill/runtime/v1/expressions.proto
```diff
@@ -0,0 +1,35 @@
...
+  OPERATION_OR = 7;
+  OPERATION_AND = 8;
+  OPERATION_BETWEEN = 9;
+  OPERATION_NOT_BETWEEN = 10;
```
begelundmuller, line 30: Redundant – would suggest using an "and" of two expressions instead. It's also ambiguous – i.e. are they inclusive or exclusive of the start and end values?

proto/rill/runtime/v1/expressions.proto
```diff
@@ -0,0 +1,35 @@
+syntax = "proto3";
```
begelundmuller, line 1: Question: not sure what I think, but do you think we should consider shorter type names given the nesting of expressions and how often we'll end up looking at them printed?

I.e. using terms like `expr` (expression), `cond` (condition), `val` (value), `op` (operation), `args` (operands), `ident` (identifer), `eq` (equal), etc.?

For example, MongoDB does that: https://www.mongodb.com/docs/manual/reference/operator/query/expr/.

proto/rill/runtime/v1/queries.proto
```diff
@@ -279,7 +280,8 @@ message MetricsViewAggregationRequest {
   TimeRange time_range = 12;
   google.protobuf.Timestamp time_start = 6; // Deprecated in favor of time_range
   google.protobuf.Timestamp time_end = 7; // Deprecated in favor of time_range
-  MetricsViewFilter filter = 8;
+  Expression filter = 8;
+  repeated MetricsViewColumnAlias aliases = 13;
```
begelundmuller, line 284: Why are `aliases` needed in other cases than `MetricsViewComparison`? For the other cases, I think it's okay to require clients to use the real names (since there's no ambiguity around base vs. comparison vs. delta values).

proto/rill/runtime/v1/queries.proto
```diff
@@ -302,17 +304,41 @@ message MetricsViewAggregationMeasure {
...
+  string name = 1;
+  bool desc = 2;
+}
+
+message MetricsViewColumnAlias {
+  string name = 1;
+  oneof alias { // Is this overkill to future proof this?
+    MetricsViewMeasureAlias measure_alias = 2;
+  }
+}
```
begelundmuller, line 317: Yeah I think we should just flatten it. In light of the comment above, just one type `MetricsViewComparisonMeasureAlias` might be enough.

proto/rill/runtime/v1/queries.proto
```diff
@@ -302,17 +304,41 @@ message MetricsViewAggregationMeasure {
...
+
+message MetricsViewMeasureAlias {
+  enum MeasureType {
+    MEASURE_TYPE_UNSPECIFIED = 0;
+    MEASURE_TYPE_COUNT = 1;
+    MEASURE_TYPE_COUNT_DISTINCT = 2;
```
begelundmuller, line 323: These can already be defined with an alias through `MetricsViewAggregationMeasure`

proto/rill/runtime/v1/queries.proto
```diff
@@ -279,7 +280,8 @@ message MetricsViewAggregationRequest {
   TimeRange time_range = 12;
   google.protobuf.Timestamp time_start = 6; // Deprecated in favor of time_range
   google.protobuf.Timestamp time_end = 7; // Deprecated in favor of time_range
-  MetricsViewFilter filter = 8;
+  Expression filter = 8;
```
begelundmuller, line 283: Need to add support for `HAVING` clauses (different from `WHERE` clauses).

Maybe we refactor to `Expression having` and `Expression where`? Or to avoid that, could keep `Expression filter` and add `Expression having` or `Expression measure_filter`?

proto/rill/runtime/v1/expressions.proto
begelundmuller: I would also suggest looking out for prior art for expressions expressed in protobufs. Two good places to look are:
- https://github.com/googleapis/googleapis
- https://buf.build/explore

proto/rill/runtime/v1/expressions.proto
AdityaHegde: These are from duckdb docs: https://duckdb.org/docs/sql/expressions/comparison_operators.html#between-and-is-not-null
I think we can convert them to compound operations in UI and keep the API simple.

proto/rill/runtime/v1/queries.proto
AdityaHegde: The idea was to use the same measure in `measures` and `filter`. Aggregation allows for count and count distinct so I thought those could be defined in aliases and used in `measures` and `filter` fields.

proto/rill/runtime/v1/queries.proto
AdityaHegde: Ya i think separating will keep it cleaner. Combined filter was not clean in the code.

AdityaHegde: Also `having` makes more sense. We can think of using `where` for the main filter in the near future.

proto/rill/runtime/v1/queries.proto
```diff
@@ -408,6 +412,21 @@ message MetricsViewTimeSeriesResponse {
...
+    MEASURE_TYPE_COMPARISON_VALUE = 2;
+    MEASURE_TYPE_ABS_DELTA = 3;
+    MEASURE_TYPE_REL_DELTA = 4;
+  }
+
+  string name = 1;
+  MeasureType type = 2;
+  repeated google.protobuf.Value args = 3;
+  string alias = 4;
+}
```
begelundmuller, line 428: 1. Would incorporate `Comparison` in the name since this is specific to that use case (no other API supports base/comparison/delta values)
2. Should reuse (and probably rename) the `MetricsViewComparisonSortType` type so we don't have two enums doing the same

proto/rill/runtime/v1/queries.proto
```diff
@@ -342,7 +344,9 @@ message MetricsViewComparisonRequest {
   repeated MetricsViewComparisonSort sort = 5;
   TimeRange time_range = 6;
   TimeRange comparison_time_range = 7;
-  MetricsViewFilter filter = 8;
+  Condition where = 8;
+  Condition having = 14;
```
begelundmuller, line 348: Consider using `Expression` directly, simpler to have one entrypoint (and matches the name of the file). Also slightly more flexible in that you can put something like just `WHERE true`

runtime/queries/metricsview_comparison_toplist.go
```diff
@@ -153,6 +166,7 @@ func (q *MetricsViewComparison) executeComparisonToplist(ctx context.Context, ol
 		return fmt.Errorf("error building query: %w", err)
 	}
 
+	fmt.Println(sql, args)
```
begelundmuller, line 169: Remove

runtime/queries/metricsview.go
```diff
@@ -292,6 +292,327 @@ func buildFilterClauseForCondition(mv *runtimev1.MetricsViewSpec, cond *runtimev
 	return fmt.Sprintf("AND (%s) ", condsClause), args, nil
 }
 
+type identifier struct {
+	mapped string
+	unnest bool
+}
```
begelundmuller, line 298: 1. Since so many different things are "identified" in these queries, perhaps call it `columnIdentifier` instead of `identifier`?
2. Maybe change `mapped` to `safeName` or `sqlName` to clarify what it contains?

runtime/queries/metricsview.go
```diff
@@ -292,6 +292,327 @@ func buildFilterClauseForCondition(mv *runtimev1.MetricsViewSpec, cond *runtimev
...
+		arg, err := pbutil.FromValue(e.Val)
+		if err != nil {
+			return "", emptyArg, err
+		}
+		return "?", []any{arg}, nil
+
+	case *runtimev1.Expression_Ident:
+		col, ok := allowedIdentifiers[e.Ident]
+		if !ok {
+			return "", emptyArg, fmt.Errorf("unknown column filter. %s", e.Ident)
```
begelundmuller, line 325: nit: use `:` instead of `.`

runtime/queries/metricsview.go
```diff
@@ -292,6 +292,327 @@ func buildFilterClauseForCondition(mv *runtimev1.MetricsViewSpec, cond *runtimev
...
+	}
+	args = append(args, subArgs...)
+
+	notKeyword := ""
+	if cond.Op == runtimev1.Operation_OPERATION_NLIKE {
+		notKeyword = "NOT"
+	}
+
+	// identify if immediate identifier has unnest
+	// TODO: do we need to do a deeper check?
```
begelundmuller, line 388: Not sure, but I don't think so? If we wanted to stay simple, might even be okay to make the assumption that the first arg to an operation is an ident

runtime/queries/metricsview.go
```diff
@@ -292,6 +292,327 @@ func buildFilterClauseForCondition(mv *runtimev1.MetricsViewSpec, cond *runtimev
...
+	unnest := false
+	ident, isIndent := cond.Exprs[0].Expression.(*runtimev1.Expression_Ident)
```
begelundmuller, line 390: `isIndent` -> `isIdent`

runtime/queries/metricsview.go
```diff
@@ -292,6 +292,327 @@ func buildFilterClauseForCondition(mv *runtimev1.MetricsViewSpec, cond *runtimev
...
+		}
+	}
+
+	// When you have "dim NOT ILIKE '...'", then NULL values are always excluded.
+	// We need to explicitly include it.
+	if cond.Op == runtimev1.Operation_OPERATION_NLIKE {
+		clause += fmt.Sprintf(" OR %s IS NULL", leftExpr)
+	}
+
+	// TODO: is `col is null` needed?
```
begelundmuller, line 415: If it was in the previous implementation, it probably still is. Maybe `git blame` can help trace why it was added? (Good example of why it's nice to add comments for non-obvious code)

runtime/queries/metricsview.go
```diff
@@ -292,6 +292,327 @@ func buildFilterClauseForCondition(mv *runtimev1.MetricsViewSpec, cond *runtimev
...
+	exclude := cond.Op == runtimev1.Operation_OPERATION_NIN
+	if exclude {
+		notKeyword = "NOT"
+	}
+
+	inHasNull := false
+	var valClauses []string
+	// Add to args, skipping nulls
+	for _, subExpr := range cond.Exprs[1:] {
+		// TODO: is a deeper check needed?
```
begelundmuller, line 440: Probably not

runtime/queries/metricsview.go
```diff
@@ -292,6 +292,327 @@ func buildFilterClauseForCondition(mv *runtimev1.MetricsViewSpec, cond *runtimev
...
+	case runtimev1.Operation_OPERATION_LT:
+		return "<"
+	case runtimev1.Operation_OPERATION_LTE:
+		return "<="
+	case runtimev1.Operation_OPERATION_GT:
+		return ">"
+	case runtimev1.Operation_OPERATION_GTE:
+		return ">="
+	}
+	return "=" // TODO: handle unknown operation type
```
begelundmuller, line 529: Maybe panic on unknown type and add a `[(validate.rules).enum.defined_only = true]` validation to the proto requests?

runtime/queries/metricsview.go
```diff
@@ -292,6 +292,327 @@ func buildFilterClauseForCondition(mv *runtimev1.MetricsViewSpec, cond *runtimev
...
+	return identifier{safeName(name), false}
+}
+
+func dimensionAliases(mv *runtimev1.MetricsViewSpec) map[string]identifier {
+	aliases := map[string]identifier{}
+	for _, dim := range mv.Dimensions {
+		aliases[dim.Name] = identifier{safeName(metricsViewDimensionColumn(dim)), dim.Unnest}
+	}
+	return aliases
+}
```
begelundmuller, line 310: The previous `buildFilterClauseForMetricsViewFilter` implementation avoided building a lookup map since the overhead of building it (allocating/assigning a ~20 element map, plus a similar number of allocations for `safeName`) would be much higher than just iterating over `mv.Dimensions` a few times (most filters apply to just one or two dimensions).

Always hard to balance optimization vs. convenience, and not sure how much this matters, but would consider staying with the format of `buildFilterClauseForMetricsViewFilter` (i.e. passing `mv` and aliases directly as args; and maybe use a boolean like `isHaving` to indicate which to use for lookups).

runtime/queries/metricsview.go
```diff
@@ -292,6 +292,327 @@ func buildFilterClauseForCondition(mv *runtimev1.MetricsViewSpec, cond *runtimev
...
+
+func buildFromExpression(expr *runtimev1.Expression, allowedIdentifiers map[string]identifier, dialect drivers.Dialect) (string, []any, error) {
```
begelundmuller, line 312: `buildFromExpression` sounds to me like "build `FROM` expression" – consider renaming to `buildExpression` or something like that

runtime/queries/filterutil.go
```diff
@@ -0,0 +1,88 @@
+package queries
+
+import (
+	"github.com/rilldata/rill/proto/gen/rill/runtime/v1"
+	"google.golang.org/protobuf/types/known/structpb"
+)
+
+func FilterColumn(col string) *runtimev1.Expression {
```
begelundmuller, line 8: Consider moving these utils to `runtime/pkg/expressionpb` and making the names match the `Expression` type (e.g. `expressionpb.NotIn` instead of `FilterNotInClause`).

begelundmuller (changes requested):

runtime/queries/metricsview.go
AdityaHegde: It was needed. I added the code but didnt remove the TODO. Explanation in the comments above the if clause above this.

runtime/queries/metricsview.go
AdityaHegde: It more like expression. Updated to be explicit.

runtime/queries/metricsview.go
AdityaHegde: Ya makes sense. We should really look at having some global cache per metrics view. Things would be very clean with a map. But for now i have updated to use mv and aliases

runtime/queries/metricsview_comparison_toplist.go
```diff
@@ -6,15 +6,14 @@ import (
...
+	_ "time/tzdata"
 
 	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
 	"github.com/rilldata/rill/runtime"
 	"github.com/rilldata/rill/runtime/drivers"
 	"github.com/rilldata/rill/runtime/pkg/pbutil"
 	"google.golang.org/protobuf/types/known/structpb"
-
-	// Load IANA time zone data
-	_ "time/tzdata"
```
begelundmuller: Should be restored – underscore imports go in a third category (not sure why the linter stopped enforcing this)

runtime/queries/metricsview.go
```diff
@@ -26,6 +26,8 @@ import (
 	"google.golang.org/protobuf/types/known/structpb"
 )
 
+var emptyMeasureAliases = make([]*runtimev1.MetricsViewComparisonMeasureAlias, 0)
```
begelundmuller, line 29: Not needed, can just pass `nil` because `range` and `len` are valid on a `nil` slice or map.

runtime/queries/metricsview_comparison_toplist.go
```diff
@@ -1030,27 +1026,22 @@ func isTimeRangeNil(tr *runtimev1.TimeRange) bool {
...
 
-func aliasesToMeasureMap(aliases []*runtimev1.MetricsViewComparisonMeasureAlias, measureAliases map[string]columnIdentifier, hasComparison bool) error {
+func validateMeasureAliases(aliases []*runtimev1.MetricsViewComparisonMeasureAlias, measures []*runtimev1.MetricsViewAggregationMeasure, hasComparison bool) error {
 	for _, alias := range aliases {
+		// We need to make sure the aliases measure is selected. Having query will fail otherwise
+		if !slices.ContainsFunc(measures, func(measure *runtimev1.MetricsViewAggregationMeasure) bool {
+			return measure.Name == alias.Name
+		}) {
+			return fmt.Errorf("measure alias not selected: %s", alias.Name)
+		}
```
begelundmuller, line 1055: Maybe redundant? IMO it's okay to pass an alias that's not used in the expression. And fine to just fail at query time if using an expression that's not selected.

runtime/queries/metricsview_comparison_toplist.go
```diff
@@ -425,13 +459,25 @@ func (q *MetricsViewComparison) buildMetricsComparisonTopListSQL(mv *runtimev1.M
...
+				)
+			} else {
+				columnsTuple = fmt.Sprintf(
+					"base.%[1]s, comparison.%[1]s AS %[2]s, base.%[1]s - comparison.%[1]s AS %[3]s, (base.%[1]s - comparison.%[1]s)/comparison.%[1]s::DOUBLE AS %[4]s",
+					safeName(m.Name),
+					safeName(m.Name+"__previous"),
+					safeName(m.Name+"__delta_abs"),
+					safeName(m.Name+"__delta_rel"),
+				)
+			}
```
begelundmuller, line 480: Is this needed because the actual aggregation happens in a sub-query? If yes, then instead of adding a new layer of aggregations, couldn't we just use a `WHERE` clause in the outer select instead? (Since a `HAVING` clause is basically equivalent to a `WHERE` clause in a wrapped query, or am I missing something?)

If the aggregation is needed, consider using `ANY_VALUE` or `LAST`, which I believe we use for such cases in other places (less ambiguous than `SUM`).

begelundmuller (changes requested):

runtime/queries/metricsview_comparison_toplist.go
AdityaHegde: Good point. I have updated to use an outer where clause.

begelundmuller (approved):

runtime/queries/metricsview_comparison_toplist_test.go
```diff
@@ -19,12 +19,18 @@ import (
 	_ "github.com/rilldata/rill/runtime/drivers/duckdb"
 )
 
+// replace with "ad_bids_druid" to test on druid
+const AdBidsProject = "ad_bids"
+
+// replace with "timestamp" to test on druid
+const AdBidsTimestamp = "timestamp"
```
begelundmuller, line 26: The test scaffolding around the analytical APIs is already pretty complex, so I don't think having constants to comment/uncomment in multiple different files is the way to add Druid support to them.

It would be really nice to add Druid support, but I think we should do it in a separate PR and figure out a proper test abstraction that would allow us to test both Druid and DuckDB in CI/CD. We should also have Druid started using Testcontainers similar to the test cases in `runtime/drivers`.

runtime/queries/metricsview_timeseries_test.go
```diff
@@ -14,8 +14,11 @@ import (
 	"google.golang.org/protobuf/types/known/structpb"
 )
 
+// replace with "timeseries_druid" to test on druid
+const TimeseriesProject = "timeseries"
```
begelundmuller, line 18: See other comment

runtime/testruntime/testruntime.go
```diff
@@ -31,6 +31,12 @@ import (
 	_ "github.com/rilldata/rill/runtime/reconcilers"
 )
 
+// testConnector replace with "druid" for testing on druid
+const testConnector = "duckdb"
+
+// testConnectorDSN replace with "http://localhost:8888/druid/v2/sql/avatica-protobuf/" for testing on druid
+const testConnectorDSN = ""
```
begelundmuller, line 38: See other comment

begelundmuller (changes requested):

runtime/queries/metricsview_comparison_toplist_test.go
AdityaHegde: Got it.

runtime/queries/metricsview.go
```diff
@@ -305,8 +305,12 @@ func columnIdentifierExpression(mv *runtimev1.MetricsViewSpec, aliases []*runtim
...
 		if alias.Alias == name {
 			switch alias.Type {
-			case runtimev1.MetricsViewComparisonMeasureType_METRICS_VIEW_COMPARISON_MEASURE_TYPE_BASE_VALUE,
-				runtimev1.MetricsViewComparisonMeasureType_METRICS_VIEW_COMPARISON_MEASURE_TYPE_UNSPECIFIED:
+			case runtimev1.MetricsViewComparisonMeasureType_METRICS_VIEW_COMPARISON_MEASURE_TYPE_UNSPECIFIED,
+				runtimev1.MetricsViewComparisonMeasureType_METRICS_VIEW_COMPARISON_MEASURE_TYPE_BASE_VALUE:
+				// using `measure_0` as is causing ambiguity error in duckdb
+				if dialect == drivers.DialectDuckDB {
+					return "base." + safeName(alias.Name), true
+				}
```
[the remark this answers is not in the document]
AdityaHegde: Added an outer query for duckdb to get rid of the ambiguity.

begelundmuller (approved):

PR #3627, opened 2023-12-05
Runtime: Reduce chance and impact of deadlocked DuckDB calls

begelundmuller:
- Ensures `LongRunning=true` for all mutation queries
- Stop the reconcile controller and log an error if a cancelled reconcile does not finish within 10 minutes
- Log an error from DuckDB driver if a connection has been held for more than one hour
- Avoid implicit connection creation in `detachAllDBs`
- Correct usage of `reopenDB`

Contributes to https://github.com/rilldata/rill-private-issues/issues/17

### DIFF
--- a/runtime/drivers/duckdb/duckdb.go
+++ b/runtime/drivers/duckdb/duckdb.go
@@ -811,6 +811,7 @@ func (c *connection) periodicallyCheckConnDurations(d time.Duration) {
 					c.logger.Error("duckdb: a connection has been held for more longer than the maximum allowed duration", zap.Int("conn_id", connID), zap.Duration("duration", time.Since(connTime)))
 				}
 			}
+			c.connTimesMu.Unlock()
 		}
 	}
 }
--- a/runtime/drivers/duckdb/duckdb.go
+++ b/runtime/drivers/duckdb/duckdb.go
@@ -2,6 +2,7 @@ package duckdb
 
 import (
 	"context"
+	databasesql "database/sql"
 	"database/sql/driver"
 	"errors"
 	"fmt"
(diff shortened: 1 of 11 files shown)


runtime/drivers/duckdb/duckdb.go
```diff
@@ -765,16 +792,42 @@ func (c *connection) periodicallyEmitStats(d time.Duration) {
...
+const maxAcquiredConnDuration = 1 * time.Hour
+
+// periodicallyCheckConnDurations periodically checks the durations of all acquired connections and logs a warning if any have been held for longer than maxAcquiredConnDuration.
+func (c *connection) periodicallyCheckConnDurations(d time.Duration) {
+	connDurationTicker := time.NewTicker(d)
+	defer connDurationTicker.Stop()
+	for {
+		select {
+		case <-c.ctx.Done():
+			connDurationTicker.Stop()
```
k-anshul, line 805: Nit: the defer on line 801 should ensure ticket is stopped?

runtime/drivers/duckdb/duckdb.go
```diff
@@ -788,7 +841,7 @@ func (c *connection) detachAllDBs() {
 		}
 
 		db := dbName(entry.Name(), version)
-		_, err = c.db.ExecContext(context.Background(), fmt.Sprintf("DETACH %s", safeSQLName(db)))
+		_, err = conn.ExecContext(ctx, fmt.Sprintf("DETACH %s", safeSQLName(db)))
```
k-anshul, line 844: Should we also add some timeout to `detach` with fatal log for timeout?

begelundmuller: Good idea, added a timeout for the close. But skipping for the reopen for now since not quite sure how long certain checkpoints can take during normal operation.

Also not sure if a go-side timeout will matter much, since if DuckDB hangs it tends not to yield back control to Go to check the cancellation....

k-anshul (approved):

PR #3634, opened 2023-12-05
Remove Babel from `web-local`

ericpgreen2:
As far as I know, our Babel dependency in `web-local` is a relic of the past. This PR removes it.


AdityaHegde (approved):
What even is this from? Jest?

PR #3639, opened 2023-12-06
Only apply `overscroll-behavior: none` to Rill Developer

### LINKED ISSUE #3574: Trackpad gesture navigation doesn't work on Google Chrome
In Google Chrome, in both Rill Cloud and Rill Developer, swiping left on my trackpad does not "go back" one page.

This trackpad gesture works in Safari, and it does work for me in Chrome on other websites.

ericpgreen2:
`overscroll-behavior: none` is CSS that prevents trackpad gesture navigation (given it's enabled at your OS level). Previously, this CSS lived in our `app.css` file, which applies to both Rill Cloud and Rill Developer. This PR move the CSS so that it only applies to Rill Developer.

Code editors, or at least vscode.dev, tend to restrict this sort of trackpad navigation so that horizontally scrolling your code file does not accidentally trigger navigation.

Closes #3574

### DIFF
--- a/web-common/src/app.css
+++ b/web-common/src/app.css
@@ -160,8 +160,6 @@ body {
   --error-bg: hsl(1, 60%, 90%);
   --error-text: hsl(1, 60%, 30%);
 
-  overscroll-behavior: none;
-
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
 }
--- a/web-common/src/layout/RillDeveloperLayout.svelte
+++ b/web-common/src/layout/RillDeveloperLayout.svelte
@@ -103,3 +103,10 @@
 </div>
 
 <NotificationCenter />
+
+<style>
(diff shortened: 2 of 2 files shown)


briangregoryholmes:
This works for me as expected on Chrome. Is it important enough that we should fire an event `beforeUnload` so that Safari users get a warning if this is done accidentally?

briangregoryholmes (approved):

ericpgreen2:
> This works for me as expected on Chrome. Is it important enough that we should fire an event `beforeUnload` so that Safari users get a warning if this is done accidentally?

Oh, fair callout. Currently, when a user attempts to navigates away from an unsaved source, we do have this dialog: 
image

It might be a little rabbit hole to reconcile two competing `beforeUnload` dialogs, so I'll leave the current behavior for now. In this PR, I mainly wanted to enable trackpad navigation on Rill Cloud.

PR #3647, opened 2023-12-07
Svelte check ignore cleanup 3

bcolloran:
more svelte check --ignore cleanups

part of  #3424


ericpgreen2 (approved):
Nice chunk of cleanups 👍

PR #3649, opened 2023-12-07
limit `--ignore` in web-local to only dev route!

bcolloran:
We can declare victory in `web-local`: the `src/routes/dev` folder can stay ignored since it's not user facing. After this pr we won't be ignoring any user facing type errors in that subfolder of the repo.

part of  #3424


ericpgreen2 (approved):
💪

PR #3651, opened 2023-12-07
svletecheck --ignore cleanup: fix most of "time-controls" folder

bcolloran:
part of https://github.com/rilldata/rill/issues/3424

@djbarnwal -- assigning this review to you since you've worked on this stuff recently (I've also assigned #3678  to you to wrap this folder up). Thanks!


web-common/src/features/entity-management/resource-status-utils.ts
```diff
@@ -106,12 +106,17 @@ export function resourceStatusStore(
       )
         return { status: ResourceStatus.Busy };
 
+      const changed =
+        !lastUpdatedOn ||
+        (res.data?.meta?.stateUpdatedOn
+          ? res.data?.meta?.stateUpdatedOn > lastUpdatedOn
+          : undefined);
+
```
djbarnwal, line 114: Not sure if we want to use `undefined` here. @ericpgreen2 would have more context.

djbarnwal (approved):
Thanks for the PR. Looks good for most of the changes. I am not sure what are implications of `changed` as `undefined`. Approving for all the other changes.

web-common/src/features/entity-management/resource-status-utils.ts
ericpgreen2: I'm not so familiar with this code (more @AdityaHegde), but it looks like it's fine as long as we protect against accessing an `undefined` attribute. And it looks like @bcolloran is doing so on line 154. After a quick look, line 154-155 is the only place where I see this `changed` attribute being used.

bcolloran: good catch @djbarnwal, thank you. The current code is 
`!lastUpdatedOn || res.data?.meta?.stateUpdatedOn > lastUpdatedOn`, and `res.data?.meta?.stateUpdatedOn` is `string | undefined`. Since `undefined > x  => false` for all `x: string`, we should default to false here.

bcolloran: (but that said we also return undefined for `changed` from this function on line 107 right above this: `return { status: ResourceStatus.Busy }`)

PR #3668, opened 2023-12-08
Fix text wrapping in the Imported Source dialog

ericpgreen2:
Before: With short source names, "successfully" was often wrapped mid-word. With long source names, the header was hard to scan quickly.
image

After: This PR moves the source name out of the dialog's header and into its body.
image

image


briangregoryholmes:
Works for me on a few different test names.

My instinct is that including the project name within this message isn't strictly necessary. When this modal appears, the project name is visible in two places elsewhere on the page. I think it might make sense for this to simply always say "Source imported successfully" without the name. The project name could be included on the next line in a lighter/smaller font. I think a confirmation message like this should be parseable with a really quick glance and the line break makes me take an extra second.

briangregoryholmes (approved):

ericpgreen2:
> This is minor, but my instinct is that including the project name within this message isn't strictly necessary. When this modal appears, the project name is visible in two places elsewhere on the page. I think it might make sense for this to simply always say "Source imported successfully" without the name. The project name could be included on the next line in a lighter/smaller font if necessary. I think a confirmation message like this should be parseable with a really quick glance and the line break makes me take an extra second.

Good idea – just incorporated. I did include the source name in the body of the dialog – because the dialog opens asynchronously, whenever the source finishes ingest, and if it's a long ingest, the user may have moved on to a different page.

PR #3669, opened 2023-12-08
make formatMsInterval tighter and add tests

### LINKED ISSUE #3655: Fix bug in `formatMsInterval` function
[As reported in Slack](https://rilldata.slack.com/archives/CTZ8XBQ85/p1701985216349899)

bcolloran:
This should fix #3655 -- this code is now as systematic as I know how to be in JS. Literally every everything value used in the function is either a literal of a defined type, a predefined constant field access, or a number, and any non-numeric input should be handle as well.

I was not able to reproduce the actual bug locally, so I can't 100% confirm that this will fix it, but it is now impossible to access the `unit` field on a possibly undefined value, because there is no more accessing of fields with runtime values.

### DIFF
--- a/web-common/src/lib/number-formatting/strategies/intervals.spec.ts
+++ b/web-common/src/lib/number-formatting/strategies/intervals.spec.ts
@@ -1,0 +1,150 @@
+import { describe, it, expect } from "vitest";
+import { formatMsInterval } from "./intervals";
+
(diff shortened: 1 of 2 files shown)


ericpgreen2 (approved):
Nice. We can merge this, then test the triggering dashboard on the staging environment to confirm.

PR #3670, opened 2023-12-08
adding unplugin-icons as icon manager

briangregoryholmes:
The intent of this PR is to simplify the current approach of creating individual Svelte components for application icons by instead using `unplugin-icons` as a management layer. In this approach, the majority of icons can be used immediately and directly by simply referencing the icon name from Figma.

For instance, the design system references some icons in the Ant Design library: https://icones.js.org/collection/ant-design

Instead of exporting/copying the SVG data from Figma, creating a new component file in the icons folder, deciding on a name, and then importing that where necessary, you could now do the following:

```svelte
<script>
    import StepForward from "~icons/ant-design/step-forward-outline"
</script>

<StepForward class="text-red-400"/>
```

Icons can be customized using Tailwind using the `text` utility classes.

While importing collections like Ant Design or Radix installs the full icon set, only the icons we actually use will be bundle into the production build

- Adds and configures the [unplugin-icons ](https://github.com/unplugin/unplugin-icons) package
- Updates the sidebar footer to use this new approach (as an example) and makes small tweaks to better match the Figma designs


briangregoryholmes:
Created new branch based on the Svelte 4 migration.

PR #3671, opened 2023-12-08
support for Add Filter button

### LINKED ISSUE #3620: Support for Add Filter button
Details here in Notion - https://www.notion.so/rilldata/Add-filter-button-cc49b174280c4dd6accb17731679e6a7?d=117128ee4700403b9ce29e373d7081da

briangregoryholmes:
Adds an "Add Filter" button to the Filters component. Also updates relevant menu designs.

Closes #3620


web-local/test/ui/dashboards.spec.ts
```diff
@@ -245,6 +245,8 @@ test.describe("dashboard", () => {
     // Filter to Facebook via leaderboard
     await page.getByRole("button", { name: "Facebook 19.3k" }).click();
 
+    await page.waitForTimeout(100);
```
AdityaHegde, line 248: Hmmm we should add a smarter wait. This arbitrary wait of 100ms might not actually wait long enough in all cases.

AdityaHegde (commented):
A few UX questions,
1. Should a check mark be shown and the dimension be disabled in add filter if it is already present in the filters?
2. Or should the dimension filter dropdown open if an existing dimension is selected in add filter drop down?

web-common/src/features/dashboards/selectors/index.ts
```diff
@@ -73,7 +73,6 @@ export const getFilterSearchList = (
           measureNames: [metricsExplorer.leaderboardMeasureName],
           timeStart: timeControls.timeStart,
           timeEnd: timeControls.timeEnd,
-          limit: "15",
```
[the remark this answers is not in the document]
briangregoryholmes: This was removed because of the design changes to the dropdown. Previously, the drop down only displayed the already selected values plus a search bar that would bring up more. The new dropdown ostensibly shows the full list of available options and even provides an opportunity to select or deselect all.

web-common/src/features/dashboards/state-managers/actions/dimension-filters.ts
```diff
@@ -33,7 +33,7 @@ export function toggleDimensionValueSelection(
   } else {
```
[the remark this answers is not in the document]
briangregoryholmes: Added a `toggleDimensionNameSelection` method

web-local/test/ui/dashboards.spec.ts
briangregoryholmes: Changed to wait for the specific selector

briangregoryholmes:
New commits up now. Regarding your UX question, I think the best option is to simply remove the already selected filters from that dropdown. The existing filters wrap onto a new line (and so are always visible) and the button is specifically for adding a filter (rather than removing or navigating to), so I think it's better UX to only present valid options. Good callout. Let me know what you think.

web-common/src/features/dashboards/selectors/index.ts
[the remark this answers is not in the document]
AdityaHegde: Hmmm would probably be better to have an explicit limit of 100. Otherwise a frontend engineer will need to look at go code to know the limit.

web-common/src/features/dashboards/state-managers/actions/dimension-filters.ts
```diff
@@ -38,6 +38,30 @@ export function toggleDimensionValueSelection(
   }
 }
 
+export function toggleDimensionNameSelection(
```
AdityaHegde, line 41: I am not sure if this is a toggle anymore. The above one either adds or removes existing one. This just adds an entry. Not a 100% sure what is a good name though.

AdityaHegde (approved):
Just a couple of small nits. Looks good otherwise.

web-common/src/features/dashboards/state-managers/actions/dimension-filters.ts
briangregoryholmes: Made it a toggle

AdityaHegde (approved):

PR #3684, opened 2023-12-12
cleanup svelte-check --ignore for  data-graphics

bcolloran:
note: full text search shows that `<Density` and `<ECDF` does not show up any where in our codebase. They include both mistyped values as well as an imports that are unavailable in our repo, so I think they either haven't been used for a long time, or may never have actually been adopted. Deleted these under the "you ain't gonna need it" principle.

part of #3424


web-common/src/components/data-graphic/marks/Line.svelte
```diff
@@ -62,7 +62,6 @@
       yScale: $yScale,
       curve,
       xAccessor,
-      pathDefined,
```
bcolloran, line 65: note: no need to pass `pathDefined = pathDoesNotDropToZero`, because `pathDoesNotDropToZero` is the dafault when this argument is omitted

web-common/src/components/data-graphic/marks/Density.svelte
bcolloran, line 1: note: full text search shows that `<Density` does not show up any where in our codebase. This file has both mistyped values as well as an import that is unavailable in our repo, so I think it hasn't been used for a long time, or it may never have actually been adopted. deleting it under the "you ain't gonna need it" principle

web-common/src/components/data-graphic/constants/types.d.ts
```diff
@@ -1,4 +1,4 @@
 export interface DomainCoordinates {
-  x: number | Date;
-  y: number | Date;
+  x?: number;
+  y?: number;
```
bcolloran, line 3: note that
(1) I could not find any instances of Date being used in this type (it always seems to be used in numeric operations, but please double check me on this)
(2) these fields are very commonly set to undefined by way of the `DEFAULT_COORDINATES` const which is used all over the place

```
export const DEFAULT_COORDINATES: DomainCoordinates = {
  x: undefined,
  y: undefined,
};
```

web-common/src/components/data-graphic/marks/Density.svelte
bcolloran, line 1: file never used

web-common/src/components/data-graphic/marks/ECDF.svelte
bcolloran, line 1: file never used

web-common/src/components/date-picker/Litepicker.svelte
```diff
@@ -251,12 +251,20 @@
...
+    /* expansion of: @apply text-gray-50 */
+    --tw-text-opacity: 1;
+    color: rgb(249 250 251 / var(--tw-text-opacity));
   }
 
   :global(.litepicker .container__tooltip:after) {
-    @apply border-t-gray-700;
+    /* expansion of: @apply border-t-gray-700; */
+    --tw-border-opacity: 1;
+    border-top-color: rgb(55 65 81 / var(--tw-border-opacity));
```
bcolloran, line 267: expanding these tailwind styles to clear a warning about use of  @apply

bcolloran:
hey @djbarnwal , I have another review for you. I think it should all be pretty straightforward, but the one place where your expertise on this part of the codebase will be a big help is double checking whether "Date" is ever actually used in the DomainCoordinate type. Thanks!

web-common/src/components/date-picker/Litepicker.svelte
ericpgreen2: I recommend `<style lang="postcss">`, which should remove these warnings.

ericpgreen2 (approved):
Looks good 💪. I only recommend making the `lang="postcss"` change before merging.

web-common/src/components/date-picker/Litepicker.svelte
bcolloran: Oooh thanks, great tip!! Didn't know we could do that in svelte

PR #3686, opened 2023-12-12
horizontal splitter overflow

briangregoryholmes:
This PR prevents the horizontal splitter from overflowing its parent container. It also updates the ModelBody and ModelWorkspace components to better match their Source counterparts.

I believe the feel and responsiveness of this interaction can be much improved by streamlining the data flow between the HorizontalSplitter component and whatever element it is intended to adjust. I would also suggest removing the tweening, which ends up feeling more like lag than a designed interaction.

This PR just includes the relatively basic adjustments needed to solve the underlying issue, but does refactor the HorizontalSplitter component in a way that I believe is easier to follow (while solving some issues present on Safari).


djbarnwal (approved):
Looks great! Thanks for simplifying the splitter component 💪

PR #3688, opened 2023-12-12
update to leaderboard test to account for inconsistent Playwright behavior

briangregoryholmes:
This PR fixes an issue that occurs when running tests locally (and sometimes in CI).

Re: `leaderboard-and-dim-table-sort.spec.ts`

1. Adds a page timeout after selecting the `Percentage change` sorting option to allow for the Leaderboard rankings to update
2. Broadens the subsequent selectors by using RegEx rather than a specific string. The reason for this is that some Playwright runs trigger the hover effect on the element, which means that the "name" selector may sometimes include an additional number (see below)

Screenshot 2023-12-12 at 6 35 39 PM

If similar issues arise in the future, I think the other selectors could be broadened without issue. Also, if these are too broad, for whatever reason, this PR could be updated with a slightly more complex RegEx to capture both hover and not hovered states.


ericpgreen2 (approved):
Great, test passes for me locally now 👍

PR #3690, opened 2023-12-13
Runtime: otel fix for db close leaks

k-anshul:
1. Fixes issue in `otelsql` which prevents duckDB `close` calls from being called properly which should hopefully fix OOM restarts.
2. Upgrades otel to `v0.26.0`
3. Removes extra code for `detach` all dbs before main duckDB handle `close`. Main handle can be closed without detaching attached DBs.


begelundmuller (approved):
LGTM – thanks!

PR #3691, opened 2023-12-13
Runtime: change reconcile and parse errors to warns

begelundmuller:
This PR changes error logs that can occur during normal usage to warning level. Examples of such errors include reconcile errors and revoked Github access.

By emitting these logs at warning level instead of error level, we reserve the error level for internal errors that are worth alerting on.


k-anshul (approved):
Just confirming if these msgs will be available to the user in local mode. Else need to use `logger.Named` for some messages like `Reconcile failed` and `Parser error` ?

begelundmuller:
> Just confirming if these msgs will be available to the user in local mode. Else need to use `logger.Named` for some messages like `Reconcile failed` and `Parser error` ?

Confirmed – these are still shown in the console

PR #3696, opened 2023-12-14
remove artificial time series chart delay

### LINKED ISSUE #3381: Time series charts render need not wait for totals API call is resolved
[Notion](https://www.notion.so/rilldata/2023-06-01-application-bug-reports-5c22a93c17a84546948ebcfe3cddefa1)

The time series charts sometimes render much later than the time it takes to resolve the API call. In the video below, it’s about a second, but in some cases, it’s much worse.

While the time series API calls do have a higher priority (30 instead of 10), it looks like the charts aren’t rendered until the totals have also resolved, and they only run at priority 10 (competing with toplists).

Is it possible to both:

- increase the priority of the totals API call, and
- do a partial render of the time series charts as soon as either the time series or total API calls have resolved (instead of waiting for them all)

briangregoryholmes:
Closes #3381 

Note: I did not have access to the dataset being used in the referenced Notion doc, so had to artificially recreate this situation while debugging.

The issue seems to be that TimeSeriesDataStore was artificially holding back data and returning `isFetching` as `true` until both the totals and primary queries were available.

During exploration, I saw no obvious UI issues related to rendering either of those pieces of data independently. As such, my proposed fix is to simply change the boolean logic of the `isFetching` property and remove the stricter "gate". This appears to solve the issue.

This PR also updates the `totals` query to a priority of 30.

### DIFF
--- a/web-common/src/features/dashboards/time-series/timeseries-data-store.ts
+++ b/web-common/src/features/dashboards/time-series/timeseries-data-store.ts
@@ -159,16 +159,6 @@ export function createTimeSeriesDataStore(ctx: StateManagers) {
         ]) => {
           let timeSeriesData = primary?.data?.data;
 
-          if (!primary?.data || !primaryTotal?.data || !unfilteredTotal?.data) {
-            return {
-              isFetching:
-                metricsView.isFetching ||
-                primary?.isFetching ||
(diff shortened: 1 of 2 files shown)


ericpgreen2 (approved):
Nice 💪

PR #3699, opened 2023-12-14
docs: hide content class

mindspank:
Adds short-hand css classes to hide pages and categories from documentation.


begelundmuller (approved):
LGTM

PR #3700, opened 2023-12-14
Runtime: unified `duckdb` connector for `motherduck` and external `db` files

### LINKED ISSUE #3642: Add generic `external DuckDB` source connector
High level tasks:

1. Define a source that points to an existing external duckdb database.
2. Also merge it with motherduck connector 

High level solution:

We will have three fields in duckdb connector :
```
DB --  either "md:..." or "path to local file"
SQL -- SQL query to run against db
NAME
```

More details mentioned here : https://rilldata.slack.com/archives/CTZ8XBQ85/p1701296087640469

k-anshul:
closes #3642

### DIFF
--- a/runtime/drivers/duckdb/transporter_duckDB_to_duckDB.go
+++ b/runtime/drivers/duckdb/transporter_duckDB_to_duckDB.go
@@ -117,11 +117,11 @@ func (t *duckDBToDuckDB) transferFromExternalDB(ctx context.Context, srcProps *d
 		var localDB, localSchema string
 		for res.Next() {
 			if err := res.Scan(&localDB, &localSchema); err != nil {
-				res.Close()
+				_ = res.Close()
 				return err
 			}
 		}
-		res.Close()
+		_ = res.Close()
 
 		// duckdb considers everything before first . as db name
 		// alternative solution can be to query `show databases()` before and after to identify db name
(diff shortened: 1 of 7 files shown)


runtime/drivers/duckdb/transporter_duckDB_to_duckDB.go
```diff
@@ -90,16 +97,64 @@ func (t *duckDBToDuckDB) Transfer(ctx context.Context, srcProps, sinkProps map[s
...
+		if err != nil {
+			return err
+		}
+		defer res.Close()
+
+		res.Next()
+		var localDB, localSchema string
+		if err := res.Scan(&localDB, &localSchema); err != nil {
+			return err
+		}
```
begelundmuller, line 122: This seems dangerous – I think `t.to` will be locked until `res.Close()` has run, but since it's deferred, it won't run until the end of the func. Not exactly sure why this doesn't deadlock, but might be because there's some logic around `Next`/`Scan` auto-closing a `rows` if scanned to the end.

Would suggest the following changes:
- Consume the rows the usual way, i.e. using `for res.Nex()` – there will be just one result, so no harm in it.
- Explicitly call `res.Close` afterwards and in the error cases

begelundmuller (changes requested):

begelundmuller (approved):

PR #3702, opened 2023-12-14
[Dashboard] Support arbitrary non-aggregate expressions in dimensions

### LINKED ISSUE #3009: Runtime: Investigate changing `column` to `expression` for dashboard dimensions
- [ ] Investigate if we can support arbitrary dimension expressions
- [ ] Change the YAML field `column` for dimensions to `expression` (keep backwards compatibility)
- [ ] If `expression` is not set, it should default to the dimension `name` (this is the same behavior as DBT)
- [ ] Update metrics APIs to alias dimension columns to the dimension `name` (oversight in earlier PRs)
- [ ] Update the frontend to always use the `dimension` name instead of `column`

AdityaHegde:
closes #3009

- [x] Add `expression` to dashboard yaml.
- [x] Update all queries and test for dimension expression.
- [x] Support unnest with dimension expressions.
- [x] Test on driud.
- [x] Introduce proper validation.


runtime/compilers/rillv1/parse_metrics_view.go
```diff
@@ -238,12 +239,6 @@ func (p *Parser) parseMetricsView(ctx context.Context, node *Node) error {
 			return fmt.Errorf("found duplicate dimension or measure name %q", dim.Name)
 		}
 		names[lower] = true
-
-		lower = strings.ToLower(dim.Column)
-		if ok := columns[lower]; ok {
-			return fmt.Errorf("found duplicate dimension column name %q", dim.Column)
-		}
-		columns[lower] = true
```
begelundmuller, line 246: If removing this, should also remove the `columns := make(map[string]bool)` on line 216 and other refs to it

runtime/queries/metricsview.go
```diff
@@ -297,7 +297,7 @@ func columnIdentifierExpression(mv *runtimev1.MetricsViewSpec, aliases []*runtim
 	// check if identifier is a dimension
 	for _, dim := range mv.Dimensions {
 		if dim.Name == name {
-			return safeName(metricsViewDimensionColumn(dim)), true
+			return metricsViewDimensionExpression(dim), true
```
begelundmuller, line 300: This will get templated into the where clause:

1. Might it be problematic to put a free form expression in here? I think at the least it will need to be wrapped in parentheses
2. Would it be better to template in just the dimension name (which the expression is aliased to in the `SELECT` clause)? But not sure if that works on Druid

runtime/queries/metricsview.go
```diff
@@ -696,6 +712,14 @@ func metricsViewDimensionColumn(dimension *runtimev1.MetricsViewSpec_DimensionV2
 	return dimension.Name
 }
 
+func metricsViewDimensionExpression(dimension *runtimev1.MetricsViewSpec_DimensionV2) string {
+	if dimension.Expression != "" {
+		return dimension.Expression
+	}
+	// if there is expression treat as a column reference
+	return safeName(metricsViewDimensionColumn(dimension))
```
begelundmuller, line 720: If possible, let's inline `metricsViewDimensionColumn` into `metricsViewDimensionExpression` to not have too much ambiguity

runtime/queries/metricsview.go
```diff
@@ -340,6 +340,22 @@ func identifierIsUnnest(mv *runtimev1.MetricsViewSpec, expr *runtimev1.Expressio
 	return false
 }
 
+func dimensionSelect(mv *runtimev1.MetricsViewSpec, dim *runtimev1.MetricsViewSpec_DimensionV2, dialect drivers.Dialect) (string, string) {
```
begelundmuller, line 343: Use named return variables to indicate what it returns  (or add a docstring about it)

runtime/reconcilers/metrics_view.go
```diff
@@ -152,6 +153,36 @@ func (r *MetricsViewReconciler) validate(ctx context.Context, mv *runtimev1.Metr
 	return errors.Join(errs...)
 }
 
+func validateDimension(
+	ctx context.Context,
+	olap drivers.OLAPStore,
+	t *drivers.Table,
+	d *runtimev1.MetricsViewSpec_DimensionV2,
+	fields map[string]*runtimev1.StructType_Field,
+) error {
```
begelundmuller, line 162: Collapse to one line similar to the style of other functions

runtime/reconcilers/metrics_view.go
```diff
@@ -152,6 +153,36 @@ func (r *MetricsViewReconciler) validate(ctx context.Context, mv *runtimev1.Metr
...
+	if d.Expression == "" {
+		if d.Column == "" {
+			return nil
+		}
+		if _, isColumn := fields[d.Column]; !isColumn {
+			return fmt.Errorf("column not found: %s", d.Column)
```
begelundmuller, line 168: Better error message – should include dimension name, something like `failed to validate dimension %q: column %q not found in table`

runtime/reconcilers/metrics_view.go
```diff
@@ -152,6 +153,36 @@ func (r *MetricsViewReconciler) validate(ctx context.Context, mv *runtimev1.Metr
...
+		}
+		return nil
+	}
+
+	err := olap.Exec(ctx, &drivers.Statement{
+		Query:  fmt.Sprintf("SELECT %s from %s GROUP BY 1", d.Expression, safeSQLName(t.Name)),
```
begelundmuller, line 174: 1. Is the `GROUP BY 1` approach supported by Druid?
2. nit: uppercase `FROM`

runtime/reconcilers/metrics_view.go
```diff
@@ -152,6 +153,36 @@ func (r *MetricsViewReconciler) validate(ctx context.Context, mv *runtimev1.Metr
...
+		DryRun: true,
+	})
+	if err == nil {
+		return nil
+	}
+	if strings.Contains(err.Error(), "GROUP BY clause cannot contain aggregates") {
+		return fmt.Errorf("dimension expression cannot have aggregates")
+	}
+	return fmt.Errorf("invalid expression for dimension %q: %w", d.Name, err)
```
begelundmuller, line 183: Would not hardcode a DuckDB error message here – consider just something like `failed to validate expression for dimension %q: %w`

begelundmuller (changes requested):

runtime/compilers/rillv1/parse_metrics_view.go
AdityaHegde: Ya removing them. It just didnt make sense anymore since we do not check duplicate of expression.

runtime/queries/metricsview.go
AdityaHegde: The problem with this is for queries where the dimension is not selected we would have issues with the column not being there. The other option is to only add it if is in the select caluse. That made these builder functions too complex.

As for adding parentheses, it gets added when the identifier is actually used.

I guess we are missing a validation to make sure there isnt sql injection happening.

runtime/queries/metricsview.go
AdityaHegde: Makes sense. There is one usage in older filters code which i will remove and merge these.

runtime/reconcilers/metrics_view.go
AdityaHegde: Group by 1 is working for druid, aggregates throw error as. expected.

runtime/queries/metricsview.go
AdityaHegde: Used named variables.

runtime/compilers/rillv1/parse_metrics_view.go
```diff
@@ -34,6 +34,7 @@ type MetricsViewYAML struct {
 		Name        string
 		Label       string
 		Column      string
+		Expression  string
```
begelundmuller, line 37: It looks like we don't have validation to ensure that only one of these are set (exactly one of them should be non-nil)

runtime/queries/metricsview.go
begelundmuller: 1. The case where the dim is not selected makes sense. Agree on not making it conditional on the select clause, seems to complicated.
2. About the parentheses, imagine if the expression is `col1 = 10` and the filter is `dim1 = true`, then you're saying the parenthesis will be added such that the where clause becomes `(col1 = 10) = true` (correct) and not `col1 = 10 = true` (which would error)?
3. How could SQL injection happen? For the where filters, I thought we were doing lookups to resolve identifiers and using `?` and args to pass values?

runtime/reconcilers/metrics_view.go
```diff
@@ -152,6 +153,27 @@ func (r *MetricsViewReconciler) validate(ctx context.Context, mv *runtimev1.Metr
 	return errors.Join(errs...)
 }
 
+func validateDimension(ctx context.Context, olap drivers.OLAPStore, t *drivers.Table, d *runtimev1.MetricsViewSpec_DimensionV2, fields map[string]*runtimev1.StructType_Field) error {
+	if d.Expression == "" {
+		if d.Column == "" {
+			return nil
```
begelundmuller, line 159: Shouldn't this be an error? (And ideally checked also at parser level)

runtime/reconcilers/metrics_view.go
```diff
@@ -152,6 +153,27 @@ func (r *MetricsViewReconciler) validate(ctx context.Context, mv *runtimev1.Metr
...
+		}
+		if _, isColumn := fields[d.Column]; !isColumn {
+			return fmt.Errorf("failed to validate dimension %q: column %q not found in table", d.Name, d.Column)
+		}
+		return nil
+	}
+
+	err := olap.Exec(ctx, &drivers.Statement{
+		Query:  fmt.Sprintf("SELECT %s FROM %s GROUP BY 1", d.Expression, safeSQLName(t.Name)),
```
begelundmuller, line 168: Should we put parentheses around the expression, i.e. `SELECT (%s) ...`? In case the expression is a sub-query

begelundmuller: Might this also be needed elsewhere?

begelundmuller (changes requested):

runtime/reconcilers/metrics_view.go
AdityaHegde: True. I guess we will never hit this without the parser running and dont need backwards compatibility right?

runtime/queries/metricsview.go
AdityaHegde: 2. Yes all binary expressions have it. `like` and `in` do not seem to have it. Lemme add that.
3. It could happen in the dimension expression, `count(*)";drop table blah; --` or something like that.

runtime/reconcilers/metrics_view.go
AdityaHegde: Yes adding them

begelundmuller (approved):
LGTM!

PR #3705, opened 2023-12-14
suggest models in the code editor

### LINKED ISSUE #3384: Suggest models in autocomplete
**Is your feature request related to a problem? Please describe.**

The autocomplete should also show modelnames.

**Describe the solution you'd like**

Models are shown in the drop-down of the autocomplete.

**Describe alternatives you've considered**

Not sure. Models could benefit from a slightly different color-formatting, as opposed to sources (?) 

**Additional context**

The autocomplete only suggests `tradesdownload` (a source). It should suggest `types` (a model) as well.
image

fwiw - the entire expression works great, it is just the auto-complete :) 

image

My apologies if this is already logged, or part of https://github.com/rilldata/rill/issues/265

Found in `version 0.36.7 – c13ed366`.

briangregoryholmes:
Closes #3384

Like the idea of coloring sources and models differently, but this PR does not address that feature.

### DIFF
--- a/web-common/src/features/models/selectors.ts
+++ b/web-common/src/features/models/selectors.ts
@@ -11,9 +11,16 @@ import {
   getRuntimeServiceListFilesQueryKey,
   runtimeServiceListFiles,
   V1ListFilesResponse,
+  V1ProfileColumn,
 } from "@rilldata/web-common/runtime-client";
 import type { QueryClient } from "@tanstack/query-core";
 import { TIMESTAMPS } from "../../lib/duckdb-data-types";
+import type { Readable } from "svelte/motion";
+import { derived } from "svelte/store";
+import {
+  createTableColumnsWithName,
+  type TableColumnsWithName,
+} from "../sources/selectors";
 
 export function useModels(instanceId: string) {
   return useFilteredResources(instanceId, ResourceKind.Model);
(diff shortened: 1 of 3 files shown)


ericpgreen2:
I started on QA, but I hit this error in my browser console, and it crashed the app:
image

I have a `select * from adbids` model, and I'm able to trigger the error when I refresh the page.

web-common/src/features/models/selectors.ts
```diff
@@ -31,6 +37,26 @@ export function useModel(instanceId: string, name: string) {
...
+    if (!allModels.data?.length) {
+      set([]);
+      return;
+    }
+
+    derived(
+      allModels.data.map((r) =>
+        createTableColumnsWithName(queryClient, instanceId, r.meta.name.name)
+      ),
+      (sourceColumnResponses) =>
```
ericpgreen2, line 54: Ah, this tipped me off to where you got this code :). Nice approach to just mirror what we have for sources. But: `sourceColumnResponses` -> `modelColumnResponses`

briangregoryholmes:
@ericpgreen2 I missed that `schema` was technically undefined until it hit the reactive code block that populated the source columns.

I moved the empty object assignment to the variable declaration, which resolved the issue on my end.

web-common/src/features/models/selectors.ts
briangregoryholmes: Renamed

ericpgreen2 (approved):
Works great!

Issue #3443, opened 2023-11-09
break out web code quality CI steps


bcolloran:
To fail faster and improve parallelism of CI, code quality checks for web can run in a separate CI job from the other tests, similar to how we currently have a separate prettier job in addition to the main web tests.

bcolloran:
updating this issue to limit scope to breaking up the tests, since https://www.notion.so/rilldata/Improving-our-e2e-tests-5aabdba1b6e243d680f4cf923565f1c3 tracks a bunch of e2e test improvements separately

bcolloran:
closed via #3706

PR #3706, opened 2023-12-14
split web tests

bcolloran:
Splits web tests up into:
- code quality (lint and type check)
- unit tests (we apparently only have unit tests in web-common and web-auth, though we have not been running the auth tests in CI, so those are added)
- e2e tests (only web-local and web-admin have this, though web-admin only seems to have a test stub for now)

Note that we have a separate GH action that runs prettier on the entire repo (prettier.yaml) so I thought it was redundant to have the prettier checks for each web folder. Easy to add back in if that is incorrect.

Looking to @himadrisingh for approval from the dev-ops perspective since I'm not super familiar with the ins-and-outs of GH actions. In particular, I don't know what the "dorny/paths-filter@v2" does, so I just copied that over as-is across the board. Please LMK if that need to be changed or can be simplified

But also fyi @ericpgreen2 @djbarnwal @AdityaHegde @briangregoryholmes to have extra eyes on this.

Partially addresses #3443, though we may want to look into parallelizing (or otherwise speeding up) the e2e at some point.

### DIFF
--- a/.github/workflows/web-test-code-quality.yml
+++ b/.github/workflows/web-test-code-quality.yml
@@ -21,22 +21,15 @@ jobs:
           filters: |
             admin:
               - ".github/workflows/web-test.yml"
-              - "admin/**"
-              - "cli/**"
-              - "runtime/**"
               - "web-admin/**"
             auth:
               - ".github/workflows/web-test.yml"
               - "web-auth/**"
             local:
               - ".github/workflows/web-test.yml"
-              - "cli/**"
-              - "runtime/**"
               - "web-local/**"
             common:
               - ".github/workflows/web-test.yml"
-              - "cli/**"
-              - "runtime/**"
               - "web-common/**"
 
       - name: Set up NodeJS
--- a/.github/workflows/web-test-code-quality.yml
+++ b/.github/workflows/web-test-code-quality.yml
@@ -1,11 +1,8 @@
-name: Build and Test web code
+name: Web code quality checks
 on:
   pull_request:
     paths:
       - ".github/workflows/web-test.yml"
-      - "admin/**"
-      - "cli/**"
-      - "runtime/**"
       - "web-admin/**"
       - "web-auth/**"
       - "web-common/**"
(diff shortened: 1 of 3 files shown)


bcolloran:
fixes #3443

.github/workflows/web-test-unit-tests.yml
```diff
@@ -0,0 +1,60 @@
...
+      - name: Filter modified codepaths
+        uses: dorny/paths-filter@v2
+        id: filter
+        with:
+          filters: |
+            admin:
+              - ".github/workflows/web-test.yml"
+              - "admin/**"
+              - "cli/**"
+              - "runtime/**"
```
ericpgreen2, line 26: Do the unit tests hit the runtime at all? We don't need to run them when `cli` or `runtime` code is changed, right?

.github/workflows/web-test-unit-tests.yml
```diff
@@ -0,0 +1,60 @@
...
+              - "web-admin/**"
+            auth:
+              - ".github/workflows/web-test.yml"
+              - "web-auth/**"
+            local:
+              - ".github/workflows/web-test.yml"
+              - "cli/**"
+              - "runtime/**"
```
ericpgreen2, line 34: See comment above

.github/workflows/web-test-unit-tests.yml
```diff
@@ -0,0 +1,60 @@
...
+              - "web-local/**"
+            common:
+              - ".github/workflows/web-test.yml"
+              - "cli/**"
+              - "runtime/**"
```
ericpgreen2, line 39: See comment above

.github/workflows/web-test-unit-tests.yml
```diff
@@ -0,0 +1,60 @@
...
+
+      - name: unit test web-common
+        if: steps.filter.outputs.common == 'true'
+        run: npm run test -w web-common
+
+      - name: Test web-auth
+        if: steps.filter.outputs.auth == 'true'
+        run: npm run test -w web-auth
+
+      - name: Make sure storybook is working
```
ericpgreen2, line 58: This name threw me off a little bit. What about "Run storybook smoketest"

ericpgreen2: And to have symmetry, the other names could be "Run web-common unit tests" and "Run web-auth unit tests"

.github/workflows/web-test-code-quality.yml
```diff
@@ -0,0 +1,73 @@
...
+      - name: Filter modified codepaths
+        uses: dorny/paths-filter@v2
+        id: filter
+        with:
+          filters: |
+            admin:
+              - ".github/workflows/web-test.yml"
+              - "admin/**"
+              - "cli/**"
+              - "runtime/**"
```
ericpgreen2, line 26: We don't need the `web` code quality tests to run when `cli` or `runtime` code changes

.github/workflows/web-test-code-quality.yml
```diff
@@ -0,0 +1,73 @@
...
+              - "web-admin/**"
+            auth:
+              - ".github/workflows/web-test.yml"
+              - "web-auth/**"
+            local:
+              - ".github/workflows/web-test.yml"
+              - "cli/**"
+              - "runtime/**"
```
ericpgreen2, line 34: Same comment as above

.github/workflows/web-test-code-quality.yml
```diff
@@ -0,0 +1,73 @@
...
+              - "web-local/**"
+            common:
+              - ".github/workflows/web-test.yml"
+              - "cli/**"
+              - "runtime/**"
```
ericpgreen2, line 39: Same comment as above

ericpgreen2:
It looks like `.prettierignore` lists pretty much all files/directories that aren't in our `web` workspaces. I'd be in favor of moving the standalone prettier check into the web code quality workflow.

bcolloran:
Thanks for the review @ericpgreen2. I removed all those extra run conditions -- I was being super timid and conservative in not taking them out :-)

I think the prettier job may pick up a handful of things that fall outside of the web code quality checks. If it's ok with you, I'd rather deal with that separately rather than rolling that audit of prettier coverage into this PR.

@himadrisingh -- please take a look at this and let us know if we're missing anything. Thank you!

himadrisingh (approved):

PR #3712, opened 2023-12-18
Runtime: Support template resolution in source properties

### LINKED ISSUE #3203: Runtime: Support template resolution in source properties
The source YAML properties are parsed into a map, which is passed directly to the transporter in [runtime/reconcilers/source.go#L385](https://github.com/rilldata/rill/blob/04724335f8baa4082d8839b686cb3ce8e78325a5/runtime/reconcilers/source.go#L385-L386).

We want to be able to use templating in source properties. For example:
```yaml
kind: source
connector: s3
path: "s3:://mybucket/{{ .env.my_tenant }}/*.parquet"
```

That requires us to traverse the source properties map and resolve templates for any string value, similar to what we do for the model SQL in [runtime/reconcilers/model.go#L410](https://github.com/rilldata/rill/blob/04724335f8baa4082d8839b686cb3ce8e78325a5/runtime/reconcilers/model.go#L410).

The template resolution must happen at reconcile time, not at parse time, since the value of variables can be changed dynamically (and may be sensitive, i.e. not to be persisted as part of the resource data in the catalog).

k-anshul:
closes #3203


runtime/reconcilers/source.go
```diff
@@ -308,7 +309,8 @@ func (r *SourceReconciler) setTriggerFalse(ctx context.Context, n *runtimev1.Res
 // ingestSource ingests the source into a table with tableName.
 // It does NOT drop the table if ingestion fails after the table has been created.
 // It will return an error if the sink connector is not an OLAP.
-func (r *SourceReconciler) ingestSource(ctx context.Context, src *runtimev1.SourceSpec, tableName string) (outErr error) {
+func (r *SourceReconciler) ingestSource(ctx context.Context, self *runtimev1.Resource, tableName string) (outErr error) {
+	src := self.Resource.(*runtimev1.Resource_Source).Source.GetSpec()
 	// Get connections and transporter
```
begelundmuller, line 314: nits:
1. Use `self.GetSource().Spec`
2. Put newline before the next section comment

runtime/reconcilers/source.go
```diff
@@ -382,8 +384,21 @@ func (r *SourceReconciler) ingestSource(ctx context.Context, src *runtimev1.Sour
...
+	props, err = resolveTemplateProps(ctx, props, rillv1.TemplateResource{
+		Meta:  self.Meta,
+		Spec:  spec,
+		State: state,
+	}, r.C)
+	if err != nil {
+		return nil, err
+	}
+
 	return props, nil
```
begelundmuller, line 402:
```suggestion
	tself := rillv1.TemplateResource{
		Meta:  self.Meta,
		Spec:  self.GetSource().Spec,
		State: self.GetSource().State,
	}

	return resolveTemplateProps(ctx, propsPB.AsMap(), tself, r.C)
```

runtime/reconcilers/util.go
```diff
@@ -154,3 +155,48 @@ func safeSQLName(name string) string {
 	}
 	return fmt.Sprintf("\"%s\"", strings.ReplaceAll(name, "\"", "\"\""))
 }
+
+func resolveTemplateProps(ctx context.Context, props map[string]any, self compilerv1.TemplateResource, c *runtime.Controller) (map[string]any, error) {
```
begelundmuller, line 159: The name sounds like it makes props for templating. Would also move contextual args to earlier in the arg list. Suggestion: `resolveTemplatedProps(ctx, c, self, props)`

runtime/reconcilers/util.go
```diff
@@ -154,3 +155,48 @@ func safeSQLName(name string) string {
...
+	}
+	vars := inst.ResolveVariables()
+
+	for key, value := range props {
+		strValue, ok := value.(string)
+		if !ok {
+			continue
+		}
+
+		strValue, err = compilerv1.ResolveTemplate(strValue, compilerv1.TemplateData{
```
begelundmuller, line 172: Two issues:
1. It doesn't traverse nested properties (nested lists/maps)
2. It creates a new `compilerv1.TemplateData` on each iteration – can just be created once

begelundmuller (changes requested):

runtime/reconcilers/util.go
k-anshul: Looked at the internal implementation of `propsPB.AsMap()` and maps and lists seems to be of type `map[string]any` and `[]any`. Adding specific handling for these types to avoid `reflection`.

begelundmuller (approved):
LGTM

PR #3719, opened 2023-12-19
Adding CONTINUOUS to comparison backwards compatibility code in url state loader

### LINKED ISSUE #3718: Cannot read properties of undefined (reading 'offsetIso')
[As reported in Slack](https://rilldata.slack.com/archives/C05693FLY7Q/p1702940524947689)

It looks like the triggering code is in `web-common/src/lib/time/comparisons/index.ts`'s `getAvailableComparisonsForTimeRange()` and/or `getComparisonTransform()`.

AdityaHegde:
closes #3718

### DIFF
--- a/web-common/src/features/dashboards/proto-state/fromProto.ts
+++ b/web-common/src/features/dashboards/proto-state/fromProto.ts
@@ -193,6 +193,9 @@ function correctComparisonTimeRange(
   comparisonTimeRange: DashboardTimeControls
 ) {
   switch (comparisonTimeRange.name as string) {
+    case "CONTIGUOUS":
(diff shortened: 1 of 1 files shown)


ericpgreen2:
I see `offsetIso` can also be blank for `TimeComparisonOption.CUSTOM`...
image

... but I don't see an entry for `CUSTOM` in `fromProto.ts`. Should there be one?

AdityaHegde:
Custom didnt change. Also we do not use offsetIso for CUSTOM and CONTINUOUS so that is fine.

ericpgreen2 (approved):

Issue #3648, opened 2023-12-07
Introduce backend flag to indicate that a deployed project is in an actual production state


AndrewRTsao:
We should implement a `has SLA` / `high priority` flag that can be set for projects via `rill sudo` to indicate that a deployed project is actually in production / business critical. This will help us determine which projects require active monitoring from our side so we can build more robust alerting / self-healing without introducing too much noise / signal either. Additionally, this should be emitted as part of log attributes (so we can alert / filter on it).

AndrewRTsao:
FYI @begelundmuller @kaspersjo

PR #3725, opened 2023-12-19
CLI: ability to check health of all projects for support users

rakeshsharma14317:


begelundmuller (changes requested):
Unfortunately this PR relies on the `projects.status` and `projects.status_message` fields to determine health, which is not really useful since these are only used for provisioning status, not reconcile status. 

Reconcile status and errors are reported at a resource-level not project-level (e.g. one source may have ingested successfully while another failed), and need to be fetched from the runtime of each deployment separately by calling `ListResources`. See the implementation of `rill project status` [here](https://github.com/rilldata/rill/blob/main/cli/cmd/project/status.go).

I think it would make more sense to implement this "health" check as follows:
1. Do a search for projects using the same API as `rill sudo project search`. Maybe even make it part of that CLI command and toggle it with a `rill sudo project search --status` flag.
2. Concurrently, request runtime details + JWT for each of the returned projects, and call `ListResources` for it.
3. Output a table containing something like the following columns for each project:
    - Count of idle resources
    - Count of idle resources with errors
    - Count of pending/running resources
    - Count of parser errors
    - Bool value for whether the health check is responsive or not (depends on https://github.com/rilldata/rill-private-issues/issues/31)
    - Bool value for provisioning error (i.e. if the project's status is not `DEPLOYMENT_STATUS_OK`)

Regarding the initial issue for this PR, I would propose limiting the initial PR to the search capabilities of `rill sudo project search`, i.e. org and project search (not user, domain, or status type search – these are trickier to resolve correctly and potentially less useful).

I would also propose completing this issue first: https://github.com/rilldata/rill/issues/3648. Combined with the above suggestions, it would make it easy to get the status of all projects with an SLA attached.

AndrewRTsao:
> Combined with the above suggestions, it would make it easy to get the status of all projects with an SLA attached.

AndrewRTsao:
> Combined with the above suggestions, it would make it easy to get the status of all projects with an SLA attached.

Agreed with this. I think the main utility of this command will be for support members to quickly check whether a triggered alert is customer impacting and in production, which warrants higher SLAs and elevated urgency / severity from our side to address. I'm not sure if it was implied in this response, while I do agree it makes sense to limit scope on the search capabilities (especially if the expanded syntax proves tricky / non-trivial to implement), we should at least be able to pass in the `has_SLA` flag as part of the search criteria (e.g. give me the project health of all production projects or production projects belonging to a specific customer).

cli/cmd/sudo/project/search.go
```diff
@@ -50,6 +57,89 @@ func SearchCmd(ch *cmdutil.Helper) *cobra.Command {
 				return err
 			}
 
+			if status {
```
begelundmuller, line 60: The names printing on line 55 should not be done when printing a status table

cli/cmd/sudo/project/search.go
```diff
@@ -50,6 +57,89 @@ func SearchCmd(ch *cmdutil.Helper) *cobra.Command {
...
+					if err != nil {
+						return fmt.Errorf("failed to list resources: %w", err)
+					}
+
+					var parser *runtimev1.ProjectParser
+					var ParserErrorCount int32
+					var IdleCount int32
+					var IdleWithErrorsCount int32
+					var PendingCount int32
+					var RunningCount int32
```
begelundmuller, line 95: 1. Not external variables, so should start with lowercase
2. Can just use `int` (instead of `int32`)

cli/cmd/sudo/project/search.go
```diff
@@ -50,6 +57,89 @@ func SearchCmd(ch *cmdutil.Helper) *cobra.Command {
...
 			}
 
+			if status {
+				var table []*projectStatusTableRow
+				ch.Printer.Println()
+				for _, name := range res.Names {
+					org := strings.Split(name, "/")[0]
+					project := strings.Split(name, "/")[1]
+
+					proj, err := client.GetProject(ctx, &adminv1.GetProjectRequest{
```
begelundmuller, line 67: Extract lines 67 to 123 to a util function, something like `func newProjectStatusTableRow(ctx, client, org, project) (*projectStatusTableRow, error)`

cli/cmd/sudo/project/search.go
```diff
@@ -58,9 +148,20 @@ func SearchCmd(ch *cmdutil.Helper) *cobra.Command {
...
 	searchCmd.Flags().StringSliceVar(&tags, "tag", []string{}, "Tags to filter projects by")
 	searchCmd.Flags().Uint32Var(&pageSize, "page-size", 50, "Number of projects to return per page")
 	searchCmd.Flags().StringVar(&pageToken, "page-token", "", "Pagination token")
 
 	return searchCmd
 }
+
+type projectStatusTableRow struct {
+	Name                string `header:"name"`
+	Organization        string `header:"organization"`
```
begelundmuller, line 161: Just make the header `org` to save space

cli/cmd/sudo/project/search.go
```diff
@@ -58,9 +148,20 @@ func SearchCmd(ch *cmdutil.Helper) *cobra.Command {
...
+	IdelCount           int32  `header:"idle"`
+	IdelWithErrorsCount int32  `header:"idle with errors"`
```
begelundmuller, line 163: Typo: should be `Idle` not `Idel`

cli/cmd/sudo/project/search.go
```diff
@@ -58,9 +148,20 @@ func SearchCmd(ch *cmdutil.Helper) *cobra.Command {
...
+	PendingCount        int32  `header:"pending"`
+	RunningCount        int32  `header:"running"`
+	ParserErrorCount    int32  `header:"parser error"`
```
begelundmuller, line 166: Make header `parser errors` (plural)

cli/cmd/devtool/data/cloud-deps.docker-compose.yml
```diff
@@ -31,14 +31,14 @@ services:
...
       - ./otel-collector-config.yaml:/etc/otel-collector-config.yaml
     ports:
-      - "1888:1888"   # pprof extension
-      - "8889:8889"   # Prometheus exporter metrics
+      - "1888:1888" # pprof extension
+      - "8889:8889" # Prometheus exporter metrics
       - "13133:13133" # health_check extension
-      - "4317:4317"   # OTLP gRPC receiver
+      - "4317:4317" # OTLP gRPC receiver
       - "55679:55679" # zpages extension
```
begelundmuller, line 42: Revert changes here (since it's not related to this PR)

begelundmuller (changes requested):

cli/cmd/devtool/data/cloud-deps.docker-compose.yml
rakeshsharma14317: this is done by linter run i guess, and if revert it i.e remove spaces from `"1888:1888" # pprof extension` in this show compile time error

begelundmuller (changes requested):

begelundmuller (approved):

PR #3728, opened 2023-12-19
move incomplete filters out of dashboard store and into a dedicated piece of state

briangregoryholmes:
This PR moves temporary/potential/incomplete filters (those that do not have dimension values) outside of the dashboard store.

This addresses a change in the runtime that happened in #3624  that requires the `in` field of a filter be a nonempty array.


ericpgreen2:
Code looks good 👍

Approving, but one UXQA nit [captured in this Jam](https://jam.dev/c/39b79edb-411b-449e-9a0a-76775b354264) to fix before merging:
- I add one filter.
- I start to add a subsequent filter and select a new dimension to filter on.
- The dimension values from the prior filter flash before rendering the dimension values of the new dimension.

ericpgreen2 (approved):

Issue #3373, opened 2023-11-03
Add `strictNullChecks` to `tsconfig.json` everywhere


ericpgreen2:

bcolloran:
currently, we have tsconfig files that set `"strictNullChecks": false` in the following directories:

- [ ] /home/bc/rill/rill-developer/web-admin/src/features/
- [ ] /home/bc/rill/rill-developer/web-admin/src/routes/
- [ ] /home/bc/rill/rill-developer/web-common/src/features/models/
- [ ] /home/bc/rill/rill-developer/web-common/src/features/sources/

Going forward, we'll want to remove those tsconfig files, push the `"strictNullChecks": false`  farther and farther down into subdirectories.

(I'll try to keep this updated, but remember that you can always do a full project text search for `"strictNullChecks": false` to see where these configs are currently set)

bcolloran:
Reassigning to you to finish up `web-admin` @ericpgreen2 . Carry the torch for me! :-)

PR #3729, opened 2023-12-19
more strictNullChecks

bcolloran:
part of #3373


web-common/src/layout/overlay-store.ts
```diff
@@ -7,4 +7,4 @@ interface Overlay {
   message?: string;
 }
 
-export const overlay = writable<Overlay>(null);
+export const overlay = writable<Overlay | null>(null);
```
bcolloran, line 10: This is explicitly set to null in web-common/src/features/sources/workspace/SourceWorkspaceHeader.svelte, so this type is needed

web-common/src/components/column-profile/ColumnProfile.svelte
```diff
@@ -28,33 +28,28 @@
...
-  );
+  $: profileColumns =
+    objectName === undefined
+      ? undefined
+      : createQueryServiceTableColumns(
+          $runtime?.instanceId,
+          objectName,
+          {},
+          { query: { keepPreviousData: true } }
+        );
```
bcolloran, line 39: here and below we just propagate `undefined` if objectName===undefined, because these fns aren't designed to handle undefined input. We then just need to add one more undefined check in the html template and all is good

web-common/src/features/entity-management/resource-selectors.ts
```diff
@@ -119,7 +119,7 @@ export function createSchemaForTable(
   ) as ReturnType<typeof createConnectorServiceOLAPGetTable>;
 }
 
-export function resourceIsLoading(resource: V1Resource) {
+export function resourceIsLoading(resource?: V1Resource) {
```
bcolloran, line 122: already checks for undefined input right below, just wasn't typed correctly

bcolloran:
@ericpgreen2 , picked off strictNullChecks in `sources`!

web-common/src/components/column-profile/ColumnProfile.svelte
ericpgreen2: Here we should make use of [TanStack Query's `enabled` flag for dependent queries](https://tanstack.com/query/v4/docs/react/guides/dependent-queries)

```
{ query: { keepPreviousData: true, enabled: !!objectName } }
```

Then we can get rid of these ternaries. Will that work?

web-common/src/features/sources/saveAndRefresh.ts
```diff
@@ -4,7 +4,10 @@ import { runtime } from "../../runtime-client/runtime-store";
 import { getFileAPIPathFromNameAndType } from "../entity-management/entity-mappers";
 import { EntityType } from "../entity-management/types";
 
-export async function saveAndRefresh(tableName: string, yaml: string) {
+export async function saveAndRefresh(
+  tableName: string,
+  yaml: string | undefined
```
ericpgreen2, line 9: `yaml` should always be string, right? If it happens to be `undefined`, can we disable the Button instead?

web-common/src/features/sources/workspace/SourceWorkspaceHeader.svelte
```diff
@@ -115,13 +115,18 @@
   }
 
   const onSaveAndRefreshClick = async (tableName: string) => {
+    // no-op if the clientYAML is still null
+    if ($sourceStore.clientYAML === null) return;
```
ericpgreen2, line 119: Similar to above. If `clientYAML` is null, can we instead disable the Button so this code path never gets hit?

web-common/src/features/sources/modal/SourceImportedModal.svelte
```diff
@@ -24,6 +27,7 @@
   }
 
   async function autoGenerateDashboard() {
+    if (dashboardFromSourceUIAction === null) return;
```
ericpgreen2, line 30: Another case – can we instead disable the Button?

web-common/src/features/sources/errors/ErrorPane.svelte
```diff
@@ -7,7 +7,7 @@
   import { useSourceFromYaml } from "../selectors";
 
   export let sourceName: string;
-  export let errorMessage: string;
+  export let errorMessage: string | undefined;
```
ericpgreen2, line 10: Can we instead only show the `ErrorPane` if an error message exists?

web-common/src/features/entity-management/name-utils.ts
```diff
@@ -37,10 +37,12 @@ export function getNextEntityName(
 }
 
 export function isDuplicateName(
-  name: string,
-  fromName: string,
-  names: Array<string | undefined>
+  name: string | null,
+  fromName: string | null,
+  names: Array<string | undefined> | undefined
 ) {
```
ericpgreen2, line 43: This function should really only be called when these parameters are non-null

ericpgreen2 (approved):
Left a variety of comments, but will approve and you can address as you see fit

web-common/src/components/column-profile/ColumnProfile.svelte
bcolloran: found another solution -- just pushed the check to upstream components so that <ColumnProfile> is only rendered if `objectName` is a string, in which case this file doesn't need to be changed at all

web-common/src/features/sources/errors/ErrorPane.svelte
bcolloran: Yep!

web-common/src/features/sources/modal/SourceImportedModal.svelte
bcolloran: disabled the button, but we still need this type check, because the function scope doesn't know about the temporal order of the events. added a comment explaining.

web-common/src/features/sources/workspace/SourceWorkspaceHeader.svelte
```diff
@@ -115,13 +115,18 @@
...
     overlay.set({ title: `Importing ${tableName}.yaml` });
     await saveAndRefresh(tableName, $sourceStore.clientYAML);
     checkSourceImported(queryClient, sourceName, filePath);
     overlay.set(null);
   };
 
   const onRefreshClick = async (tableName: string) => {
+    // no-op if connector is undefined
+    if (connector === undefined) return;
```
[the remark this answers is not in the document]
bcolloran: we can disable the button, but we can't take out this check because of the way the component is set up. In practice, I think this code path will never actually be hit anyway. This whole thing is ugly anyway (see eg the try/catch with a no-op directly below) and could use a refactor, but I'm gonna make the call that that's out of scope for this PR

PR #3732, opened 2023-12-20
Runtime: error instead of panic when creating models on Druid

begelundmuller:
Prevents the panics observed in these [logs](https://rill.datadoghq.com/logs?query=service%3Arill-cloud%20%40error%3A%2A%20%40msg%3A%22panic%20in%20reconciler%22&agg_m=count&agg_q=kube_namespace%2Chost%2C%40env&agg_t=count&cols=host%2Cservice&index=main&messageDisplay=inline&refresh_mode=sliding&source=monitor_notif&stream_sort=desc&view=spans&viz=stream&from_ts=1703075402000&to_ts=1703075702000&live=false)

### DIFF
--- a/runtime/drivers/druid/olap.go
+++ b/runtime/drivers/druid/olap.go
@@ -13,32 +13,32 @@ var _ drivers.OLAPStore = &connection{}
 
 // AddTableColumn implements drivers.OLAPStore.
 func (c *connection) AddTableColumn(ctx context.Context, tableName, columnName, typ string) error {
-	panic("not implemented")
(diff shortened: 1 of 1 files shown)


k-anshul (approved):
Do we want to raise errors instead of panics for developer errors ?

begelundmuller:
> Do we want to raise errors instead of panics for developer errors ?

Yeah we do, but in this case a user can trigger this case by adding a model to a Druid-backed project. So for user errors, we don't want to panic.

PR #3734, opened 2023-12-20
debounce update model when typing in code editor

briangregoryholmes:


ericpgreen2 (approved):
Looks good 👍

Only thing is I see the `<Editor />` component has some old debouncing no-ops and there's an unused `models/utils/Debounce.ts` file in the codebase. I like your approach better! Mind removing the old code before merging?

PR #3736, opened 2023-12-20
Introduce ShadCN UI via the `DropdownMenu` component set

ericpgreen2:
This PR adds [ShadCN-Svelte](https://www.shadcn-svelte.com/) to the repository, starting with the `DropdownMenu` component set.

Specifically, I've:
- Followed the ShadCN-Svelte [manual installation docs](https://www.shadcn-svelte.com/docs/installation)
- Added the [Dropdown Menu](https://www.shadcn-svelte.com/docs/components/dropdown-menu) component set via: `npx shadcn-svelte@latest add dropdown-menu`
- Created a few Storybook stories to showcase some examples

This PR will be followed by a PR that uses this new dropdown menu in the Cloud UI (#3737).

PR #3737, opened 2023-12-20
Use ShadCN's `DropdownMenu` in the Cloud UI

### LINKED ISSUE #3630: Style the Scheduled Report dropdown menus according to new mocks

ericpgreen2:
This PR follows https://github.com/rilldata/rill/pull/3736, which introduces the ShadCN `DropdownMenu` component set. This PR uses the new dropdown menu in the following places in the Cloud UI:
- The top bar's breadcrumbs' dropdown menus
- The one-element menu that opens upon clicking a dashboard's "Share" button
- The user avatar's dropdown menu in the top-right-hand corner of the application
- The scheduled report page's dropdown (which shows the "Edit report", "Delete report" actions)

Closes #3630 

Note: In the user avatar menu, I'd prefer if the "View As" submenu opened on click rather than on hover. I think I'm blocked by a bug with controlled dropdown menus. I've opened up the following issue in ShadCN-Svelte: https://github.com/huntabyte/bits-ui/issues/245


bcolloran (approved):
LGTM, digging it.

web-admin/src/features/authentication/AvatarButton.svelte
```diff
@@ -1,17 +1,18 @@
...
-    Popover,
-    PopoverButton,
-    PopoverPanel,
-  } from "@rgossiaux/svelte-headlessui";
-  import { MenuItem } from "@rilldata/web-common/components/menu";
-  import Menu from "@rilldata/web-common/components/menu/core/Menu.svelte";
-  import { createPopperActions } from "svelte-popperjs";
+    DropdownMenu,
+    DropdownMenuContent,
+    DropdownMenuItem,
```
briangregoryholmes, line 6: This is a nit, but I think `import * as DropdownMenu` and then accessing `Content`, `Root`, `Trigger`, etc. keeps this a lot cleaner.

web-admin/src/features/authentication/AvatarButton.svelte
```diff
@@ -32,64 +33,45 @@
...
       >
-    </Menu>
-  </PopoverPanel>
-</Popover>
+        <svelte:fragment slot="manage-project">
+          <!-- TODO: open this on click, rather than on hover  -->
+          <DropdownMenuSub>
+            <DropdownMenuSubTrigger>View as</DropdownMenuSubTrigger>
+            <DropdownMenuSubContent
+              class="min-w-[150px] max-w-[300px] min-h-[150px] max-h-[190px]"
```
briangregoryholmes, line 58: The ViewAsUserPopover will overflow without changing the layout algorithm for this. You can just add `flex flex-col` to the class.

Screenshot 2023-12-21 at 2 07 01 PM

web-admin/src/features/authentication/AvatarButton.svelte
```diff
@@ -32,64 +33,45 @@
...
-  const [popperRef2, popperContent2] = createPopperActions();
 </script>
 
-<Popover class="relative" let:close={close1}>
-  <PopoverButton use={[popperRef1]}>
+<DropdownMenu>
+  <DropdownMenuTrigger>
     <img
       src={$user.data?.user?.photoUrl}
       alt="avatar"
```
briangregoryholmes, line 42: Unrelated to this PR, really, but do we need a fallback here or are we confident that `photoUrl` will always be defined?

web-admin/src/features/view-as-user/ViewAsUserPopover.svelte
```diff
@@ -44,48 +40,27 @@
...
+<div class="px-0.5 pt-1 pb-2 text-[10px] text-gray-500 text-left">
+  Test your <a target="_blank" href="https://docs.rilldata.com/develop/security"
+    >security policies</a
+  > by viewing this project from the perspective of another user.
+</div>
+<div class="pb-1">
+  <Search bind:value={searchText} autofocus={false} />
+</div>
+{#if visibleUsers.length > 0}
+  <div class="overflow-auto pb-1">
```
briangregoryholmes, line 52: Another nit, but I would explicitly use `DropdownMenuGroup`. ShadCN has a nice ScrollArea component that we should use eventually, but it's not available in the Svelte library yet.

briangregoryholmes (commented):
Looks really great and solves a ton of little problems in one fell swoop. Added some comments about small implementation details.

briangregoryholmes:
Regarding your issue with the `SubTrigger`, I agree that it's unintuitive. You have to also set `disabled={true}` on the SubTrigger and then you can add a control function via `on:click`

web-admin/src/features/authentication/AvatarButton.svelte
ericpgreen2: Agreed, and just updated. I was relying on VSCode auto-imports, so I'm curious – would you do this `import *` by hand, or do you happen to have a VSCode trick?

web-admin/src/features/dashboards/share/ShareDashboardButton.svelte
```diff
@@ -29,29 +19,18 @@
   }
 </script>
 
-<Popover class="relative">
-  <PopoverButton use={[popperRef]}>
+<DropdownMenu>
+  <DropdownMenuTrigger>
     <Button type="secondary">Share</Button>
```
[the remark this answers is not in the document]
ericpgreen2: Good point. Instead of adding builders to our `Button` component, I used our `buttonClasses()` function to apply the `secondary` button classes to the `DropdownMenuTrigger` itself.

web-admin/src/features/authentication/AvatarButton.svelte
ericpgreen2: Ah, good catch. Fixed.

web-admin/src/features/view-as-user/ViewAsUserPopover.svelte
ericpgreen2: Done

web-admin/src/features/navigation/BreadcrumbItem.svelte
```diff
@@ -20,27 +27,32 @@
...
-            : 'text-gray-500'}"
-          on:click={toggleMenu}
-        >
-          <CaretDownIcon size="14px" />
-        </button>
-      </WithSelectMenu>
+      <DropdownMenu>
+        <DropdownMenuTrigger>
+          <button
+            class="flex flex-col justify-center items-center transition-transform hover:translate-y-[2px] {isCurrentPage
```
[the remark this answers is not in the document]
ericpgreen2: Done

ericpgreen2:
> Regarding your issue with the `DropdownMenuSub`, I agree that it's unintuitive. You have to also set `disabled={true}` on the `SubTrigger` and then you can add a control function via `on:click`

Oh, nice find. Works 👍

web-admin/src/features/authentication/AvatarButton.svelte
ericpgreen2: It does look like this could be protected better, though we are [always getting a string in response](https://vscode.dev/github/rilldata/rill/blob/main/admin/server/auth/handlers.go#L171).

PR #3747, opened 2023-12-22
Add Salesforce Connector

cwarden:
## Checklist
- [ ] Manual verification
- [ ] Unit test coverage
- [ ] E2E test coverage
- [ ] Needs manual QA?

#### Issue addressed: 
Use Salesforce Bulk API as a source.

#### Details:
Add Salesforce connector supporting the FileIterator interface using the Bulk API with CSV jobs.

Each source requires a SOQL Query and SObject name.

Credentials can be provided for the source or the connector.

Each bulk job batch is retrieved as a single CSV file for rill to ingest.

## Steps to Verify
1. Add a new Source
2. Select Salesforce
3. Enter Username, Password (with token), SOQL Query, and SObject Name


k-anshul:
Hey @cwarden 

Thanks a lot for working on this.
We are both very happy and impressed to see such important contributions from community members. 
I understand this is still a draft but thought to pass some high level feedback on this : 
There seems to be some boiler plate code around authentication and the bulk API call and I wonder if some library already achieves the same. We already use libraries for `s3/gcs/bigquery/snowflake` etc to do something similar.
I took a quick look and found this [library](https://github.com/simpleforce/simpleforce) but need to evaluate if it supports the features we need to build a MVP connector here. Please let me know what are your thoughts on this and any other library suggestions you might have.

Thank you and looking forward to contribute on this with you.

cwarden:
> Thanks a lot for working on this. We are both very happy and impressed to see such important contributions from community members. I understand this is still a draft but thought to pass some high level feedback on this : There seems to be some boiler plate code around authentication and the bulk API call and I wonder if some library already achieves the same. We already use libraries for `s3/gcs/bigquery/snowflake` etc to do something similar. I took a quick look and found this [library](https://github.com/simpleforce/simpleforce) but need to evaluate if it supports the features we need to build a MVP connector here. Please let me know what are your thoughts on this and any other library suggestions you might have.

I'm using the [force cli library](https://github.com/ForceCLI/force/tree/master/lib), which I maintain.  It's grown out of the command-line tool and some of the code dates back to Go 1.1 so there are some rough edges, but it has been heavily used.

I plan to clean up the authentication in this PR so it doesn't use a credential store.

I'm not sure if it makes sense to move some of the Bulk API handling into the library, but I'm open to suggestions.

runtime/drivers/salesforce/authentication.go
```diff
@@ -0,0 +1,109 @@
...
+func jwtLogin(endpoint string, options authenticationOptions, provider Authenticable) (*force.Force, error) {
+	assertion, err := provider.JwtAssertionForEndpoint(endpoint, options.Username, options.JWT, options.ConnectedApp)
+	if err != nil {
+		return nil, err
+	}
+	return provider.JwtLoginAtEndpoint(endpoint, assertion)
+}
+
+func (provider forceProvider) JwtAssertionForEndpoint(endpoint, username, keyfile, clientID string) (string, error) {
+	return force.JwtAssertionForEndpoint(endpoint, username, keyfile, clientID)
```
k-anshul, line 90: The library expects a path to the key file whereas a token is being directly passed here.

runtime/drivers/salesforce/authentication.go
```diff
@@ -0,0 +1,109 @@
...
+func (provider forceProvider) JwtLoginAtEndpoint(endpoint, assertion string) (*force.Force, error) {
+	session, err := force.JWTLoginAtEndpoint(endpoint, assertion)
+	if err != nil {
+		return nil, fmt.Errorf("JWT authentication failed: %w", err)
+	}
+
+	return force.NewForce(&session), nil
+}
+
+func (provider forceProvider) ForceSoapLoginAtEndpoint(endpoint, username, password string) (*force.Force, error) {
```
k-anshul, line 102: Does this method work in all cases? I got below error which seem to signify that we also need to pass some security token for this type of login to work.
```
failed to ingest source: Authentication failed: SOAP authentication failed: Login failed: LOGIN_MUST_USE_SECURITY_TOKEN: Invalid username, password, security token; or user locked out. Are you at a new location? When accessing Salesforce--either via a desktop client or the API--from outside of your company’s trusted networks, you must add a security token to your password to log in. To get your new security token, log in to Salesforce. From your personal settings, enter Reset My Security Token in the Quick Find box, then select Reset My Security Token.
```

runtime/drivers/salesforce/authentication.go
```diff
@@ -0,0 +1,109 @@
...
+		return nil, err
+	}
+
+	switch {
+	case isJWTSelected:
+		return jwtLogin(endpoint, options, provider)
+	case isSOAPSelected:
+		return provider.ForceSoapLoginAtEndpoint(endpoint, options.Username, options.Password)
+	case !isUsernameSelected:
+		return nil, fmt.Errorf("username missing")
```
k-anshul, line 53: Can this error be thrown earlier ? I see `Username` is being used in both `jwtLogin` and `ForceSoapLoginAtEndpoint`.

runtime/drivers/salesforce/authentication.go
```diff
@@ -0,0 +1,109 @@
...
+	isEndpointSelected := len(options.Endpoint) > 0
+
+	if isEndpointSelected {
+		// URL needs to have scheme lest the force cli lib chokes
+		uri, err := url.Parse(options.Endpoint)
+		if err != nil {
+			return defaultEndpoint, errors.New("Unable to parse endpoint: " + options.Endpoint)
+		}
+
+		if uri.Host == "" {
```
k-anshul, line 68: Shouldn't this be `uri.Schema == ""` ?

k-anshul: Also can be simplified to 
```
if u.Scheme == "" {
	u.Scheme = "https"
}
```

runtime/drivers/salesforce/bulk.go
```diff
@@ -0,0 +1,292 @@
...
+	"strconv"
+	"time"
+
+	force "github.com/ForceCLI/force/lib"
+	"github.com/ForceCLI/force/lib/record_reader"
+	"go.uber.org/zap"
+)
+
+// Forceable represents an object which can talk to salesforce
+type Forceable interface {
```
k-anshul, line 22: Either the library should expose an interface or we should directly use `force.Force`. I don't think its required here.

runtime/drivers/salesforce/bulk.go
```diff
@@ -0,0 +1,292 @@
+package salesforce
```
k-anshul, line 1: TBH I still don't think its the right place to have the logic contained in the file here. I think the library should encapsulate the logic here and should provide APIs to start the job and retrieve results(may be via some iterator). You can take a look at the bigquery SDK: https://github.com/googleapis/google-cloud-go/blob/0020ab23b0d1344cf03b2bfa5e2afe947b70240a/bigquery/query.go#L373

It provides APIs to create a client(similar to a force session here), create and start a job via `client.Query.Read` method(similar to `makeBulkJob` and `startJob` here) and exposes an iterator to query results(similar to `fetchBatchResult`) here.

runtime/drivers/salesforce/sql_store.go
```diff
@@ -0,0 +1,171 @@
...
+
+	var username, password, endpoint, key, clientID string
+	if srcProps.Username != "" { // get from src properties
+		username = srcProps.Username
+	} else if u, ok := c.config["username"].(string); ok && u != "" { // get from driver configs
+		username = u
+	} else {
+		return nil, fmt.Errorf("the property 'username' is required for Salesforce. Provide 'username' in the YAML properties or pass '--env connector.salesforce.username=...' to 'rill start'")
+	}
+	if srcProps.Endpoint != "" { // get from src properties
```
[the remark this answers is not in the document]
cwarden: Updated in faea05276b5691f43bc3811ad69587fad3d618fb

runtime/drivers/salesforce/sql_store.go
```diff
@@ -0,0 +1,171 @@
...
+		Username:     username,
+		Password:     password,
+		JWT:          key,
+		Endpoint:     endpoint,
+		ConnectedApp: clientID,
+	}
+
+	session, err := authenticate(authOptions)
+	if err != nil {
+		return nil, fmt.Errorf("Authentication failed: %w", err)
```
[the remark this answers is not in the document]
cwarden: Updated in 6fc5b102bd88c8cd149db325ef47fdc82f058096.

runtime/drivers/salesforce/authentication.go
```diff
@@ -0,0 +1,109 @@
...
+	JwtAssertionForEndpoint(endpoint string, username string, keyfile string, clientID string) (string, error)
+	JwtLoginAtEndpoint(endpoint string, assertion string) (*force.Force, error)
+	ForceSoapLoginAtEndpoint(endpoint string, userName string, password string) (*force.Force, error)
+}
+
+// a concrete implementation of Authenticable
+type forceProvider struct{}
+
+func authenticate(options authenticationOptions) (*force.Force, error) {
+	if options.ConnectedApp == "" {
```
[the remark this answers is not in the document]
cwarden: Updated the `authenticate` function in 3e25cdcf1770d7b9f15fdd4acc59931129507510 to return an error if the connected app client id is not set.

runtime/drivers/salesforce/authentication.go
```diff
@@ -0,0 +1,109 @@
...
+	isSOAPSelected := len(options.Password) > 0
+
+	endpoint, err := endpoint(options)
+	if err != nil {
+		return nil, err
+	}
+
+	switch {
+	case isJWTSelected:
+		return jwtLogin(endpoint, options, provider)
```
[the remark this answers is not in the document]
cwarden: Removed `Authenticable` interface and `forceProvider` implementation in 3b31ef72a6d10315ca714c5f592ee84f7cf5c7a6.

runtime/drivers/salesforce/authentication.go
cwarden: The convention in Salesforce is to [append the token to the password if a token is required](https://developer.salesforce.com/docs/atlas.en-us.api.meta/api/sforce_api_calls_login.htm).

runtime/drivers/salesforce/authentication.go
cwarden: Addressed in 8b973d169b894a27a2792ebf344c828ad452941a.

runtime/drivers/salesforce/authentication.go
cwarden: Added tests in a2446ca17a34da31fb0876cbbc49eb7da67ee26f to illustrate support for when `authenticationOptions.Endpoint` contains a hostname rather than a URL.

runtime/drivers/salesforce/bulk.go
cwarden: Removed Forceable interface in de5424725d167b58dbbedb0273f12933eeeeccc0.

cwarden:
Thanks for the thorough code review, @k-anshul.  You're seeing some of the artifacts of this code being extracted from an existing application, such as the interfaces that were used to facilitate testing.

runtime/drivers/salesforce/bulk.go
k-anshul: Yeah I agree that moving this to a library requires lot more thinking on multiple use cases.

I also happen to notice that salesforce has a newer version of the Bulk API which automatically performs chunking and provides result from a single endpoint similar to what I suggested to be done upstream.
Could you please evaluate that? Salesforce also recommends to use Bulk API 2.0 instead of Bulk API for simpler integration.

k-anshul (changes requested):

runtime/drivers/salesforce/authentication.go
cwarden: I misunderstood how `URL` works.  I didn't realize that `URL.Host` could be the empty string, and `URL.String()` would still return the hostname (or the URL after `URL.Scheme` is set).

Updated in 4b048742431412d67003cc0c4fd85a55ca39d9f4.

runtime/drivers/salesforce/bulk.go
cwarden: I expect there will be new problems to troubleshoot moving to Bulk API 2.0.  The docs suggest that you can get [query timeouts](https://developer.salesforce.com/docs/atlas.en-us.api_asynch.meta/api_asynch/bulk_api_2_0_troubleshoot_query_timeouts.htm), and may get [timeouts if you choose the wrong number of records to retrieve at a time](https://developer.salesforce.com/docs/atlas.en-us.api_asynch.meta/api_asynch/query_get_job_results.htm).

I'm more comfortable using the original Bulk API at this point as we've already identified the rough edges and have been successfully using it for a decade.

k-anshul (commented):
Thanks for the PR @cwarden LGTM 
I will also let @begelundmuller take a look at this. 
Two things:
1. I discovered that whenever I am submitting an incorrect query, the application gets stuck. Could you please take a look at this ? Sample incorrect query : `SELECT Id, IsDeleted, Name, Type, ParentId, Rating, Site FROM Account WHERE Rating is not null`
2. Would you also be interested in adding docs for the connector ? It can also be taken up in a separate PR. PR for reference : https://github.com/rilldata/rill/pull/3147/files

cwarden:
> Thanks for the PR @cwarden LGTM I will also let @begelundmuller take a look at this. Two things:
> 
> 1. I discovered that whenever I am submitting an incorrect query, the application gets stuck. Could you please take a look at this ? Sample incorrect query : `SELECT Id, IsDeleted, Name, Type, ParentId, Rating, Site FROM Account WHERE Rating is not null`

Fixed in d82fb1f2458d9f5af3d799d41d66d5754df2a645.

> 2. Would you also be interested in adding docs for the connector ? It can also be taken up in a separate PR. PR for reference : https://github.com/rilldata/rill/pull/3147/files

Added in 8951aec3d07b335ad529192c2c14ae333b8f1d08.

k-anshul (approved):
Thanks

PR #3750, opened 2023-12-27
buffered zap logger

### LINKED ISSUE #3131: Runtime: Logging improvements
- [x] Implement the `GetLogs` and `WatchLogs` APIs for a controller's logger
- [x] Implement a `rill project logs` CLI command for getting recent logs and with a flag for streaming logs (similar to `kubectl logs`)
- [ ] Move the logger to `runtime/registry.go` – see comment [here](https://github.com/rilldata/rill/pull/3170#issuecomment-1775732356)
- [ ] Use the same logger for connectors/drivers (to propagate e.g. source download logs to `rill project logs` as well)
- [ ] Deprecate the `Name("console")` hack and use debug-level logging for hiding low-level logs from the console (also see the `initLogger` function in `cli/pkg/local/app.go`, which this will enable cleaning up).
- [ ] Propagate source name to transporter logs

pjain1:
Fixes #3131 

Since non console namespaced logs are changed to debug level, `runtime-cloud` does not output to console logs like 
```
...
{"level":"info","ts":1704707695.429421,"caller":"runtime/registry.go:346","msg":"controller starting","instance_id":"..."}
{"level":"info","ts":1704707695.430704,"caller":"runtime/registry.go:357","msg":"controller ready","instance_id":"..."}
...
```
unless the `RILL_RUNTIME_LOG_LEVEL` is set to `debug`


runtime/registry.go
```diff
@@ -354,7 +368,7 @@ func (r *registryCache) restartController(iwc *instanceWithController) {
 				close(iwc.readyCh)
 				r.mu.Unlock()
 
-				r.logger.Info("controller ready", zap.String("instance_id", iwc.instance.ID))
+				r.logger.Debug("controller ready", zap.String("instance_id", iwc.instance.ID))
```
begelundmuller, line 371: Should also use `iwc.logger`?

runtime/registry.go
```diff
@@ -366,7 +380,7 @@ func (r *registryCache) restartController(iwc *instanceWithController) {
 			r.mu.Unlock()
 
 			if errors.Is(err, iwc.ctx.Err()) {
-				r.logger.Warn("controller stopped", attrs...)
+				r.logger.Debug("controller stopped", attrs...)
```
begelundmuller, line 383: Should also use `iwc.logger`?

runtime/registry.go
```diff
@@ -333,18 +347,18 @@ func (r *registryCache) restartController(iwc *instanceWithController) {
 			// Before starting the controller, sync the repo.
 			// This is necessary for resources (usually sources or models) that reference files in the repo,
 			// and may be triggered before the project parser is triggered and syncs the repo.
-			r.logger.Info("syncing repo", zap.String("instance_id", iwc.instance.ID))
+			iwc.logger.Debug("syncing repo", zap.String("instance_id", iwc.instance.ID))
```
begelundmuller, line 350: I think `zap.String("instance_id", iwc.instance.ID)` is redundant when using `iwc.logger` since it's already passed in `With`. This applies a few places in this function.

cli/pkg/local/app.go
```diff
@@ -264,14 +264,6 @@ func (a *App) Serve(httpPort, grpcPort int, enableUI, openBrowser, readonly bool
 		Readonly:         readonly,
 	}
 
-	// Create server logger.
-	// It only logs error messages when !verbose to prevent lots of req/res info messages.
-	lvl := zap.ErrorLevel
-	if a.Verbose {
-		lvl = zap.DebugLevel
-	}
-	serverLogger := a.BaseLogger.WithOptions(zap.IncreaseLevel(lvl))
```
begelundmuller, line 273: Instead of making log levels on the server configurable (by passing it into `LoggingUnaryServerInterceptor`, `ServeGRPC`, etc.), I think we should keep using info logs in the server and then keep this snippet here instead.

begelundmuller (changes requested):

runtime/server/controller.go
```diff
@@ -176,7 +176,7 @@ func (s *Server) WatchResources(req *runtimev1.WatchResourcesRequest, ss runtime
 			var err error
 			r, access, err = s.applySecurityPolicy(ss.Context(), req.InstanceId, r)
 			if err != nil {
-				s.logger.Info("failed to apply security policy", zap.String("name", n.Name), zap.Error(err))
+				s.logger.Debug("failed to apply security policy", zap.String("name", n.Name), zap.Error(err))
```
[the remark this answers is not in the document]
pjain1: should we make it `error` ?

runtime/pkg/logutil/handler.go
```diff
@@ -1,45 +1,41 @@
...
-	return d.logs.Add(record)
+func (d *BufferedZapCore) Write(entry zapcore.Entry, fields []zapcore.Field) error {
+	return d.logs.AddZapEntry(entry, fields)
+}
+
+func (d *BufferedZapCore) With(fields []zapcore.Field) zapcore.Core {
+	// TODO: should ideally create a clone of the core with new log buffer and append the new fields
+	clone := *d
+	clone.fields = append(clone.fields, fields...)
+	return &clone
```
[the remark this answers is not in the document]
pjain1: 1. May be I was just confused that `With` should return a new independent logger but that does not seem like it.
2.  Not sure I understand how it will overwrite as `append` adds to the end

runtime/pkg/logutil/handler.go
```diff
@@ -1,45 +1,41 @@
...
-	"go.uber.org/zap/exp/zapslog"
-	"golang.org/x/exp/slog"
+	"go.uber.org/zap/zapcore"
 )
 
-type DuplicatingHandler struct {
-	zapHandler *zapslog.Handler
-	logs       *logbuffer.Buffer
+type BufferedZapCore struct {
+	fields []zapcore.Field
```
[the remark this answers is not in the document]
pjain1: true

cli/pkg/local/app.go
pjain1: I think I removed this because the logic is duplicated in `initLogger` function [here](https://github.com/rilldata/rill/blob/main/cli/pkg/local/app.go#L480). Also If we keep info level for grpc calls then it will start to show in console logs which were not being shown earlier, hence passing debug level for local rill-dev runtime and info level for standalone/cloud runtime.

runtime/pkg/logutil/handler.go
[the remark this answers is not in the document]
pjain1: For 2 I understand what you mean - the fields for the original logger will also get the appended fields so we need to create a new slice so original array remains as it is.

runtime/server/controller.go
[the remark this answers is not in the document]
begelundmuller: Thought about it, but I think it will trigger if a user has broken template logic in their security policy. And we don't really want error-level logs for user errors.

runtime/pkg/logutil/handler.go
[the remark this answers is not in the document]
begelundmuller: Actually it's a little more complicated – see this example: https://go.dev/play/p/ghSHN5YV9Du.

Basically slices represent ranges on top of an underlying allocated array. So while `With` won't change the slice of the original, it can corrupt the slice of another logger previously created using `With`.

cli/pkg/local/app.go
begelundmuller: This code is slightly different – basically what it does is it increases the level of the server logger to error, which means anything below error will be discarded. Therefore, the server can print at info level without anything being output on local:
```go
// a.BaseLogger outputs at info-level
serverLogger := a.BaseLogger.WithOptions(zap.IncreaseLevel(zap.ErrorLevel))
// a.BaseLogger still outputs at info-level
// But serverLogger only forwards error+ level logs to a.BaseLogger
```

runtime/pkg/logutil/handler.go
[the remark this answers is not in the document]
pjain1: ok interesting, thanks for the explanation.

cli/pkg/local/app.go
pjain1: Went through the code again. We have parent logger created in `NewApp`, that logger has two cores - file and console and depending on verbose flag the level will be either debug or info. Now this parent logger is given to runtime and controller creates a new duplicating logger from this. This logger is also given to server which by defaults logs only error however with verbose it can log at debug level because parent logger (a.BaseLogger) will also have debug level set. 

So few observations - 
1. Irrespective of verbose flag, server grpc logs goes only to file never console as they are not console namespaced.
2. When verbose is set the code above `zap.IncreaseLevel(zap.DebugLevel)` is essentially a no-op as BaseLogger already have debug level so we can remove this and simplify.

Now in the current PR -
Few observations -
1. If we keep the code above then server logger will allow debug+ level logs and the since the console core of the parent logger allows info+, all the info level grpc calls will go to console which is different behaviour from earlier but seems ok. This was the reason to send explicit flag. Just an observation we can keep the code.
2. While creating the parent logger, console core can be created directly using `zapcore.NewCore(consoleEncoder, zapcore.Lock(os.Stdout), zapcore.InfoLevel)` instead of `zapfilter.NewFilteringCore(zapcore.NewCore(consoleEncoder, zapcore.Lock(os.Stdout), logLevel), zapfilter.MinimumLevel(zapcore.InfoLevel))` since even if debug level is set for parent, console will not get them because of `zapfilter.MinimumLevel(zapcore.InfoLevel)` rule.

begelundmuller: > 2. When verbose is set the code above zap.IncreaseLevel(zap.DebugLevel) is essentially a no-op as BaseLogger already have debug level so we can remove this and simplify.

It's not quite a no-op because it changes the server logger level from error (as set on line 269) to debug. But you're right it could have been simplified slightly like this:
```go
serverLogger := a.BaseLogger
if !a.Verbose {
  serverLogger = a.BaseLogger.WithOptions(zap.IncreaseLevel(zap.ErrorLevel))
}
```

> 1. If we keep the code above then server logger will allow debug+ level logs and the since the console core of the parent logger allows info+, all the info level grpc calls will go to console which is different behaviour from earlier but seems ok. This was the reason to send explicit flag. Just an observation we can keep the code.

By default, `a.Verbose` is false, so all the server-level info logs will go neither to the console nor the log file. That's what we want since server-level info logs are very chatty and not really useful on local.

But if `a.Verbose` is true, then after this change, it will be as you say that the server info logs will go to the console. And I think that's an improvement – because if you pass `--verbose` it's better to output as much as possible to the console.

> 2. While creating the parent logger, console core can be created directly using zapcore.NewCore(consoleEncoder, zapcore.Lock(os.Stdout), zapcore.InfoLevel) instead of zapfilter.NewFilteringCore(zapcore.NewCore(consoleEncoder, zapcore.Lock(os.Stdout), logLevel), zapfilter.MinimumLevel(zapcore.InfoLevel)) since even if debug level is set for parent, console will not get them because of zapfilter.MinimumLevel(zapcore.InfoLevel) rule.

Yeah looks like this can be simplified. But one difference here – if `a.Verbose` is true, we should change the console log level to `zapcore.DebugLevel`. Since as stated above, if someone passes `--verbose` we just want to output as much as possible to the console.

runtime/registry.go
```diff
@@ -452,3 +456,14 @@ func (r *registryCache) ensureProjectParser(ctx context.Context, instanceID stri
 		r.logger.Error("could not create project parser", zap.Error(err), zap.String("instance_id", instanceID), observability.ZapCtx(ctx))
 	}
 }
+
+func (r *registryCache) getLogbuffer(instanceID string) (*logbuffer.Buffer, error) {
```
begelundmuller, line 460: Maybe move up to after `getController` to keep similar getters grouped

runtime/pkg/logutil/handler.go
```diff
@@ -21,13 +21,15 @@ func (d *BufferedZapCore) Enabled(level zapcore.Level) bool {
 }
 
 func (d *BufferedZapCore) Write(entry zapcore.Entry, fields []zapcore.Field) error {
+	fields = append(d.fields, fields...)
```
begelundmuller, line 24: Won't this suffer from a similar potential issue of overwriting the underlying array of `d.fields` (which is shared)?

runtime/pkg/logbuffer/buffer.go
```diff
@@ -27,30 +27,28 @@ func NewBuffer(maxMessageCount int, maxBufferSize int64) *Buffer {
 	}
 }
 
-func (b *Buffer) Add(record slog.Record) error {
+func (b *Buffer) AddZapEntry(entry zapcore.Entry, fields []zapcore.Field) error {
 	size := 0
 	attrs := make(map[string]any)
```
begelundmuller, line 32: Use `make(map[string]any, len(fields))` to optimize allocations

runtime/pkg/logutil/handler.go
begelundmuller: We can save a redundant map allocation by passing `d.fields` and `fields` separately to `AddZapEntry`, and having `AddZapEntry` combine the two in the map it builds. I.e. refactor to something like `AddZapEntry(entry, d.fields, fields)`.

begelundmuller (changes requested):

runtime/pkg/logutil/handler.go
pjain1: or I can flip the variables to append call like `fields = append(fields, d.fields...)` ?

begelundmuller: By passing directly to `AddZapEntry` (which allocates a map for it), we would save an allocation. And since this happens for every log, that’s a lot of allocations.

`append(fields, d.fields...)` would be safe, but still do a new underlying allocation in most cases.

Issue #3744, opened 2023-12-22
Make it possible to adjust the height of the "Model Data" dashboard module


ericpgreen2:
In `RowsViewer.svelte`, the height is currently fixed at `h-72`. The height should be dynamic and controlled via dragging the "Model Data" header up/down.

kensuke-b-sasaki:
Hi @ericpgreen2 , I'm k_ssk on discord, thank you for creating issues for new feature.
I really appreciate for your cooperation & looking forward to the future implementation!

kensuke-b-sasaki:
@ericpgreen2 
Today I upgrade to v0.39.0 and confirmed this change is applied.
Really appreciate for your help!

ericpgreen2:
Terrific! Don't hesitate to reach out if there's anything else that would improve your experience.

PR #3751, opened 2023-12-27
Allow resizing the height of the "Model Data" dashboard module in Rill Developer

ericpgreen2:
Closes https://github.com/rilldata/rill/issues/3744

### DIFF
--- a/web-common/src/features/dashboards/rows-viewer/RowsViewerAccordion.svelte
+++ b/web-common/src/features/dashboards/rows-viewer/RowsViewerAccordion.svelte
@@ -6,12 +6,20 @@
   import { formatCompactInteger } from "@rilldata/web-common/lib/formatters";
   import { createQueryServiceMetricsViewTotals } from "@rilldata/web-common/runtime-client";
   import { runtime } from "@rilldata/web-common/runtime-client/runtime-store";
+  import { writable } from "svelte/store";
   import { useDashboardStore } from "web-common/src/features/dashboards/stores/dashboard-stores";
+  import { drag } from "../../../layout/drag";
+  import type { LayoutElement } from "../../../layout/workspace/types";
   import { featureFlags } from "../../feature-flags";
   import ExportModelDataButton from "./ExportModelDataButton.svelte";
   import RowsViewer from "./RowsViewer.svelte";
 
   export let metricViewName: string;
+
+  const INITIAL_HEIGHT_EXPANDED = 300;
+  const MIN_HEIGHT_EXPANDED = 30;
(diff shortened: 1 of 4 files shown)


briangregoryholmes:
Approving as it's a blocker, but I'd like to propose a different approach for this throughout our UI and will do so in another PR.

I've marked two nits for review.

There's also some unintended cursor snapping behavior, but this should be addressed by my PR.

web-common/src/features/dashboards/rows-viewer/RowsViewerAccordion.svelte
```diff
@@ -92,26 +100,45 @@
...
 </script>
 
 <div>
-  <button
-    aria-label="Toggle rows viewer"
-    class="w-full bg-gray-100 h-7 text-left px-2 border-t border-t-gray-200 text-xs text-gray-800 flex items-center gap-1"
-    on:click={toggle}
+  <div
+    class="flex items-center px-2 h-7 w-full bg-gray-100 border-t border-t-gray-200 {isOpen
+      ? '!cursor-move'
```
briangregoryholmes, line 113: This should be `cursor-ns-resize` or `cursor-row-resize`

web-common/src/layout/drag.ts
```diff
@@ -1,4 +1,15 @@
...
+
+interface DragParams {
+  store: Writable<LayoutElement>;
+  minSize?: number;
+  maxSize?: number;
+  reverse?: boolean;
+  orientation?: "horizontal" | "vertical";
+}
+
+export function drag(node, params: DragParams) {
```
briangregoryholmes, line 12: This was an issue before the PR, but the `mousemove` function needs to prevent the default browser behavior. Resizing selects elements in Safari.

briangregoryholmes (approved):

web-common/src/layout/drag.ts
ericpgreen2: Nice catch, fixed 👍

web-common/src/features/dashboards/rows-viewer/RowsViewerAccordion.svelte
ericpgreen2: Yeah, just switched to `cursor-ns-resize`. Originally I chose `cursor-move` to mirror the behavior of the other `use:drag` instances, but, you're right, this is more correct.

ericpgreen2:
> Approving as it's a blocker, but, after looking at how this works throughout our application, I'd like to propose a different approach for this and will do so in another PR.

Sounds good, feel free 👍

PR #3754, opened 2023-12-28
Runtime: Prevent file watcher errors when a file is rapidly created and deleted

begelundmuller:
I saw this error in a project and I hope this will fix it:
```
2023-12-28T13:47:32.266	ERROR	Stopped watching for file changes	{"err": "watch failed: no such file or directory"}
2023-12-28T13:47:32.266	WARN	Reconcile failed	{"name": "parser", "kind": "ProjectParser", "elapsed": "10.016s", "error": "watch failed: no such file or directory"}
```

### DIFF
--- a/runtime/drivers/file/watcher.go
+++ b/runtime/drivers/file/watcher.go
@@ -48,7 +48,7 @@ func newWatcher(root string) (*watcher, error) {
 		buffer:      make(map[string]drivers.WatchEvent),
 	}
 
-	err = w.addDir(root, false)
+	err = w.addDir(root, false, true)
 	if err != nil {
 		w.watcher.Close()
 		return nil, err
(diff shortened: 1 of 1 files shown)


runtime/drivers/file/watcher.go
```diff
@@ -193,6 +194,9 @@ func (w *watcher) runInner() error {
 func (w *watcher) addDir(path string, replay bool) error {
 	err := w.watcher.Add(path)
 	if err != nil {
+		if os.IsNotExist(err) || errors.Is(err, unix.ENOENT) {
```
k-anshul, line 197: Does the upstream code handle cases when its a genuine error that file being watched doesn't exist ? 
Could it be problematic to not return an error if trying to watch a non existent directory ?

begelundmuller: No it doesn't for the root directory case, good catch! Should be fixed now :)

k-anshul (approved):

PR #3758, opened 2024-01-01
Introduce backend flag to indicate that a deployed project is in an a…

rakeshsharma14317:


begelundmuller (changes requested):
Change of plan – @nishantmonu51 requested to turn this into a generic tagging feature. Suggested implementation:
- Instead of `ProdSLA bool`, we would have `ProdTags []string`
- Instead of `rill sudo project search --sla`, we would have `rill sudo project search --tag tag1 --tag tag2 ...` (use `cmd.Flags().StringSliceVar` to read multiple values into a `[]string`).

nishantmonu51:
Nit: Instead of `ProdTags []string` we can just have `Tags []string`

admin/database/postgres/migrations/0019.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE projects ADD COLUMN tags TEXT[];
```
begelundmuller, line 1: Add `NOT NULL DEFAULT '{}'::text[]`

admin/database/database.go
```diff
@@ -265,6 +267,7 @@ type Project struct {
 	ProdSlots            int       `db:"prod_slots"`
 	ProdTTLSeconds       *int64    `db:"prod_ttl_seconds"`
 	ProdDeploymentID     *string   `db:"prod_deployment_id"`
+	Tags                 Tags      `db:"tags"`
```
begelundmuller, line 270: Try just `[]string` after adding the `NOT NULL` change (and resetting the dev postgres) to the migration as suggested in another comment.

If that still doesn't work, try `pgtype.Array[string]` instead. It should be possible to get rid of the custom `Tags` type.

admin/deployments.go
```diff
@@ -277,6 +277,7 @@ func (s *Service) HibernateDeployments(ctx context.Context) error {
 			ProdSlots:            proj.ProdSlots,
 			Region:               proj.Region,
 			ProdTTLSeconds:       proj.ProdTTLSeconds,
+			Tags:                 proj.Tags,
 			ProdDeploymentID:     nil,
```
begelundmuller, line 281: Should keep same order as in struct, i.e. `Tags` come last

admin/projects.go
```diff
@@ -91,6 +91,7 @@ func (s *Service) CreateProject(ctx context.Context, org *database.Organization,
 		ProdSlots:            proj.ProdSlots,
 		Region:               proj.Region,
 		ProdTTLSeconds:       proj.ProdTTLSeconds,
+		Tags:                 proj.Tags,
 		ProdDeploymentID:     &depl.ID,
```
begelundmuller, line 95: Order

admin/server/projects.go
```diff
@@ -276,6 +282,7 @@ func (s *Server) CreateProject(ctx context.Context, req *adminv1.CreateProjectRe
 		GithubURL:            &req.GithubUrl,
 		GithubInstallationID: &installationID,
 		ProdVariables:        req.Variables,
+		Tags:                 req.Tags,
 		ProdTTLSeconds:       prodTTL,
```
begelundmuller, line 286: Order

cli/cmd/sudo/project/search.go
```diff
@@ -50,6 +52,7 @@ func SearchCmd(ch *cmdutil.Helper) *cobra.Command {
 			return nil
 		},
 	}
+	searchCmd.Flags().StringArrayVar(&tags, "tag", []string{}, "Tags to filter projects by")
```
begelundmuller, line 55: Not sure, but I think this should be `StringSliceVar` to support multiple flags mapped to one array? Like `--tag tag1 --tag tag2`

cli/cmd/sudo/project/search.go
```diff
@@ -9,9 +9,10 @@ import (
 func SearchCmd(ch *cmdutil.Helper) *cobra.Command {
 	var pageSize uint32
 	var pageToken string
+	var tags []string
 
 	searchCmd := &cobra.Command{
-		Use:   "search",
+		Use:   "search <pattern>",
 		Args:  cobra.ExactArgs(1),
```
begelundmuller, line 16: Let's make the pattern optional, like `search [<pattern>]` and have it default to `%` if not set (which means match all)

cli/cmd/sudo/tags/get.go
```diff
@@ -0,0 +1,43 @@
...
+
+	"github.com/rilldata/rill/cli/pkg/cmdutil"
+	adminv1 "github.com/rilldata/rill/proto/gen/rill/admin/v1"
+	"github.com/spf13/cobra"
+)
+
+func GetCmd(ch *cmdutil.Helper) *cobra.Command {
+	getCmd := &cobra.Command{
+		Use:   "get",
+		Args:  cobra.ExactArgs(2),
```
begelundmuller, line 14: Should be `get <org> <project>`

cli/cmd/sudo/tags/set.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"github.com/rilldata/rill/cli/pkg/cmdutil"
+	adminv1 "github.com/rilldata/rill/proto/gen/rill/admin/v1"
+	"github.com/spf13/cobra"
+)
+
+func SetCmd(ch *cmdutil.Helper) *cobra.Command {
+	var tags []string
+	setCmd := &cobra.Command{
+		Use:   "set <organization> <project> --tags <tag1> --tags <tag2> ...",
```
begelundmuller, line 14: Flags are not part of the usage + it's `--tag`, not `--tags` here

proto/rill/admin/v1/api.proto
```diff
@@ -298,6 +298,14 @@ service AdminService {
     };
   }
 
+  // Add a SudoUpdateTags endpoint for superusers to update tags
+  rpc SudoUpdateTags(SudoUpdateTagsRequest) returns (SudoUpdateTagsResponse) {
+    option (google.api.http) = {
+      patch: "/v1/superuser/tags",
```
begelundmuller, line 304: Suggestion: `/v1/superuser/projects/tags`

proto/rill/admin/v1/api.proto
```diff
@@ -639,6 +648,7 @@ message CreateProjectRequest {
   string prod_branch = 9;
   string github_url = 10;
   map<string, string> variables = 11;
+  repeated string tags = 13;
```
begelundmuller, line 651: Must not be allowed as part of `CreateProject` since only superusers should be able to add tags

go.mod
```diff
@@ -56,6 +56,7 @@ require (
 	github.com/joho/godotenv v1.5.1
 	github.com/kelseyhightower/envconfig v1.4.0
 	github.com/lensesio/tableprinter v0.0.0-20201125135848-89e81fc956e7
+	github.com/lib/pq v1.10.9
```
begelundmuller, line 59: This dependency isn't used – try running `go mod tidy`

cli/cmd/sudo/project/search.go
rakeshsharma14317: Looks like both are working fine, `StringSliceVar` make more sense here, updating this 👍

go.mod
rakeshsharma14317: using it to get rid of the custom Tags type.

admin/database/database.go
```diff
@@ -255,18 +258,19 @@ type Project struct {
...
+	GithubInstallationID *int64         `db:"github_installation_id"`
+	Subpath              string         `db:"subpath"`
+	ProdBranch           string         `db:"prod_branch"`
+	ProdVariables        Variables      `db:"prod_variables"`
+	ProdOLAPDriver       string         `db:"prod_olap_driver"`
+	ProdOLAPDSN          string         `db:"prod_olap_dsn"`
+	ProdSlots            int            `db:"prod_slots"`
+	ProdTTLSeconds       *int64         `db:"prod_ttl_seconds"`
+	ProdDeploymentID     *string        `db:"prod_deployment_id"`
+	Tags                 pq.StringArray `db:"tags"`
```
begelundmuller, line 271: We don't use `lib/pq`, we use `jackc/pgx` for Postgres. Does this work? If yes, you can just put `[]string` here. If not, should use `pgtype.Array[string]`.

admin/server/projects.go
```diff
@@ -174,8 +175,13 @@ func (s *Server) SearchProjectNames(ctx context.Context, req *adminv1.SearchProj
 		return nil, err
 	}
 	pageSize := validPageSize(req.PageSize)
+	var projectNames []string
 
-	projectNames, err := s.admin.DB.FindProjectPathsByPattern(ctx, req.NamePattern, token.Val, pageSize)
+	if req.Tags != nil && len(req.Tags) > 0 {
+		projectNames, err = s.admin.DB.FindProjectPathsByPatternAndTags(ctx, req.NamePattern, token.Val, req.Tags, pageSize)
```
begelundmuller, line 181: nit: move the newline up one line (so declaration of `projectNames` goes in same group as assignment)

cli/cmd/sudo/tags/set.go
```diff
@@ -0,0 +1,55 @@
...
+
+			tags := res.Project.Tags
+			fmt.Printf("Project: %s\n", res.Project.Name)
+			fmt.Printf("Organization: %s\n", res.Project.OrgName)
+			fmt.Printf("Tags: %v\n", tags)
+
+			return nil
+		},
+	}
+	setCmd.Flags().StringArrayVar(&tags, "tag", []string{}, "Tags to set on the project")
```
begelundmuller, line 52: Should also use `StringSliceVar` here

cli/cmd/sudo/tags/get.go
```diff
@@ -0,0 +1,43 @@
...
+				Name:             args[1],
+			})
+			if err != nil {
+				return err
+			}
+
+			tags := res.Project.Tags
+			fmt.Printf("Project: %s\n", res.Project.Name)
+			fmt.Printf("Organization: %s\n", res.Project.OrgName)
+			fmt.Printf("Tags: %v\n", tags)
```
begelundmuller, line 36: Would be nicer to do a comma-separated list, something like `fmt.Printf("Tags: %s\n", strings.Join(res.Project.Tags, ","))`

begelundmuller (changes requested):

admin/database/database.go
rakeshsharma14317: I tried both ways you suggested, only works if we use pq.StringArray or custom impl of Tags

begelundmuller (approved):

PR #3776, opened 2024-01-04
Wait for sub processes to be killed in E2E tests

### LINKED ISSUE #3738: Test for leaderboard and dimension table sorting is failing sporadically in CI
It seems to fail ~50% of the time.

image

AdityaHegde:
closes #3738


begelundmuller (approved):

web-local/test/ui/utils/startRuntimeForEachTest.ts
```diff
@@ -50,13 +63,20 @@ export function startRuntimeForEachTest() {
   });
 
   test.afterEach(async () => {
-    const processExit = new Promise((resolve) => {
-      childProcess.on("exit", resolve);
```
begelundmuller, line 54: 1. Might it be a problem that it was registering the `on` callback inside the promise, which will run after the kill signal has been sent instead of after? Might be better to register the callback maybe even when creating the process.
2. Based on some googling, it seems the `"close"` event would be safer than the `"exit"` event

begelundmuller (approved):

PR #3778, opened 2024-01-04
Adding support for sections in searchable filters

AdityaHegde:
In preparation of adding measure filters we need measures and dimensions sections in the add filter drop down.

This PR only adds this grouping feature to our searchable filter component. Visit http://localhost:3000/dev/grouped-searchable-filter to check how it will look.

### DIFF
--- a/web-common/src/components/menu/core/MenuItem.svelte
+++ b/web-common/src/components/menu/core/MenuItem.svelte
@@ -127,6 +127,8 @@
   $: bgColor = dark
     ? `${hovered ? "bg-gray-600" : ""}`
     : `${hovered ? "bg-gray-200" : ""}`;
+
+  $: console.log(icon, selected);
 </script>
 
 <button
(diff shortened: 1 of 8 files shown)


web-common/src/components/menu/core/MenuItem.svelte
```diff
@@ -127,6 +127,8 @@
   $: bgColor = dark
     ? `${hovered ? "bg-gray-600" : ""}`
     : `${hovered ? "bg-gray-200" : ""}`;
+
+  $: console.log(icon, selected);
```
ericpgreen2, line 131: Leftover console log

web-local/src/routes/dev/grouped-searchable-filter/+page.svelte
ericpgreen2, line 1: In the future, it'd be better to use Storybook rather than a `/dev` route, so these showcases aren't included in the production build.

ericpgreen2 (approved):
Looks good 👍 

Nit: in the multiselect, I noticed the checkmarks are a different color when one item is selected...

image

... versus when two items are selected...

image

AdityaHegde:
I havent changed the color scheme. I believe that is to indicate that once selected, there should atleast be a single value. Either way it is a design concern and out of scope of this PR.

PR #3779, opened 2024-01-04
Minor follow ups to measure filter API changes

AdityaHegde:
While adding UI support for measure filters and the new filter format some changes were needed to not crash the API. 

- Adding some empty checks here.
- Also updating dashboard url state to keep go changes separate.


runtime/queries/metricsview_aggregation.go
```diff
@@ -503,7 +505,9 @@ func (q *MetricsViewAggregation) buildMetricsAggregationSQL(mv *runtimev1.Metric
 		if err != nil {
 			return "", nil, err
 		}
-		havingClause = "HAVING " + havingClause
+		if strings.TrimSpace(havingClause) != "" {
+			havingClause = "HAVING " + havingClause
+		}
```
begelundmuller, line 510: It already checks that `q.Having != nil` – how can the it evaluate to an empty having clause if it's not nil?

begelundmuller (approved):

runtime/queries/metricsview_aggregation.go
AdityaHegde: This is to gracefully handle cases with empty AND/OR or IN filters.

PR #3781, opened 2024-01-05
Cloud UI: add resource-level status to the project status page

ericpgreen2:
This PR enhances the Cloud UI project status page with resource-level status. Previously, we synthesized the resource-level status and displayed one project-wide status. This hid rich reconciliation data from the user and led to confusion. With this PR, we split out and display individual resource status (including execution state, errors, last refresh time, and next refresh time).

This PR builds off of #3786, which adds ShadCN's Data Table component set.

Here's an example of what the new page looks like:
image


web-admin/src/features/projects/status/ProjectParseErrors.svelte
```diff
@@ -0,0 +1,56 @@
...
+  import Spinner from "@rilldata/web-common/features/entity-management/Spinner.svelte";
+  import {
+    ResourceKind,
+    SingletonProjectParserName,
+  } from "@rilldata/web-common/features/entity-management/resource-selectors";
+  import { EntityStatus } from "@rilldata/web-common/features/entity-management/types";
+  import { createRuntimeServiceGetResource } from "@rilldata/web-common/runtime-client";
+  import { runtime } from "@rilldata/web-common/runtime-client/runtime-store";
+
+  $: parserErrors = createRuntimeServiceGetResource(
```
AdityaHegde, line 11: The intent here is to only show parse errors and not reconcile errors right? Because in my project this shows no errors but there are a couple resources with reconcile errors.

web-admin/src/features/projects/status/display-utils.ts
```diff
@@ -0,0 +1,31 @@
+import { V1ReconcileStatus } from "@rilldata/web-common/runtime-client";
+
+export function prettyReconcileStatus(status: V1ReconcileStatus) {
```
AdityaHegde, line 3: Would javascript objects as maps be cleaner than these switch cases?

web-common/src/components/tag/Tag.svelte
```diff
@@ -1,14 +1,39 @@
...
+    | "amber"
+    | "blue"
+    | "purple";
+
+  export let color: Color = "gray";
 
   function getColorClass(color: string) {
     switch (color) {
-      case "default":
+      case "gray":
```
AdityaHegde, line 17: Same thing here. Feels like a map will be better here.

web-admin/src/features/projects/status/ProjectResourcesTable.svelte
```diff
@@ -0,0 +1,116 @@
...
+<div class="rounded-md border">
+  <Table.Root {...$tableAttrs}>
+    <Table.Header>
+      {#each $headerRows as headerRow}
+        <Subscribe rowAttrs={headerRow.attrs()}>
+          <Table.Row>
+            {#each headerRow.cells as cell (cell.id)}
+              <Subscribe attrs={cell.attrs()} let:attrs props={cell.props()}>
+                <Table.Head {...attrs}>
+                  <Render of={cell.render()} />
```
AdityaHegde, line 92: This doesnt feel like svelte, more like react. Does the library not support slots?

AdityaHegde (approved):
Small comments but looks good otherwise.

web-admin/src/features/projects/status/ProjectParseErrors.svelte
ericpgreen2: Correct – those (resource-specific) reconcile errors should be shown in the table above in the "Error" column. Do you see them there?

web-admin/src/features/projects/status/ProjectResourcesTable.svelte
ericpgreen2: It doesn't seem to. All examples use `<Render />` and the [docs](https://svelte-headless-table.bryanmylee.com/docs/getting-started/quick-start#applying-to-markup) show this info box:
image

web-admin/src/features/projects/status/display-utils.ts
ericpgreen2: Could definitely go either way on it. In these cases (small sets) I slightly prefer the simplicity & encapsulation of the `switch` statement.

Issue #3679, opened 2023-12-11
Make sure we're type checking non-svelte files in CI (TSC checks)


bcolloran:
I just realized that the `--no-tsconfig` option to svelte-check means we're not type checking any of our plain `.ts` files :'(  Need to rectify that.

# UPDATES
- 2024-01-09 : added initial tsc check with whitelist to allow very common error codes: https://github.com/rilldata/rill/pull/3785

# LATEST  STATUS (2024-01-09)

running `npx tsc --noEmit | grep "error TS" | sed -n 's/.*\(error TS[0-9]*:\).*/\1/p' | sort | uniq -c | sort -nr`

shows that we have the following number of each TS error code:
```
    194 error TS18048:
    167 error TS2345:
     98 error TS2322:
     41 error TS18047:
     23 error TS2532:
     17 error TS2339:
     13 error TS2538:
     11 error TS2769:
     11 error TS18046:
      1 error TS2614:
```

and `npx tsc  --noEmit` shows that we have the following number of errors per file:
```
Found 576 errors in 90 files.

Errors  Files
    37  web-admin/src/features/dashboards/listing/selectors.ts:44
     5  web-admin/src/features/errors/error-utils.ts:107
     1  web-admin/src/features/projects/selectors.ts:27
     9  web-admin/src/features/scheduled-reports/selectors.ts:32
    13  web-admin/src/features/view-as-user/clearViewedAsUser.ts:25
    14  web-common/src/components/button-group/ButtonGroup.spec.ts:73
     8  web-common/src/components/data-graphic/actions/outline.ts:20
     1  web-common/src/components/data-graphic/marks/segment.ts:65
    25  web-common/src/components/date-picker/datetime.ts:26
     1  web-common/src/components/date-picker/util.ts:4
     2  web-common/src/components/editor/indent-guide/index.ts:59
     3  web-common/src/components/editor/line-status/line-number-gutter.ts:58
     3  web-common/src/components/editor/line-status/line-status-gutter.ts:52
     1  web-common/src/components/editor/line-status/state.ts:35
     2  web-common/src/components/notifications/notificationStore.ts:86
    14  web-common/src/features/dashboards/actions/index.ts:16
     4  web-common/src/features/dashboards/dashboard-utils.ts:13
     1  web-common/src/features/dashboards/dimension-table/dimension-table-export-utils.ts:23
     5  web-common/src/features/dashboards/dimension-table/dimension-table-utils.ts:218
[...]

PR #3785, opened 2024-01-05
add tsc check to CI

bcolloran:
part of #3679, adds typescript checks to CI to check TS error codes, and whitelists some error codes so that we can adopt checks incrementally.

also fixes this first batch of low hanging fruit error codes:
```
  8 error TS2454:
      8 error TS2304:
      7 error TS2582:
      7 error TS2554:
      3 error TS2741:
      3 error TS2722:
      3 error TS2353:
      2 error TS2349:
      1 error TS2614:
      1 error TS2531:
      1 error TS2464:
      1 error TS2366:
      1 error TS2352:
      1 error TS2307: 
```


web-common/src/components/data-graphic/marks/segment.ts
```diff
@@ -1,39 +1,50 @@
+function pathIsDefined(yAccessor: string) {
```
bcolloran, line 1: this function is only ever needed in this file, moving it here

web-common/src/components/data-graphic/marks/segment.ts
```diff
@@ -1,39 +1,50 @@
...
+    );
+  };
+}
+
 /**
  * Helper function to compute the contiguous segments of the data
  * based on https://github.com/pbeshai/d3-line-chunked/blob/master/src/lineChunked.js
  */
-export function computeSegments(lineData, defined, isNext = () => true) {
+export function computeSegments(
```
bcolloran, line 16: Note that `computeSegments` is only ever called in one place, and the optional argument `isNext = () => true` was not given. `isNext` was only used in the chunk
```
      if (isNext(lastDatum, d)) {
        ...
      } else {
       ...
     }
```

Since `isNext` always returns true, the if block is always used and the else block is never used.

Additionally, `lastDatum` is only used as an argument to `isNext`, so it's not need either.


Finally, the `defined` arg to this fn is `pathIsDefined(yAccessor)` at the only call site, and `pathIsDefined` is never used anywhere other than at that call site, so pulling that into this file and inlining it in this fn

bcolloran:
Oh by the way @ericpgreen2, when you're reviewing this pull request, it may be helpful to look at the individual commits, which describe the reasoning behind the changes I was making when I was cleaning things up

ericpgreen2 (approved):
LGTM!

PR #3786, opened 2024-01-06
Add ShadCN's Data Table component set

ericpgreen2:
This PR adds [ShadCN Svelte's "Data Table"](https://www.shadcn-svelte.com/docs/components/data-table) component set to `web-common/components/table-shadcn`. The PR includes a basic Storybook story to show it in action.

This PR will be followed by another that uses this component on the Cloud UI project status page.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24323,6 +24323,20 @@
         "dequal": "^2.0.2"
       }
     },
+    "node_modules/svelte-headless-table": {
+      "version": "0.18.1",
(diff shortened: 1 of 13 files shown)


briangregoryholmes:
Looks good to me. Would you want to keep this pure or could I extend these components for use elsewhere in the application?

briangregoryholmes (approved):

ericpgreen2:
> Looks good to me. Would you want to keep this pure or could I extend these components for use elsewhere in the application?

I didn't modify/style these ShadCN components at all. I'm game if we want to customize them a bit. If possible, we should just avoid too much style proliferation (though control-plane tables and data-plane tables probably differ).

Issue #3547, opened 2023-11-23
Internationalization Support  for the UI (i18n)


fredguth:
Problem Description:
I am using Rill and have encountered a significant barrier for non-English speaking users. Many of my users are Portuguese speakers, and they face difficulties as the Rill UI is currently available only in English. This language barrier hinders the usability and accessibility of Rill for a broader audience.

Proposed Solution:
I propose the adoption of Paraglide.js as an internationalization (i18n) library. Paraglide.js is a new library known for its efficiency (due to Message Functions tree shaking) and ease of integration. By integrating Paraglide.js, we can make Rill more accessible to non-English speakers, starting with Portuguese localization.

Implementation Suggestion:

Integrate Paraglide.js: Begin by integrating Paraglide.js into Rill's frontend (starting by web-local/web-common and then web-admin).

Localization Files: Create localization files for Portuguese, allowing for future expansion to other languages.

Refactor UI Codebase: Update the UI codebase to utilize Paraglide.js, replacing hardcoded English text with references to the localization files.

Community Contribution: Encourage the community to contribute translations for other languages, enhancing Rill's global usability.
Impact of the Feature:

This feature will significantly improve Rill's accessibility for non-English speaking users and open up opportunities for broader usage globally. It aligns with the ethos of open-source projects by fostering inclusivity and diversity.

Additional Considerations:

Testing: Ensure thorough testing of the internationalized UI to maintain a consistent user experience across different languages.
Documentation: Update documentation to reflect the new internationalization feature and guide contributors on adding more languages.

I believe that this feature will be a valuable addition to Rill and will greatly enhance its appeal and usability worldwide. I look forward to discussing this further with the community and contributors.

ericpgreen2:
@fredguth, thanks very much for opening this issue and pushing this forward. A few initial thoughts/questions:

- Paraglide looks pretty neat. My one hesitation is that it’s very new (still considered a [prerelease](https://inlang.com/m/gerre34r/library-inlang-paraglideJs?view=changelog)). What alternatives did you consider? Can you speak to the pros/cons compared to established alternatives?
- To best manage community-contributed translations, we should setup a l10n management platform. Do you have experience with these tools and have any recommendations?
- For Rill Developer, the “Add Source” form uses text that comes from the backend. At some point, we’ll need a i18n solution for the backend as well. Are your users mainly dashboard consumers or additionally dashboard creators? If only dashboard consumers, we can consider the backend another time.

[...]

PR #3788, opened 2024-01-08
[do not merge] feat/i18n brazilian portuguese

fredguth:
## Checklist
- [ ] Manual verification
- [ ] Unit test coverage
- [ ] E2E test coverage
- [ ] Needs manual QA?

#### Issue addressed: 

#3547
#### Details:

This PR intention is to present a minimal working example of how this feature might work.
- installed svelte-i18n package in web-common
- used svelte-i18n in welcome message ("i18n" string shows it is reading from json file)
- used svelte-i18n in ProjectTitle but it seems to be overwritten by yaml file specification.
- had some difficulties understanding the monorepo, so please let me know if I need to change anything.
- there is a hardcoded reference to https://github.com/rilldata/rill.git in the project and I could not run it when the remote was my fork
- change browser language to Brazilian Portuguese and check project main page
## Steps to Verify

./rill devtool start localcode co


web-common/src/layout/index.ts
```diff
@@ -1 +1,12 @@
...
+import '../lib/i18n' // Import to initialize. Important :)
+import { locale, waitLocale } from 'svelte-i18n'
+import type { LayoutLoad } from './$types'
+
+export const load: LayoutLoad = async () => {
+	if (browser) {
+		locale.set(window.navigator.language)
+	}
+	await waitLocale()
+}
```
ericpgreen2, line 12: 1. Can you encapsulate all the i18n initialization logic (both the logic in this file and the logic in `lib/i18n/index.ts`) into one function `initializei18n()` and call the function from the top of `web-local/src/routes/+layout.svelte`?
2. No need for any checks for `if(browser)`. Because we're running SvelteKit in SPA-mode, there's no SSR and `browser` will always be `true`.

web-common/src/lib/i18n/index.ts
```diff
@@ -0,0 +1,13 @@
...
+import { init, register } from 'svelte-i18n'
+
+const defaultLocale = 'en'
+
+register('en', () => import('./locales/en.json'))
+register('pt-BR', () => import('./locales/pt-BR.json'))
+
+init({
+	fallbackLocale: defaultLocale,
+	initialLocale: browser ? window.navigator.language : defaultLocale,
```
ericpgreen2, line 12: Again, no need for the `browser` check – it'll always be `true`

ericpgreen2 (changes requested):
Fred, this is definitely on the right track – great to see it working! I've left some comments inline.

On your comment about "Untitled Rill Project", I've found that `web-common/src/features/welcome/constants.ts` has an `EMPTY_PROJECT_TITLE` constant, and it looks like we're not using that everywhere we should.

web-common/package.json
```diff
@@ -70,5 +70,8 @@
     "yaml": "^2.1.3",
     "yup": "^0.32.11"
   },
-  "type": "module"
+  "type": "module",
+  "dependencies": {
+    "svelte-i18n": "^4.0.0"
+  }
```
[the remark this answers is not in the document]
fredguth: That is right.  I usually do that as well, I guess I just followed the svelte-i18n tutorial and didn't pay attention to this point.  Nice catch!

web-common/src/lib/i18n/locales/en.json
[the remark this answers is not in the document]
fredguth: No problem. But my usecase is for developer-facing text, though. I can keep it in a fork.  Is there any hint in the code where something in web-common is developer-facing?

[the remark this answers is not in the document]
fredguth: ok!

web-common/src/layout/index.ts
fredguth: ok, I started this way and then afterwards though I would only need translation in web-common.  I can make the change.

fredguth:
I was wondering if there will be anywhere in the UI for the user to change the preferred language.  At the moment we show the window.navigator.language and the user can't change it in the webapp.

web-common/src/layout/index.ts
ericpgreen2: The function can live in `web-common` – then we can reuse it easily in the other apps (`web-admin`, `web-auth`). Just moving its invocation to `+layout.svelte` will make it very explicit that it's one of the functions that's called on startup.

ericpgreen2:
> I was wondering if there will be anywhere in the UI for the user to change the preferred language. At the moment we show the window.navigator.language and the user can't change it in the webapp.

That's a great question. A language setting is something we should add to the app at some point. We don't yet have a design for this. For the time being, mind if we cut scope and rely on the user's browser setting?

nishantmonu51:
Thanks @fredguth for the PR, closing this PR as of now due to inactivity. Please reopen it when ready to proceed.

PR #3806, opened 2024-01-09
add query retries in case druid returns 429

pjain1:


runtime/pkg/common/retryutil.go
```diff
@@ -0,0 +1,53 @@
...
+			return nil
+		}
+
+		if !shouldRetry(err) {
+			return err
+		}
+
+		sleepMillis := nextRetrySleepMillis(try)
+
+		time.Sleep(time.Duration(sleepMillis) * time.Millisecond)
```
begelundmuller, line 44: This has an issue that `ctx` cancellation will not cancel the sleep, so it may hang until retry before checking the context. Other places, we use the `github.com/eapache/go-resiliency/retrier` package for retries, might be worth using that instead – it has a `RunCtx` function which checks for cancellation during sleeping periods.

begelundmuller (changes requested):

runtime/pkg/common/retryutil.go
pjain1: changed to use that. only thing is the fuzzy multiplier here tries to prevents retry hotspots, not sure how practical it is though.

begelundmuller (approved):
LGTM

PR #3809, opened 2024-01-09
Begelundmuller/fix racy runtime test

begelundmuller:
Fixes the race condition that repeatedly causes race condition tests to fail on merge to main (e.g. [this one](https://github.com/rilldata/rill/actions/runs/7462333741/job/20304396670))


begelundmuller:
Replaced by https://github.com/rilldata/rill/pull/3810

PR #3810, opened 2024-01-09
Runtime: fix failing race condition test

begelundmuller:
Fixes the race condition that repeatedly causes race condition tests to fail on merge to main (e.g. [this one](https://github.com/rilldata/rill/actions/runs/7462333741/job/20304396670))

### DIFF
--- a/runtime/registry.go
+++ b/runtime/registry.go
@@ -122,6 +122,7 @@ type registryCache struct {
 }
 
 type instanceWithController struct {
+	instanceID    string
 	instance      *drivers.Instance
 	controller    *Controller
 	controllerErr error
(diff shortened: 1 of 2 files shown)


runtime/registry.go
```diff
@@ -261,7 +262,10 @@ func (r *registryCache) add(inst *drivers.Instance) {
 		panic(fmt.Errorf("instance %q already open", inst.ID))
 	}
 
-	iwc := &instanceWithController{instance: inst}
+	iwc := &instanceWithController{
+		instanceID: inst.ID,
```
k-anshul, line 266: Why is `instanceID` separately required ?

k-anshul (approved):

PR #3811, opened 2024-01-09
druid group by alias fix

pjain1:
Druid does not support group by alias so if the `column` or even `expression` now is different from the `name` in the dimension definition, Druid throws error like ` Column '<name>' not found in any table`.


nishantmonu51:
Not a blocker for this PR: is there a quick test that we can plan around this ?

begelundmuller (approved):
LGTM

PR #3814, opened 2024-01-09
Dashboard: Remove calls to `TableColumns` API

ericpgreen2:
Addresses the second bullet here: https://github.com/rilldata/rill/issues/3803


djbarnwal (changes requested):
Getting these warnings in the console - 

image

Extra props are being passed to `NoTimeDimensionCTA`.

Also saw a reference of  `createQueryServiceTableColumns` being used in `RowViewer` component used in the dashboard. It's behind a feature flag in the same manner as this component. Do we have to change that as well?

ericpgreen2:
> Extra props are being passed to `NoTimeDimensionCTA`.

Ah, thanks for catching that. Just fixed.
 
> Also saw a reference of `createQueryServiceTableColumns` being used in `RowViewer` component used in the dashboard. It's behind a feature flag in the same manner as this component. Do we have to change that as well?

We'd really only need to fix it once we promote the feature to the Cloud UI, but I just fixed it now anyways so we don't forget.

djbarnwal (changes requested):
Seeing some issues with the Rows table.

1. Opening the table throws a 429 error. 
2. Selecting a leaderboard value after opening the table crashes the app

image

djbarnwal (approved):
Tested locally, works as expected! 
Approving so that you can merge after resolving the merge conflicts.

PR #3815, opened 2024-01-10
Frontend integration of measure filters 

AdityaHegde:
Reference design doc: https://www.notion.so/rilldata/Measure-Filters-1c408bb456e64e2284392847d700b130


web-common/src/features/dashboards/filters/measure-filters/MeasureFilterMenu.svelte
```diff
@@ -0,0 +1,170 @@
...
+    initialValues: {
+      dimension: dimensionName,
+      operation,
+      value1,
+      value2,
+    },
+    validationSchema: yup.object({
+      dimension: yup.string().required("Required"),
+      operation: yup.string().required("Required"),
+      value1: yup.number().required(),
```
mindspank, line 70: NIT: Form error validation looks kinda ugly as it takes the property name

image

web-common/src/components/forms/Select.svelte
```diff
@@ -18,15 +22,15 @@
 </script>
 
 <div class="flex flex-col gap-y-2">
```
briangregoryholmes, line 24: The Figma includes some placeholder text for these Select components: https://www.figma.com/file/Qt6EyotCBS3V6O31jVhMQ7/RILL?type=design&node-id=7116-382806&mode=design&t=8UeglpfHSPS7bIzz-0

web-common/src/features/dashboards/filters/measure-filters/MeasureFilterMenu.svelte
```diff
@@ -0,0 +1,170 @@
...
+  <div class="p-4">
+    <form
+      autocomplete="off"
+      class="flex flex-col gap-y-3"
+      id="measure"
+      on:submit|preventDefault={handleSubmit}
+    >
+      <Select
+        bind:value={$form["dimension"]}
+        detach
```
briangregoryholmes, line 130: Should these have a more explicit `on:change` trigger like the Input component? There's a reactive statement on line 103 that tracks the `$form["operation"]` value, but not the `$form["dimension"]` value.

briangregoryholmes (changes requested):
Looks like the right approach overall. There are a few nits, but a couple important functional issues that I'll describe here:

1. Start to add a measure filter, but fill out the value before selecting the dimension or threshold. The filter will not be added.
2. Add a measure filter. Select a different dimension option and close out of the menu. You'll see the measure does not update.
3. Add a measure filter. Change the Dimension to something else and then update the value. The visible measure pill does not update, but the dashboard does. If you close out of the menu and attempt to remove the pill, you'll see that the two filters appear to be "stacked" on top of each other.

web-common/src/features/dashboards/filters/measure-filters/MeasureFilter.svelte
```diff
@@ -0,0 +1,91 @@
...
+  function handleDismiss() {
+    if (!expr) {
+      dispatch("remove");
+    } else {
+      active = false;
+    }
+  }
+</script>
+
+<WithTogglableFloatingElement
```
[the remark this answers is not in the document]
AdityaHegde: Ya I am hoping we will do a refactor at once. Sadly this feature is a blocker for alerts so we can take the refactor up as a follow up.

AdityaHegde:
Good catch! Have fixed the issues. We really should move to shadcn's select components. The ones we have are lacklustre.

briangregoryholmes (approved):

PR #3816, opened 2024-01-10
Fixing the sort type in export calls

AdityaHegde:
`sortType` is the correct type for sorting. This was changed when we changed the API but missed changing these places in UI.


k-anshul (approved):
Approving after experience pass. 
Code review not done.

PR #3817, opened 2024-01-10
Fix POST queries to honour the abort signal

### LINKED ISSUE #3710: Query cancellation not working
Evident when clicking around on a Druid-backed dashboard and looking at the browser's network tab. It's also evident when looking at a large volume DuckDB-backed dashboard.

[Reported in Slack](https://rilldata.slack.com/archives/C0415QK8DK6/p1702667416409829)

AdityaHegde:
closes #3710

## Checklist
- [x] Manual verification
- [ ] Unit test coverage
- [ ] E2E test coverage
- [x] Needs manual QA?

#### Issue addressed: 
The queries in dashboard are not being cancelled when changing the filters and other settings. This is apparent while using druid or a large project.

#### Details:
We actually call cancel queries. But the orval generated code is not passing on the abort signal to our client. So cancellations are not honoured for these. The root cause is that these queries are POST requests overridden to be queries instead of mutations. This seems to have an edge case in orval of not adding signal properly.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,13 +1,27 @@
 {
+  // TODO: SvelteKit recommends extending their config, but it currently triggers type import errors throughout application that we need to fix
+  // "extends": "./.svelte-kit/tsconfig.json",
   "compilerOptions": {
     "moduleResolution": "node",
     "module": "es2020",
     "lib": ["es2020", "DOM"],
     "target": "es2020",
+    /**
+      svelte-preprocess cannot figure out whether you have a value or a type, so tell TypeScript
+      to enforce using \`import type\` instead of \`import\` for Types.
+    */
     "importsNotUsedAsValues": "error",
+    /**
+      ignoreDeprecations is recommended by the SvelteKit maintainers until SvelteKit 2.0.
+      See: https://github.com/sveltejs/kit/issues/8650#issuecomment-1607282327
+    */
     "ignoreDeprecations": "5.0",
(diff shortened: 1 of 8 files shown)


tsconfig.json
ericpgreen2, line 1: Do we have to remove the comments in this file?

web-common/src/runtime-client/add-signal-argument.ts
ericpgreen2, line 1: I really like how this is encapsulated into this one file. To help our future selves – Could you add a comment at the top of this file to 1) explain the motivation, and 2) describes the input / output of the transformation?

ericpgreen2 (commented):
Overall, I like the encapsulation of the transformation. Some comments inline, and also:
- How do you feel about how resilient it is to time?
- Did you check with Orval community to see if anyone had run into this issue before? I think the maintainer is quite responsive, so creating a quick issue could be good.

tsconfig.json
AdityaHegde: Nope. This was a temporary change.

AdityaHegde:
It is not guaranteed to be stable with newer versions. I have fixed the version of orval for now.

Ideally there should be a fix from orval. Created a discussion for now: https://github.com/anymaniax/orval/discussions/1151

ericpgreen2:
In trying to figure out if query cancellation is working, I found myself in this state after navigating quickly away from a loading dashboard, then returning to it:

image

PR #3818, opened 2024-01-10
Add kubernetes provisioner and provisioner set support

kaspersjo:
This PR rewrites the `provisioner` package to support multiple provisioners and adds a new Kubernetes provisioner type to dynamically provision runtimes.


himadrisingh:
Lets resolve the conflicts and push the latest changes.

admin/provisioner/kubernetes.go
```diff
@@ -0,0 +1,303 @@
...
+	svc := &apiv1.Service{}
+	sts := &appsv1.StatefulSet{}
+
+	// Resolve the templates and decode into Kubernetes API resources
+	for k, v := range map[string]any{
+		p.Spec.TemplatePaths.HTTPIngress: httpIng,
+		p.Spec.TemplatePaths.GRPCIngress: grpcIng,
+		p.Spec.TemplatePaths.Service:     svc,
+		p.Spec.TemplatePaths.StatefulSet: sts,
+	} {
```
begelundmuller, line 138: I'm wondering if we could avoid the external templated files and instead just define the base resources directly in code, and override the relevant properties here. Or are the specs pretty internal/custom?

begelundmuller (changes requested):
This PR looks very good! Put some comments below, apart from those, one other question – could we put some instructions for testing the Kubernetes provisioner in a `README.md` in the `admin/provisioner` directory. Optionally (I realize this might not be worth the trouble), could also add a test somewhere for the Kubernetes provisioner that starts Minikube using Testcontainers.

admin/admin.go
```diff
@@ -11,21 +12,23 @@ import (
 )
 
 type Options struct {
-	DatabaseDriver  string
-	DatabaseDSN     string
-	ProvisionerSpec string
-	ExternalURL     string
+	DatabaseDriver     string
+	DatabaseDSN        string
+	ProvisionerSet     string
```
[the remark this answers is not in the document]
kaspersjo: Agree - done

admin/admin.go
```diff
@@ -54,21 +57,27 @@ func New(ctx context.Context, opts *Options, logger *zap.Logger, issuer *auth.Is
...
+	// Create provisioner set
+	provSet, err := provisioner.NewSet(opts.ProvisionerSet, db, logger)
 	if err != nil {
 		return nil, err
 	}
 
+	// Verify that the specified default provisioner is in the provisioner set
+	_, ok := provSet[opts.DefaultProvisioner]
+	if !ok {
+		return nil, fmt.Errorf("default provisioner '%s' is not in the provisioner set", opts.DefaultProvisioner)
```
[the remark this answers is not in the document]
kaspersjo: Neat - I've updated all the occurrences.

admin/database/postgres/migrations/0021.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE deployments
+ADD COLUMN provisioner TEXT DEFAULT 'static' NOT NULL,
+ADD COLUMN provision_id TEXT,
+ADD COLUMN runtime_version TEXT;
```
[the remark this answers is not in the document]
kaspersjo: Agree - done

admin/deployments.go
```diff
@@ -36,13 +40,53 @@ func (s *Service) createDeployment(ctx context.Context, opts *createDeploymentOp
...
+	if !ok {
+		return nil, fmt.Errorf("provisioner '%s' is not in the provisioner set", opts.Provisioner)
+	}
+
+	// Resolve runtime version
+	var runtimeVersion string
+	if opts.ProdVersion == "latest" {
+		// Resolve latest version from config
+		if opts.VersionNumber != "" {
+			runtimeVersion = opts.VersionNumber
```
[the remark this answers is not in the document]
kaspersjo: Changed the logic slightly here to support `latest` version and added SerVer validation for the version coming form the `ldflags`.

admin/deployments.go
```diff
@@ -36,13 +40,53 @@ func (s *Service) createDeployment(ctx context.Context, opts *createDeploymentOp
...
+		// Verify provided semVer is valid
+		_, err := version.NewVersion(opts.ProdVersion)
+		if err != nil {
+			return nil, err
+		}
+		runtimeVersion = opts.ProdVersion
+	}
+
+	// Create provision ID
+	provisionID := strings.ReplaceAll(uuid.New().String(), "-", "")
```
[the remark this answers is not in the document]
kaspersjo: Make sense - done

admin/deployments.go
```diff
@@ -36,13 +40,53 @@ func (s *Service) createDeployment(ctx context.Context, opts *createDeploymentOp
...
-	alloc, err := s.Provisioner.Provision(ctx, &provisioner.ProvisionOptions{
-		OLAPDriver: opts.ProdOLAPDriver,
-		Slots:      opts.ProdSlots,
-		Region:     opts.Region,
+	alloc, err := p.Provision(ctx, &provisioner.ProvisionOptions{
+		ProvisionID:    provisionID,
+		RuntimeVersion: runtimeVersion,
+		OLAPDriver:     opts.ProdOLAPDriver,
+		Slots:          opts.ProdSlots,
+		Annotations:    opts.Annotations.toMap(),
```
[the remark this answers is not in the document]
kaspersjo: Removed - it was passed to the old provisioner, so I just kept it as is, but I took a look and it doesn't seem to be used for anything.

admin/projects.go
```diff
@@ -131,7 +134,7 @@ func (s *Service) TeardownProject(ctx context.Context, p *database.Project) erro
 // UpdateProject updates a project and any impacted deployments.
 // It runs a reconcile if deployment parameters (like branch or variables) have been changed and reconcileDeployment is set.
 func (s *Service) UpdateProject(ctx context.Context, proj *database.Project, opts *database.UpdateProjectOptions) (*database.Project, error) {
-	requiresReset := (proj.Region != opts.Region) || (proj.ProdSlots != opts.ProdSlots)
+	requiresReset := (proj.Provisioner != opts.Provisioner) || (proj.ProdSlots != opts.ProdSlots) || (proj.ProdVersion != opts.ProdVersion)
```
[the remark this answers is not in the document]
kaspersjo: Updated the deployment logic to include the ability to use the provisioner to update/patch the image version without resetting the project. It's used in a new `upgrade_latest_version_projects` job, so that should take care of smoothly rolling out a new version without a lot of downtime.

In `UpdateProject` I think it's still the safest option to keep resetting the project on a version change, since we don't know if it's a version upgrade, downgrade or potentially a pre-release, so resetting the project just seems like the safest options for now and then we can refine this later on.

admin/provisioner/kubernetes.go
kaspersjo: I thought quite a bit about this when deciding on the templates and I totally understand the desire to have the resources directly defined in code, but there's just too many environment specific custom configurations of the Kubernetes resources needed for it to really be a viable option.

admin/projects.go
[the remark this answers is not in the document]
begelundmuller: Okay that makes sense. As long as regular releases don't lead to `proj.ProdVersion`?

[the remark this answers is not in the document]
kaspersjo: A release will never lead to changes to `proj.ProdVersion` by itself, the default version will be `latest` and in that case a release will trigger a deployment update, which in turn will bump the provisioned runtime to the latest release version (using the corresponding SemVar) without resetting the project.

admin/provisioner/kubernetes.go
```diff
@@ -250,8 +251,18 @@ func (p *KubernetesProvisioner) AwaitReady(ctx context.Context, provisionID stri
...
+		return err
+	}
+
+	// As a final step we make sure the runtime can be reached, we retry on failure, to account for potential small delays in network config propagation
+	retryClient := retryablehttp.NewClient()
+	retryClient.RetryMax = 3
+	retryClient.RetryWaitMin = 1 * time.Second
+	retryClient.RetryWaitMax = 3 * time.Second
+	retryClient.Logger = nil // Disable inbuilt logger
+	_, err = retryClient.Get(p.getHost(provisionID) + "/v1/ping")
```
begelundmuller, line 263: Use `url.JoinPath` to avoid concerns about whether the host has a trailing slash or not

admin/worker/upgrade_latest_version_projects.go
```diff
@@ -0,0 +1,93 @@
...
+		if len(projs) != 0 {
+			afterName = projs[len(projs)-1].Name
+		}
+
+		// Process batch
+		for _, proj := range projs {
+			err := w.upgradeAllDeploymentsForProject(ctx, proj, latestVersion)
+			if err != nil {
+				// We log the error, but continues to the next deployment
+				w.logger.Error("upgrade latest version projects: failed to upgrade project deployments", zap.String("project", proj.ID), zap.String("version", latestVersion), observability.ZapCtx(ctx))
```
begelundmuller, line 50: Use `project_id` instead of `project` for the key – for consistency with other logs, so searching is easier :)

admin/worker/upgrade_latest_version_projects.go
```diff
@@ -0,0 +1,93 @@
...
+
+	// Get project organization, we need this to create the deployment annotations
+	org, err := w.admin.DB.FindOrganization(ctx, proj.OrganizationID)
+	if err != nil {
+		return err
+	}
+
+	for _, depl := range depls {
+		if depl.RuntimeVersion != latestVersion {
+			w.logger.Info("upgrade latest version projects: upgrading deployment", zap.String("deployment", depl.ID), zap.String("provision", depl.ProvisionID), zap.String("instance", depl.RuntimeInstanceID), zap.String("version", latestVersion), observability.ZapCtx(ctx))
```
begelundmuller, line 73: Same – `deployment_id`, `provision_id`, `instance_id`

admin/worker/upgrade_latest_version_projects.go
```diff
@@ -0,0 +1,93 @@
...
+				Variables:       proj.ProdVariables,
+				Annotations:     w.admin.NewDeploymentAnnotations(org, proj),
+				EvictCachedRepo: false,
+			})
+			if err != nil {
+				w.logger.Error("upgrade latest version projects: failed to upgrade deployment", zap.String("deployment", depl.ID), zap.String("version", latestVersion), observability.ZapCtx(ctx))
+				return err
+			}
+
+			w.logger.Info("upgrade latest version projects: upgraded deployment", zap.String("deployment", depl.ID), zap.String("version", latestVersion), observability.ZapCtx(ctx))
```
begelundmuller, line 88: A few other cases here

admin/projects.go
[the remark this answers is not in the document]
begelundmuller: Perfect, that sounds great

admin/worker/upgrade_latest_version_projects.go
kaspersjo: I just followed the naming convention from one of the other jobs? Should we change them to reflect the convention you mention here?
https://github.com/rilldata/rill/blob/a2614c0c4f300a2b702eeb2eaf2b83b70e0f7d50/admin/worker/reset_all_deployments.go#L48

admin/worker/upgrade_latest_version_projects.go
kaspersjo: Updated

admin/worker/upgrade_latest_version_projects.go
kaspersjo: Updated

admin/provisioner/kubernetes.go
kaspersjo: Done

admin/worker/upgrade_latest_version_projects.go
kaspersjo: Updated - I also updated `reset_all_deployments` to use the same convention

admin/provisioner/kubernetes.go
```diff
@@ -260,7 +261,11 @@ func (p *KubernetesProvisioner) AwaitReady(ctx context.Context, provisionID stri
 	retryClient.RetryWaitMin = 1 * time.Second
 	retryClient.RetryWaitMax = 3 * time.Second
 	retryClient.Logger = nil // Disable inbuilt logger
-	_, err = retryClient.Get(p.getHost(provisionID) + "/v1/ping")
+	pingURL, err := url.JoinPath(p.getHost(provisionID), "/v1/ping")
+	if err != nil {
+		return err
+	}
+	_, err = retryClient.Get(pingURL)
```
begelundmuller, line 268: If you follow the docs through to the underlying `net/http` call, I think you might find that it is required to get the response and call `.Close()` on the response body. See docstring here: https://pkg.go.dev/net/http#Get

kaspersjo: Added - good catch

admin/provisioner/kubernetes.go
kaspersjo: I've added a `README.md` with some initial config instructions and some simplified template examples, in my opinion this should be enough for now and internal Rill people can refer to our private infra repos for our internal configurations and templates. Creating a simple developer workflow for the Kubernetes part of the provisioner is hard and it's something we should tackle as a separate task if we feel it's worth the effort to support.

begelundmuller (approved):

PR #3826, opened 2024-01-11
hide navbar when browser width drops below 768

briangregoryholmes:
This PR hides the `Navigation` sidebar when the browser width drops below 768px. This only happens when the window is being made smaller.


ericpgreen2 (approved):
Nice touch 👍

PR #3829, opened 2024-01-11
Adding security policy when no include clauses match

AdityaHegde:
When there are some include clauses and none of them match, we end up including all measures/dimensions.

Adding a boolean `ExcludeAll` that is set when some include conditions are present but non are met.

### DIFF
--- a/runtime/server/controller.go
+++ b/runtime/server/controller.go
@@ -320,11 +320,20 @@ func (s *Server) applySecurityPolicyMetricsView(ctx context.Context, instID stri
 
 // applySecurityPolicyIncludesAndExcludes rewrites a metrics view based on the include/exclude conditions of a security policy.
 func (s *Server) applySecurityPolicyMetricsViewIncludesAndExcludes(mv *runtimev1.MetricsViewV2, policy *runtime.ResolvedMetricsViewSecurity) (*runtimev1.MetricsViewV2, bool) {
-	if policy == nil || (len(policy.Include) == 0 && len(policy.Exclude) == 0) {
+	if policy == nil {
 		return mv, false
 	}
-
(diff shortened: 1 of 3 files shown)


runtime/server/controller.go
```diff
@@ -320,12 +320,21 @@ func (s *Server) applySecurityPolicyMetricsView(ctx context.Context, instID stri
...
 // applySecurityPolicyIncludesAndExcludes rewrites a metrics view based on the include/exclude conditions of a security policy.
 func (s *Server) applySecurityPolicyMetricsViewIncludesAndExcludes(mv *runtimev1.MetricsViewV2, policy *runtime.ResolvedMetricsViewSecurity) (*runtimev1.MetricsViewV2, bool) {
-	if policy == nil || (len(policy.Include) == 0 && len(policy.Exclude) == 0) {
+	if policy == nil {
 		return mv, false
 	}
-
 	mv = proto.Clone(mv).(*runtimev1.MetricsViewV2)
 
+	if policy.ExcludeAll {
```
pjain1, line 328: Instead of this flag, wouldn't it work if we just check `if len(policy.Include) == 0 && len(policy.Exclude) == 0 ` here

AdityaHegde: We wouldnt want to exclude everything if there were some exclude clauses but none of them evaluated to true.

AdityaHegde: Adding a `ExcludeAll` felt more explicit and controlled by the security module.

pjain1: ok got it

pjain1 (approved):

runtime/server/controller.go
pjain1: lets add a test case in follow up because the current test cases passes with this condition as well

PR #3836, opened 2024-01-11
remove whitespace pre-wrap in MultiMetricMouseoverLabel

briangregoryholmes:
In Safari, the inclusion of `whitespace: pre-wrap` appears to cause an issue where the value is placed incorrectly (see below). Removing this appears to maintain the intended placement in all browsers. However, it may have been added to address another visual issue.

Screenshot 2024-01-11 at 3 47 05 PM


djbarnwal (approved):
Don't recall why it was added there. Did QA on the charts and didn't find any bugs.

PR #3843, opened 2024-01-12
Cloud UI: Fix scroll issues

### LINKED ISSUE #3841: Fix scroll issue in Cloud UI

ericpgreen2:
This PR addresses scrolling issues in the Cloud UI. Previously, the `h-screen` class on the `<main>` tag combined with `VerticalScrollContainer`'s `h-full` class caused some content to be hidden. The issue stemmed from `VerticalScrollContainer` inheriting `h-screen` from its parent, yet not wrapping the navbar. This resulted in the navbar displacing the `VerticalScrollContainer` and its contents, obscuring content equivalent to the navbar's height.

Closes #3841

To verify:
- Ensure scroll works on the dashboard when there are many line charts & leaderboards
- Ensure scroll works on the org page and the project page


bcolloran (approved):
Looks good me @ericpgreen2 

FYI @briangregoryholmes -- I know you have ideas for improving our layout code, so just flagging this as background for an area where we think our current approach could use some tuning.

PR #3844, opened 2024-01-15
Fix aggregation on dimenison with timezone

AdityaHegde:
Adding timezone to dimension expression seems to break because it adds a query arg. This seems like a bug where `?` is not expanded properly, since manually using timezone instead of `?` works.

For now using numbered group by to fix the issue.

### DIFF
--- a/runtime/queries/metricsview_aggregation.go
+++ b/runtime/queries/metricsview_aggregation.go
@@ -437,6 +437,10 @@ func (q *MetricsViewAggregation) buildMetricsAggregationSQL(mv *runtimev1.Metric
 			return "", nil, err
 		}
 		selectCols = append(selectCols, fmt.Sprintf("%s as %s", expr, safeName(d.Name)))
+		// Using expr was causing issues with query arg expansion in duckdb.
+		// Using column name is not possible either since it will take the original column name instead of the aliased column name
+		// But using numbered group we can exactly target the correct selected column.
+		// Note that non-timestamp columns also use numbered group by for constancy.
 		groupCols = append(groupCols, fmt.Sprintf("%d", len(selectCols)))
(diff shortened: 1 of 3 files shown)


runtime/queries/metricsview_aggregation.go
```diff
@@ -427,7 +427,7 @@ func (q *MetricsViewAggregation) buildMetricsAggregationSQL(mv *runtimev1.Metric
 			if unnestClause != "" {
 				unnestClauses = append(unnestClauses, unnestClause)
 			}
-			groupCols = append(groupCols, safeName(d.Name))
+			groupCols = append(groupCols, fmt.Sprintf("%d", len(selectCols)))
```
begelundmuller, line 430: Can we add a comment here and on line 440 about what this does and why it's required. I think this fix might seem confusing/redundant to future maintainers.

runtime/queries/metricsview_aggregation.go
```diff
@@ -437,7 +437,11 @@ func (q *MetricsViewAggregation) buildMetricsAggregationSQL(mv *runtimev1.Metric
 			return "", nil, err
 		}
 		selectCols = append(selectCols, fmt.Sprintf("%s as %s", expr, safeName(d.Name)))
-		groupCols = append(groupCols, expr)
+		// Using expr was causing issues with query arg expansion in duckdb.
+		// Using column name is not possible either since it will take the original column name instead of the aliased column name
+		// But using numbered group we can exactly target the correct selected column.
+		// Note that non-timestamp columns also use numbered group by for constancy.
```
egor-ryashin, line 443: "*the* non-timestamp columns also use *the* numbered group-by"

egor-ryashin (changes requested):

PR #3847, opened 2024-01-15
Fix Use With Proxy

cwarden:
Include a hostname, localhost, when connecting to the GRPC endpoint.

Fixes bug in which `rill start` fails to start up when a proxy server is configured because rill sends an invalid request to the proxy.  Using mitmproxy, for example, will log the following error when rill tries to connect to the GRPC endpoint using only the port:
Bad HTTP request line: b'CONNECT :49009 HTTP/1.1'


begelundmuller (approved):
LGTM

PR #3748, opened 2023-12-25
Add Euro type

Fredehagelund92:
Please go the the `Preview` tab and select the appropriate sub-template:

* [Applications](?expand=1&template=applications_template.md)
* [Others](?expand=1&template=fallback_template.md)

PR #3850, opened 2024-01-15
Add euro format

bcolloran:
continuing the work started by @Fredehagelund92  in PR  https://github.com/rilldata/rill/pull/3748


web-common/src/lib/number-formatting/strategies/per-range.default-percent.spec.ts
```diff
@@ -39,14 +39,17 @@ const defaultGenericNumTestCases: [number, string][] = [
...
-  [0.1234686 / 100, "0.12%"],
 
+  // FIXME: rounding to 2 decimals not working as desired
+  // [999.994 / 100, "999.99%"], // ACTUALLY GETTING '1.0k%'
+  // [999.99 / 100, "999.99%"], // ACTUALLY GETTING '1.0k%'
+  // [999.1234686 / 100, "999.12%"], // ACTUALLY GETTING '999.1%'
+  // [789.1234686 / 100, "789.12%"], // ACTUALLY GETTING '789.1%'
+  // [89.1234686 / 100, "89.12%"], // ACTUALLY GETTING '89.1%'
+  // [9.1234686 / 100, "9.12%"], // ACTUALLY GETTING '9.1%'
+  // [0.1234686 / 100, "0.12%"], // ACTUALLY GETTING '0.1%'
```
bcolloran, line 50: @ericpgreen2 -- note that this has been entirely disabled for months. I've commented out individual cases that are not working exactly as desired (though they are small edge cases that have not been bothersome in practice), but I reckon it's better to have most of the cases running and these ones commented out with FIXME than to turn off the whole file

bcolloran:
@ericpgreen2 This PR adds a `currency_eur` format preset, and does some amount of the cleanup we talked about the other day without going too deep -- I can revisit if there are product needs that would point us in that direction.

web-common/src/lib/number-formatting/humanizer-types.ts
```diff
@@ -107,12 +137,44 @@ export type NumberParts = {
...
+
+  // not actually used for formatting, but needed to calculate the
+  // px sizes of maxWidthsInSample and maxWidthsPossible
+  // alignDecimal?: boolean;
+
+  // If `true`, use upper case "E" for exponential notation;
+  // If `false` or `undefined`, use lower case
+  upperCaseEForExponent?: boolean;
+};
+
```
ericpgreen2, line 173: What's going on with the options that are commented out? Remove?

ericpgreen2 (approved):
Nice, thanks @Fredehagelund92 for getting this started, and thanks @bcolloran for getting it over-the-line!

As @bcolloran and I discussed, this is a short-path to add the Euro. When adding many more currencies, we should think about a more scalable strategy/syntax.

PR #3856, opened 2024-01-16
Handle mouseover tooltip for non string dimension values

### LINKED ISSUE #3855: Dashboard hangs when comparing by numeric dimension and filtering by null value
**Describe the bug**
Dashboard will crash / hang when user attempts to compare by a numeric dimension, which contain null values, and then applies a filter on these null values (intentionally or unintentionally). Looking at the browser console, an `Uncaught (in promise) TypeError: Cannot read properties of null (reading 'length')` error will be thrown and the user won't be able to take any more actions nor will the dashboard be responsive until you refresh the page.

**To Reproduce**
1. Open a dashboard that contains some numeric dimensions (at least one of which has null values)
2. Apply some filters, change the time range, etc. (everything should work)
3. Switch the comparison to the numeric dimension that contains null values
4. Apply a filter on this same dimension for only `null`

**Expected behavior**
Dashboard should continue to be responsive and user can continue to interact / make changes. There should be no error thrown in the browser console (that requires a hard refresh at least).

**Actual behavior**
Dashboard stops being responsive and user cannot interact anymore without doing a refresh. There is an `Uncaught (in promise) TypeError: Cannot read properties of null (reading 'length')` error in the console.

**Screenshots**
Dev tools console

AdityaHegde:
closes #3855

### DIFF
--- a/web-common/src/features/dashboards/time-series/DimensionValueMouseover.svelte
+++ b/web-common/src/features/dashboards/time-series/DimensionValueMouseover.svelte
@@ -16,6 +16,8 @@
   $: x = point?.[xAccessor];
 
   function truncate(str) {
+    if (!str?.length) return str;
+
     const truncateLength = 34;
 
     if (str.length > truncateLength) {


djbarnwal (approved):
Thanks for the PR

PR #3857, opened 2024-01-16
Admin: change tags to annotations

begelundmuller:
Changes:

- Changes superuser project "tags" (string array) to "annotations" (string KV-pairs).
- The annotations are propagated to the runtime and set as attributes on *some* telemetry (such as events and request/response logs). 

Example usage:

```bash
rill sudo annotations set org project --annotation foo=bar --annotation hello=world
rill sudo project search --annotation foo=bar
rill sudo project search --annotation foo=*
```


admin/projects.go
```diff
@@ -133,9 +134,12 @@ func (s *Service) TeardownProject(ctx context.Context, p *database.Project) erro
 func (s *Service) UpdateProject(ctx context.Context, proj *database.Project, opts *database.UpdateProjectOptions) (*database.Project, error) {
 	requiresReset := (proj.Region != opts.Region) || (proj.ProdSlots != opts.ProdSlots)
 
+	log.Printf("HERE: projects=%v AND opts=%v", proj.Annotations, opts.Annotations)
```
kaspersjo, line 137: Some left behind debug log?

cli/cmd/sudo/annotations/tags.go
kaspersjo, line 1: Nitpick but `annotations.go` seems to be a better name?

kaspersjo (approved):
Two small comments, but otherwise looks good to me!

admin/projects.go
begelundmuller: Whoops 😅

cli/cmd/sudo/annotations/tags.go
begelundmuller: Agree!

PR #3858, opened 2024-01-16
rework button component and match design system

briangregoryholmes:
- Updates styles to better match design system
- Adds the ability to customize via the class prop
- Removes usage of `buttonClasses` function and co-locates styles within component


djbarnwal (approved):
Looks good!

PR #3859, opened 2024-01-16
support log level for project logs

pjain1:


pjain1:
`npx prettier --check` is failing on a generated code file `web-common/src/runtime-client/gen/query-service/query-service.ts`. I can manually fix it by running npx prettier write but that would defeat the purpose of auto-generated code.

pjain1:
looks like orval generating one extra empty line at the end of the file that is causing issue

proto/rill/runtime/v1/api.proto
```diff
@@ -509,6 +510,7 @@ message WatchLogsRequest {
   string instance_id = 1;
   bool replay = 2;
   int32 replay_limit = 3 [(validate.rules).int32 = {gte: -1}];
+  string level = 4;
```
begelundmuller, line 513: Use `LogLevel` type instead, like `LogLevel level = 4`

proto/rill/runtime/v1/api.proto
```diff
@@ -528,6 +530,7 @@ message WatchResourcesRequest {
   string instance_id = 1;
   string kind = 2;
   bool replay = 3;
+  string level = 4;
```
begelundmuller, line 533: Use `LogLevel` type

runtime/server/instances.go
```diff
@@ -277,3 +294,20 @@ func toString(connectors []*runtimev1.Connector) []string {
...
+	case "WARN":
+		return runtimev1.LogLevel_LOG_LEVEL_WARN
+	case "ERROR":
+		return runtimev1.LogLevel_LOG_LEVEL_ERROR
+	case "FATAL":
+		return runtimev1.LogLevel_LOG_LEVEL_FATAL
+	default:
+		return runtimev1.LogLevel_LOG_LEVEL_UNSPECIFIED
+	}
+}
```
begelundmuller, line 313: After switching to `LogLevel` proto type, this can be moved to the CLI side

runtime/server/instances.go
```diff
@@ -201,11 +202,19 @@ func (s *Server) GetLogs(ctx context.Context, req *runtimev1.GetLogsRequest) (*r
...
 	}
 
-	logs, err := s.runtime.InstanceLogs(ctx, req.InstanceId)
+	if req.Level == "" {
+		req.Level = "INFO"
+	}
+	lvl := toRuntimeLogLevel(req.Level)
+	if lvl == runtimev1.LogLevel_LOG_LEVEL_UNSPECIFIED {
+		return nil, status.Errorf(codes.InvalidArgument, "invalid log level %s", req.Level)
+	}
```
begelundmuller, line 211: After switching to `LogLevel` proto type, can replace this with a check that changes `UNSPECIFIED` to `INFO` (since `UNSPECIFIED` is the default value if the log level isn't explicitly passed, as might be the case from old clients).

runtime/pkg/logbuffer/buffer.go
```diff
@@ -100,17 +103,32 @@ func (b *Buffer) GetLogs(asc bool, limit int) []*runtimev1.Log {
...
+	if limit > 0 {
+		logs = logs[limit:]
+	}
+
+	if !asc {
+		// this is a rare case, only when the user has specified asc=false
+		// reverse the logs
+		for i, j := 0, len(logs)-1; i < j; i, j = i+1, j-1 {
+			logs[i], logs[j] = logs[j], logs[i]
+		}
```
begelundmuller, line 131: Consider using `slices.Reverse(logs)` instead

begelundmuller (changes requested):
Everything else looks good!

runtime/pkg/logbuffer/buffer.go
```diff
@@ -27,35 +26,30 @@ func NewBuffer(maxMessageCount int, maxBufferSize int64) *Buffer {
...
+	// encode fields using zapcore.Encoder, append to entryFields since coreFields will be empty in many cases
+	// until logger is created `With` method.
+	entryFields = append(entryFields, coreFields...)
+	// send empty entry as we want to store message separately
+	fieldsBuf, err := enc.EncodeEntry(zapcore.Entry{}, entryFields)
+	if err != nil {
+		return err
 	}
+	defer fieldsBuf.Free()
+	payload := fieldsBuf.String()
```
begelundmuller: It seems odd to pass the Zap encoder along here. Maybe we should move this Zap-specific logic to `BufferedZapCore`. Then this function could be refactored to something like `AddEntry(lvl runtimev1.LogLevel, t time.Time, msg string, payload string) error`.

runtime/pkg/logutil/handler.go
```diff
@@ -2,17 +2,26 @@ package logutil
...
 	logs   *logbuffer.Buffer
+	enc    zapcore.Encoder
 }
 
 func NewBufferedZapCore(logs *logbuffer.Buffer) *BufferedZapCore {
+	encCfg := zap.NewDevelopmentEncoderConfig()
+	encCfg.NameKey = zapcore.OmitKey
+	encCfg.EncodeLevel = nil
+	encCfg.EncodeTime = nil
+	consoleEncoder := zapcore.NewConsoleEncoder(encCfg)
```
begelundmuller, line 20: Is there a reason not to use `zapcore.NewJSONEncoder()`? Since this is for a JSON payload and only the fields are output

begelundmuller (approved):
LGTM!

PR #3861, opened 2024-01-17
Fixing all dimension values having the same measure value

AdityaHegde:
Selecting compare by dimension in the time dimensional table would result in all dimension values getting the same measure value. This was because the filter was getting sent properly.


web-common/src/features/dashboards/time-series/multiple-dimension-queries.ts
```diff
@@ -243,6 +243,7 @@ export function getDimensionValueTimeSeries(
             ([value, timeseries]) => {
               let prepData = timeseries?.data?.data;
               if (!timeseries?.isFetching) {
+                console.log(value, timeseries?.data);
```
ericpgreen2, line 246: Leftover console log

ericpgreen2 (approved):
Looks good, just a leftover console log.

PR #3863, opened 2024-01-17
Update color theme 3860

bcolloran:
will fix #3860

PR #3866, opened 2024-01-18
Better theme palette generation

AdityaHegde:
This is a prototype with an updated algorithm to get the color palette from the input color.

- Find the color with the lowest [distance](https://www.vis4.net/chromajs/#chroma-distance) with any color in the list of [tailwind colors](https://tailwindcss.com/docs/customizing-colors)
- Calculate the shift in hue and saturation of the matched color compared to input color.
- Get the palette this matched color belongs to and apply the shift to all colors.

Primary color is now available in the dashboard. Secondary color is available as well since we updated the spinner to be primary-500 to secondary-500.


jkhwu:
@AdityaHegde I cloned this branch, ran the app locally, but http://localhost:3000/dev/color-palette didn't go anywhere. I'll Slack you for help.

jkhwu:
Checked the theme color palette and looks good, will check back in after it's applied to dashboards. Thanks @AdityaHegde !

bcolloran:
@AdityaHegde and @jkhwu -- request that we merge this after https://github.com/rilldata/rill/pull/3863

jkhwu:
Just checked these colors applied to the dashboard charts and pills (without the spinner updates). They look good to me.

web-common/package.json
```diff
@@ -36,9 +36,11 @@
     "@tanstack/svelte-table": "^8.9.5",
     "@tanstack/svelte-virtual": "^3.0.1",
     "@testing-library/svelte": "^4.0.0",
+    "@types/chroma-js": "^2.4.3",
     "@vitest/coverage-c8": "^0.31.0",
     "autoprefixer": "^10.4.13",
     "bits-ui": "^0.11.8",
+    "chroma-js": "^2.4.2",
     "clsx": "^2.0.0",
     "colorjs.io": "^0.4.5",
```
bcolloran, line 45: can we use either colorjs.io or chroma-js? I don't think we need both.

web-local/src/routes/dev/color-palette/+page.svelte
```diff
@@ -0,0 +1,47 @@
...
+  let matchedPalette: Color[];
+  $: if (input) {
+    try {
+      const inputColor = chroma(input);
+      const [palette, shift] = getPaletteAndShift(inputColor);
+      matchedPalette = palette;
+      inputPalette = applyShiftToPalette(palette, shift);
+    } catch (err) {
+      // no-op
+    }
```
bcolloran, line 22: Can we remove this try/catch? None of these functions seem like they should be fallible. If one is, then we should try to debug it. If the try/catch is absolutely needed because a library function from chroma.js is failing let's remove that dep and use colorio.js

web-common/src/features/themes/color-config.ts
```diff
@@ -1,3 +1,321 @@
...
+
+// Usable preset converted from the above generated one
+export const TailwindColorPresetsConverted: Color[][] = [];
+for (const colorName in TailwindColorPresets) {
+  const colors: Color[] = [];
+  for (const shade of TailwindColorPresets[colorName]) {
+    colors.push(chroma(shade));
+  }
+  TailwindColorPresetsConverted.push(colors);
+}
```
bcolloran, line 317: no big deal, but this could flatten down to single defintion:
```suggestion
export const TailwindColorPresetsConverted: Color[][] = [];
for (const colorName in TailwindColorPresets) {
  const colors: Color[] = [];
  for (const shade of TailwindColorPresets[colorName]) {
    colors.push(chroma(shade));
  }
  TailwindColorPresetsConverted.push(colors);
}
```

```suggestion
export const TailwindColorPresetsConverted: Color[][] = TailwindColorPresets.map(
  (name) =>  TailwindColorPresets[name].map((shade) =>chroma(shade))
)
```

bcolloran (approved):
Looks good, a couple small change request that you can handle as you see fit @AdityaHegde, but I'm going to mark this as approved, I don't feel the need for a second review (but am happy to review again if you  want to discuss anything)

web-common/package.json
AdityaHegde: chroma-js has a method that was not in colorjs.io. I can remove colorjs.io

PR #3867, opened 2024-01-18
Runtime: custom user agent for motherduck

k-anshul:


k-anshul:
Passing version seemed like some additional work so skipped it for now : https://github.com/rilldata/rill-private-issues/issues/80#issuecomment-1898090880

begelundmuller (approved):

PR #3868, opened 2024-01-18
Support toggle click in leaderboards to exclusively select a dimension value

### LINKED ISSUE #3848: Toggle Click in Leaderboards
_Problem_

When wanting to cycle through 1 filter at a time within a dimension it requires two clicks. First de-filter the existing value and then filter again on the new value. It's one extra click but also causes unnecessary queries being sent to the backend.

Lets assume we have a dimension Region with the values North, South, East & West. In the normal filter behaviour to look at just 1 region value at a time you would have to 

1. Click North - Filter by North
2. Click North - De-selects North - Filters by all regions
3. Click South

_Proposed Solution_

Introduce a Toggle click via cmd + click / control + click (windows). A toggle click would be same behaviour as Clear Any Filters in Dimension + Filter by clicked value. With the same assumed dimension as above,

1. Click North - Filter by North
2. Cmd + Click South - Clear Filter - Filter by South

If multiple pre-existing values exists in a dimension when a cmd + click is done we should clear all previous filters and apply the clicked on.

AdityaHegde:
closes #3848

This adds a simple cmd/ctrl + click handler to leaderboard and dimension table. Doing this on an unselected value will select just that value and deselect everything else. Doing this on a selected value will just deselect it similar to a normal click.

### DIFF
--- a/web-common/src/features/dashboards/state-managers/actions/dimension-filters.ts
+++ b/web-common/src/features/dashboards/state-managers/actions/dimension-filters.ts
@@ -11,6 +11,7 @@ export function toggleDimensionValueSelection(
   dimensionName: string,
   dimensionValue: string,
   keepPillVisible?: boolean,
+  uniqueValue?: boolean,
 ) {
   // if we are able to update the filters, we must cancel any queries
   // that are currently running.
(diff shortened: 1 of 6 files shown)


ericpgreen2:
This interaction feels great! Only thing is we should probably have a line item in the tooltip to callout that this interaction is possible.

web-common/src/features/dashboards/state-managers/actions/dimension-filters.ts
```diff
@@ -11,6 +11,7 @@ export function toggleDimensionValueSelection(
   dimensionName: string,
   dimensionValue: string,
   keepPillVisible?: boolean,
+  uniqueValue?: boolean,
```
ericpgreen2, line 14: This parameter is a little hard to understand on its own. Maybe `isExclusiveFilter` and add a little comment to explain it?

ericpgreen2:
I'm seeing:
- The tooltip in the expanded dimension table doesn't include the informational copy
- The interaction doesn't work in the Time Dimension Detail table (but maybe we can keep this as future work)

ericpgreen2 (approved):

AdityaHegde:
Ya some other tool tips also do not show up in dimension table. Also we decided this feature will not be added to TDD for now.

PR #3872, opened 2024-01-19
Fix export not honouring the old filter

AdityaHegde:
This is a temporary fix for 0.39 to apply the old filter properly.


begelundmuller (approved):
LGTM – notice the linter issue before merging

PR #3879, opened 2024-01-19
MySQL connector

esevastyanov:


k-anshul (commented):
Are we not using `duckdb-mysql` extension for this ?
We decided to build our own connector for postgres since the postgres_scanner extension had licensing issues. 
`duckdb-mysql` extension is MIT licensed so I believe we can use the extension directly similar to `sqlite`.
I also noticed that `postgres_scanner` extension is also MIT licensed now so I believe we can deprecate our current connector and use duckdb extension directly(in a separate PR).
https://github.com/duckdb/postgres_scanner/pull/117

esevastyanov:
@k-anshul you are right - both extensions can be used now. However, we decided to go with our own connectors and then migrate (if possible) to `duckdb-mysql` and `postgres_scanner`.

web-common/src/features/sources/modal/yupSchemas.ts
```diff
@@ -135,6 +135,19 @@ export function getYupSchema(connector: V1ConnectorSpec) {
...
+      return yup.object().shape({
+        sql: yup.string().required("sql is required"),
+        sourceName: yup
+          .string()
+          .matches(
+            /^[a-zA-Z_][a-zA-Z0-9_]*$/,
+            "Source name must start with a letter or underscore and contain only letters, numbers, and underscores",
+          )
+          .required("Source name is required"),
+        output_location: yup.string(),
```
k-anshul, line 148: `output_location` and `workgroup` not required for MySQL.

runtime/drivers/mysql/parser.go
```diff
@@ -0,0 +1,206 @@
+package mysql
+
+import (
+	"database/sql"
+	"fmt"
+	reflect "reflect"
```
k-anshul, line 6: Can be just reflect.

runtime/drivers/mysql/parser.go
```diff
@@ -0,0 +1,206 @@
...
+		return new(sql.NullFloat64), nil
+	default:
+		return nil, fmt.Errorf("numericMapper: unsupported scan type %v", st.Name())
+	}
+}
+
+type bitMapper struct{}
+
+func (m *bitMapper) runtimeType(reflect.Type) (*runtimev1.Type, error) {
+	return &runtimev1.Type{Code: runtimev1.Type_CODE_BYTES}, nil
```
k-anshul, line 151: Shouldn't bit be string?

runtime/drivers/mysql/sql_store.go
```diff
@@ -0,0 +1,212 @@
...
+		return nil, err
+	}
+
+	for i := range r.schema.Fields {
+		v := dests[i]
+		if v == nil {
+			r.row[i] = nil
+			continue
+		}
+		if reflect.ValueOf(v).Kind() == reflect.Ptr {
```
k-anshul, line 126: Could you please explain this if block? Some questions:
1. Wouldn't this always be a ptr since dest always returns pointer to some type ?
2. Under what conditions can it be a slice ?

k-anshul (changes requested):
Can you also add a unit test that adds data from mysql to duckdb with hopefully as many types as possible. 
Please check `transporter_postgres_to_duckDB_test.go` for reference (It became useful for me just now when I was migrating `go-duckdb` to later version and realised appender API is broken bcz of which some datatypes no longer works).

runtime/drivers/mysql/parser.go
```diff
@@ -0,0 +1,224 @@
...
+	// date and time
+	m[("DATE")] = &dateMapper{}
+	m[("DATETIME")] = &dateMapper{}
+	m[("TIMESTAMP")] = &dateMapper{}
+	m[("YEAR")] = &int16Mapper{}
+	// TIME is scanned as bytes and can be converted to string
+	m[("TIME")] = &charMapper{}
+
+	// json
+	m[("JSON")] = &jsonMapper{}
```
begelundmuller, line 65: Why parenthesis around the map key strings?

runtime/drivers/mysql/sql_store.go
```diff
@@ -0,0 +1,216 @@
...
+
+	conf, err := mysql.ParseDSN(dsn)
+	if err != nil {
+		return nil, err
+	}
+	conf.ParseTime = true // if set to false, time is scanned as an array rather than as time.Time
+
+	db, err := sql.Open("mysql", conf.FormatDSN())
+	if err != nil {
+		db.Close()
```
begelundmuller, line 41: I don't think `db.Close` should be called if `Open` fails (`db` will be `nil` in that case and may give nil access panic)

runtime/drivers/mysql/sql_store.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+
+	// Scan expects destinations to be pointers
+	dests := make([]any, len(cts))
+	for i := range dests {
+		dests[i], err = r.fieldMappers[i].dest(cts[i].ScanType())
+		if err != nil {
+			return nil, err
+		}
+	}
```
begelundmuller, line 115: Allocating a new row map for each row might cause a lot of allocations. IIRC, the Postgres implementation explicitly tries to avoid that by caching the row in the iterator and re-using it.

I must also admit I get a little lost in the reflection code in this implementation and don't quite understand why it's needed. Why can't we map the `ColumnTypes` to a slice of values with the appropriate types (just using a regular switch and return of empty value of the type we want), and then `Scan` into that and return it directly? (Which would not require reflection.)

begelundmuller (changes requested):

runtime/drivers/mysql/parser.go
```diff
@@ -0,0 +1,206 @@
...
+	runtimeType(st reflect.Type) (*runtimev1.Type, error)
+	dest(st reflect.Type) (any, error)
+}
+
+// refer https://github.com/go-sql-driver/mysql/blob/master/fields.go for base types
+func getDBTypeNameToMapperMap() map[string]mapper {
+	m := make(map[string]mapper)
+
+	// bit
+	m[("BIT")] = &bitMapper{}
```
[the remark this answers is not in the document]
esevastyanov: Removed parenthesis

runtime/drivers/mysql/parser.go
esevastyanov: Converted to a bit string

runtime/drivers/mysql/sql_store.go
esevastyanov: 1. Yes
2. Binary columns

However, I completely modified this block

runtime/drivers/mysql/parser.go
esevastyanov: Removed the parenthesis

runtime/drivers/mysql/sql_store.go
esevastyanov: Row slice is re-used on every `Next` call but I also added a re-use of column destinations
Removed reflection and moved the dereferencing to `switch` statements

runtime/drivers/mysql/sql_store.go
```diff
@@ -0,0 +1,189 @@
...
+		if errors.Is(err, sql.ErrNoRows) {
+			return nil, fmt.Errorf("no results found for the query")
+		}
+		return nil, err
+	}
+
+	cts, err := r.rows.ColumnTypes()
+	if err != nil {
+		return nil, err
+	}
```
begelundmuller, line 105: This might be an expensive operation and should be avoided on each row scan

runtime/drivers/mysql/sql_store.go
```diff
@@ -0,0 +1,189 @@
...
+
+	for i := range r.schema.Fields {
+		// Dereference destinations and fill the row
+		r.row[i], err = r.fieldMappers[i].value(r.fieldDests[i])
+		if err != nil {
+			return nil, err
+		}
+	}
```
begelundmuller, line 125: I think there might still be room to simplify this code. You can pass an `any` value to `Scan`. What if we just did something like:
```go
// in setSchema
r.dests = make([]any, len(r.schema.Fields))
for i := range r.dests {
  r.dests[i] = new(any)
}

// in Scan
err = r.rows.Scan(r.dests...)
if err != nil {...}

for i := r.schema.Fields {
  r.row[i], err = r.mappers[i].value(dests[i])
}
```

runtime/drivers/mysql/parser.go
```diff
@@ -0,0 +1,307 @@
...
+	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
+)
+
+type mapper interface {
+	runtimeType(st reflect.Type) (*runtimev1.Type, error)
+	// dest returns a pointer to a destination value that can be used in Rows.Scan
+	dest(st reflect.Type) (any, error)
+	// value dereferences a pointer created by dest
+	value(p any) (any, error)
+}
```
begelundmuller, line 18: I believe we should still be able to avoid reflection in this code. Based on another common, `dest` might not be needed, and `runtimeType` doesn't need the `st` parameter – the only case where it's currently accessed is in `numericMapper`, and for that case, you can just refactor to this:

```go
type numericMapper struct {
  code runtime.Type_CODE
}

// in getDBTypeNameToMapperMap
m["TINYINT"] = numericMapper{code: runtimev1.Type_CODE_INT8}
...
```

runtime/drivers/mysql/parser.go
```diff
@@ -0,0 +1,307 @@
...
+	// date and time
+	m["DATE"] = &dateMapper{}
+	m["DATETIME"] = &dateMapper{}
+	m["TIMESTAMP"] = &dateMapper{}
+	m["YEAR"] = &numericMapper{}
+	// TIME is scanned as bytes and can be converted to string
+	m["TIME"] = &charMapper{}
+
+	// json
+	m["JSON"] = &jsonMapper{}
```
begelundmuller, line 68: Since these structs dont have state, I think they don't need to be pointers? For example, can just set `= bitMapper{}` instead of `= &bitMapper{}` (would need to change the member function signatures to work on non-pointer types)

begelundmuller (changes requested):

runtime/drivers/mysql/sql_store.go
esevastyanov: This requires implementing some conversions that `Scan` has. For example, numeric types are represented as `[]uint8`. But to use `Scan`'s conversions one needs to appropriately prepare destinations - `dest` function

runtime/drivers/mysql/parser.go
esevastyanov: Reflection, in this case, doesn't affect performance as it is used only to store scan types (see [mysql driver](https://github.com/go-sql-driver/mysql/blob/master/fields.go#L120)).
`TINYINT` might be scanned as `uint8`, `int8` and `sql.NullInt64` depending on `NOT NULL` and `UNSIGNED` column props

runtime/drivers/mysql/sql_store.go
begelundmuller: Okay, that's a shame, but makes sense

k-anshul (approved):

PR #3880, opened 2024-01-19
Process Kafka producer events (delivery, errors, etc)

esevastyanov:


runtime/pkg/activity/kafka.go
```diff
@@ -53,12 +57,72 @@ func NewKafkaSink(brokers, topic string, logger *zap.Logger) (*KafkaSink, error)
...
+			if s.activity != nil {
+				// Be optimistic - if delivery failed because of a temporary issue then
+				// the following activity event might be eventually delivered
+				s.activity.Emit(context.Background(), "activity_kafka_delivery_failure", float64(deliveryFailureCount),
+					attribute.String("error", lastDeliveryError.Error()))
+			}
+			deliveryFailureCount = 0
+			lastDeliveryError = nil
+		}
+	}
```
begelundmuller, line 103: We should use OpenTelemetry for this. OpenTelemetry is still our primary choice for internal observability. I think we should aim to keep the activity client to the original use case of gathering events for usage, billing and uptime reporting.

That will also avoid the cyclic dependency between the sink and activity client, which is a nice benefit.

Eventually, we should probably find a way to emit OpenTelemetry data to the Rill metrics project as well.

runtime/pkg/activity/kafka.go
```diff
@@ -53,12 +57,72 @@ func NewKafkaSink(brokers, topic string, logger *zap.Logger) (*KafkaSink, error)
...
+			switch ev := e.(type) {
+			case *kafka.Message:
+				if ev.TopicPartition.Error != nil {
+					deliveryFailureCount++
+					lastDeliveryError = ev.TopicPartition.Error
+				} else {
+					deliverySuccessCount++
+				}
+			case kafka.Error:
+				logger.Warn("Kafka sink: producer error", zap.String("error", ev.Error()))
```
begelundmuller, line 117: We generally won't notice warnings. Should we make this error level?

begelundmuller (changes requested):

runtime/pkg/activity/kafka.go
esevastyanov: Lets use error level for now but this error might be a duplicate of what is logged by `forwardKafkaLogEventToLogger`

runtime/pkg/activity/kafka.go
```diff
@@ -53,12 +63,42 @@ func NewKafkaSink(brokers, topic string, logger *zap.Logger) (*KafkaSink, error)
...
+
+	go sink.processProducerEvents(producer, logger)
+
+	return &sink, nil
+}
+
+func (s *KafkaSink) processProducerEvents(producer *kafka.Producer, logger *zap.Logger) {
+	for {
+		select {
+		case e := <-producer.Events():
```
begelundmuller, line 82: I wonder if the `producer.Events` channel might be closed when `s.producer.Close()` is called. Just to be safe, might be worth doing:
```
case e, ok := <-producer.Events():
  if !ok {
    return
  }
  ...
```

runtime/pkg/activity/kafka.go
```diff
@@ -143,3 +184,10 @@ func retry(maxRetries int, delay time.Duration, fn func() error, retryOnErrFn fu
 	}
 	return err
 }
+
+func Must[T any](v T, err error) T {
```
begelundmuller, line 188: Can be made internal (i.e. lowercase `must`)

begelundmuller (approved):

PR #3882, opened 2024-01-22
Cleaning up old filters references from UI

### LINKED ISSUE #3849: Export from dimension table does not retain filters in export data
**Describe the bug**
When exporting, the filter settings should retain values so that the export is filtered to the same

**To Reproduce**
1. Filter by any dimension value(s)
2. Expand another dimension table
3. Select export csv

**Expected behavior**
Expected behavior is export would be filtered to selected. Instead, the values are exported fully

AdityaHegde:
closes #3849

Some of the places were still using the older filters from the dashboard store.

This PR removes the old filter object from the dashboard store entity and removes all references to it. The only remaining reference is in `fromProto` for backwards compatibility from older url state.


ericpgreen2 (commented):
Nice to see this cleanup. So can we now remove the `filter` parameter from each API definition and the generated client code? (An observation: it seems Orval does not convey the OpenAPI `title` attribute in the generated code, so it's not clear to anyone using the client that the `filter` parameter is deprecated.)

ericpgreen2:
UX issue: with the Adbids dataset, I filtered the Publisher dimension for "oo", clicked "Select all", and the app crashed

image

AdityaHegde:
We need the old params for now since reports might have it. Lets do a follow up for that and figure out a way to migrate.

ericpgreen2 (approved):

PR #3885, opened 2024-01-22
add toggle button to pivot toolbar

briangregoryholmes:
This PR adds a toggle button to the Pivot Toolbar component to show/hide the drag-and-drop panels.


djbarnwal (approved):
Looks good. One nit - Can we remove the focus on the toggle button after it's clicked otherwise it remains in the focus state with a border around. 
image

PR #3892, opened 2024-01-23
Fetch the latest tag from github

himadrisingh:


.github/workflows/cli-release.yml
```diff
@@ -65,6 +65,8 @@ jobs:
 
           if [[ ${{ env.PUBLISH_NIGHTLY }} == 'true' ]]; then
             export NIGHTLY="--snapshot"
+            LATEST_TAG=$(curl -s https://api.github.com/repos/rilldata/rill/releases/latest | jq -r '.name')
+            echo "GORELEASER_CURRENT_TAG=${LATEST_TAG}" >> .release-env
```
begelundmuller, line 69: Any way to do this using `git` commands or built-in features in either Goreleaser or GH actions? Wondering if this might be fragile given it's an external network request and relies on a specific response format.

.goreleaser.yaml
```diff
@@ -110,7 +110,7 @@ checksum:
   name_template: checksums.txt
 
 snapshot:
-  name_template: "{{ incpatch .Version }}-snapshot"
+  name_template: "{{ incminor .Version }}-snapshot"
```
begelundmuller, line 113: What about changing this to `-nightly` instead of `-snapshot`? Checked two projects, Goreleaser and Rust, they both use `-nightly`. It might seem more natural to users

himadrisingh, line 113:
```suggestion
  name_template: "{{ incminor .Version }}-nightly"
```

.github/workflows/cli-release.yml
```diff
@@ -65,6 +65,8 @@ jobs:
 
           if [[ ${{ env.PUBLISH_NIGHTLY }} == 'true' ]]; then
             export NIGHTLY="--snapshot"
+            LATEST_TAG=$(curl -s https://api.github.com/repos/rilldata/rill/releases/latest | jq -r '.name')
```
himadrisingh, line 68:
```suggestion
            LATEST_TAG=$(git describe --tags $(git rev-list --tags --max-count=1))
```

begelundmuller (approved):

PR #3894, opened 2024-01-23
Refactor reports to just have the queryArgs. Build the state url in the UI.

AdityaHegde:
Currently the dashboard url for a report is filled in by the UI. This is hard to manage when creating a report manually using file artifacts. Instead we can generate the dashboard url using the query args in the UI and redirect to that.

This PR adds,
- [x] Add a path `/<org>/<proj>/reports/<report>/open`. Map the query object to the url state and redirect to the dashboard.
- [x] Update backend to use this new path instead of using the url passed in annotations.
- [x] Loading spinner in the `/open` route.
- [x] Error handling in the `/open` route.
- [x] Cleanup of old code.


nishantmonu51:
is this backwards compatible ?

AdityaHegde:
It would take the query args from the older reports, ignoring the state url.

proto/rill/admin/v1/api.proto
```diff
@@ -1066,6 +1066,7 @@ message GetReportMetaRequest {
   string branch = 2;
   string report = 3;
   map<string, string> annotations = 4;
+  string execution_time = 5;
```
begelundmuller, line 1069: Can we make this a `google.protobuf.Timestamp` and then move the string formatting into the url builder in `externalURLs.reportOpen`?

begelundmuller (changes requested):

begelundmuller (commented):
Backend changes look good to me now!

ericpgreen2:
I tested by creating a new report, clicking on the "Open in browser" button in the email, and I hit this error:
image

web-admin/src/features/scheduled-reports/query-mapper.ts
```diff
@@ -0,0 +1,277 @@
...
+  timeRangeSummary: V1TimeRangeSummary;
+  executionTime: string;
+};
+
+type ParsedReportQuery = {
+  ready: boolean;
+  state?: string;
+  metricsView?: string;
+  error?: string;
+};
```
ericpgreen2, line 48: I'm noticing we have a lot of places in the codebase where we use a derived store to combine the results from two queries. Rather than a type like this one that mixes query status (`ready`, `error`) with query results (`state`, `metricsView`), it could be good if we created a generic type to hold the query status, like...

```typescript
type BaseQuery<T> = {
  isReady: boolean;
  error?: string
  data: T
}
```

which we could use to wrap the query results type...

```typescript
type DashboardState = {
    metricsView: string
    state: string
}

type DashboardStateQuery = BaseQuery<DashboardState>;
```

ericpgreen2: FWIW TanStack Query v5 will give us a first-class way to combine multiple queries, and the above approach mirrors TanStack.

web-admin/src/features/scheduled-reports/query-mapper.ts
```diff
@@ -0,0 +1,277 @@
...
+
+  return newReq;
+}
+
+function mapMetricsViewAggregationRequest({
+  req,
+  dashboard,
+}: QueryMapperArgs<V1MetricsViewAggregationRequest>) {
+  if (req.where) dashboard.whereFilter = req.where;
+  // TODO
```
ericpgreen2, line 176: Mind clarifying that these TODOs should be completed when we offer scheduled reports for the given query?

web-admin/src/features/scheduled-reports/query-mapper.ts
```diff
@@ -0,0 +1,277 @@
...
+        timeRangeSummary.data,
+      );
+      queryMapper({
+        dashboard,
+        req,
+        metricsView: metricsViewResource.data,
+        timeRangeSummary: timeRangeSummary.data.timeRangeSummary,
+        executionTime,
+      });
+      console.log(dashboard);
```
ericpgreen2, line 141: Leftover console log

web-admin/src/features/scheduled-reports/query-mapper.ts
```diff
@@ -0,0 +1,277 @@
...
+      };
+    },
+  );
+}
+
+/**
+ * This method corrects the underscore naming to camel case.
+ * This is the drawback of storing the request object as is.
+ */
+function correctRequest(req: Record<string, any>): Record<string, any> {
```
ericpgreen2, line 155: The comment is helpful, but maybe even the function name could be `convertSnakeToCamelCase()`

web-admin/src/features/scheduled-reports/query-mapper.ts
```diff
@@ -0,0 +1,277 @@
...
+        metricsViewName,
+        metricsViewResource.data,
+        timeRangeSummary.data,
+      );
+      queryMapper({
+        dashboard,
+        req,
+        metricsView: metricsViewResource.data,
+        timeRangeSummary: timeRangeSummary.data.timeRangeSummary,
+        executionTime,
```
ericpgreen2, line 139: Is it purposeful that the function returns `void`? Might be more clear if it returned the dashboard state object

web-admin/src/features/scheduled-reports/query-mapper.ts
ericpgreen2, line 1: Name suggestion: `query-mapper.ts` -> `get-dashboard-state-for-report.ts`

web-admin/src/features/scheduled-reports/query-mapper.ts
```diff
@@ -0,0 +1,277 @@
...
+  state?: string;
+  metricsView?: string;
+  error?: string;
+};
+
+/**
+ * Reports manually written through file artifacts won't have the UI to feed the url state.
+ * Hence we are building the state from the query args in the report.
+ */
+export function parseReportQuery(
```
ericpgreen2, line 54: Name suggestion: `getDashboardStateForReport(...)`

web-admin/src/features/scheduled-reports/query-mapper.ts
```diff
@@ -0,0 +1,277 @@
...
+        };
+      }
+
+      initLocalUserPreferenceStore(metricsViewName);
+      const dashboard = getDefaultMetricsExplorerEntity(
+        metricsViewName,
+        metricsViewResource.data,
+        timeRangeSummary.data,
+      );
+      queryMapper({
```
ericpgreen2, line 134: Name suggestion: `getDashboardState`

ericpgreen2 (commented):
Overall, code is clear! Mostly left some name and type recommendations.

web-admin/src/features/scheduled-reports/query-mapper.ts
AdityaHegde: UI only adds MetricsViewComparison right now. Need to clarify if we need to add support for others in the 1st cut. Lemme double check with product before merging this PR.

web-admin/src/features/scheduled-reports/query-mapper.ts
AdityaHegde: It updated the dashboard in place. It felt like an overkill to make it copy again and return it, since we are creating the dashboard object before passing it to the mapper.

web-admin/src/features/scheduled-reports/query-mapper.ts
AdityaHegde: Good point. I will not refactor other places but update this derived state.

ericpgreen2:
I just created a report and tried to open it, but now I hit a different error: 
image

The intended report view: 
image

web-common/src/features/compound-query-result.ts
ericpgreen2, line 1: I think it'd be better if this lived in `web-common/src/runtime-client` or `web-common/src/lib`

web-admin/src/features/scheduled-reports/query-mapper.ts
ericpgreen2: Probably not needed for the first cut because we don't yet mention Report YAML in our docs anywhere.

My point was that the TODOs confused me until I realized that we don't yet offer scheduled reports for these queries. A note like 'TODO: Scheduled reports not yet available for this query' could help clarify.

ericpgreen2 (approved):
Works for me now! Approving so you can merge after fixing CI tests.

begelundmuller (approved):

PR #3898, opened 2024-01-24
optimize toplist to have only needed measures

pjain1:


pjain1:
E2E is failing because on changing the time period `No comparison` is getting changed to Comparison, test is expecting No comparison. This is happening on main as well.

begelundmuller (changes requested):

runtime/queries/metricsview_comparison_toplist.go
```diff
@@ -36,9 +36,18 @@ type MetricsViewComparison struct {
 	// backwards compatibility
 	Filter *runtimev1.MetricsViewFilter `json:"filter"`
 
+	// internal use
+	MeasuresMeta map[string]*MeasureMeta `json:"-"`
```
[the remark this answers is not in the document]
pjain1: This is needed in export as well so made it top level as general export just calls `Resolve` and not individual method and assumes all measures has derived comparisons.

runtime/queries/metricsview_comparison_toplist.go
```diff
@@ -36,9 +36,18 @@ type MetricsViewComparison struct {
...
+
 	Result *runtimev1.MetricsViewComparisonResponse `json:"-"`
 }
 
+type MeasureMeta struct {
```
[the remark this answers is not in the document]
pjain1: we can name it `mvMeasureMeta`, its used in general export so will work for both toplist and comparetoplist meta

AndrewRTsao:
Since we had our code freeze last Friday (but since this will be something we'll want to have in place for Automatad), once we have a final fix in place, what would be the best way to make sure this is part of our 0.40 release? 

@ericpgreen2 @begelundmuller Any chance we could merge this with our release candidate or would this require cutting a new hotfix (not familiar with our internal engineering processes)?

bcolloran:
Hey, I just happened to look at this and see that it disables some client side code that I wrote last month. Not saying not to merge it if it's causing bigger problems elsewhere, but disabling that code _will_ cause a regression in a bug that was fixed recently, so from a process perspective I think it would have been appropriate to look at the git blame for the code being disabled and flag the author to discuss the impact and alternatives. I only happened to look at this by PR chance, and would have missed the regression being introduced.

ericpgreen2:
> Since we had our code freeze last Friday (but since this will be something we'll want to have in place for Automatad), once we have a final fix in place, what would be the best way to make sure this is part of our 0.40 release?
> 
> @ericpgreen2 @begelundmuller Any chance we could merge this with our release candidate or would this require cutting a new hotfix (not familiar with our internal engineering processes)?

Andrew, given this PR is considered a release blocker, we'll add it (cherry-pick it) to the release branch once it's merged to `main`.

pjain1:
@bcolloran that code removal was just a hack to get past the query cancellations that I was seeing. Just reached out to Aditya during IST hours as I wanted to deploy this code on cloud for testing and thats why pushed the change to the branch. Its not supposed to be merged, I will revert that commit. @AdityaHegde mentioned that applications teams will discuss internally on how to implement a proper fix.

pjain1:
@bcolloran removed that commit

bcolloran:
> @bcolloran that code removal was just a hack to get past the query cancellations that I was seeing. Just reached out to Aditya during IST hours as I wanted to deploy this code on cloud for testing and thats why pushed the change to the branch. Its not supposed to be merged, I will revert that commit. @AdityaHegde mentioned that applications teams will discuss internally on how to implement a proper fix.

Gotcha, thanks @pjain1! Sorry to be sticking my nose in where it doesn't belong in that case!!

ericpgreen2:
@pjain1, the frontend code LGTM.

runtime/queries/metricsview_comparison_toplist.go
```diff
@@ -36,9 +37,18 @@ type MetricsViewComparison struct {
 	// backwards compatibility
 	Filter *runtimev1.MetricsViewFilter `json:"filter"`
 
+	// internal use
+	MeasuresMeta map[string]metricsViewMeasureMeta `json:"-"` // ignore this field when marshaling
```
begelundmuller, line 41: Should be lowercase `measuresMeta` if not accessed externally

egor-ryashin (approved):

runtime/queries/metricsview_comparison_toplist.go
```diff
@@ -1045,51 +1126,55 @@ func timeRangeClause(tr *runtimev1.TimeRange, mv *runtimev1.MetricsViewSpec, dia
...
-		case runtimev1.MetricsViewComparisonMeasureType_METRICS_VIEW_COMPARISON_MEASURE_TYPE_COMPARISON_VALUE,
-			runtimev1.MetricsViewComparisonMeasureType_METRICS_VIEW_COMPARISON_MEASURE_TYPE_ABS_DELTA,
-			runtimev1.MetricsViewComparisonMeasureType_METRICS_VIEW_COMPARISON_MEASURE_TYPE_REL_DELTA:
-			if !hasComparison {
-				return fmt.Errorf("comparison not enabled for alias %s", alias.Alias)
-			}
+func validateAlias(aliases []*runtimev1.MetricsViewComparisonMeasureAlias, comparisonMap map[string]bool) error {
+	for _, a := range aliases {
+		if a.Type != runtimev1.MetricsViewComparisonMeasureType_METRICS_VIEW_COMPARISON_MEASURE_TYPE_BASE_VALUE && !comparisonMap[a.Name] {
+			return fmt.Errorf("alias measure '%s' is not present in the comparison measures", a.Name)
```
begelundmuller, line 1172: The old one checked explicitly for the comparison cases. I think that might be better because a) I think we also allow passing `METRICS_VIEW_COMPARISON_MEASURE_TYPE_UNIDENTIFIED`, b) new measure types that are not comparison-related might be added in the future.

runtime/queries/metricsview_comparison_toplist.go
```diff
@@ -1045,51 +1126,55 @@ func timeRangeClause(tr *runtimev1.TimeRange, mv *runtimev1.MetricsViewSpec, dia
...
+				s.SortType = runtimev1.MetricsViewComparisonMeasureType_METRICS_VIEW_COMPARISON_MEASURE_TYPE_ABS_DELTA
 			case runtimev1.MetricsViewComparisonSortType_METRICS_VIEW_COMPARISON_SORT_TYPE_REL_DELTA:
-				comparisonSort.SortType = runtimev1.MetricsViewComparisonMeasureType_METRICS_VIEW_COMPARISON_MEASURE_TYPE_REL_DELTA
+				s.SortType = runtimev1.MetricsViewComparisonMeasureType_METRICS_VIEW_COMPARISON_MEASURE_TYPE_REL_DELTA
+			}
+		}
+
+		if s.SortType == runtimev1.MetricsViewComparisonMeasureType_METRICS_VIEW_COMPARISON_MEASURE_TYPE_UNSPECIFIED {
+			return fmt.Errorf("sort measure '%s' does not have a sort type", s.Name)
+		}
```
begelundmuller, line 1155: I think previously you could pass an unspecified sort type?

runtime/queries/metricsview_comparison_toplist.go
```diff
@@ -99,13 +104,75 @@ func (q *MetricsViewComparison) Resolve(ctx context.Context, rt *runtime.Runtime
...
+		if expand {
+			outer += 4
+		} else {
+			outer++
+		}
+		inner++
+	}
+	return nil
+}
```
begelundmuller, line 174: Always hard to say about premature optimization, but since this gets hit dozens of times per dashboard load, I think there are some low hanging optimizations that can be easily fixed here:

1. The number of measures is usually quite small, so instead of allocating/populating/garbage collecting maps for `measureMap` and `comparisonMap`, consider just iterating over the input slices when building `q.MeasureMeta`
2. Can populate and pass `q.MeasuresMeta` to `validateSort` and `validateAlias` instead of the separate maps
3. For the case where `len(q.ComparisonMeasures) == 0 && compare`, there is no need to validate the comparison measures

Another question – what about case sensitivity? I'm not sure, but I wonder if the old implementation was case insensitive?

begelundmuller (changes requested):

runtime/queries/metricsview_comparison_toplist.go
pjain1: I commonized common checks for both toplist and compare toplist cases so using the comparisonMap, toplist case will have comparisonMap empty. 
a. Yes it was being passed earlier but the new code disallows that as was not sure whats the use of that and that makes the semantics loose. For example, currently there is a edge case where front end will send a relative sort type for non comparison toplist query, earlier there was no strict check in the toplist logic so it was working, just wanted to avoid similar cases. 
b. May be we use `METRICS_VIEW_COMPARISON_MEASURE_TYPE_BASE_VALUE` for them, if not I can revert it back to what it was previously.

runtime/queries/metricsview_comparison_toplist.go
pjain1: yes I am disallowing that as unlike alias, its a sort measure so makes sense to have a valid sort type, also catches some edge cases as described above. Unless we plan to use it for non measures it makes sense.

pjain1: reverted the change

runtime/queries/metricsview_comparison_toplist.go
pjain1: reverted the change

runtime/queries/metricsview_comparison_toplist.go
pjain1: fixed

runtime/queries/metricsview_comparison_toplist.go
```diff
@@ -117,48 +117,56 @@ func (q *MetricsViewComparison) Resolve(ctx context.Context, rt *runtime.Runtime
...
 	}
 
-	err := validateSort(q.Sort, comparisonMap, compare)
+	err := validateSort(q.Sort, q.ComparisonMeasures, compare)
 	if err != nil {
 		return err
 	}
 
-	err = validateAlias(q.Aliases, comparisonMap)
+	err = validateMeasureAliases(q.Aliases, q.ComparisonMeasures, compare)
```
begelundmuller, line 152: Sorry, maybe it wasn't clear before, but why not use `q.measuresMeta` to do these validations? I.e. build it first, validate after. That way, it's just one map allocation and only one nested loop (for building `q.measuresMeta`).

runtime/queries/metricsview_comparison_toplist.go
```diff
@@ -99,13 +104,84 @@ func (q *MetricsViewComparison) Resolve(ctx context.Context, rt *runtime.Runtime
...
+				if m.Name == cm {
+					found = true
+					break
+				}
+			}
+			if !found {
+				return fmt.Errorf("comparison measure '%s' is not present in the measures list", cm)
+			}
+		}
+	}
```
begelundmuller, line 145: This validation can also be pushed to after the `q.measuresMeta` is built and avoid the nested loop

begelundmuller (approved):

PR #3902, opened 2024-01-25
add icons to explore/pivot tabs and resolve navigation bug

briangregoryholmes:
Adds tab icons and resolves bug related to tab state when navigating between dashboards.


djbarnwal (approved):
Looks good!

PR #3911, opened 2024-01-25
update feature flag implementation

briangregoryholmes:
This PR updates the feature flag implementation in the following ways:

- Fixes typing issues previously present in the file.
- Supports consumption of feature flags via URL Search Params as a comma separated list.
- Creates a custom feature flag store that:
  -  Provides three helper methods (set, update and toggle) to dynamically change an arbitrary selection of flags without impacting existing flags.
  - Automatically derives getters for each flag so that they can be individually subscribed to in components that need them (rather than subscribing to the entire store)
- Updates the usage of the features flag store throughout the application


ericpgreen2 (approved):
Great!

PR #3919, opened 2024-01-29
Fix scroll on embedded dashboards

ericpgreen2:
The scroll logic in `web-admin/src/routes/+layout.svelte` checks to see if the user is on a dashboard page. Previously, we only checked dashboards at the route `/{org}/{project}/{dashboard}`. With this PR, we additionally check embedded dashboards served from `/-/embed`.


bcolloran (approved):
I don't know how to QA embedded pages, but this seems totally reasonable so happy to approve (we can go over the QA steps after this week's rush)

PR #3923, opened 2024-01-29
#3922 Remove -headless from the image

himadrisingh:


kaspersjo (changes requested):
I suggest we keep `rill-headless` as the image name, but just remove the `-headless` part from `main.Version` ldflag in this line, since that flag is used to resolve the version at runtime. https://github.com/rilldata/rill/blob/a82cbe970a558e85315b79e35615e4b4ad817685/.github/workflows/rill-cloud.yml#L47

kaspersjo (approved):

PR #3924, opened 2024-01-29
Integrate new filter object into pivot prototype

AdityaHegde:
Please go the the `Preview` tab and select the appropriate sub-template:

* [Others](?expand=1&template=fallback_template.md)

### DIFF
--- a/web-common/src/features/dashboards/pivot/pivot-expansion.ts
+++ b/web-common/src/features/dashboards/pivot/pivot-expansion.ts
@@ -1,7 +1,11 @@
 import type { StateManagers } from "@rilldata/web-common/features/dashboards/state-managers/state-managers";
+import {
+  createAndExpression,
+  createInExpression,
+} from "@rilldata/web-common/features/dashboards/stores/filter-utils";
 import type { TimeRangeString } from "@rilldata/web-common/lib/time/types";
 import type {
-  MetricsViewFilterCond,
+  V1Expression,
   V1MetricsViewAggregationResponseDataItem,
 } from "@rilldata/web-common/runtime-client";
 import { Readable, derived, writable } from "svelte/store";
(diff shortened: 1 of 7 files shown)


web-common/src/features/dashboards/pivot/pivot-infinite-scroll.ts
```diff
@@ -115,14 +116,7 @@ export function getColumnFiltersForPage(
...
-  Object.entries(pageGroups).forEach(([colDimensionName, values]) => {
-    colFilters.push({
-      name: colDimensionNames[parseInt(colDimensionName)],
-      in: Array.from(values),
-    });
-  });
-
-  return colFilters;
+  return Object.entries(pageGroups).map(([colDimensionName, values]) =>
+    createInExpression(colDimensionName, Array.from(values)),
```
djbarnwal, line 120: The variable name is mistyped leading to an error while adding row dimensions. Can be changed to - 

 ```
  return Object.entries(pageGroups).map(([colDimensionId, values]) =>
    createInExpression(
      colDimensionNames[parseInt(colDimensionId)],
      Array.from(values),
    ),
  );
```

File: web-common/src/features/dashboards/pivot/pivot-expansion.ts [hunk 2]
```diff
@@ -190,8 +183,10 @@ export function queryExpandedRowMeasureValues(
         timeFilters,
       );
 
-      // TODO: Merge filters
-      // mergeFilters(rowNestFilters, sortFilters)
+      const mergedFilter = mergeFilters(
+        createAndExpression(rowNestFilters),
+        config.whereFilter,
+      );
```
djbarnwal, line 189: Don't have to add this here now. We can merge with global filter later at the query level. 

Idea to add mergeFilters here was to add - 
1. Filters for a sub row
2. Filters for sorted column

I can take care of this outside of this PR.

djbarnwal (commented):
Thanks for add the new filter expressions. Works well for most parts but seeing some data inconsistencies. My idea was to merge the global filters at the very end inside `createPivotAggregationRowQuery` rather than at util level. 

image

djbarnwal (approved):
Looks good! Thanks for the changes

PR #3928, opened 2024-01-29
collapsible pivot drag area and persistent chip styling

briangregoryholmes:
- Adds a new PivotDrag component to enable design system specifics around drag areas including collapsibility and styling
- Adds new types to capture the structure of chip data
- Adds helper functions for collecting information about column and row values in the needed format
- Makes chips removable in row/column drag sections

### DIFF
--- a/web-common/src/features/dashboards/pivot/pivot-utils.ts
+++ b/web-common/src/features/dashboards/pivot/pivot-utils.ts
@@ -7,6 +7,7 @@ import {
 } from "@rilldata/web-common/lib/time/types";
 import type {
   MetricsViewFilterCond,
+  MetricsViewSpecDimensionV2,
   MetricsViewSpecMeasureV2,
   V1MetricsViewAggregationSort,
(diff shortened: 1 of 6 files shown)


web-common/src/features/dashboards/pivot/pivot-utils.ts
```diff
@@ -51,6 +53,85 @@ export function getDimensionsInPivotColumns(
...
+
+  return {
+    measures: measures.map((m) => {
+      return { id: m, title: m, type: PivotChipType.Measure };
+    }),
+    dimensions: dimensions.map((d) => {
+      return { id: d, title: d, type: PivotChipType.Dimension };
+    }),
+    time: time.map((t) => {
+      return { id: t, title: t, type: PivotChipType.Time };
```
djbarnwal, line 88: Lets use the `label` for titles here instead of `name` which is a dev identifier.

djbarnwal (approved):

Issue #3905, opened 2024-01-25
Update default spinner color to match new primary/secondary colors


jkhwu:
**Spinner**

Update our default spinner colors to replace pink/orange with our primary/secondary colors

![Screenshot 2024-01-25 at 9.45.48 AM.png](https://prod-files-secure.s3.us-west-2.amazonaws.com/ee565765-7662-4ad7-b80c-88457fa83d0c/bc8ee8e8-a697-4ea4-8cf3-c34078e15775/Screenshot_2024-01-25_at_9.45.48_AM.png)

https://www.figma.com/file/Qt6EyotCBS3V6O31jVhMQ7/RILL?type=design&node-id=458-12484&mode=design&t=2SlPTPxwRkk2Srpl-11

ericpgreen2:
As @bcolloran's PR #3933 shows, the spinner could look funky if the primary/secondary colors have large contrast...

![image](https://github.com/rilldata/rill/assets/14206386/164a9a50-391f-4416-b9c2-a16b261a0095)

What if the spinner just used the `secondary` color to make the gradient?

bcolloran:
2 cents: if users choose goofy looking colors that look terrible together, it will look terrible all over the app and the gradient with be a small thing. But if they choose nice looking colors, a gradient between two colors is much more elegant than just two shades of a single color. Another option (rather than using only `secondary`) would be to add a much lighter gradient stop in the middle of the spinner or something like that.

I'd be interested in nerding out on this after the brand update, but I think we should merge #3933 for now b/c this is an edge case, and open up a new issue to track refinements later if desired.

bcolloran:
also, by gradient interpolation happens in srgb, which is the ugliest possible option:

![image](https://github.com/rilldata/rill/assets/2380975/5f3597b5-3fcb-4ffc-b7b9-e09a24ce8595)

https://davidjohnstone.net/lch-lab-colour-gradient-picker

it looks like[ safari supports specifying a color space on gradients](https://css-tricks.com/color-spaces-for-more-interesting-css-gradients/), but that's no help to us. 

We could manually specify a gradient with like 10 steps in HSL or LCH and then the rgb interpolation between those smaller steps wouldn't be so ugly. That's what I'd recommend, but, again, as follow up work

jkhwu:
Let's merge https://github.com/rilldata/rill/pull/3933 and proceed with primary/secondary colors. We can alter later if we see examples of that looking bad.

bcolloran:
[...]

PR #3933, opened 2024-01-29
fix spinner gradient

### LINKED ISSUE #3913: Possible regression to custom themes
A non-blocking issue experienced while testing themes with https://github.com/rilldata/rill/pull/3863# 

It seems to be taking the primary color input ok, but the secondary color input is no longer changing the spinner color as it did before.

Here's a gif of 2 themes I'm trying out: the first one asks for the same shade of blue for primary and secondary, the 2nd one should have yellow for primary and lime for the spinner. Instead you can see that the spinner stays default pink in both cases.

Jan-25-2024 22-36-40 testing branding and themes

bcolloran:
fixes #3913 

default brand theme:
firefox_OCp4dYBc7c

custom user theme from orange to green
firefox_N3ldMBhFsY

### DIFF
--- a/web-admin/src/features/projects/ProjectBuilding.svelte
+++ b/web-admin/src/features/projects/ProjectBuilding.svelte
@@ -24,12 +24,7 @@
 <CtaLayoutContainer>
   <CtaContentContainer>
     <div class="h-36">
-      <Spinner
-        bg="linear-gradient(90deg, #22D3EE -0.5%, #6366F1 98.5%)"
-        status={EntityStatus.Running}
-        size="7rem"
-        duration={725}
-      />
+      <Spinner status={EntityStatus.Running} size="7rem" duration={725} />
     </div>
     <CtaHeader variant="bold"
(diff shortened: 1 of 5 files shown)


bcolloran:
fixes https://github.com/rilldata/rill/issues/3905 and https://github.com/rilldata/rill/issues/3913

jkhwu (changes requested):
At the time of my review, it seems the spinner color is darker than intended in the design, which was secondary 500 - primary 500. I think they're both set to 700 in this branch. Here's a gif of what I'm seeing.
Jan-29-2024 23-13-04

bcolloran:
@jkhwu updated to 500/500. Sorry about that, when I was doing the fix I was looking at the theme regression bug #3913 and trying to match that previous state, and I didn't spot the callout for the new gradients in #3905

jkhwu (approved):

ericpgreen2 (approved):

PR #3947, opened 2024-01-30
fix slow leaderboard loads

bcolloran:
like https://github.com/rilldata/rill/pull/3946/files, I think this fixes https://github.com/rilldata/rill-private-issues/issues/112, though I don't have an objective measure.

This PR uses an additional store in `getStateManagers`, but what actually seems to be the thing that does the trick is wrapping the update to that store in a conditional.

### DIFF
--- a/web-common/src/features/dashboards/leaderboard/LeaderboardHeader.svelte
+++ b/web-common/src/features/dashboards/leaderboard/LeaderboardHeader.svelte
@@ -20,7 +20,7 @@
   const {
     selectors: {
       contextColumn: {
-        widthPx,
+        contextColumn,
         isDeltaAbsolute,
         isDeltaPercent,
         isPercentOfTotal,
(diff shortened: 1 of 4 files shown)


web-common/src/features/dashboards/leaderboard/LeaderboardHeader.svelte
```diff
@@ -124,7 +130,7 @@
           on:click={toggleSortByActiveContextColumn}
           class="shrink flex flex-row items-center justify-end"
           aria-label="Toggle sort leaderboards by context column"
-          style:width={$widthPx}
+          style:width={widthPx}
```
ericpgreen2, line 133: Are we ever using other units besides "px"? It'd feel better if we appended the "px" at the very end, and move around `widthPx` as a `number` rather than a `string`.

ericpgreen2 (approved):
Fixes the query cancellation issue for me 💪

PR #3948, opened 2024-01-31
Remove query cancellation from StateManagers

bcolloran:
experiment with removing query cancellation

Here's what I see in a couple conditions 

(1) state as of changes in this PR:
- all manually query cancellation removed
- `observeContextColumnWidth` reactive statement commented out

Results:
- everything above the fold loads in about 15 seconds,
- loading keeps going for stuff below the fold
- no queries cancelled

https://github.com/rilldata/rill/assets/2380975/5ed313ac-b47a-4fed-84d0-6d3f0fd163d6 

(2) Main before this PR, but without `observeContextColumnWidth`
- manual query cancellation in place
- `observeContextColumnWidth` reactive statement commented out

Results:
- everything above the fold loads in about 15 seconds,
- loading keeps going for stuff below the fold
- no queries cancelled
https://github.com/rilldata/rill/assets/2380975/194cfdfc-769a-4fdb-9acb-ad7b5d115205


ericpgreen2 (commented):
It doesn't appear like these `cancelQueries` calls are adding any value to the dashboard. Out-of-the-box query cancellation works just fine without these explicit calls. Can you go ahead and remove the calls and related code?

However, for now can you keep the calls in `ModelBody.svelte` and `resource-invalidations.ts`? IIRC those are there as a ham-fisted way to cancel/invalidate outstanding queries.

bcolloran:
@AdityaHegde and @djbarnwal  would love to get your takes on this also before we go forward with any changes.

@ericpgreen2 and I have been looking at query cancellation in the context of https://github.com/rilldata/rill-private-issues/issues/112, and think that we might be doing extra manual cancellations in places where tanstack should be able to do that automatically. I think that as a team, we need to audit all of those manual cancellation and understand why they are needed, and if they can be removed.

@AdityaHegde, I've assigned this to you because I think you probably know the most about the intersection between runtime and client here. I think "done" on this would be code comments for any places we want to retain manual query cancellation. I don't mind making the changes in this PR if you just want to put those comments as review comments.

djbarnwal:
I haven't personally worked a lot with query cancellation and relied on tanstack for it. I am up for removing manual cancellation if they don't have any performance implications.

web-common/src/features/entity-management/resource-invalidations.ts
```diff
@@ -149,16 +149,16 @@ async function invalidateRemovedResource(
   switch (resource.meta.name.kind) {
     case ResourceKind.Source:
     case ResourceKind.Model:
-      queryClient.cancelQueries({
-        predicate: (query) => isProfilingQuery(query, resource.meta.name.name),
-      });
+      // queryClient.cancelQueries({
```
AdityaHegde, line 152: We need these since there could be a race condition in request for profile sent and in queue in backend vs the entity being deleted.

AdityaHegde: Same applies to line 158 and line 99 in ModelBody.

bcolloran:
@ericpgreen2 cleaning this up to get it over the line -- removed changes in ModelBody.svelte and resource-invalidations.ts, per you and @AdityaHegde's suggestions

PR #3951, opened 2024-01-31
queries.proto docstrings

egor-ryashin:


proto/rill/runtime/v1/queries.proto
```diff
@@ -54,6 +54,22 @@ service QueryService {
...
+  // | 2             | 2                   | 0                   | 0                  | Safari    |
+  // | 1             | 0                   | 1                   | N/A                | Chrome    |
+  // | 0             | 4                   | -4                  | -1.0               | Firefox   |
+  //
+  // ie. toplist:
+  // | measure1 | measure2 | dimension |
+  // |----------|----------|-----------|
+  // | 2        | 45       | Safari    |
+  // | 1        | 350      | Chrome    |
+  // | 0        | 25       | Firefox   |
```
bcolloran, line 72: Omg amazing! this goes so far above and beyond what I expected, and is so informative and helpful!!! <3  <3  <3

bcolloran (approved):
This is great! THANK YOU!!! 😍

begelundmuller (approved):

web-common/src/proto/gen/rill/runtime/v1/queries_pb.ts
```diff
@@ -3758,16 +3975,22 @@ export class ColumnTimeRangeResponse extends Message<ColumnTimeRangeResponse> {
...
 
   /**
+   * Not null
+   *
    * @generated from field: google.protobuf.Timestamp max = 2;
    */
   max?: Timestamp;
 
   /**
+   * Not null
```
bcolloran, line 3992: I think "not null" in a lot of these cases is a little confusing. If the type is `interval?: TimeRangeSummary_interval`, what i need to know is whether or not it's optional, not whether it's null or not -- null is not mentioned anywhere

egor-ryashin: I meant `not optional` by saying `not null`, should I rename it?

egor-ryashin: We can have 3 cases:
```
{
  "a": "string",
  "field": object
}
```
```
{
  "a": "string",
  "field": null
}
```
and 
```
{
 "a": "string"
}
```
`Not null` for `field` says last 2 cases aren't possible. `Not optional` doesn't negates the 2nd one though.

bcolloran: I think the way I'd recommend clarifying this distinction is by specifically calling out when a field may be null. When combined with optional/required, that could give us 4 cases I suppose. 

(case 1) "Required, Not null" or "Not optional, Not null"-- We know the field must be present; ("Required" is probably a bit better, I think that is more standard.)
(case 2) "Required, Nullable" -- We  know that the field must be present, but it may be null
(case 3) "Optional, Not Null"

(maybe case 4) "Optional, Nullable" 

I don't think there is any reason to try to collapse both of those orthogonal concepts down into a single term though, it's better to be full explicit I think. Thanks @egor-ryashin !

PR #3952, opened 2024-01-31
Add alert related APIs

AdityaHegde:
This is essentially a copy of repots API with 2 changes.
- Alerts cannot be triggered adhoc (based on UI mocks) so the trigger endpoint is removed.
- Alerts need to get the virtual file (based on UI mocks) so a new endpoint is added.

TODOs
- [x] Use virtual file to return the yaml.


begelundmuller (approved):
This looks solid!

PR #3953, opened 2024-01-31
Fix bug in line charts when the time controls are not ready

ericpgreen2:
Closes https://github.com/rilldata/rill-private-issues/issues/120

The workflow cited in the issue above triggered the `MeasureChart.svelte` component to render with an `undefined` time grain. That crashed the application.

This PR prevents the time series charts from rendering until the `timeControlsStore` has resolved with a defined `interval`.

### DIFF
--- a/web-common/src/features/dashboards/time-series/MetricsTimeSeriesCharts.svelte
+++ b/web-common/src/features/dashboards/time-series/MetricsTimeSeriesCharts.svelte
@@ -19,9 +19,9 @@
   import { getOrderedStartEnd } from "@rilldata/web-common/features/dashboards/time-series/utils";
   import { EntityStatus } from "@rilldata/web-common/features/entity-management/types";
   import { adjustOffsetForZone } from "@rilldata/web-common/lib/convertTimestampPreview";
-  import { TIME_GRAIN } from "@rilldata/web-common/lib/time/config";
   import { getAdjustedChartTime } from "@rilldata/web-common/lib/time/ranges";
   import { TimeRangePreset } from "@rilldata/web-common/lib/time/types";
+  import { TIME_GRAIN } from "../../../lib/time/config";
   import { runtime } from "../../../runtime-client/runtime-store";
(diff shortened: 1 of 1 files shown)


web-common/src/features/dashboards/time-series/MetricsTimeSeriesCharts.svelte
```diff
@@ -204,114 +204,117 @@
...
-          ? EntityStatus.Running
-          : EntityStatus.Idle}
-        on:expand-measure={() => {
-          metricsExplorerStore.setExpandedMeasureName(
-            metricViewName,
-            measure.name,
-          );
-        }}
-      />
+{#if $timeControlsStore.ready}
```
bcolloran, line 207: Did you investigate whether this conditional can be pushed deeper into the component tree?

I'm fine with going ahead and merging this to clear this bug as a quick stop gap fix, but I suspect we can be more granular here. I'll note that the whole `time-series` folder is one of the ones that [still needs to be cleaned up for strict null checks](https://github.com/rilldata/rill/issues/3424), which is presumably how this bug is possible in the first place.

So I'll happy be happy to dig deeper and clean this up when I get to this file, but in the meantime, I propose adding a FIXME comment before this condition that links back to this PR and the issue and explains briefly that we added this non-granular quick fix, but that we think we can do better. A little reminder like that will help me to be less afraid to refactor this guard out when I get to cleaning up this file

bcolloran (approved):
approving, but please add a FIXME comment before merging

web-common/src/features/dashboards/time-series/MetricsTimeSeriesCharts.svelte
ericpgreen2: Good call. I just made a more precise conditional, which removed the culpable type error 👍

bcolloran: love it! 👍

Issue #3907, opened 2024-01-25
Finer brand color adjustments to dashboard


jkhwu:
https://www.figma.com/file/Qt6EyotCBS3V6O31jVhMQ7/RILL?type=design&node-id=3224-55814&mode=design&t=2SlPTPxwRkk2Srpl-11

**Note:** When we have time we should go back and make the finer color adjustments shown above, such as the text color for Measure and Dimension labels (I believe this will better prompt users to click them to discover our other surfaces).

Also button colors in the upper right.

bcolloran:
Closing this -- some of the designs have been updated in the last month and some of this has already been implemented. Recommend opening new GH issues to track remaining design work.

PR #3954, opened 2024-01-31
use brand colors for bignum and leaderboard names

bcolloran:
partial fix of #3907, teeny tweak to colors in a couple places

image

### DIFF
--- a/web-common/src/app.css
+++ b/web-common/src/app.css
@@ -36,6 +36,10 @@
     @apply text-gray-600 dark:text-gray-200;
   }
 
+  .ui-copy-primary {
+    @apply text-primary-700 dark:text-primary-200;
+  }
+
   .ui-copy-strong {
     @apply text-gray-900 dark:text-white font-semibold;
   }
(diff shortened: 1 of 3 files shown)


jkhwu (approved):
Looks good @bcolloran !

ericpgreen2 (approved):

ericpgreen2:
I've noticed that the same header color should be applied to the dimension detail table as well.

jkhwu:
@ericpgreen2 I actually chose to change these colors to make them look like links and encourage users to mouse over and click to drill down (so far no first-time users we've observed have discovered Dimension Detail view on their own). So by that logic we may not need to apply the same header color to dimension detail view, since there's not another surface that would be revealed if we click that.

bcolloran:
@jkhwu, when you decide on which solution you prefer, would you mind capturing that decision in a new issue so they are tracked? They are likely to get lost in the shuffle as a comment in a merged PR. Thanks!

PR #3956, opened 2024-02-01
Force convert dimension values while loading from the url state

AdityaHegde:
We are facing issues with data type mismatch from the url state object and the actual dimension type. This was fixed on main but older urls with broken state are not handled.

This PR force converts the dimension values from the url state to keep things consistent.


runtime/queries/metricsview_data_types.go
```diff
@@ -0,0 +1,171 @@
...
+
+	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
+	"github.com/rilldata/rill/runtime"
+	"github.com/rilldata/rill/runtime/drivers"
+)
+
+type MetricsViewDataTypes struct {
+	MetricsViewName    string                               `json:"metrics_view_name,omitempty"`
+	MetricsView        *runtimev1.MetricsViewSpec           `json:"-"`
+	ResolvedMVSecurity *runtime.ResolvedMetricsViewSecurity `json:"security"`
```
begelundmuller, line 18: In the alerts PR, I'm making some refactors to stop passing full metrics view specs in here. Would recommend doing the same here (can wait until the other PR is merged so the correct funcs are available in this package). See [this diff](https://github.com/rilldata/rill/pull/3783/files#diff-4a01247b922ef6106d44f5553dc3fba7130fadb020de3ff8260e29b71f9c4e8e) for `runtime/queries/metricsview_aggregation.go` for an example.

runtime/queries/metricsview_data_types.go
```diff
@@ -0,0 +1,171 @@
...
+
+	Result []*runtimev1.MetricsViewDataType `json:"-"`
+}
+
+var _ runtime.Query = &MetricsViewDataTypes{}
+
+func (q *MetricsViewDataTypes) Key() string {
+	return fmt.Sprintf("TableColumns:%s", q.MetricsViewName)
+}
```
begelundmuller, line 27: Key prefix needs updating

runtime/queries/metricsview_data_types.go
```diff
@@ -0,0 +1,171 @@
...
+
+func (q *MetricsViewDataTypes) Deps() []*runtimev1.ResourceName {
+	return []*runtimev1.ResourceName{
+		{Kind: runtime.ResourceKindSource, Name: q.MetricsView.Table},
+		{Kind: runtime.ResourceKindModel, Name: q.MetricsView.Table},
+		{Kind: runtime.ResourceKindMetricsView, Name: q.MetricsViewName},
+	}
+}
```
begelundmuller, line 35: The deps should only return immediate dependencies. The responsibility to resolve the DAG should be an upstream concern (at the moment, this is handled by invalidating the cache based on `StateUpdatedOn`, which is increased in all cases where a ref changes).

proto/rill/runtime/v1/queries.proto
```diff
@@ -511,6 +518,21 @@ message MetricsViewTimeRangeResponse {
...
+}
+
+message MetricsViewDataTypesResponse {
+  repeated MetricsViewDataType data_types = 1;
+}
+
+message MetricsViewDataType {
+  string name = 1;
+  string type = 2;
+}
```
begelundmuller, line 534: It would be preferable to return a `rill.runtime.v1.StructType`, which is our generic type for representing schemas (e.g. see `MetricsViewAggregationResponse`). Then it will work predictably for both DuckDB and Druid (and any other OLAP backends we add in future).

runtime/queries/metricsview_data_types.go
```diff
@@ -0,0 +1,171 @@
...
+
+	return olap.WithConnection(ctx, priority, false, false, func(ctx context.Context, ensuredCtx context.Context, _ *sql.Conn) error {
+		// views return duplicate column names, so we need to create a temporary table
+		temporaryTableName := tempName("profile_columns_")
+
+		err = olap.Exec(ctx, &drivers.Statement{
+			Query:            fmt.Sprintf(`CREATE TEMPORARY TABLE "%s" AS (%s)`, temporaryTableName, q.buildMetricsViewDataTypesSQL()),
+			Priority:         priority,
+			ExecutionTimeout: defaultExecutionTimeout,
+		})
```
begelundmuller, line 77: As addressed previously, would suggest two refactors:

1. Return a `rill.runtime.v1.StructType` instead of raw type names, so we have predictable types for any OLAP backend (so the frontend doesn't get tied too closely to DuckDB type names specifically). It's available in the `Schema` property of the `drivers.Result` type returned from OLAP queries – it should be possible to construct a query that doesn't scan any data, but returns all dims and measures as fields just to get the types.
2. Make sure it also supports Druid – we need to support it as a first-class citizen for all metrics-related queries

proto/rill/runtime/v1/queries.proto
```diff
@@ -95,6 +95,13 @@ service QueryService {
     };
   }
 
+  // MetricsViewDataTypes Get the data types of measures and dimensions
+  rpc MetricsViewDataTypes(MetricsViewDataTypesRequest) returns (MetricsViewDataTypesResponse) {
```
begelundmuller, line 99: Maybe just call the API `MetricsViewTypes` or `MetricsViewSchema`? One less word ¯\_(ツ)_/¯

AdityaHegde: `MetricsViewSchema` Sounds better

runtime/queries/metricsview_schema.go
```diff
@@ -0,0 +1,118 @@
...
+	if groupBy != "" {
+		groupBy = fmt.Sprintf("GROUP BY %s", groupBy)
+	}
+
+	return fmt.Sprintf(
+		`SELECT %[1]s FROM "%[2]s" %[3]s %[4]s LIMIT 1`,
+		columns,                         // 1
+		q.TableName,                     // 2
+		strings.Join(unnestClauses, ""), // 3
+		groupBy,                         // 4
```
begelundmuller, line 116: 1. Use `safeName(q.TableName)` to escape it correctly
2. I wonder if putting the `LIMIT 1` at the end will cause a full table scan? Maybe doing `SELECT ... FROM (SELECT * FROM tbl LIMIT 1) GROUP BY ...` might protect against that? (Also curious if `LIMIT 0` would be adequate enough to still get the schema back?)

AdityaHegde: Updated to use `LIMIT 0`, was the fasted based on brenchmarks.

ericpgreen2:
Leaving a note that we discussed:

With the new call to `MetricsViewSchema`, it's not ideal that we need (another) network request before loading a dashboard. In the future, we could at least batch the call with the `TimeRangeSummary` request, but we should consider if these could be avoided outright.

ericpgreen2:
Approving from the frontend perspective, so we can test on the staging environment with our list of old URLs.

ericpgreen2 (approved):

PR #3961, opened 2024-02-01
convert measure filter dropdown to ShadCN components and resolve validation issues

briangregoryholmes:
The purpose of this PR is to convert the `MeasureFilterMenu` and `MeasureFilter` components to ShadCN and improve the UX around validation when adjusting measure filter parameters.

It also solves a bug where the `preventDefault` call on the MeasureChart function prevented the measure filter form from submitting when clicking outside.

Regarding validation UX, the new menu now restores the last previously valid state of the form on close.  If you change the threshold, for instance, from "Less Than" to "Between", the new form fields are no longer valid w/r/t the API. If you dismiss the menu in this state, the last valid state is restored, so you'll see "Less Than" when the menu is reopened. This is to bring consistency between what you see in the Chip (which is always valid w/r/t the API) and what is shown in the menu. This is true of other invalid states, like entering strings instead of numbers.

Additionally, this PR attempts to simplify the form code and provide more robust yup validation that accounts for different form states. To that end, this PR removes the "b" and "nb" notation that appeared to be local to this component and replaces it with `V1Operation` values, which is what was being sent to the API and stored in the serialized proto. This may be the wrong approach.

To achieve this, the following changes were made:

- Add the ShadCN builder pattern to the Chip so that it can be used as a menu trigger
- Remove the WithTogglableFloatingElement components and replace with ShadCN's DropdownMenu
- Move the conditional variable assignment in MeasureFilterMenu into the initialValue declaration
- Add a conditional to the schema for `value2`
- Change `InputV2` component to only display errors when a value is present and the input is not currently focused
- Remove preventDefault call in `onScrubStart` function
- Add the `onScrub` listener only when `onScrubStart` has fired

### DIFF
--- a/web-common/src/features/dashboards/filters/measure-filters/MeasureFilter.svelte
+++ b/web-common/src/features/dashboards/filters/measure-filters/MeasureFilter.svelte
@@ -1,10 +1,11 @@
 <script lang="ts">
+  import * as DropdownMenu from "@rilldata/web-common/components/dropdown-menu/";
   import { Chip } from "@rilldata/web-common/components/chip";
   import {
     ChipColors,
     measureChipColors,
   } from "@rilldata/web-common/components/chip/chip-types";
-  import WithTogglableFloatingElement from "@rilldata/web-common/components/floating-element/WithTogglableFloatingElement.svelte";
+
   import Tooltip from "@rilldata/web-common/components/tooltip/Tooltip.svelte";
   import TooltipContent from "@rilldata/web-common/components/tooltip/TooltipContent.svelte";
   import TooltipTitle from "@rilldata/web-common/components/tooltip/TooltipTitle.svelte";
@@ -18,10 +19,11 @@
   export let name: string;
   export let label: string | undefined = undefined;
   export let colors: ChipColors = measureChipColors;
-  export let expr: V1Expression | undefined;
+  export let expr: V1Expression | undefined = undefined;
+
+  const dispatch = createEventDispatcher();
 
   let active = !expr;
-  const dispatch = createEventDispatcher();
 
   function handleDismiss() {
     if (!expr) {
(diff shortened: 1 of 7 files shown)


web-common/src/features/dashboards/filters/measure-filters/MeasureFilter.svelte
```diff
@@ -32,60 +34,59 @@
...
+        {label}
+        builders={[builder]}
+        on:remove={() => dispatch("remove")}
+        outline
+        removable
+      >
+        <!-- remove button tooltip -->
+        <svelte:fragment slot="remove-tooltip">
+          <slot name="remove-tooltip-content">
+            remove {label}
```
ericpgreen2, line 71: We should use sentence capitalization: "Remove [label]"

web-common/src/features/dashboards/filters/measure-filters/MeasureFilter.svelte
```diff
@@ -32,60 +34,59 @@
...
+        <MeasureFilterBody {dimensionName} {expr} {label} slot="body" />
+      </Chip>
+      <div slot="tooltip-content" transition:fly={{ duration: 100, y: 4 }}>
+        <TooltipContent maxWidth="400px">
+          <TooltipTitle>
+            <svelte:fragment slot="name">{name}</svelte:fragment>
+            <svelte:fragment slot="description">{label || ""}</svelte:fragment>
+          </TooltipTitle>
+          {#if $$slots["body-tooltip-content"]}
+            <slot name="body-tooltip-content">click to edit the values</slot>
```
ericpgreen2, line 84: Again, sentence capitalization: "Click to edit the values"

web-common/src/features/dashboards/filters/measure-filters/MeasureFilterMenu.svelte
```diff
@@ -29,141 +30,132 @@
     },
   } = getStateManagers();
 
+  const initialValues = {
+    dimension: dimensionName,
```
ericpgreen2, line 34: It appears in practice this `dimensionName` is an empty string. Thoughts on initializing an actual dimension?

ericpgreen2 (approved):
Looks great to me! Left two small UX nits. Approving so you can merge at your discretion.

AdityaHegde (approved):

web-common/src/features/dashboards/filters/measure-filters/MeasureFilterMenu.svelte
briangregoryholmes: It's an empty string only when the measure filter form hasn't yet been submitted. So, when restoring from proto state (or after submission), it does have a value. I think it would be better give it a more explicit Option<string> type, but I'll put that in a different PR.

PR #3973, opened 2024-02-06
Persist some dashboard state to local storage and use it when returning to a dashboard

### LINKED ISSUE #3932: Feat: Restore User State
We would like to restore user UI state whenever a user comes back to a dashboard.

UI State to restore:
- Measure list (which measures are visible)
- Dimension/Leaderboard list (which dimensions are visible)
- Active Measure and sort order

Only the last state is needed, no need to retain a history of state.

It's not expected for last state to sync across devices but if possible lets go for it.

AdityaHegde:
## Checklist
- [x] Manual verification
- [ ] Unit test coverage
- [ ] E2E test coverage
- [x] Needs manual QA?

#### Issue addressed: 
closes #3932

#### Details:
We are storing the selection of measures, dimensions, selected leaderboard measure and leaderboard sort settings.

## Steps to Verify
1. Create a few dashboards.
2. Navigate to them and change the settings mentioned in details sections.
3. Close the page and go back to those dashboards.
4. The above settings should be retained.


web-common/src/features/dashboards/user-preferences.ts
ericpgreen2, line 1: "user preferences" seems like the wrong term. What about "last view state" or "most recent state"?

web-common/src/features/dashboards/user-preferences.ts
```diff
@@ -10,6 +14,13 @@ import type { Writable } from "svelte/store";
  */
 export interface LocalUserPreferences {
```
ericpgreen2, line 15: Why are these _local_ user preferences? They apply to cloud too, right?

web-common/src/features/dashboards/show-hide-selectors.ts
```diff
@@ -73,13 +74,20 @@ function createShowHideStore<Item>(
       };
     },
   ) as ShowHideSelectorStore;
+  const updateMethod = `updateVisible${fieldInMeta.replace(/\w/, (s) =>
+    s.toUpperCase(),
+  )}` as "updateVisibleMeasures" | "updateVisibleDimensions";
+  const localUserPreferences = getLocalUserPreferences();
```
ericpgreen2, line 80: This feels a bit hacky. I think it stems from the `createShowHideStore(...)` function signature. Rather than...
```typescript
function createShowHideStore<Item>(
  metricsViewName: string,
  metricsView: CreateQueryResult<V1MetricsView, RpcStatus>,
  fieldInMeta: keyof Pick<V1MetricsView, "dimensions" | "measures">,
  visibleFieldInStore: keyof Pick<
    MetricsExplorerEntity,
    "visibleMeasureKeys" | "visibleDimensionKeys"
  >,
  allVisibleFieldInStore: keyof Pick<
    MetricsExplorerEntity,
    "allMeasuresVisible" | "allDimensionsVisible"
  >,
  labelSelector: (i: Item) => string,
)
```
... this could be cleaner / simpler:
```typescript
function createShowHideStore<Item>(
  metricsViewName: string,
  type: "dimensions" | "measures",
)
```

web-common/src/features/dashboards/stores/DashboardStateProvider.svelte
```diff
@@ -11,8 +11,9 @@
   import { EntityStatus } from "../../entity-management/types";
 
   export let metricViewName: string;
+  export let additionalKey = ""; // used in cloud to add org and project
```
ericpgreen2, line 14: I think it'd be a cleaner interface if the `org` and `project` were the optional props. Then the key could be composed within the `DashboardStateProvider` component and not be leaked outside of it.

ericpgreen2 (changes requested):

web-common/src/features/dashboards/stores/dashboard-stores.ts
```diff
@@ -182,8 +182,10 @@ const metricViewReducers = {
         metricsView,
         fullTimeRange,
       );
+      mergeDashboardWithLocalPreferences(state.entities[name]);
```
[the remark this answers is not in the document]
AdityaHegde: True. But since this is updating the state it will be performant if we didnt replace the entire dashboard state. Will be especially impactful if and when we use immer.

web-common/src/features/dashboards/user-preferences.ts
AdityaHegde: Ya makes sense. I will just keep the local user preferences alone. Will create a separate store for dashboard state.

web-common/src/features/dashboards/state-managers/state-managers.ts
```diff
@@ -124,6 +130,10 @@ export function createStateManagers({
     contextColWidthDefaults,
   );
 
+  // TODO: once we move everything from dashboard-stores to here, we can get rid of the global
+  initPersistentDashboardStore(metricsViewName + (extraKey ?? ""));
```
ericpgreen2, line 134: I'd expect this to be org-project-dashboard, not dashboard-org-project, to match our page route hierarchy

web-common/src/features/dashboards/state-managers/state-managers.ts
```diff
@@ -68,9 +72,11 @@ export function getStateManagers(): StateManagers {
 export function createStateManagers({
   queryClient,
   metricsViewName,
+  extraKey,
 }: {
   queryClient: QueryClient;
   metricsViewName: string;
+  extraKey?: string;
```
ericpgreen2, line 79: `extraKeyPrefix`?

web-common/src/features/dashboards/stores/persistent-dashboard-state.ts
```diff
@@ -0,0 +1,79 @@
...
+        delete pd.sortDirection;
+        return pd;
+      });
+    },
+  };
+}
+
+export type PersistentDashboardStore = Readable<PersistentDashboardState> &
+  ReturnType<typeof persistentDashboardActions>;
+export function createPersistentDashboardStore(storeKey: string) {
```
ericpgreen2, line 55: Rather than creating a new store, did you consider taking our existing dashboard store and persisting it to local storage? Pros/cons? I'm a bit hesitant that we'll have to keep two stores in sync with the same data.

ericpgreen2 (changes requested):
The feature isn't working as expected for me. When I select filters on a dashboard, close the page, then re-open the dashboard, my filters are not applied. Am I missing something? Here's a JAM: https://jam.dev/c/f4899bfc-4d2a-4a16-96fe-83d6bcd7a2fd

I also left various comments inline.

web-common/src/features/dashboards/stores/persistent-dashboard-state.ts
AdityaHegde: A small section of the store is to be persisted. Check the details section for the exact fields. So it made sense to create a new store.

web-common/src/features/dashboards/state-managers/state-managers.ts
AdityaHegde: Not sure i understand. Is this comment in the wrong place?

AdityaHegde:
@ericpgreen2 Check the details section. Only a very small settings are to be persisted as per product.

web-common/src/features/dashboards/stores/persistent-dashboard-state.ts
```diff
@@ -0,0 +1,79 @@
...
+        delete pd.leaderboardMeasureName;
+        delete pd.dashboardSortType;
+        delete pd.sortDirection;
+        return pd;
+      });
+    },
+  };
+}
+
+export type PersistentDashboardStore = Readable<PersistentDashboardState> &
```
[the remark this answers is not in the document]
AdityaHegde: Since we use the suffix as `state` and the svelte store as `store` this would be confusing.

web-common/src/features/dashboards/stores/dashboard-stores.ts
```diff
@@ -182,8 +182,12 @@ const metricViewReducers = {
         metricsView,
         fullTimeRange,
       );
+      state.entities[name] = mergeDashboardWithLocalPreferences(
```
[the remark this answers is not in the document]
AdityaHegde: `restorePersistedDashboardState` feels better.

web-common/src/features/dashboards/stores/persistent-dashboard-state.ts
ericpgreen2: Got it, I didn't realize that only a subset of the store is to be persisted.

Though I'm still wondering if it'd be simpler to persist the store in its entirety, then we can just choose which fields to restore in the `restorePersistedDashboardState()` function. Then it'd be easy to add/subtract fields we want to restore by just editing that one function.

Else, needing to remember to update both stores, I worry the two stores will often get out-of-sync and the feature will appear buggy.

ericpgreen2: Another option to avoid synchronization issues – the persisted dashboard store could be a `derived` store that consumes the root dashboard store

ericpgreen2:
> @ericpgreen2 Check the details section. Only a very small settings are to be persisted as per product.

Oh, I hadn't realized that. Yes, then for these fields it works like a charm!

web-common/src/features/dashboards/stores/persistent-dashboard-state.ts
AdityaHegde: True. We can do that once the immer solution is in.

ericpgreen2: Okay, just opened up an issue to track this: https://github.com/rilldata/rill-private-issues/issues/164

ericpgreen2 (approved):

PR #3977, opened 2024-02-06
Added output format flag for CSV, TSV or JSON output for all commands

rakeshsharma14317:
**Usage** 
use `--format [csv,json,tsv]` with any command and output will be in required format, Better usecase would be to use this flag for **list** commands where we expected couple of rows

### DIFF
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -132,6 +132,7 @@ func runCmd(ctx context.Context, ver config.Version) error {
 	rootCmd.SilenceUsage = true
 	// we want to override some error messages
 	rootCmd.SilenceErrors = true
+	rootCmd.PersistentFlags().Var(printer.NewFormatValue(printer.Human, &format), "format", "Output format. Possible values: [human, json, csv, tsv]")
 	rootCmd.PersistentFlags().BoolP("help", "h", false, "Print usage") // Overrides message for help
 	rootCmd.PersistentFlags().BoolVar(&cfg.Interactive, "interactive", true, "Prompt for missing required parameters")
 	rootCmd.Flags().BoolP("version", "v", false, "Show rill version") // Adds option to get version by passing --version or -v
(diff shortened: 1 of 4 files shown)


cli/cmd/root.go
```diff
@@ -132,6 +132,7 @@ func runCmd(ctx context.Context, ver config.Version) error {
 	rootCmd.SilenceUsage = true
 	// we want to override some error messages
 	rootCmd.SilenceErrors = true
+	rootCmd.PersistentFlags().Var(printer.NewFormatValue(printer.Human, &format), "format", "Output format. Possible values: [human, json, csv, tsv]")
 	rootCmd.PersistentFlags().BoolP("help", "h", false, "Print usage") // Overrides message for help
 	rootCmd.PersistentFlags().BoolVar(&cfg.Interactive, "interactive", true, "Prompt for missing required parameters")
```
begelundmuller, line 137: Move the format option to after "interactive"

begelundmuller: Also, can we remove support for TSV? Don't think that'll ever be useful

cli/pkg/printer/printer.go
```diff
@@ -9,6 +10,7 @@ import (
 	"strings"
 
 	"github.com/fatih/color"
+	"github.com/gocarina/gocsv"
```
begelundmuller, line 13: What's wrong with the native `encoding/csv` package? It would be nice to avoid a new external dependency for this

cli/pkg/printer/printer.go
```diff
@@ -71,10 +108,64 @@ func (p *Printer) PrintResource(v interface{}) error {
...
+	if p.resourceOut != nil {
+		out = p.resourceOut
+	}
+
+	type csvvaluer interface {
+		MarshalCSVValue() interface{}
+	}
+	if c, ok := v.(csvvaluer); ok {
+		v = c.MarshalCSVValue()
+	}
```
begelundmuller, line 131: What is the need for this? It doesn't look like anything in our code implements `MarshalCSVValue`

begelundmuller (changes requested):

begelundmuller (approved):

PR #3983, opened 2024-02-07
Add pivot state to url

AdityaHegde:
This PR adds all the fields in `PivotState` to the url state.


djbarnwal (approved):
Tested locally, works smoothly. Thanks for the changes!

PR #3986, opened 2024-02-07
initial time grain support for pivot table

briangregoryholmes:
This PR adds the initial code necessary for supporting time granularity in the pivot table.

- Changes row/column data structure in the pivot store
- Removes related utils from pivot-utils
- Adds selectors for rows and columns
- Changes setPivotRows and setPivotColumns function to match new structure
- Adds a new PivotChip component to wrap Chip
- Displays a dropdown menu (not fully styled) for time pills in the pivot header
- Does not support adding multiple time pills (requires more API work)

### DIFF
--- a/web-common/src/features/dashboards/stores/dashboard-stores.ts
+++ b/web-common/src/features/dashboards/stores/dashboard-stores.ts
@@ -32,6 +32,7 @@ import {
   SortDirection,
   SortType,
 } from "web-common/src/features/dashboards/proto-state/derived-types";
+import { PivotChipType, type PivotChipData } from "../pivot/types";
 
 export interface MetricsExplorerStoreType {
   entities: Record<string, MetricsExplorerEntity>;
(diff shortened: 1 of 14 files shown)


djbarnwal (commented):
Hey, briefly looked into the PR. I see that we are introducing `PivotChipData` as a type for the rows and columns
``` 
export type PivotChipData = {
  id: string;
  title: string;
  type: PivotChipType;
};
```
This is in the dashboard state which is later serialized in the proto. I am hesitant on adding generated info like `type` and `title` which can be extracted later using just the `id`. My primary concern is the URL length and proto complexity. 
cc @AdityaHegde if you have any thoughts on this

AdityaHegde:
Looking at the type `PivotChipData` we should be able to map a name to measure/dimension and then the type based on its type. So we can still store just the names in the url and map them back. Lemme know if there are any challenges in mapping it back.

web-common/src/features/dashboards/stores/dashboard-stores.ts
```diff
@@ -229,15 +231,41 @@ const metricViewReducers = {
...
+      value.forEach((val) => {
+        if (val.type === PivotChipType.Measure) {
+          measures.push(val);
+        } else {
+          dimensions.push(val);
+        }
+      });
+
+      metricsExplorer.pivot.columns = {
+        ...metricsExplorer.pivot.columns,
```
djbarnwal, line 265: Is this needed here?

web-common/src/features/dashboards/time-dimension-details/TDDHeader.svelte
```diff
@@ -112,21 +113,47 @@
...
+            type: PivotChipType.Dimension,
+          },
+        ]
+      : [];
+    metricsExplorerStore.createPivot(
+      metricViewName,
+      { dimension: rowDimensions },
+      {
+        dimension: [
+          { id: timeDimension, title: timeDimension, type: PivotChipType.Time },
```
djbarnwal, line 146: Here `id` is the grain right and not the name of time dimension?

web-common/src/features/dashboards/time-dimension-details/TDDHeader.svelte
```diff
@@ -112,21 +113,47 @@
...
+      pivot.columns.measure.length ||
+      pivot.columns.dimension.length
+    ) {
       showReplacePivotModal = true;
     } else {
       createPivot();
     }
   }
+
+  // This needs to be reworked - bgh
```
djbarnwal, line 127: Is this a stale comment?

djbarnwal (approved):
Left a couple of comments. Approving so you can merge with the changes in at your convenience.

web-common/src/features/dashboards/time-dimension-details/TDDHeader.svelte
briangregoryholmes: Not at the time. Looks like you reworked it in https://github.com/rilldata/rill/pull/4002

PR #3987, opened 2024-02-08
Fix POST queries not getting cancelled in profiling queries

AdityaHegde:
## Checklist
- [x] Manual verification
- [ ] Unit test coverage
- [ ] E2E test coverage
- [x] Needs manual QA?

#### Issue addressed: 
We are still seeing issues with query cancellation outside of dashboard.

#### Details:
Adding `signal` setting to all POST profiling queries.


ericpgreen2 (approved):
The changes make sense, but I'm not seeing much difference in behavior. On both `main` and this branch, I observe profiling queries get cancelled when I edit a model. On both branches though, it seems only a fraction of the outstanding queries get cancelled. Regardless, this PR looks beneficial.

AdityaHegde:
Ya it didnt make too much of a difference. My guess is the http request queue handles it all. But if there was a timeseries request in flight then this could help.

PR #3997, opened 2024-02-08
Fix #3992, and other cleanups

### LINKED ISSUE #3992: TypeErrors in dimension table crash the app
image

image

bcolloran:
Fixes #3992, and some related type cleanups on the side


web-common/src/features/dashboards/dimension-table/DimensionValueHeader.svelte
```diff
@@ -38,11 +40,14 @@
 
   $: atLeastOneSelected = !!selectedIndex?.length;
 
-  const getCellProps = (row) => {
-    const value = rows[row.index][column.name];
+  const getCellProps = (row: VirtualItem) => {
+    const value = rows[row.index]?.[column.name];
```
bcolloran, line 44: I believe this invalid array access was the actual cause of the crash, everything else is just cleanup to make it clearer what data we're dealing with here

ericpgreen2 (approved):
Sweet

PR #3998, opened 2024-02-08
Read-only filter chips

ericpgreen2:
This PR refactors the filter chips to accommodate the "read-only" mode that we need for the Create Alert dialog in PR #3881.

Here are the normal, interactive chips:
image

Here are the read-only versions:
image


web-common/src/features/dashboards/filters/dimension-filters/DimensionFilter.svelte
```diff
@@ -0,0 +1,82 @@
...
+
+  $: isInclude = !$dashboardStore.dimensionFilterExcludeMode.get(name);
+
+  let isOpen = false;
+  let searchText = "";
+  let allValues: Record<string, string[]> = {};
+  let topListQuery: ReturnType<typeof getFilterSearchList> | undefined;
+
+  $: if (isOpen) {
+    topListQuery = getFilterSearchList(StateManagers, {
```
AdityaHegde, line 33: We should disable this query if it is readonly. Even though `isOpen` is not set to true right now we might end up changing it. Perhaps just a comment would be enough to make sure future devs will keep it in mind?

AdityaHegde (approved):
Just a small nit. Otherwise LGTM

web-common/src/features/dashboards/filters/dimension-filters/DimensionFilter.svelte
ericpgreen2: Kk, I just added the conditional now.

PR #3999, opened 2024-02-09
Add component set for multi-step dialog tabs

ericpgreen2:
This PR adds a new component set for multi-step dialog tabs. These tabs are only to be used for multi-step dialogs, such as the Create Alert dialog in PR #3881.

The tabs are bootstrapped from ShadCN Svelte's tab component set, but they're heavily modified.

image

Additionally, this PR does the following cleanup:
- Upgrades Storybook (which was crashing on launch for me)
- Renames `components/dialog-v2` to `components/dialog`


web-common/src/components/dialog/tabs/__stories__/DialogTabs.stories.svelte
```diff
@@ -0,0 +1,86 @@
...
+    DialogTitle,
+  } from "@rgossiaux/svelte-headlessui";
+  import { Meta, Story } from "@storybook/addon-svelte-csf";
+  import * as DialogTabs from "../";
+  import Button from "../../../button/Button.svelte";
+
+  let selectedTabIndex = 0;
+  const tabs = ["Tab 1", "Tab 2", "Tab 3"];
+
+  function handleNextTab() {
```
AdityaHegde, line 14: Should we wrap this next/back/done functionality in a component as well?

AdityaHegde (approved):
Just a small question

web-common/src/components/dialog/tabs/__stories__/DialogTabs.stories.svelte
ericpgreen2: Oh interesting idea – maybe if/when we have another multi-panel dialog we can look at doing that.

PR #4000, opened 2024-02-09
Use `esnext` (as does SvelteKit's `tsconfig.json`)

ericpgreen2:
Upon upgrading some dependencies in #3999, I started to get these errors:

image

This PR updates the root `.tsconfig`'s `lib`, `module`, and `target` to use `esnext`. FWIW, this is the same configuration as in SvelteKit's `.tsconfig`, and it'd be nice to converge.


bcolloran (approved):
lgtm!

PR #4003, opened 2024-02-09
Pass through measure filters to pivot queries

AdityaHegde:
Treating aggregation as timeseries and sending in `ResolvedMeasureFilter` to all the queries fired in pivot queries.

### DIFF
--- a/web-common/src/features/dashboards/pivot/pivot-data-store.ts
+++ b/web-common/src/features/dashboards/pivot/pivot-data-store.ts
@@ -1,3 +1,4 @@
+import { measureFilterResolutionsStore } from "@rilldata/web-common/features/dashboards/filters/measure-filters/measure-filter-utils";
 import { mergeFilters } from "@rilldata/web-common/features/dashboards/pivot/pivot-merge-filters";
 import { useMetricsView } from "@rilldata/web-common/features/dashboards/selectors/index";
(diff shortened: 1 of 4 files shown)


web-common/src/features/dashboards/pivot/pivot-data-store.ts
```diff
@@ -46,8 +47,13 @@ import {
  */
 function getPivotConfig(ctx: StateManagers): Readable<PivotDataStoreConfig> {
   return derived(
-    [useMetricsView(ctx), ctx.dashboardStore, useTimeControlStore(ctx)],
-    ([metricsView, dashboardStore, timeControls]) => {
+    [
+      useMetricsView(ctx),
+      ctx.dashboardStore,
+      useTimeControlStore(ctx),
+      measureFilterResolutionsStore(ctx),
```
djbarnwal, line 54: Since measure filter have no change in the pivot flow do you think we can just directly use it in `createPivotAggregationRowQuery` for all queries that call it?

AdityaHegde: It feels like this is the perfect place. Since all settings from dashboard needed for querying that is not changed is passed through here it fits pretty well. It also avoids adding derived and set to get the async data from this piece.

djbarnwal (approved):

PR #4009, opened 2024-02-09
Aggregation filter using JOIN

egor-ryashin:


runtime/queries/metricsview_aggregation.go
```diff
@@ -417,7 +417,12 @@ func (q *MetricsViewAggregation) buildMetricsAggregationSQL(mv *runtimev1.Metric
 
 	groupCols := make([]string, 0, len(q.Dimensions))
 	unnestClauses := make([]string, 0)
+	joinConditions := make([]string, 0, len(q.Dimensions))
+	selfJoinCols := make([]string, 0, len(q.Dimensions)+1)
+	selfJoinTableAlias := tempName("self_join")
+	nonNullValue := tempName("non_null")
 	var args []any
+	var subSelectArgs []any
```
begelundmuller, line 425: I'm afraid this takes the code to a barely readable level and will make future changes too hard and error-prone. I don't have a great proposal, but somehow this filter case needs to be isolated better in the code (and not interleaved with the existing implementation with various if statements throughout) and also have an appropriate docstring explaining the case being handled.

I wonder if it can either be refactored to a single `if` statement at the end of the `buildMetricsAggregationSQL` function, or otherwise we may need to have a dedicated `buildMetricsAggregationSingleMeasureFilterSQL` function that gets invoked instead?

runtime/queries/metricsview_aggregation.go
```diff
@@ -579,34 +602,61 @@ func (q *MetricsViewAggregation) buildMetricsAggregationSQL(mv *runtimev1.Metric
...
+					SELECT %[10]s FROM %[2]s %[3]s %[9]s %[5]s %[6]s
+				) %[7]s 
+				ON (%[8]s) %[13]s %[11]s  OFFSET %[12]d
+				`,
+				strings.Join(selfJoinCols, ", "),      // 1
+				safeName(mv.Table),                    // 2
+				strings.Join(unnestClauses, ""),       // 3
+				whereClause,                           // 4
+				groupClause,                           // 5
+				havingClause,                          // 6
```
begelundmuller, line 630: Wouldn't the `HAVING` clause need to be applied after the left join?

runtime/queries/metricsview_aggregation.go
```diff
@@ -579,34 +602,61 @@ func (q *MetricsViewAggregation) buildMetricsAggregationSQL(mv *runtimev1.Metric
...
+				strings.Join(joinConditions, " AND "), // 8
+				filterWhereClause,                     // 9
+				strings.Join(selectCols, ", "),        // 10
+				limitClause,                           // 11
+				q.Offset,                              // 12
+				orderClause,                           // 13
+			)
+
+			args = append(args, subSelectArgs...)
+			fmt.Println("sql ", sql, args)
```
begelundmuller, line 641: Leftover print statement

begelundmuller (changes requested):

runtime/queries/metricsview_aggregation.go
egor-ryashin: `args` collection prevents from refactoring it to a single `if`

runtime/queries/metricsview_aggregation.go
```diff
@@ -579,32 +572,88 @@ func (q *MetricsViewAggregation) buildMetricsAggregationSQL(mv *runtimev1.Metric
...
+		/*
+			Example:
+			SELECT t.d1, t.d2, t.d3, t2.m1 (SELECT d1, d2, d3, m1 FROM t WHERE ...  GROUP BY d1, d2, d3 HAVING ... ) t LEFT JOIN (
+				SELECT d1, d2, d3, m1 FROM t WHERE ... AND (d4 = 'Safari') GROUP BY d1, d2, d3 HAVING ...
+			)  t2 ON (COALESCE(t.d1, 'val') = COALESCE(t2.d1, 'val') and COALESCE(t.d2, 'val') = COALESCE(t2.d2, 'val') and ...) ORDER BY ...
+
+			This JOIN mirrors functionality of SELECT d1, d2, d3, m1 FILTER (WHERE d4 = 'Safari') FROM t WHERE... GROUP BY d1, d2, d3
+			bacause FILTER cannot be applied for arbitrary measure, ie sum(a)/1000
+		*/
+		if len(q.Measures) == 1 && q.Measures[0].Filter != nil {
```
begelundmuller, line 584: We should return an error if more than 1 measure has a filter (since that's not supported) or if it is used with `pivot_on`

runtime/queries/metricsview_aggregation.go
```diff
@@ -579,32 +572,88 @@ func (q *MetricsViewAggregation) buildMetricsAggregationSQL(mv *runtimev1.Metric
...
+				orderClause = "ORDER BY " + strings.Join(jsc, ", ")
+			}
+			sql = fmt.Sprintf(`
+					SELECT %[1]s FROM (
+						SELECT %[10]s FROM %[2]s %[3]s %[4]s %[5]s %[6]s 
+					) %[2]s 
+					LEFT JOIN (
+						SELECT %[10]s FROM %[2]s %[3]s %[9]s %[5]s %[6]s
+					) %[7]s 
+					ON (%[8]s) %[13]s %[11]s  OFFSET %[12]d
```
begelundmuller, line 619: I might be missing something, but the `having` clause seems to be applied in the sub-selects, but shouldn’t it be outside, after the join? For example, if you have having `measure1 < 10`, some dimension values might be excluded from the first sub-selects which are included in the second sub-select.

runtime/queries/metricsview_aggregation.go
```diff
@@ -579,32 +572,88 @@ func (q *MetricsViewAggregation) buildMetricsAggregationSQL(mv *runtimev1.Metric
...
+				strings.Join(selfJoinCols, ", "),      // 1
+				safeName(mv.Table),                    // 2
+				strings.Join(unnestClauses, ""),       // 3
+				whereClause,                           // 4
+				groupClause,                           // 5
+				havingClause,                          // 6
+				selfJoinTableAlias,                    // 7
+				strings.Join(joinConditions, " AND "), // 8
+				measureWhereClause,                    // 9
+				strings.Join(selectCols, ", "),        // 10
```
begelundmuller, line 630: If using the same `selectCols` in both sub-selects, then won't it be computing the measure twice, one without the measure filter and the second with the measure filter? The non-filtered value would not be used, so should be cheaper to not compute it.

begelundmuller (commented):
This looks cleaner. Have you also validated that it works on Druid?

runtime/queries/metricsview_aggregation.go
```diff
@@ -579,32 +584,127 @@ func (q *MetricsViewAggregation) buildMetricsAggregationSQL(mv *runtimev1.Metric
...
+			)  t2 ON (COALESCE(t.d1, 'val') = COALESCE(t2.d1, 'val') and COALESCE(t.d2, 'val') = COALESCE(t2.d2, 'val') and ...)
+			WHERE t2.m1 > 10
+			ORDER BY ...
+			LIMIT 100
+			OFFSET 0
+
+			This JOIN mirrors functionality of SELECT d1, d2, d3, m1 FILTER (WHERE d4 = 'Safari') FROM t WHERE... GROUP BY d1, d2, d3
+			bacause FILTER cannot be applied for arbitrary measure, ie sum(a)/1000
+		*/
+		if filterCount == 1 {
```
begelundmuller, line 600: Can we pull this case into a separate function? The level of nesting and complexity here makes it hard to follow

runtime/queries/metricsview_aggregation.go
begelundmuller: This comment is still unaddressed

runtime/queries/metricsview_aggregation.go
```diff
@@ -579,32 +584,127 @@ func (q *MetricsViewAggregation) buildMetricsAggregationSQL(mv *runtimev1.Metric
...
+			if dialect == drivers.DialectDruid {
+				druidGroupBy = groupClause
+			}
+
+			sql = fmt.Sprintf(`
+					SELECT %[1]s FROM (
+						SELECT %[10]s FROM %[2]s %[3]s %[4]s %[5]s %[6]s 
+					) %[2]s 
+					LEFT JOIN (
+						SELECT %[10]s FROM %[2]s %[3]s %[9]s %[5]s %[6]s
```
begelundmuller, line 658: `%[6]s` represents the having clause – does it make sense to have it in the subqueries when it will be enforced after the left join? 

At least in the first sub-query, we should be able to completely remove calculation of the measure (see other comment), so having it there seems weird.

begelundmuller (changes requested):

runtime/queries/metricsview_aggregation.go
egor-ryashin: I'm not sure it will be much more performant than "SELECT DISTINCT", but it requires additional algorithm branching

runtime/queries/metricsview_aggregation.go
egor-ryashin: assuming Druid won't push down the WHERE, the subselect generates more rows for joining

runtime/queries/metricsview.go
```diff
@@ -242,6 +242,10 @@ func columnIdentifierExpression(mv *runtimev1.MetricsViewSpec, aliases []*runtim
 			switch alias.Type {
 			case runtimev1.MetricsViewComparisonMeasureType_METRICS_VIEW_COMPARISON_MEASURE_TYPE_UNSPECIFIED,
 				runtimev1.MetricsViewComparisonMeasureType_METRICS_VIEW_COMPARISON_MEASURE_TYPE_BASE_VALUE:
+				splits := strings.Split(alias.Name, ".")
+				if len(splits) > 1 {
+					return safeName(splits[0]) + "." + safeName(splits[1]), true
+				}
```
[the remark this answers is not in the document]
egor-ryashin: The other way I need to refactor it to pass information that it's a complex name

runtime/queries/metricsview_aggregation.go
egor-ryashin: it will be long signature function

begelundmuller (approved):

PR #4020, opened 2024-02-12
Add button for selecting pivot fields via dropdown

briangregoryholmes:
This PR adds a button to the pivot columns and rows drag lists that allows the user to select pivot fields via dropdown. To support this, the dimension and measure aggregators have been moved to the pivot selectors.

### DIFF
--- a/web-common/src/features/dashboards/state-managers/selectors/pivot.ts
+++ b/web-common/src/features/dashboards/state-managers/selectors/pivot.ts
@@ -1,7 +1,47 @@
 import type { DashboardDataSources } from "./types";
+import { PivotChipType } from "../../pivot/types";
 
 export const pivotSelectors = {
   showPivot: ({ dashboard }: DashboardDataSources) => dashboard.pivot.active,
   rows: ({ dashboard }: DashboardDataSources) => dashboard.pivot.rows,
   columns: ({ dashboard }: DashboardDataSources) => dashboard.pivot.columns,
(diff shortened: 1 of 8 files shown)


web-common/src/features/dashboards/state-managers/selectors/pivot.ts
```diff
@@ -1,7 +1,47 @@
...
+import { PivotChipType } from "../../pivot/types";
 
 export const pivotSelectors = {
   showPivot: ({ dashboard }: DashboardDataSources) => dashboard.pivot.active,
   rows: ({ dashboard }: DashboardDataSources) => dashboard.pivot.rows,
   columns: ({ dashboard }: DashboardDataSources) => dashboard.pivot.columns,
+  measures: ({ metricsSpecQueryResult, dashboard }: DashboardDataSources) => {
+    const measures =
+      metricsSpecQueryResult.data?.measures?.filter(
+        (d) => d.name && dashboard.visibleMeasureKeys.has(d.name),
```
djbarnwal, line 11: Wondering if we should only limit to visible measures and dimensions in the pivot. We don't have controls to change the visibility in the pivot view. Probably better if we show all available dimensions and measures cc @mindspank What do you recon we should do here?

djbarnwal (approved):
Looks good and works as intended 👍

web-common/src/features/dashboards/state-managers/selectors/pivot.ts
mindspank: @djbarnwal Yes I would display all dimensions and measures

PR #4024, opened 2024-02-12
Temporary workaround for DragList quirk

briangregoryholmes:
This PR enables a temporary workaround for the pivot drag list to resolve a limitation with the chosen drag-and-drop library. A more permanent solution will be put in place after discussing with the team.

The limitation is that child elements of the designated drag area are automatically turned into draggable elements behind the scenes, meaning our AddField button cannot be a child of that element without presenting (somewhat non-obvious) cursor and UX issues.

As such, this PR moves the AddField button outside of the drag area temporarily.

### DIFF
--- a/web-common/src/components/button/Button.svelte
+++ b/web-common/src/components/button/Button.svelte
@@ -229,7 +229,7 @@
   /* ADD BUTTON STYLES */
 
   .add {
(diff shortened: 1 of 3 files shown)


nishantmonu51:
may not be relevant anymore, closing it

PR #4033, opened 2024-02-13
Adityahegde/alerts page

AdityaHegde:
## Checklist
- [ ] Manual verification
- [ ] Unit test coverage
- [ ] E2E test coverage
- [ ] Needs manual QA?

### DIFF
--- a/web-common/src/features/dashboards/state-managers/selectors/dimension-filters.ts
+++ b/web-common/src/features/dashboards/state-managers/selectors/dimension-filters.ts
@@ -99,27 +99,36 @@ export function getDimensionFilterItems(
   dashData: AtLeast<DashboardDataSources, "dashboard">,
 ) {
   return (dimensionIdMap: Map<string, MetricsViewSpecDimensionV2>) => {
-    if (!dashData.dashboard.whereFilter) return [];
-
-    const filteredDimensions: DimensionFilterItem[] = [];
-    const addedDimension = new Set<string>();
-    forEachIdentifier(dashData.dashboard.whereFilter, (e, ident) => {
-      if (addedDimension.has(ident) || !dimensionIdMap.has(ident)) return;
-      const dim = dimensionIdMap.get(ident);
-      if (!dim) {
-        return;
-      }
-      addedDimension.add(ident);
-      filteredDimensions.push({
-        name: ident,
-        label: getDimensionDisplayName(dim),
-        selectedValues: getValuesInExpression(e),
-      });
-    });
-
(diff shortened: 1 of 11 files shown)


ericpgreen2:
A few observations in the page below:
- We're missing an Alert breadcrumb
- The context menu shows "Edit report" and "Delete report". Edit is going to take some effort to refactor the form (similar to the Scheduled Report form), so let's remove that one for now.
- "Report created by" should read "Alert created by"
- In the header, there's an extra bullet to the right of "Eric Green"
- "New cases for > "10"" should read "New cases > 10"
- "Last 1 run" should read "Last 1 check"
- "Checking" should read "Checked" (given "Failed" indicates the alert check is complete)

image

web-common/src/features/dashboards/state-managers/selectors/dimension-filters.ts
```diff
@@ -99,27 +99,36 @@ export function getDimensionFilterItems(
...
-        selectedValues: getValuesInExpression(e),
-      });
+    return getDimensionFilterItemsForFilter(
+      dimensionIdMap,
+      dashData.dashboard.whereFilter,
+    );
+  };
+}
+
+export function getDimensionFilterItemsForFilter(
```
ericpgreen2, line 109: This function name is a little confusing. Maybe  just `getDimensionFilters()` would be more readable?

web-common/src/features/dashboards/state-managers/selectors/measure-filters.ts
```diff
@@ -85,37 +85,59 @@ export const getMeasureFilterItems = (
...
+          addedMeasure,
+        ),
+      );
     }
 
     return filteredMeasures;
   };
 };
 
+export function getMeasureFilterItemsForFilter(
```
ericpgreen2, line 102: Similarly, maybe `getMeasureFilters()` would be more readable?

ericpgreen2: Though I see there's already a `getMeasureFilters()` function in `features/dashboards/filters/measure-filters/measure-filter-utils.ts`. Do we need both? Is it that the function below is a small transformation on top of the existing function?

ericpgreen2 (approved):
This is great! Just a couple questions on the filter utils.

web-common/src/features/dashboards/state-managers/selectors/measure-filters.ts
AdityaHegde: They are all slightly different. Makes sense to rename the one in `measure-filter-utils.ts`

Issue #3980, opened 2024-02-06
Fix: Measure filter seems to filter out extra data


ericpgreen2:
JAM video: https://jam.dev/c/7caa6124-464b-4208-be58-295ecb67d8ec

In the video, note that filtering for "New cases" > 0, both 'Panama' (50 new cases) and 'Turks and Caicos Islands' (10 new cases) get filtered out.

Here's a public dashboard where you can observe the behavior: [COVID dashboard](https://ui.rilldata.com/ericpgreen/rill-test-covid/covid?state=CgUKA1A0VxgFKgltZWFzdXJlXzE4AEgBWAFgBGoTQW1lcmljYS9Mb3NfQW5nZWxlc3gCgAEBmAH%252F%252F%252F%252F%252F%252F%252F%252F%252F%252F%252F8BogEqGigICBIkGiIICRILCgljb250aW5lbnQSERIPGg1Ob3J0aCBBbWVyaWNhqgEwCghsb2NhdGlvbhIkGiIICBIeGhwIBRILCgltZWFzdXJlXzESCxIJEQAAAAAAAAAA).

AdityaHegde:
This needs backend refactor with subqueries to get the accurate support.

nishantmonu51:
Reassigning to @begelundmuller for getting prioritized on platform side.

begelundmuller:
> This needs backend refactor with subqueries to get the accurate support.

@AdityaHegde can you elaborate on what the issue is here?

egor-ryashin:
That's because `toplist` query wasn't given a filter while `compare-toplist` had it. This the response from `toplist`:
```
{
    "meta": [
        {
            "name": "domain",
            "type": "CODE_STRING",
            "nullable": true
        },
        {
            "name": "total_records",
            "type": "CODE_INT64",
            "nullable": true
        }
    ],
    "data": [
        {
            "domain": "facebook.com",
            "total_records": 15609
        },
        {
            "domain": "msn.com",
            "total_records": 15526
        },
        {
            "domain": "google.com",
            "total_records": 15119
        },
        {
            "domain": "news.yahoo.com",
            "total_records": 14870
        },
        {
            "domain": "instagram.com",
            "total_records": 13104
        },
        {
            "domain": "sports.yahoo.com",
            "total_records": 12889
        },
        {
            "domain": "news.google.com",
            "total_records": 12883
        }
    ]
}
```
This is the response from `compare-toplist`:
```
{
    "rows": [
        {
[...]

PR #4034, opened 2024-02-13
Fix measure filter being inconsistent with certain data sets.

AdityaHegde:
## Checklist
- [x] Manual verification
- [ ] Unit test coverage
- [ ] E2E test coverage
- [x] Needs manual QA?

#### Issue addressed: 
partially addresses #3980

#### Details:
Since measure filters make a frontend query instead of a subquery in the backend we can face issues with not all values being included since there is a limit on the query result.

This PR makes this consistent by adding a sort to the toplist query used to fetch the results. The correct solution would be a backend sub query.


ericpgreen2:
This didn't seem to make any difference for my measure filter in the COVID dashboard: https://jam.dev/c/8a0b332c-616b-4a8e-9aa0-7f7e80b78df0

But I'll approve in case you think there's a marginal benefit to this new code.

ericpgreen2 (approved):

PR #4039, opened 2024-02-13
Temporarily limit pivot UI to one time pill

briangregoryholmes:
This PR puts in place a temporary UI limit of one time pill per pivot table. It does this by creating a piece of derived state called `hasTimePill`.


djbarnwal:
Closing in favor of #4043

Issue #3989, opened 2024-02-08
TypeError in Time Dimension Details crashes the app


ericpgreen2:
<img width="539" alt="image" src="https://github.com/rilldata/rill/assets/14206386/56f0981b-81e1-44e1-bf40-6d59291dfd2b">

bcolloran:
part of #3679

bcolloran:
the crash is in transposeArray:

```js
export function transposeArray(arr, rowCount, columnCount) {
  const columnarBody = [];

  for (let i = 0; i < columnCount; i++) {
    const column = [];
    for (let j = 0; j < rowCount; j++) {
      column.push(arr[j][i]);   // ------------- CRASHES HERE------------------
    }
    columnarBody.push(column);
  }

  return columnarBody;
}
```

there is a stop gap solution in https://github.com/rilldata/rill/pull/4135, but @djbarnwal will need to do type cleanups in the surrounding files to get a more robust behavior. Reassigning to @djbarnwal for a complete fix.

bcolloran:
possibly closed by #4042 and @djbarnwal has additional work that might help in https://github.com/rilldata/rill/pull/4154. Going to close this since we can't directly reproduce, can always re-open if this error code shows up in telemetry again

PR #4042, opened 2024-02-13
Type cleanups

bcolloran:
part of #3424, hopefully helps with #3989, but it's not clear what exactly is causing that error.

Note that per conversation with @djbarnwal we're updating `MetricsExplorerEntity.selectedTimezone` to
(a) be required, not optional)
(b) default to "UTC" if not specified, 
(c) use the value "UTC" in preference to "Etc/UTC" in all cases

### DIFF
--- a/web-common/src/components/data-graphic/elements/GraphicContext.svelte
+++ b/web-common/src/components/data-graphic/elements/GraphicContext.svelte
@@ -10,17 +10,17 @@ for any of its children.
   import { getContext, hasContext, onMount } from "svelte";
   import { contexts } from "../constants";
   import {
+    ScaleType,
     cascadingContextStore,
     initializeMaxMinStores,
     initializeScale,
     pruneProps,
   } from "../state";
-  import {
+  import type {
     ExtremumResolutionStore,
-    ScaleType,
     SimpleDataGraphicConfiguration,
     SimpleDataGraphicConfigurationArguments,
-  } from "../state/types";
+  } from "@rilldata/web-common/components/data-graphic/state/types";
 
   export let width: number | undefined = undefined;
   export let height: number | undefined = undefined;
@@ -35,8 +35,8 @@ for any of its children.
   export let bodyBuffer: number | undefined = undefined;
   export let marginBuffer: number | undefined = undefined;
 
-  export let xType: ScaleType = ScaleType.date;
-  export let yType: ScaleType = ScaleType.number;
+  export let xType: ScaleType = ScaleType.DATE;
+  export let yType: ScaleType = ScaleType.NUMBER;
 
   export let xMin: number | undefined | Date = undefined;
   export let xMax: number | undefined | Date = undefined;
--- a/web-common/src/components/data-graphic/elements/GraphicContext.svelte
+++ b/web-common/src/components/data-graphic/elements/GraphicContext.svelte
@@ -15,8 +15,9 @@ for any of its children.
     initializeScale,
     pruneProps,
   } from "../state";
-  import type {
+  import {
     ExtremumResolutionStore,
+    ScaleType,
     SimpleDataGraphicConfiguration,
     SimpleDataGraphicConfigurationArguments,
   } from "../state/types";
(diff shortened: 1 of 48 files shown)


web-common/src/features/dashboards/time-series/MetricsTimeSeriesCharts.svelte
```diff
@@ -70,26 +71,19 @@
...
-      renderedMeasures = $metricsView.data?.measures.filter(
-        (_, i) => $showHideMeasures.selectedItems[i],
-      );
-    }
-  }
+  $: renderedMeasures = $metricsView.data?.measures?.filter(
+    expandedMeasureName
+      ? (measure) => measure.name === expandedMeasureName
+      : (_, i) => $showHideMeasures.selectedItems[i],
+  );
```
bcolloran, line 78: consolidating into one declaration allows automatic type inference

web-common/src/components/data-graphic/elements/SimpleDataGraphic.svelte
```diff
@@ -34,7 +35,7 @@ A simple composable container for SVG-based data graphics.
   export let shareYScale = true;
 
   export let mouseoverValue: DomainCoordinates | undefined = undefined;
-  export let hovered = undefined;
+  export let hovered = false;
```
bcolloran, line 38: @djbarnwal -- I believe that hover state should always default to `false` unless explicitly set (here and in other files), rather than using undefined. Please LMK if that will cause any problems

web-common/src/components/data-graphic/marks/types.ts
```diff
@@ -1,7 +1,7 @@
 export interface Point {
   x: number;
   y: number;
-  value: string;
+  value?: string;
```
bcolloran, line 4: this is already treated as optional, there are guards against it being undefined in web-common/src/components/data-graphic/marks/MultiMetricMouseoverLabel.svelte line 204

web-common/src/components/data-graphic/state/types.d.ts
```diff
@@ -31,8 +31,8 @@ export interface SimpleDataGraphicConfigurationArguments {
   bottom: number;
   fontSize: number;
   textGap: number;
-  xType: string; // FIXME: we should have an enum here
-  yType: string; // FIXME: we should have an enum here
+  xType: ScaleType;
+  yType: ScaleType;
```
bcolloran, line 35: fixed this ancient FIXME

web-common/src/features/dashboards/time-dimension-details/TDDTable.svelte
```diff
@@ -26,8 +26,8 @@
   export let excludeMode: boolean;
   export let sortDirection: boolean;
   export let sortType: SortType;
-  export let highlightedCol: number;
-  export let scrubPos: { start: number; end: number };
+  export let highlightedCol: number | undefined;
+  export let scrubPos: { start?: number; end?: number };
```
bcolloran, line 30: already treated as optional, see check below

web-common/src/components/data-graphic/elements/GraphicContext.svelte
```diff
@@ -94,7 +95,12 @@ for any of its children.
       marginBuffer,
       id,
     }),
-  };
+    // FIXME: this cast is not provably safe,
+    // but we've been treating this as if it's true
+    // for at least 19 months, so hopefully it's safe enough.
+    // In any case, this cascading context store approach is
+    // ridiculously over engineered and needs to be rethought
+  } as SimpleDataGraphicConfigurationArguments;
```
djbarnwal, line 103: Just to add, this is most probably not a safe cast. I remember seeing missing properties at couple of places some time back. We should get back to this when we have more bandwidth to handle our tech debt

web-common/src/components/data-graphic/elements/SimpleDataGraphic.svelte
djbarnwal: Sounds about right. Shouldn't cause any issue.

djbarnwal (approved):
Thanks for the cleanups. Looks good.

web-common/src/components/data-graphic/elements/GraphicContext.svelte
bcolloran: Thanks for flagging this @djbarnwal !!! -- since you know this is unsafe for sure, I'm going to remove this cast before merging

PR #4048, opened 2024-02-14
Add manual ref for UI created alert

AdityaHegde:
Alerts need manual refs to be able to get the dashboard source. This PR adds it to the API and to the UI.


admin/server/alerts.go
```diff
@@ -444,6 +444,8 @@ func (s *Server) yamlForManagedAlert(opts *adminv1.AlertOptions, alertName, owne
 	res := alertYAML{}
 	res.Kind = "alert"
 	res.Title = opts.Title
+	// manually add the ref
+	res.Refs = []string{fmt.Sprintf("dashboard/%s", opts.DashboardName)}
```
begelundmuller, line 448: Can we put `MetricsView/%s` here instead (already supported by the parser)? There's some talk about changing "dashboard" to mean custom dashboards, so would be nice not to have to migrate the virtual alert artifacts

admin/server/alerts.go
```diff
@@ -471,6 +473,8 @@ func (s *Server) yamlForCommittedAlert(opts *adminv1.AlertOptions) ([]byte, erro
...
```
begelundmuller, line 477: Same comment

proto/rill/admin/v1/api.proto
```diff
@@ -1440,7 +1440,8 @@ message AlertOptions {
   string interval_duration = 2;
   string query_name = 3;
   string query_args_json = 4;
-  repeated string recipients = 5;
-  bool email_renotify = 6;
-  uint32 email_renotify_after_seconds = 7;
+  string dashboard_name = 5;
```
begelundmuller, line 1443: Maybe `metrics_view_name`?

begelundmuller (approved):
Approving but left some comments below

PR #4052, opened 2024-02-14
API for generating metrics views using AI

begelundmuller:
Changes:

- Adds a `GenerateMetricsViewYAML` runtime API that generates and validates a metrics view for a given table and writes it to a file.
  - By default, it will output the same metrics view definition as our current "Autogenerate dashboard"
  - If you pass `use_ai=true`, it will attempt to generate the metrics view using AI and fall back to the old template if that fails
- The AI-generated metrics views have the following properties:
  - Only uses AI to generate the title and measures. The same logic as our existing "autogenerate dashboard" is used for the other properties (such as the dimensions).
  - Removes any generated measures that do not pass validation
  - Applies a 1 minute timeout to the AI generation
  - If anything fails during generation or validation, it will log the error and fall back to our existing metrics view generator
- Adds abstractions for an "AI driver" in the runtime, which is implemented by our admin service. Other AI drivers can be added in the future (e.g. to support bringing your own OpenAI account).
- `rill start` sets up our admin service as the default AI service using the user's current admin configuration from `~/.rill` (API URL and token if logged in).

Development instructions:

- Run `rill devtool start cloud` to start the admin service locally. This automatically changes your CLI to direct admin queries to localhost instead of `admin.rilldata.com` (when you stop the devtool, it switches back).
- Run `go run ./cli start <path to project>`
- You can now generate a metrics view by calling `POST http://localhost:9009/v1/instances/default/files/generate-metrics-view` with body `{"table": "<insert>", "path": "/dashboards/name.yaml", "use_ai": true}`


admin/ai/openai.go
```diff
@@ -0,0 +1,53 @@
...
+func NewOpenAI(apiKey string) (Client, error) {
+	c := openai.NewClient(apiKey)
+
+	return &openAI{
+		client: c,
+		apiKey: apiKey,
+	}, nil
+}
+
+func (c *openAI) Complete(ctx context.Context, msgs []*adminv1.CompletionMessage) (*adminv1.CompletionMessage, error) {
```
nishantmonu51, line 27: Additional object - 
{
"completiton_id" : "UUID"
}

runtime/server/generate_metrics_view.go
```diff
@@ -0,0 +1,418 @@
...
+		doc.Measures[ie.Idx] = nil
+	}
+	for idx := 0; idx < len(doc.Measures); {
+		if doc.Measures[idx] == nil {
+			doc.Measures = slices.Delete(doc.Measures, idx, idx+1)
+		} else {
+			idx++
+		}
+	}
+
```
nishantmonu51, line 206: Emit a metric - 
{
"completiton_id" : "UUID", 
valid_metrics_count: 5, 
invalid_metrics_count: 4, 
user_id and other dimensions that we emit today..
}

PR #4053, opened 2024-02-14
fix type errors in GraphicContext

bcolloran:
fix type errors in GraphicContext. Untangles that mess a bit by:
- cleaning up `cascadingContextStore`. It's only called in one place, so it doesn't need generic params and a static argument can be pushed into the function
- making the the default overrides more explicit rather than spreading `DEFAULT` and `pruneProps(...)` objects  on top of each other, which leads to ambiguous types.


djbarnwal (changes requested):
Chart labels are displaced with these changes

image

bcolloran:
I think this approach didn't work out @djbarnwal, and I don't have enough time left to try again :-(

cleaning this up will have to wait for a future refactor -- hopefully you can just tear out the whole context store concept and replace it with something saner rather than trying to clean up the types on the context store.

PR #4059, opened 2024-02-15
Add row based virtualization to pivot table

briangregoryholmes:
This PR adds row based virtualization to the pivot table. Column virtualization will be added separately.

TanStack's virtualization library and flexRenderer are not performant enough for our use case and will need to be replaced.

### DIFF
--- a/web-common/src/features/dashboards/pivot/PivotTable.svelte
+++ b/web-common/src/features/dashboards/pivot/PivotTable.svelte
@@ -11,9 +11,17 @@
   import type { Readable } from "svelte/motion";
   import { derived } from "svelte/store";
   import type { PivotDataRow, PivotDataStore } from "./types";
+  import {
+    createVirtualizer,
+    defaultRangeExtractor,
+  } from "@tanstack/svelte-virtual";
   import ArrowDown from "@rilldata/web-common/components/icons/ArrowDown.svelte";
 
   export let pivotDataStore: PivotDataStore;
+
+  const OVERSCAN = 50;
+  const ROW_HEIGHT = 24;
+  const HEADER_HEIGHT = 30;
 
   const stateManagers = getStateManagers();
   const { dashboardStore, metricsViewName } = stateManagers;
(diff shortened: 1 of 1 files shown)


web-common/src/features/dashboards/pivot/PivotTable.svelte
```diff
@@ -12,9 +12,17 @@
...
   import type { PivotDataRow, PivotDataStore } from "./types";
+  import {
+    createVirtualizer,
+    defaultRangeExtractor,
+  } from "@tanstack/svelte-virtual";
 
   export let pivotDataStore: PivotDataStore;
 
+  const OVERSCAN = 80;
+  const ROW_HEIGHT = 24;
```
djbarnwal, line 23: 24 seems a bit too tight for row height. The mock has height as 28 for each row?

web-common/src/features/dashboards/pivot/PivotTable.svelte
```diff
@@ -12,9 +12,17 @@
...
   import { derived } from "svelte/store";
   import type { PivotDataRow, PivotDataStore } from "./types";
+  import {
+    createVirtualizer,
+    defaultRangeExtractor,
+  } from "@tanstack/svelte-virtual";
 
   export let pivotDataStore: PivotDataStore;
 
+  const OVERSCAN = 80;
```
djbarnwal, line 22: Overscan of 80 seems excessive? Probably can make it 30-40

web-common/src/features/dashboards/pivot/PivotTable.svelte
```diff
@@ -44,15 +52,38 @@
...
+
+  $: virtualRows = $virtualizer.getVirtualItems();
+  $: totalRowSize = $virtualizer.getTotalSize();
+
+  $: [before, after] = virtualRows.length
+    ? [
+        (virtualRows[1]?.start ?? virtualRows[0].start) - ROW_HEIGHT,
+        totalRowSize - virtualRows[virtualRows.length - 1].end,
+      ]
+    : [0, 0];
```
djbarnwal, line 86: Would be nice to add some description of what this does and why do we need it

web-common/src/features/dashboards/pivot/PivotTable.svelte
```diff
@@ -67,117 +98,93 @@
...
-                      </span>
-                    {/if}
-                  </button>
-                {:else}
-                  <button class="w-full h-full"></button>
-                {/if}
-              </div>
-            </th>
-          {/each}
+          <td colspan={headerGroups.length} style:height="{before}px"> </td>
```
djbarnwal, line 146: Why do we need this? Wondering if we can follow Svelte virtual's example for sticky rows as well?
https://github.com/TanStack/virtual/blob/main/examples/svelte/sticky/src/App.svelte

djbarnwal (commented):
Have added some questions

web-common/src/features/dashboards/pivot/PivotTable.svelte
briangregoryholmes: Where are you seeing 28? The cells in Figma are 24.

web-common/src/features/dashboards/pivot/PivotTable.svelte
briangregoryholmes: I've spent a good bit of time tinkering with this value. There is no downside to setting it at 80. Any browser is more than capable of stably rendering ~200 rows. What they can't do, within the context of TanStack's virtualization paradigm, is quickly mount/unmount/rerender huge chunks rows, which is what happens when you have a lower overscan. Virtualization, frankly, is not needed on views that have less than 2000 total cells, so this is a way of bypassing virtualization for those smaller tables. Setting it at 80 means that you get very little flashing even when scrolling quickly. This will vary from browser to browser, of course, but there's a rough example in the screenshots below. Additionally, setting it higher means that your cell widths are being derived from a large set of values, which will cause less column size snapping in the short term.

Scrolling with OVERSCAN at 30:
Screenshot 2024-02-16 at 9 07 51 AM

Scrolling with OVERSCAN at 80:
Screenshot 2024-02-16 at 9 07 27 AM

web-common/src/features/dashboards/pivot/PivotTable.svelte
briangregoryholmes: Added

web-common/src/features/dashboards/pivot/PivotTable.svelte
briangregoryholmes: The rendering approach in this example uses absolute positioning of rows, which is not performant enough for our use case and does not work with well with rendering table elements as opposed to divs.

See: https://github.com/TanStack/virtual/discussions/476

See: https://github.com/TanStack/virtual/issues/585#issuecomment-1716247313

See: https://github.com/TanStack/virtual/issues/656

briangregoryholmes:
@djbarnwal Responded to your questions and updated the comments.

web-common/src/features/dashboards/pivot/PivotTable.svelte
djbarnwal: Oh, you are right!

djbarnwal (approved):

PR #4066, opened 2024-02-16
Adding offset support to available_time_ranges

### LINKED ISSUE #4004: Feat: Custom time ranges should support offsets
**Problem**
Today when defining custom ranges you can define ISO durations that uses Max(timestamp)/Today as the end period.
We should introduce the ability to define a custom end period/offset so that users can craft custom ranges such as 

- Yesterday
- Previous Month
- Previous Quarter
- Previous Year

AC:

- A user can define custom periods like the ones above
- A user can specify comparison offsets that inherits the offset from the time range
- Timerange: Previous Year Comparison Period: 2 years ago

AdityaHegde:
## Checklist
- [ ] Manual verification
- [ ] Unit test coverage
- [ ] E2E test coverage
- [ ] Needs manual QA?

#### Issue addressed: 
closes #4004


AdityaHegde:
Closing in favour of #4096 . We will revisit the way we configure arbitrary time range with offsets in the future.

PR #4067, opened 2024-02-16
Fix: Add label for empty string in column header

djbarnwal:
If one of the values of a dimension is "" (empty string), the app would crash. This PR adds a label `<empty string>` for such cases.

### DIFF
--- a/web-common/src/features/dashboards/pivot/PivotExpandableCell.svelte
+++ b/web-common/src/features/dashboards/pivot/PivotExpandableCell.svelte
@@ -26,7 +26,7 @@
   {#if value === "LOADING_CELL"}
     {""}
   {:else if value === ""}
-    {"<empty string>"}
+    {"\u00A0"}
   {:else}
     {value}
   {/if}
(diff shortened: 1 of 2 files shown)


web-common/src/features/dashboards/pivot/PivotExpandableCell.svelte
```diff
@@ -22,7 +22,14 @@
       </div>
     </button>
   {/if}
-  {value === "LOADING_CELL" ? "" : value}
+
+  {#if value === "LOADING_CELL"}
+    {""}
+  {:else if value === ""}
+    {"<empty string>"}
```
pjain1, line 29: is it not possible to display just blank instead of printing this ?

djbarnwal: Displaying blank string messes up the styling and also is not supported by the table framework we use. `null` is already a reserve keyword so went with `<empty string>` which is adopted by some browser and tools when you print an empty string.

djbarnwal: @mindspank thoughts?

pjain1: may be I can merge for now and product can decide correct semantics here

djbarnwal: Switched to NBSP unicode character after discussing with Alex

pjain1 (approved):

PR #4072, opened 2024-02-16
Send all relevant measures to dimension display query

AdityaHegde:
Not a 100% sure when this broke, but if we have a measure hidden but it is selected as the leaderboard measure dimension display breaks. This is because we call sort on the selected leaderboard measure without actually selecting it in the query.

Sending all relevant measures to the query so that the app doesnt crash.


djbarnwal (approved):
Tested locally, works fine. 
Approving so that you can merge post lint fixes

PR #4076, opened 2024-02-16
Update pivot sidebar layout mechanism

briangregoryholmes:
This PR updates the sidebar layout mechanism to display as many pills as possible while still maintaining the visibility of all three sections.

### DIFF
--- a/web-common/src/features/dashboards/pivot/PivotDrag.svelte
+++ b/web-common/src/features/dashboards/pivot/PivotDrag.svelte
@@ -9,22 +9,12 @@
   export let items: PivotChipData[];
   export let collapsed = false;
 
-  let showMore = false;
-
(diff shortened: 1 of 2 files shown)


djbarnwal (commented):
Wondering if we can just use the required space when we have less measures.
 
image

With the current approach I see a lot of whitespace. In most cases dimensions will have the most number of chips compared to Time and Measure so this would be pretty common.

briangregoryholmes:
@djbarnwal Alright, unfortunately, only Safari seems to support setting a `flex-basis` in conjunction with `max-height: fit-content`, so I just moved this conditional clamping to a JS implementation. Works as intended across all browsers now.

djbarnwal (approved):

PR #4078, opened 2024-02-17
Updating to Docusaurus v3

AndrewRTsao:
We are on an older version and should upgrade to Docusaurus v3. Followed the steps described here:
https://docusaurus.io/docs/migration/v3
https://docusaurus.io/blog/preparing-your-site-for-docusaurus-v3

This is a major upgrade but we should pull the trigger now to be on the latest and greatest prior to our upcoming docs refactor (and GA launch). Notably, from using `docusaurus-mdx-checker`, there weren't that many breaking changes that needed to be fixed prior to the migration. There was 1 `<` that was flagged but it just needed to be commented out because we were essentially giving an example and the <example> was detected as a bad use of a tag. 

Additionally, moving to Docusaurus v3 implies upgrade Node, React, TS, and quite a few other dependencies. I ended up having to delete `package-lock.json` to debug but the major dependency issue seemed to be around redocusaurus, which only recently added support for Docusaurus 3.0 and required setting `"redoucsaurus": "^2.0.0"`. Would definitely appreciate a thorough review to see if we can get around recreating `package-lock.json`, cleaning `node_modules`, running `npm run clean`, and clearing the npm cache. Not sure also if this affects some of our downstream automations.

### DIFF
--- a/docs/docs/deploy/credentials/athena.md
+++ b/docs/docs/deploy/credentials/athena.md
@@ -29,7 +29,7 @@ If you do not have the AWS CLI installed and authenticated, follow these steps:
 
 3. If your organization does not have SSO configured:
 
-    a. Follow the steps described under [How to create an AWS service account using the AWS Management Console](#how-to-create-an-aws-service-account-using-the-aws-management-console), which you will find below on this page.
+    a. Follow the steps described in [How to create an AWS service account using the AWS Management Console](./s3.md#how-to-create-an-aws-service-account-using-the-aws-management-console).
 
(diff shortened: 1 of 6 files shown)


AndrewRTsao:
The changes to athena.md is because it seems like Docusaurus will now check for broken / unresolved soft links and flag them before you can successfully run a local build. One of the benefits of moving to v3!

himadrisingh (approved):
lgtm

mindspank (approved):
Functionality wise LGTM. Haven't spotted any differences in the look-n-feel of docs after upgrade.

PR #4082, opened 2024-02-17
Fix invalid fetching state

djbarnwal:
For cases where we have column dimensions and row dimension without any measures, the table view shows a fetching spinner. We should instead show the empty state for now


mindspank (approved):
lgtm

PR #4083, opened 2024-02-17
Format month, quarter and week headers

djbarnwal:
Add quarter number for quarters, week number for week and mention full year for month.


mindspank (approved):
LGTM

PR #4096, opened 2024-02-20
Add UI support for previous complete presets

### LINKED ISSUE #4004: Feat: Custom time ranges should support offsets

AdityaHegde:
## Checklist
- [x] Manual verification
- [ ] Unit test coverage
- [ ] E2E test coverage
- [x] Needs manual QA?

#### Issue addressed: 
closes #4004 , #4120

#### Details:
This PR adds time range presets for 
- Yesterday (rill-PDC)
- Previous Week Complete (rill-PWC)
- Previous Month Complete (rill-PMC)
- Previous Quarter Complete (rill-PQC)
- Previous Year Complete (rill-PYC)

We are adding just the presets for now and not the support for supplying arbitrary time range with offsets.

- [x] Add config in the UI for time ranges.
- [x] Add validation support in backend.
- [x] Add config in the UI for comparison time ranges.
- [x] Add correct settings for reports/alerts.


djbarnwal (changes requested):
Seeing some discrepancy with the transformations applied 
Below is the range for Last 7 days - 
image

Previous month complete should show data for Nov and not Oct
image

Similarly, for Previous year complete, the data should be for the year 2020

image

Can see similar issues with the other time presets as well.

web-common/src/features/dashboards/time-controls/time-range-store.ts
```diff
@@ -91,6 +100,12 @@ export function timeRangeSelectionsSelector([
   } else {
     latestWindowTimeRanges = LATEST_WINDOW_TIME_RANGES;
     periodToDateRanges = PERIOD_TO_DATE_RANGES;
+    previousCompleteDateRanges = PREVIOUS_COMPLETE_DATE_RANGES;
+    hasDefaultInRanges =
```
AdityaHegde, line 104: This should fix #4120

djbarnwal (approved):
Looks code from application perspective. Seeing some discrepancy in previous period time comparison for some cases (for ex. Previous quarter complete only shows two points of comparison) but that can be addressed as a follow up.

PR #4117, opened 2024-02-21
Alert preview table updates

AdityaHegde:
Makes small adjustment to alert preview table.
- [x] Refactor to remove too much code from component.
- [x] Pass in split by time grain.
- [x] Pass in measure filters.
- [x] Add a throttle to refetches.
- [x] Add a loading spinner


web-common/src/features/alerts/alert-preview-data.ts
```diff
@@ -0,0 +1,135 @@
...
+    having: sanitiseExpression(undefined, params.criteria),
+    timeRange: {
+      start:
+        params.splitByTimeGrain && timeControls.timeEnd
+          ? getOffset(
+              new Date(timeControls.timeEnd),
+              params.splitByTimeGrain,
+              TimeOffsetType.SUBTRACT,
+            )
+          : timeControls.timeStart,
```
ericpgreen2, line 103: Is this offset actually needed? Adding the time + time grain to the `dimensions` would make this unnecessary, right?

ericpgreen2 (changes requested):

web-common/src/features/alerts/alert-preview-data.ts
```diff
@@ -0,0 +1,135 @@
...
+      end: timeControls.timeEnd,
+    },
+  };
+}
+
+function getAlertPreviewQueryOptions(
+  ctx: StateManagers,
+  params: AlertPreviewParams,
+  timeControls: TimeControlState,
+  resolvedMeasureFilters: ResolvedMeasureFilter,
```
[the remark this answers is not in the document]
AdityaHegde: True. But this is supposed to disable the query when timeControls and measureFilters are being computed asynchronously. Sadly ctx doesnt have that info.

web-common/src/features/alerts/alert-preview-data.ts
AdityaHegde: No was getting some error with pivot query. So had this for just the last time segment

web-common/src/features/alerts/alert-preview-data.ts
```diff
@@ -0,0 +1,167 @@
...
+> {
+  const {
+    selectors: {
+      measureFilters: { getResolvedFilterForMeasureFilters },
+      timeRangeSelectors: { timeControlsState },
+    },
+  } = ctx;
+  const timeControls = get(timeControlsState);
+  // 1st get is to fetch the selector, 2nd get is to get async data.
+  const resolvedMeasureFilters = get(get(getResolvedFilterForMeasureFilters));
```
ericpgreen2, line 147: Just a note we won't be able to use this in the Edit Alert dialog where there's no dashboard (and accompanying state) on the page, but let's refactor in a follow-on.

ericpgreen2 (approved):
Nice, it's working well!

PR #4129, opened 2024-02-21
Prefil alert options from the underlying dashboard

AdityaHegde:
Options like selected measure and dimension can be pre-filled from the dashboard settings.
- [x] Default the measure to the selected leaderboard measure.
- [x] Default the dimension to the active dimension.
- [x] Restrict the measure in the criteria to the selected measure.


ericpgreen2 (approved):
Nice 👍

Issue #3991, opened 2024-02-08
TypeError in line chart mouseover crashes the app


ericpgreen2:
<img width="551" alt="image" src="https://github.com/rilldata/rill/assets/14206386/2351a885-0ac9-4a32-b5c9-662700e8ad60">

bcolloran:
part of #3679

bcolloran:
related to https://github.com/rilldata/rill/issues/3424

PR #4133, opened 2024-02-21
prevent indexing into undefined value

bcolloran:
this should fix https://github.com/rilldata/rill/issues/3991 -- the function `bisectData` can _explicitly_ return undefined--

```
export function bisectData(
  value: Date,
  direction: "left" | "right" | "center",
  accessor: string,
  data: ArrayLike<unknown>,
  returnPos = false,
) {
  if (!data?.length) return;
  const bisect = bisector((d) => d[accessor])[direction];

  if (returnPos) return value !== undefined ? bisect(data, value) : undefined;
  return value !== undefined ? data[bisect(data, value)] : undefined;
}
```
-- but that case is not handled.

This PR should prevent the TypeError in #3991 (which will cause an app crash), but it doesn't fix the underlying typing problems. This may cause weird behavior in the client, but that's probably better than a crash.

My recommendation would be: merge this PR (with any revisions suggested by @djbarnwal) to fix the crash, but then probably @djbarnwal will need to dig into this to get a full fix with double checks for desired behavior-- there is a huge amount of untyped code in utils.ts that is beyond my understanding, and it feels dicey for me to do a deep dive there with only a few days left.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8474,6 +8474,12 @@
         "@types/node": "*"
       }
     },
+    "node_modules/@types/d3-array": {
+      "version": "3.2.1",
+      "resolved": "https://registry.npmjs.org/@types/d3-array/-/d3-array-3.2.1.tgz",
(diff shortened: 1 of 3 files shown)


djbarnwal (approved):

djbarnwal:
The whole data graphics directory needs to be strongly typed. We should pick a week where we can prioritize this tech debt. @nishantmonu51

PR #4135, opened 2024-02-21
prevent crash from 3989

bcolloran:
fixes the crash in 3989, but does not fix the underlying code


djbarnwal:
Merged this branch to #4154. Closing in favor of that.

PR #4140, opened 2024-02-22
Cancel alert confirmation

AdityaHegde:
Cancelling an alert creation is instant right now. We need to make sure the user doesnt lose progress we need a confirmation.

Note that multiple dialogs is not supported right now. Was not sure if shadcn's variant supports this. So just disabling the main dialog while showing the cancel dialog.


ericpgreen2:
This PR should be coupled with the "X" button in the top-right of the dialog. Because the confirmation prompt is much more valuable if you're quitting from further along in the flow.

image

ericpgreen2:
I also think it feels a little funky without the overlapping dialogs. If it requires moving to ShadCN's dialog, let's consider this future work until we've finished the rest of the items on the burndown list.

ericpgreen2:
Janet and I discussed we should use this copy:

image

ericpgreen2:
Let's consider this future work and close this PR for now?

AdityaHegde:
Closing this for now. We need shadcn dialogs

PR #4144, opened 2024-02-22
Update buttons to match Figma, add Storybook story to simplify QA

### LINKED ISSUE #3909: Update brand colors on Rill Cloud pages
**Login/Signup Pages**

https://www.figma.com/file/Qt6EyotCBS3V6O31jVhMQ7/RILL?type=design&node-id=5131-162648&mode=design&t=2SlPTPxwRkk2Srpl-11

**Link/Text Buttons**

https://www.figma.com/file/nqqazRo1ckU9ooC9ym9weI/Rill-Design-System?type=design&node-id=1156-254709&mode=design&t=Scidzo2gfdWvh1ax-11

Screenshot 2024-01-24 at 2 02 06 PM

**Brand buttons**

https://www.figma.com/file/nqqazRo1ckU9ooC9ym9weI/Rill-Design-System?type=design&node-id=1155-311775&mode=design&t=Scidzo2gfdWvh1ax-11

bcolloran:
fixes #3909


bcolloran:
A parting gift for @jkhwu ;-)  which makes good on some work I'd started earlier, this PR does the following:

- update the buttons to match [the current state of the design system in Figma](https://www.figma.com/file/nqqazRo1ckU9ooC9ym9weI/Rill-Design-System?type=design&node-id=74-246&mode=design&t=Uw4vfgNKotpEJRrK-0)
- start migrating the <Button> component usage semantics match those of the design system, and add warnings for incorrect usage in the meantime
- add storybook story providing an overview of variants that is displayed in the same layout as Figma, in order to make updating and QAing easier (and this bit is a gift to the application team as well as Janet :-) )
- update postcss configs so that we can use the nice nested sass syntax to organize things


To see the storybook  story, run `cd web-common && npm run storybook` from your rill folder, then go to http://localhost:6006/?path=/story/button-stories--button-table-like-figma

Screenshots of stories shown below. Note that there are some combinations of button attributes in the storybook tables below that are not included in the design system -- console warnings are shown if those invalid combos are used in code, and the fallbacks shown will be used.

image

image

image

briangregoryholmes:
This is really helpful overall, though I am seeing some reversions here and there.

The `shape` prop seems not to work because of the way sizing has been changed (comes after the shape classes in the styles), but maybe you have a different intention here. See the Button in the PivotToolbar, which isn't square even after updating to the new prop. I would also just directly put the `shape` prop on the class, rather than doing the conditional in line 59/60.

Not sure what the behavior of the `loading` prop is, exactly, but it doesn't seem to be working correctly. Styles seem okay if I just change the default to `false`. Without that, the add buttons in Pivot are the primary color and the Deploy/Dashboard buttons are the wrong color and have no hover star.

If you could leave the `add` class in there for now, I can make a wrapper component later, but I think we'll need an `icon` type or a `zeroPadding` prop.

I think, in general, the console warnings are not necessary. We're the only ones interfacing with this and I don't believe they would be caught/stripped out in CI or the build process. I would rather those conditionals get built into the selectors.

Could get rid of the conditional for dashed and just add `.dashed` to the secondary styles.

briangregoryholmes (changes requested):
This is really helpful overall, though I am seeing some reversions here and there.

The `shape` prop seems not to work because of the way sizing has been changed (comes after the shape classes in the styles), but maybe you have a different intention here. See the Button in the PivotToolbar, which isn't square even after updating to the new prop. I would also just directly put the shape prop on the class, rather than doing the conditional in line 59/60.

Not sure what the behavior of the `loading` prop is, exactly, but it doesn't seem to be working correctly. Styles seem okay if I just change the default to false. Without that, the add buttons in Pivot are the primary color and the Deploy/Dashboard buttons are the wrong color and have no hover state.

If you could leave the `add` class in there for now, I can make a wrapper component later, but I think we'll need an `icon` type that can maybe take the icon component as a prop or a `padding` prop and a `rounded` prop.

I think, in general, the console warnings are not necessary. We're the only ones interfacing with this and I don't believe they would be caught/stripped out in CI or the build process. I would rather those conditionals get built into the selectors.

Could get rid of the conditional for dashed and just add `.dashed` to the secondary styles.

bcolloran:
Thanks for the review @briangregoryholmes, this is ready for another look


> This is really helpful overall, though I am seeing some reversions here and there.

Glad this is helpful, sorry for any reversions! I'm not shocked, b/c I know the design system I was working off of is just one of several things in flight. Hopefully having the storybook will be useful in filling in those other things as well.

Also, if I'm not able to get this over the line in the next two days, obvi please take whatever is useful here and push it in whatever direction you prefer.

> The `shape` prop seems not to work because of the way sizing has been changed (comes after the shape classes in the styles), but maybe you have a different intention here. See the Button in the PivotToolbar, which isn't square even after updating to the new prop. 

nope that was just an oversight, fixed

> would also just directly put the shape prop on the class, rather than doing the conditional in line 59/60.

np, done!

> Not sure what the behavior of the `loading` prop is, exactly, but it doesn't seem to be working correctly. Styles seem okay if I just change the default to false. Without that, the add buttons in Pivot are the primary color and the Deploy/Dashboard buttons are the wrong color and have no hover state.

`loading` is just laying the framework for the ability to match the "status" button kind from the design system, but it's not wired up here. Feel free to take that in another direction if you wish, just wanted to have the styling pieces in place for when the rest implementation is ready to be prioritized. 

But importantly: yes, that prop should have defaulted to false!!

> If you could leave the `add` class in there for now, I can make a wrapper component later, but I think we'll need an `icon` type that can maybe take the icon component as a prop or a `padding` prop and a `rounded` prop.

`add` is still in there (`ButtonKind="add"), I didn't take it out b/c I noticed it had been put in recently. Did it regress? I tried to get the styles to match, if it doesn't look right, can you point me at the figma for that? (Or I can plug in the styles if you know what they should be)

`icon` is out of scope for this PR i think

> I think, in general, the console warnings are not necessary. We're the only ones interfacing with this and I don't believe they would be caught/stripped out in CI or the build process. I would rather those conditionals get built into the selectors.

took out the warnings.

> Could get rid of the conditional for dashed and just add `.dashed` to the secondary styles.

I looked at doing this with an `@extends .secondary`, but our css preprocessor isn't set up for that ATM, so the conditional was the most expedient way to do this, and I didn't want to spend a ton of time there with just a couple days left. As with all above: hopefully this will be a useful starting point, please build off it in whatever direction you prefer! :-)

bcolloran:
@briangregoryholmes and @ericpgreen2 -- I'm not going to have time to address any additional rounds of feedback on this PR. Reassigning to @ericpgreen2 to track/reassign/finish this

ericpgreen2:
Temporarily closing this PR, but we'll pick it back up next sprint or later this quarter.

PR #4146, opened 2024-02-22
Fix case sensitivity bug in dimension/measure names validation

begelundmuller:


AdityaHegde (approved):

AdityaHegde:
Merging since the test failure is unrelated.

PR #4154, opened 2024-02-23
Stricter types for Time dimension detail table

djbarnwal:
This PR adds stricter types for Time dimension detail table. There are still some type warnings but we need better runtime support for some types.
For ex-  values in `V1MetricsViewAggregationResponseDataItem` should not be typed as `any`


web-common/src/features/dashboards/time-dimension-details/time-dimension-data-store.ts
```diff
@@ -247,20 +269,31 @@ function prepareTimeData(
       tableData?.map((v) => {
         if (v[measureName] === null && v[`comparison.${measureName}`] === null)
           return null;
-        return formatter(v[measureName] + v[`comparison.${measureName}`]);
+
+        const total =
+          ((v[measureName] as MeasureValue) || 0) +
+          ((v[`comparison.${measureName}`] as MeasureValue) || 0);
```
bcolloran, line 275: All of this casting as MeasureValue is a code smell -- it indicates something is wrong here. It looks to me like the type `TimeSeriesDatum` may be defined too expansively. Here are some ideas of things to look into:
- can we remove the possibility that `V1TimeSeriesValueRecords` is ever present in the `[key:string]` union type? I don't see `V1TimeSeriesValueRecords` being used basically anywhere, is it actually used? if so, that type should be called out where it is used.
- instead of casting, you could have a wrapper fn that does type narrowing, perhaps [using the `is` operator](V1TimeSeriesValueRecords) or [the `typeof` operator](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#typeof-type-guards).
- You could try a generic type for `TimeSeriesDatum` and only allow the possibility of `V1TimeSeriesValueRecords` when explictly needed something like (this is not exactly right)
```
export interface TimeSeriesDatum<T extends V1TimeSeriesValueRecords|undefined = undefined> extends V1TimeSeriesValue {
  ts_comparison?: Date;
  ts_position?: Date;
  [key: string]: string | number | undefined | T;
}
```


But the use of all this casting shows that something is wrong, and that we should try to clean up the upstream types a bit.

bcolloran (changes requested):
Nice work @djbarnwal. My main comment overall is: be careful of casting. If you find yourself doing it very much it means something is wrong and you should look into the upstream types you are working with and wrap them so that the need to cast doesn't flow into your downstream code.

For us, our API types are annoying limited, which means types don't flow through the system as easily as we'd like, but rather than casting at the endpoint, you should use type guards and wrappers as close to the data ingestion point as possible.

bcolloran (approved):
Nice!

PR #4160, opened 2024-02-26
[Alert Create/Edit] Show the dashboard time range and send it in queryArgs

AdityaHegde:
Updates the way we handle time range.
- [x] Show the time range selected in the dashboard in the filters section.
- [x] Use the time range instead of split by grain in creating the alert.
- [x] Update the measure filter to not show `for` if dimension is not specified.

### DIFF
--- a/web-common/src/features/alerts/data-tab/AlertDialogDataTab.svelte
+++ b/web-common/src/features/alerts/data-tab/AlertDialogDataTab.svelte
@@ -31,7 +31,8 @@
       label: d.label?.length ? d.label : d.expression,
     })) ?? [];
 
-  $: hasAtLeastOneFilter = $form.whereFilter.cond.exprs.length > 0;
+  $: hasAtLeastOneFilter =
+    $form.whereFilter.cond.exprs.length > 0 || $form["timeRange"];
 </script>
 
 <div class="flex flex-col gap-y-3">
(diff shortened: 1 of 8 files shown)


web-common/src/features/dashboards/filters/ReadonlyTimeRange.svelte
ericpgreen2, line 1: Capitalization nit on the file name: "ReadOnlyTimeRange"

web-common/src/features/alerts/data-tab/AlertDialogDataTab.svelte
```diff
@@ -31,7 +31,8 @@
       label: d.label?.length ? d.label : d.expression,
     })) ?? [];
 
-  $: hasAtLeastOneFilter = $form.whereFilter.cond.exprs.length > 0;
+  $: hasAtLeastOneFilter =
+    $form.whereFilter.cond.exprs.length > 0 || $form["timeRange"];
```
ericpgreen2, line 35: We should probably get rid of this now that the time range filter pill will always be present

ericpgreen2 (commented):
Left some comments. Also, check out the full discussion in Slack: https://rilldata.slack.com/archives/C06638A72TH/p1709254991171789

AdityaHegde:
Closing this since the design has changed significantly and my understanding is obsolete.

PR #4162, opened 2024-02-26
Fix multiple updates of pivot state leading to cancelled requests

AdityaHegde:
The top level pivot config was depending on dashboard store 3 times. One directly, another through time controls and last one through measure filter resolutions.

This PR updates it to be just once. Time controls and measure filter resolutions are computed directly from the resolved dashboard state.


djbarnwal (approved):
Tested on druid and duckdb dashboards, works well. Don't see cancelled queries while building the pivot anymore.

PR #4167, opened 2024-02-26
Update Bookmarks API to support new designs

AdityaHegde:
There has been significant design changes to bookmarks. We need to update the APIs to make sure all those features can be added. Ref: [notion](https://www.notion.so/rilldata/Bookmarks-1766199e3f074010bfe5e3e97922a31a#385014d5263740aeac870735dc06dcab)

- [x] Add global and user specific bookmark support.
- [ ] Update the list bookmarks to get user and global bookmarks.
- [ ] Testing the APIs


admin/database/database.go
```diff
@@ -179,9 +179,10 @@ type DB interface {
 	DeleteProjectInvite(ctx context.Context, id string) error
 	UpdateProjectInviteRole(ctx context.Context, id, roleID string) error
 
-	FindBookmarks(ctx context.Context, projectID, userID string) ([]*Bookmark, error)
+	FindBookmarks(ctx context.Context, projectID, dashboardName, userID string) ([]*Bookmark, error)
```
begelundmuller, line 182: Maybe now that we have the resource kind/name terminology, we should generalize this to `resourceKind, resourceName string`?

admin/database/database.go
```diff
@@ -649,21 +650,34 @@ type InsertProjectInviteOptions struct {
 type Bookmark struct {
 	ID            string
 	DisplayName   string    `db:"display_name"`
+	Description   string    `db:"description"`
 	Data          []byte    `db:"data"`
 	DashboardName string    `db:"dashboard_name"`
 	ProjectID     string    `db:"project_id"`
 	UserID        string    `db:"user_id"`
+	IsGlobal      bool      `db:"is_global"`
```
begelundmuller, line 658: Would suggest `Shared` instead of `IsGlobal` since "global" implies everybody has access to it, whereas it's actually just the people with access to the resource who gets access to it

admin/server/bookmarks.go
```diff
@@ -96,16 +96,22 @@ func (s *Server) CreateBookmark(ctx context.Context, req *adminv1.CreateBookmark
 		permissions.ReadProd = true
 	}
 
+	if !permissions.ManageProject {
+		req.IsGlobal = false // only users that can manage the project can
+	}
```
begelundmuller, line 101: Better to return an error when permissions are breached than have (maybe unexpected) side effects.

Like, if a reader role sets `is_global=true`, it probably implies a missing check on the frontend

begelundmuller (changes requested):

admin/database/postgres/migrations/0021.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE bookmarks RENAME COLUMN dashboard_name TO resource_name;
+ALTER TABLE bookmarks ADD COLUMN resource_kind TEXT DEFAULT 'MetricsView';
```
begelundmuller, line 2: Should this be `rill.runtime.v1.MetricsView`?

AdityaHegde: Yes!

begelundmuller (approved):

Issue #4163, opened 2024-02-26
Unable to move chips around in Pivot Header drag list


djbarnwal:
I am unable to shuffle or move chips from Rows to columns and vice versa in the pivot header. They are getting locked into their positions once they are dropped into the header drag list.

ericpgreen2:
Closed by #4207

PR #4168, opened 2024-02-26
Fix: Chip sorting in pivot header

briangregoryholmes:
This PR fixes an issues where the pivot field chips in the `PivotHeader` component were not sortable.

Closes:
#4163

### DIFF
--- a/web-common/src/features/dashboards/pivot/DragList.svelte
+++ b/web-common/src/features/dashboards/pivot/DragList.svelte
@@ -73,7 +73,7 @@
   {#each items as item (item.id)}
     <button
       class="item"
-      title={item.type}
+      title={item.title}
       data-type={item.type}
       animate:flip={{ duration: flipDurationMs }}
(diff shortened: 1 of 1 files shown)


web-common/src/features/dashboards/pivot/DragList.svelte
```diff
@@ -34,13 +37,33 @@
...
+    window.removeEventListener("mouseup", onMouseUp);
+  }
+
+  function handleMouseDown(
+    e: MouseEvent & {
+      currentTarget: EventTarget & HTMLButtonElement;
+    },
+  ) {
+    const type = e.currentTarget.dataset.type as PivotChipType;
+    console.log(type);
```
djbarnwal, line 52: Stale console log here

djbarnwal (commented):
Seems to work but the drag feels very jittery. Is that something you can reproduce as well?

briangregoryholmes:
Closing in favor of #4207

PR #4172, opened 2024-02-26
update tsc checks to filter by file and TS error code

bcolloran:
Updating the whitelist script to check by file _and_ TS code. This will prevent newTS errors from entering the codebase unless there is already an error with the same error code in the updated file. The whitelist can be gradually tightened as the codebase improves.


ericpgreen2 (approved):
Awesome!

PR #4192, opened 2024-02-27
fix Dashboard: number discrepancy between dimension table context column and its tooltip (#4143)

### LINKED ISSUE #4143: Dashboard: number discrepancy between dimension table context column and its tooltip
Hover over the context column in [this dashboard](https://ui.rilldata.com/ericpgreen/rill-test-covid/covid?state=CgUKA1A0VxgFIgAqCW1lYXN1cmVfMDIIbG9jYXRpb244AUgBWAFgAmoTQW1lcmljYS9Mb3NfQW5nZWxlc3gCgAEBmAH%252F%252F%252F%252F%252F%252F%252F%252F%252F%252F%252F8BogEEGgIICPgBAA%253D%253D) to see the faulty behavior.

Both in the table and in the tooltip, we should be rounding numbers up not down.

[Reported in Slack](https://rilldata.slack.com/archives/CTZ8XBQ85/p1708544911033699)

image

bcolloran:
Fixes #4143 

before
image

after
image

@ericpgreen2, I was starting to work through a more comprehensive fix, but I was getting bogged down. This is kind of a nasty hack, but I thought it would be better to get a fix in than to spend a bunch of time trying to untangle a good fix if I might risk running out of time.


ericpgreen2 (approved):
Agree this code/component is a bit confusing. Works as a patch fix 👍

PR #4207, opened 2024-02-29
Refactor: Pivot drag and drop

briangregoryholmes:
This PR removes the `svelte-dnd-action` dependency and implements bespoke drag and drop functionality for our use case, solving some functionality and performance issues.

Optimization ongoing, but should be ready for merge.


djbarnwal (commented):
Looks good functionality wise. Some broader thoughts -
- Dragging measures into rows from columns should not remove the measure. This would be consistent with what we already have now
- I see a lot of native JS DOM manipulations. Is there a way to avoid/abstract this and do it in a way which is closer to svelte? We already have a `portal` component. Is that something that can be used here for the dragged item?

briangregoryholmes:
@djbarnwal I have some issues with the existing Portal component and would love to factor it out eventually in favor of an action. I've added an example of that to this implementation and refactored the code accordingly. I've also restructured some things to address your note.

Unclear about your first comment, however. Measures can't be dragged into rows. Do you mean dimensions? It was my understanding that this should be possible. Is there a technical reason not to allow it?

djbarnwal:
@briangregoryholmes Sounds good, do you prefer to merge the Portal PR first followed by this?

For the second point, previously if you would drag a measure from column to a row, no change would happen. The measure would stay in columns zone. Now if you do it, the measure is removed from the Pivot table altogether. Probably not a big issue now that I think of it.

briangregoryholmes:
@djbarnwal Ah, I see what you mean. Easy enough to add that back. I'd like to get #4215 and #4216 merged first.

@mindspank Can you take a look at what we're referring to. The question is really about whether dragging a measure pill out of columns and dropping it on the rows zone (which isn't valid) should remove the pill from measures (as it would if you dragged onto the table, for instance) or be treated as a "mistake" and keep the pill in columns.

mindspank:
@briangregoryholmes I think removing the pill makes sense

djbarnwal (approved):

PR #4209, opened 2024-02-29
Runtime: Custom APIs Yaml parser

k-anshul:


proto/rill/runtime/v1/resources.proto
```diff
@@ -418,3 +419,29 @@ message ExecutionError {
...
+// MetricSQL is used for querying against a metrics view
+// If sql is set, it is parsed and executed as metrics SQL
+// Alternatively, you can also pass parameters similar to MetricsViewAggregation
+message MetricSQL {
+  string sql = 1;
+  string metrics_view = 2;
+  repeated string measures = 3;
+  string where = 4;
+  string limit = 5;
+}
```
begelundmuller, line 445: I think we can avoid putting the specific resolver properties here – it should be sufficient to parse them in the resolver implementation. With reference to the `APIResolverOptions` proposal in the design doc, I would consider something like this:
```proto3
message APISpec {
  string resolver = 1;
  google.protobuf.Struct resolver_properties = 2;
}
```

runtime/api.go
```diff
@@ -0,0 +1,38 @@
...
+
+	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
+	"github.com/rilldata/rill/runtime/drivers"
+)
+
+var ErrAPINotFound = fmt.Errorf("API not found")
+
+var BuiltinAPIs = map[string]*runtimev1.API{}
+
+func (r *Runtime) APIForName(ctx context.Context, instanceID, name string, reqParams map[string]any) (*runtimev1.API, error) {
```
begelundmuller, line 16: `reqParams` aren't needed here – only at resolve time

runtime/api.go
```diff
@@ -0,0 +1,38 @@
...
+	ctrl, err := r.Controller(ctx, instanceID)
+	if err != nil {
+		return nil, err
+	}
+
+	resource, err := ctrl.Get(ctx, &runtimev1.ResourceName{Name: name, Kind: ResourceKindAPI}, false)
+	if err != nil {
+		if errors.Is(err, drivers.ErrResourceNotFound) {
+			return nil, ErrAPINotFound
+		}
```
begelundmuller, line 29: I think we can just return the error outright and remove the `ErrAPINotFound` value. The caller can just check for `drivers.ErrResourceNotFound` directly if they need to handle a not-found case (or to produce a nicer error message).

runtime/api.go
```diff
@@ -0,0 +1,38 @@
...
+func (r *Runtime) APIForName(ctx context.Context, instanceID, name string, reqParams map[string]any) (*runtimev1.API, error) {
+	if api, ok := BuiltinAPIs[name]; ok {
+		return api, nil
+	}
+	ctrl, err := r.Controller(ctx, instanceID)
+	if err != nil {
+		return nil, err
+	}
+
+	resource, err := ctrl.Get(ctx, &runtimev1.ResourceName{Name: name, Kind: ResourceKindAPI}, false)
```
begelundmuller, line 25: nit: `Kind` comes before `Name` in the struct

runtime/api.go
```diff
@@ -0,0 +1,38 @@
...
+			return nil, ErrAPINotFound
+		}
+		return nil, err
+	}
+
+	api := resource.GetApi()
+	if api == nil {
+		return nil, ErrAPINotFound
+	}
+	return api, nil
```
begelundmuller, line 37: nit: don't need to check the `nil` case, `ctrl.Get` is guaranteed to match the requested `Kind` (would be panic-worthy if that wasn't the case)

runtime/compilers/rillv1/parse_api.go
```diff
@@ -0,0 +1,53 @@
...
+	"fmt"
+	"strings"
+
+	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
+)
+
+// APIYAML
+type APIYAML struct {
+	SQL        string         `yaml:"sql"`
+	MetricsSQL *MetricSQLYAML `yaml:"metrics_sql"`
```
begelundmuller, line 14: 1. Would suggest changing `metrics_sql` to `metrics` (since `SQL` is a property of it)
2. Since the `MetricsSQLYAML` type isn't used, let's use an anonymous struct, similar to the other YAML types in this package (e.g. see `ReportYAML`).

runtime/compilers/rillv1/parse_api.go
```diff
@@ -0,0 +1,53 @@
...
+	MetricsView string   `yaml:"metrics_view"`
+	Measures    []string `yaml:"measures"`
+	Where       string   `yaml:"where"`
+	Limit       string   `yaml:"limit"`
+}
+
+// parseAPI parses an API definition and adds the resulting resource to p.Resources.
+func (p *Parser) parseAPI(ctx context.Context, node *Node) error {
+	tmp := &APIYAML{}
+	err := p.decodeNodeYAML(node, false, tmp)
```
begelundmuller, line 28: Should have `true` for `knownFields`

runtime/compilers/rillv1/parse_api.go
```diff
@@ -0,0 +1,53 @@
...
+	if err != nil {
+		return err
+	}
+
+	if tmp.SQL == "" && tmp.MetricsSQL == nil {
+		return fmt.Errorf("no SQL provided")
+	}
```
begelundmuller, line 35: Would suggest an error such as `missing resolver`

runtime/compilers/rillv1/parse_api.go
```diff
@@ -0,0 +1,53 @@
...
+
+	r.APISpec.Sql = strings.TrimSpace(tmp.SQL)
+	if tmp.MetricsSQL != nil {
+		r.APISpec.Metrics = new(runtimev1.MetricSQL)
+		r.APISpec.Metrics.Sql = strings.TrimSpace(tmp.MetricsSQL.SQL)
+		r.APISpec.Metrics.MetricsView = tmp.MetricsSQL.MetricsView
+		r.APISpec.Metrics.Measures = tmp.MetricsSQL.Measures
+		r.APISpec.Metrics.Where = strings.TrimSpace(tmp.MetricsSQL.Where)
+		r.APISpec.Metrics.Limit = strings.TrimSpace(tmp.MetricsSQL.Limit)
+	}
```
begelundmuller, line 50: It should be an error if two resolvers are configured for the same resource (in this case SQL and metrics SQL). There should only be one resolver.

runtime/compilers/rillv1/parse_api.go
```diff
@@ -0,0 +1,53 @@
...
+	"context"
+	"fmt"
+	"strings"
+
+	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
+)
+
+// APIYAML
+type APIYAML struct {
+	SQL        string         `yaml:"sql"`
```
begelundmuller, line 13: As an exception, the YAML shouldn't have a `SQL` property because that's already parsed by the `parseNode` function and can be accessed on `node.SQL`. This will automatically enable having SQL in a separate `.sql` file.

runtime/server/api.go
```diff
@@ -0,0 +1,109 @@
...
+	"io"
+	"net/http"
+
+	"github.com/rilldata/rill/runtime"
+	"github.com/rilldata/rill/runtime/pkg/observability"
+	"github.com/rilldata/rill/runtime/server/auth"
+	"go.opentelemetry.io/otel/attribute"
+)
+
+// nolint
```
begelundmuller, line 16: Why `nolint`?

runtime/server/server.go
```diff
@@ -191,6 +191,16 @@ func (s *Server) HTTPHandler(ctx context.Context, registerAdditionalHandlers fun
...
 
+	// custom APIs
+	err = gwMux.HandlePath("GET", "/v1/instances/{instance_id}/api/{name=**}", auth.GatewayMiddleware(s.aud, s.APIForName))
+	if err != nil {
+		panic(err)
+	}
+	err = gwMux.HandlePath("POST", "/v1/instances/{instance_id}/api/{name=**}", auth.GatewayMiddleware(s.aud, s.APIForName))
+	if err != nil {
+		panic(err)
+	}
```
begelundmuller, line 202: Would suggest registering as a normal HTTP handler directly on `httpMux` (more portable/native).

As of Go 1.22, it supports path params as well (see the runtime proxy PR for an example).

begelundmuller (changes requested):

proto/rill/runtime/v1/resources.proto
k-anshul: Makes sense. 
But a user would not be expected to set resolver for custom APIs right ? We would need to infer it basis the properties. For built-in APIs a resolver would be required.

runtime/server/api.go
k-anshul: It had some commented code around how resolver is expected to get called. It is temporary. Can be removed once resolver PR is integrated.

runtime/compilers/rillv1/parse_api.go
k-anshul: Understood. Also considering other review comments I believe even `metrics` isn't required and we can directly map `resolver_properties` to `map[string]any` and pass to `APIResolverOptions` ?

begelundmuller (approved):

PR #4211, opened 2024-02-29
Handle API error for timeseries

### LINKED ISSUE #4188: Time Dimension Detail: when a request has failed, show an error state rather than a perpetual loading spinner
[Slack discussion here](https://rilldata.slack.com/archives/C02T907FEUB/p1709058133734979)

djbarnwal:
Partially fixes #4188
This checks for error in the timeseries data store and propagates that to the TDD table. 

We do not currently check for error on the individual timeseries API which is called for each row of the TDD table. Since we are planning to move away from individual timeseries API to a batch aggregation API, will add the error handling directly there.

### DIFF
--- a/web-common/src/features/dashboards/time-dimension-details/TimeDimensionDisplay.svelte
+++ b/web-common/src/features/dashboards/time-dimension-details/TimeDimensionDisplay.svelte
@@ -201,9 +201,7 @@
       </div>
     </div>
   </div>
-{/if}
-
-{#if formattedData && comparisonCopy}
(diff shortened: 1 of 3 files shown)


ericpgreen2:
To check this out, I manually set `{#if $timeDimensionDataStore?.isError}` to `{#if true}`. I see this:

image

Looks like a bug with the TDD table below the error message. I'm not sure if this case would ever happen in reality, but looks like an "else if" statement could be safer.

ericpgreen2 (approved):
Approving so you can merge at your discretion.

PR #4216, opened 2024-02-29
Refactor: Replace Portal component with action

briangregoryholmes:
This PR removes usage of the Portal component and replaces it with a more stable and safe action.

Previously, the Portal component would remove/append the portal to the DOM every time it was needed and the creation of one portal did not mean that it would be re-used for another element. Its implementation as a component also obfuscated exactly what was happening in ways that caused it to be used incorrectly (portals within portals), creating multiple undifferentiated instances (see below).

Screenshot 2024-02-29 at 4 28 01 PM

Instead, this PR adds a stable portal element with the ID `rill-portal` to `app.html`. The portal action now appends/removes from this single zone by default. It can be placed directly on the element intended to be rendered elsewhere, making it clearer exactly what is happening.

If necessary, additional portal zones can be added to `app.html`. The portal action  takes an optional parameter of a CSS selector (string) or HTMLElement.


web-common/src/components/modal/ModalContainer.svelte
```diff
@@ -73,16 +73,19 @@
 
 <svelte:window on:keydown={handleKeydown} />
 
-<Portal>
+<!-- <Portal> -->
```
ericpgreen2, line 76: Leftover comment

web-local/src/app.html
ericpgreen2, line 1: We should add the same element to `web-admin/src/app.html`

web-common/src/features/dashboards/time-series/MeasureZoom.svelte
ericpgreen2, line 1: I'm hitting this error when selecting a subrange on a line chart:

image

ericpgreen2 (approved):
Generally LGTM! But left a couple small comments and hit one bug to resolve before merging.

djbarnwal (changes requested):
Seeing a few bugs

Dropdowns don't work as before - 
https://github.com/rilldata/rill/assets/4402679/af5d4c5e-e3a1-4f15-a7a3-f3510366e6ed

Scrubbing crashes the app. This is because the "Zoom" button uses the `FloatingElement` component but doesn't require a Portal

briangregoryholmes:
Just noticed one thing that I think is the cause of the failing test. Looking into it.

djbarnwal (commented):
The zoom button's height is bigger than the container it is placed in. The reason behind using `FloatingElement` for the zoom button was to show the zoom button fully. If you see the zoom button in this branch vs main, you can see the top of the button being cropped

**Main**
image

**Branch**
image

With reference to menus still seeing some weird error state. Jam here - https://jam.dev/c/49e4acb6-3b79-4d47-9b03-0a6c8891b4b1

djbarnwal (approved):

PR #4217, opened 2024-02-29
Alerts UI: handle incomplete Criteria

ericpgreen2:
This PR adds specific copy to the Alert Preview section for when Criteria are not fully specified.

The PR refactors the `AlertDataPreview` component into two distinct components: `DataPreview` for tab 1 and `AlertPreview` for tab 2. This refactor is motivated by:
- The Data Preview should not include the Alert's criteria in its API call.
- Each tab's Preview has its own helper text.


web-common/src/features/dashboards/stores/filter-utils.ts
```diff
@@ -253,3 +254,47 @@ export const sanitiseExpression = (
   if (!where?.cond?.exprs?.length) return undefined;
   return where;
 };
+
+export function isExpressionIncomplete(expression: V1Expression): boolean {
+  // Check if the operation is unspecified at any level of the condition.
+  function isOperationUnspecified(cond: V1Condition): boolean {
```
AdityaHegde, line 260: Any reason fro this to be inside the `isExpressionIncomplete` method?

AdityaHegde: Same with `isValDefinedAndNonEmpty`

AdityaHegde (approved):
Just a couple of nits.

web-common/src/features/dashboards/stores/filter-utils.ts
ericpgreen2: No strong preference – just moved them outside

Issue #4132, opened 2024-02-21
Relax restrictions for object naming conventions


AndrewRTsao:
Currently, we enforce the following object for object names: `/^[a-zA-Z_][a-zA-Z0-9_]*$/`. This regex appears to be a legacy frontend restriction and we should now be able to support any names that the underlying DB can support (i.e. DuckDB / Druid). We should relax these restrictions as end users may want to use names that include periods, hyphens, and/or spaces as an example.

Further context: https://rilldata.slack.com/archives/C04UY5NTXA9/p1708468146105439

PR #4230, opened 2024-03-04
Relax name constraints

ericpgreen2:
This PR relaxes file name constraints. Now, the only restriction on file names is that they cannot contain the following special characters: /, <, >, :, ", \\, |, ?, or *

Additionally, this PR fixes a long-standing issue where the frontend would sometimes include an extra leading `/` when referencing a file name in an API call, e.g. `http://localhost:9009/v1/instances/default/files/-//models/model1.sql`.

Closes https://github.com/rilldata/rill/issues/4132


web-common/src/features/entity-management/name-utils.ts
```diff
@@ -1,3 +1,8 @@
+export const nameRegex = /^[^<>:"/\\|?*]+$/;
```
AdityaHegde, line 1: Global constants like this should probably have 1st char capitalised.

AdityaHegde (approved):
Just a small nit

web-common/src/features/entity-management/name-utils.ts
ericpgreen2: Good point – though changed to ALL_CAPS

PR #4232, opened 2024-03-04
Api resolver

pjain1:


runtime/resolvers/sql.go
```diff
@@ -0,0 +1,263 @@
...
+}
+
+type SQLResolver struct {
+	resolvedSQL string
+	deps        []*runtimev1.ResourceName
+	olap        drivers.OLAPStore
+	releaseFunc func()
+}
+
+func New(ctx context.Context, opts *runtime.APIResolverOptions) (runtime.APIResolver, error) {
```
begelundmuller, line 32: nit: Seems like it doesn't need to be exported. And since we'll be adding more resolvers, should probably be something like `newSQL`.

begelundmuller (approved):
LGTM

PR #4236, opened 2024-03-04
Offer CLI flags for serving Rill Developer over HTTPS

briangregoryholmes:
This PR adds `--cert` and `--key` flags to the Rill CLI that, when provided, enable serving the application over HTTPS. Both are required and are at least validated as being correct file paths before starting the server. If neither are provided, the application is served normally over HTTP.


runtime/pkg/graceful/http.go
```diff
@@ -11,26 +11,37 @@ import (
 
 const httpShutdownTimeout = 15 * time.Second
 
+type Options struct {
+	Port     int
+	CertPath string
+	KeyPath  string
```
begelundmuller, line 17: But it's fine to keep these names since hopefully it's clear from the surrounding context that it's TLS related

runtime/pkg/graceful/http.go
```diff
@@ -11,26 +11,37 @@ import (
 
 const httpShutdownTimeout = 15 * time.Second
 
+type Options struct {
```
begelundmuller, line 14: This package also has some non-serving related logic in the `context.go` file, so I would suggest renaming this type to `ServeOptions`

runtime/pkg/debugserver/debugserver.go
```diff
@@ -12,5 +12,10 @@ import (
 
 func ServeHTTP(ctx context.Context, port int) error {
 	srv := &http.Server{} // An empty server will serve http.DefaultServeMux
-	return graceful.ServeHTTP(ctx, srv, port)
+
+	options := graceful.Options{
+		Port: port,
+	}
+
+	return graceful.ServeHTTP(ctx, srv, options)
```
begelundmuller, line 20:
```suggestion
	return graceful.ServeHTTP(ctx, srv, graceful.Options{
		Port: port,
	})
```

begelundmuller (changes requested):

cli/cmd/start/start.go
```diff
@@ -116,6 +118,20 @@ func StartCmd(ch *cmdutil.Helper) *cobra.Command {
 				return err
 			}
 
+			// If keypath or certpath provided, but not the other, display error
+			// If keypath and certpath provided, check if the file exists
+			if (tlsCertPath != "" && tlsKeyPath == "") || (tlsCertPath == "" && tlsKeyPath != "") {
+				return fmt.Errorf("both --cert and --key must be provided")
```
begelundmuller, line 124:
```suggestion
				return fmt.Errorf("both --tls-cert and --tls-key must be provided")
```

begelundmuller (approved):

PR #4237, opened 2024-03-04
Add vega chart editor and renderer

djbarnwal:
Add Vega lite chart rendered and editor


web-common/src/features/charts/render/utils.ts
```diff
@@ -0,0 +1,170 @@
...
+  return new Set(specs.flatMap((o) => Object.keys(o)));
+}
+
+interface SpecChanges {
+  width: false | number;
+  height: false | number;
+  isExpensive: boolean;
+}
+
+export function computeSpecChanges(
```
ericpgreen2, line 89: Can you add some comments to this function? E.g. Why do we do this? What classifies a change as expensive?

web-common/src/features/charts/render/VegaLiteRenderer.svelte
```diff
@@ -0,0 +1,30 @@
...
+  <p>{error}</p>
+{:else}
+  <VegaEmbed
+    {spec}
+    {data}
+    {signalListeners}
+    options={vegaLiteOptions}
+    bind:view
+    on:onNewView
+    on:onError={(e) => (error = e.detail)}
```
ericpgreen2, line 28: - "on" is redundant in these event names
- we're using kebab case for these events elsewhere in the app

So, what about `on:new-view` and `on:error`?

web-common/src/features/charts/render/VegaEmbed.svelte
```diff
@@ -0,0 +1,147 @@
...
+  let chartContainer: HTMLElement;
+
+  $: {
+    if (!shallowEqual(data, prevData)) {
+      update();
+    }
+    prevData = data;
+  }
+
+  $: {
```
ericpgreen2, line 40: There's a lot going on in this reactive statement. Could you somehow make it more readable, e.g. with comments or smaller functions?

ericpgreen2:
I realize this PR is a quick first-pass, but leaving some UX findings:
- At the base of the YAML editor, it'd be good to see a warning when the code isn't valid, like what we have in the Metrics Editor
- I’m sometimes _but not always_ getting keystroke-by-keystroke feedback
- You can remove the placeholder we use in the Metrics Editor
image
- This inspector toggle can be hidden with `showInspectorToggle: false` on the `WorkspaceHeader` component
image

ericpgreen2 (approved):
Approving so you can feel free to merge after addressing the code comments, in case you want to work on the UX items in follow-ups.

web-common/package.json
```diff
@@ -67,6 +67,7 @@
     "svelte-forms-lib": "^2.0.1",
     "svelte-headless-table": "^0.18.1",
     "svelte-preprocess": "^5.0.4",
+    "svelte-vega": "^2.1.0",
```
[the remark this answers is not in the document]
djbarnwal: Vega has multiple sub packages being used. We are also using `vega`, `vega-lite` (In `types.ts`). At some point we will also introduce `vega-tooltip`. `svelte-vega` packages all of them together. Any reason we would want to individually mention all packages?

djbarnwal:
Initially I imported the whole `svelte-vega` package (https://github.com/vega/svelte-vega/tree/main/packages/svelte-vega)  in our code base so that its easier to modify and make changes but seems like we don't need it at this point. Have replaced the internal files with `VegaLite` component which is an exported member of `svelte-vega`

PR #4240, opened 2024-03-05
Alerts time range refactor

AdityaHegde:
This PR updates how we handle time range in alerts.
- [x] Use the dashboard time range as the query time range.
- [x] Use splitByGrain for only the alert run interval.
- [x] Add the time range from dashboard to the the filters section.

### DIFF
--- a/web-common/src/features/alerts/EditAlertDialog.svelte
+++ b/web-common/src/features/alerts/EditAlertDialog.svelte
@@ -45,6 +45,7 @@
       name: alertSpec.title as string,
       snooze: getSnoozeValueFromAlertSpec(alertSpec),
       recipients: alertSpec?.emailRecipients?.map((r) => ({ email: r })) ?? [],
+      splitByTimeGrain: alertSpec.intervalsIsoDuration ?? "",
       ...extractAlertFormValues(queryArgsJson, metricsViewSpec),
     },
     validationSchema: alertFormValidationSchema,
@@ -99,5 +100,5 @@
   <DialogOverlay
     class="fixed inset-0 bg-gray-400 transition-opacity opacity-40"
   />
-  <BaseAlertForm isEditForm={true} {formState} on:close />
+  <BaseAlertForm {formState} isEditForm={true} on:close />
 </Dialog>
--- a/web-common/src/features/alerts/delivery-tab/AlertDialogDeliveryTab.svelte
+++ b/web-common/src/features/alerts/delivery-tab/AlertDialogDeliveryTab.svelte
@@ -1,6 +1,7 @@
 <script lang="ts">
   import FormSection from "@rilldata/web-common/components/forms/FormSection.svelte";
   import Select from "@rilldata/web-common/components/forms/Select.svelte";
+  import { AlertIntervalOptions } from "@rilldata/web-common/features/alerts/delivery-tab/intervals";
   import { SnoozeOptions } from "@rilldata/web-common/features/alerts/delivery-tab/snooze";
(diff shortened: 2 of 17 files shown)


web-common/src/features/alerts/EditAlertDialog.svelte
```diff
@@ -45,6 +45,7 @@
       name: alertSpec.title as string,
       snooze: getSnoozeValueFromAlertSpec(alertSpec),
       recipients: alertSpec?.emailRecipients?.map((r) => ({ email: r })) ?? [],
+      splitByTimeGrain: alertSpec.intervalsIsoDuration ?? "",
```
ericpgreen2, line 48: Let's call this `interval` to mirror the API parameter or `evaluationInterval` to mirror the form label.

web-common/src/features/dashboards/filters/FilterChipsReadOnly.svelte
```diff
@@ -36,14 +41,18 @@ The main feature-set component for dashboard filters
 </script>
 
 <div class="relative flex flex-row flex-wrap gap-x-2 gap-y-2 items-center">
+  {#if timeRange}
+    <ReadOnlyTimeRange {timeRange} />
```
ericpgreen2, line 45: It'd be nice symmetry with `DimensionFilterReadOnlyChip` and `MeasureFilterReadOnlyChip` if this component were named `TimeRangeReadOnlyChip`

web-common/src/features/alerts/delivery-tab/AlertDialogDeliveryTab.svelte
```diff
@@ -14,6 +15,36 @@
...
     title="Schedule"
   />
+  <FormSection
+    description="Choose the interval at which the alert is evaluated since the prior data refresh."
+    title="Evaluation interval"
+    tooltip
+  >
+    <Select
+      bind:value={$form["splitByTimeGrain"]}
+      id="splitByTimeGrain"
```
ericpgreen2, line 25: Ditto as mentioned above – the form field should be `interval` or `evaluationInterval`

ericpgreen2:
Some UX items:

1. It looks like the time range provided to the Data and Alert Previews is All Time, not what's inherited from the dashboard. Here, the All Time vaccinations is 52.5B and the Last 4 weeks vaccinations is 470k. <br>
image

2. On the Alert Detail page, can you:
- Remove "Split by time grain"
- Add "Evaluation interval" to the right of "Schedule" and show whatever is selected ("None" / "Hourly" / "Daily" / "Weekly") or "X seconds" (if defined through code and not one of the default options)<br>
image

3. Rather than this placeholder "None / Hourly / Daily / Weekly", mind switching it so the initial value is "None". Sorry, the Figma mock didn't reflect this.<br>
image

web-common/src/features/alerts/EditAlertDialog.svelte
```diff
@@ -90,6 +103,18 @@
...
+    const formValues = extractAlertFormValues(
+      queryArgsJson,
+      $metricsViewSpec.data,
+      $timeRange.data,
+    );
+    for (const fk in formValues) {
+      $form[fk] = formValues[fk];
+    }
+    console.log(formValues);
+  }
```
ericpgreen2, line 117: Is this code necessary?

ericpgreen2 (approved):
Looks good!

web-common/src/features/alerts/EditAlertDialog.svelte
AdityaHegde: Yes. Loading of metrics view spec and time range is async. So we need this to be loaded into the form. The other option is to get the args to the component, but this felt cleaner.

PR #4246, opened 2024-03-05
Runtime: `metrics_view` SQL resolver 

k-anshul:


runtime/resolvers/metricsview_sql.go
```diff
@@ -0,0 +1,257 @@
...
+	"golang.org/x/exp/maps"
+)
+
+func init() {
+	runtime.RegisterAPIResolverInitializer("Metrics", newMetricsViewSQL)
+}
+
+// sqlIdenitifer is regex pattern to identify a SQL idenitifier. The idenitifier may be wrapped in double quotes.
+// Additionally if double quotes are present in idenitifer, it is escaped with additional double quotes.
+var sqlIdenitifer = `[a-zA-z_][a-zA-Z0-9_]*|"(?:[^"]|"")*"`
```
begelundmuller, line 25: nit: misspelling of `identifier`

runtime/resolvers/metricsview_sql.go
```diff
@@ -0,0 +1,257 @@
...
+var (
+	aggregateRegex = regexp.MustCompile(fmt.Sprintf(`(?i)AGGREGATE\((?:(%s)\.)?(%s)\)`, sqlIdenitifer, sqlIdenitifer))
+	fromMVRegex    = regexp.MustCompile(fmt.Sprintf(`(?i)FROM\s+(%s)`, sqlIdenitifer))
+)
+
+func newMetricsViewSQL(ctx context.Context, opts *runtime.APIResolverOptions) (runtime.APIResolver, error) {
+	sql := opts.API.Spec.ResolverProperties.Fields["sql"].GetStringValue()
+	if sql == "" {
+		return nil, errors.New("no sql query found for metrics sql resolver")
+	}
```
begelundmuller, line 36: 1. I think there might be a risk of nil panics for this: `.ResolverProperties.Fields` (in case no properties are set, not sure how that serializes/deserializes).
2. Even though there's only `sql` now, I would consider adding a `metricsSQLProperties` struct and deserializing to that using `mapstructure`, similar to what we do for connector properties. It makes it easier to get an overview of available properties and to add new options.

runtime/resolvers/metricsview_sql.go
```diff
@@ -0,0 +1,257 @@
...
+		// example query = select dim1, aggregate(mv."my measure") from mv
+		// captures AGGREGATE("mv name"."my measure"), my name, measure
+		aggMatches := aggregateRegex.FindAllStringSubmatch(sql, -1)
+		for _, aggMatch := range aggMatches {
+			metricView := unquote(aggMatch[1])
+			var expr string
+			var found bool
+			if metricView == "" {
+				if len(mvToMeasureExprMap) > 1 {
+					return "", nil, fmt.Errorf("measure should be precedded with metric_view if api references more than one metric view")
```
begelundmuller, line 131: nit: spelling of `precedded`. Also would suggest a more explicit error message, something like `ambiguous reference to measure %q: use a fully qualified name such as "metrics_view.measure"`

runtime/server/api.go
```diff
@@ -83,3 +85,16 @@ func (s *Server) APIForName(w http.ResponseWriter, req *http.Request) {
 		return
 	}
 }
+
+func errorResponse(err error) []byte {
```
begelundmuller, line 89: Use this new util instead – search for `httputil.Handler` in the code for example usage: https://github.com/rilldata/rill/blob/c752b64e89640b0299b21c27af3332df66a75bf6/runtime/pkg/httputil/httputil.go#L12

begelundmuller (approved):

begelundmuller (changes requested):
Approved by mistake. Marking changes requested to prevent accidental merging. (Though it seems pretty close to ready.)

Issue #4178, opened 2024-02-27
Custom Dashboards: create a Dashboard canvas


ericpgreen2:
A Dashboard canvas's definition is discussed [here in Notion](https://www.notion.so/rilldata/Runtime-support-for-Vega-charts-and-dashboards-36460ad3593645cb852251970bd5e13d?pvs=4#476d9fc18be1402ab13d758cffc62bad). 

We should create a UI component that takes a Dashboard canvas spec as input and renders the referenced Chart components in the defined grid layout.

This can be done with mock Chart components while #4175 and #4176 are in progress.

Issue #4179, opened 2024-02-27
Custom Dashboards: setup Custom Dashboard YAML editor with “Preview” button


ericpgreen2:
In Rill Developer, when a user navigates to a custom dashboard:
- We should show a full-screen YAML editor to edit the [Custom Dashboard spec discussed here](https://www.notion.so/rilldata/Runtime-support-for-Vega-charts-and-dashboards-36460ad3593645cb852251970bd5e13d?pvs=4#476d9fc18be1402ab13d758cffc62bad). 
- In the top-right of the page, we should include a "Preview" button.
- Clicking on the "Preview" button should open up a new tab that renders the Custom Dashboard

PR #4249, opened 2024-03-05
Add CustomDashboard editing/preview surfaces

briangregoryholmes:
This is a draft PR for adding the editing and preview surfaces for Custom Dashboards.

Closes:
#4178
#4179


briangregoryholmes:
@ericpgreen2 @mindspank @AdityaHegde 

Here's a PR based on our discussion this morning. Absolutely still in progress, but wanted to to get some initial feedback.

A few notes:

I think the dashboard definition is "narrow" enough that we could easily display this as a two column layout rather than only offering full screen views. I found flipping back and forth to be a little annoying.

I've updated the DashboardComponent definition to have `x`, `y`, `width` and `height` values instead of simply `columns` and `rows`. Naming is absolutely up for discussion, but the general idea is that these are to be provided in grid units.

@AdityaHegde Let me know if I've changed things on the proto side correctly. I really only changed `resources.proto` and the dashboard parser and then ran `make proto.generate`

Not sure if it matters, but the `int64` type for these seems like overkill. Feels like we could make them uint8, but that type isn't recognized in `resources.proto`

ericpgreen2:
> I think the dashboard definition is "narrow" enough that we could easily display this as a two column layout rather than only offering full screen views. I found flipping back and forth to be a little annoying.

I'm all for showing this in a two-column layout. These dashboards should be responsive to small screen widths anyways (a current deficiency of the Metrics Explorer, which makes a two-column layout for it less practical at the moment).

In a responsive dev environment, maybe it makes sense to label the screen size somewhere in the Dashboard Preview, with a note/tooltip that says your dashboard layout will be different on different screen sizes.

> I've updated the DashboardComponent definition to have `x`, `y`, `width` and `height` values instead of simply `columns` and `rows`. Naming is absolutely up for discussion, but the general idea is that these are to be provided in grid units.
>
> I also removed the idea of setting the number of rows at the grid level and added a gap property. I didn't really see how setting rows would be used in a way that felt intuitive. I've just set this up to ingest the column count and create a grid that features square cells with a row count equal to whatever the max calculated row would be based on the component definitions.

On first pass, these ergonomics feel good to me! @mindspank, I know you've thought more about the Dashboard YAML, what do you think?

Playing around with it, I'm thinking a toggle to show gridlines on the Dashboard Preview could be really helpful.

AdityaHegde:
Hey @briangregoryholmes . Yes, the proto files are generated correctly. We could use `uint32` instead of int64.

briangregoryholmes:
Added grid line toggle.

ericpgreen2:
Checked in on the latest progress. At a macro level, the editing experience feels pretty good!

The main item that concerns me is responsiveness. Currently, the grid has the same number of columns for all screen sizes, which leads to dashboard miniaturization on smaller screens. If we were to take any of the in-the-wild custom dashboards that @mindspank shared in Slack, they wouldn't look very good on mobile and you'd have to zoom in to read them. We should outline the options and trade-offs for different approaches.

mindspank:
In general responsiveness does not work well for dashboards for a few reasons

- order is lost when grid is collapsing
- certain visualization types and information density does not work on smaller screen sizes
- you have different tasks you perform on mobile VS desktop

If we'd want to support mobile then I would suggest you'd had to define a separate grid with different content either as a mode in the dashboard or a different dashboard with a mobile type

mindspank:
There are also two levels of responsiveness, the grid and the visualization itself. For custom charts we don't have a lot of control but for Rill authored visualizations the different components of a chart should have a priority order and then we would layout (or not layout) depending on available size. 
Such as axis and value labels, label collision and entire axises.

See https://picassojs.studio/#stacked-bar-chart for example

web-common/src/features/custom-dashboards/VegaLiteRenderer.svelte
```diff
@@ -0,0 +1,33 @@
+<script lang="ts">
```
AdityaHegde, line 1: Can we reuse the render in features/charts ? If it is significantly different, better to rename this component with `Embed` in it.

AdityaHegde (changes requested):

web-common/src/features/custom-dashboards/CustomDashboard.svelte
```diff
@@ -0,0 +1,207 @@
...
+    resizingIndex !== null && mousePosition && startMouse
+      ? vector.subtract(mousePosition, startMouse)
+      : null;
+
+  $: resizeDimenions =
+    resizingIndex !== null && resizeOffset && initialResizeElementDimensions
+      ? vector.add(resizeOffset, initialResizeElementDimensions)
+      : null;
+
+  function handleMouseUp() {
```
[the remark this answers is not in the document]
briangregoryholmes: Not seeing where this will leak handlers, but this was refactored a bit in the latest code and that may have addressed your concern.

web-common/src/features/custom-dashboards/DashVegaRenderer.svelte
ericpgreen2, line 1: I see `features/charts/render/VegaLiteRenderer.svelte` is quite similar. Can we consolidate on one of these?

ericpgreen2 (approved):
I left a few comments, but generally looks good to be merged and we can iterate in follow-on PRs 👍

web-common/src/features/custom-dashboards/Chart.svelte
```diff
@@ -0,0 +1,61 @@
...
+  class="h-full w-full overflow-hidden pointer-events-none"
+  bind:this={container}
+>
+  {#if error}
+    <p>{error}</p>
+  {:else if !parsedVegaSpec}
+    <p>Chart not available</p>
+  {:else}
+    <DashVegaRenderer
+      spec={parsedVegaSpec}
```
AdityaHegde, line 56: DashVegaRenderer seem to want `data` is there a layer missing here? Because of this, I dont see the chart at all even if it is valid.

web-common/src/features/custom-dashboards/Element.svelte
```diff
@@ -0,0 +1,125 @@
...
+            on:change
+          />
+        {/each}
+
+        <div
+          class="size-full overflow-hidden"
+          class:shadow-lg={interacting}
+          style:border-radius="{radius}px"
+        >
+          <Chart {chartName} />
```
AdityaHegde, line 101: This seem to trigger a rerender of chart while dragging. Would be extremely inefficient on complex charts. Perhaps we could lock it while the mouse is held?

AdityaHegde (commented):
Mostly nits. The issue about chart not rendering is a blocker IMO

web-common/src/features/custom-dashboards/CustomDashboard.svelte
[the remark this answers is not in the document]
AdityaHegde: I think when i reviewed there were manual `addEventListeners` But the new code doesnt have that anymore with `svelte:window` so this is not a concern

web-common/src/features/custom-dashboards/Element.svelte
```diff
@@ -0,0 +1,125 @@
...
+  import { createEventDispatcher, onMount } from "svelte";
+  import { writable } from "svelte/store";
+  import ResizeHandle from "./ResizeHandle.svelte";
+  import * as ContextMenu from "@rilldata/web-common/components/context-menu";
+  import type { V1DashboardComponent } from "@rilldata/web-common/runtime-client";
+  import { goto } from "$app/navigation";
+
+  const zIndex = writable(0);
+
+  const options = [0, 0.5, 1];
```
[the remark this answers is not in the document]
briangregoryholmes: I'll try this out after merge.

PR #4255, opened 2024-03-06
Druid HTTP API driver

egor-ryashin:
Tested manually using dashboards UI.


runtime/drivers/druid/druid_api_sql_driver.go
```diff
@@ -0,0 +1,354 @@
...
+
+	bodyReader := bytes.NewReader(b)
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, c.dsn, bodyReader)
+	if err != nil {
+		return nil, err
+	}
+
+	req.Header.Add("Content-Type", "application/json")
+	// nolint:all
```
begelundmuller, line 60: Why the need for `nolint`?

runtime/drivers/druid/druid_api_sql_driver.go
```diff
@@ -0,0 +1,354 @@
...
+}
+
+func init() {
+	sql.Register("druid", &druidSQLDriver{})
+}
+
+type sqlConnection struct {
+	client *http.Client
+	dsn    string
+}
```
begelundmuller, line 38: A Druid driver for `database/sql` should be a standalone component, so can we put this in `runtime/drivers/druid/druidsqldriver` or something like that?

runtime/drivers/druid/druid.go
```diff
@@ -14,14 +14,16 @@ import (
...
 
 func init() {
-	drivers.Register("druid", driver{})
+	drivers.Register("druid", &driversDriver{})
 }
 
-type driver struct{}
+type driversDriver struct{}
+
+var _ drivers.Driver = &driversDriver{}
```
begelundmuller, line 22: Can we keep the old type name here?

runtime/drivers/druid/druid.go
```diff
@@ -14,14 +14,16 @@ import (
...
 
 // Open connects to Druid using Avatica.
 // Note that the Druid connection string must have the form "http://host/druid/v2/sql/avatica-protobuf/".
-func (d driver) Open(config map[string]any, shared bool, client activity.Client, logger *zap.Logger) (drivers.Handle, error) {
+func (d *driversDriver) Open(config map[string]any, shared bool, client activity.Client, logger *zap.Logger) (drivers.Handle, error) {
```
begelundmuller, line 26: What is the new DSN connection format that is expected?

runtime/drivers/druid/druid_api_sql_driver.go
```diff
@@ -0,0 +1,354 @@
...
+
+func (s *stmt) Exec(args []driver.Value) (driver.Result, error) {
+	return nil, fmt.Errorf("unsupported")
+}
+
+type JSONWalker struct {
+	dec *json.Decoder
+	err error
+}
```
begelundmuller, line 259: The walker makes the logic pretty hard to follow – I'm guessing it's for performance? I wonder if there's a way to have Druid return NDJSON instead or some other kind of result pagination, so we could just use normal struct deserialization?

begelundmuller (commented):
Also how can we use this to add support for query priorities in Druid?

runtime/drivers/druid/druid_api_sql_driver.go
egor-ryashin: ND arrays are possible:
```
["__time","aos","id","EXPR$3"]
["TIMESTAMP","OTHER","BIGINT","BOOLEAN"]
["2010-01-01T00:00:00.000Z","[{\"b\":\"gamma\",\"a\":1},{\"b\":\"beta\",\"a\":2}]",111,true]
```
Objects are possible but it's a waste of CPU and network:
```
{"__time":{"sqlType":"TIMESTAMP"},"aos":{"sqlType":"OTHER"},"id":{"sqlType":"BIGINT"},"EXPR$3":{"sqlType":"BOOLEAN"}}
{"__time":"2010-01-01T00:00:00.000Z","aos":"[{\"b\":\"gamma\",\"a\":1},{\"b\":\"beta\",\"a\":2}]","id":111,"EXPR$3":true}
```

runtime/drivers/druid/druid_api_sql_driver.go
egor-ryashin: we need to carry the body to allow streaming
image

runtime/drivers/druid/druid.go
egor-ryashin: like this `http://localhost:8888/druid/v2/sql`

runtime/drivers/druid/druid_api_sql_driver.go
egor-ryashin: fixed

egor-ryashin:
> Also how can we use this to add support for query priorities in Druid?

we can add `priority` in the request (https://druid.apache.org/docs/latest/querying/query-context), we need to enable lanes on the Druid side (https://druid.apache.org/docs/latest/configuration/#highlow-laning-strategy).

runtime/drivers/druid/olap.go
```diff
@@ -70,12 +70,12 @@ func (c *connection) Exec(ctx context.Context, stmt *drivers.Statement) error {
...
 	if stmt.DryRun {
-		// TODO: Find way to validate with args
-		prepared, err := c.db.PrepareContext(ctx, stmt.Query)
+		rows, err := c.db.QueryxContext(ctx, stmt.Query, stmt.Args...)
 		if err != nil {
 			return nil, err
 		}
-		return nil, prepared.Close()
+
+		return nil, rows.Close()
```
begelundmuller, line 78: This is a dry run, so it can't execute the query. The dry runs are called a lot during metric expression validation.

runtime/drivers/druid/druid.go
```diff
@@ -9,19 +9,21 @@ import (
 	"github.com/rilldata/rill/runtime/pkg/activity"
 	"go.uber.org/zap"
 
-	// Load calcite avatica driver for druid
-	_ "github.com/apache/calcite-avatica-go/v5"
+	// Load Druid database/sql driver
+	_ "github.com/rilldata/rill/runtime/drivers/druid/druidsqldriver"
```
begelundmuller, line 13: This is missing from the PR?

Can you fix merge conflicts and run CI?

begelundmuller (changes requested):

runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go
```diff
@@ -0,0 +1,267 @@
...
+
+	"github.com/google/uuid"
+)
+
+type druidSQLDriver struct{}
+
+var _ driver.Driver = &druidSQLDriver{}
+
+func (a *druidSQLDriver) Open(dsn string) (driver.Conn, error) {
+	client := http.Client{Timeout: time.Second * 10}
```
begelundmuller, line 22: That seems like a very short timeout? What if a query runs for e.g. 2 minutes?

runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go
```diff
@@ -0,0 +1,267 @@
...
+	s := make([]string, len(values))
+	for i, v := range values {
+		vv, ok := v.(string)
+		if !ok {
+			return nil
+		}
+		s[i] = vv
+	}
+	return s
+}
```
begelundmuller, line 55: nit: put util functions at the end of the file. See the style guide on function ordering: https://github.com/uber-go/guide/blob/master/style.md#function-grouping-and-ordering

runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go
```diff
@@ -0,0 +1,267 @@
...
+	return v, nil
+}
+
+func toStringArray(values []any) []string {
+	s := make([]string, len(values))
+	for i, v := range values {
+		vv, ok := v.(string)
+		if !ok {
+			return nil
+		}
```
begelundmuller, line 51: Seems like a case that should error?

runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go
```diff
@@ -0,0 +1,267 @@
...
+}
+
+type sqlConnection struct {
+	client *http.Client
+	dsn    string
+}
+
+var _ driver.QueryerContext = &sqlConnection{}
+
+func emptyTransformer(v any) (any, error) {
```
begelundmuller, line 41: nit: rename to `identityTransformer`

runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go
```diff
@@ -0,0 +1,267 @@
...
+					err := json.Unmarshal([]byte(v.(string)), &l)
+					if err != nil {
+						return nil, err
+					}
+					return l, nil
+				}
+			} else if c == "OTHER" {
+				transformers[i] = func(v any) (any, error) {
+					var l map[string]any
+					err := json.Unmarshal([]byte(v.(string)), &l)
```
begelundmuller, line 122: Won't the decoder already have parsed the value as JSON? Or are the values encoded as JSON at multiple levels in Druid?

runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go
```diff
@@ -0,0 +1,267 @@
...
+	return dr.columns
+}
+
+func (dr *druidRows) Close() error {
+	return dr.closer.Close()
+}
+
+func (dr *druidRows) Next(dest []driver.Value) error {
+	var a []any
+	err := dr.dec.Decode(&a)
```
begelundmuller, line 178: Instead of allocating a new array on each call to `Next`, could we maybe move it to a property on the `druidRows` and re-use the same array on each iteration?

runtime/drivers/druid/druid_test.go
```diff
@@ -104,10 +104,11 @@ func TestDruid(t *testing.T) {
 	brokerURL, err := container.PortEndpoint(ctx, "8082/tcp", "http")
 	require.NoError(t, err)
 
-	avaticaURL, err := url.JoinPath(brokerURL, "/druid/v2/sql/avatica-protobuf/")
+	druidAPIURL, err := url.JoinPath(brokerURL, "/druid/v2/sql")
```
begelundmuller, line 107: Can you also update our docs with the new expected API URL format (including username/password)

begelundmuller (changes requested):

runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go
egor-ryashin: decoder parses a string value, because complex types outputed as strings, see example https://github.com/rilldata/rill/pull/4255#discussion_r1522015728

begelundmuller (commented):
Just missing updating the DSN format in the docs in `rill/docs/docs/reference/olap-engines/druid.md`, then we can merge this.

pjain1:
does not look like it supports immediate query cancellation, we will have to send a delete request with queryId when context is cancelled.

pjain1:
can you check this, but getting this error locally when connecting to metrics cluster, this [project](https://github.com/rilldata/rill-qa/tree/main/rill-druid-metrics)
```
2024-03-26T15:16:59.646 WARN    Reconcile failed        {"name": "druid_requests_datasource", "kind": "MetricsView", "elapsed": "1.153s", "error": "could not find table \"druid-requests\": invalid character '<' looking for beginning of value"}
```
its originating at [`druid/olap.go`](https://github.com/rilldata/rill/blob/main/runtime/drivers/druid/olap.go#L173)

egor-ryashin:
> does not look like it supports immediate query cancellation, we will have to send a delete request with queryId when context is cancelled.

added

pjain1:
issue was that my project was still using older avatica connection string

runtime/drivers/druid/druid.go
```diff
@@ -26,15 +26,17 @@ var spec = drivers.Spec{
...
+	drivers.Register("druid", &driver{})
+	drivers.RegisterAsConnector("druid", &driver{})
 }
 
 type driver struct{}
 
+var _ drivers.Driver = &driver{}
+
 // Open connects to Druid using Avatica.
 // Note that the Druid connection string must have the form "http://host/druid/v2/sql/avatica-protobuf/".
```
begelundmuller, line 38: Need to update this comment

runtime/drivers/druid/druid.go
```diff
@@ -43,7 +45,7 @@ func (d driver) Open(config map[string]any, shared bool, client *activity.Client
 		return nil, fmt.Errorf("require dsn to open druid connection")
 	}
 
-	db, err := sqlx.Open("avatica", dsn)
+	db, err := sqlx.Open("druid", dsn)
```
begelundmuller, line 48: If the `dsn` contains `/avatica-protobuf`, can we try to parse and rewrite it to the new format? (For backwards compatibility.)

begelundmuller (changes requested):

begelundmuller (approved):
Approved, but there's a linter error that needs fixing

PR #4262, opened 2024-03-07
Remove resize listener action

briangregoryholmes:
This PR removes the `createResizeListenerFactory` module used in various places across the codebase as `use:listenToNodeResize`. In all instances, this action was used to capture the dimensions of an underlying DOM element. However, Svelte provides a native binding for this (`bind:clientHeight` and `bind:clientWidth`) and this PR replaces usage of the action with this binding.

While the Svelte binding has slightly more overhead than `ResizeObserver`, I propose making this change because it is used so infrequently throughout our codebase, removes the need for creating a store for the observed element, and the native binding will be updated to use ResizeObserver in Svelte 5 (https://github.com/sveltejs/svelte/issues/7583).

That said, we could also replace this with `bind:borderBoxSize`, which uses the ResizeObserver API already, though it is slightly less ergonomic.

This PR also removes the `getEltSize` function, which was only used in the `Dashboard` component and captured the total size of an element including margin. However, the element it was applied to does not have margin to account for.

In a few instances, knowing the dimension of the element is not necessary, but I'll address that in a separate PR.


ericpgreen2 (approved):
Very much like this line of thinking 👍

PR #4264, opened 2024-03-07
Encode dashboard url state only once

### LINKED ISSUE #3955: `encodeURIComponent()` is twice called on URL state
There are two instances of URL encoding: 
1. All parameters are URL-encoded in `web-common/src/runtime-client/fetchWrapper.ts`
2. The `state` object is specifically URL-encoded in `web-common/src/features/dashboards/proto-state/dashboard-url-state.ts` 

We should only URL-encode the state object once, not twice.

AdityaHegde:
closes #3955

Using `URL::searchParams` and `URL::toString` does the encoding for us. So an additional encoding is not necessary.


nishantmonu51 (approved):
LGTM, one que: does it breaks any previously generated stateful URLs ?

AdityaHegde:
Yes verified by saving old url and using on a new binary built from this branch.

PR #4265, opened 2024-03-07
Add typesafe context helper

briangregoryholmes:
This is a work in progress PR to add a wrapper function around our use of `setContext` and `getContext` that provides "type safety", reduces the need for namespaced context keys, and limits the number of imports required when getting context.


briangregoryholmes:
Will reopen

PR #4273, opened 2024-03-07
Add types for D3 packages

briangregoryholmes:
This PR adds types for the remaining D3 packages. This should help with some of the `no-unsafe-assignment` false positives.


ericpgreen2 (approved):
Approving so you can merge after fixing the CI errors. If that one `data-graphic` file is a rabbit hole, feel free to add it to the `svelte-check` ignore list and we can circle back later.

PR #4277, opened 2024-03-08
Update chart to add metrics sql support

AdityaHegde:
This PR replaces the old `queryName` and `queryArgs` with metrics sql support.

New format,
```
data:
  metrics_sql: |
    SELECT domain, count(*) as impression
    FROM AdBids_model
    GROUP BY domain
```

```
data:
  api: adbids_per_publisher
```

New API,
```
➜  rill git:(main) curl http://localhost:9009/v1/instances/default/charts/adbids_domain_chart/data
[{"domain":"sports.yahoo.com","impression":12889},{"domain":"news.google.com","impression":12883},{"domain":"facebook.com","impression":15609},{"domain":"instagram.com","impression":13104},{"domain":"news.yahoo.com","impression":14870},{"domain":"msn.com","impression":15526},{"domain":"google.com","impression":15119}]
```

Note: if any chart was created, the db file will have to be deleted.


runtime/server/queries_charts.go
```diff
@@ -0,0 +1,116 @@
...
+
+func (s *Server) GetChartData(w http.ResponseWriter, req *http.Request) error {
+	// TODO: telemetry
+
+	ctx := req.Context()
+	instanceId := req.PathValue("instance_id")
+	chartName := req.PathValue("chart_name")
+	// TODO: is a separate auth needed?
+	if !auth.GetClaims(ctx).CanInstance(instanceId, auth.ReadMetrics) {
+		return ErrForbidden
```
k-anshul, line 28: This returns grpc error which I think will map to a bad request. Should use `httputil.Error`.

runtime/server/queries_charts.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	ch, err := resolveChart(ctx, s.runtime, instanceId, chartName)
+	if err != nil {
+		return err
```
k-anshul, line 33: Same as above, it will again map to `http.StatusBadRequest`.

runtime/server/queries_charts.go
```diff
@@ -0,0 +1,116 @@
+package server
```
k-anshul, line 1: may be this should be `charts_api.go` >

runtime/server/server.go
```diff
@@ -208,6 +208,9 @@ func (s *Server) HTTPHandler(ctx context.Context, registerAdditionalHandlers fun
 	// custom API handler
 	observability.MuxHandle(httpMux, "/v1/instances/{instance_id}/api/{name...}", observability.Middleware("runtime", s.logger, auth.HTTPMiddleware(s.aud, httputil.Handler(s.apiForName))))
 
+	// custom Chart Data handler
+	observability.MuxHandle(httpMux, "/v1/instances/{instance_id}/charts/{chart_name...}", observability.Middleware("runtime", s.logger, auth.HTTPMiddleware(s.aud, httputil.Handler(s.GetChartData))))
```
k-anshul, line 212: Path also had `/data` suffix?

runtime/server/queries_charts.go
```diff
@@ -0,0 +1,116 @@
...
+		return httputil.Error(http.StatusInternalServerError, err)
+	}
+
+	return nil
+}
+
+func resolveChart(ctx context.Context, rt *runtime.Runtime, instanceID, name string) (*runtimev1.ChartSpec, error) {
+	ctrl, err := rt.Controller(ctx, instanceID)
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
```
k-anshul, line 101: should not return grpc errors otherwise everything gets mapped to bad request (which although is correct in this case). See `httputil.WriteError`.

runtime/server/server.go
AdityaHegde: Ah yes will add that.

runtime/server/queries_charts.go
AdityaHegde: Ya updating this method to just return the error as is. The caller will handle the not found and to use `httputil.Errorf`

proto/rill/runtime/v1/resources.proto
```diff
@@ -396,9 +396,10 @@ message Chart {
 
 message ChartSpec {
   string title = 1;
-  string query_name = 2;
-  string query_args_json = 3;
-  string vega_lite_spec = 4;
+  string metrics_sql = 2;
+  string api = 3;
+  map<string, string> args = 4;
```
begelundmuller, line 401: Instead of putting the YAML properties here, I would suggest having `data_resolver` and `data_resolver_properties` mirroring the `APISpec`. And then we should have one common YAML structure for parsing a `data:` clause (which we will want to use in all of APIs, charts, alerts, reports, etc.).

This would be similar to the `parseScheduleYAML` function (see [here](https://github.com/rilldata/rill/blob/9b768c083048a1897955b80b6bacb5e002feb1c3/runtime/compilers/rillv1/parse_source.go#L107-L108)) in the compiler. We might have this:

```go
type DataYAML struct {
  SQL string `...`
  MetricsSQL string `...`
  // Can add others. Like "api" for referencing another existing API.
}

func parseDataYAML(raw *DataYAML) (resolver string, resolverProps map[string]any, error) {
  ...
}

// In charts:
type ChartYaml struct {
  // ...
  Data *DataYAML `yaml:"data"`
}

// In alerts:
type AlertYAML struct {
  // ...
  Data *DataYAML `yaml:"data"` // This will eventually replace/deprecate the "query:" property
}

// In APIs:
type APIYAML struct {
  // ...
  Data *DataYAML `yaml:",inline"` // Inline since API's don't nest under "data":
}

// etc.
```

begelundmuller (changes requested):

runtime/compilers/rillv1/parse_chart.go
```diff
@@ -40,14 +38,24 @@ func (p *Parser) parseChart(node *Node) error {
...
 
 	if tmp.Data.API != "" {
 		node.Refs = append(node.Refs, ResourceName{Kind: ResourceKindAPI, Name: tmp.Data.API})
 	}
 
+	// Convert resolver properties to structpb.Struct before inserting the resource (since we can't return errors after that point)
+	resolverPropsPB, err := structpb.NewStruct(resolverProps)
+	if err != nil {
+		return fmt.Errorf("encountered invalid property type: %w", err)
+	}
```
begelundmuller, line 57: Actually, can we also move this into the `parseDataYAML` function, and return a `structpb` value instead of a map? Since otherwise, we will need to duplicate it every place that has a data section.

runtime/server/charts_api.go
```diff
@@ -119,3 +101,40 @@ func resolveChart(ctx context.Context, rt *runtime.Runtime, instanceID, name str
...
+			return nil, errors.New("api name is missing")
+		}
+		api, err = rt.APIForName(ctx, instanceID, apiName)
+		if err != nil {
+			return nil, err
+		}
+
+	default:
+		return nil, errors.New("unknown resolver")
+	}
```
begelundmuller, line 137: I would suggest refactoring `runtime.APIResolverOptions` to take `Resolver` and `ResolverProperties` as direct options instead of an `API`.

Since the point is to make the resolvers generic enough that all the code between the parser and the actual resolver doesn't need to know about the different available options.

begelundmuller (changes requested):

runtime/server/charts_api.go
begelundmuller: To make that work, you probably also need to add a new resolver called "API" that proxies to another resolver (and maybe checks to avoid infinite recursion if you want to be fancy?)

begelundmuller (approved):

PR #4279, opened 2024-03-08
Runtime: shorthands for configuring "dev" and "prod" envs

begelundmuller:
- This PR adds shorthands for configuring environment-specific overrides for the `dev` and `prod` environments.
- This saves one level of nesting for configuring our two primary environments.
- The shorthands work in both individual YAML files and for `rill.yaml`.
- We still support the current `env:environment_name:` syntax. This keeps environment-specific overrides available for custom environments.

Examples:
```yaml
# sources/my-source.yaml
connector: s3
path: s3://bucket/**/*.parquet
dev:
  path: s3://bucket/year=2024/month=01/**/*.parquet

# rill.yaml
prod:
  models:
    materialize: true
```

### DIFF
--- a/runtime/compilers/rillv1/parse_rillyaml.go
+++ b/runtime/compilers/rillv1/parse_rillyaml.go
@@ -38,7 +38,7 @@ type rillYAML struct {
 	Title string `yaml:"title"`
 	// Description of the project
 	Description string `yaml:"description"`
-	// The project's default OLAP connector to use (can be overriden in the individual resources)
+	// The project's default OLAP connector to use (can be overridden in the individual resources)
 	OLAPConnector string `yaml:"olap_connector"`
 	// Connectors required by the project
 	Connectors []struct {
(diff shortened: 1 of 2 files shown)


runtime/compilers/rillv1/parse_rillyaml.go
```diff
@@ -34,20 +34,34 @@ type VariableDef struct {
 
 // rillYAML is the raw YAML structure of rill.yaml
 type rillYAML struct {
-	Title         string `yaml:"title"`
-	Description   string `yaml:"description"`
+	// Title of the project
+	Title string `yaml:"title"`
+	// Description of the project
+	Description string `yaml:"description"`
+	// The project's default OLAP connector to use (can be overriden in the individual resources)
```
begelundmuller, line 41:
```suggestion
	// The project's default OLAP connector to use (can be overridden in the individual resources)
```

k-anshul (approved):

PR #4280, opened 2024-03-08
Fix: typing issues related to bisectData

briangregoryholmes:
This PR fixes some typing/ergonomic issues with the `bisectData` function as well as related components.

### DIFF
--- a/web-common/src/features/dashboards/time-series/DimensionValueMouseover.svelte
+++ b/web-common/src/features/dashboards/time-series/DimensionValueMouseover.svelte
@@ -44,7 +44,7 @@
   $: yValues = pointsData.map((dimension) => {
     if (!x) return { y: null, fillClass: undefined, name: "" };
 
-    const [bisected] = bisectData(
+    const { entry: bisected } = bisectData(
       new Date(x),
       "center",
       xAccessor,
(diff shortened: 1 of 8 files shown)


web-common/src/features/dashboards/time-series/DimensionValueMouseover.svelte
```diff
@@ -52,18 +62,18 @@
 
   $: points = yValues
     .map((dimension) => {
-      const y = dimension.y;
+      const y = Number(dimension.y);
```
djbarnwal, line 65: `Number(null)` casts y as `0` which is not correct. This makes `currentPointIsNull` always false. We need to preserve the value of y as `null`

web-common/src/features/dashboards/time-series/round-to-nearest-time-unit.ts
```diff
@@ -1,28 +1,28 @@
...
     year: "month",
     month: "day",
     week: "weekday",
     day: "hour",
     hour: "minute",
     minute: "second",
-    second: "milli",
+    second: "millisecond",
+    millisecond: "millisecond",
+    quarter: "month",
```
djbarnwal, line 18: Nit: would be nice to place `quarter` in the right order i.e. below `year`

djbarnwal (changes requested):

briangregoryholmes:
There are some issues up/down stream of this that present dimension.y as possibly being of type `string` or `Date` (in addition to a number or nullish value), but fixing that would be a more substantial change, though one I intend to get to at some point.

This is just intended to solve a small chunk. As such, I've changed the `measureValueFormatter` to accept a type `string` without changing its functionality (it'll just pass that string through) and changed the yValue mapping to pass strings and nulls through to that function. That means that the only thing that should be getting converted to a number here is a Date, which I'm guessing shouldn't really surface as a Y value. If you think that's a more real possibility, I can add a handler for that possibility and pass it through to the mouse over formatter as a string.

djbarnwal (approved):
This looks good. Data Graphics as a whole needs a bunch of type fixes and this is a great first step. Don't think we ever have `Date` in Y axis so we should be good here.

PR #4281, opened 2024-03-09
Rework duplicate source modal

briangregoryholmes:
This PR moves the duplicate source modal flow from its own dedicated component into the `AddSourceModal` component.

Previously, these concerns were separated in a way that caused UI issues, as seen below.

Blocker for #4216 

Before:
Screenshot 2024-03-08 at 7 23 32 PM

After:
Screenshot 2024-03-08 at 7 24 05 PM


AdityaHegde:
There one flow where the duplicate source dialog is used independently. That is when you drag drop a file without the `Add Source` modal open. Not a 100% sure if it makes sense to open `Add Source` dialog in that case? Either way duplicate source check there is broken with this PR.

briangregoryholmes:
Wasn't aware of the `FileDrop` component. In my view, I think it makes sense for `AddSourceModal` to be a singleton, and I've updated the PR as such. This should solve the issue you pointed out.

AdityaHegde (approved):

PR #4229, opened 2024-03-04
Updates to ChartInteractions, time series chart layout and MeasureBigNumber

briangregoryholmes:
This PR makes a series of related changes to the primary components used in the Dashboard to solve some UI bugs, inconsistencies with Figma and to prepare for #4216 and Custom Dashboards.

- Changes the `MeasureBigNumber` component to have a fixed width of 118 and height of 85, to match Figma
- Conditionally renders MeasureBigNumber as a `div` or `button` element depending on context
- Removes the use of CSS Grid in the `TimeSeriesChartContainer` to allow for better layout placement and scroll behavior
- Removes the use of Portal for `ChartInteractions`, solving a UI issue where its position would not update when resizing the sidebar
- Shifts the Y-axis labeling for the TimeSeries measures to the left to match their alignment on Figma
- Removes unnecessary wrapper elements in the dashboard

PR #4282, opened 2024-03-09
Fix: Measure overflow reversion

briangregoryholmes:
A fix in #4202 was reverted by #4229.


djbarnwal (changes requested):
Seeing some misalignment when the labels for comparison is long (3 digits, currency and percentage). Probably we can lower the size of comparison change labels? 

image

djbarnwal (approved):

PR #4283, opened 2024-03-09
Refactor workspace layout

briangregoryholmes:
This PR refactors the process by which we store, update and render the various layout parameters associated with a workspace to improve rendering performance and, hopefully, simplify the DX around how all of this works.

To that end, the following changes were made:

-Remove the use of context to pass around layout parameters
-Create new `WorkspaceLayout` and `Workspaces` stores using our new state management model
-Rely on CSS/Svelte transitions to animate in/out panel sliding rather than manual tween calculations
-Add a new `Resizer` component
-Collect the table and inspector state into a single localStorage item
-Debounce updating localStorage when state changes

This should solve a variety of UI responsiveness issues when interacting with these surfaces (e.g. toggling the panels should no longer experience frame drops). Additional performance improvements will come once the `PreviewTable` has been refactored.

### DIFF
--- a/web-common/src/features/generic-yaml-editor/YamlWorkspace.svelte
+++ b/web-common/src/features/generic-yaml-editor/YamlWorkspace.svelte
@@ -6,7 +6,7 @@
   export let fileName: string;
 </script>
 
-<WorkspaceContainer assetID={fileName} inspector={false}>
+<WorkspaceContainer inspector={false}>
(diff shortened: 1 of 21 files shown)


ericpgreen2 (approved):
Great! A couple small things:

1. The HorizontalSplitter's z-index has an issue<br>image<br>image
2. This is in the browser console when navigating to a dashboard
image

PR #4284, opened 2024-03-09
Runtime: Review/refactor custom APIs

begelundmuller:
- Use the same parser for resolvers in the API and Chart YAML
- Support the SQL resolver for chart data
- Change `metrics:sql:` to `metrics_sql:`
- Move `safeSQLName` to `EscapeIdentifier` on drivers.Dialect
- Fix bug where refs in resolvers didn't work (they used the shorthand kind string instead of the fully qualified kind strings, e.g. `Source` instead of `rill.runtime.v1.Source`)
- Rename `APIResolver` to `Resolver` – since they are also used for charts, and soon will be for alerts and reports
- Add docstrings at various levels to explain resolvers and how they work
- Clean up the resolver types, e.g. take a Go map instead of `structpb` and use `Refs()` instead of `Deps()` for alignment with reconciler terminology
- Use an MD5 hash for the query cache key to avoid very long cache keys – for example, SQL resolver returns the full SQL as the cache key
- Fix issue where charts that invoke an existing API did not pass the args provided in the YAML to the API resolver
- Prevent possible infinite recursion if the API proxy resolver proxies to itself
- For metrics SQL, use the connector defined on the metrics view(s) instead of the default OLAP
- Correctly set the `.export` bool in the SQL resolver (to true if the SQL is for an export)
- Make sure `{{ ref }}` tags in metrics SQL are added to the resolver's refs (for query cache invalidation)
- Return 400 instead of 500 HTTP errors for resolver errors (almost always due to malformed SQL or missing args)


runtime/resolvers/metrics_sql.go
```diff
@@ -0,0 +1,245 @@
...
+			return "", "", nil, err
+		}
+		mvToMeasureExprMap[metricView] = measureToExprMap
+		sql = strings.ReplaceAll(sql, match[0], fromQry)
+	}
+
+	// Expand AGGREGATE expressions
+	if len(mvToMeasureExprMap) > 0 {
+		// Example: SELECT dim1, AGGREGATE(mv."my measure") FROM mv
+		// The regex captures [AGGREGATE("mv name"."my measure"), my name, measure]
```
k-anshul, line 155: Nit: `mv name` and not `my name`

runtime/resolvers/metrics_sql.go
```diff
@@ -0,0 +1,245 @@
...
+//
+// This is a simple implementation that uses regular expressions. It does not support all SQL features.
+// It works by:
+//
+// 1. Expanding AGGREGATE(measure) into actual aggregate expressions from the metrics view definition.
+// 2. Converting "FROM metrics_view" clauses to nested SELECTs on the underlying table with filters based on the metrics view's security policy.
+//
+// Example transformation:
+// - Input: SELECT dim1, AGGREGATE(measure1) FROM metrics_view GROUP BY dim1
+// - Output: SELECT col1 AS dim1, SUM(col2) AS measure1 FROM underlying_model GROUP BY dim1
```
k-anshul, line 104: The output is not how it is expanded right now. 
1. `dimension` is not yet supported. 
2. `AGGREGATE(measure1)` is expanded as just `SUM(col2)` right now.
Though both can be fixed with the planned changes for support for dimension expressions.

k-anshul (approved):
LGTM

runtime/resolvers/sql.go
```diff
@@ -19,121 +18,146 @@ import (
...
+		Priority: r.priority,
 	})
 	if err != nil {
-		return runtime.Result{}, err
+		return nil, err
 	}
 	defer res.Close()
 
-	var out []map[string]interface{}
+	var out []map[string]any
```
pjain1, line 115: trying to understand, what difference will this make ?

k-anshul:
Just realised, on this point : `Make sure {{ ref }} tags in metrics SQL are added to the resolver's refs (for query cache invalidation)`
I believe we are leaning on to cases where a user can either target a `metrics_view` or `source/model` using ref notations and not both (to handle dimension expression using regex). The previous code I believe handled both and the current code only either.

runtime/resolvers/sql.go
```diff
@@ -183,87 +207,58 @@ func (r *SQLResolver) generalExport(ctx context.Context, w io.Writer, opts *runt
...
-	})
-	if err != nil {
-		return "", nil, err
+		refs = append(refs, runtime.ResourceNameFromCompiler(ref))
 	}
 
-	if dialect == drivers.DialectDuckDB {
-		ast, err := duckdbsql.Parse(sql)
+	// For DuckDB, we can do ref inference using the SQL AST (similar to the rillv1 compiler).
+	if dialect == drivers.DialectDuckDB && !meta.UsesTemplating {
```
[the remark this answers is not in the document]
begelundmuller: It's a good point – I'm torn on this. The issue is that in reconcile, we have a rule that if templating is used, no additional inference is done. So if we do it differently here, it will be inconsistent and potentially weirder to understand for devs.

I'll add the extra DuckDB parsing back for now, and layer it on top of any `{{ ref }}` in the templating. But will also raise a discussion about this inference in general, it's a little confusing IMO.

runtime/resolvers/metrics_sql.go
begelundmuller: Changed to a `TODO` :)

runtime/resolvers/sql.go
begelundmuller: It's the same, just syntactic sugar

begelundmuller:
> Just realised, on this point : `Make sure {{ ref }} tags in metrics SQL are added to the resolver's refs (for query cache invalidation)` I believe we are leaning on to cases where a user can either target a `metrics_view` or `source/model`(using ref notations) and not both (to handle dimension expression using regex). The previous code I believe handled both and the current code only either.

Good catch. I adapted the code to append the metric view to any refs discovered by the regular SQL resolver. It doesn't hurt to have additional refs in there, so doesn't matter that it will also discover and add the underlying model.

As I wrote in the comment on Parag's note about, I think this ref inference is a little weird. I also just discovered that DuckDB ref detection doesn't seem to work for subqueries in select fields. I think we should contemplate moving to something more explicit.

I think we may also need to tighten the constraints on metrics SQL in the upcoming changes.

k-anshul:
> I think we may also need to tighten the constraints on what you can do in metrics SQL for the upcoming changes to support dimensions.

Yeah I also believe if we are going ahead with regex based approach its better to just support `select dim, aggregate(measure) from tbl group by dim` to begin with and expand further with if requested by users.

runtime/resolver.go
```diff
@@ -2,109 +2,163 @@ package runtime
...
+	sum := hex.EncodeToString(hash.Sum(nil))
+	key := fmt.Sprintf("inst:%s:resolver:%s:hash:%s", opts.InstanceID, opts.Resolver, sum)
 
 	// Try to get from cache
-	if val, ok := opts.Runtime.queryCache.cache.Get(key); ok {
+	if val, ok := r.queryCache.cache.Get(key); ok {
+		log.Printf("CACHE HIT")
 		return val.([]byte), nil
 	}
+	log.Printf("CACHE MISS")
```
pjain1, line 146: is this intentional ?

PR #4288, opened 2024-03-11
Fix measure filter stopped applying

AdityaHegde:
## Checklist
- [x] Manual verification
- [ ] Unit test coverage
- [ ] E2E test coverage
- [x] Needs manual QA?

#### Issue addressed: 
Measure filters were not applying since a condition was incorrect.

#### Details:
Fixed the incorrect condition for keeping `dimensionThresholdFilters` and `toplists` in sync.

## Steps to Verify
1. Create any dashboard.
2. Add measure filter.
3. The data should update as and when the measure filter is changed.

### DIFF
--- a/web-common/src/features/dashboards/filters/measure-filters/MeasureFilterBody.svelte
+++ b/web-common/src/features/dashboards/filters/measure-filters/MeasureFilterBody.svelte
@@ -45,8 +45,11 @@
     class="font-bold text-ellipsis overflow-hidden whitespace-nowrap"
     style:max-width={labelMaxWidth}
   >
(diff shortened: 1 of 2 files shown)


djbarnwal (commented):
Getting an error on the Bids Sample Dash

```
code	2
message	"Conversion Error: Could not convert string 'circulars.adadapted.com' to DOUBLE"
details	[]
```

djbarnwal:
Also seeing a case for the NYC 311 data dashboard where filter is >=2200 but in agency I can see a data point with 1.9k value

image

djbarnwal (approved):

PR #4291, opened 2024-03-11
Bookmark followups based on feedback

AdityaHegde:
- Add the search bar.
- Add a message when there are no default/shared bookmarks.
- Move the description just below the name in the forms.


ericpgreen2:
Some UX items:

1. I'm unable to use the Bookmark's Search input when on the TDD view

2. Per the Figma mocks, it looks like the Edit and Delete icons should have their own hover state. But it'd be good to check with @jkhwu how that hover state should be reconciled with the hover state of the full bookmark item.<br><br>image

3. The "Filters" text is subtly a different font style (both size and color) than the other form item labels <br><br>image

AdityaHegde:
Thanks have addressed all your comments except the hover state. Will check that and raise a separate PR with any potential feedback from the demo.

PR #4296, opened 2024-03-11
Telemetry for success/failure of AI-based metrics view generation

begelundmuller:
Closes https://github.com/rilldata/rill-private-issues/issues/186

### DIFF
--- a/runtime/server/generate_metrics_view.go
+++ b/runtime/server/generate_metrics_view.go
@@ -92,13 +92,14 @@ func (s *Server) GenerateMetricsViewFile(ctx context.Context, req *runtimev1.Gen
 
 	// Try to generate the YAML with AI
 	var data string
-	var aiSucceeded bool
+	var aiResult *generateMetricsViewYAMLWithAIResult
+	var aiErr error
 	if req.UseAi {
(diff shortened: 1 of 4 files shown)


runtime/server/generate_metrics_view.go
```diff
@@ -94,12 +95,32 @@ func (s *Server) GenerateMetricsViewFile(ctx context.Context, req *runtimev1.Gen
...
+		if res != nil {
+			attrs = append(attrs,
+				attribute.Int("valid_measures_count", res.validMeasures),
+				attribute.Int("invalid_measures_count", res.invalidMeasures),
+			)
+		}
+		if err != nil {
+			attrs = append(attrs, attribute.String("error", err.Error()))
+		}
+		s.activity.Record(ctx, activity.EventTypeLog, "ai_generated_metrics_view_yaml", attrs...)
```
esevastyanov, line 123: This might be recorded as a metric and allowed in a `filterSink` as
```
		sink := activity.NewFilterSink(intakeSink, func(e activity.Event) bool {
			return e.EventType != activity.EventTypeMetric || name == "ai_generated_metrics_view_yaml_elapsed_ms"
		})
```
not very elegant as it requires to maitain a list of allowed metric names. but ideally, we anyway need to create a list of constants that represent all registered application metric names so that it is possible to quickly check if there are any duplicate names
so it is up to you, i'm fine with the existing solution

esevastyanov (approved):

PR #4304, opened 2024-03-12
Make dimension/measure in dashboard config insensitive

AdityaHegde:
`default_comparison => dimension` , `default_dimensions` and `default_measures` were doing an exact check. But dimension and measure names are case insensitive.

Making sure we do a case insensitive check when we load the default dashboard config.

### DIFF
--- a/web-common/src/features/dashboards/stores/dashboard-store-defaults.ts
+++ b/web-common/src/features/dashboards/stores/dashboard-store-defaults.ts
@@ -18,6 +18,8 @@ import { isoDurationToFullTimeRange } from "@rilldata/web-common/lib/time/ranges
 import type { TimeComparisonOption } from "@rilldata/web-common/lib/time/types";
 import { DashboardState_ActivePage } from "@rilldata/web-common/proto/gen/rill/ui/v1/dashboard_pb";
 import type {
+  MetricsViewSpecDimensionV2,
+  MetricsViewSpecMeasureV2,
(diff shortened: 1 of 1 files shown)


web-common/src/features/dashboards/stores/dashboard-store-defaults.ts
```diff
@@ -129,13 +133,21 @@ export function getDefaultMetricsExplorerEntity(
...
+        )
       : new Set(defaultMeasureNames),
     allMeasuresVisible:
       !metricsView.defaultMeasures?.length ||
       metricsView.defaultMeasures?.length === defaultMeasureNames.length,
     visibleDimensionKeys: metricsView.defaultDimensions?.length
-      ? new Set(metricsView.defaultDimensions)
+      ? new Set(
+          metricsView.defaultDimensions
+            .map((dd) => normaliseName(dd, metricsView.dimensions ?? []))
```
djbarnwal, line 148: Nit: Instead of using `enitities ?? []` everywhere, maybe handle `undefined` entities in `normaliseName` directly

djbarnwal (approved):
Looks good. Mentioned a nit.

web-common/src/features/dashboards/stores/dashboard-store-defaults.ts
AdityaHegde: Good point. Updated

PR #4306, opened 2024-03-12
Fix format of ducksql sample select query

AdityaHegde:
A query like this was erroring out,
```
select * from read_parquet('data/sales_data_more_variability_aggregated.parquet') USING SAMPLE 10%
```

This is because json format wont retain `.0`. This leads to a json like this,
```
"sample": {
  "sample_size": {
    "type": {
      "id": "DOUBLE",
      "type_info": null
    },
    "is_null": false,
    "value": 10 // Since this doesnt have a `.0` duckdb seem to assume it is uint
  },
  "is_percentage": true,
  "method": "System",
  "seed": -1
},
```

We need to change the `type.id` and match the type.

### DIFF
--- a/runtime/pkg/duckdbsql/ast_rewrite.go
+++ b/runtime/pkg/duckdbsql/ast_rewrite.go
@@ -301,6 +301,56 @@ func createGenericValue(key string, val any) (astNode, error) {
 	return n, err
 }
 
+func createStandaloneValue(val any) (astNode, error) {
+	var t string
+	switch vt := val.(type) {
+	// these are not supported for standalone value
+	// case map[string]any:
+	// case []any:
+	case bool:
+		t = "BOOLEAN"
+	case int, int32:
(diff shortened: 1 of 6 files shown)


runtime/pkg/duckdbsql/ast_rewrite.go
```diff
@@ -301,6 +301,56 @@ func createGenericValue(key string, val any) (astNode, error) {
...
+		if math.Floor(vt) == vt {
+			val = forceConvertToNum[int64](vt)
+			t = "BIGINT"
+		}
+	case string:
+		t = "VARCHAR"
+		val = fmt.Sprintf(`"%s"`, vt)
+	// TODO: others
+	default:
+		return nil, nil
```
begelundmuller, line 339: What others?

begelundmuller (approved):

PR #4307, opened 2024-03-12
Error telemetry for rill developer

### LINKED ISSUE #3387: Catch application runtime bugs in an error boundary and record stacktraces in telemetry
For capturing runtime errors:
- Svelte does not have a native error boundary. See: https://github.com/sveltejs/rfcs/pull/46.
- The oft-cited third-party Svelte library is the [Crown Framework](https://github.com/CrownFramework/svelte-error-boundary), but it is unmaintained.
- Hooking into `window.onerror` and `window.onunhandledrejection` seems like one valid option.
```js
window.onerror = (msg, url, line, col, error) => {
    // capture error & emit telemetry
    // show a full error page
};
window.onunhandledrejection = (event) => {
    // capture error & emit telemetry
    // show a full error page
};
```
- Sentry implies their [Svelte SDK](https://docs.sentry.io/platforms/javascript/guides/svelte/) captures runtime errors

For recording stacktraces in telemetry:
- Let's use Sentry
- Ideally we proxy the requests through our backend

AdityaHegde:
closes #3387

This adds telemetry to rill developer. Note that we do not send error telemetry in dev mode.


web-common/src/metrics/ErrorEventHandler.ts
```diff
@@ -37,12 +104,16 @@ export class ErrorEventHandler {
...
   }
 
-  public fireHTTPErrorBoundaryEvent(
+  private fireHTTPErrorBoundaryEvent(
     api: string,
     status: string,
     message: string,
     screenName: MetricsEventScreenName,
   ) {
+    if (dev) {
```
ericpgreen2, line 113: Nit:
1. The `metricsService` already keeps track of `dev` via the `is_dev` "common field". Is there a way to avoid this redundancy?
2. If not, `isDev` could be a property of the `ErrorEventHandler` class, rather than `dev` being pulled in here as a global.

ericpgreen2 (changes requested):
Great to see. Left some small comments that would be good to fix before merging.

web-admin/src/features/errors/error-utils.ts
```diff
@@ -19,23 +18,7 @@ import { ErrorStoreState, errorStore } from "./error-store";
...
-      return;
-    } else {
-      errorEvent?.fireHTTPErrorBoundaryEvent(
-        query.queryKey[0] as string,
-        error.response?.status + "" ?? error.status,
-        (error.response?.data as RpcStatus)?.message ?? error.message,
-        screenName,
-      );
-    }
+    errorEvent?.handleSvelteQueryError(error, query);
```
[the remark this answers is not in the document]
AdityaHegde: 1. 👍 
2. Data for it can be pulled async and it might not be initialised. We really need a greater refactor of the metrics service honestly.

web-common/src/metrics/ErrorEventHandler.ts
AdityaHegde: Dont want to pull information from `metricsService`. Better to send through `ErrorEventHandler` constructor.

ericpgreen2 (approved):

PR #4321, opened 2024-03-12
Runtime: Add interactive result limit of 10k rows for SQL resolver

begelundmuller:


runtime/resolvers/sql.go
```diff
@@ -115,6 +117,10 @@ func (r *sqlResolver) ResolveInteractive(ctx context.Context) (*runtime.Resolver
 
 	var out []map[string]any
 	for res.Rows.Next() {
+		if len(out) >= sqlResolverInteractiveRowLimit {
```
pjain1, line 120: this may not protect backend but only runtime as backends may materialize full results before returning results, whats the purpose ? for backend should we add limit clause but it can be tricky

runtime/resolvers/sql.go
```diff
@@ -17,6 +17,8 @@ import (
 	"google.golang.org/protobuf/types/known/structpb"
 )
 
+const sqlResolverInteractiveRowLimit = 10000
```
nishantmonu51, line 20: Nit: 
1. Can this be made configurable via some project specific config ? Either via undocumented rill.yaml or admin command. 
2. 10K seems a bit more restrictive, how about increasing the limit to 100K ?

begelundmuller: This is not for exports, only interactive API requests. For something like 100k rows, for interactive queries, shouldn’t we push people to use pagination? (I.e. make multiple requests and have args like `{{ .args.page_offset }}`.)

The network limit for responses is currently 4MB, so 100k seems quite likely to fail at network level anyway.

If you want, I can make this configurable. I'll do it in a follow-up PR because I'll need to add some new abstractions for it :)

nishantmonu51: sounds good.

nishantmonu51 (approved):

runtime/resolvers/sql.go
begelundmuller: I updated the PR to wrap the SQL in an outer select that pushes the limit into the DB. Might not be perfect for all edge cases, but seems pretty good for now.

pjain1 (approved):

PR #4323, opened 2024-03-12
Generate chart yaml using AI

AdityaHegde:
- [x] Endpoint to send prompt to chatgpt.
- [x] Parse chatgpt response and generate the yaml (happens in the UI right now).
- [x] Hook up a menu item to chart to generate the vega lite chart spec.
- [x] Support generating a chart directly from a table/model.
- [x] Support generating a chart directly from a metrics view.

These will have a follow up PR
- Prompt tuning.


runtime/resolver.go
```diff
@@ -34,6 +34,8 @@ type Resolver interface {
 	ResolveInteractive(ctx context.Context) (*ResolverResult, error)
 	// ResolveExport resolve data for export (e.g. downloads or reports).
 	ResolveExport(ctx context.Context, w io.Writer, opts *ResolverExportOptions) error
+	// ResolveSchema returns just the schema for the query
+	ResolveSchema(ctx context.Context) (*runtimev1.StructType, error)
```
begelundmuller, line 38: Can we just change `ResolverResult` to include the result schema when calling `ResolveInteractive`?

The `ResolveSchema` implementation anyway runs the full query right now, and I'm not sure we'll be able to avoid running the query to get the schema for some resolvers.

Running the query is probably anyway faster and cheaper than invoking and waiting for the AI, so shouldn't be a bottleneck :)

runtime/server/generate_chart.go
```diff
@@ -0,0 +1,174 @@
...
+	"google.golang.org/grpc/status"
+)
+
+func (s *Server) GenerateChartSpec(ctx context.Context, req *runtimev1.GenerateChartSpecRequest) (*runtimev1.GenerateChartSpecResponse, error) {
+	// TODO: telemetry
+
+	// Must have edit permissions on the repo
+	if !auth.GetClaims(ctx).CanInstance(req.InstanceId, auth.EditRepo) {
+		return nil, ErrForbidden
+	}
```
begelundmuller, line 24: `EditRepo` is a slightly weird permission if it doesn't write a file. But maybe a good proxy permission for now.

proto/rill/runtime/v1/api.proto
```diff
@@ -497,6 +504,19 @@ message GenerateMetricsViewFileResponse {
   bool ai_succeeded = 1;
 }
 
+message GenerateChartSpecRequest {
+  string instance_id = 1 [(validate.rules).string = {pattern: "^[_\\-a-zA-Z0-9]+$"}];
+  string chart = 2;
+  string table = 3;
+  string metrics_view = 4;
```
begelundmuller, line 511: Since only one of these should be set, use a `oneof`.

But it seems weird to be able to pass a `chart` here, since it's kind of self-referential. And with Vega spec validation, it won't even be possible to create a chart without already having a Vega spec set in it?

Also, how are you planning on adding metrics view support?

I'm wondering if it would make more sense to pass `resolver` and `resolver_properties` here, since that's what the chart will be based on. And then having a separate AI endpoint for generating the resolver from e.g. a table name (generate plain SQL resolver) or a metrics view (generate metrics SQL)

runtime/server/generate_chart.go
```diff
@@ -0,0 +1,174 @@
...
+	return resp, nil
+}
+
+// chartYAMLSystemPrompt returns the static system prompt for the chart generation AI.
+func chartYAMLSystemPrompt() string {
+	return `
+You are an agent whose only task is to suggest relevant chart based on a table schema.
+You should suggest the vega config for the chart based on the data.
+Replace the data field in vega lite json with,
+{ "name": "table" }
```
begelundmuller, line 155: With metrics view generation, it seemed that less fixed rule were better. So maybe consider saying "don't add a 'data' field". And then just add this programmatically afterwards.

runtime/server/generate_chart.go
```diff
@@ -0,0 +1,174 @@
...
+
+  "sql": "<the sql query used to fetch the data for the above chart. replace new line with \n>"
+}
+`
+}
+
+func chartYAMLUserPrompt(tblName, userPrompt string, schema *runtimev1.StructType) string {
+	prompt := fmt.Sprintf(`
+%s on the table named %q, which has the following schema:
+`, userPrompt, tblName)
```
begelundmuller, line 169: Since the user prompt might not match the sentence here, would suggest separating it out. Like:
```
Prompt provided by the user: %s

Based on a table named %q with schema: ...
```

runtime/server/generate_chart.go
```diff
@@ -0,0 +1,174 @@
...
+
+	// Convert the vega lite spec json to string
+	spec, err := json.Marshal(resp.VegaLiteSpec)
+	if err != nil {
+		return nil, err
+	}
+
+	return &runtimev1.GenerateChartSpecResponse{
+		VegaLiteSpec: string(spec),
+		Sql:          resp.SQL,
```
begelundmuller, line 94: If the AI is anyway generating the resolver also, why not just output the whole file here, like for metrics view generation?

begelundmuller (changes requested):

runtime/server/generate_chart.go
AdityaHegde: I am thinking there might be different configuration in the chart that would need to be untouched. Would probably be better for the yaml to be built in the UI in that case.

runtime/resolver.go
AdityaHegde: True since the final query would have to be run in the UI for preview this shouldnt be a problem.

runtime/server/generate_chart.go
AdityaHegde: One thing I am seeing is chatgpt does a good job of giving proper indentation. Parsing, updating and reformatting will lose this. So probably it might be good to keep it as is.

proto/rill/runtime/v1/api.proto
```diff
@@ -119,6 +120,20 @@ service RuntimeService {
...
+      body: "*"
+    };
+  }
+
+  rpc GenerateChartSpec(GenerateChartSpecRequest) returns (GenerateChartSpecResponse) {
+    option (google.api.http) = {
+      post: "/v1/instances/{instance_id}/files/generate-chart",
+      body: "*"
+    };
+  }
```
begelundmuller, line 135: 1. Add a docstring for each API (can be short, like for the other APIs) :) 
2. Since these endpoints don't generate files, they shouldn't use the `/files`/ path. For consistency between all the generator APIs, you might also consider refactoring `GenerateMetricsViewFile` to `GenerateMetricsView` and have the frontend write the file.

proto/rill/runtime/v1/api.proto
```diff
@@ -497,6 +512,30 @@ message GenerateMetricsViewFileResponse {
   bool ai_succeeded = 1;
 }
 
+message GenerateResolverRequest {
+  string instance_id = 1 [(validate.rules).string = {pattern: "^[_\\-a-zA-Z0-9]+$"}];
+  string prompt = 2;
+  string table = 3;
+  string connector = 4;
```
begelundmuller, line 519: nit: define `connector` before `table` – that's the order in all other protos with these two props

proto/rill/runtime/v1/api.proto
```diff
@@ -497,6 +512,30 @@ message GenerateMetricsViewFileResponse {
...
+  string metrics_view = 5;
```
begelundmuller, line 520: Maybe add a comment that if this is set, both `connector` and `table` should be empty

runtime/resolver.go
```diff
@@ -86,7 +95,7 @@ type ResolveOptions struct {
 }
 
 // Resolve resolves a query using the given options.
-func (r *Runtime) Resolve(ctx context.Context, opts *ResolveOptions) ([]byte, error) {
+func (r *Runtime) Resolve(ctx context.Context, opts *ResolveOptions) (*CachedResolverResult, error) {
```
begelundmuller, line 98: Would suggest not using a pointer, this is a very light struct

runtime/resolver.go
```diff
@@ -40,10 +41,18 @@ type Resolver interface {
...
 	// Data is a JSON encoded array of objects.
 	Data []byte
+	// Schema is the schema for the Data
+	Schema *runtimev1.StructType
 	// Cache indicates whether the result can be cached.
 	Cache bool
 }
 
+// CachedResolverResult is subset of ResolverResult that is cached
+type CachedResolverResult struct {
```
begelundmuller, line 51: Would suggest calling this `ResolveResult` because it is a result returned from `Resolve` (note: not "resolve", not "resolver"). And then move the definition to after `the `ResolveOptions` struct.

"Cached" is an implementation detail, and it will actually not always have been served from cache :)

runtime/resolver.go
```diff
@@ -139,28 +148,44 @@ func (r *Runtime) Resolve(ctx context.Context, opts *ResolveOptions) ([]byte, er
 
 	// Try to get from cache
 	if val, ok := r.queryCache.cache.Get(key); ok {
-		return val.([]byte), nil
+		res, ok := val.(*CachedResolverResult)
+		if ok {
+			return res, nil
+		}
+		// rerun the query if cache cannot be converted to *CachedResolverResult
```
begelundmuller, line 155: Having an unconditional cast is actually preferable. This would be a severe developer error and it should rightfully panic in this case.

runtime/resolver.go
```diff
@@ -139,28 +148,44 @@ func (r *Runtime) Resolve(ctx context.Context, opts *ResolveOptions) ([]byte, er
...
 	// Load with singleflight
 	val, err := r.queryCache.singleflight.Do(ctx, key, func(ctx context.Context) (any, error) {
 		// Try cache again
 		if val, ok := r.queryCache.cache.Get(key); ok {
-			return val, nil
+			res, ok := val.(*CachedResolverResult)
+			if ok {
+				return res, nil
+			}
+			// rerun the query if cache cannot be converted to *CachedResolverResult
```
begelundmuller, line 166: No need to cast here – it gets cast again after the singleflight `Do`.

runtime/resolver.go
```diff
@@ -139,28 +148,44 @@ func (r *Runtime) Resolve(ctx context.Context, opts *ResolveOptions) ([]byte, er
...
 	})
 	if err != nil {
 		return nil, err
 	}
-	return val.([]byte), nil
+	res, ok := val.(*CachedResolverResult)
+	if !ok {
+		return nil, errors.New("failed to parse query response")
+	}
+	return res, nil
```
begelundmuller, line 190: Same comment about the cast – this should definitely panic!

runtime/server/generate_chart.go
```diff
@@ -0,0 +1,134 @@
...
+	"github.com/rilldata/rill/runtime/server/auth"
+	"go.opentelemetry.io/otel/attribute"
+)
+
+func (s *Server) GenerateChartSpec(ctx context.Context, req *runtimev1.GenerateChartSpecRequest) (*runtimev1.GenerateChartSpecResponse, error) {
+	observability.AddRequestAttributes(ctx,
+		attribute.String("args.instance_id", req.InstanceId),
+		attribute.String("args.resolver", req.Resolver),
+		attribute.String("args.resolver_property", marshalResolverProperties(req.ResolverProperties.AsMap())),
+		// attribute.String("args.prompt", req.Prompt), // Adding this might be a privacy issue
```
begelundmuller, line 23: These metrics are not emitted on local, so you can uncomment for now for completeness. If we ever emit these from local, we will anyway need to go over all endpoints and add hashing for sensitive fields.

runtime/server/generate_chart.go
begelundmuller: Okay, you can keep if you want. But note that Go's JSON encoder is able to output indented JSON.

runtime/server/generate_chart.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	if api, ok := resolverProperties["api"]; ok {
+		if apiStr, ok := api.(string); ok {
+			return apiStr
+		}
+	}
+
+	return ""
+}
```
begelundmuller, line 134: The goal of the resolver abstraction is to avoid resolver-specific switches littered across the codebase. So can we change this to a normal JSON serialization?

runtime/server/generate_resolver.go
```diff
@@ -0,0 +1,236 @@
...
+	"google.golang.org/protobuf/types/known/structpb"
+)
+
+func (s *Server) GenerateResolver(ctx context.Context, req *runtimev1.GenerateResolverRequest) (*runtimev1.GenerateResolverResponse, error) {
+	observability.AddRequestAttributes(ctx,
+		attribute.String("args.instance_id", req.InstanceId),
+		attribute.String("args.connector", req.Connector),
+		attribute.String("args.table", req.Table),
+		attribute.String("args.metrics_view", req.MetricsView),
+		// attribute.String("args.prompt", req.Prompt), // Adding this might be a privacy issue
```
begelundmuller, line 29: Same comment as in the other API

runtime/server/generate_resolver.go
```diff
@@ -0,0 +1,236 @@
...
+	if req.Table != "" && req.Connector != "" {
+		// Get table info
+		tbl, err := olap.InformationSchema().Lookup(ctx, req.Table)
+		if err != nil {
+			return nil, status.Error(codes.InvalidArgument, "table not found")
+		}
+
+		start := time.Now()
+
+		sql, err := s.generateResolverForTable(ctx, req.InstanceId, req.Prompt, tbl.Name, dialect, tbl.Schema)
```
begelundmuller, line 76: Would suggest returning `resolver, resolverProps, err` here instead of `sql` (otherwise the function should be `generateSQLResolverForTable`)

runtime/server/generate_resolver.go
```diff
@@ -0,0 +1,236 @@
...
+		}
+		err = s.runtime.Query(ctx, req.InstanceId, q, 0)
+		if err != nil {
+			return nil, err
+		}
+
+		sql, err := s.generateResolverForMetricsView(ctx, req.InstanceId, req.Prompt, req.MetricsView, dialect, q.Result.Schema)
+		if err != nil {
+			return nil, err
+		}
```
begelundmuller, line 122: Same comment – mismatch between func name and return value

begelundmuller (changes requested):

runtime/server/generate_chart.go
AdityaHegde: While true, I dont think any library can do partial indentation like,
```
 {
    "$schema": "https://vega.github.io/schema/vega-lite/v5.json",
    "data": { "name": "table" },
    "mark": "arc",
    "encoding": {
      "theta": {"field": "total_records", "type": "quantitative"},
      "color": {"field": "publisher", "type": "nominal"}
    },
    "width": 300,
    "height": 300
  }
```

proto/rill/runtime/v1/api.proto
AdityaHegde: for 2 will raise a separate PR to refactor `GenerateMetricsViewFile`. This PR is already getting large.

djbarnwal (approved):

runtime/server/generate_chart.go
begelundmuller: That's probably true :) Makes sense

begelundmuller (approved):

PR #4329, opened 2024-03-12
Feat: set flags via rill.yaml

briangregoryholmes:
This PR enables the setting of feature flags via the rill.yaml file and updates the formatting expectations of localStorage keys to match the YAML and URL params.

- Adds a "features" key to the YAML spec that takes a list of strings
- Updates the FeaturesFlag store to use the new class based structure
- Adds a subscriber function that listens for changes to the rill.yaml and updates keys accordingly
- Adds a "system" or "user" scope to limit who can change feature flags
- Updates `rillYAML` and `RillYAML` types in the platform code

As an example, if you add (or remove) the following in your `rill.yaml`:

```yaml
features:
  - customDashboards
```
The UI should respond immediately (after a debounce) to toggle that feature on or off. Adding something like `adminServer` would not enable that feature as this flag has been created with `system` scope, but no warning is surfaced.
 
A flag needs only to be set in one location to be enabled, so if you have `customDashboards` set in the URL, localStorage and YAML, then it won't be turned off in the UI until it's removed from all three locations.

While this will not be a widely used feature, I would argue that we should remove the ability to set flags in the URL/localStorage now that it's enabled in the project YAML.

### DIFF
--- a/runtime/api.go
+++ b/runtime/api.go
@@ -4,10 +4,32 @@ import (
 	"context"
 
 	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
+	"google.golang.org/protobuf/types/known/structpb"
 )
 
+// BuiltinAPIs is a map of built-in APIs (i.e. predefined APIs that are not created dynamically from a project's YAML files.)
 var BuiltinAPIs = map[string]*runtimev1.API{}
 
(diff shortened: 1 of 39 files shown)


nishantmonu51 (approved):
LGTM, 👍

PR #4336, opened 2024-03-13
Add vega lite spec validation for charts

AdityaHegde:
This adds vega lite spec validation based on the spec : https://vega.github.io/schema/vega-lite/v5.json
We are using github.com/santhosh-tekuri/jsonschema/v5 to run the validation.


runtime/compilers/rillv1/parse_chart.go
```diff
@@ -13,6 +18,8 @@ type ChartYaml struct {
 	VegaLite   string           `yaml:"vega_lite"`
 }
 
+var vegaLiteSchema = jsonschema.MustCompile("https://vega.github.io/schema/vega-lite/v5.json")
```
begelundmuller, line 21: Let's vendor this spec into our repo (so charts work on slow/no internet connection).

Would suggest putting it in `runtime/compilers/rillv1/data/vega-lite-v5.json` and using the `embed` package here to build it into the binary.

begelundmuller (changes requested):

begelundmuller (approved):

PR #4341, opened 2024-03-13
Docs: Document when `{{ ref ... }}` tags are needed when using templating

begelundmuller:
Closes https://github.com/rilldata/rill-private-issues/issues/215


AndrewRTsao:
Incorporated these changes already on the refactored docs but will also merge for now so we can get them into our current docs until we push the refresh

AndrewRTsao (approved):

PR #4343, opened 2024-03-13
Runtime: Support OLAP connector overrides in analytical APIs

begelundmuller:
This PR enables a single Rill project to serve dashboards backed by different OLAP databases.

For some time, we have supported overriding the OLAP connector in model and dashboard YAML files. However, those overrides were not used in the analytical APIs, making the resulting resources broken in the UI. This PR fixes that.

Note that due to the UI locking down modeling when the default OLAP database is not DuckDB, to leverage multi-OLAP support, you must set DuckDB as the default OLAP connector (and then specify a different connector in the YAML for those dashboards that should not use DuckDB).

Closes https://github.com/rilldata/rill-private-issues/issues/194.

### DIFF
--- a/runtime/connections.go
+++ b/runtime/connections.go
@@ -111,13 +111,19 @@ func (r *Runtime) AI(ctx context.Context, instanceID string) (drivers.AIService,
 	return ai, release, nil
 }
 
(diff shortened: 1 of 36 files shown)


k-anshul (approved):

AndrewRTsao:
@begelundmuller I presume this didn't make it into v0.42 but will be available in the upcoming patch release?

PR #4334, opened 2024-03-13
Docs: metrics sql syntax and capabilities

k-anshul:
Should be merged after custom API docs are merged.

PR #4346, opened 2024-03-14
custom api docs

pjain1:
Fixes https://github.com/rilldata/rill-private-issues/issues/199. 

Will need to fix few things in https://github.com/rilldata/rill/pull/4334 like path and give some examples of using templating once this is merged


pjain1:
build is failing because `metrics-sql.md` is missing which is part of https://github.com/rilldata/rill/pull/4334. Let me just copy the changes to this PR.

docs/docs/develop/custom-apis/index.md
```diff
@@ -0,0 +1,35 @@
...
+
+Rill allows you to create custom APIs to pull data out in a flexible manner. You can write custom SQL queries and expose them as an API endpoint.
+
+## Create a custom API
+
+To create a custom API, create a new yaml file under `apis` directory in your Rill project. Currently, 
+we support two types of custom APIs:
+
+1. **SQL API**: You can write a SQL query and expose it as an API endpoint. This is useful when you want to directly 
+    write queries against a [model](../sql-models.md) that you have created. It should have the following structure:
```
k-anshul, line 16: technically a user can also query source right ? Are we not documenting that ?

docs/docs/integration/custom-api.md
```diff
@@ -0,0 +1,43 @@
...
+-H "Authorization: Bearer <token>"
+```
+
+There are two types of bearer tokens that you can use to access the custom APIs:
+1. **Service Account Token**: You can use a service account token to access the custom APIs.
+    Read more about [Service Account Tokens](../reference/cli/service). 
+
+    :::note
+    1. Service accounts have full access to all APIs so if there are security policies defined for metrics view being used in metrics_sql API, they will not be enforced.
+    2. Also since there is no user context available,m this means if the api being called uses `{{ .user.<attr> }}` in SQL templating, it will fail.
```
k-anshul, line 25: > Also since there is no user context available,m

This has extra m.

docs/docs/develop/custom-apis/metrics-sql-api.md
```diff
@@ -0,0 +1,23 @@
+---
+title: Metrics SQL API
+sidebar_label: Metrics SQL API
+sidebar_position: 30
```
k-anshul, line 4: I think we can change `sidebar_position` to have the following sequence : `sql-api`, `metrics-sql-api` and `metrics-sql`.

k-anshul: May be instead of keeping separate pages for `sql-api` and `metrics-sql-api` we can combine them in the main page itself. Just a thought.

docs/docs/develop/custom-apis/index.md
pjain1: yeah I was trying to avoid that since for druid there are no sources and in the metrics view we put datasource name under `model` so was not sure. Now I think about it, it should be ok to add that. What do you think ?

docs/docs/integration/custom-api.md
pjain1: thanks

docs/docs/develop/custom-apis/index.md
pjain1: added it for now

docs/docs/develop/custom-apis/metrics-sql-api.md
pjain1: fixed positions, separate pages look more cleaner to me

AndrewRTsao:
@pjain1 I just pushed a new commit to try to align this branch with our recent docs refactor and merging these changes in from main. See #4355. 

I would suggest taking a moment to review the changes to make sure the structure still makes sense given the new custom api docs we're pushing here (had to move some things around).

docs/docs/integrate/custom-api.md
```diff
@@ -2,10 +2,10 @@
 title: "Custom API Integration"
 description: How to integrate custom APIs with your application
 sidebar_label: "Custom API Integration"
-sidebar_position: 20
+sidebar_position: 11
 ---
 
-Rill exposes [custom APIs](../develop/custom-apis/index.md) you have created under `apis` folder as HTTP endpoints 
+Rill exposes [custom APIs](/integrate/custom-apis/index.md) you have created under `apis` folder as HTTP endpoints
```
pjain1, line 8: so there is not `develop` section now ?

pjain1 (commented):
👍

docs/docs/integrate/custom-apis/_category_.yml
```diff
@@ -0,0 +1,4 @@
+position: 14
```
pjain1, line 1: should this section go under `build` now as previously it was under `develop` ?

AndrewRTsao: The original thinking behind the refactored docs structure is an end-to-end user journey, where a new person that discovers Rill could easily follow our docs to get an initial project deployed and set up (on Rill Cloud). If you're curious, you can check out this [Notion doc](https://www.notion.so/rilldata/Rill-Docs-Structure-52de7fe35d5a4e598ecf835ccaf4807d).

Logically speaking, we should be putting this section under either `build` or `integrate`. I would consider custom APIs more of an "advanced" concept and not really necessary for most end users so I personally think it makes the most sense to keep it under `integrate` (which is really more for developers). However, I can see the argument for `build` as well as it's still part of the building process, just for more advanced project setups. @mindspank Did you want to make a judgment call on where these docs should live in our new structure?

pjain1: Ok makes sense, feel free to merge  either way you decide. I am ok with both.

mindspank: Feels like integrate is the correct place

nishantmonu51 (approved):

PR #4348, opened 2024-03-14
Alerts: Fix data at watermark boundary is excluded

begelundmuller:
Also includes some minor comment and logging improvements in the alerts reconciler.

Closes https://github.com/rilldata/rill-private-issues/issues/198.

### DIFF
--- a/runtime/queries/resource_watermark.go
+++ b/runtime/queries/resource_watermark.go
@@ -114,6 +114,10 @@ func (q *ResourceWatermark) resolveMetricsView(ctx context.Context, rt *runtime.
 	}
 
 	if !t.IsZero() {
+		// Hacky workaround for the following issue: the watermark is used as the *exclusive* upper bound for time ranges.
+		// We need to add a small delta to ensure the row with the watermark is included in the result.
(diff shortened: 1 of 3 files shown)


runtime/queries/resource_watermark.go
```diff
@@ -114,6 +114,10 @@ func (q *ResourceWatermark) resolveMetricsView(ctx context.Context, rt *runtime.
 	}
 
 	if !t.IsZero() {
+		// Hacky workaround for the following issue: the watermark is used as the *exclusive* upper bound for time ranges.
+		// We need to add a small delta to ensure the row with the watermark is included in the result.
+		t = t.Add(time.Second)
```
AdityaHegde, line 119: Would just a ms be enough?

AdityaHegde (approved):
Just a small comment. Otherwise LGTM

runtime/queries/resource_watermark.go
begelundmuller: I tried it and it didn't work outright with ClickHouse – apparently they have timestamps at second granularity by default

PR #4371, opened 2024-03-18
Slack integration for alerts and reports

esevastyanov:
- [x] slack driver to deliver notifications to channels and application "messages" tab (by email)
- [x]  slack driver to deliver notifications using webhooks
- [x] refactored alert yaml/options/spec
- [x] refactored report yaml/options/spec
- [x] backward compatibility for alert yaml
- [x] backward compatibility for report yaml
- [x] docs
- [x] slack application
- [x] metrics
- [x] unit tests
- [x] polish message templates

alert spec has a `notify` section
```
...
notify:
  on_recover: true
  renotify: true
  renotify_after: 24h
  email:
    recipients:
      - recipient@example.com
      ...
  slack:
    emails:
      - recipient@example.com
      ...
    channels:
      - alerts
      ...
    webhooks:
      - https://...
      ...
```
report spec has `notify` section
```
notify:
  email:
    recipients:
      - recipient@example.com
      ...
  slack:
    emails:
      - recipient@example.com
      ...
    channels:
      - alerts
      ...
    webhooks:
      - https://...
      ...
```
slack driver only needs a bot token to post messages to channels and "messages" tab
```
start devproject --env connector.slack.bot_token=xoxb-...
```

Required scopes for a Slack application

Permission | Description
-- | --
chat:write | Send messages
users:read | View people in a workspace (required only to search a user by an email)
users:read.email | View email addresses of people in a workspace (required only to search a user by an email)
incoming-webhook | Post messages to specific channels in Slack


esevastyanov:
@AdityaHegde could you please review the change of UI components? The only thing is done there - I renamed api parameters related to email recipients of alerts and reports

k-anshul (changes requested):

admin/server/reports.go
```diff
@@ -261,7 +268,7 @@ func (s *Server) UnsubscribeReport(ctx context.Context, req *adminv1.Unsubscribe
 		return nil, status.Error(codes.InvalidArgument, "user is not subscribed to report")
 	}
 
-	if len(opts.Recipients) == 0 {
+	if len(opts.EmailRecipients) == 0 && len(opts.SlackEmails) == 0 && len(opts.SlackChannels) == 0 && len(opts.SlackWebhooks) == 0 {
```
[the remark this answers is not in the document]
esevastyanov: I guess the ability to unsubscribe is a requirement for private notifications. As for group notifications (slack webhooks, channels), there is a way to update the corresponding subscription via `EditReport`

AdityaHegde (approved):
Looks good from a UI perspective

runtime/drivers/slack/notifier.go
```diff
@@ -0,0 +1,186 @@
...
+		}
+		err := slack.PostWebhook(webhook, &payload)
+		if err != nil {
+			return fmt.Errorf("slack webhook error: %w", err)
+		}
+	}
+	return nil
+}
+
+type SlackNotifierProps = runtimev1.NotifierSpec_Slack
```
k-anshul, line 165: Why is this separate type required ?

esevastyanov: Removed

proto/rill/runtime/v1/resources.proto
```diff
@@ -299,16 +300,35 @@ message AlertSpec {
...
+  bool notify_on_recover = 1;
+  bool notify_on_fail = 2;
+  bool notify_on_error = 3;
+  bool renotify = 4;
+  uint32 renotify_after_seconds = 5;
+  repeated NotifierSpec notifiers = 6;
+}
+
+message NotifierSpec {
+  string connector = 1;
```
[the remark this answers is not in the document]
esevastyanov: Removed the connector name

runtime/drivers/slack/notifier.go
```diff
@@ -0,0 +1,186 @@
...
+	htemplate "html/template"
+	"time"
+
+	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
+	"github.com/rilldata/rill/runtime/drivers"
+	"github.com/slack-go/slack"
+)
+
+func (h *handle) SendScheduledReport(s *drivers.ScheduledReport, spec drivers.NotifierSpec) error {
+	slackSpec, err := h.validateSlackSpec(spec)
```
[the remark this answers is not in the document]
esevastyanov: Removed

esevastyanov:
@begelundmuller what do you think about going with `slack.users` instead of `slack.emails`?

proto/rill/runtime/v1/resources.proto
```diff
@@ -296,13 +300,34 @@ message AlertSpec {
...
+
+message EmailNotifierSpec {
+  repeated string recipients = 1;
+}
+
+message SlackNotifierSpec {
+  repeated string emails = 1;
+  repeated string channels = 2;
+  repeated string webhooks = 3;
 }
```
begelundmuller, line 331: A few thoughts:
1. I would avoid calling this `NotifierSpec` because it sounds like a spec for a "notifier" resource, but no such resource exists. All the other `XXXSpec` types refer to resource types. Instead, I would propose calling it just `message Notifier` (similar to `message Schedule` which is referenced by multiple resource specs).
2. I feel ambivalent about hard-coding the properties for the different notification implementations (email, Slack, etc.). We did that with source properties and also analytical APIs, and we ended up refactoring both. It doesn't fit nicely with the extensible nature of our connectors (especially if we need to make some of them proprietary, i.e. stored in a private repo). I would consider mirroring what we did with source properties and with resolvers (e.g. see the `SourceSpec` and `APISpec` types), e.g.:
```proto
message Notifier {
  string connector = 1;
  google.protobuf.struct properties = 2;
}
```
3. Lastly, we now have a level of nesting where both the notifiers are repeated and the individual properties (like `emails` or `channels`) are repeated. Would it maybe make sense to only have `repeated` at the root to simplify things, and still have the same behavior? I'm thinking of this kind of refactor (not sure it's a good idea, just a suggestion):
```yaml
notify:
- connector: slack
   channel: #channel-foo
- connector: slack
   channel: #channel-bar
- connector: email
   recipient: a@example.com
- connector: email
   recipient: b@example.com
```

runtime/drivers/notifier.go
```diff
@@ -0,0 +1,49 @@
...
+	"time"
+
+	"github.com/rilldata/rill/proto/gen/rill/runtime/v1"
+)
+
+// Notifier sends notifications.
+type Notifier interface {
+	SendAlertStatus(s *AlertStatus, spec NotifierSpec) error
+	SendScheduledReport(s *ScheduledReport, spec NotifierSpec) error
+}
```
begelundmuller, line 14: Considering the generic suggestion above, the usage of this could be refactored to:
```go
# pseudo-code
for _, notifier := range alert.Spec.Notifiers { // Could be done in parallel
  connector := runtime.AcquireHandle(notifier.Connector)
  notifier := connector.AsNotifier(notifier.Properties) // Notice the recipient properties are passed here
  notifier.SendAlertStatus(alertStatus)
}
```

begelundmuller (changes requested):

begelundmuller:
> @begelundmuller what do you think about going with `slack.users` instead of `slack.emails`?

I like that idea, but don't have a strong opinion on it.

proto/rill/runtime/v1/resources.proto
esevastyanov: As for the item 3, if a connector requires additional props passed, then these props are duplicated for each connector recipient (user/channel/webhook)

runtime/reconcilers/report.go
```diff
@@ -301,10 +302,11 @@ func (r *ReportReconciler) sendReport(ctx context.Context, self *runtimev1.Resou
...
 
 	sent := false
-	for _, notifier := range rep.Spec.NotifySpec.Notifiers {
-		switch notifier.Spec.(type) {
-		case *runtimev1.NotifierSpec_Email:
-			for _, recipient := range notifier.GetEmail().Recipients {
+	for _, notifier := range rep.Spec.Notifiers {
+		switch notifier.Connector {
+		case "email":
+			recipients := pbutil.ToSliceString(notifier.Properties.AsMap()["recipients"].([]any))
```
begelundmuller, line 308: Would also suggest a `val, ok := ...` type assertion here to keep it safe from panics

runtime/server/controller.go
```diff
@@ -394,17 +395,19 @@ func (s *Server) applySecurityPolicyReport(ctx context.Context, r *runtimev1.Res
...
+			for _, recipient := range recipients {
 				if recipient == email {
 					return r, true, nil
 				}
 			}
-		case *runtimev1.NotifierSpec_Slack:
-			for _, recipient := range notifier.GetSlack().Emails {
-				if recipient == email {
+		case "slack":
+			users := pbutil.ToSliceString(notifier.Properties.AsMap()["users"].([]any))
```
begelundmuller, line 408: Same comment about safe assertions

runtime/server/controller.go
```diff
@@ -440,9 +443,22 @@ func (s *Server) applySecurityPolicyAlert(ctx context.Context, r *runtimev1.Reso
...
+		switch notifier.Connector {
+		case "email":
+			recipients := pbutil.ToSliceString(notifier.Properties.AsMap()["recipients"].([]any))
+			for _, recipient := range recipients {
+				if recipient == email {
+					return r, true, nil
+				}
+			}
+		case "slack":
+			users := pbutil.ToSliceString(notifier.Properties.AsMap()["users"].([]any))
```
begelundmuller, line 456: Same

begelundmuller (changes requested):

runtime/drivers/notifier.go
```diff
@@ -1,20 +1,17 @@
...
-	SendScheduledReport(s *ScheduledReport, spec NotifierSpec) error
+	SendAlertStatus(s *AlertStatus) error
+	SendScheduledReport(s *ScheduledReport) error
 }
 
-type NotifierSpec = interface{}
-
 type AlertStatus struct {
 	ToEmail        string
 	ToName         string
```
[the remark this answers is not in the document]
esevastyanov: Will remove this once email notifier is created analogously to Slack notifier. Placed a todo reminder

runtime/drivers/slack/notifier.go
```diff
@@ -2,129 +2,168 @@ package slack
...
+	users     []string
+	channels  []string
+	webhooks  []string
+	templates *template.Template
+}
+
+func newNotifier(token string, props map[string]any) *notifier {
+	users := pbutil.ToSliceString(props[UsersField].([]any))
+	channels := pbutil.ToSliceString(props[ChannelsField].([]any))
+	webhooks := pbutil.ToSliceString(props[WebhooksField].([]any))
```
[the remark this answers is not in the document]
esevastyanov: Moved Slack props encode/decode into Slack notifier so that it is possible to reuse them

begelundmuller:
Also note the merge conflicts

runtime/reconcilers/alert.go
```diff
@@ -130,7 +130,7 @@ func (r *AlertReconciler) Reconcile(ctx context.Context, n *runtimev1.ResourceNa
 	adhocTrigger := a.Spec.Trigger
 	specHashTrigger := a.State.SpecHash != specHash
 	refsTrigger := a.State.RefsHash != refsHash && a.Spec.RefreshSchedule != nil && a.Spec.RefreshSchedule.RefUpdate
-	scheduleTrigger := a.State.NextRunOn != nil && !a.State.NextRunOn.AsTime().After(time.Now())
+	scheduleTrigger := a.State.NextRunOn != nil && a.State.NextRunOn.AsTime().Before(time.Now())
```
begelundmuller, line 133: This is not equivalent – specifically, if `NextRunOn == time.Now()`, we should trigger

esevastyanov: Reverted and fixed linter changes

begelundmuller (approved):

k-anshul (approved):

PR #4372, opened 2024-03-18
Update iframe embedding documentation

AndrewRTsao:
Update the iframe embedding documentation to capture the latest changes, including the ability to embed a project-level dashboard list or individual dashboard resource (with the ability to toggle navigation). Based on #4367 and #4247. 

Rather than documenting every available explicit option, I decided to only include:
- `resource`
- `navigation`
- `theme`
- `user_id`
- `user_email`
- `attributes`
- `ttl_seconds`

Including the other options (`organization`, `project`, `branch`, `kind`, `state`) felt either redundant because it's already specified in the URL itself or could potentially introduce confusion since it's not apparently obvious at the moment when users would want to set a different `branch`, `kind`, or `state` (this could change in the future).


docs/docs/integrate/embedding.md
```diff
@@ -266,12 +271,46 @@ The API accepts the following parameters:
...
-| attributes | Json payload to be put in the access token, used to pass attributes to the dashboard for enforcing policies. When using this make sure to pass all the attributes used in your security policy like `email`, `domain` and `admin` | No (either this or `user_email`) |
-| ttl_seconds | The time to live for the iframe URL                                                                                                                                                             | No (Default: 86400)              |
+| resource | The name of the dashboard to embed                                                                                                                                              | No (if not specified, `navigation` should be set to `true`)                              |
+| navigation | Boolean whether to enable navigation and allow users to navigate to other dashboards (set to `false` if you wish to hard embed and allow access to only a single dashboard)                                                                                                                           | No (either set to `true` or `false`) |
```
ericpgreen2, line 275: The default value is effectively `false`

docs/docs/integrate/embedding.md
```diff
@@ -266,12 +271,46 @@ The API accepts the following parameters:
...
+| theme | If [themes](/build/dashboards/customize.md#changing-themes--colors) are being used, the specific theme to pass to the embedded dashboard                                                                                                                                          | No (set to the name of the theme) |
+| user_id | The id of the user to embed the dashboard for                                                                                                                                                | No (only `user_id`, `user_email`, or `attributes` can be specified) |
```
ericpgreen2, line 277: I'm not following "only `user_id`, `user_email`, or `attributes` can be specified"

ericpgreen2 (approved):
Looks good! 2 small comments.

docs/docs/integrate/embedding.md
AndrewRTsao: I'm not exactly sure I get this. Would you mind elaborating?

docs/docs/integrate/embedding.md
AndrewRTsao: It was mostly based on this [code block](https://github.com/rilldata/rill/blob/7ae6b2960ac5faf82c4f1067f4930743f468d157/proto/rill/admin/v1/api.proto#L646-L653) and the description by Benjamin in #4247. Even in our [current docs](https://arc.net/l/quote/ibilcpyz), we mention that you can either optionally pass in `user_email` or `attributes` so I simply extended this description.

docs/docs/integrate/embedding.md
AndrewRTsao: Actually I think I understand. Navigation defaults to `false` unless otherwise specified.

docs/docs/integrate/embedding.md
AndrewRTsao: Made this a bit more clear in the description

PR #4356, opened 2024-03-15
Admin: Scaffolding for metrics project integration for autoscaling

begelundmuller:
This is a prototype PR that makes the admin service capable of querying custom APIs on the https://github.com/rilldata/rill-cloud-metrics project.

This enables dogfooding Rill for various analytics use cases, such as autoscaling, health checks, quota enforcement, usage-based billing, etc.

Setting a metrics project is optional. If not set, the admin service will still work (to avoid circular dependencies!).
You can provide a metrics project by setting the following environment variable:
```
RILL_ADMIN_METRICS_PROJECT=rilldata/internal-rill-cloud-metrics
```

The PR includes an initial integration with a prototype `autoscaler-slots-recommendations` custom API and adds scaffolding for an autoscaling job that leverages those recommendations (not actually implemented).

Todo for a complete autoscaling solution:
- [ ] Implement autoscaling logic in `admin/worker/run_autoscaler.go`
- [ ] Set the `RILL_ADMIN_METRICS_PROJECT` environment variable in our different environments
- [ ] Test it :)

PR #4376, opened 2024-03-19
feat: update slot by autoscaler

Kavinjsir:
#### Description

Implement `runAutoscaler` to  apply recommended slots for projects

### DIFF
--- a/admin/worker/run_autoscaler.go
+++ b/admin/worker/run_autoscaler.go
@@ -2,10 +2,16 @@ package worker
 
 import (
 	"context"
+	"time"
 
+	"github.com/rilldata/rill/admin/database"
 	"github.com/rilldata/rill/admin/metrics"
 	"go.uber.org/zap"
 )
+
+const legacyRecommendTime = 24 * time.Hour
+
+const scaleThreshold = 0.10
 
 func (w *Worker) runAutoscaler(ctx context.Context) error {
 	recs, ok, err := w.allRecommendations(ctx)
@@ -18,10 +24,44 @@ func (w *Worker) runAutoscaler(ctx context.Context) error {
 	}
 
 	for _, rec := range recs {
-		// TODO: Add autoscaling logic based on the recommendation here.
-		// Consider checking rec.UpdatedOn to avoid making autoscaling decision on stale data!
-		w.logger.Info("autoscaler recommendation", zap.String("project_id", rec.ProjectID), zap.Int("recommended_slots", rec.RecommendedSlots))
-		break
+		// if UpdatedOn is too old, the recommendation is stale and may not be trusted.
+		if time.Since(rec.UpdatedOn) >= legacyRecommendTime {
+			w.logger.Debug("skipping autoscaler: the recommendation is stale", zap.String("project_id", rec.ProjectID), zap.Time("recommendation_updated_on", rec.UpdatedOn))
+			break
+		}
+
+		if rec.RecommendedSlots <= 0 {
+			w.logger.Debug("skipping autoscaler: the recommend slot is <= 0", zap.String("project_id", rec.ProjectID), zap.Int("recommendation_slots", rec.RecommendedSlots))
+			break
+		}
+
+		targetProject, err := w.admin.DB.FindProject(ctx, rec.ProjectID)
+		if err != nil {
+			w.logger.Debug("failed to find project:", zap.String("project_id", rec.ProjectID), zap.Error(err))
+			break
+		}
+
+		if !shouldScale(targetProject.ProdSlots, rec.RecommendedSlots) {
+			w.logger.Debug("skipping autoscaler: target slots are within threshold of original slots",
+				zap.Int("project_slots", targetProject.ProdSlots),
+				zap.Int("recommend_slots", rec.RecommendedSlots),
+				zap.Float64("scale_threshold_percentage", scaleThreshold),
+				zap.String("project_id", targetProject.ID),
+			)
+			break
+		}
+
+		opts := &database.UpdateProjectOptions{
+			ProdSlots: rec.RecommendedSlots,
+		}
+
+		updatedProject, err := w.admin.UpdateProject(ctx, targetProject, opts)
+		if err != nil {
+			w.logger.Error("failed to autoscale:", zap.String("project_id", rec.ProjectID), zap.Error(err))
+			break
+		}
+
+		w.logger.Debug("succeeded in autoscaling:", zap.String("project_id", updatedProject.Name), zap.Int("project_slots", updatedProject.ProdSlots))
 	}
 
 	return nil
@@ -52,4 +92,10 @@ func (w *Worker) allRecommendations(ctx context.Context) ([]metrics.AutoscalerSl
 	}
 
 	return recs, true, nil
+}
+
+func shouldScale(originSlots, recommendSlots int) bool {
+	lowerBound := float64(originSlots) * (1 - scaleThreshold)
+	upperBound := float64(originSlots) * (1 + scaleThreshold)
+	return float64(recommendSlots) < lowerBound || float64(recommendSlots) > upperBound
 }
--- a/admin/worker/worker.go
+++ b/admin/worker/worker.go
@@ -9,6 +9,7 @@ import (
 	"github.com/rilldata/rill/admin"
 	"github.com/rilldata/rill/runtime/pkg/graceful"
 	"github.com/rilldata/rill/runtime/pkg/observability"
+	"github.com/robfig/cron/v3"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/metric"
@@ -56,7 +57,7 @@ func (w *Worker) Run(ctx context.Context) error {
 		return w.schedule(ctx, "upgrade_latest_version_projects", w.upgradeLatestVersionProjects, 6*time.Hour)
 	})
 	group.Go(func() error {
-		return w.schedule(ctx, "run_autoscaler", w.runAutoscaler, 6*time.Hour)
+		return w.scheduleCron(ctx, "run_autoscaler", w.runAutoscaler, "CRON_TZ=America/Los_Angeles 0 0 * * 1") // Run the scaling job at 00:00 on every Monday.
 	})
 
 	// NOTE: Add new scheduled jobs here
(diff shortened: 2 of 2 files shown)


Kavinjsir:
The is a continue work from https://github.com/rilldata/rill/pull/4356, which should partially resolve https://github.com/rilldata/rill-private-issues/issues/143.
It is work in progress now.

admin/worker/run_autoscaler.go
```diff
@@ -18,10 +20,23 @@ func (w *Worker) runAutoscaler(ctx context.Context) error {
...
+		if duration < 24*time.Hour {
+			w.logger.Debug("skipping autoscaler: the project has been scaled recently", zap.String("project_id", rec.ProjectID), zap.Time("project_updated_on", rec.UpdatedOn))
+			break
+		}
+
+		opts := &database.UpdateProjectOptions{
+			ProdSlots: rec.RecommendedSlots,
+		}
+
+		proj, err := w.admin.DB.UpdateProject(ctx, rec.ProjectID, opts)
```
Kavinjsir, line 33: TODO:
Consider using admin client `UpdateProject` method instead of direct DB update since we may also want to:
- trigger restart the project
- enable observability for project update

Kavinjsir (commented):
Hold at this moment to finish TODO in the comment.

admin/worker/run_autoscaler.go
```diff
@@ -18,10 +20,29 @@ func (w *Worker) runAutoscaler(ctx context.Context) error {
...
 	for _, rec := range recs {
-		// TODO: Add autoscaling logic based on the recommendation here.
-		// Consider checking rec.UpdatedOn to avoid making autoscaling decision on stale data!
-		w.logger.Info("autoscaler recommendation", zap.String("project_id", rec.ProjectID), zap.Int("recommended_slots", rec.RecommendedSlots))
-		break
+		duration := time.Since(rec.UpdatedOn)
+		if duration < 24*time.Hour {
+			w.logger.Debug("skipping autoscaler: the project has been scaled recently", zap.String("project_id", rec.ProjectID), zap.Time("project_updated_on", rec.UpdatedOn))
+			break
+		}
```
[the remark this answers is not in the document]
Kavinjsir: Oh, okay, thx for clarifying!

Updating it to:
- check if the update of recommendation was over **7 days**, then skip
- since the autoscaler will be scheduled for a fixed time every week, I'm not going to add logic here for _checking when the slots were most recently updated_.

@begelundmuller  How would you like that?

admin/worker/run_autoscaler.go
```diff
@@ -18,10 +20,29 @@ func (w *Worker) runAutoscaler(ctx context.Context) error {
...
+
+		targetProject, err := w.admin.DB.FindProject(ctx, rec.ProjectID)
+		if err != nil {
+			w.logger.Warn("failed to find project:", zap.String("project_id", rec.ProjectID), zap.Error(err))
```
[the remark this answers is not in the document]
Kavinjsir: Sounds good! Changing to `Debug`

admin/worker/run_autoscaler.go
```diff
@@ -18,10 +20,29 @@ func (w *Worker) runAutoscaler(ctx context.Context) error {
...
+			break
+		}
+
+		opts := &database.UpdateProjectOptions{
+			ProdSlots: rec.RecommendedSlots,
+		}
```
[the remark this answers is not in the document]
Kavinjsir: I think this is a great concern 👍🏼  and have also talked with @kaspersjo :

As the essence to have a careful control here is knowing _**when to scale**_ and _**how to scale**_:
- On one hand, we want to avoid frequent or unnecessary scaling, such as slots from `30 -> 31` or `60 -> 58`.
- On the other hand, we need the scaler to be responsive when "necessary"

@kaspersjo raised the idea to have a fixed number of slots difference (`abs(current_slots - recommend_slots) <= min_slots_scale_limit`), which can be clearer than a percentage threshold.

Well, from a double thought, I'd prefer a percentage threshold (`min_slots_scale_percentage` than a fix number of scale limit `min_slots_scale_limit`:
- Imagine we have two projects with different scales (say `A` with 100 slots and `B` with 5 slots). As we want to reduce unnecessary scaling that leads to downtime, defining a literal `min_slots_scale_limit` is hard:
    - If it is defined small, like `1-20`, it does not help control `A`.
    - And it is hard to find a "bigger" number, say `20`, which maybe fit for `A`, but probably won't make sense for a project which can be scaled from `20` to `40`.
- So there is a balance to maintain, and I propose to define `min_slots_scale_percentage`, like `10%`:
    - It should be useful for bigger projects (slots >= 50)
    - For small projects, it is a tradeoff, but I think it is acceptable, assuming restarting a small project is easier/quicker than restarting a bigger project.

admin/worker/run_autoscaler.go
```diff
@@ -18,10 +20,29 @@ func (w *Worker) runAutoscaler(ctx context.Context) error {
...
+
+		updatedProject, err := w.admin.UpdateProject(ctx, targetProject, opts)
+		if err != nil {
+			w.logger.Warn("failed to autoscale:", zap.String("project_id", rec.ProjectID), zap.Error(err))
```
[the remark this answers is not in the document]
Kavinjsir: Sounds good!

admin/worker/worker.go
```diff
@@ -94,6 +95,35 @@ func (w *Worker) schedule(ctx context.Context, name string, fn func(context.Cont
 	}
 }
 
+func (w *Worker) scheduleCron(ctx context.Context, name string, fn func(context.Context) error, cronExpr string) error {
```
Kavinjsir, line 98: > For this initial autoscaler, since we still haven't solved the downtime issue during auto-scaling, I would also suggest running the autoscaler at a time we know has little traffic, e.g. at 00:00 PST.
> 
> To make that work, you might need to make a small change to the background job scheduler.

@begelundmuller Good catch! So I implemented this `scheduleCron` function to enable job scheduling by `cronjob`, which is actually a suggestion from @kaspersjo .
How would you like this?

begelundmuller: I like this!

admin/worker/worker.go
```diff
@@ -94,6 +95,35 @@ func (w *Worker) schedule(ctx context.Context, name string, fn func(context.Cont
...
+	loc, err := time.LoadLocation("America/Los_Angeles")
+	if err != nil {
+		return err
+	}
```
begelundmuller, line 102: I would recommend removing this (since we consistently use UTC on the server side), and instead passing the time zone directly in the crontab (which is supported in `robfig/cron/v3`). I think the format is: `CRON_TZ=America/Los_Angeles 0 0 * * 1` (but please verify)

admin/worker/worker.go
```diff
@@ -94,6 +95,35 @@ func (w *Worker) schedule(ctx context.Context, name string, fn func(context.Cont
...
+		now := time.Now().In(loc)
+		nextRun := schedule.Next(now)
+		waitDuration := time.Until(nextRun)
+
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-time.After(waitDuration):
+		}
+	}
```
begelundmuller, line 124: I think you need to move the `runJob` call to *after* the `select`. Otherwise, the job would run every time the process restarts, which is not desired for cron-configured jobs.

admin/worker/run_autoscaler.go
```diff
@@ -2,11 +2,23 @@ package worker
...
 import (
 	"context"
+	"time"
 
+	"github.com/rilldata/rill/admin/database"
 	"github.com/rilldata/rill/admin/metrics"
 	"go.uber.org/zap"
 )
 
+const legacyRecommendTime = 7 * 24 * time.Hour
```
begelundmuller, line 12: What do you think of reducing this to 1 day? IMO if the pipeline has been down for several days, I would prefer we don't autoscale until it's back up.

admin/worker/run_autoscaler.go
```diff
@@ -2,11 +2,23 @@ package worker
...
+
+const scaleThreshold = 0.10
+
+func shouldScale(originSlots, recommendSlots int) bool {
+	lowerBound := float64(originSlots) * (1 - scaleThreshold)
+	upperBound := float64(originSlots) * (1 + scaleThreshold)
+	return float64(recommendSlots) < lowerBound || float64(recommendSlots) > upperBound
+}
```
begelundmuller, line 20: nit: put util functions at the end of the file. See the style guide on function ordering: https://github.com/uber-go/guide/blob/master/style.md#function-grouping-and-ordering

admin/worker/run_autoscaler.go
```diff
@@ -18,10 +30,39 @@ func (w *Worker) runAutoscaler(ctx context.Context) error {
...
+			w.logger.Debug("skipping autoscaler: target slots are within threshold of original slots",
+				zap.Int("project_slots", targetProject.ProdSlots),
+				zap.Int("recommend_slots", rec.RecommendedSlots),
+				zap.Float64("scale_threshold_percentage", scaleThreshold),
+				zap.String("project_id", targetProject.ID),
+			)
+			break
+		}
+
+		opts := &database.UpdateProjectOptions{
```
begelundmuller, line 55: Can we add an extra check here for `if rec.RecommendedSlots <= 0` just to be completely safe? (The recommender should always return `>= 1`, but just in case something went wrong there.)

begelundmuller (changes requested):

admin/worker/run_autoscaler.go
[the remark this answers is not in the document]
begelundmuller: I like your approach here. Having a percentage also allows us to scale at the small end, e.g. if a project is very small, we can scale down from 2 slots (default) to 1 slot.

admin/worker/run_autoscaler.go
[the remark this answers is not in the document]
begelundmuller: Makes sense, but I think we should reduce to 1 day (to be conservative and not scale in case of broken data pipelines)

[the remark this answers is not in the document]
Kavinjsir: Agree.

(Resolving this conversation as it is followed up by https://github.com/rilldata/rill/pull/4376#discussion_r1538978689)

admin/worker/worker.go
Kavinjsir: Sounds good! The format you provide here looks good to me, see reference:

https://pkg.go.dev/github.com/robfig/cron/v3#hdr-Usage

https://github.com/robfig/cron/blob/v3.0.1/parser.go#L95

admin/worker/worker.go
Kavinjsir: Oh, that's a good catch!  Just moved the logic based on you suggestion

admin/worker/run_autoscaler.go
Kavinjsir: Yeah, that totally makes sense!

Perhaps in the future we may organize the logic here in a proper place as we are having two places to handle the data logic from this specific custom api.

admin/worker/run_autoscaler.go
Kavinjsir: Sure thing! Just added.

admin/worker/run_autoscaler.go
begelundmuller: Yeah, if you prefer, feel free to refactor this to put it into the custom API if you want (e.g. the custom API could return a `stale` boolean to indicate a stale recommendation).

begelundmuller (approved):
Looks great now!

PR #4379, opened 2024-03-19
Avoid 404s for get resource while modelling

AdityaHegde:
While actively modelling or updating metrics config we need not refetch the resource query since we get the updated resource object in watch-resource API.

This will also avoid 404s when the resource went into an error state because of any change in the editor.


AdityaHegde:
Will revisit this after asset explorer changes

PR #4381, opened 2024-03-19
Runtime: rename source fix for external table storage

k-anshul:
Here is the state of source renames for `external storage enabled` case:

1. Renames were never supposed to work for `external storage enabled`.
2. But worked for some time initially due to another bug which was fixed here : https://github.com/rilldata/rill/pull/3291
post which it stopped working :)


nishantmonu51 (approved):
👍

PR #4382, opened 2024-03-19
Debounce olap list invalidation when renamed

AdityaHegde:
While `renameFrom` is set in the resource. Olap might not be updated at this point. So adding a debouncer to reset the query after 250ms.


ericpgreen2:
Leaving my Slack comment here:

Semantically, it feels like the runtime shouldn’t send the “renamedFrom” event until the table has actually been renamed. So rather than debouncing the client, can the runtime wait to send the event until the event has actually taken place?

PR #4386, opened 2024-03-20
Refactor dashboard state sync with metrics view spec

AdityaHegde:
During bookmark changes we added a layer to add loading of home bookmark. This PR refactors the code to streamline the sync code and generalise for local and cloud.

### DIFF
--- a/web-admin/src/features/bookmarks/selectors.ts
+++ b/web-admin/src/features/bookmarks/selectors.ts
@@ -3,6 +3,7 @@ import {
   type V1Bookmark,
 } from "@rilldata/web-admin/client";
 import { useProjectId } from "@rilldata/web-admin/features/projects/selectors";
+import type { CompoundQueryResult } from "@rilldata/web-common/features/compound-query-result";
 import { getDashboardStateFromUrl } from "@rilldata/web-common/features/dashboards/proto-state/fromProto";
 import {
   useMetricsView,
@@ -103,6 +104,42 @@ export function searchBookmarks(
     personal: bookmarks?.personal.filter(matchName) ?? [],
     shared: bookmarks?.shared.filter(matchName) ?? [],
   };
+}
+
+export function getHomeBookmark(
+  queryClient: QueryClient,
+  instanceId: string,
+  orgName: string,
+  projectName: string,
+  metricsViewName: string,
+): CompoundQueryResult<string> {
+  return derived(
+    getBookmarks(
+      queryClient,
+      instanceId,
+      orgName,
+      projectName,
+      metricsViewName,
+    ),
+    (bookmarks) => {
+      if (bookmarks.isFetching) {
+        return {
+          isFetching: true,
+          error: "",
+        };
+      } else if (bookmarks.isError) {
+        return {
+          isFetching: false,
+          error: bookmarks.error,
+        };
+      }
+      return {
+        isFetching: false,
+        error: "",
+        data: bookmarks.data.home?.resource?.data,
+      };
+    },
+  );
 }
 
 export function getPrettySelectedTimeRange(
--- a/web-admin/src/features/bookmarks/selectors.ts
+++ b/web-admin/src/features/bookmarks/selectors.ts
@@ -106,7 +106,7 @@ export function searchBookmarks(
   };
 }
 
-export function getHomeBookmark(
+export function getHomeBookmarkData(
   queryClient: QueryClient,
   instanceId: string,
   orgName: string,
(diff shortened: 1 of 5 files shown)


web-common/src/features/dashboards/stores/DashboardStateProvider.svelte
```diff
@@ -1,53 +1,27 @@
...
-  $: if (
-    $metricsView.data &&
-    $metricsViewSchema.data &&
-    ($timeRangeQuery.data || !$hasTimeSeries.data)
-  ) {
-    syncDashboardStateLocal();
-  }
-
-  $: ready = metricViewName in $metricsExplorerStore.entities;
+  const dashboardStateSync = createDashboardStateSync(
```
ericpgreen2, line 14: Same comment as above – it's hard to interpret what this store represents

web-common/src/features/dashboards/stores/DashboardStateProvider.svelte
```diff
@@ -1,53 +1,27 @@
...
+    getStateManagers(),
+    writable({
+      isFetching: false,
+      error: "",
+      data: $page.url.searchParams.get("state") ?? "",
+    }),
```
ericpgreen2, line 20: Feels a little hacky. Rather could this `data` object be computed via fallback/default logic inside the `createDashboardStateSync()` function?

ericpgreen2 (approved):
Nice step forward in cleaning this up. Approving, noting:
- I left some code cleanup comments
- I haven't QA'd this – assuming you're confident in it

web-admin/src/features/bookmarks/selectors.ts
```diff
@@ -105,6 +106,42 @@ export function searchBookmarks(
   };
 }
 
+export function getHomeBookmark(
+  queryClient: QueryClient,
+  instanceId: string,
+  orgName: string,
+  projectName: string,
+  metricsViewName: string,
+): CompoundQueryResult<string> {
```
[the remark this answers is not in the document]
AdityaHegde: Let this be `getHomeBookmarkData` that makes it easy to consume in dashboard state sync.

web-admin/src/features/dashboards/DashboardAdminStateProvider.svelte
```diff
@@ -1,74 +1,33 @@
...
-  const metricsViewSchema = createQueryServiceMetricsViewSchema(
-    $runtime.instanceId,
-    metricViewName,
-  );
-
   const queryClient = useQueryClient();
-  $: bookmarks = getBookmarks(
+  $: bookmarks = getHomeBookmark(
     queryClient,
     $runtime?.instanceId,
```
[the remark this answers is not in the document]
AdityaHegde: The global `$runtime` store might not be initialised when this is run, especially when it is loaded in the bookmarks dropdown which is in the header. We cannot derive from an undefined store so it is simpler to add it here with a `?`.

While `getHomeBookmark` can have a layer that depends on runtime lets keep it consistent with `getBookmarks` that it depends on.

web-admin/src/features/dashboards/DashboardAdminStateProvider.svelte
```diff
@@ -1,74 +1,33 @@
...
-    !$bookmarks.isFetching
-  ) {
-    syncDashboardStateLocal();
-  }
-
-  $: ready = metricViewName in $metricsExplorerStore.entities;
+  const dashboardStateSync = createDashboardStateSync(
+    getStateManagers(),
+    bookmarks,
+  );
```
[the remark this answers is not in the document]
AdityaHegde: Added a comment. Lets keep the method name the same `createDashboardStateSync` but the variable can be updated to mean that the dashboard store is initialised.

web-admin/src/features/dashboards/DashboardAdminStateProvider.svelte
```diff
@@ -1,74 +1,33 @@
...
-  const hasTimeSeries = useModelHasTimeSeries(stateManagers);
-  const timeRangeQuery = createTimeRangeSummary(stateManagers);
-  const metricsViewSchema = createQueryServiceMetricsViewSchema(
-    $runtime.instanceId,
-    metricViewName,
-  );
-
   const queryClient = useQueryClient();
-  $: bookmarks = getBookmarks(
+  const homeBookmark = getHomeBookmarkData(
```
ericpgreen2, line 16: Looks like this should be reactive?

AdityaHegde: Since none of those can change during a dashboard being open it can be non reactive. Also we have a key on `metricsViewName` so this will always have the correct store.

PR #4388, opened 2024-03-20
Add rill theme to custom charts

djbarnwal:
First pass at adding Rill theme to custom charts. This might change soon with design input

image

image


ericpgreen2 (approved):
Looks great!

PR #4390, opened 2024-03-20
Refactor: Simplify sidebar layout and related positioning

briangregoryholmes:
This PR simplifies the process by which we animate/layout the sidebar and update the positioning of impacted elements.

- Removes the dependency on a secondary component (WorkspaceContainer) to properly layout a page
- Improves animation performance by moving the slide animations to CSS transition (though the current PreviewTable still causes frame drops in Workspaces)
- Removes a somewhat opaque use of context that was used to enable this previously
- Sets up `WorkspaceContainer` for a future simplification since it is now only used to layout a three panel workspace
- Fixes an issue in Safari (and sometimes Chrome) where the sidebar control button tooltip would persist

### DIFF
--- a/web-common/src/app.css
+++ b/web-common/src/app.css
@@ -236,3 +236,13 @@ body {
 a {
   @apply text-primary-500 hover:text-primary-600 font-medium;
 }
+
+.slide {
+  transition-property: padding-left;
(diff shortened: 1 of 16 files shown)


ericpgreen2 (approved):
So great!

PR #4391, opened 2024-03-20
Feat: Add auto-save toggle to model workspace

### LINKED ISSUE #4301: Manual Trigger for saving models
Refer to more details here - https://www.notion.so/rilldata/Make-Rill-dev-usable-at-scale-21ef9fb859874a86aa8d5e4cb808f830

briangregoryholmes:
This PR adds a manual trigger for saving models and persists this preference in local storage via the WorkspaceLayout store.

Closes #4301

### DIFF
--- a/web-common/src/components/forms/Switch.svelte
+++ b/web-common/src/components/forms/Switch.svelte
@@ -2,12 +2,15 @@
   import { Switch as SwitchPrimitive } from "bits-ui";
   import { cn } from "@rilldata/web-common/lib/shadcn";
 
-  type $$Props = SwitchPrimitive.Props;
+  type $$Props = SwitchPrimitive.Props & {
+    small?: boolean;
+  };
   // type $$Events = SwitchPrimitive.Events;
 
   let className: $$Props["class"] = undefined;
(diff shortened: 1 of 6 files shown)


briangregoryholmes:
@nishantmonu51 @AdityaHegde In the PRD, there's seems to be a differentiation between "Save" and "Profile". I believe we talked about this last week, but can you clarify what that means? Currently, updating the model invalidates the `modelQuery`. When this gets invalidated, the `tableQuery` will refetch. Is the PRD saying that we should disentangle these two queries? "Save" would run the `updateModel` mutation, but the `tableQuery` won't be refetched until you click Profile?

AdityaHegde (approved):

nishantmonu51:
@briangregoryholmes : Yes that was the initial intent, that the profile only works when a user explicitly wants to do that. Doing it as part of save is also fine with dev partition change.

nishantmonu51 (approved):
Just did a quick experience pass, LGTM

nishantmonu51:
@briangregoryholmes : Found 1 issue during experience pass - 
Make changes to model, navigate to other model, choose yes, leave model, It doesn't work

PR #4392, opened 2024-03-20
Rill Developer: Fix dashboard's "Edit model" button

ericpgreen2:
Quick fix for [this bug reported in Slack](https://rilldata.slack.com/archives/CTZ8XBQ85/p1710963629492489).

Not all dashboards have an underlying Model. Some dashboards are built directly on top of a Source or Table resource. This PR hides the "Edit model" menu item when there is no Model to edit.


briangregoryholmes (approved):
Works for me. I realize this predates this PR, but I do find it sort of unintuitive that the selector is called `useDashboards` instead of `useValidDashboards`. I would expect `useDashboards` to return everything.

ericpgreen2:
> Works for me. I realize this predates this PR, but I do find it sort of unintuitive that the selector is called `useDashboards` instead of `useValidDashboards`. I would expect `useDashboards` to return everything.

Totally agree – just changed.

PR #4393, opened 2024-03-21
Refactor common code to use full paths

AdityaHegde:
To replace the left nav with a free form folder structure we need to make sure our code uses paths everywhere instead of mapping name and type to a path. This PR is the 1st step where we update common methods to use paths.


web-common/src/features/entity-management/actions.ts
```diff
@@ -35,24 +34,22 @@ export async function renameFileArtifact(
...
     httpRequestQueue.removeByName(name);
     if (showNotification) {
       notifications.send({ message: `Deleted ${getLabel(type)} ${name}` });
     }
 
     if (get(appScreen)?.name === name) {
-      const route = getRouteFromName(getNextEntityName(names, name), type);
-
-      await goto(route);
+      await goto(getNextEntityName(allPaths, name));
```
ericpgreen2, line 52: When we migrate, navigation to the "next entity name" will be confusing in cases where users do not have a `sources/`, `models/`, `dashboards/` directory structure. So instead, to avoid too much magic in the code, I think we should navigate to the root `/` in all cases.

web-common/src/features/sources/sourceUtils.ts
```diff
@@ -4,7 +4,10 @@ import type {
 } from "@rilldata/web-common/runtime-client";
 import { getFilePathFromNameAndType } from "../entity-management/entity-mappers";
 import { EntityType } from "../entity-management/types";
-import { extractFileExtension, sanitizeEntityName } from "./extract-table-name";
+import {
+  extractFileExtension,
+  sanitizeEntityName,
+} from "web-common/src/features/sources/extract-file-name";
```
ericpgreen2, line 10: This import looks off – shouldn't it include "@rilldata/web-common/..."?

web-common/src/features/entity-management/file-selectors.ts
```diff
@@ -7,6 +7,7 @@ import { createRuntimeServiceListFiles } from "@rilldata/web-common/runtime-clie
 export function useMainEntityFiles(
   instanceId: string,
   prefix: "sources" | "models" | "dashboards" | "charts" | "custom-dashboards",
+  transform = (name: string) => name,
```
ericpgreen2, line 10: Rather than this transform parameter, can we instead just edit the `createRuntimeServiceListFiles` selector function to return file paths not file names?

ericpgreen2 (approved):
Looks pretty good 👍

web-common/src/features/charts/ChartMenuItems.svelte
```diff
@@ -1,20 +1,21 @@
 <script lang="ts">
+  import { getFileAPIPathFromNameAndType } from "@rilldata/web-common/features/entity-management/entity-mappers";
   import NavigationMenuItem from "@rilldata/web-common/layout/navigation/NavigationMenuItem.svelte";
   import { runtime } from "../../runtime-client/runtime-store";
   import { deleteFileArtifact } from "../entity-management/actions";
   import { EntityType } from "../entity-management/types";
-  import { useChartFileNames } from "./selectors";
+  import { useChartRoutes } from "./selectors";
 
   export let chartName: string;
```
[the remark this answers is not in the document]
AdityaHegde: Agree. This PR makes an effort to not touch the components to update such params.

web-common/src/features/entity-management/actions.ts
AdityaHegde: Hmm true. I thought we could navigate to the next file in the folder. But i havent seen any editor do that.

PR #4395, opened 2024-03-21
Runtime: metrics_sql parser 

k-anshul:
The parser transforms input `metrics_sql` into a regular SQL query. In addition to features listed in the [docs](https://docs.rilldata.com/integrate/custom-apis/metrics-sql#supported-sql-features) following features are also supported/changed:

1. No need to specify `AGGREGATE` clause. Can simply do `SELECT dim, agg FROM metrics_view`.
2. No need to specify `GROUP BY` clause if selecting measure and dimension both. `GROUP BY` is implicitly added.
3. `WHERE`, `HAVING` and `ORDER BY` clauses can reference any dimension from the metrics_view and not just the selected dimension.
4. Ability to specify custom aliases.
5. Supports `date_trunc` function.
6. Operations on selected dimensions and measures similar to filter clauses. 

Closes https://github.com/rilldata/rill-private-issues/issues/207


mindspank:
@k-anshul you adding docs or want me to do it?

k-anshul:
@mindspank Haven't started on the docs yet. Could you please take it ?
Thanks.

go.mod
```diff
@@ -65,6 +65,8 @@ require (
 	github.com/marcboeker/go-duckdb v1.5.5
 	github.com/mazznoer/csscolorparser v0.1.3
 	github.com/mitchellh/mapstructure v1.5.0
+	github.com/pingcap/tidb v1.1.0-beta.0.20231124053542-069631e2ecfe
+	github.com/pingcap/tidb/pkg/parser v0.0.0-20231124053542-069631e2ecfe
```
begelundmuller, line 69: These are adding a lot of indirect deps. I wonder if removing the direct `tidb` import might reduce the indirect deps. It seems to only derive from this import:
```
	_ "github.com/pingcap/tidb/pkg/types/parser_driver"
```

Is there any way to avoid that?

runtime/pkg/metricssql/parser.go
```diff
@@ -0,0 +1,607 @@
...
+		dimExprList := maps.Keys(t.dimExprToCol)
+		// sort so that consistent group by sql is produced
+		slices.Sort(dimExprList)
+		for i, expr := range dimExprList {
+			if i != 0 {
+				sb.WriteString(", ")
+			}
+			sb.WriteString(expr)
+		}
+	} // todo :: support group by if only dimensions are selected
```
begelundmuller, line 157: If possible, it would be nice to support this now for both dimensions and measures. Having just a dimension is an easy way to get distinct dimension values. Having just a measure is equivalent to a total query.

runtime/pkg/metricssql/parser.go
```diff
@@ -0,0 +1,607 @@
...
+		return "", err
+	}
+
+	return sb.String(), nil
+}
+
+func (t *transformer) fromQueryForMetricsView(mv *runtimev1.Resource) (string, error) {
+	// Dialect to use for escaping identifiers. Currently hardcoded to DuckDB.
+	// TODO: Make dynamic based on actual dialect of the OLAP connector used by the referenced metrics view.
+	dialect := drivers.DialectDuckDB
```
begelundmuller, line 533: Can this TODO be resolved easily now?

begelundmuller (changes requested):

go.mod
k-anshul: So there is a test driver mentioned in the quick start which is part of `pkg/parser` only. The doc says to use it for tests and use `tidb/parser_driver` for advanced use cases. I see all our test cases are passing when I also use test_driver so may be for now we can use it and in future if we find anything that doesn't work we can make a switch ?

runtime/pkg/metricssql/parser.go
k-anshul: So it is possible to specify just dimension or just measure now also. This is specifically for cases when we want to select just dimensions and group by on one or more dimensions. I can handle that case as well.

k-anshul: Okay I realised this can not be achieved till we allow some aggregate functions over dimensions. 
So now we can already do : `select dim1, dim2 from mv` or `select measure1 from mv` but not `select sum(dim1), dim2 from mv group by dim2`. 
I will also added support for `select distinct col1 [,col2...] from mv` type query as well.
Any other queries that you have in your mind ?

go.mod
k-anshul: https://github.com/pingcap/tidb/blob/master/pkg/parser/docs/quickstart.md#parse-sql-text

runtime/pkg/metricssql/parser.go
begelundmuller: Got it – we don't need (maybe even don't want) to support aggregations over dimensions, so that should be fine

begelundmuller (approved):

PR #4396, opened 2024-03-21
Unified artifact store for storing path to resource map

AdityaHegde:
We had 2 store to save information about a file artifact : `fileArtifactsStore` and `resourcesStore`.
This PR merges the 2 and moves to the class way of storing data.

Tries to parse the file to infer `kind` and `name` if a resource doesnt exist.


web-common/src/features/entity-management/file-artifacts-store.ts
```diff
@@ -1,131 +1,241 @@
...
+      return f;
+    });
+  }
+
+  public updateArtifact(resource: V1Resource) {
+    resource.meta?.filePaths?.map(correctFilePath).forEach((filePath) => {
+      this.artifacts[filePath] ??= new FileArtifact(filePath);
+      this.artifacts[filePath].updateResource(resource);
+    });
+  }
```
ericpgreen2, line 120: I was wondering why these methods are on the `FileArtifactsStore`, rather than on the `FileArtifact` class itself. Then I realized that each update updates not just one artifact, but potentially >1.

So improved naming could be the plural `updateArtifacts` (and will assume `isReconciling` will be updated as a part of that function). Further, `updateFileArtifactsForResource` could be most clear.

web-common/src/features/entity-management/file-artifacts-store.ts
```diff
@@ -1,131 +1,241 @@
...
+
+  public deleteResource(resource: V1Resource) {
```
ericpgreen2, line 122: Naming suggestion: `deleteFileArtifactsForResource`

web-common/src/features/entity-management/file-artifacts-store.ts
```diff
@@ -1,131 +1,241 @@
...
+    filePath: string,
+  ): Readable<V1ResourceName | undefined> {
+    return derived(this.getFileArtifact(filePath), (artifact, set) =>
+      artifact.name.subscribe(set),
+    );
+  }
 
-      if (!errorsForPaths.has(filePath)) {
-        errorsForPaths.set(filePath, []);
+  public getReconcilingItems() {
```
ericpgreen2, line 152: Naming suggestion: `getReconcilingFiles`

web-common/src/features/entity-management/file-artifacts-store.ts
```diff
@@ -1,131 +1,241 @@
...
+    this.files.update((f) => {
+      delete f[filePath];
+      return f;
+    });
+  }
+
+  public setResource(resource: V1Resource) {
+    this.updateArtifact(resource);
+    this.updateReconciling(resource);
+  }
```
ericpgreen2, line 102: Hopefully we can delete this in favor of `updateArtifactsForResource` (see below)

web-common/src/features/entity-management/file-artifacts-store.ts
```diff
@@ -1,131 +1,241 @@
...
+      // Dummy store to not break component code
+      new FileArtifact(filePath),
     );
+  }
 
-    errors.forEach((error) => {
-      const filePath = correctFilePath(error.filePath);
+  public getLastStateUpdatedOn(filePath: string) {
+    return this.artifacts[correctFilePath(filePath)]?.lastStateUpdatedOn;
+  }
```
ericpgreen2, line 142: I'd expect this method to be on the `FileArtifact` class itself, not on the parent store

web-common/src/features/entity-management/file-artifacts-store.ts
```diff
@@ -1,131 +1,241 @@
...
-  subscribe,
-  ...fileArtifactsEntitiesReducers,
-};
-
-export function getIsFileReconcilingStore(file: string) {
-  return derived(
-    fileArtifactsStore,
-    ($store) => $store.entities[file]?.isReconciling,
-  );
+  }
```
ericpgreen2, line 231: I'd also expect all these methods to be on the `FileArtifact` class itself, not on the parent container

web-common/src/features/entity-management/file-artifacts-store.ts
```diff
@@ -1,131 +1,241 @@
...
-      ) {
-        return;
-      }
+  public getResourceNameForFile(
+    filePath: string,
+  ): Readable<V1ResourceName | undefined> {
+    return derived(this.getFileArtifact(filePath), (artifact, set) =>
+      artifact.name.subscribe(set),
+    );
+  }
```
ericpgreen2, line 150: I'm not following why we need a `derived` store just to listen to the `FileArtifact`'s `name` store. Can we instead:
1. Get the given `FileArtifact` from the `FileArtifactsStore`
2. Get the FileArtifact's `name` store
3. Subscribe to the `name` store

ericpgreen2: Then we can remove the `SubStore` stuff?

ericpgreen2 (changes requested):

web-common/src/features/entity-management/file-artifacts-store-new.ts
```diff
@@ -0,0 +1,211 @@
...
+} from "@rilldata/web-common/runtime-client";
+import type { QueryClient } from "@tanstack/svelte-query";
+import { derived, get, type Readable, writable } from "svelte/store";
+
+export class FileArtifact {
+  public readonly path = writable<string>("");
+
+  public readonly name = writable<V1ResourceName | undefined>(undefined);
+
+  public readonly jsonRepresentation = writable<
```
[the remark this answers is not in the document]
AdityaHegde: Stale comment

web-common/src/features/dashboards/DashboardMenuItems.svelte
```diff
@@ -0,0 +1,134 @@
...
+    <Explore slot="icon" />
+    <div class="flex gap-x-2 items-center">
+      Generate chart with AI
+      <WandIcon class="w-3 h-3" />
+    </div>
+    <svelte:fragment slot="description">
+      {#if hasErrors}
+        Dashboard has errors
+      {/if}
+    </svelte:fragment>
```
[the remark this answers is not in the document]
AdityaHegde: Merge issue. It should have the disable as well.

web-common/src/features/entity-management/file-artifacts-store.ts
```diff
@@ -1,131 +1,241 @@
...
+export class FileArtifact {
+  public readonly path: string;
+
+  public readonly name = writable<V1ResourceName | undefined>(undefined);
+
+  /**
+   * Last time the state of the resource `kind/name` was updated
+   * used to make sure we do not have unnecessary refreshes
+   */
+  public lastStateUpdatedOn: string | undefined;
```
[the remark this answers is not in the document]
AdityaHegde: The point of this is to avoid triggering a query to resource. This is set in the resources listener. Writable is not needed since it is not used as a reactive data.

web-common/src/features/entity-management/file-artifacts-store.ts
```diff
@@ -1,131 +1,241 @@
...
+  public readonly files = writable<Record<string, boolean>>({});
+
+  /**
+   * Map of all files and its individual store
+   */
+  private readonly artifacts: Record<string, FileArtifact> = {};
+
+  // Actions
+
+  public async init(instanceId: string) {
```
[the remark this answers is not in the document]
AdityaHegde: Good idea.

web-common/src/features/entity-management/file-artifacts-store.ts
AdityaHegde: It is to handle race condition where `FileArtifact` might not have been added yet. `SubStore` stuff is a generic way of doing this in other places.

web-common/src/features/entity-management/file-artifacts-store.ts
AdityaHegde: We would need to get the file artifact for a path anyway. So it makes sense to add this method here itself.

web-common/src/features/entity-management/file-artifacts-store.ts
```diff
@@ -1,131 +1,241 @@
...
-};
-
-const fileArtifactsEntitiesReducers = {
-  setName(path: string, name: string) {
-    createOrUpdateFileArtifact(path, (entityData) => {
-      entityData.name = name;
+  }
+
+  public updateFile(filePath: string) {
+    if (get(this.files)[filePath]) return;
```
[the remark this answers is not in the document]
AdityaHegde: We would need to make sure to add `FileArtifact` for a path. This method handles that. Since changes to file are returned in a different watch request this is needed.

web-common/src/features/entity-management/file-artifacts-store.ts
```diff
@@ -1,131 +1,241 @@
...
+
+  public deleteResource() {
+    this.name.set(undefined);
+  }
+}
+
+export class FileArtifactsStore {
+  /**
+   * Map of all files and whether it is reconciling or not.
+   * If an entry is present here then there should be on in {@link artifacts} as well
```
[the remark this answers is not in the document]
AdityaHegde: Initially i was thinking of mirroring backend by not having an instance of `FileArtifact` when backend doesnt have a resource for a file. This does make it a little complex in that, we need to read files as well when a resource is in error state.

web-common/src/features/entity-management/file-artifacts-store.ts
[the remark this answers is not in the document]
AdityaHegde: Actually lets keep this as init. Mainly because the instance of fileArtifactStore could be shared. Also because init will be an async method so it will have to be called in the constructor anyway.

web-common/src/features/entity-management/file-artifacts-store.ts
AdityaHegde: It doesnt actually returns files, rather resource names. `getReconcilingResourceNames` makes more sense.

web-common/src/features/entity-management/file-artifacts-store.ts
AdityaHegde: I refactored to instead add an instance in the getter if there isnt one. This will get rid of `derived` here.

web-common/src/features/entity-management/file-artifacts-store.ts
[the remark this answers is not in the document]
ericpgreen2: So would this be an accurate explanation/comment: "We save this data here from the `WatchResources` response to avoid an extra call to `GetResource`"? 

Though we could do better by using `queryClient.setQueryData()` in the `WatchResources` listener, and accessing this state through the TanStack Query cache for the `GetResource` API, right? And the same goes for the `reconciling` store?

web-common/src/features/entity-management/file-artifacts-store.ts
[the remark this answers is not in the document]
ericpgreen2: Ah the async concern makes sense. I wonder if there's anything to learn from how we're [instantiating a feature flag listener in the feature flag store's constructor](https://github.com/rilldata/rill/blob/main/web-common/src/features/feature-flags.ts#L53-L76).

web-common/src/features/entity-management/file-artifacts-store.ts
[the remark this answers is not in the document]
ericpgreen2: Cool, I was originally understanding `FileArtifact` to exist regardless of whether or not an associated Resource existed.

So it looks like we can remove this `files` map now?

web-common/src/layout/RillDeveloperLayout.svelte
```diff
@@ -36,7 +39,7 @@
       commitHash: config.build_commit,
     });
 
-    return resourcesStore.init(config.instance_id);
+    return fileArtifactsStore.init(queryClient, config.instance_id);
```
ericpgreen2, line 42: Can we move this to `web-local/src/routes/(application)/+layout.svelte`? It feels very related to the Watch Flies/Resources clients

ericpgreen2: Further, do we need to formally initialize this store at all? Instead, will the Watch Files/Resources clients return all files/resources when they start-up, and we can populate the store with these responses as they stream in?

web-common/src/lib/getSubStore.ts
ericpgreen2, line 1: I'm not seeing this used anymore, so looks like we can delete it?

web-common/src/features/entity-management/file-artifacts-store.ts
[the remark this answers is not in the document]
AdityaHegde: We wont need continuous reactivity feature flags has, since the updates are already streamed in.

web-common/src/features/entity-management/file-artifacts-store.ts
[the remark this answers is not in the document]
AdityaHegde: Yes, just this variable is not removed yet. Nothing reads or write to it.

web-common/src/features/entity-management/file-artifacts-store.ts
```diff
@@ -1,131 +1,297 @@
...
 
-export function getFileArtifactReconciliationErrors(
-  fileState: FileArtifactsState,
-  fileName: string,
-): V1ReconcileError[] {
-  const path = Object.keys(fileState?.entities)?.find((key) => {
-    return key.endsWith(fileName);
-  });
-  return fileState?.entities?.[path]?.errors;
+export class FileArtifactsStore {
```
ericpgreen2, line 190: Nit: this isn't a store, so can we name it just `FileArtifacts`? And maybe change the file name too to `entity-management/file-artifacts.ts`?

web-common/src/features/entity-management/file-artifacts-store.ts
ericpgreen2: Yeah I get that. It just took me a little while to understand what was going on. The plural `updateArtifacts` is good.

web-common/src/features/entity-management/file-artifacts-store.ts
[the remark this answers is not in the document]
AdityaHegde: Yes, I can add more info stating that it saves data from `WatchResources`.

That would be ideal, but i was facing some issues with consistency. Here is a PR where i tried it: https://github.com/rilldata/rill/pull/4379 . As I mentioned yesterday, we should revisit this once asset explorer is finished.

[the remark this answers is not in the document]
ericpgreen2: Sweet, makes sense. Sounds good.

ericpgreen2 (approved):
Nice cleanup! As discussed, we'll hold off on merging this until after we've cut the `v0.43.0` release.

PR #4397, opened 2024-03-21
Druid Explore Exactify mode

egor-ryashin:


runtime/queries/metricsview_comparison_toplist.go
```diff
@@ -18,21 +19,22 @@ import (
...
+	Limit                 int64                                          `json:"limit,omitempty"`
+	Offset                int64                                          `json:"offset,omitempty"`
+	Sort                  []*runtimev1.MetricsViewComparisonSort         `json:"sort,omitempty"`
+	Where                 *runtimev1.Expression                          `json:"where,omitempty"`
+	Having                *runtimev1.Expression                          `json:"having,omitempty"`
+	Filter                *runtimev1.MetricsViewFilter                   `json:"filter"` // Backwards compatibility
+	Aliases               []*runtimev1.MetricsViewComparisonMeasureAlias `json:"aliases,omitempty"`
+	Exact                 bool                                           `json:"exact"`
+	SecurityAttributes    map[string]any                                 `json:"security_attributes,omitempty"`
+	NoExploreExactifyMode bool                                           `json:"druid_exactify,omitempty"`
```
nishantmonu51, line 37: Can we make it a config read from rill.yaml instead ? with the default value here set to true.

runtime/queries/metricsview_comparison_toplist.go
```diff
@@ -113,12 +115,69 @@ func (q *MetricsViewComparison) Resolve(ctx context.Context, rt *runtime.Runtime
...
+			ttr := q.TimeRange
+			q.TimeRange = q.ComparisonTimeRange
+			err = q.executeToplist(ctx, olap, mv, priority, security)
+			if err != nil {
+				return err
+			}
+
+			q.TimeRange = ttr
+		}
+		q.addDimsAsFilter()
```
nishantmonu51, line 147: We need not execute the query again if the prev query had all the measures already computes and gives us less results  than the limit...
e.g.  e.g cardinality is < 10 and we asked for Top100 then our first query already gave exact results..

runtime/queries/metricsview_comparison_toplist.go
```diff
@@ -113,12 +115,69 @@ func (q *MetricsViewComparison) Resolve(ctx context.Context, rt *runtime.Runtime
...
+
+		if q.Exact {
+			return fmt.Errorf("Cannot calculate Exact when Druid Exactify is requested")
+		}
+
+		// Druid-based `exactify` approach:
+		// The previous query collected a list of dimensions. The following query will filter in those dimensions
+		// to reduce the topN list to fit in memory to exclude discrepancies due incomplete topN tables merged by the previous query
+		if q.isBase() || q.isDeltaComparison() {
+			err = q.executeToplist(ctx, olap, mv, priority, security)
```
nishantmonu51, line 133: list of measures can be optimized here, only compute the measures which are used in sort order. Only do it if we have more than 4 measures.

nishantmonu51 (approved):

runtime/queries/metricsview_comparison_toplist.go
nishantmonu51: this is a good query for all leaderboards, for expanded table it will depend on the number of measures.

runtime/queries/metricsview_comparison_toplist.go
nishantmonu51: rename exactify flag to useApproximateTopN, by default set to false

begelundmuller, line 37: This boolean appears to not be checked anywhere?

runtime/queries/metricsview_comparison_toplist.go
```diff
@@ -112,13 +114,97 @@ func (q *MetricsViewComparison) Resolve(ctx context.Context, rt *runtime.Runtime
 		return err
 	}
 
+	// comparison toplist
 	if !isTimeRangeNil(q.ComparisonTimeRange) {
+		// execute toplist for base and get dim list
+		// create and add filter and execute comprison toplist
+		// remove strict limits in comp toplist sql
+		if drivers.DialectDruid != olap.Dialect() || q.Exact {
```
begelundmuller, line 122: I'm worried about how inlined/nested/scattered this fix is since the comparison logic is already tricky to follow. I'm wondering if it could we moved into a separate branch of logic without altering the existing implementation too much? For example, something like this on line 117:

```go
if olap.Dialect() == drivers.DialectDruid && !q.Exact {
  return q.executeDruidApproximate(ctx, olap, mv, priority, security)
}
```

egor-ryashin:
Benchmarks
```
BenchmarkMetricsViewsToplist_Druid-10                              	      79	  22788884 ns/op	    8774 B/op	     154 allocs/op
BenchmarkMetricsViewsToplist_Druid_Exactify-10                     	      82	  16689934 ns/op	    8746 B/op	     153 allocs/op
BenchmarkMetricsViewsToplist_Druid_high_cardinality-10             	      16	  74560154 ns/op	   14953 B/op	     206 allocs/op
BenchmarkMetricsViewsToplist_Druid_Exactify_high_cardinality-10    	       8	 138951797 ns/op	   19994 B/op	     219 allocs/op
```

runtime/queries/metricsview_comparison_toplist.go
egor-ryashin: done

begelundmuller (approved):
Approving and will address the outstanding comments in a follow up PR

PR #4398, opened 2024-03-21
Refactor: Pivot table cleanup

briangregoryholmes:
WIP


djbarnwal (changes requested):
Seeing some UX issues
- Horizontal scroll has an issue with z-index
- In some cases I am unable to resize back the column after reaching full width

https://github.com/rilldata/rill/assets/4402679/71468494-24ab-4c92-833c-85211cc0caa1

web-common/src/features/dashboards/pivot/PivotTable.svelte
```diff
@@ -124,26 +137,83 @@
       }
     }
   };
+
+  function calculateFirstColumnWidth(firstColumnName: string) {
+    const rows = $pivotDataStore.data;
+
+    const isTimeDimension = firstColumnName.startsWith("__time");
```
djbarnwal, line 144: `__time` is not the default column name. Some of our dashboard usually have that as the key but in practice it could be anything. You would want to use `isTimeDimension` function here

web-common/src/features/dashboards/pivot/PivotTable.svelte
```diff
@@ -124,26 +137,83 @@
...
+
+    // Dates are displayed as shorter values
+    if (isTimeDimension) return MIN_COL_WIDTH;
+
+    const samples = extractSamples(
+      rows.map((row) => row[firstColumnName]),
+    ).filter((v): v is string => typeof v === "string");
+
+    const averageValueLength =
+      samples.reduce((sum, value) => sum + value.length, 0) / rows.length;
```
djbarnwal, line 154: Why not use the max length here and then use `Math.min(MAX_INIT_COL_WIDTH, pixelLenght + 30)`
Since we have a bound here, taking the max would make the best effort to accommodate all values until a certain width.

web-common/src/features/dashboards/pivot/PivotTable.svelte
```diff
@@ -124,26 +137,83 @@
...
               {@const sortDirection = header.column.getIsSorted()}
               <th
+                class:first-column={hasDimension && i === 0}
                 colSpan={header.colSpan}
-                class:with-row-dimension={rows.length > 1}
+                class:with-row-dimension={hasDimension}
+                class="relative"
               >
+                {#if i === 0 && hasDimension && firstColumnWidth !== undefined}
+                  <Resizer
```
djbarnwal, line 208: Tanstack table has an inbuilt resizing prop -
https://tanstack.com/table/v8/docs/guide/column-sizing

Do you think it makes sense to use that for all the columns?

briangregoryholmes: May revisit, but saw frame drops when using the Tanstack table approach.

djbarnwal (changes requested):
The table experience seems very flaky. Some issues captured below - 

- Scrolling is jittery when scrolling back up sometimes
- The totals row is not consistently sticky
- Resizing the first column, resizes the label without resizing the whole column itself.

https://github.com/rilldata/rill/assets/4402679/8c0ae4a1-9c25-4428-8aff-ab8b83ecb204

web-common/src/features/dashboards/pivot/pivot-column-definition.ts
```diff
@@ -180,8 +176,8 @@ export function getColumnDefForPivot(
 
     return {
       label: measure?.label || m,
-      formatter: createMeasureValueFormatter<null | undefined>(measure),
```
djbarnwal, line 183: All table measure formatting across the application uses `createMeasureValueFormatter`. Would be better to stick with it.

web-common/src/features/dashboards/pivot/PivotTable.svelte
```diff
@@ -123,192 +174,201 @@
...
+
+    let middleStartIndex = Math.floor((arr.length - sectionSize) / 2);
+    const middle = arr.slice(middleStartIndex, middleStartIndex + sectionSize);
+
+    const last = arr.slice(-lastSectionSize);
+
+    return [...first, ...middle, ...last];
+  }
+
+  // $: console.log(rows[0].getVisibleCells());
```
djbarnwal, line 222: Stale log

web-common/src/features/dashboards/pivot/PivotExpandableRow.svelte
```diff
@@ -0,0 +1,113 @@
...
+          </div>
+        {/if}
+      </svelte:element>
+    </td>
+  {/if}
+
+  {#each hasDimension ? cells.slice(1) : cells as cell, i (cell.id)}
+    {@const value = cell.getValue()}
+    {@const format = String(cell.column.columnDef.cell)}
+    <td class:border-r={i && (i + 1) % measureCount === 0}>
```
djbarnwal, line 53: There's a line missing on the right of row totals

image

djbarnwal (changes requested):
This is removing/changing some of the existing functionalities in place.

1. Loading cells do not show up when fetching new data. We are using `assembled` as a prop for that. 
2. Formatting UX is different compared to main. For `druid_requests_datasource` dashboard seeing non formatted cells.
3. Changing the way all cells are being rendered. For ex empty characters is not being replaced with `\u00A0` while rendering.

There's a lot to unpack here.

I would prefer if we can keep the changes here to resizing, setting column widths and address the later issues in more depth in separate PRs.

PR #4408, opened 2024-03-22
Updagraded gosnowflake as v1.8.0 has an issue with arrow batches

esevastyanov:
https://github.com/snowflakedb/gosnowflake/pull/1068


nishantmonu51 (approved):

begelundmuller (changes requested):
Looking at this, it seems like there's an error scenario where the `writeGrp` fails early (e.g. because of missing file permissions or a storage limit error), but the `fetchGrp` keeps downloading for a long time. 

Similarly, this PR removes the `ctx` check in the `writeGrp` goroutine, which might lead to a lot of redundant writing if the `fetchGrp` fails. 

I think this can be solved by using a single `errgroup` to coordinate all the goroutines (and having a `if ctx.Err() != nil` check inside the for-loop of the writing goroutine).

begelundmuller (approved):

PR #4414, opened 2024-03-25
Add "Export" button to pivot table

### LINKED ISSUE #4271: Feat: Export in Pivot Tables
https://www.notion.so/rilldata/Export-Reports-in-Pivot-Table-d7d5b6a7cfdd4f698404bf033ddacaeb?pvs=4

ericpgreen2:
Closes #4271

### DIFF
--- a/web-common/src/features/dashboards/pivot/pivot-export.ts
+++ b/web-common/src/features/dashboards/pivot/pivot-export.ts
@@ -1,0 +1,80 @@
+import { getResolvedMeasureFilters } from "@rilldata/web-common/features/dashboards/filters/measure-filters/measure-filter-utils";
+import { sanitiseExpression } from "@rilldata/web-common/features/dashboards/stores/filter-utils";
+import type {
+  V1ExportFormat,
+  V1TimeGrain,
+  createQueryServiceExport,
+} from "@rilldata/web-common/runtime-client";
+import { get } from "svelte/store";
+import { runtime } from "../../../runtime-client/runtime-store";
+import type { StateManagers } from "../state-managers/state-managers";
+import { PivotChipType } from "./types";
+
+export default async function exportPivot({
+  ctx,
+  query,
+  format,
+  timeDimension,
+}: {
+  ctx: StateManagers;
+  query: ReturnType<typeof createQueryServiceExport>;
+  format: V1ExportFormat;
+  timeDimension: string;
+}) {
+  const metricsView = get(ctx.metricsViewName);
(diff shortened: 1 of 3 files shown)


web-common/src/features/dashboards/pivot/pivot-export.ts
```diff
@@ -0,0 +1,78 @@
...
+          instanceId: get(runtime).instanceId,
+          limit: undefined, // the backend handles export limits
+          measures: columns.measure.map((m) => {
+            return {
+              name: m.id,
+            };
+          }),
+          metricsView,
+          offset: "0",
+          pivotOn: columns.dimension.map((d) => d.id),
```
djbarnwal, line 64: Column dimensions can have Time dimension as well, does this statement take care of that case as well?

djbarnwal (changes requested):
Pivot table with dimension column in them don't work for me 

https://github.com/rilldata/rill/assets/4402679/e1680627-ecc2-4f6a-8bc4-9ad32187607a

web-common/src/features/dashboards/pivot/pivot-export.ts
ericpgreen2: Oh, the `pivotOn` parameter only takes an array of strings. It seems like we'll need some Platform support to account for this – need to be able to provide the time grain and time zone in addition to the column ID.

ericpgreen2:
> Pivot table with dimension column in them don't work for me

Ah, thanks for finding. Just opened up this issue for Platform: https://github.com/rilldata/rill/issues/4452

web-common/src/features/dashboards/pivot/pivot-export.ts
ericpgreen2: Oh, updated to account for time dimensions. Though it won't support nested time grains (waiting on https://github.com/rilldata/rill/issues/4416).

ericpgreen2:
> Pivot table with dimension column in them don't work for me

Ah, fixed!

djbarnwal:
Update: There's some bug around how date fields are being handled. This is being fixed here #4473. Not sure if that is the source of data discrepancy but while on the call with Eric saw that is was only an issue for dashboard with Date fields

ericpgreen2:
After merging `main`, I'm still seeing a discrepancy. The export does not include all dimension values that the UI shows.

In the below case, here are values in the UI not present in the export:
- `cooking-therapy.com`
- `pluto.tv`
- `arkadium.com`

image

briangregoryholmes (approved):

Issue #3808, opened 2024-01-09
Dashboard: Date picker is hard to use on small screens


ericpgreen2:
[As described in Slack.](https://rilldata.slack.com/archives/C02T907FEUB/p1704804137011539)

This belongs in a greater epic to make the dashboard responsive.

PR #3824, opened 2024-01-11
remove Litepicker and update TimeRangeSelector to new menu system

briangregoryholmes:
To resolve #3808, I propose removing Litepicker from our codebase. This PR does not yet remove the Litepicker code, but simply replaces its functionality with native date pickers.

Litepicker has been archived as of December 18th and the codebase is no longer supported. Additionally, the structure and API of the library does not mesh well with the rest of our components. For instance, Litepicker does not provide a straightforward way to control where the element renders, a requirement for using it with the ShadCN menus.

I would normally include in this PR a proposal for a replacement date picker, perhaps the one from ShadCN, but the current implementation, wherein the date picker is presented as a submenu, has a number of inherent UX problems that would require input from other team members. The intent of this PR is to fix the underlying usability issue and update the menu.

This PR also solves an issue where the min/max boundaries on the inputs were not functional as min/max are not attributes on the text input type.

PR #4422, opened 2024-03-26
Feat: Super Pill and Range Picker

briangregoryholmes:
Fixes [#4679](https://github.com/rilldata/rill/issues/4679)
PRD: https://www.notion.so/rilldata/Time-Super-pill-b26f581630a14680b05fb5f2cd2d41e6?pvs=4

This PR incorporates a variety of updates related to the new Time Super Pill. It started as a refactor of the calendar picker, the original description of which can be seen below.

This implementation has been put together with the Time Control store refactor in mind, but sticks to the existing state, pulling only a few helpers or types from the new time controls file. As such, a few things that have just been copied verbatim into new locations will be simplified when integrated with the new store.

To Do:

- [x] Add custom range selection to comparison (needs feedback)
- [x] Resolve UI issues with narrower screens (feedback welcome)
- [x] Ensure implementation matches Figma as close as possible
- [x] Move subrange to Zoom button
- [x] Consider solutions to button shifting issue (based on changing range strings)
- [x] Update tests
- [x] Add comparison selector to Time Series view
- [x] Resolve inclusive/exclusive date display issues

Out of Scope:

- [ ] Add zoom functionality from PRD (likely out of scope for this PR)

--------------------------------
This PR is a feature-complete, but WIP proposal (initially started with #3824) that both updates the `TimeRangeSelector` and `TimeComparisonSelector` menus to the new ShadCN components and makes a related and technically necessary change to our date picker implementation. Additionally, there is some simplification being done in preparation for the refactor of the time control state management.

Looking for experience passes around the new UX, technical review (mostly around date handling), design input and general feedback.

- Removes the use of Litepicker
- Adds custom `Month` and `Calendar` components
- Removes the use of domain-less text fields as a secondary form of input
- Adds `Select` components from ShadCN
- Uses dropdown UI elements for month, day and year selection
- Adds the ability to nudge the selection by different time grains
- Fixes a critical bug that would extend the selected range by one day
- Explicitly labels the start and end dates as inclusive/exclusive

Tagging @mindspank and @jkhwu 

Screenshot 2024-03-26 at 1 02 59 AM

Closes: #3808


ericpgreen2:
As mentioned in check-in, first @mindspank will review the UX. After his green light, one of @AdityaHegde / @djbarnwal / myself will do the code review.

ericpgreen2:
Code looks good. I see @jkhwu just finalized the [Time "Superpill" PRD](https://www.notion.so/rilldata/Time-Super-pill-Rearrange-filter-bar-and-selectors-b26f581630a14680b05fb5f2cd2d41e6), so I think her UX feedback will be important.

jkhwu:
Thanks for your patience, here are my thoughts:
Screenshot 2024-04-19 at 7 37 04 PM

1. Those little step buttons at the bottom felt a bit buggy, and they may not be necessary. Sometimes (esp after changing the dates manually), those arrow buttons only changed one of the dates, and once when I tried stepping by year it didn't work for either date. Sometimes the year stepper lets you get to a year much further in the future than the dropdown selectors allow. I'd prefer omitting those buttons for now.

2. When using the calendar layout to click on dates, I found myself frequently entering this state (where the start date was after the end date). Could we simply disallow this from ever happening so I don't trigger a bunch of errors by clicking dates? I propose that instead of entering this state, we simply assume that the earlier selected date is always the start date. Would that be possible? 
Screenshot 2024-04-19 at 7 47 16 PM

3. Could we set the extended calendar panel to not move around? When trying to advance the months using the arrows, I kept closing the panel by accident because the arrow moves around depending on how many days are in the month. Would be nice to keep those arrows fixed in place.
Apr-19-2024 19-55-22

Just a note in response to Eric's comment that this component will remain available even after we implement the time super-pill concept, so would be great to spend the time polishing its usability.

jkhwu:
https://www.notion.so/rilldata/UXQA-Time-superpill-d7ed5d224f4e4a829bf588eb3ee27d82?pvs=4
^ Where @ericpgreen2  and I have started listing things from experience pass

mindspank:
Adding some comments to the same list

mindspank:
@ericpgreen2 @briangregoryholmes Are we splitting timeseries exports into its own issue?

ericpgreen2:
UXQA –

- Given the super pill can turn _on_ a time range comparison, I'd expect to be able to turn _off_ a time range comparison. It feels weird to have to go somewhere else to do the inverse operation, and I initially had trouble finding how to do it. One idea – maybe we shouldn't display "vs [comparison period]" when there's no comparison, and maybe there should be explicit "add time comparison" / "remove time comparison" (icon) buttons. @jkhwu, @mindspank?

- With "All time" selected:
  - I'd expect the "Pan time range back" button to be disabled. image
  - Clicking/enabling the comparison "custom period" doesn't seem to do anything meaningful

ericokuma:
Hi y'all, I've done a Product Review and documented my thoughts here:

https://www.notion.so/rilldata/EO-Product-Review-notes-25f671118e0f4bc7a1bd6073091581ab?pvs=4

jkhwu:
Just reviewed, this looks good to me!

ericpgreen2:
I love the new always-on leaderboard comparisons!

UXQA –

1. I navigate from "Last 7 days" comparing "Previous week" to  "Last 14 days". The compared range has updated to show the previous 14 days / previous period, but the label still says "Previous week".

    image

    image

2. On "Last 12 months", I initially see ["Previous period", "Previous Year"] are available options. After selecting "Previous period", "Previous year" disappears as an option.

    In other cases, if I have this right, it looks like we don't show "Previous period" when we have already have an equivalent "Previous X". E.g. when the time range is "Last 24 Hours", we have "Previous day", and we don't show "Previous period". So, it seems like we should do the same for "Last 12 months" and "Previous year".

3. The Custom Time Range picker isn't very keyboard accessible. It doesn't need to be perfect for this first pass, but I most noticed it when I pressed "tab" to try to move from the Start Date to End Date, the dialog closed on me, and the selection wasn't applied.

4. (Existing behavior) I noticed that Pivot comparisons don't work when a Time dimension is included. Maybe we could be explicit about this and toggle off the Comparison in the Super Pill.

    image

briangregoryholmes:
1. Addressed, though this is technically a behavior change. Default comparison period for `LAST_14_DAYS` was `WEEK` instead of `CONTIGUOUS`. You can see in your screenshot that the comparison interval overlaps the primary interval. This feels wrong to me, though, so I think the behavior change is correct.
2. This is a leap year bug that I can't fully address as part of this PR. It will not be present after I move time controls to the new store/logic.
3. Addressed.
4. Added a workaround for just the date picker, but this looks to be true of all our menus. Tab closes the dropdown. I didn't immediately see a way to disable this using shadcn/Bits APIs.
5. Made these states mutually exclusive for now.
6. Asking Janet/Mike.

ericpgreen2 (approved):
So good! Approving for you to merge at your discretion.

jkhwu:
There are three areas that we want to follow up with soon but that were considered too big of an effort for this PR:

1. Showing comparison lines in multi-line visualizations (when dimension comparisons are on)
2. Showing delta columns in Time Dimension Detail
3. Showing delta columns in Pivot, even when a time pill is being used.

cc @ericokuma and @mindspank for consideration in upcoming sprints.

PR #4425, opened 2024-03-26
Refactor sources to use filePath instead of sourceName

AdityaHegde:
Refactors most sources components to use filePath instead of sourceName.

The navigation components are untouched since they will change with asset explorer.


web-common/src/features/sources/navigation/SourceMenuItems.svelte
```diff
@@ -119,7 +116,7 @@
       return;
     }
     try {
-      await refreshSource(connector, tableName, runtimeInstanceId);
+      await refreshSource(connector, tableName, sourceName, runtimeInstanceId);
```
ericpgreen2, line 119: Here I see `tableName` is passed in as the `filePath` parameter. Is that a mistake?

ericpgreen2 (approved):
Approving, but a couple UX issues to resolve before merging:
1. Renaming a source causes it to disappear on me
image
2. I see the whole inspector turned grayscale. It's not like this on `main`, so maybe pulling in `main` will resolve it.
image

web-common/src/features/sources/navigation/SourceMenuItems.svelte
AdityaHegde: Yes it should be filePath

PR #4427, opened 2024-03-26
Hide the Metrics Inspector for OLAP engines that don't support modeling

ericpgreen2:
Closes https://github.com/rilldata/rill-private-issues/issues/234

The Metrics Inspector is enabled for DuckDB-backed projects:
image

The Metrics Inspector is hidden for ClickHouse & Druid projects:
image


AndrewRTsao:
Would this be configurable or hardcoded?

ericpgreen2:
> Would this be configurable or hardcoded?

We've currently hardcoded that ClickHouse and Druid don't support modeling, so this PR just piggybacks on that setting.

AndrewRTsao:
Sounds good. At some point, there will likely be a need to support modeling using CH, so something to keep in mind (but does not need to be solved for immediately).

nishantmonu51 (approved):

PR #4430, opened 2024-03-26
Highlight JSON and SQL in Charts Editor

djbarnwal:
Highlight JSON and SQL in charts editor.
Using `SQLite` dialect for now for the SQL. We will have to replace this with our Metrics SQL dialect later.

image


web-common/src/components/editor/presets/yamlWithJsonAndSql.ts
```diff
@@ -0,0 +1,38 @@
...
+    blockLiteralCount = 0;
+  }
+  if (!node || !activateOnNodes.has(node.name)) {
+    return null;
+  }
+  blockLiteralCount++;
+
+  if (blockLiteralCount === 1) {
+    return {
+      parser: SQLite.language.parser,
```
ericpgreen2, line 22: We've defined a CodeMirror SQL dialect for DuckDB in `web-common/src/features/models/workspace/Editor.svelte`. You could move it to a common location then import it here?

web-common/src/components/editor/presets/yamlWithJsonAndSql.ts
```diff
@@ -0,0 +1,38 @@
+import { jsonLanguage } from "@codemirror/lang-json";
+import { SQLite } from "@codemirror/lang-sql";
+import { yamlLanguage } from "@codemirror/lang-yaml";
+import { LRLanguage } from "@codemirror/language";
+import { parseMixed } from "@lezer/common";
+
+const activateOnNodes = new Set(["BlockLiteralContent"]);
```
ericpgreen2, line 7: Nit: a comment to explain where this "BlockLiteralContent" terminology is coming from would be helpful. Link to docs?

ericpgreen2 (approved):
This is great!

web-common/src/components/editor/presets/yamlWithJsonAndSql.ts
djbarnwal: There is no documentation around the new YAML parser. It was added couple of months ago. This mostly came through a hit and trial using the new parser.

PR #4442, opened 2024-03-26
feat: move alerts to top bar

mindspank:
This PR:

- Moves Create Alert entrypoint to the top bar
- Removes the entry point from leaderboard
- Dynamic defaults in the Alerts form depending on UI state
- Updates icons in listings

### DIFF
--- a/web-admin/src/features/alerts/CreateAlert.svelte
+++ b/web-admin/src/features/alerts/CreateAlert.svelte
@@ -1,6 +1,5 @@
 <script lang="ts">
   import { onMount } from "svelte";
-  import { page } from "$app/stores";
   import Tooltip from "@rilldata/web-common/components/tooltip/Tooltip.svelte";
   import TooltipContent from "@rilldata/web-common/components/tooltip/TooltipContent.svelte";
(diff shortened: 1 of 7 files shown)


web-admin/src/features/alerts/CreateAlert.svelte
```diff
@@ -16,27 +18,30 @@
   // This ensures Rill Developer doesn't try and fail to import the admin-client.
   let CreateAlertDialog;
   onMount(async () => {
-    CreateAlertDialog = (await import("./CreateAlertDialog.svelte")).default;
+    CreateAlertDialog = (
+      await import(
+        "@rilldata/web-common/features/alerts/CreateAlertDialog.svelte"
+      )
+    ).default;
   });
```
ericpgreen2, line 26: Now that we're moving this entrypoint to a `web-admin`-specific component, we can remove this dynamic/conditional import.

ericpgreen2 (approved):
This is great! 

With this change, I'm seeing we now should be able to move all the alerts code from `web-common` to `web-admin`. Feel free to do that in this PR, else we can do it in a follow-on cleanup PR.

mindspank:
Lets leave it in web-common for now if we decide to surface alerts also for local

PR #4444, opened 2024-03-26
Runtime : Support multiple schemas with one live connection

k-anshul:
1. Added two new optional fields, database and db_schema, in metrics_view definition.
2. Added support for specifying database and schema in information_schema lookup calls and profiling APIs.

TBD : Add similar support for druid.


proto/rill/runtime/v1/queries.proto
```diff
@@ -938,6 +938,8 @@ message TimeSeriesValue {
 message TableCardinalityRequest {
   string instance_id = 1;
   string connector = 4;
+  string database = 5;
+  string db_schema = 6;
```
begelundmuller, line 942: It's missing handling for a bunch of the other analytical APIs – try searching for `string table_name`

proto/rill/runtime/v1/api.proto
```diff
@@ -504,6 +504,8 @@ message UnpackEmptyResponse {}
 message GenerateMetricsViewFileRequest {
   string instance_id = 1 [(validate.rules).string = {pattern: "^[_\\-a-zA-Z0-9]+$"}];
   string connector = 2;
+  string database = 6;
+  string db_schema = 7;
   string table = 3;
```
begelundmuller, line 509: The terminology of `database` and `db_schema` feels off – it seems weird to type out `database` in one case and use a `db` abbreviation in another. What about just `database`, `schema`, and `table`? (It's annoying that `schema` can have multiple meanings, but that's a problem in so many tools that people are hopefully used to it by now. If we really want to disambiguate, an option would be `database_schema`.)

runtime/drivers/clickhouse/information_schema.go
```diff
@@ -18,46 +19,67 @@ func (c *connection) InformationSchema() drivers.InformationSchema {
...
+	if err != nil {
+		return nil, err
+	}
+	defer func() { _ = release() }()
+
+	var databases []string
+	dbs, ok := i.c.config["databases"].(string)
+	if ok {
+		databases = strings.Split(dbs, ",")
+	} else {
```
begelundmuller, line 32: 1. It feels weird to ask the user to explicitly pass a comma-separated list of database names like this. Could we just just always list the contents of all (non-system) databases/schemas?
2. Config properties should be parsed using `mapstructure` and added to the connector spec, not read and cast directly from the map

runtime/drivers/clickhouse/information_schema.go
```diff
@@ -18,46 +19,67 @@ func (c *connection) InformationSchema() drivers.InformationSchema {
...
 }
 
-func (i informationSchema) Lookup(ctx context.Context, name string) (*drivers.Table, error) {
-	q := `
+func (i informationSchema) Lookup(ctx context.Context, db, schema, name string) (*drivers.Table, error) {
+	var q string
+	var args []any
+	// table_catalog and table_schema both means the name of the database in which the table is located in clickhouse.
+	// we use either db or schema arg to set table_schema
+	if db == "" && schema == "" {
```
begelundmuller, line 79: It's weird that ClickHouse returns the same for both, but in our conceptual model, arguably these are just "schemas", not "databases". To prevent them showing up with duplicate prefixes "foo_schema.foo_schema.table_name" in the UI, should we maybe return `database` as the empty string always? If `database` is empty (or indeed, if `schema` is empty), our analytical APIs should be disregard them in the implementation.

runtime/server/generate_metrics_view.go
```diff
@@ -195,11 +197,15 @@ func (s *Server) generateMetricsViewYAMLWithAI(ctx context.Context, instanceID,
 			doc.Connector = connector
 		}
 		doc.Table = tblName
+		doc.Database = tbl.Database
+		doc.Schema = tbl.DatabaseSchema
```
begelundmuller, line 201: We really shouldn't do this for cases where people are using the connector's default database/schema. It bloats the metrics YAML, which is already our most complex YAML artifact.

An option to easily check it might be to include two booleans on `drivers.Table` indicating whether the database and schema respectively is the default one.

begelundmuller (changes requested):

proto/rill/runtime/v1/queries.proto
k-anshul: I only added support for `Tablewide profiling APIs` bcz those are the ones supposed to be used for external tables. Do we want this support in other profiling APIs as well ?

proto/rill/runtime/v1/api.proto
k-anshul: I actually kept it just `schema` earlier but later changed it since some of the API responses already had a schema field and thought to keep same across request and response. `database_schema` felt too verbose but I can keep it to reduce the weirdness.

runtime/drivers/clickhouse/information_schema.go
k-anshul: 1. I felt it gives some control to the user. There will always be cases when user passes a database in their connection string and would just be interested in looking at the tables of the passed database vs all tables in all databases. May be we can keep it a boolean config like `scan_all_databases` with default value as `true`?
2. Ack.

runtime/drivers/clickhouse/information_schema.go
k-anshul: In clickhouse's terminology they are `database`. For example the SQL is `SHOW databases`, `current_database()` and `SHOW schemas` and `current_schema()`.  
> To prevent them showing up with duplicate prefixes

Yeah I was just returning the `database` field and empty `schema` field in information_schema calls. May be I can stop reading `schema` field completely in this lookup call ?

proto/rill/runtime/v1/queries.proto
k-anshul: Added support in other APIs as well for future use cases.

runtime/drivers/clickhouse/information_schema.go
begelundmuller: A boolean would be better, but I would suggest not doing it at all for now and waiting until someone requests it. We may also want to handle it differently, e.g. by collapsing by db/schema name in the UI instead of pushing it as a config key.

runtime/drivers/clickhouse/information_schema.go
begelundmuller: Okay maybe I'm missing something. Usually, you can run queries like `select * from table` (uses default DB and schema), `select * from schema.table` (uses default DB), and `select * from db.schema.table`. Are you saying in Clickhouse, you can only do `select * from table` (uses default db/schema) and `select * from db.db.table`?

Because if you can do `select * from db.table` in ClickHouse, then it sounds like that's actually a schema in our terminology (and SQL terminology), and I would argue we should then map the "database" in ClickHouse to the `schema` field in our generic output.

k-anshul: Yes we can do `select * from db.table` and `select * from table` and not `select * from db.db.table`. Sure we can map it to schema.

runtime/server/generate_metrics_view.go
```diff
@@ -150,7 +152,9 @@ type generateMetricsViewYAMLWithres struct {
 
 // generateMetricsViewYAMLWithAI attempts to generate a metrics view YAML definition from a table schema using AI.
 // It validates that the result is a valid metrics view. Due to the unpredictable nature of AI (and chance of downtime), this function may error non-deterministically.
-func (s *Server) generateMetricsViewYAMLWithAI(ctx context.Context, instanceID, dialect, connector, tblName string, isDefaultConnector, isModel bool, schema *runtimev1.StructType) (*generateMetricsViewYAMLWithres, error) {
+func (s *Server) generateMetricsViewYAMLWithAI(ctx context.Context, instanceID, dialect, connector string, tbl *drivers.Table, isDefaultConnector, isModel bool) (*generateMetricsViewYAMLWithres, error) {
+	schema := tbl.Schema
+	tblName := tbl.Name
```
begelundmuller, line 157: Remove these shorthand variables, they save very few chars

runtime/validate.go
```diff
@@ -152,3 +152,17 @@ func safeSQLName(name string) string {
 	}
 	return fmt.Sprintf("\"%s\"", strings.ReplaceAll(name, "\"", "\"\""))
 }
+
+func fullyQualifiedTableName(t *drivers.Table) string {
```
begelundmuller, line 156: Align naming with the other place this function exists. Actually, consider using `olap.Dialect().EscapeTable(t.Database, t.DatabaseSchema, t.Name)`.

begelundmuller (changes requested):

runtime/queries/table_head.go
```diff
@@ -13,11 +13,13 @@ import (
...
 type TableHead struct {
-	Connector string
-	TableName string
-	Limit     int
-	Result    []*structpb.Struct
-	Schema    *runtimev1.StructType
+	Connector    string
+	DatabaseName string
+	SchemaName   string
+	TableName    string
```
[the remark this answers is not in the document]
k-anshul: Thanks for pointing. These were added initially so didn't notice.

runtime/drivers/druid/olap.go
```diff
@@ -245,10 +245,12 @@ func (i informationSchema) scanTables(rows *sqlx.Rows) ([]*drivers.Table, error)
...
-				Database:       database,
-				DatabaseSchema: schema,
-				Name:           name,
-				Schema:         &runtimev1.StructType{},
+				Database:                database,
+				DatabaseSchema:          schema,
+				Name:                    name,
+				Schema:                  &runtimev1.StructType{},
+				IsDefaultDatabase:       true,
+				IsDefaultDatabaseSchema: true,
```
begelundmuller, line 253: Same:
```suggestion
				Database:                database,
				DatabaseSchema:          schema,
				IsDefaultDatabase:       true,
				IsDefaultDatabaseSchema: true,
				Name:                    name,
				Schema:                  &runtimev1.StructType{},
```

k-anshul:
This requires changes on UI to handle database and schema field.
Moving the PR to draft state till UI changes are ready.

begelundmuller (approved):

PR #4462, opened 2024-03-28
Show a warning indicator for Tables with unsupported data types

ericpgreen2:
Closes https://github.com/rilldata/rill-private-issues/issues/245

image

image


nishantmonu51 (commented):
Can we list the names of unsupported columns and their DB types in the message ?

ericpgreen2:
> Can we list the names of unsupported columns and their DB types in the message ?

Yep, added.

k-anshul (approved):
LGTM from experience pass POV

PR #4465, opened 2024-03-28
Fix: Perpetual loading state and infinite loop query

briangregoryholmes:
This PR is a fix for an issue where dashboards would fall into a perpetual "loading" state when viewing as a mock user with a limited security policy.

This was solved by altering the `createDashboardStateSync` function to return a greater specificity of state than a simple ready/not ready boolean.

As such, even when the dashboard store has failed to load data properly (which would be the case when a user does not have appropriate access), the Dashboard component will still render and handle the `mockUserHasNoAccess` conditional. This is somewhat of a stopgap solution for what should eventually be a more substantial reorganization of how this works.

Because the conditional rendering block doesn't exist in `DashboardStateProvider` in the same way, a conditional had to be added to the `createTimeRangeSummary` query that disables it when `metricsView` has errors. Without this, the `DashboardURLStateProvider` component would mount and it's use of `useDashboardUrlSync` would cause the parent component to fall back into a loading state, which would loop infinitely as it mounted/unmounted. Again, this requires a more significant reorganization. 

Generally, my view is that these are not actually providers and so should not be conditionally blocking the rendering of components further down the tree. The `Dashboard` and its children should be responsible for their own loading states.

### DIFF
--- a/web-common/src/runtime-client/invalidation.ts
+++ b/web-common/src/runtime-client/invalidation.ts
@@ -60,6 +60,7 @@ export async function invalidateAllMetricsViews(
 ) {
   // First, refetch the resource entries (which returns the available dimensions and measures)
   await queryClient.refetchQueries({
+    type: "active",
     predicate: (query) =>
       typeof query.queryKey[0] === "string" &&
       query.queryKey[0].startsWith(`/v1/instances/${instanceId}/resource`),
--- a/web-common/src/features/dashboards/selectors/index.ts
+++ b/web-common/src/features/dashboards/selectors/index.ts
@@ -19,7 +19,6 @@ import type {
 } from "@tanstack/svelte-query";
 import { Readable, derived } from "svelte/store";
 import type { StateManagers } from "../state-managers/state-managers";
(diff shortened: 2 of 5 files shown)


web-common/src/runtime-client/invalidation.ts
```diff
@@ -60,6 +60,7 @@ export async function invalidateAllMetricsViews(
 ) {
   // First, refetch the resource entries (which returns the available dimensions and measures)
   await queryClient.refetchQueries({
+    type: "active",
```
AdityaHegde, line 63: Does this not lead to stale data in dashboards not in view?

AdityaHegde (approved):
Just left a question

web-common/src/runtime-client/invalidation.ts
briangregoryholmes: Yes and no. The only time this function is called (directly or not) is in the context of updating the user. Viewing as a user is specific to a dashboard, not a project. So, yes, only the currently active queries for the dashboard you're looking at are refetched. The data of other dashboards is never refetched/invalidated/etc, so it's not so much "stale" as it is unaltered.

PR #4468, opened 2024-03-28
Fix: slot autoscaler

Kavinjsir:
#### Description

Fix some issues of slot scaler

#### Motivation

Partially resolves: https://github.com/rilldata/rill-private-issues/issues/143

### DIFF
--- a/.github/workflows/go-lint.yml
+++ b/.github/workflows/go-lint.yml
@@ -22,7 +22,7 @@ jobs:
         uses: golangci/golangci-lint-action@v3
         with:
           # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
-          version: v1.56.0
+          version: v1.57.2
 
           # Optional: working directory, useful for monorepos
           # working-directory: somedir
--- a/admin/metrics/client.go
+++ b/admin/metrics/client.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"net/url"
 	"path"
+	"strings"
 	"time"
 )
 
@@ -38,7 +39,18 @@ type AutoscalerSlotsRecommendation struct {
 // AutoscalerSlotsRecommendations invokes the "autoscaler-slots-recommendations" API endpoint to get a list of recommendations for the number of slots to use for projects.
 func (c *Client) AutoscalerSlotsRecommendations(ctx context.Context, limit, offset int) ([]AutoscalerSlotsRecommendation, error) {
 	// Create the URL for the request
-	uri, err := url.Parse(c.RuntimeHost)
+	var runtimeHost string
+
+	// In production, the REST and gRPC endpoints are the same, but in development, they're served on different ports.
+	// TODO: move to http and grpc to the same c.RuntimeHost for local development.
+	// Until we make that change, this is a convenient hack for local development (assumes REST on port 8081).
+	if strings.Contains(c.RuntimeHost, "localhost") {
+		runtimeHost = "http://localhost:8081"
+	} else {
+		runtimeHost = c.RuntimeHost
+	}
+
+	uri, err := url.Parse(runtimeHost)
 	if err != nil {
 		return nil, err
 	}
--- a/admin/worker/run_autoscaler.go
+++ b/admin/worker/run_autoscaler.go
@@ -16,6 +16,7 @@ const scaleThreshold = 0.10
 func (w *Worker) runAutoscaler(ctx context.Context) error {
 	recs, ok, err := w.allRecommendations(ctx)
 	if err != nil {
+		w.logger.Error("failed to autoscale: unable to fetch recommended slots", zap.Error(err))
 		return err
 	}
 	if !ok {
@@ -27,18 +28,18 @@ func (w *Worker) runAutoscaler(ctx context.Context) error {
 		// if UpdatedOn is too old, the recommendation is stale and may not be trusted.
 		if time.Since(rec.UpdatedOn) >= legacyRecommendTime {
 			w.logger.Debug("skipping autoscaler: the recommendation is stale", zap.String("project_id", rec.ProjectID), zap.Time("recommendation_updated_on", rec.UpdatedOn))
-			break
+			continue
 		}
 
 		if rec.RecommendedSlots <= 0 {
 			w.logger.Debug("skipping autoscaler: the recommend slot is <= 0", zap.String("project_id", rec.ProjectID), zap.Int("recommendation_slots", rec.RecommendedSlots))
-			break
+			continue
 		}
 
 		targetProject, err := w.admin.DB.FindProject(ctx, rec.ProjectID)
 		if err != nil {
 			w.logger.Debug("failed to find project:", zap.String("project_id", rec.ProjectID), zap.Error(err))
-			break
+			continue
 		}
 
 		if !shouldScale(targetProject.ProdSlots, rec.RecommendedSlots) {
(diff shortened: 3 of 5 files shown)


admin/worker/worker.go
```diff
@@ -110,9 +112,7 @@ func (w *Worker) scheduleCron(ctx context.Context, name string, fn func(context.
 			return nil
 		case <-time.After(waitDuration):
 			err := w.runJob(ctx, name, fn)
-			if err != nil {
-				return err
```
Kavinjsir, line 114: I think it better not to exit the admin process when autoscaler is not available.

admin/worker/run_autoscaler.go
```diff
@@ -61,7 +62,7 @@ func (w *Worker) runAutoscaler(ctx context.Context) error {
 			break
 		}
 
-		w.logger.Debug("succeeded in autoscaling:", zap.String("project_id", updatedProject.Name), zap.Int("project_slots", updatedProject.ProdSlots))
+		w.logger.Info("succeeded in autoscaling:", zap.String("project_id", updatedProject.Name), zap.Int("project_slots", updatedProject.ProdSlots))
```
Kavinjsir, line 65: I think it better to leave a log line production env when slots for a project is updated.

admin/worker/run_autoscaler.go
```diff
@@ -27,18 +28,18 @@ func (w *Worker) runAutoscaler(ctx context.Context) error {
 		// if UpdatedOn is too old, the recommendation is stale and may not be trusted.
 		if time.Since(rec.UpdatedOn) >= legacyRecommendTime {
 			w.logger.Debug("skipping autoscaler: the recommendation is stale", zap.String("project_id", rec.ProjectID), zap.Time("recommendation_updated_on", rec.UpdatedOn))
-			break
+			continue
```
Kavinjsir, line 31: Replace `break` with `continue` to allow the loop goes through all projects mentioned by the recommendation.

admin/worker/run_autoscaler.go
```diff
@@ -48,20 +49,30 @@ func (w *Worker) runAutoscaler(ctx context.Context) error {
...
+			Provisioner:          targetProject.Provisioner,
+			Annotations:          targetProject.Annotations,
+		})
 
-		updatedProject, err := w.admin.UpdateProject(ctx, targetProject, opts)
 		if err != nil {
 			w.logger.Error("failed to autoscale:", zap.String("project_id", rec.ProjectID), zap.Error(err))
-			break
+		} else {
+			w.logger.Info("succeeded in autoscaling:", zap.String("project_id", updatedProject.Name), zap.Int("project_slots", updatedProject.ProdSlots))
```
Kavinjsir, line 74: Switching from `Debug` to `Info` here as I think it better leave logs for autoscaling successful events

admin/worker/worker.go
```diff
@@ -111,7 +113,7 @@ func (w *Worker) scheduleCron(ctx context.Context, name string, fn func(context.
 		case <-time.After(waitDuration):
 			err := w.runJob(ctx, name, fn)
 			if err != nil {
-				return err
+				w.logger.Error("Failed to run the cronjob", zap.String("cronjob_name", name), zap.Error(err))
```
Kavinjsir, line 116: The reason to update the code here is:
When the autoscaling work fails, I think it better to AVOID exiting the admin process.

---

Assuming the unavailability of autoscaler is:
- not harmful to the main process
- can be temporarily unavailable.

admin/metrics/client.go
```diff
@@ -38,7 +39,13 @@ type AutoscalerSlotsRecommendation struct {
...
 func (c *Client) AutoscalerSlotsRecommendations(ctx context.Context, limit, offset int) ([]AutoscalerSlotsRecommendation, error) {
 	// Create the URL for the request
-	uri, err := url.Parse(c.RuntimeHost)
+	var runtimeHost string
+	if strings.Contains(c.RuntimeHost, "localhost") {
+		runtimeHost = "http://localhost:8081"
+	} else {
+		runtimeHost = c.RuntimeHost
+	}
+	uri, err := url.Parse(runtimeHost)
```
Kavinjsir, line 48: @begelundmuller this is an update based on the slack thread you left at:
https://rilldata.slack.com/archives/C06PPCTGZ3P/p1711545926984779?thread_ts=1711536145.216599&cid=C06PPCTGZ3P

Which is only an attempt for local debugging, as we may mostly practicing it in `test` or `staging` env, I am also good if you think it unnecessary.

begelundmuller: Looks correct, though please add a comment/TODO about why it's needed and when it can be removed (similar to the other place I linked you to)

admin/worker/run_autoscaler.go
```diff
@@ -48,20 +49,30 @@ func (w *Worker) runAutoscaler(ctx context.Context) error {
...
-			break
+			continue
 		}
 
-		opts := &database.UpdateProjectOptions{
-			ProdSlots: rec.RecommendedSlots,
-		}
+		updatedProject, err := w.admin.UpdateProject(ctx, targetProject, &database.UpdateProjectOptions{
+			ProdSlots:            rec.RecommendedSlots,
+			Name:                 targetProject.Name,
```
begelundmuller, line 57: Can you move `ProdSlots` in the same place in the list as it appears in the struct definition? It makes future refactors easier :)

admin/worker/run_autoscaler.go
```diff
@@ -48,20 +49,30 @@ func (w *Worker) runAutoscaler(ctx context.Context) error {
...
+			ProdTTLSeconds:       targetProject.ProdTTLSeconds,
+			Provisioner:          targetProject.Provisioner,
+			Annotations:          targetProject.Annotations,
+		})
 
-		updatedProject, err := w.admin.UpdateProject(ctx, targetProject, opts)
 		if err != nil {
 			w.logger.Error("failed to autoscale:", zap.String("project_id", rec.ProjectID), zap.Error(err))
-			break
+		} else {
```
begelundmuller, line 73: Can put `continue` in the `if err != nil {...}` case, and then unnest the else case.

See "Reduce nesting" in this style guide: https://github.com/uber-go/guide/blob/master/style.md#reduce-nesting

admin/worker/run_autoscaler.go
```diff
@@ -88,7 +99,11 @@ func (w *Worker) allRecommendations(ctx context.Context) ([]metrics.AutoscalerSl
 			break
 		}
 		recs = append(recs, batch...)
-		offset += limit
+		if len(batch) >= limit {
+			offset += limit
+		} else {
+			break
+		}
```
begelundmuller, line 106: Here you can also reduce nesting by changing to something like:
```suggestion
		if len(batch) < limit {
			break
		}
		offset += limit
```

admin/worker/worker.go
begelundmuller: Seems reasonable – can you make the same change in the old `func (w *Worker) schedule` function?

cli/cmd/admin/start.go
```diff
@@ -99,7 +99,8 @@ func StartCmd(ch *cmdutil.Helper) *cobra.Command {
 
 			// Init logger
 			cfg := zap.NewProductionConfig()
-			cfg.Level.SetLevel(conf.LogLevel)
+			// TODO: Set DebugLevel temporarily for debuggin purpose, revert to InfoLevel before merge.
+			cfg.Level.SetLevel(zapcore.DebugLevel)
```
begelundmuller, line 103: Needs to be reverted

admin/worker/worker.go
```diff
@@ -57,7 +57,9 @@ func (w *Worker) Run(ctx context.Context) error {
 		return w.schedule(ctx, "upgrade_latest_version_projects", w.upgradeLatestVersionProjects, 6*time.Hour)
 	})
 	group.Go(func() error {
-		return w.scheduleCron(ctx, "run_autoscaler", w.runAutoscaler, "CRON_TZ=America/Los_Angeles 0 0 * * 1") // Run the scaling job at 00:00 on every Monday.
+		// return w.scheduleCron(ctx, "run_autoscaler", w.runAutoscaler, "CRON_TZ=America/Los_Angeles 0 0 * * 1") // Run the scaling job at 00:00 on every Monday.
+		// TODO: The following line is only for testing purpose to run scaler frequently for debugging, remove it before merge.
+		return w.scheduleCron(ctx, "run_autoscaler", w.runAutoscaler, "CRON_TZ=America/Los_Angeles * * * * *") // Run the scaling job for every minite.
```
begelundmuller, line 62: Needs to be reverted before merging

begelundmuller (changes requested):

admin/worker/run_autoscaler.go
Kavinjsir: Sure 👍🏼

admin/worker/run_autoscaler.go
Kavinjsir: 👍🏼

admin/worker/worker.go
Kavinjsir: Sure thing!

.github/workflows/go-lint.yml
```diff
@@ -22,7 +22,7 @@ jobs:
         uses: golangci/golangci-lint-action@v3
         with:
           # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
-          version: v1.56.0
+          version: v1.57.2
```
Kavinjsir, line 25: The reason for bumping the version here is to resolve an unexpected lint issue in CI:
https://github.com/rilldata/rill/actions/runs/8511596615

Removed `timeCmpSimplify` also because the latest release has dropped this function.

cli/cmd/admin/start.go
```diff
@@ -99,7 +99,7 @@ func StartCmd(ch *cmdutil.Helper) *cobra.Command {
 
 			// Init logger
 			cfg := zap.NewProductionConfig()
-			cfg.Level.SetLevel(conf.LogLevel)
+			cfg.Level.SetLevel(zapcore.InfoLevel)
```
begelundmuller, line 102: I think this needs to be reverted?

begelundmuller (changes requested):
All the other changes look great and seem ready to merge!

cli/cmd/admin/start.go
Kavinjsir: My bad.. using the wrong variable, just reverted back, thx for pointing out!

begelundmuller (approved):

PR #4469, opened 2024-03-28
fix: event error in cloud

mindspank:
behaviourEvents are not defined in rill cloud
We should follow up on how to unify events across cloud and local

### DIFF
--- a/web-common/src/features/dashboards/tab-bar/TabBar.svelte
+++ b/web-common/src/features/dashboards/tab-bar/TabBar.svelte
@@ -47,15 +47,17 @@
       selectedTab.label === "Pivot",
     );
 
-    behaviourEvent.fireNavigationEvent(
-      $metricsViewName,
-      BehaviourEventMedium.Tab,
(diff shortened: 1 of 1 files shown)


web-common/src/features/dashboards/tab-bar/TabBar.svelte
```diff
@@ -47,15 +47,17 @@
...
-    behaviourEvent.fireNavigationEvent(
-      $metricsViewName,
-      BehaviourEventMedium.Tab,
-      MetricsEventSpace.Workspace,
-      MetricsEventScreenName.Dashboard,
-      selectedTab.label === "Pivot"
-        ? MetricsEventScreenName.Pivot
-        : MetricsEventScreenName.Explore,
-    );
+    if (behaviourEvent) {
```
ericpgreen2, line 50: @AdityaHegde why do you think this is necessary? I see `web-admin`'s `initCloudMetrics()` function creates the `BehaviourEventFactory`

PR #4470, opened 2024-03-29
Refac: Move totals row out of data

djbarnwal:
Separate the totals data row rather than clubbing it with the overall pivot data. 
Also makes sure that totals data row is always present when needed ensuring fixed array size while making changes to pivot/scrolling.


briangregoryholmes:
Are you able to return the columns in the same order as the table? I think currently the first few columns are returned at the end of the object.

briangregoryholmes (approved):

djbarnwal:
> Are you able to return the columns in the same order as the table? I think currently the first few columns are returned at the end of the object.

Hard to guarantee order of keys in the objects. Surely something which can be done to make it work but sounds very error prone as order of objects is not guaranteed in JS and code changes upstream can affect the order.

PR #4471, opened 2024-03-29
Refac: Replace TDD multiple timeseries API with Aggregation API

djbarnwal:
Current implementation of TDD has 1 query for each dimension value. This leads to huge number of reactive updates and a slow table.

This PR batches the call to 50 values at once. So instead of 250 timeseries queries we have 5 aggregation queries (configurable as needed). This helps us make the TDD table many folds faster.


web-common/src/features/dashboards/time-series/utils.ts
```diff
@@ -134,17 +136,80 @@ export function getOrderedStartEnd(start: Date, stop: Date) {
...
+}
+
+export function transformAggregateDimensionData(
+  timeDimension: string,
+  dimensionName: string,
+  measures: string[],
+  dimensionValues: string[],
+  timeSeriesData: V1TimeSeriesValue[],
+  response: V1MetricsViewAggregationResponseDataItem[],
+) {
```
ericpgreen2, line 157: Mind being more clear about what transformation this is? Maybe either add a return type, code comment, or a more descriptive function name.

ericpgreen2 (commented):
I tested this on the OpenRTB Auctions dashboard and I'm seeing "no data" everywhere. E.g.:

image

image

djbarnwal:
The table was not working because of a prior issue with the API response https://rilldata.slack.com/archives/C05PTA6DK0S/p1711622792070359
That has been addressed in #4473. Merged main and it works fine now

ericpgreen2 (approved):
So much better UX!

PR #4472, opened 2024-03-29
Runtime: clickhouse version check and compatibly queries

k-anshul:
Minimum supported version : `22.7`


k-anshul:
I ran unit tests with `23` and `22.7`. Need to see what should be done in longer term.

begelundmuller (approved):
LGTM

AndrewRTsao:
Pushed this info to our docs (under Reference > OLAP Engines > ClickHouse). If this ever changes, we should also make sure to update our docs accordingly.

PR #4477, opened 2024-03-31
Improve reconcilation performance for external dashboards

nishantmonu51:
Improves reconcilation time on druid based dashboard having 45 dims and 30 measures from 50 sec to 4 sec.


pjain1:
@nishantmonu51 main change is its doing all dims and measures in one go and if there are any errors then go one by one on them to find actual failing one ?

pjain1 (approved):

nishantmonu51:
Yes, its trying to improve reconcilation time by doing the check in one shot rather than one by one. Prepares one query instead of 70.

PR #4479, opened 2024-04-01
Runtime: Add query logging flags for Druid and ClickHouse

begelundmuller:
We already support `--var connector.duckdb.log_queries=true` to enable query logging for Druid. This PR adds support for:
- Passing `--var connector.druid.log_queries=true` for Druid query logging
- Passing `--var connector.clickhouse.log_queries=true` for ClickHouse query logging

### DIFF
--- a/runtime/drivers/clickhouse/olap.go
+++ b/runtime/drivers/clickhouse/olap.go
@@ -55,6 +55,11 @@ func (c *connection) WithConnection(ctx context.Context, priority int, longRunni
 }
 
 func (c *connection) Exec(ctx context.Context, stmt *drivers.Statement) error {
+	// Log query if enabled (usually disabled)
+	if c.config.LogQueries {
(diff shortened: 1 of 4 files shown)


runtime/drivers/clickhouse/olap.go
```diff
@@ -76,6 +76,11 @@ func (c *connection) Exec(ctx context.Context, stmt *drivers.Statement) error {
 }
 
 func (c *connection) Execute(ctx context.Context, stmt *drivers.Statement) (res *drivers.Result, outErr error) {
+	// Log query if enabled (usually disabled)
+	if c.config.LogQueries {
```
k-anshul, line 80: Also needs this change in `Exec` method. Unlike duckDB and druid, clickhouse's `Exec` method does not internally call `Execute` because the underlying driver errors out if running `Query` method for a SQL that does not return any rows.

begelundmuller: Good catch, fixed!

PR #4488, opened 2024-04-02
Support repo creation as part of rill deploy by requesting Github write access

k-anshul:
Changes in Github App Permissions for Repository : 

1. `New`

image

2. Updated from `Read only` to `Read-Write`
 
image


cli/cmd/deploy/deploy.go
```diff
@@ -358,6 +388,169 @@ func loginWithTelemetry(ctx context.Context, ch *cmdutil.Helper, redirectURL str
...
+			if err != nil {
+				return fmt.Errorf("failed to open git repo: %w", err)
+			}
+		}
+
+		wt, err := repo.Worktree()
+		if err != nil {
+			return fmt.Errorf("failed to get worktree: %w", err)
+		}
+
```
begelundmuller, line 470: Any easy way to bail/prompt if the repo is bigger than e.g. 100 MB?

cli/cmd/deploy/deploy.go
```diff
@@ -358,6 +388,169 @@ func loginWithTelemetry(ctx context.Context, ch *cmdutil.Helper, redirectURL str
...
+		if err != nil {
+			return fmt.Errorf("failed to create remote: %w", err)
+		}
+
+		// push the changes
+		if err := repo.PushContext(ctx, &git.PushOptions{Auth: &http.BasicAuth{Username: "x-access-token", Password: pollRes.AccessToken}}); err != nil {
+			return fmt.Errorf("failed to push to remote %q : %w", *githubRepository.HTMLURL, err)
+		}
+
+		printer.ColorGreenBold.Printf("\nRepository %q created successfully. Local changes pushed to remote.\n\n", *githubRepository.Name)
```
begelundmuller, line 495: Would suggest printing repo created message right after it is created in case of errors during push. And then print something like "pushing local project to Github..." while doing the push.

cli/cmd/deploy/deploy.go
```diff
@@ -358,6 +388,169 @@ func loginWithTelemetry(ctx context.Context, ch *cmdutil.Helper, redirectURL str
...
+
+	ch.Print("\nRequest submitted for creating repository. Checking completion status\n")
+	pollCtx, cancel := context.WithTimeout(ctx, 10*time.Minute)
+	defer cancel()
+	for {
+		select {
+		case <-pollCtx.Done():
+			return nil, pollCtx.Err()
+		case <-time.After(5 * time.Second):
+			// Ready to check again.
```
begelundmuller, line 543: Can we check more often than every 5 seconds? Maybe every 2 seconds. To make it feel snappier.

cli/cmd/deploy/deploy.go
```diff
@@ -358,6 +388,169 @@ func loginWithTelemetry(ctx context.Context, ch *cmdutil.Helper, redirectURL str
...
+			// The users who installed app before we started including repo:write permissions need to accept permissions
+			// and then only we can create repositories.
+			return nil, fmt.Errorf("rill app does not have permissions to create github repository. Visit `https://github.com/settings/installations` to accept new permissions or reinstall app and try again")
+		}
+
+		if !strings.Contains(err.Error(), "name already exists") {
+			return nil, fmt.Errorf("failed to create repository: %w", err)
+		}
+		// there is a name conflict
+		tempRepoName = repoName + fmt.Sprintf("_v%v", i)
```
begelundmuller, line 527: Should we maybe prompt for a name instead?

begelundmuller (changes requested):

cli/cmd/deploy/deploy.go
k-anshul: The input from the user can also be already taken. May be we can do this in for loop similar to here.

cli/cmd/deploy/deploy.go
```diff
@@ -153,19 +157,45 @@ func DeployFlow(ctx context.Context, ch *cmdutil.Helper, opts *Options) error {
...
+			if !cmdutil.ConfirmPrompt("Do you want to continue?", "", true) {
 				ch.PrintfBold(githubSetupMsg)
 				return nil
 			}
-			return err
+			// ideally this should be clubbed with login similar to main github flow to reduce switch between CLI and browser
+			// but it is not possible right now since we need an input from user for the org where they want to create the repo.
+			if err := createGithubRepoFlow(ctx, ch, localGitPath); err != nil {
+				return err
+			}
```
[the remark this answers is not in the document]
k-anshul: Yeah we can login and connect to github in one go. Modified the `github/post-auth-redirect` endpoint to handle both cases.

cli/cmd/deploy/deploy.go
k-anshul: I could find this command : `git count-objects -v -H` 
But the problem is this works only after files have been added into the repository. And adding bigger files itself takes lot of time (Took ~4 minutes to add a file of size ~4.5GB). 
Can't think of any other easy way.

cli/cmd/deploy/deploy.go
```diff
@@ -358,6 +381,192 @@ func loginWithTelemetry(ctx context.Context, ch *cmdutil.Helper, redirectURL str
...
+			ch.Print("\nYou also have access to organization(s)\n\n")
+			repoOwner = cmdutil.SelectPrompt("Select Github account", repoOwners, pollRes.Account)
+		}
+		// create and verify
+		githubRepository, err := createGithubRepository(ctx, ch, pollRes, localGitPath, repoOwner)
+		if err != nil {
+			return err
+		}
+
+		printer.ColorGreenBold.Printf("\nRepository %q created successfully\n\n", *githubRepository.Name)
```
begelundmuller, line 465: Can we print the URL instead? I think that would be more useful for people. Like `Successfully created repository on "https://github.com/account/project"`

cli/cmd/deploy/deploy.go
begelundmuller: Okay, then let's like it like this for now

cli/cmd/deploy/deploy.go
```diff
@@ -358,6 +381,192 @@ func loginWithTelemetry(ctx context.Context, ch *cmdutil.Helper, redirectURL str
...
+		// git commit -m
+		_, err = wt.Commit("Auto committed by Rill", &git.CommitOptions{All: true})
+		if err != nil {
+			if !errors.Is(err, git.ErrEmptyCommit) {
+				return fmt.Errorf("failed to commit files to git: %w", err)
+			}
+		}
+
+		// Create the remote
+		_, err = repo.CreateRemote(&config.RemoteConfig{Name: "origin", URLs: []string{*githubRepository.HTMLURL}})
```
begelundmuller, line 498: Is the HTML URL the right one to use here? I thought it was supposed to be the `GitURL`

cli/cmd/deploy/deploy.go
```diff
@@ -358,6 +381,192 @@ func loginWithTelemetry(ctx context.Context, ch *cmdutil.Helper, redirectURL str
...
+		if err != nil {
+			return fmt.Errorf("failed to create remote: %w", err)
+		}
+
+		// push the changes
+		if err := repo.PushContext(ctx, &git.PushOptions{Auth: &http.BasicAuth{Username: "x-access-token", Password: pollRes.AccessToken}}); err != nil {
+			return fmt.Errorf("failed to push to remote %q : %w", *githubRepository.HTMLURL, err)
+		}
+
+		ch.Print("Local changes pushed to remote\n\n")
```
begelundmuller, line 508: Would suggest something like `Successfully pushed your local project to Github` to mirror the earlier log, and since this is the first push, there aren't really any "changes".

begelundmuller (changes requested):

web-admin/src/routes/-/github/connect/+page.svelte
```diff
@@ -42,11 +42,13 @@
...
-          githubUrl={remote}
-        />
-      </CtaMessage>
+      {#if remote}
+        <CtaMessage>
+          Please grant access to your repository<br /><GithubRepoInline
+            githubUrl={remote}
+          />
+        </CtaMessage>
+      {/if}
```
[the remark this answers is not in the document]
k-anshul: There is already a msg `Connect to Github`. I think that should be sufficient.
```
<CtaHeader>Connect to Github</CtaHeader>
<CtaMessage>
  Rill projects deploy continuously when you push changes to Github.
</CtaMessage>
```

cli/pkg/cmdutil/prompt.go
```diff
@@ -48,17 +48,17 @@ func ConfirmPrompt(msg, help string, def bool) bool {
...
+func InputPrompt(msg, def string) string {
 	prompt := &survey.Input{
 		Message: msg,
 		Default: def,
 	}
 	result := def
 	if err := survey.AskOne(prompt, &result); err != nil {
 		fmt.Printf("Prompt failed %v\n", err)
-		return "", err
+		os.Exit(1)
```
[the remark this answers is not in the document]
k-anshul: We were already doing `os.Exit` in other prompts so I just followed the same here. Can revert here and others can be handled separately ?

cli/cmd/deploy/deploy.go
```diff
@@ -358,6 +381,192 @@ func loginWithTelemetry(ctx context.Context, ch *cmdutil.Helper, redirectURL str
...
+		if !cmdutil.ConfirmPrompt("Do you want to continue?", "", true) {
+			ch.PrintfBold(githubSetupMsg)
+			return nil
+		}
+
+		repoOwner := pollRes.Account
+		if len(pollRes.Organizations) > 0 {
+			repoOwners := []string{pollRes.Account}
+			repoOwners = append(repoOwners, pollRes.Organizations...)
+			ch.Print("\nYou also have access to organization(s)\n\n")
```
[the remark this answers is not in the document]
k-anshul: I felt it gives a subtle nudge to the user that he is also part of some organization. I can remove this as well.

web-admin/src/routes/-/github/connect/+page.svelte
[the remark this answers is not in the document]
k-anshul: image

cli/cmd/deploy/deploy.go
k-anshul: Both works I guess. 
We can do `git clone https://github.com/rilldata/rill` or `git clone https://github.com/rilldata/rill.git` as well.

cli/cmd/deploy/deploy.go
```diff
@@ -153,19 +157,29 @@ func DeployFlow(ctx context.Context, ch *cmdutil.Helper, opts *Options) error {
...
+			// It's not a valid remote for Github. We still navigate user to login and then ask user to chhose either to create repo manually or let rill create one for them.
+			silent := false
 			if !ch.IsAuthenticated() {
-				err := loginWithTelemetry(ctx, ch, "")
+				err := loginWithTelemetryAndGithubRedirect(ctx, ch, "")
 				if err != nil {
-					ch.PrintfWarn("Login failed with error: %s\n", err.Error())
+					return fmt.Errorf("login failed with error: %w", err)
 				}
-				fmt.Println()
```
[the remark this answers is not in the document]
k-anshul: I didn't understand. This was for the case when there was no git remote found. The space was between login output and the git instructions. The flow is completely overhauled now so I guess this should not be required now.

web-admin/src/routes/-/github/connect/+page.svelte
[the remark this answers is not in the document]
begelundmuller: Got it, yeah that seems fine

begelundmuller (approved):

PR #4493, opened 2024-04-03
Refactor: Unify Source and Model workspaces

briangregoryholmes:
This PR unifies the source and model workspaces into a single `+page.svelte` file to resolve a variety of navigation related bugs and latency issues.

Reasoning:
- Reduce reflow and thrown click handler violations
- Improve Lighthouse performance score
- Use of key blocks + navigation causes issues with transitions
- Latency issues when navigating causes uncontrolled file/data fetching that can throw
- Components are being unnecessarily mounted/unmounted rather than simply having their props change
- Fixes an issue where `formatConnectorType` was returning the wrong property
- Removes use of source store as files are not persisted after navigating and the blob already exists in the queryClient cache
- Eliminate redundant/stale code
- Remove navigation side effect from handleEntityRename
- Behavior (errors, autosave, panel toggling, refresh state) is currently out of sync between models/sources and unifying workspaces allows us to update them in tandem (though some differences are maintained in this PR)
- New structure should eventually work or any asset type, supporting the move to the asset explorer
- Fix incorrect or redundant typing
- Reduced file count and code complexity lowers onboarding for new developers


ericpgreen2:
I didn't realize the scope on this, but nice to see the cleanup! Some UXQA findings:
1. I'm seeing that sometimes I don't land on the intended model page, but instead get redirected to the root page. JAM: https://jam.dev/c/fa21242d-76cc-4bac-8dff-a7d640b3c2c3
2. The "Refresh" button doesn't work for Local Files sources. JAM: https://jam.dev/c/0a24f64a-16f8-4187-bbc4-7c2310c6c94e
3. Switching between sources does not change the source name in the Workspace Header. This header should read "nyc311_reduced":
image
4. The dialog now says you're leaving a "model" not a "source"
image
5. When the left sidebar is closed, the "open sidebar" button overlaps with the Source name
image

Otherwise looks good, so I'll approve for you to merge at your discretion.

ericpgreen2 (approved):

briangregoryholmes:
@ericpgreen2 Would love it if you could take another look. Also tagging @AdityaHegde since he seems to have a PR open that touches these files.

Essentially, I had to just nip this thing in the bud rather than trying to do patch fixes. This PR is now a total rework of the workspace routing and combines the model and source workspaces into a single `+page.svelte` file. It solves a ton of issues, removes a lot of code and will make the new data loading model a lot more straightforward (not included in this PR). That said, it's obviously a huge change and needs other eyes.

The `+page` file is currently pretty long, but I think relatively straightforward It'll get a lot simpler once we move the data loading out and can unify the Editor, error handling and inspectors. I can actually do the error handling in this PR if we can decide on which version (model or source) is preferred.

AdityaHegde (approved):

PR #4494, opened 2024-04-03
Telemetry and source rename fixes

AdityaHegde:
Fixes a couple of things,
1. Behaviour telemetry is not present on cloud. This PR disables it when the global is not present. We should look into enabling in the future.
2. Source rename doesnt invalidate the olap table list properly. This PR tracks a rename and invalidates it. We use `renamedFrom` in the resource meta to check if it was set and then unset. Reconciler uses it internally to tracks renames, but unsets it once reconciler is done renaming.

### DIFF
--- a/web-common/src/features/dashboards/tab-bar/TabBar.svelte
+++ b/web-common/src/features/dashboards/tab-bar/TabBar.svelte
@@ -47,7 +47,8 @@
       selectedTab.label === "Pivot",
     );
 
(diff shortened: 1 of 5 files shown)


nishantmonu51 (approved):

ericpgreen2:
I'm getting a flicker in the Tables list whenever I rename a source. JAM: https://jam.dev/c/541534cd-7000-4906-9d5b-6722d9f425df

PR #4495, opened 2024-04-03
Support type hint and lint for metrics editor

djbarnwal:
Add type hints and lint for metrics editor. This PR also adds a JSON Schema for the metrics editor. In long term we would want to auto generate the schema from our Typescript types and doc strings. Since we are not there yet and not all types are consolidated, this JSON schema has be updated manually (along with docs and go validator) when we update our schema.

Future Improvements - 
- Auto completion based on enum 
- Add link and markdown support in tooltips/description


ericpgreen2 (commented):
This is great! My one worry is the hardcoded JSON schema will be a PITA to continually synchronize with the source-of-truth on the backend. Maybe there's a way we can generate it programmatically? @begelundmuller, thoughts?

PR #4503, opened 2024-04-03
PIVOT exporting dimension alias fix

egor-ryashin:


runtime/queries/metricsview_aggregation.go
```diff
@@ -683,6 +684,7 @@ func (q *MetricsViewAggregation) buildMetricsAggregationSQL(mv *runtimev1.Metric
 		)
 	}
 
+	fmt.Println(sql, args)
```
begelundmuller, line 687: Probably needs to be removed :)

begelundmuller (approved):

PR #4505, opened 2024-04-03
Enable ListNotifierConnectors for Slack

begelundmuller:
- Returns info about registered Slack connectors ([PR](https://github.com/rilldata/rill/pull/4371)) in `ListNotifierConnectors`.
- Note that even if the response does not contain a "slack" driver, the "slack" connector can still be used anonymously if a Slack webhook is provided to the alert. A Slack connector only needs to be configured to support delivery to users and channels.
- The `ListNotifierConnectors` API does not yet return info about email connector.


runtime/server/connectors.go
```diff
@@ -106,6 +107,28 @@ func (s *Server) ListNotifierConnectors(ctx context.Context, req *runtimev1.List
 		}
 	}
 
+	// Connectors may be implicitly defined just by adding variables in the format "connector.<name>.<property>".
```
esevastyanov, line 110: Slack connector is defined even if bot token isn't there but only for webhooks

begelundmuller: Yeah the abstractions definitely get blurry with that kind of "anonymous" functionality. In this case, it's useful for the `ListNotifierConnectors` only to return connectors that have been configured somehow (with vars or by adding to `rill.yaml`). Will add a note in the PR about it.

esevastyanov (approved):

PR #4510, opened 2024-04-04
Runtime: use a single directory for all instance data

### LINKED ISSUE #4497: Runtime: Use a single directory for all instance data
An instance may currently create files in many different directories – namely in the directory containing the DuckDB file (specified when the instance is created) and in various temporary directories that are created ad-hoc.

We want to centralize all data related to an instance in a single directory (which of course can have any number of subdirectories). This will provide two advantages:
- We can emit a metric of the total disk usage for an instance (to be used for quota/usage/billing tracking)
- We can easily clean up an instance's data when it's deleted

k-anshul:
closes #4497 

- Post this change for newly deployed projects, the data will be stored in `data/instance_id/connector` directory **so projects will get reset once this change is deployed.**
- In memory duckdb is handled with special DSN prefix`:memory:` similar to duckdb.


cli/cmd/runtime/start.go
```diff
@@ -83,6 +83,9 @@ type Config struct {
 	// AllowHostAccess controls whether instance can use host credentials and
 	// local_file sources can access directory outside repo
 	AllowHostAccess bool `default:"false" split_words:"true"`
+	// DataDir stores data for all instances like duckdb file, temporary downloaded file etc.
+	// The data for each instance is stored in a child directory named instance_id
+	DataDir string `split_words:"true"`
```
begelundmuller, line 88: 1. Needs to be added to the shared `.env` file (`rill devtool dotenv upload cloud`). Should probably point to `dev-cloud-state`.
2. Need to remember to get infra to set this on the runtimes. They also need to think apply it to the Kubernetes provisioned runtimes.

cli/pkg/local/app.go
```diff
@@ -186,6 +187,9 @@ func NewApp(ctx context.Context, opts *AppOptions) (*App, error) {
 	if opts.OlapDriver == DefaultOLAPDriver && olapDSN == DefaultOLAPDSN {
 		defaultOLAP = true
 		olapDSN = path.Join(dbDirPath, olapDSN)
+		// Set path to ensure that OLAP does not store data in runtime.Options.DataDir
```
begelundmuller, line 190: This is still the same as `DataDir`. Maybe note that otherwise it would be put in a nested directory and we need this for backwards compatibility.

runtime/drivers/admin/admin.go
```diff
@@ -276,7 +278,10 @@ func (h *Handle) checkHandshake(ctx context.Context) error {
 	}
 
 	if h.repoPath == "" {
-		h.repoPath, err = os.MkdirTemp("", "admin_driver_repo")
+		if err := os.Mkdir(h.config.DataDir, os.ModePerm); err != nil && !errors.Is(err, fs.ErrExist) {
+			return err
+		}
+		h.repoPath, err = os.MkdirTemp(h.config.DataDir, "admin_driver_repo")
```
begelundmuller, line 284: It's problematic to create temporary files in the `data_dir/instance_id/connector` directory because it makes it hard to clear out old `tmp` files.

Probably we should centralize temp files in `data_dir/instance/tmp/` or maybe even `data_dir/tmp/`? And clear the `tmp` directory each time the runtime is started (since when not using the system temp folder, it doesn't get auto-cleared).

Regarding accessing it here in the connector, we could pass in both a `DataDir` and a `TempDir`?

runtime/drivers/blob/blobdownloader.go
```diff
@@ -116,7 +118,7 @@ var _ drivers.FileIterator = &blobIterator{}
 func NewIterator(ctx context.Context, bucket *blob.Bucket, opts Options, l *zap.Logger) (drivers.FileIterator, error) {
 	opts.validate()
 
-	tempDir, err := os.MkdirTemp(os.TempDir(), "blob_ingestion")
+	tempDir, err := os.MkdirTemp(opts.DataDir, "blob_ingestion")
```
begelundmuller, line 121: Same temp data issue

runtime/drivers/duckdb/olap_test.go
```diff
@@ -9,13 +9,14 @@ import (
 	"testing"
 	"time"
 
+	"strings"
+
```
begelundmuller, line 13: Should be one group

runtime/drivers/github/github.go
```diff
@@ -47,22 +49,38 @@ func init() {
...
+	err = json.Unmarshal([]byte(conf.DSN), &dsn)
 	if err != nil {
 		return nil, err
 	}
 
-	tempdir, err := os.MkdirTemp("", "github_repo_driver")
+	if err := os.Mkdir(conf.DataDir, os.ModePerm); err != nil && !errors.Is(err, fs.ErrExist) {
+		return nil, err
+	}
+	tempdir, err := os.MkdirTemp(conf.DataDir, "github_repo_driver")
```
begelundmuller, line 83: Same temp data issue

runtime/registry.go
```diff
@@ -61,7 +64,16 @@ func (r *Runtime) WaitUntilIdle(ctx context.Context, instanceID string, ignoreHi
...
-	return r.registryCache.create(ctx, inst)
+	if err := r.registryCache.create(ctx, inst); err != nil {
+		return err
+	}
+
+	if r.opts.DataDir != "" {
+		if err := os.Mkdir(filepath.Join(r.opts.DataDir, inst.ID), os.ModePerm); err != nil && !errors.Is(err, fs.ErrExist) {
+			return err
+		}
+	}
```
begelundmuller, line 75: It would be better to do this in `registryCache.add` to make it self-recovering in case the directory is removed.

runtime/registry.go
```diff
@@ -110,6 +122,11 @@ func (r *Runtime) DeleteInstance(ctx context.Context, instanceID string, dropOLA
 		if err != nil {
 			r.logger.Error("could not drop database", zap.Error(err), zap.String("instance_id", instanceID), observability.ZapCtx(ctx))
 		}
+
+		// We drop all data if dropOLAP is true
+		if err := os.RemoveAll(filepath.Join(r.opts.DataDir, instanceID)); err != nil {
+			r.logger.Error("could not drop instance data directory", zap.Error(err), zap.String("instance_id", instanceID), observability.ZapCtx(ctx))
+		}
```
begelundmuller, line 129: `dropOLAP` was added just for DuckDB, we never use it (and probably shouldn't use it) for external OLAPs.

So how about just always deleting the data dir and removing the `dropOLAP` parameter and `drivers.Drop` function altogether?

begelundmuller (changes requested):

cli/cmd/runtime/start.go
k-anshul: Ack

cli/pkg/local/app.go
k-anshul: Sorry I didn't understand. `dbDirPath` is `<projectPath>/tmp`  and olapDSN is `<DefaultOLAPDSN>(main.db)` which is also the path to duckdb file currently.

runtime/drivers/admin/admin.go
k-anshul: I thought we should rely on connectors to delete any temporary files which they should be able to do in most cases apart from crashes.
Storing temporary files in a separate directory is also a good idea but I would prefer to keep `data_dir/instance/tmp` for following reasons even if it will add extra code to delete temporary files.

1. Max storage space will not track downloaded data in case we want to track that. 
2. Also I believe the main purpose of this exercise was to make sure that we add all data for an instance in one directory only.

cli/pkg/local/app.go
begelundmuller: I just meant to clarify the comment. Right now it says "ensure that OLAP does not store data in runtime.Options.DataDir", but I think that is the opposite of what it does – because in the default case it is setting `DataDir == "./tmp"` and `olapDSN == "./tmp/main.db"`, right?

runtime/drivers/admin/admin.go
begelundmuller: > I thought we should rely on connectors to delete any temporary files

On the happy path, yes, but things crash and have bugs. We should prevent temp data from accumulating in a persistent directory over time.

> I would prefer to keep `data_dir/instance/tmp`

Yeah you're right, of course we need to use that to ensure correct disk usage metrics :)

runtime/runtime.go
```diff
@@ -108,6 +116,27 @@ func (r *Runtime) GetInstanceAttributes(ctx context.Context, instanceID string)
...
+		return
+	}
+
+	for _, match := range matches {
+		err := os.RemoveAll(match)
+		if err != nil {
+			logger.Warn("failed to remove temp directory", zap.String("dir", match), zap.Error(err))
+		}
+	}
+}
```
begelundmuller, line 138: It might be cleaner to do this in `registryCache.add` (i.e. before creating the empty `tmp` directory, delete any existing `tmp` directory).

k-anshul: Yeah I also thought so but actually that may not handle all failures cases like during `DeleteInstance instance was deleted from registry but controller did not stop and server crashed. 
Since this is an extra check just to ensure we handle all failure cases as best as possible, I felt this is the best place to do that.

begelundmuller (approved):

PR #4515, opened 2024-04-04
UI to support multiple schemas in an OLAP engine

ericpgreen2:
- All API calls that reference a Table now specify the Table's specific `database` and `databaseSchema`
- Table names show their fully-qualified table names
- Druid, ClickHouse, and DuckDB OLAP engines are supported


k-anshul:
Hi Eric 

Please find the responses below:
> What should we do about a source/model connector's database and databaseSchema? Just omit for now?

As of now both models and sources are created in the default database and default schema so we can skip for now.

>I'm getting the following reconcile error when generating a dashboard based off a Druid table:

The druid was exposing database field as well but as confirmed with Parag, druid doesn't support db field while querying so made changes to do the handling. Thanks for reporting. This should be fixed now in your PR.

PR #4517, opened 2024-04-05
Disable search for non-string dimensions

AdityaHegde:
This PR disables search for non-string dimensions in,
- [x] Filter pill
- [x] Dimension Table
- [ ] Time Dimension Detail (?)
- [ ] Pivot (?)

Also removes a usage of old filters in getting the filter list in pills. Also updates to use MetricsViewComparison instead of MetricsViewToplist (deprecated)


djbarnwal (approved):
Thanks for the changes. 

I will create a follow up PR for TDD on similar lines.
Pivot doesn't have search yet so shouldn't be an issue.

PR #4519, opened 2024-04-05
 Init Runtime Store

briangregoryholmes:
This PR initializes the runtime store with default values and, in Rill Developer, removes the use of the RuntimeProvider component, which is not necessary.

Additionally, it cleans up the process by which we derive the runtime URL.

This is a necessary change to enable data fetching in `+` files as the http-client requires that runtime be defined.


ericpgreen2 (commented):
I'm getting an error when running the Cloud UI. It looks like the feature flag store is trying and failing to find the `rill.yaml` file – looks like we missed this when it was added, but that should only happen in Rill Developer.
image

briangregoryholmes:
Alright, should be okay now. The types kind of threw me off. `instanceId` is treated as a string throughout the application with the `undefined` state being an empty string.  There are some query wrappers that run when they should't but are disabled so long as "instanceId" is falsy. Should probably move to an explicit `string | null` type and restructure , but the exposure to that change across the app is huge.

For my own sanity, I'll try to explain the host behavior. In Rill developer, the host should be `http://localhost:9009` when in dev mode. Instead of using the `RILL_DEV` flag in package.json, I've changed it to use the SvelteKit flag. In testing and prod, the host should be an empty string so that the requests are made via relative paths.

In web-admin, there's another httpClient for the admin service that uses the Axios. It runs on "http://localhost:8080" or, in production, on a host that is injected via an environment variable. This PR doesn't touch that.

web-admin also uses the runtime service, but the host is conditional based on the project and no calls should be made to it until you have a host and instance id. However, the layout does have hooks like `useDashboards` being called in the tree above the `RuntimeProvider`. Because of that, the web-common runtime store needs to be initialized with empty strings so that those queries are not enabled.

After digging into this, the only thing that really needs to be a writable is the instanceId, though that all also get moved to mostly existing inside `+` files. Nothing subscribes to the JWT (and, in my opinion, it shouldn't really be touched in client side code) and the host should be captured/updated from within the httpClient via closure.

ericpgreen2 (approved):

PR #4526, opened 2024-04-05
Refactor measure filter components to be agnostic of proto types

AdityaHegde:
For ease of use mapping to and from proto type `V1Expression` should be separate from the measure filter forms and pills.

This PR is a 1st step towards this. It refactors all components dealing with measure filter editing and the pills to use a new format that can be mapped directly to the form.
Note: this makes things easier for comparison in alerts.

A follow up PR will refactor our dashboard state to store this directly.

### DIFF
--- a/web-common/src/features/dashboards/filters/measure-filters/MeasureFilter.svelte
+++ b/web-common/src/features/dashboards/filters/measure-filters/MeasureFilter.svelte
@@ -8,6 +8,11 @@
   import Tooltip from "@rilldata/web-common/components/tooltip/Tooltip.svelte";
   import TooltipContent from "@rilldata/web-common/components/tooltip/TooltipContent.svelte";
   import TooltipTitle from "@rilldata/web-common/components/tooltip/TooltipTitle.svelte";
+  import {
+    mapExprToMeasureFilter,
+    mapMeasureFilterToExpr,
+    MeasureFilterEntry,
+  } from "@rilldata/web-common/features/dashboards/filters/measure-filters/measure-filter-entry";
   import MeasureFilterBody from "@rilldata/web-common/features/dashboards/filters/measure-filters/MeasureFilterBody.svelte";
   import MeasureFilterMenu from "@rilldata/web-common/features/dashboards/filters/measure-filters/MeasureFilterMenu.svelte";
   import type { V1Expression } from "@rilldata/web-common/runtime-client";
@@ -31,10 +36,13 @@
       active = false;
     }
   }
+
+  // TODO: in the next round of refactor update upstream to use MeasureFilterEntry
+  let filter: MeasureFilterEntry | undefined;
+  $: filter = mapExprToMeasureFilter(expr);
 </script>
 
 <DropdownMenu.Root
-  preventScroll
   bind:open={active}
   onOpenChange={(open) => {
     if (!open) {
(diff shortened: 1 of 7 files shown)


web-common/src/features/dashboards/filters/measure-filters/MeasureFilter.svelte
```diff
@@ -31,10 +36,13 @@
       active = false;
     }
   }
+
+  // TODO: in the next round of refactor update upstream to use MeasureFilterEntry
+  let filter: MeasureFilterEntry | undefined;
```
briangregoryholmes, line 41: Nit: Just my preference, but I generally like avoiding the secondary declaration unless it's for caching reasons or a prop. Ensures the types are being accurately inferred and saves a couple lines.

web-common/src/features/dashboards/filters/measure-filters/MeasureFilterBody.svelte
```diff
@@ -6,36 +6,42 @@
...
+  $: if (filter) {
+    switch (filter.operation) {
+      case MeasureFilterOperation.GreaterThan:
+      case MeasureFilterOperation.GreaterThanOrEquals:
+      case MeasureFilterOperation.LessThan:
+      case MeasureFilterOperation.LessThanOrEquals:
+        shortLabel =
+          MeasureFilterOptions.find(
+            (o) =>
+              // svelte-check is throwing an error here stating `filter could be undefined` so we need this
```
briangregoryholmes, line 32: Is undefined a valid state? Seems like there should be a conditional further up the tree that just prevents this from rendering.

web-common/src/features/dashboards/filters/measure-filters/MeasureFilterMenu.svelte
```diff
@@ -3,23 +3,22 @@
...
+  } from "@rilldata/web-common/features/dashboards/filters/measure-filters/measure-filter-options";
+  import { getStateManagers } from "@rilldata/web-common/features/dashboards/state-managers/state-managers";
   import { createEventDispatcher } from "svelte";
   import { createForm } from "svelte-forms-lib";
   import * as yup from "yup";
 
   export let dimensionName: string;
   export let name: string;
-  export let expr: V1Expression | undefined;
+  export let filter: MeasureFilterEntry | undefined;
```
briangregoryholmes, line 21: Same thing. Can we get rid of the undefined and just wrap the use of this component in an `#if` block?

briangregoryholmes (approved):
Awesome. A lot cleaner and easy to follow. Added a couple nits and just a question about whether we should have undefined as an option or instead just have a conditional higher up the tree.

web-common/src/features/dashboards/filters/measure-filters/MeasureFilter.svelte
AdityaHegde: The case where this is undefined is when the menu is opened for the 1st time. Same applies to other components as well.

web-common/src/features/dashboards/filters/measure-filters/MeasureFilterBody.svelte
AdityaHegde: For this piece of code there is a undefined check at the beginning of this reactive statement, so this is a false positive in svelte-check.

For the case where filter could be null is when the filter pill was just added.

PR #4529, opened 2024-04-07
Refactor: Preview Table 

briangregoryholmes:
This PR adds a performance-focused, zero dependency `VirtualTable` component with row and column virtualization. In this PR, it is used to replace all instances of `PreviewTable` across the application, but the component has been designed in such a way that it should also work as a replacement for `DimensionTable` and `TDDTable` in the future (with expected modifications). That said, reviews of this PR should consider it only as a replacement for `PreviewTable`. It features an improved handling of Tooltips, but one that is separate from the proposal for a global overhaul of that system.

You should see improvements in page load times, table render times, scroll performance, column resize frame rate, workspace responsiveness and heap snapshot size when viewing the table.


ericpgreen2 (approved):
Much better performance! Nice work 👏

djbarnwal (changes requested):
Thanks for the PR Brian. Couple of regressions - 

- The table for model/sources is not styled right
- The column size is not adaptive. I would expect the columns with smaller values and header (in length) to have a thinner column. (Not a blocker but good to have if it doesn't add performance issues)
image

briangregoryholmes:
@djbarnwal Got the Firefox one. Column sizes are set variably, but I was applying too much buffer, so that's been addressed as well.

djbarnwal (approved):

PR #4533, opened 2024-04-08
Refactor the breadcrumbs

ericpgreen2:
This PR breaks up the one `Breadcrumbs` component into individual components for each breadcrumb (Org, Project, Dashboard, Report, Alert).

Mainly, this PR is to prepare for the Rill Developer "Preview mode", which will re-use the `DashboardBreadcrumb` component.

Additionally, this structure better ensures we do not have Query Observers mounted for resources that are not relevant to the given page (e.g. when on a Dashboard page, there's no need for the Reports query to be active).


briangregoryholmes (approved):
Looks good and a lot easier to follow. For some of the items, it looks like you could set the href prop on DropdownMenu.Item directly and bypass the onSelect callback.

PR #4552, opened 2024-04-09
Update ListFiles to return directories and add MakeDir endpoint

AdityaHegde:
To prepare for asset explorer,
1. We need to get empty directories in ListFiles to allow users to add files to them.
2. We need a way to create a new directory. Adding `MakeDir` for this.

Note: We do not need this in watch files as of now, since that is used to react to assets.


runtime/drivers/drivers_test.go
```diff
@@ -21,10 +21,10 @@ import (
 // as subtests of TestAll.
 func TestAll(t *testing.T) {
 	var matrix = []func(t *testing.T, fn func(driver string, shared bool, cfg map[string]any)) error{
-		withDuckDB,
+		//withDuckDB,
 		withFile,
-		withPostgres,
-		withSQLite,
+		//withPostgres,
+		//withSQLite,
```
begelundmuller, line 27: Needs to be reverted

runtime/drivers/repo.go
```diff
@@ -39,3 +40,8 @@ type RepoObjectStat struct {
 }
 
 var ErrFileAlreadyExists = errors.New("file already exists")
+
+type FileEntry struct {
+	Path  string
+	IsDir bool
+}
```
begelundmuller, line 47: Should we call it `DirEntry` to align with Go's type name for it?

AdityaHegde: Ya makes sense

proto/rill/runtime/v1/api.proto
```diff
@@ -78,6 +78,13 @@ service RuntimeService {
     };
   }
 
+  rpc CreateDirectory(CreateDirectoryRequest) returns (CreateDirectoryResponse) {
```
begelundmuller, line 81: Add a docstring :)

proto/rill/runtime/v1/api.proto
```diff
@@ -78,6 +78,13 @@ service RuntimeService {
...
+    option (google.api.http) = {
+      post: "/v1/instances/{instance_id}/files/make-dir/-/{path=**}",
```
begelundmuller, line 83: `make-dir` probably needs to be changed

proto/rill/runtime/v1/api.proto
```diff
@@ -448,6 +460,15 @@ message PutFileResponse {
   string file_path = 1; // TODO: Redundant, should be removed (but frontend currently uses it)
 }
 
+message CreateDirectoryRequest {
```
begelundmuller, line 463: Docstring

proto/rill/runtime/v1/api.proto
```diff
@@ -448,6 +460,15 @@ message PutFileResponse {
...
+  string instance_id = 1 [(validate.rules).string = {pattern: "^[_\\-a-zA-Z0-9]+$"}];
+  string path = 2 [(validate.rules).string.min_len = 1];
+  // Indicates whether to ignore the request if the directory already exists
+  bool ignore_if_exists = 3;
+}
+
+message CreateDirectoryResponse {}
```
begelundmuller, line 470: Docstring

runtime/compilers/rillv1/parser.go
```diff
@@ -172,11 +172,16 @@ type Parser struct {
...
-	paths, err := repo.ListRecursive(ctx, "rill.{yaml,yml}")
+	files, err := repo.ListRecursive(ctx, "rill.{yaml,yml}", true)
 	if err != nil {
 		return nil, fmt.Errorf("could not list project files: %w", err)
 	}
 
+	var paths []string
+	for _, file := range files {
+		paths = append(paths, file.Path)
+	}
```
begelundmuller, line 183: Should be removed and `files` renamed to `paths` (look at the diff vs. main)

runtime/compilers/rillv1/parser.go
```diff
@@ -255,11 +260,16 @@ func (p *Parser) reload(ctx context.Context) error {
...
-	paths, err := p.Repo.ListRecursive(ctx, "**/*.{env,sql,yaml,yml}")
+	files, err := p.Repo.ListRecursive(ctx, "**/*.{env,sql,yaml,yml}", true)
 	if err != nil {
 		return fmt.Errorf("could not list project files: %w", err)
 	}
 
+	var paths []string
+	for _, file := range files {
+		paths = append(paths, file.Path)
+	}
```
begelundmuller, line 271: Same comment

runtime/repos.go
```diff
@@ -54,6 +56,23 @@ func (r *Runtime) PutFile(ctx context.Context, instanceID, path string, blob io.
...
 }
 
+func (r *Runtime) MakeDir(ctx context.Context, instanceID, path string, ignoreIfExists bool) error {
+	repo, release, err := r.Repo(ctx, instanceID)
+	if err != nil {
+		return err
+	}
+	defer release()
+
+	// TODO: Handle ignoreIfExists
```
begelundmuller, line 66: Can we fix this todo (or remove the `ignoreIfExists` option entirely)?

runtime/server/repos.go
```diff
@@ -62,14 +70,14 @@ func (s *Server) WatchFiles(req *runtimev1.WatchFilesRequest, ss runtimev1.Runti
 	defer release()
 
 	if req.Replay {
-		paths, err := repo.ListRecursive(ss.Context(), "**")
+		files, err := repo.ListRecursive(ss.Context(), "**", true)
```
begelundmuller, line 73: Won't you need to be told about new/deleted directories in `WatchFiles`?

proto/rill/runtime/v1/api.proto
```diff
@@ -397,7 +404,12 @@ message ListFilesRequest {
 
 // Response message for RuntimeService.ListFiles
 message ListFilesResponse {
-  repeated string paths = 1;
+  repeated FileEntry files = 1;
+}
+
+message FileEntry {
```
begelundmuller, line 410: 1. Add a docstring
2. If we're calling it `DirEntry` internally now, we should probably also call it that here to align names

begelundmuller (changes requested):

runtime/repos.go
AdityaHegde: Ya i think it makes sense to just remove it. Let it error if the folder exists.

begelundmuller (approved):

PR #4558, opened 2024-04-09
Admin same port fix

egor-ryashin:


runtime/pkg/graceful/http.go
```diff
@@ -56,3 +52,9 @@ func ServeHTTP(ctx context.Context, server *http.Server, options ServeOptions) e
 
 	return serveErr
 }
+
+func noReusePort(network, address string, conn syscall.RawConn) error {
+	return conn.Control(func(descriptor uintptr) {
+		syscall.SetsockoptInt(int(descriptor), syscall.SOL_SOCKET, syscall.SO_REUSEADDR, 0)
+	})
+}
```
begelundmuller, line 60: This seems like a potentially dangerous thing to change. Have you found any references to indicate it is safe to use? Issues I'm worried about:
- Will it work on WSL?
- What if someone restarts Rill rapidly – might they get a port error then? For example if the browser is still trying to connect.
- Could it impact restarts in production?

nishantmonu51:
closing as stale PR. Please reopen if necessary.

PR #4559, opened 2024-04-09
Runtime: Add missing dry run support in `Exec` method of clickhouse driver

k-anshul:


runtime/drivers/clickhouse/olap.go
```diff
@@ -60,6 +60,25 @@ func (c *connection) Exec(ctx context.Context, stmt *drivers.Statement) error {
...
+
+		// TODO: Find way to validate with args
+		name := uuid.NewString()
+		_, err = conn.ExecContext(ctx, fmt.Sprintf("CREATE VIEW %q AS %s", name, stmt.Query))
+		if err != nil {
+			return err
+		}
+
+		_, err = conn.ExecContext(context.Background(), fmt.Sprintf("DROP VIEW %q", name))
+		return err
```
begelundmuller, line 79: Did you consider `EXPLAIN`? Wondering if it would reduce the chance of orphaned resources or flashes in the table listings

k-anshul: Good idea. Actually just realised that we should not use `CREATE` for external connections. Having write access is not guaranteed. 
Let me check if `EXPLAIN` has any latency issues.

begelundmuller (approved):

PR #4562, opened 2024-04-09
fix: avoid crash from missing project

Kavinjsir:
### Description

Log a warning for an invalid metrics project specified by the environment variable instead of throwing an error.

### Motivation

Currently, if a metrics project is specified in the environment variable, failure to find the project leads to the crash of the admin service.
I believe the main process of the admin service should not be affected by the metrics project or slot autoscaler.

### DIFF
--- a/admin/admin.go
+++ b/admin/admin.go
@@ -80,9 +80,10 @@ func New(ctx context.Context, opts *Options, logger *zap.Logger, issuer *auth.Is
 	if opts.MetricsProjectOrg != "" && opts.MetricsProjectName != "" {
 		proj, err := db.FindProjectByName(ctx, opts.MetricsProjectOrg, opts.MetricsProjectName)
 		if err != nil {
-			return nil, fmt.Errorf("error looking up metrics project: %w", err)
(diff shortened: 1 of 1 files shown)


admin/admin.go
```diff
@@ -80,9 +80,10 @@ func New(ctx context.Context, opts *Options, logger *zap.Logger, issuer *auth.Is
 	if opts.MetricsProjectOrg != "" && opts.MetricsProjectName != "" {
 		proj, err := db.FindProjectByName(ctx, opts.MetricsProjectOrg, opts.MetricsProjectName)
 		if err != nil {
-			return nil, fmt.Errorf("error looking up metrics project: %w", err)
+			logger.Sugar().Warnf("error looking up metrics project: %w", err)
```
begelundmuller, line 83: This feels slightly like an anti-pattern. It's cleaner to fail if an invalid value is provided (and then handle the project-not-created-yet case by not setting the env var until later).

But if this is really needed – can we at least change this log to error-level? (Since it's a real error that might indicate a misconfigured system.)

Kavinjsir:
Close this ticket, as we will be follow the principle to validate the env if set for scaler.

PR #4563, opened 2024-04-09
Introducing `RILL_ADMIN_AUTOSCALER_CRON` Environment Variable

Kavinjsir:
#### Purpose

Allow configuration of the slot autoscaler cron job frequency.

#### Background

Previously, the run_autoscaler cron job frequency was hardcoded to _00:00 Monday PT_.

#### Change

The new `RILL_ADMIN_AUTOSCALER_CRON` variable enables flexible scheduling, facilitating more frequent runs in test or debug modes.

#### Usage

Set `RILL_ADMIN_AUTOSCALER_CRON` to the desired cron schedule, based on the [robfig-cron syntax](https://pkg.go.dev/github.com/robfig/cron/v3#hdr-Usage). e.g., `CRON_TZ=Asia/Tokyo 30 04 * * *` for _04:30 Tokyo time every day_.

### DIFF
--- a/admin/admin.go
+++ b/admin/admin.go
@@ -21,6 +21,7 @@ type Options struct {
 	VersionNumber      string
 	MetricsProjectOrg  string
 	MetricsProjectName string
+	AutoscalerCron     string
 }
 
 type Service struct {
@@ -35,6 +36,7 @@ type Service struct {
 	issuer           *auth.Issuer
 	VersionNumber    string
 	metricsProjectID string
+	AutoscalerCron   string
 }
 
 func New(ctx context.Context, opts *Options, logger *zap.Logger, issuer *auth.Issuer, emailClient *email.Client, github Github, aiClient ai.Client) (*Service, error) {
@@ -97,6 +99,7 @@ func New(ctx context.Context, opts *Options, logger *zap.Logger, issuer *auth.Is
 		issuer:           issuer,
 		VersionNumber:    opts.VersionNumber,
 		metricsProjectID: metricsProjectID,
+		AutoscalerCron:   opts.AutoscalerCron,
 	}, nil
 }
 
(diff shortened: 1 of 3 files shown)


cli/cmd/admin/start.go
```diff
@@ -77,6 +77,7 @@ type Config struct {
 	ActivitySinkKafkaBrokers string                 `default:"" split_words:"true"`
 	ActivityUISinkKafkaTopic string                 `default:"" split_words:"true"`
 	MetricsProject           string                 `default:"" split_words:"true"`
+	AutoscalerCron           string                 `default:"" split_words:"true"`
```
begelundmuller, line 80: Change the default to `default:"CRON_TZ=America/Los_Angeles 0 0 * * 1"`

admin/admin.go
```diff
@@ -97,6 +99,8 @@ func New(ctx context.Context, opts *Options, logger *zap.Logger, issuer *auth.Is
 		issuer:           issuer,
 		VersionNumber:    opts.VersionNumber,
 		metricsProjectID: metricsProjectID,
+		// TODO: Validate the syntax, if err or missing, then default to "CRON_TZ=America/Los_Angeles 0 0 * * 1"
```
begelundmuller, line 102: Can remove this TODO – see other comment about setting a default.

As for validation, the cron library will do it when it is invoked and return the error, so we don't need to do additional validation here IMO.

admin/admin.go
```diff
@@ -35,6 +36,7 @@ type Service struct {
 	issuer           *auth.Issuer
 	VersionNumber    string
 	metricsProjectID string
+	AutoScalerCron   string
```
begelundmuller, line 39: nit: Change to `AutoscalerCron` (two words, not three)

admin/admin.go
```diff
@@ -21,6 +21,7 @@ type Options struct {
 	VersionNumber      string
 	MetricsProjectOrg  string
 	MetricsProjectName string
+	AutoScalerCron     string
```
begelundmuller, line 24: Same nit, change to `AutoscalerCron`

begelundmuller (changes requested):

cli/cmd/admin/start.go
Kavinjsir: Sounds good!

admin/admin.go
Kavinjsir: Agree!

admin/admin.go
Kavinjsir: Good Catch!

PR #4572, opened 2024-04-10
Add CustomDashboardEmbed and generic Component

briangregoryholmes:
This PR reworks the organization of the CustomDashboard feature so that it can support a read-only embedding of a CustomDashboard as well as generic (non-chart) Components.


ericpgreen2 (approved):
Code looks good 👍

web-common/src/features/custom-dashboards/CustomDashboardPreview.svelte
```diff
@@ -164,65 +163,37 @@
...
+  {showGrid}
+  {gapSize}
+  {gridCell}
+  {radius}
+  {changing}
+  bind:contentRect
+  on:scroll={handleScroll}
+  on:click={deselect}
+>
+  {#each charts as chart, i (i)}
```
mindspank, line 179: NIT: Should use a more generic wording then charts as we will also render non-charts

PR #4561, opened 2024-04-09
Add Preview Button to Dashboard

briangregoryholmes:
This PR adds a new `Preview` button to the workspace for a dashboard yaml. As such, it changes the navigation behavior when clicking a dashboard asset to direct the user to the `/edit` page and removes the sidebar when viewing a dashboard, changes which required some restructuring of the `Dashboard` component.

As part of the refactor, the dashboard editing workspace has also been updated to flatten the tree structure and ensure that data access is orchestrated from the top down, limiting direct imports of state/writables/etc from child components.

Frankly, the layout across the app will simplify quite a bit once the file explorer is in place, so how things are organized in this PR is subject to change.

PR #4576, opened 2024-04-10
Make Breadcrumbs environment agnostic

briangregoryholmes:
This PR moves the Breadcrumbs component into `web-common` and makes it environment agnostic, allowing it to be added to the `web-local` Dashboard layout as a header. This is a preliminary step for #4561.


web-common/src/components/navigation/breadcrumbs/BreadcrumbItem.svelte
```diff
@@ -0,0 +1,81 @@
...
+  $: nextChild = $page.route.id?.split(regex)[depth + 1];
+</script>
+
+<li class="flex items-center gap-x-2 px-2">
+  <div class="flex flex-row gap-x-1 items-center">
+    {#if selected}
+      <a
+        href={!isCurrentPage && selected.href
+          ? selected.href + nextChild
+          : "#top"}
```
ericpgreen2, line 26: Confirming, it seems like "#top" is supposed to give the user feedback that they did indeed click the link, even though the link doesn't direct them anywhere?

ericpgreen2 (approved):
Code looks good, just on the dashboard page:
1. I'm missing the Dashboard breadcrumb
2. Clicking on the Project breadcrumb does not navigate me back to the Project page

image

ericpgreen2:
A couple more UXQA findings (though these might already be on your radar for a follow-on PR):
- The Rill icon overlaps the show/hide sidebar menu (though this won't be an issue when we move the dashboard preview to its own tab)
image
- With the breadcrumbs on local, we should remove the redundant title below the breadcrumbs / above the time controls. And we should move the Edit Metrics & Deploy buttons up into the top bar.
image

web-common/src/components/navigation/breadcrumbs/BreadcrumbItem.svelte
briangregoryholmes: I think it probably doesn't really matter, but #top is just acknowledging that the element is a link that references the current page as opposed to a fully qualified link that happens to be the same as the page you're on.

briangregoryholmes:
Slipped my mind to add the dashboards level to the `web-admin`. Added back now. Your second comment is on my radar for the next PR.

PR #4582, opened 2024-04-11
Runtime: Emit instance heartbeat events 

k-anshul:
Emits an event named `instance_heartbeat` with a `data_dir_size_bytes` attribute for the size of the instance's data directory.

Closes https://github.com/rilldata/rill-private-issues/issues/284


cli/cmd/runtime/start.go
```diff
@@ -91,6 +91,7 @@ type Config struct {
 	ActivitySinkKafkaBrokers string `default:"" split_words:"true"`
 	// Kafka topic of an activity client's sink
 	ActivitySinkKafkaTopic string `default:"" split_words:"true"`
+	EmitInstanceHeartbeat  bool   `default:"false" split_words:"true"`
```
begelundmuller: Let's not make this configurable at this stage. You can already disable all metrics (by setting`ActivitySinkType` to noop as we do in dev), and I don't think we need more granular control than that at this stage.

runtime/registry.go
```diff
@@ -502,3 +503,54 @@ func (r *registryCache) ensureProjectParser(ctx context.Context, instanceID stri
...
+}
+
+func newInstanceHeartbeatEmitter(rt *Runtime, duration time.Duration) *instanceHeartbeatEmitter {
+	return &instanceHeartbeatEmitter{
+		rt:       rt,
+		duration: duration,
+	}
+}
+
+func (i *instanceHeartbeatEmitter) emit() {
```
begelundmuller: This doesn't really require a separate struct – it can just be a function on `registryCache`, like `go r.emitHeartbeats()` in `newRegistryCache`. 

The registry cache already has a `baseCtx` which you can use instead of `closeChan` for cancellation.

runtime/registry.go
```diff
@@ -502,3 +503,54 @@ func (r *registryCache) ensureProjectParser(ctx context.Context, instanceID stri
...
+	ticker := time.NewTicker(i.duration)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			instances, err := i.rt.registryCache.list()
+			if err != nil {
+				i.rt.logger.Warn("failed to send instance heartbeat event, instance listing failed", zap.Error(err))
```
begelundmuller: That's worth an error-level log. Something would have to go very wrong for that to happen.

runtime/registry.go
```diff
@@ -502,3 +503,54 @@ func (r *registryCache) ensureProjectParser(ctx context.Context, instanceID stri
...
+				return
+			}
+			for _, instance := range instances {
+				var dataDirSize int64
+				f, err := os.Stat(filepath.Join(i.rt.opts.DataDir, instance.ID))
+				if err == nil {
+					dataDirSize = f.Size()
+				}
```
begelundmuller: Does that work? Might be wrong, but wouldn't think that does a recursive walk

runtime/registry.go
```diff
@@ -502,3 +503,54 @@ func (r *registryCache) ensureProjectParser(ctx context.Context, instanceID stri
...
+	ticker := time.NewTicker(i.duration)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			instances, err := i.rt.registryCache.list()
+			if err != nil {
+				i.rt.logger.Warn("failed to send instance heartbeat event, instance listing failed", zap.Error(err))
+				return
```
begelundmuller: Return here is weird since the error doesn't cause a crash. So it might look like it recovered. Probably better to `continue` and let it keep printing error messages once per minute.

runtime/registry.go
```diff
@@ -502,3 +503,54 @@ func (r *registryCache) ensureProjectParser(ctx context.Context, instanceID stri
...
+			}
+			for _, instance := range instances {
+				var dataDirSize int64
```
begelundmuller: Nesting getting a little deep here, consider pulling the instance-specific logic into a separate `emitHeartbeatForInstance(inst)`

runtime/registry.go
k-anshul: Cool. Added in `init` since that's where cache is populated.

begelundmuller (approved):

cli/cmd/runtime/start.go
```diff
@@ -208,6 +208,7 @@ func StartCmd(ch *cmdutil.Helper) *cobra.Command {
 				ControllerLogBufferSizeBytes: conf.LogBufferSizeBytes,
 				AllowHostAccess:              conf.AllowHostAccess,
 				DataDir:                      conf.DataDir,
+				EmitInstanceHeartbeat:        true,
```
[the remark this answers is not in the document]
k-anshul: I kept it to ensure we don't enable it for local and tests. 
For local I believe we don't want to send any events. Also `ActivitySinkType` is not `NOP` for local ?

runtime/registry.go
```diff
@@ -183,7 +184,9 @@ func (r *registryCache) init(ctx context.Context) error {
 			return err
 		}
 	}
-
+	if r.rt.opts.EmitInstanceHeartbeat {
+		r.emitHeartbeats()
+	}
```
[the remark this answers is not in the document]
k-anshul: OOPS. I keep missing spawning a new goroutine.

PR #4587, opened 2024-04-12
Add directories to file watcher and also populate kind for AI generated metrics view

AdityaHegde:
For asset explorer we need to get file watch events for directories. Adding `is_dir` to denote that an event is a directory.

Also updating generate metrics endpoint to populate `kind` to support generating it anywhere in the project.


begelundmuller (approved):
LGTM

Issue #4541, opened 2024-04-09
The Back button in the Add Source modal closes the dialog


ericpgreen2:
To reproduce:
1. Open the Add Source dialog
2. On page 1, select any connector
3. Click the back button
4. The dialog is closed, rather than directed back to page 1

PR #4591, opened 2024-04-12
Associate history entries with AddSourceModal state

briangregoryholmes:
This PR adds history entries associated with the state of the Add Source modal. This is something that is trivial and has a pretty clean API in SvelteKit 2.0, but is kind of cumbersome using native browser APIs. As such, this is kind of a patchy approach and another reason to move to 2.0 as soon as possible.

Closes: #4541


ericpgreen2 (approved):
Looks good 👍

PR #4593, opened 2024-04-12
Exporting limit error message

egor-ryashin:


runtime/queries/metricsview_aggregation.go
```diff
@@ -172,6 +172,10 @@ func (q *MetricsViewAggregation) Resolve(ctx context.Context, rt *runtime.Runtim
 				return err
 			}
 
+			if *q.Limit > 0 && int64(len(data)) > *q.Limit {
```
begelundmuller, line 175: Should handle the case where `q.Limit == nil` (to avoid nil access panics)

begelundmuller (approved):

PR #4609, opened 2024-04-15
fix double column handling in calcite go driver

pjain1:
see this - https://github.com/rilldata/calcite-avatica-go/pull/3


nishantmonu51 (commented):
👍

PR #4613, opened 2024-04-15
Update AI feature flag scope

briangregoryholmes:
This PR changes the scope of the AI feature flag from `user` to `system`, meaning it can't be set/overwritten via localStorage, URL or the YAML file.

Looking forward, the feature flags store should probably be broken up into three components: a `config` store that is project/organization specific and depends on data only available via the API, a `preferences` store for things that can be toggled on and off and are not intended to be hidden, and a `feature flags` store for things disabled by default and enabled manually. Also, there are certain things, like `adminServer` that don't need to be reactive or controlled during runtime that should be constants determined by environment variables.

### DIFF
--- a/web-common/src/features/feature-flags.ts
+++ b/web-common/src/features/feature-flags.ts
@@ -15,8 +15,9 @@ class FeatureFlag {
   private _system = false;
   private state = writable(false);
 
-  constructor(scope: "user" | "system" = "user") {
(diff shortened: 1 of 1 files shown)


ericpgreen2:
Can we set the AI flag to default to `true`?

briangregoryholmes:
@ericpgreen2 Added.

ericpgreen2 (approved):

PR #4617, opened 2024-04-16
Runtime: Support for time_range functions

k-anshul:
Added two function:
1. `time_range_start(duration, [unit, col])` and `time_range_end(duration, [unit, col])`
2. `time_range` will be added in later phases.


runtime/pkg/duration/duration.go
```diff
@@ -32,11 +32,11 @@ var (
 	}
 	daxOffsetNotations = map[string]StandardDuration{
 		"PP": {},
```
k-anshul, line 34: What should we do wrt `PP`? Add an explicit check that it is not allowed in time_range functions

k-anshul:
No clue why e2e tests are failing on this PR. The changes are only in `runtime/pkg` and pretty much self contained.

runtime/resolvers/metrics_sql.go
```diff
@@ -47,7 +47,12 @@ func newMetricsSQL(ctx context.Context, opts *runtime.ResolverOptions) (runtime.
 		return nil, err
 	}
 
-	compiler := metricssqlparser.New(ctrl, opts.InstanceID, opts.UserAttributes)
+	sqlArgs := &sqlArgs{}
+	if err := mapstructure.Decode(opts.Args, sqlArgs); err != nil {
+		return nil, err
+	}
+
+	compiler := metricssqlparser.New(ctrl, opts.InstanceID, opts.UserAttributes, sqlArgs.Priority)
```
begelundmuller, line 55: A little confusing to overload `sqlArgs` – maybe add an explicit `metricsSQLArgs` type?

begelundmuller (approved):

runtime/resolvers/metrics_sql.go
k-anshul: Sure. Fixed.

AndrewRTsao:
@k-anshul Should we add a snippet to our docs to call out support for time_range functions?

PR #4619, opened 2024-04-16
Improve drag drop UX in file explorer and other fixes

AdityaHegde:
- [x] Improve drag drop UX.
  - Add a small distance before triggering the drag
  - Add icons to the dragged element.
  - Disable drag-drop for rill.yaml
- [x] Optimise the list files APIs to not make multiple calls.


ericpgreen2 (approved):
Approving for you to merge after fixing the above-mentioned regression.

PR #4625, opened 2024-04-17
Remove project label from Rill Dev Breadcrumbs

briangregoryholmes:
To match the updated Figma mocks, this PR removes the project label from the Rill Developer Breadcrumbs. It also turns the Rill logo into a link that takes you back to the root page.


ericpgreen2:
Nit: looks like it's missing some padding between the Rill icon and the dashboard name

Observed:
image

Design:
image

ericpgreen2 (approved):

PR #4629, opened 2024-04-17
fix: flicker during rename in the new file explorer

AdityaHegde:
In the new file explorer we use name+kind to decide what type of editor we use. When renaming it switching from undefined to a valid one. This causes a flicker from entity's editor to fallback editor back to entity's editor. By proactively adding name+kind to the target path we can avoid this.


ericpgreen2 (approved):
Much better UX!

PR #4630, opened 2024-04-17
TDD chart improvements

djbarnwal:
This PR - 
- adds correct tooltip content for the charts 
- enables all chart type by default
- change the style of the tooltip box
- displays proper label for the fields in the tooltip
- Improves default area chart

### DIFF
--- a/web-common/src/features/charts/render/VegaLiteRenderer.svelte
+++ b/web-common/src/features/charts/render/VegaLiteRenderer.svelte
@@ -39,8 +39,22 @@
   />
 {/if}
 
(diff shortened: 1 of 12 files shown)


ericpgreen2 (approved):
Code looks good.

Though when I select the Stacked Bar and Stacked Area charts, I don't see any stacking. I think you mentioned that comparisons are still a TODO, so I think that's why?

image

image

djbarnwal:
Time comparison is still a TODO. You need to select dimension values to start comparing them and see them stacked.

PR #4634, opened 2024-04-17
Move to flat ESLint config

briangregoryholmes:
This PR moves our ESLint configuration to the new [flat config](https://eslint.org/blog/2022/08/new-config-system-part-2/) format, partly in preparation for [ESLint v9](https://eslint.org/blog/2024/04/eslint-v9.0.0-released/) and partly to resolve a few issues with our monorepo setup. Legacy config versions are being deprecated.

To trigger linting across all repos, it changes all `rel="noreferrer"` attributes to the correct `rel="noreferrer noopener` and adds a key to the each block in `web-auth`'s `Auth` component.

To enable the ESLint VSCode extension, you'll need to either install the latest prerelease in the marketplace or turn on `eslint.experimental.useFlatConfig`, see: https://github.com/microsoft/vscode-eslint/issues/1644

Motivation: The use of `__dirname` in the current configuration does not correctly reference the corresponding `tsconfig.json` for each project, which is at least one of the issues causing false positive ESLint warnings. The options to solve that are either move to the new flat config or have an additional ESLint configuration per repo.


ericpgreen2 (approved):
Sounds good to me.

PR #4649, opened 2024-04-18
Fix rename of a folder not correctly identifying inner entities are renamed

AdityaHegde:
When a folder is renamed files within the new folder is considered as new files rather than renames. This is an attempt at fixing it by going through all files in the cache that have the prefix of a deleted folder.

### DIFF
--- a/runtime/compilers/rillv1/parser.go
+++ b/runtime/compilers/rillv1/parser.go
@@ -246,6 +246,16 @@ func (p *Parser) IsIgnored(path string) bool {
 	return !pathIsYAML(path) && !pathIsSQL(path) && !pathIsDotEnv(path)
 }
 
+func (p *Parser) FindFilesInDir(dir string) []string {
+	var files []string
+	for path := range p.resourcesForPath {
+		if strings.HasPrefix(path, dir) {
+			files = append(files, path)
+		}
+	}
+	return files
+}
+
 // reload resets the parser's state and then parses the entire project.
 func (p *Parser) reload(ctx context.Context) error {
 	// Reset state
--- a/runtime/compilers/rillv1/parser.go
+++ b/runtime/compilers/rillv1/parser.go
@@ -246,14 +246,24 @@ func (p *Parser) IsIgnored(path string) bool {
 	return !pathIsYAML(path) && !pathIsSQL(path) && !pathIsDotEnv(path)
 }
 
-func (p *Parser) FindFilesInDir(dir string) []string {
-	var files []string
+func (p *Parser) IsInRestrictedDir(path string) bool {
+	return strings.HasPrefix(path, "/tmp")
+}
+
+// TrackedPathsInDir returns the paths under the given directory that the parser currently has cached results for.
+func (p *Parser) TrackedPathsInDir(dir string) []string {
+	// Ensure dir has a trailing path separator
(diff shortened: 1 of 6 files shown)


runtime/compilers/rillv1/parser.go
```diff
@@ -246,6 +246,16 @@ func (p *Parser) IsIgnored(path string) bool {
...
 
+func (p *Parser) FindFilesInDir(dir string) []string {
+	var files []string
+	for path := range p.resourcesForPath {
+		if strings.HasPrefix(path, dir) {
+			files = append(files, path)
+		}
+	}
+	return files
+}
```
begelundmuller, line 257: 1. Needs docstring
2. Needs to ensure a trailing slash on the dir

Suggestion:

```suggestion
// TrackedPathsInDir returns the paths under the given directory that the parser currently has cached results for.
func (p *Parser) TrackedPathsInDir(dir string) []string {
	// Ensure dir has a trailing path separator
	if len(dir) > 0 && dir[len(dir)-1] != '/' {
		dir += '/'
	}

	// Find paths
	var paths []string
	for path := range p.resourcesForPath {
		if strings.HasPrefix(path, dir) {
			paths = append(paths, path)
		}
	}
	return paths
}
```

runtime/reconcilers/project_parser.go
```diff
@@ -173,16 +174,25 @@ func (r *ProjectParserReconciler) Reconcile(ctx context.Context, n *runtimev1.Re
...
+		deletedPaths := make([]string, 0)
 		for _, e := range events {
 			if e.Dir {
 				continue
 			}
 			if parser.IsIgnored(e.Path) {
+				if !strings.HasPrefix(e.Path, "/tmp") && e.Type == runtimev1.FileEvent_FILE_EVENT_DELETE {
+					// only check for deleted folders that are not inside /tmp
+					deletedPaths = append(deletedPaths, parser.FindFilesInDir(e.Path)...)
+				}
```
begelundmuller, line 186: 1. Skipping `/tmp` should be done separately at the start of the loop (might make sense to have a general exclude list global)
2. Why is the directory deletion case not handled under `if e.Dir { ...`?

runtime/reconcilers/project_parser.go
```diff
@@ -173,16 +174,25 @@ func (r *ProjectParserReconciler) Reconcile(ctx context.Context, n *runtimev1.Re
...
 				continue
 			}
 			changedPaths = append(changedPaths, e.Path)
 		}
 
+		// prepend deleted paths and dedupe
+		deletedPaths = append(deletedPaths, changedPaths...)
+		changedPaths = arrayutil.Dedupe(deletedPaths)
```
begelundmuller, line 194: If we anyway do a dedupe, why not just remove `deletedPaths`, append directly to `changedPaths`, and do a dedupe on it? (As an optimization, would make sense to have a `bool` for whether deleted paths were added to avoid a dedupe in regular cases.)

begelundmuller (changes requested):

runtime/reconcilers/project_parser.go
AdityaHegde: There were previous iterations where it made sense to have a separate list. Will remove it

begelundmuller (approved):

PR #4650, opened 2024-04-18
Stop adding additional forward slash in url and other fixes

AdityaHegde:
- [x] Stop adding an additional leading forward slash in the page url
- [x] Fix custom dashboard getting an extra `.yaml`
- [x] Add warning when moving entity out of natural folder.


web-common/src/features/entity-management/actions.ts
```diff
@@ -40,6 +49,19 @@ export async function renameFileArtifact(
...
+    if (
+      fromResName?.kind &&
+      FolderToResourceKind[removeLeadingSlash(fromFolder)] ===
+        fromResName.kind &&
+      fromFolder !== toFolder
+    ) {
+      notifications.send({
+        message: `Moving ${fromName} out of native folder. Please make sure to add "kind" key to denote the type.`,
+      });
+    }
```
ericpgreen2, line 64: This notification is firing even when moving a file into a subdirectory, say from `models/abc.sql` to `models/sub-dir/abc.sql`. We don't need to manually add a `kind` in subdirectories, right?

ericpgreen2 (approved):
👍 

One UX regression: the active file is no longer highlighted in the sidebar. Here I should see `model.sql` in an active state:

image

web-common/src/features/entity-management/actions.ts
AdityaHegde: Good point. It only needs the root folder to be `models`,`sources` etc

PR #4652, opened 2024-04-18
Use a single editor instance for fallback editors

AdityaHegde:
This adds support to change extensions dynamically. For now we only use it in the fallback editor. But we would need some significant refactor to achieve this in other entities' editors.

Also removes the dedicated route for rill.yaml


web-common/src/features/editor/Editor.svelte
```diff
@@ -50,8 +50,26 @@
...
+          // any extensions passed as props
+          ...newExtensions,
+          // establish a basic editor
+          base(),
+          // this will catch certain events and dispatch them to the parent
+          bindEditorEventsToDispatcher(dispatch),
+        ],
+      }),
+    );
+  }
```
ericpgreen2, line 67: This seems to work, just noting this docs page has a couple different recommended ways to change extensions/languages dynamically: https://codemirror.net/examples/config. See "Top-level reconfiguration" and "Automatic language detection".

[This ChatGPT answer](https://chat.openai.com/share/e/1e74c4d6-f66e-4524-9e9d-09c5e1dec202) also proposed the approach stated in "Top-level reconfiguration".

ericpgreen2 (approved):

web-common/src/features/editor/Editor.svelte
AdityaHegde: Ideally ya we should use compartments. But until we know how the end refactor will look we can replace the entire list of extensions between editors.

PR #4653, opened 2024-04-18
Rows Accordion: use Resizer, remove drag action

briangregoryholmes:
With the new Preview Table component merged, this PR simplified the RowsViewerAccordion and updates it to use the new Resizer component, solving a few small UI issues.

The drag action has also been removed since it is no longer used in the codebase.


ericpgreen2 (approved):
A couple nits:
1. Looks like the unfilled space switched from gray to white. I'm not sure which is better, but calling it out in case you hadn't seen it. 
2. The right-most column in the table could use a right border:
image

PR #4657, opened 2024-04-18
Add preliminary components to cloud custom dashboard page

briangregoryholmes:
Adds structure to the custom dashboard page for `web-admin`. It is expected that this route may throw errors depending on your chart/custom-dashboard spec as Chart data fetching, which is currently using a manual client, throws uncaught errors.

This also removes a conditional check that was being used to apply `h-screen` on the root component for `web-admin`. I didn't see any obvious issues caused by removing this, though we should probably rework the layout for `web-admin` to allow for a sticky `TopNavigationBar` and scrollable tables.

### DIFF
--- a/web-admin/src/routes/+layout.svelte
+++ b/web-admin/src/routes/+layout.svelte
@@ -1,7 +1,6 @@
 <script lang="ts">
   import { beforeNavigate } from "$app/navigation";
   import { page } from "$app/stores";
-  import { isMetricsExplorerPage } from "@rilldata/web-admin/features/navigation/nav-utils";
   import { initCloudMetrics } from "@rilldata/web-admin/features/telemetry/initCloudMetrics";
   import NotificationCenter from "@rilldata/web-common/components/notifications/NotificationCenter.svelte";
   import {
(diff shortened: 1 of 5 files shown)


web-admin/src/routes/+layout.svelte
```diff
@@ -49,9 +42,7 @@
 
 <RillTheme>
   <QueryClientProvider client={queryClient}>
-    <main
-      class="flex flex-col min-h-screen {onMetricsExplorerPage && 'h-screen'}"
-    >
+    <main class="flex flex-col min-h-screen h-screen">
```
ericpgreen2, line 45: I like the idea to simplify this, but one regression is that pages w/ scroll no longer have a bottom padding

Before:
image

After:
image

ericpgreen2 (approved):
Nice 👍 

A somewhat unrelated nit I just found navigating between dashboards. The background of the `CustomDashboardEmbed` is subtly a different color from the Metrics Explorer header. Looks like it's `bg-gray-100` vs `bg-slate-50`.

web-admin/src/routes/+layout.svelte
briangregoryholmes: Added bottom padding to the content container which seems to be used on all the routes that need it.

PR #4662, opened 2024-04-19
File Explorer: small cleanups

ericpgreen2:
- Applies `flex-none` so carets and icons don't resize when the sidebar's width changes
- Uses `<a>` elements not `<button>` element for file navigation; moves the context menu `<button>` outside of the `<a>` tag
- Corrects the hover state for dotfiles
- Modifies e2e tests to identify files with full file path
- Adds a "No tables found" note for new projects without any tables
- Improves the default Model and Dashboard files
- Adds `/.git` as a protected directory


web-common/src/features/file-explorer/NavFile.svelte
```diff
@@ -51,36 +50,34 @@
...
   class="w-full h-6 group pr-2 text-left flex justify-between gap-x-1 items-center
   {isProtectedDirectory || isDotFile
-    ? 'text-gray-500'
+    ? 'text-gray-500 hover:text-gray-500'
     : 'text-gray-900 hover:text-gray-900'}
   {isCurrentFile ? 'bg-slate-100' : ''} 
   font-medium hover:bg-slate-100"
   {id}
-  on:click={() => navigate(filePath)}
+  href={`/files${filePath}`}
```
AdityaHegde, line 68: This might be a problem for firing telemetry. We cant really add beforeNavigation since we wont have the source of the action (`left-nav` or `cta` etc)

briangregoryholmes: We should eventually be instrumenting telemetry from a single place in the code, a la https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/web/opentelemetry-instrumentation-user-interaction?ref=asserts.ai

ericpgreen2: Merged `main` and resolved the telemetry conflict – used both `href` and `on:click` as in `DashboardCTAs.svelte`

ericpgreen2: @briangregoryholmes, very much agree. We should consider along with this workstream: https://github.com/rilldata/rill-private-issues/issues/163

briangregoryholmes:
Not sure if this PR was meant to address it, but the nav entries do still have nested interactive elements. The dropdown trigger is inside the <a> tag.

ericpgreen2:
> Not sure if this PR was meant to address it, but the nav entries do still have nested interactive elements. The dropdown trigger is inside the `<a>` tag.

Oh, missed that – fixed.

AdityaHegde (approved):

PR #4684, opened 2024-04-22
Rill Developer: Hide Sources & Models for ClickHouse & Druid

ericpgreen2:
- Sources & Models are not currently supported for ClickHouse- and Druid-backed projects
- This PR hides the Sources & Models options in the "Add [resource]" button when the application is started with either of these OLAP engines.

### DIFF
--- a/web-common/src/features/entity-management/AddAssetButton.svelte
+++ b/web-common/src/features/entity-management/AddAssetButton.svelte
@@ -21,6 +21,7 @@
   import { directoryState } from "../file-explorer/directory-store";
   import { handleEntityCreate } from "../file-explorer/new-files";
   import { addSourceModal } from "../sources/modal/add-source-visibility";
+  import { useIsModelingSupportedForCurrentOlapDriver } from "../tables/selectors";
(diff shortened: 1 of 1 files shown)


AdityaHegde (approved):

AndrewRTsao:
@ericpgreen2 Will this still be true now that Pinot is an available OLAP engine?

ericpgreen2:
> @ericpgreen2 Will this still be true now that Pinot is an available OLAP engine?

Oh, we currently have [this configuration](https://github.com/rilldata/rill/blob/release-0.44/web-common/src/features/tables/olap-config.ts#L1-L2) to determine whether the OLAP driver supports sources & models. I haven't heard – should we add Pinot to that list?

begelundmuller:
> Might be worth double checking with @nishantmonu51 or @begelundmuller but we just added [support for Pinot](https://github.com/rilldata/rill/pull/4534) as a new OLAP engine and I don't think we can support modeling there either.

Yeah it doesn't support modeling at the moment.

PR #4685, opened 2024-04-22
Refactor: Leaderboard Grid

briangregoryholmes:
This PR removes the use of the `VirtualizedGrid` component and instead relies on the `IntersectionObserver` API to lazy load Leaderboard queries.

This was done to reduce code complexity and align the implementation with the lazy loading approach that will  be taken with the time series charts.

Additionally, the loading state of the leaderboards has been changed the following:

Screenshot 2024-04-22 at 5 17 58 PM

This is not a requirement of the PR and is open to feedback, but it does solve two related issues:

1) The scrollbar no longer erratically resizes as data "beneath the fold" is fetched
2) For most leaderboards, the vertical height of the loading state now more closely matches the loaded state

This is somewhat of a stopgap UI improvement and loading states for the Leaderboards will be further refined once they're converted to HTML tables.


ericpgreen2 (approved):
I like it. The taller loading state also improves loading performance, especially for big Druid-backed dashboards. Currently when a dashboard is opened, we unnecessarily fire `toplist` queries for bottom-of-the-screen leaderboards that are ultimately pushed off-screen & unmounted when the data arrives for the top-of-the-screen leaderboards.

Issue #4686, opened 2024-04-22
When adding multiple unnest dimensions, dashboard hangs


cohenscottr:
**Describe the bug**
When setting unnest: true for multiple dimensions, the dashboard does not load.

Validated by trying each metric individually and dashboard succeeds. 

**To Reproduce**
1. Utilize the attached data
2. Auto create dash
3. Add fields for:

  - label: Req Model Api
    column: req_api
    description: ""
    unnest: true
  - label: Req Mimes
    column: req_mimes
    description: ""
    unnest: true
    
Dashboard crashes

[unnest_model_20240422164516.csv](https://github.com/rilldata/rill/files/15069500/unnest_model_20240422164516.csv)

PR #4690, opened 2024-04-23
Fix issues around unnest in dashboard

### LINKED ISSUE #4691: Fix: Having multiple unnest dimensions shows a blank screen in preview
Having multiple unnest dimensions shows a blank screen in preview, the metrics view schema query errors out.
Add this to a dashboard created from the AdBids.csv file source,
```
  - label: Domain Parts
    name: domain_parts
    expression: "string_split(domain, '.')"
    unnest: true
  - label: Domain Parts 2
    name: domain_parts_2
    expression: "string_split(domain, '.')"
    unnest: true
```
Error,
```
Binder Error: Duplicate alias "tbl" in query!
```

Another issue is that this error is not caught in validation. Validation code doesnt honour the unnest boolean.

AdityaHegde:
closes #4691 , #4686

- [x] Fix multiple dimensions having unnest breaking certain queries.
- [x] Update validation to check for unnest

### DIFF
--- a/runtime/drivers/olap.go
+++ b/runtime/drivers/olap.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/google/uuid"
 	"github.com/jmoiron/sqlx"
 	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
 )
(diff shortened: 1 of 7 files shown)


runtime/drivers/olap.go
```diff
@@ -182,3 +183,36 @@ func (d Dialect) EscapeTable(db, schema, table string) string {
...
+	}
+
+	return sel, fmt.Sprintf(`, LATERAL UNNEST(%s) %s(%s)`, dim.Expression, unnestTableName, unnestColName)
+}
+
+func tempName(prefix string) string {
+	return prefix + strings.ReplaceAll(uuid.New().String(), "-", "")
+}
+
+func (d Dialect) MetricsViewDimensionExpression(dimension *runtimev1.MetricsViewSpec_DimensionV2) string {
```
begelundmuller, line 208: Is it correct to call this directly? Could seem like unnest is not handled in those places where it's used directly.

AdityaHegde: Good point. I just renamed it to export the method. Lemme take a look at its usages.

AdityaHegde: In all the cases it is used in expressions and not select so it wont be a problem. We can do a follow up to update one usage in the metrics view comparison query for druid as a follow up.

begelundmuller (approved):

PR #4694, opened 2024-04-23
File explorer add support for force deleting a non-empty directory

AdityaHegde:
- [x] Add a confirmation to force delete a non-empty directory
- [x] Deleting a folder for an active file should redirect to `/`


begelundmuller (approved):
Approving from backend perspective

web-common/src/features/file-explorer/ForceDeleteConfirmation.svelte
```diff
@@ -0,0 +1,28 @@
...
+
+  function handleClose() {
+    open = false;
+  }
+</script>
+
+<Dialog on:close={handleClose} {open}>
+  <div slot="title">
+    Folder is not empty. Are you sure you want to delete everything?
+  </div>
```
ericpgreen2, line 16: Can we change the title to: "Are you sure you want to delete everything?"

And add a `body` section: "This folder is not empty. All contained items will be deleted."

web-common/src/features/file-explorer/ForceDeleteConfirmation.svelte
```diff
@@ -0,0 +1,28 @@
+<script lang="ts">
+  import { Button } from "@rilldata/web-common/components/button";
+  import Dialog from "@rilldata/web-common/components/dialog/Dialog.svelte";
```
ericpgreen2, line 3: (Nice-to-have, but by no means a blocker): it looks like ShadCN-Svelte has a specific [Alert Dialog component](https://www.shadcn-svelte.com/docs/components/alert-dialog) for these use cases.

web-common/src/features/file-explorer/ForceDeleteConfirmation.svelte
ericpgreen2, line 1: Nit: mind including `Dialog` in the component name, e.g. `ForceDeleteConfirmationDialog.svelte`?

web-common/src/features/file-explorer/transform-file-list.ts
```diff
@@ -61,3 +62,12 @@ export function transformFileList(files: V1DirEntry[]): Directory {
 
   return rootDirectory;
 }
+
+export function findDirectory(root: Directory | undefined, filePath: string) {
```
ericpgreen2, line 66: I think it'd be simpler if this didn't accept an `undefined` root directory

web-common/src/features/file-explorer/utils.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export function isCurrentActivePage(
+  filePath: string,
+  currentFile: string | undefined,
+  isDir: boolean,
+) {
+  filePath = removeLeadingSlash(filePath);
+  currentFile = removeLeadingSlash(currentFile ?? "");
+  return (
+    currentFile !== "" &&
```
ericpgreen2, line 11: Adding a guard clause at the top of the function could be better:
```
if (currentFile === undefined) return false
```

Further, I'd wonder if the function even needs to accept an `undefined` `currentFile`.

ericpgreen2 (changes requested):
I played around with it, and UXQA-wise:
1. I'm not able to delete a non-empty directory. JAM: https://jam.dev/c/089c392e-78d9-4144-9470-a3a93be0173f
2. I'm seeing this error in the console
image

web-common/src/features/file-explorer/ForceDeleteConfirmation.svelte
AdityaHegde: Good idea. Seems like it has some dependancy on button. Lets add it in a separate PR and move all alerts to use this.

web-common/src/features/file-explorer/utils.ts
AdityaHegde: `$page.params.file` could be undefined when user is on home. Moving that check here for now. Felt cleaner to deal with `$page` here itself.

AdityaHegde:
I couldnt reproduce the 1st problem. Did you restart the backend server after checking this code out?

ericpgreen2:
One UX issue:

When viewing a file and force deleting its directory, I see our current 404/error page. I'd expect to be re-directed to root. JAM: https://jam.dev/c/0d3b3c42-88c0-45be-878e-a209819d38d6

ericpgreen2 (approved):

PR #4695, opened 2024-04-23
Add support for Git hash as a valid deployment runtime version

kaspersjo:


admin/deployments.go
```diff
@@ -430,6 +434,25 @@ func (s *Service) NewDeploymentAnnotations(org *database.Organization, proj *dat
...
+func (s *Service) ValidateRuntimeVersion(ver string) error {
+	// Verify version is a valid SemVer, a full Git commit hash or 'latest'
+	if ver != "latest" {
+		_, err := version.NewVersion(ver)
+		if err != nil {
+			// Not a valid SemVer, try as a full commit hash
+			matched, err2 := regexp.MatchString(`\b([a-f0-9]{40})\b`, ver)
+			if err2 != nil {
+				return err2
+			}
```
begelundmuller, line 446: nit: can just reuse the `err` variable instead of declaring `err2`

begelundmuller (approved):

admin/deployments.go
kaspersjo: Updated

PR #4696, opened 2024-04-23
DuckDb HAVING clause with the same alias name

egor-ryashin:


proto/rill/runtime/v1/expression.proto
```diff
@@ -10,6 +10,7 @@ message Expression {
     google.protobuf.Value val = 2;
     Condition cond = 3;
   }
+  bool having = 4;
```
begelundmuller, line 13: This is an internal implementation detail, so should not leak into the API type. Can we move it to a separate `bool` passed internally alongside the expression instead?

begelundmuller (changes requested):

begelundmuller (commented):
LGTM

begelundmuller (approved):

PR #4697, opened 2024-04-23
Backend number formatter

esevastyanov:
The code is based on the logic implemented on the frontend
Type and file names are kept mostly the same for consistency and ease of updating


runtime/pkg/formatter/count_digits.go
```diff
@@ -0,0 +1,15 @@
...
+func countDigits(numStr string) int {
+	re := regexp.MustCompile(`\D`)
+	return len(re.ReplaceAllString(numStr, ""))
+}
+
+// countNonZeroDigits counts all non-zero numeric digits in a string.
+func countNonZeroDigits(numStr string) int {
+	re := regexp.MustCompile(`[^1-9]`)
+	return len(re.ReplaceAllString(numStr, ""))
+}
```
begelundmuller, line 15: If you move the `MustCompile` results into globals, it will only need to compile the regex strings once instead of on every invocation

runtime/pkg/formatter/format_measure_value.go
```diff
@@ -0,0 +1,86 @@
...
+	formatter
+	logger *zap.Logger
+}
+
+func (f *MeasureValueFormatter) Format(value any) string {
+	str, err := f.stringFormat(value)
+	if err == nil {
+		return str
+	}
+	f.logger.Warn("failed to format value, returning non-formatted value", zap.Any("value", value), zap.Error(err))
```
begelundmuller, line 25: We should avoid logging from this util library. Better to return an error and then log the error in the caller if desired.

runtime/pkg/formatter/format_measure_value.go
```diff
@@ -0,0 +1,86 @@
...
+	return &MeasureValueFormatter{f, logger}, nil
+}
+
+func NewMeasureValuePresetFormatter(preset string, useUnabridged bool, logger *zap.Logger) (*MeasureValueFormatter, error) {
+	f, err := presetFormatter(preset, useUnabridged)
+	if err != nil {
+		return nil, err
+	}
+	return &MeasureValueFormatter{f, logger}, nil
+}
```
begelundmuller, line 60: Can we decouple this package from the API types and just focus on the format options? For example, having something like `NewFormatter(opts)` or `NewPresetFormatter(preset, unabridged)` and `NewD3Formatter(format)`?

runtime/pkg/formatter/format_with_order_of_magnitude.go
```diff
@@ -0,0 +1,129 @@
...
+	// Set to true to pad with insignificant zeros.
+	// Integers will be padded with zeros if this is set.
+	padInsignificantZeros bool,
+	// Set to `true` to leave a trailing "." in the case
+	// of non-integers formatted to e0 with 0 fraction digits.
+	// Even if this is `true` integers WILL NOT be formatted with a trailing "."
+	trailingDot bool,
+	// strip commas from output?
+	stripCommas bool,
+) numberParts {
```
begelundmuller, line 54: Use an options type here, e.g. `struct formatNumWithOrderOfMagOptions`. If you need to line break a function definition, it's a good time to switch to an options type :)

runtime/pkg/formatter/intervals.go
```diff
@@ -0,0 +1,219 @@
...
+ *
+ * The strategy is to:
+ * - show two digits of precision
+ * - prefer to show two integer digits in a smaller unit
+ * - if that is not possible, show a floating point number in a larger unit with one digit of precision (e.g. 1.2 days)
+ *
+ * see https://www.notion.so/rilldata/Support-display-of-intervals-and-formatting-of-intervals-e-g-25-days-in-dashboardsal-data-t-8720522eded648f58f35421ebc28ee2f
+ */
+
+func formatMsInterval(ms float64) string {
```
begelundmuller, line 54: Use the usual Go-style comments (`//` on each line). And avoid an empty line before the function definition (it decouples the docstring so the editor hints disappear). Applies a few places.

begelundmuller (changes requested):

begelundmuller (approved):
LGTM

PR #4704, opened 2024-04-24
Fix label for dimensions in pivot

djbarnwal:
For dimension label in pivot where name and column are different.

Fixes https://github.com/rilldata/rill-private-issues/issues/248


ericpgreen2 (approved):
It looks like this fixed the issue in the pivot table body. Though, when coming from TDD, I'm seeing the measure & dimension pills are using the `name` not the `label`. Approving for you to merge when you like.

image

PR #4708, opened 2024-04-24
Add ignore_paths support in rill.yaml

AdityaHegde:
We want to allow users to ignore certain paths through rill.yaml. A new config `ignore_paths` will be added.

- [x] Initial instantiation of repo should include `ignore_paths`.
- [x] Add support for `ignore_paths` in file driver.
- [x] Add support for `ignore_paths` in admin driver.


runtime/compilers/rillv1/init.go
```diff
@@ -18,7 +18,8 @@ func IsInit(ctx context.Context, repo drivers.RepoStore, instanceID string) bool
 func InitEmpty(ctx context.Context, repo drivers.RepoStore, instanceID, title string) error {
 	mockUsersInfo := "# These are example mock users to test your security policies.\n# For more information, see the documentation: https://docs.rilldata.com/manage/security"
 	mockUsers := "mock_users:\n- email: john@yourcompany.com\n- email: jane@partnercompany.com"
-	rillYAML := fmt.Sprintf("compiler: %s\n\ntitle: %q\n\n%s\n\n%s", Version, title, mockUsersInfo, mockUsers)
+	ignoredPaths := "ignored_paths:\n- /.idea\n- /.vscode" // TODO: get a list from product
+	rillYAML := fmt.Sprintf("compiler: %s\n\ntitle: %q\n\n%s\n\n%s\n\n%s", Version, title, mockUsersInfo, mockUsers, ignoredPaths)
```
begelundmuller, line 22: What do you think about not adding it by default to avoid clutter? And just describing it in the docs. (Assuming our defaults are good enough for most people.)

runtime/drivers/repo.go
```diff
@@ -46,3 +47,31 @@ type DirEntry struct {
 	Path  string
 	IsDir bool
 }
+
+// ignoredPaths is a list of paths that are ignored by the parser.
+var ignoredPaths = []string{
+	"/tmp",
+	"/.git",
+	"/node_modules",
+}
```
begelundmuller, line 56: Let's also add `.DS_Store`.

Can you think of others that would be nice?

begelundmuller (changes requested):

runtime/compilers/rillv1/init.go
AdityaHegde: This was the ask from product. But if we have hardcoded paths then this might not be necessary. @mindspank Any comments?

mindspank: Hmm feels like with our hardcoded paths we don't need defaults in rill.yaml

runtime/drivers/repo.go
begelundmuller: Having editor folders here makes sense to me – don't see any reason people would ever want to edit `.vscode` inside of Rill :)

AdityaHegde: Cool. I will move `.vscode` and `.idea` here from the init.

begelundmuller (approved):
LGTM

PR #4713, opened 2024-04-24
Rill Developer: Fix Welcome workflow

ericpgreen2:
Upon selecting an example project, the user is now directed to the dashboard YAML, rather than to the dashboard Preview. This both gives the user context for what a Rill project is and it buys time for the chosen data to load. The "Preview" button is now disabled until the resource is ready (notably until all dependent sources have been ingested).

In its implementation, this PR simplifies `(viz)/dashboard/[name]/+page.svelte` and mirrors the `+page.ts` data loading approach taken in `(viz)/custom/[name]/+page.svelte`.

### DIFF
--- a/web-common/src/features/welcome/ProjectCards.svelte
+++ b/web-common/src/features/welcome/ProjectCards.svelte
@@ -59,7 +59,6 @@
       },
     });
     await goto(firstPage);
-    //  consider additionally opening up the project in a new tab
   }
 </script>
 
--- a/web-common/src/features/welcome/ProjectCards.svelte
+++ b/web-common/src/features/welcome/ProjectCards.svelte
@@ -21,14 +21,14 @@
       description: "Monitoring cloud infrastructure",
       image:
         "bg-[url('$img/welcome-bg-cost-monitoring.png')] bg-no-repeat bg-cover",
-      firstPage: "/dashboard/customer_margin_dash",
+      firstPage: "/files/dashboards/customer_margin_dash.yaml",
     },
     {
       name: "rill-openrtb-prog-ads",
       title: "OpenRTB Programmatic Ads",
       description: "Real-time Bidding (RTB) advertising",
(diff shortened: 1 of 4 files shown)


briangregoryholmes:
This is great. Couple things that this brings up, though they may not need to be addressed now.

1) We should probably get some better error pages at this point. I know Janet's been working on one. Not sure of the status.
2) There used to be some functionality that would redirect you to the `/edit` page in certain conditions. Not sure that it's strictly necessary, but this PR doesn't necessarily cover the case where you have a successful resource query, but the resource has an error (see below). Again, not strictly an issue, but we should either redirect or have a Dashboard error message in a case like this.
3) It's probably time to add a call to `$app/navigation`'s invalidate function in the watch request clients so that these load functions run again when necessary. Though, we do have to resolve this issue with extraneous events first (see: https://rilldata.slack.com/archives/CTCJ58H3M/p1712333249276039).

briangregoryholmes (approved):

ericpgreen2:
> 1. We should probably get some better local error pages at this point. I know Janet's been working on one. Not sure of the status. Are the cloud ones generic enough to reuse?

Yeah, agreed, we can move the Cloud pages to `web-common` and re-use those.

> 2. There used to be some functionality that would redirect you to the `/edit` page in certain conditions. Not sure that it's strictly necessary, but this PR doesn't cover the case where you have a successful resource query, but the resource has an error (see below). Again, not a dealbreaker, but we should either redirect or have a Dashboard error message in a case like this.

Yeah, I removed that redirect because, if you're editing the dashboard YAML in an external IDE, that redirect would be quite annoying. I have an improvement in error handling that uses the Cloud error page – I'll submit that in a follow-on PR.

> 3. It's probably time to add a call to `$app/navigation`'s invalidate function in the watch request clients so that these load functions run again when necessary. Though, we do have to resolve this issue with extraneous events first (see: https://rilldata.slack.com/archives/CTCJ58H3M/p1712333249276039).

I agree. In general our resource invalidation code could use some cleanup, so I'll leave that project out of this PR.

PR #4715, opened 2024-04-25
Cleanup old usage of ListFiles and using the folder name to identify entities

AdityaHegde:
In quite a few place we were still using the old method of using folder name to identify entities. Instead fetching all file names.

This will have the side effect of needing unique file name across all entities. This should be dealt with in a follow up.


ericpgreen2:

ericpgreen2:
I like this cleanup. I'm not sure if this was supposed to address it, but I'm still seeing the name collision error that [Andrew reported in Slack](https://rilldata.slack.com/archives/C02T907FEUB/p1713982682749269). [Here's a JAM](https://jam.dev/c/9f2392a3-d5f7-427b-81f7-b11887623689).

Given it's possible to override the `name` key within any YAML/SQL file, I'm wondering if we should be fetching existing names via `ListResources` rather than `ListFiles`. Thoughts?

web-common/src/features/custom-dashboards/AddChartMenu.svelte
```diff
@@ -1,23 +1,24 @@
 <script lang="ts">
+  import { useAllFileNames } from "@rilldata/web-common/features/entity-management/file-selectors";
```
ericpgreen2, line 2: **Issue**: The addition of 'useAllFileNames' for fetching file names introduces a broader scope than the original 'useChartFileNames', potentially including files that are not charts. This could lead to unintended files being listed in the chart addition menu. <br> **Fix**: Implement a filtering mechanism to ensure only chart-related files are included in the 'chartFileNames' array. This could involve filtering based on file metadata or adopting a naming convention for chart files. <br> **Code Suggestion**: 
 ```
 -  $: allNamesQuery = useAllFileNames(queryClient, $runtime.instanceId);
 +  $: allNamesQuery = useAllFileNames(queryClient, $runtime.instanceId).then(files => files.filter(file => isChartFile(file)));
 ```

web-common/src/features/sources/createModel.ts
```diff
@@ -1,11 +1,11 @@
 import { removeLeadingSlash } from "@rilldata/web-common/features/entity-management/entity-mappers";
+import { fetchAllFileNames } from "@rilldata/web-common/features/entity-management/file-selectors";
```
ericpgreen2, line 2: **Performance Issue**: Introduction of 'fetchAllFileNames' without filtering for model-specific files could lead to unnecessary processing and performance overhead when creating models from sources. <br> **Fix**: Optimize the 'fetchAllFileNames' function to allow for entity-specific filtering or introduce a new method tailored for model creation scenarios to ensure only relevant files are processed, minimizing performance overhead. <br> **Code Suggestion**: 
 ```
 +import { fetchModelSpecificFileNames } from "@rilldata/web-common/features/entity-management/file-selectors";
 -import { fetchAllFileNames } from "@rilldata/web-common/features/entity-management/file-selectors";
 ```

web-common/src/features/sources/createModel.ts
```diff
@@ -19,8 +19,8 @@ export async function createModelFromSource(
   folder = removeLeadingSlash(folder);
 
   // Get new model name
-  const modelNames = await getModelNames(queryClient, instanceId);
-  const newModelName = getName(`${sourceName}_model`, modelNames);
+  const allNames = await fetchAllFileNames(queryClient, instanceId);
+  const newModelName = getName(`${sourceName}_model`, allNames);
```
ericpgreen2, line 23: **Security Issue**: Changing the method of fetching model names to 'fetchAllFileNames' without proper validation can lead to unauthorized access if the file names include sensitive information or system files. <br> **Fix**: Add validation to ensure that only authorized file names are fetched and processed. Implement filtering based on user permissions and roles. <br> **Code Suggestion**: 
 ```
 -  const allNames = await fetchAllFileNames(queryClient, instanceId);
 -  const newModelName = getName('${sourceName}_model', allNames);
 +  const allNames = await fetchAllFileNames(queryClient, instanceId).filter(name => isValidFileName(name));
 +  const newModelName = getName('${sourceName}_model', allNames.filter(name => userCanAccessFileName(name)));
 ```

web-common/src/features/sources/modal/FileDrop.svelte
```diff
@@ -36,7 +34,7 @@
 
     const uploadedFiles = uploadTableFiles(
       Array.from(files),
-      [$sourceNames?.data ?? [], $modelNames?.data ?? []],
+      $allNames.data ?? [],
       $runtime.instanceId,
     );
     for await (const { tableName, filePath } of uploadedFiles) {
```
ericpgreen2, line 40: **Security Issue**: The removal of specific file name selectors and replacement with 'useAllFileNames' increases the risk of processing unauthorized files, as it broadens the scope of accessible files without apparent restrictions or validations. <br> **Fix**: Reinstate specific file name selectors with added security checks to ensure only authorized files are processed. Implement a validation layer to filter out unauthorized access based on user roles and permissions. <br> **Code Suggestion**: 
 ```
 -      $allNames.data ?? [],
 +      $allNames.data.filter(name => userIsAuthorizedToAccessFile(name)) ?? [],
 ```

web-common/src/features/sources/modal/file-upload.ts
```diff
@@ -94,15 +94,15 @@ async function checkForDuplicate(
...
+): Promise<string | undefined> {
   const currentTableName = getTableNameFromFile(file.name);
 
   try {
     const isDuplicate = duplicateValidator(currentTableName);
     if (isDuplicate) {
       const userResponse = await getResponseFromModal(currentTableName);
       if (userResponse == DuplicateActions.Cancel) {
-        return;
+        return undefined;
```
ericpgreen2, line 105: **Security Issue**: The function 'checkForDuplicate' allows for a user response to dictate file handling without proper authorization checks, potentially leading to unauthorized file access or overwrite. <br> **Fix**: Ensure that user responses are validated against their permissions to prevent unauthorized file access or overwrite actions. <br> **Code Suggestion**: 
 ```
 -       const userResponse = await getResponseFromModal(currentTableName);
 -       if (userResponse == DuplicateActions.Cancel) {
 -         return undefined;
 -       } else if (userResponse == DuplicateActions.KeepBoth) {
 -         return incrementedNameGetter(currentTableName);
 -       } else if (userResponse == DuplicateActions.Overwrite) {
 +       if (userHasPermissionForAction(userResponse, currentTableName)) {
 +         const userResponse = await getResponseFromModal(currentTableName);
 +         if (userResponse == DuplicateActions.Cancel) {
 +           return undefined;
 +         } else if (userResponse == DuplicateActions.KeepBoth) {
 +           return incrementedNameGetter(currentTableName);
 +         } else if (userResponse == DuplicateActions.Overwrite) {
 ```

web-common/src/features/sources/modal/file-upload.ts
```diff
@@ -26,7 +23,7 @@ import {
  */
 export async function* uploadTableFiles(
   files: Array<File>,
-  [models, sources]: [Array<string>, Array<string>],
+  allNames: Array<string>,
   instanceId: string,
   goToIfSuccessful = true,
 ): AsyncGenerator<{ tableName: string; filePath: string }> {
```
ericpgreen2, line 29: **Scalability Issue**: The change from specific model and source name checks to a generic allNames check in uploadTableFiles function simplifies the code but may inadvertently increase the risk of handling large datasets inefficiently. <br> **Fix**: Reintroduce more granular checks or implement a caching strategy for name validation to ensure scalability. <br> **Code Suggestion**: 
 ```
 -  allNames: Array<string>,
 +  [models, sources]: [Array<string>, Array<string>],
 ```

web-common/src/features/custom-dashboards/AddChartMenu.svelte
ericpgreen2: I (real Eric) agree with this. This "AddChart" button should only surface valid charts. What about using `ListResources()` and filtering for charts?

web-common/src/features/sources/createModel.ts
ericpgreen2: I (real Eric again) also agree with this one. I believe we're okay to use the same name across different resource kinds. So we should only be fetching the model names here.

web-common/src/features/sources/modal/file-upload.ts
ericpgreen2: Agree, I think this should only use Source names.

ericpgreen2:

web-common/src/features/entity-management/file-artifacts.ts
```diff
@@ -205,6 +212,13 @@ export class FileArtifacts {
    * Map of all files and its individual store
    */
   private readonly artifacts: Record<string, FileArtifact> = {};
+  /**
+   * A cache of {@Link FileArtifact}s by {@link ResourceKind}
+   * This avoids going through every file artifact to filter the by kind
+   */
+  private readonly artifactsByKind: Partial<
+    Record<ResourceKind, FileArtifact[]>
+  > = {};
```
ericpgreen2, line 221: IMO it significantly increases the code complexity to have two sources-of-truth that need to be synchronized. E.g. needing to pass `handleArtifactNameChange` to the `FileArtifact` constructor adds an unfortunate coupling between the contained item and the container.

Can we remove the `artifactsByKind` cache and instead add a method like this:
```
 public getArtifactsByKind(kind: ResourceKind): FileArtifact[] {
    return Object.values(this.artifacts).filter(artifact => {
      const name = get(artifact.name);
      return name?.kind === kind;
    });
  }
```

The performance impact of dynamic filtering shouldn't be perceived by a user until the project has (many) thousands of files.

web-common/src/features/entity-management/file-artifacts.ts
```diff
@@ -337,6 +351,35 @@ export class FileArtifacts {
...
+      const index = this.artifactsByKind[prevName.kind].find(
+        (a: FileArtifact) => a.path === fileArtifact.path,
+      );
+      if (index >= 0) this.artifactsByKind[prevName.kind].splice(index, 1);
+    }
+    if (!name?.kind) return;
+    this.artifactsByKind[name.kind] ??= [];
+    this.artifactsByKind[name.kind].push(fileArtifact);
+  }
 }
```
ericpgreen2, line 383: This `handleArtifactNameChange` callback is quite confusing – see comment and suggestion above.

ericpgreen2 (changes requested):

web-common/src/features/entity-management/file-artifacts.ts
AdityaHegde: Ya makes sense.

ericpgreen2 (approved):

PR #4720, opened 2024-04-26
Fix common console error in the Model Workspace

ericpgreen2:
The `isLocalFileConnector()` function was being called for Models, triggering this YAML parse error in the console:

image

In this case, the `enabled` flag of the `isLocalFileConnector` TanStack Query call _is_ set to `false`, so it seems like a separate call to `GetFile` has been triggering the `select` function.


AdityaHegde (approved):
Good catch!

PR #4724, opened 2024-04-26
Invalidate olap table list when a new source/model is added

AdityaHegde:
List of OLAP tables is not being invalidated when a new source/model is added. Since we could get multiple events when a new entity is added we need a state in fileArtifacts for new entities.


ericpgreen2 (approved):
Works nicely, but adding this complexity to the `FileArtifact` class feels like a hack to compensate for something that should really be fixed server-side. CC @briangregoryholmes who mentioned he's looking at these events.

PR #4728, opened 2024-04-26
Fix edge cases around rename

AdityaHegde:
- [x] Just after a rename Model/Source resource would be idle with old table name. This lead to us resending the column profile leading to no table found error. Added a check to make sure table name is updated before invalidating the queries.
- [x] Every now and then a rename would send WRITE event for both source and destination, leading to errored page with old file name. This was because fsnotify would set CHMOD and RENAME to true. Flipping the condition fixes this


begelundmuller (approved):
LGTM

PR #4729, opened 2024-04-26
Support domain whitelisting for projects

### LINKED ISSUE #4496: Admin: Support domain whitelisting for projects
We currently support domain whitelisting at the org-level, but not at the project-level.

esevastyanov:
Closes #4496

### DIFF
--- a/cli/cmd/sudo/whitelist/add.go
+++ b/cli/cmd/sudo/whitelist/add.go
@@ -9,7 +9,7 @@ import (
 func AddCmd(ch *cmdutil.Helper) *cobra.Command {
 	addCmd := &cobra.Command{
 		Use:   "add <org> [<project>] <domain> <role>",
-		Args:  cobra.ExactArgs(3),
+		Args:  cobra.RangeArgs(3, 4),
 		Short: "Whitelist users from a domain",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			ctx := cmd.Context()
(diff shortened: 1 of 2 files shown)


cli/cmd/sudo/whitelist/add.go
```diff
@@ -8,7 +8,7 @@ import (
 
 func AddCmd(ch *cmdutil.Helper) *cobra.Command {
 	addCmd := &cobra.Command{
-		Use:   "add <org> <domain> <role>",
+		Use:   "add <org> [<project>] <domain> <role>",
 		Args:  cobra.ExactArgs(3),
```
begelundmuller, line 12: `ExactArgs` no longer correct here

cli/cmd/sudo/whitelist/add.go
```diff
@@ -19,11 +19,42 @@ func AddCmd(ch *cmdutil.Helper) *cobra.Command {
...
+					Organization: org,
+					Domain:       domain,
+					Role:         role,
+				})
+				if err != nil {
+					return err
+				}
+
+				ch.PrintfSuccess("Success\n")
+			}
```
begelundmuller, line 49: Needs to return here – else it will try to run the project flow after the org flow

begelundmuller (changes requested):

begelundmuller (approved):

PR #4731, opened 2024-04-26
Measure labels and formatting in alert emails

esevastyanov:


runtime/queries/proto.go
```diff
@@ -62,16 +62,18 @@ func ProtoToQuery(q *runtimev1.Query, attrs map[string]any) (runtime.Query, erro
 }
 
 // ProtoFromJSON builds a proto query from a query name, JSON args, and optional execution time.
-func ProtoFromJSON(qryName, qryArgsJSON string, executionTime *time.Time) (*runtimev1.Query, error) {
+func ProtoFromJSON(qryName, qryArgsJSON string, executionTime *time.Time) (*runtimev1.Query, string, error) {
```
begelundmuller, line 65: Can we split this out into a `func MetricsViewFromQuery(qry *runtimev1.Query) (name string, ok bool)` function instead? It would need to reproduce the `switch`, but would nicely separate the concerns (e.g. in case we need to use `ProtoFromJSON` for non-metrics queries in the future.)

runtime/reconcilers/alert.go
```diff
@@ -882,12 +887,34 @@ func calculateExecutionTimes(a *runtimev1.Alert, watermark, previousWatermark ti
...
+						}
+						break
+					}
+				}
+				if measureLabel == "" {
+					measureLabel = k
+				}
+				res[measureLabel] = formatValue(f, v, logger)
+			}
+			return res, true, nil
```
begelundmuller, line 917: 1. Can we pull this out into a util function to avoid too much nesting? E.g. `func formatResultRow(row map[string]any, metricsView) map[string]any`
2. Do you think it would make sense to log the error from `NewPresetFormatter` and output the non-formatted values? Instead of failing the alert due to an error in the format preset.

runtime/reconcilers/alert.go
```diff
@@ -896,15 +923,35 @@ func extractQueryResultFirstRow(q runtime.Query) (map[string]any, bool, error) {
 			res := make(map[string]any)
 			res[q.DimensionName] = row.DimensionValue
 			for _, v := range row.MeasureValues {
-				res[v.MeasureName] = v.BaseValue.AsInterface()
+				var measureLabel string
+				var f formatter.Formatter
+				for _, measure := range measures {
+					if v.MeasureName == measure.Name {
```
begelundmuller, line 929: Same comments as above about a util func (probably a different one since the logic is quite different) and maybe falling back to non-formatted measures instead of failing

begelundmuller (changes requested):

begelundmuller (approved):
LGTM!

PR #4732, opened 2024-04-26
Ensure consistent query keys for runtimeServiceListFiles

briangregoryholmes:
This PR removes the use of a glob parameter from calls to `runtimeServiceListFiles` to ensure that all uses of that query have the same key. Among the many issues we have with invalidation, the existence of multiple independent calls to list files meant that that they were either 1) not being invalidated when they should have been or 2) firing redundant queries for the same underlying data.

This is a somewhat speculative PR and it's possible I've missed something about the need for the glob parameter.


ericpgreen2 (approved):
Agree with all this. Nice complement to [this PR](https://github.com/rilldata/rill/pull/4715) – CC @AdityaHegde, there's just a little bit of overlap.

PR #4734, opened 2024-04-26
Revert bits-ui version update

briangregoryholmes:
This update seems to have broken, somewhat inconsistently, a single test. Reverting for now, but will revisit.


ericpgreen2 (approved):
As mentioned, I'm still seeing the test fail sporadically, but approving just in case you think this helps & want to merge.

briangregoryholmes:
See: #4739

PR #4735, opened 2024-04-26
Runtime: Dashboard components with generic renderers and optional inline definition

begelundmuller:
Changes:

- Changed the `components:` property to `items:` in the dashboard YAML
  - The idea is that `item = component + grid info`
- Renamed the `Chart` resource to `Component`
- Supports inline component definition from a dashboard:
  - if `items[*].component` is a `string`, it is treated as a reference to an externally defined component
  - if `items[*].component` is a `map`, it is treated as an inline component definition
- Added generic `renderer` and `renderer_properties` fields to `Component`
- Defines three renderers with the following formats:
  - `"renderer": "vega_lite", "rendererProperties": {"spec": "..."}`
  - `"renderer": "markdown", "rendererProperties": {"contents": "..."}`
  - `"renderer": "image", "rendererProperties": {"url": "..."}`
- Inline component definitions are emitted as separate `Component` resources called `[dashboard name]--component[idx]`
- Changes the `/charts/{name}/data` API to `/components/{name}/data`
- Changes the `GenerateChartSpec` API to `GenerateRenderer`. It now returns two values: `renderer` and `renderer_properties`.

See this [design doc](https://www.notion.so/rilldata/Dashboards-Supporting-non-chart-components-bb2e448153ff4a9a8fd22bda426ede91) for the new component syntax.

This PR currently breaks the frontend since these are breaking changes.


proto/rill/runtime/v1/resources.proto
```diff
@@ -431,12 +432,11 @@ message DashboardSpec {
...
 
 message DashboardComponent {
-  string chart = 1;
+  string component = 1;
   uint32 x = 2;
   uint32 y = 3;
   uint32 width = 4;
   uint32 height = 5;
-  string markdown = 6;
   uint32 font_size = 7;
```
begelundmuller, line 440: Should this stay in the `DashboardComponent` or be moved to the `renderer_properties` for the `markdown` renderer? If the latter, it may require a syntax change since currently it's:
```yaml
kind: component
markdown: Hello world
```
not:
```yaml
kind: component
markdown:
  contents: Hello world
  font_size: 12
```

Another idea is to not support a separately defined font size and let people do it within Markdown using for example `<span>` tags?

proto/rill/runtime/v1/resources.proto
```diff
@@ -425,18 +426,17 @@ message DashboardSpec {
...
-  repeated DashboardComponent components = 4;
+  repeated DashboardItem items = 4;
 }
 
 message DashboardState {}
 
-message DashboardComponent {
-  string chart = 1;
+message DashboardItem {
+  string component = 1;
```
begelundmuller, line 435: After fetching the dashboard, each item's component must be fetched separately using `GetResource`. Is that okay, or would you prefer if we added an API for fetching a dashboard and all of its components in a single request? It could look something like this:
```proto
message GetDashboardRequest {
  string instance_id = 1;
  string name = 2;
}

message GetDashboardResponse {
  Resource dashboard = 1;
  map<string, Resource> components = 2;
}
```

AdityaHegde: It would only be advantageous if a new resource is not emitted for each inline resource. But all inlined resources would get invalidated if one changes in that case, so i think this is preferable.

AdityaHegde (approved):

proto/rill/runtime/v1/resources.proto
begelundmuller: I was thinking that the response `components` would return *every* referenced component (both inline and non-inline). So they could still be treated the same. Does that change your assessment or not?

AdityaHegde: It would probably help in caching for non-inline components. So not sending it inline is better.

briangregoryholmes: Yeah, I don't mind this. It might simplify the read only versions of the dashboards, but it's fine as is.

begelundmuller: Got it, won't add this then :)

PR #4738, opened 2024-04-28
Runtime: load env on start and reload on change

### LINKED ISSUE #4491: Runtime: Support editing variables (and `.env`) without restart/reset
Currently, a restart is required for changes that impact metrics views, and a *reset* is required for changes that impact sources/models.

Task list:

- [ ] On start, load `.env` and update the instance's variables *before* starting the controller (avoids the race condition described in [#193](https://github.com/rilldata/rill-private-issues/issues/193) and [#4523](https://github.com/rilldata/rill/issues/4523))
- [ ] When the `.env` is changed, restart the controller
- [ ] For resources that use variables, incorporate the relevant variable values in the execution hash (to ensure correct invalidation)

k-anshul:
closes #4491


runtime/reconcilers/model.go
```diff
@@ -132,8 +132,12 @@ func (r *ModelReconciler) Reconcile(ctx context.Context, n *runtimev1.ResourceNa
 		materialize = true
 	}
 
+	sql, err := r.resolveTemplateSQL(ctx, self)
+	if err != nil {
+		return runtime.ReconcileResult{Err: err}
+	}
 	// Use a hash of execution-related fields from the spec to determine if something has changed
```
begelundmuller, line 139:
```suggestion

	// Resolve variables before computing the execution hash to ensure we re-trigger when a variable is updated
	sql, err := r.resolveTemplateSQL(ctx, self)
	if err != nil {
		return runtime.ReconcileResult{Err: err}
	}

	// Use a hash of execution-related fields from the spec to determine if something has changed
```

runtime/reconcilers/project_parser.go
```diff
@@ -293,14 +293,16 @@ func (r *ProjectParserReconciler) reconcileParser(ctx context.Context, inst *dri
...
 		diff = nil
 	}
 
 	// Update state from rill.yaml and .env
 	if diff == nil || diff.ModifiedDotEnv {
-		err := r.reconcileProjectConfig(ctx, parser)
+		err := r.reconcileProjectConfig(ctx, parser, restartController)
 		if err != nil {
 			return err
 		}
```
begelundmuller, line 308: If setting `restartController` to `true`, this reconciler's `ctx` will be cancelled right after, so I guess it should return here? And since you also restart when `diff.Reloaded`, I guess the handling for when `diff == nil` later in this file is redundant?

runtime/compilers/rillv1/parser.go
```diff
@@ -524,6 +524,12 @@ func (p *Parser) parsePaths(ctx context.Context, paths []string) error {
 		if pathIsRillYAML(b) {
 			return 1
 		}
+		if pathIsDotEnv(a) {
+			return -1
+		}
+		if pathIsDotEnv(b) {
+			return 1
+		}
```
begelundmuller, line 532: Why does `.env` need to be parsed before other resources? I don't think we read variables from inside the parser.

The reason `rill.yaml` is parsed first is because you can specify resource defaults in `rill.yaml` that are applied when parsing other files.

cli/pkg/local/app.go
```diff
@@ -182,9 +184,12 @@ func NewApp(ctx context.Context, opts *AppOptions) (*App, error) {
...
 		"rill.stage_changes":      "false",
 	}
-	for k, v := range opts.Variables {
-		vars[k] = v
+	maps.Copy(vars, opts.Variables)
+	envMap, err := godotenv.Read(filepath.Join(projectPath, ".env"))
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
+		return nil, err
 	}
+	maps.Copy(vars, envMap)
```
begelundmuller, line 192: This has some problems:
- Only works on local, not in cloud (`.env` is usually gitignored, but people may have reasons not to, e.g. if using it for non-sensitive variables)
- Duplicates `.env` parsing logic from the parser
- Does not reload on controller restart

I think this can be addressed by:
- Implementing it before starting the controller (in `runtime/registry.go` around line 419) – then it works also in cloud, and also for controller restarts. 
- Implementing a function `rillv1.ParseDotEnv` that only parses the .env file, similar to `rillv1.ParseRillYAML`

begelundmuller (changes requested):

runtime/compilers/rillv1/parser.go
k-anshul: Yeah I did these changes to test something else and realised we don't read variables in parser but missed removing these changes. Will remove.

cli/pkg/local/app.go
k-anshul: The idea behind these changes was to set variables in an instance during application boot up on local. For all other cases i.e. cloud and controller restart on local, the env variables should already be set in instance in function `reconcileProjectConfig`.
Alternatively what you are suggesting can also be done.

begelundmuller: Got it, it's a good point that in subsequent runs it will already have been updated. 

Still wondering about these two issues:
1. On the first run in cloud, there will be a flash of errors like we currently see on local. It would be nice to avoid that, especially if we start sending alerts for project errors.
2. We should try to avoid a controller restart right after startup – it causes a bunch of cancels/reruns of reconcilers, which even if safe, can still seem weird. So maybe we should also incorporate the logic for variables from `rill.yaml` into this code?

rill-dev:

runtime/reconcilers/project_parser.go
```diff
@@ -358,10 +360,7 @@ func (r *ProjectParserReconciler) reconcileProjectConfig(ctx context.Context, pa
 	}
 	inst.ProjectVariables = vars
 
-	// TODO: Passing "false" guards against infinite cancellations and restarts of the controller,
-	// but it also ignores potential consistency issues where we update connector config without evicting cached connctions,
-	// or where we update variables and don't re-evaluate all resources.
-	err = r.C.Runtime.EditInstance(ctx, inst, false)
+	err = r.C.Runtime.EditInstance(ctx, inst, restartController)
```
rill-dev, line 363: **Issue**: The introduction of the 'restartController' variable improves flexibility but lacks validation or default setting, which might lead to runtime errors if not properly handled elsewhere in the code. <br> **Fix**: Ensure that 'restartController' is validated or safely defaulted before this point in the code. If the variable is expected to be set externally, consider implementing a safeguard check or default assignment. <br> **Code Suggestion**: 
 ```
 + if restartController == nil {
 +     restartController = false // Default to false if not explicitly set
 + }
 err = r.C.Runtime.EditInstance(ctx, inst, restartController)
 ```

cli/pkg/local/app.go
```diff
@@ -182,9 +184,12 @@ func NewApp(ctx context.Context, opts *AppOptions) (*App, error) {
 		"rill.download_row_limit": "0", // 0 means unlimited
 		"rill.stage_changes":      "false",
 	}
-	for k, v := range opts.Variables {
-		vars[k] = v
+	maps.Copy(vars, opts.Variables)
+	envMap, err := godotenv.Read(filepath.Join(projectPath, ".env"))
```
rill-dev, line 188: **Issue**: Directly reading the .env file within the "NewApp" function increases coupling and reduces testability. It also handles the file path construction inline, which can lead to errors if not handled carefully. <br> **Fix**: Refactor the environment loading logic into a separate function that can be easily tested and reused. Also, ensure that the file path handling is robust against common errors. <br> **Code Suggestion**: 
 ```
 - envMap, err := godotenv.Read(filepath.Join(projectPath, ".env"))
 + envMap, err := loadEnvFile(projectPath)

 // Somewhere in the same file
 + func loadEnvFile(projectPath string) (map[string]string, error) {
 +   path := filepath.Join(projectPath, ".env")
 +   return godotenv.Read(path)
 + }
 ```

runtime/reconcilers/project_parser.go
```diff
@@ -293,14 +293,16 @@ func (r *ProjectParserReconciler) reconcileParser(ctx context.Context, inst *dri
 		return parseErrsErr
 	}
 
+	// not setting restartController=true when diff is actually nil prevents infinite restarts
+	restartController := diff != nil && (diff.ModifiedDotEnv || diff.Reloaded)
```
rill-dev, line 297: **Performance Issue**: The logic to determine 'restartController' might lead to scenarios where the controller does not restart when needed, due to not checking other conditions that might require a restart. <br> **Fix**: Refactor the condition to include checks for other configuration changes that should trigger a controller restart. <br> **Code Suggestion**: 
 ```
 + restartController := diff != nil && (diff.ModifiedDotEnv || diff.Reloaded || otherCondition)
   // Treat reloads the same as a fresh parse (where there's no diff)
   if diff != nil && diff.Reloaded {
       diff = nil
   }
 ```

runtime/reconcilers/project_parser.go
rill-dev, line 363: **Performance Issue**: The change from a hardcoded 'false' to 'restartController' in the EditInstance method call may lead to performance issues due to potential unnecessary restarts of the controller, which can be resource-intensive. <br> **Fix**: Implement a mechanism to determine when a restart is truly necessary based on the changes made to the instance. This could involve checking if the changes affect critical configurations that require a restart to take effect. <br> **Code Suggestion**: 
 ```
 - err = r.C.Runtime.EditInstance(ctx, inst, false)
 + // Check if critical configurations that require restart are changed
 + needRestart := checkNeedForRestart(inst)
 + err = r.C.Runtime.EditInstance(ctx, inst, needRestart)
 ```

cli/pkg/local/app.go
rill-dev, line 192: **Performance Issue**: The use of godotenv.Read to load environment variables from a file during runtime can lead to performance issues due to I/O operations. This is especially critical in a function that might be called frequently or in a performance-sensitive context. <br> **Fix**: To enhance performance, consider caching the environment variables at application startup and using in-memory storage for subsequent accesses. This avoids repeated I/O operations. <br> **Code Suggestion**: 
 ```
 -	envMap, err := godotenv.Read(filepath.Join(projectPath, ".env"))
 -	if err != nil and !errors.Is(err, os.ErrNotExist) {
 -		return nil, err
 -	}
 -	maps.Copy(vars, envMap)
 +	if envMap == nil {
 +		envMap, err = godotenv.Read(filepath.Join(projectPath, ".env"))
 +		if err != nil and !errors.Is(err, os.ErrNotExist) {
 +			return nil, err
 +		}
 +	}
 +	maps.Copy(vars, envMap)
 ```

runtime/reconcilers/project_parser.go
rill-dev, line 363: **Security Issue**: The code change introduces a potential security risk by dynamically controlling the restart behavior of the controller based on external input without proper validation. This could lead to unauthorized system behavior manipulation, such as denial of service (DoS) if the controller is forced to restart repeatedly. <br> **Fix**: Implement input validation for the 'restartController' variable to ensure it is derived from a trusted source and under strict conditions that prevent abuse. Additionally, consider implementing rate limiting or checks to prevent abuse of the restart feature. <br> **Code Suggestion**: 
 ```
 - err = r.C.Runtime.EditInstance(ctx, inst, restartController)
 + if validateRestartController(restartController) {
 +     err = r.C.Runtime.EditInstance(ctx, inst, restartController)
 + } else {
 +     err = fmt.Errorf("Invalid restart control parameter")
 + }
 ```

cli/pkg/local/app.go
rill-dev, line 192: **Security Issue**: The use of 'godotenv.Read' to load environment variables from a '.env' file without proper error handling and security checks can lead to sensitive information leakage or unauthorized access if the environment variables contain sensitive data. <br> **Fix**: Implement additional checks to ensure that the '.env' file does not contain sensitive information that should not be loaded into the application context. Additionally, restrict access to the '.env' file to prevent unauthorized reading or modification. <br> **Code Suggestion**: 
 ```
 + if err != nil and !errors.Is(err, os.ErrNotExist) {
 +     log.Error("Failed to read .env file", err)
 +     return nil, err
 + }
 + if !validateEnvMap(envMap) {
 +     log.Error("Invalid .env file content")
 +     return nil, fmt.Errorf("invalid .env file content")
 + }
 + maps.Copy(vars, envMap)
 ```

runtime/reconcilers/source.go
```diff
@@ -183,7 +188,7 @@ func (r *SourceReconciler) Reconcile(ctx context.Context, n *runtimev1.ResourceN
 
 	// Execute ingestion
 	r.C.Logger.Info("Ingesting source data", zap.String("name", n.Name), zap.String("connector", connector))
-	ingestErr := r.ingestSource(ctx, self, stagingTableName)
+	ingestErr := r.ingestSource(ctx, self, srcConfig, driversSink(stagingTableName))
```
rill-dev, line 191: **Security Issue**: The code change introduces a potential security vulnerability by not validating the 'srcConfig' and 'driversSink(stagingTableName)' parameters before using them in the 'ingestSource' method. This could lead to SQL injection if the input is not properly sanitized. <br> **Fix**: Ensure that 'srcConfig' and 'driversSink(stagingTableName)' are validated to contain safe values before they are used in the database query. <br> **Code Suggestion**: 
 ```
 - ingestErr := r.ingestSource(ctx, self, srcConfig, driversSink(stagingTableName))
 + // Validate srcConfig and stagingTableName before use
 + validatedSrcConfig, err := validateSrcConfig(srcConfig)
 + if err != nil {
 +     return runtime.ReconcileResult{Err: fmt.Errorf("invalid source config: %w", err)}
 + }
 + validatedStagingTableName, err := validateStagingTableName(stagingTableName)
 + if err != nil {
 +     return runtime.ReconcileResult{Err: fmt.Errorf("invalid staging table name: %w", err)}
 + }
 + ingestErr := r.ingestSource(ctx, self, validatedSrcConfig, driversSink(validatedStagingTableName))
 ```

cli/pkg/local/app.go
rill-dev, line 192: **Code Structure Issue**: The use of 'maps.Copy' for copying environment variables and opts.Variables could lead to unintended overwrites if keys overlap. The order of copying could be crucial and needs clear documentation or reconsideration to prevent potential bugs. <br> **Fix**: Ensure that the order of copying variables is intentional and document the behavior clearly. If the order is not intended, consider merging the variables with a method that handles key conflicts explicitly. <br> **Code Suggestion**: 
 ```
 -	maps.Copy(vars, opts.Variables)
 +	// Ensure no key conflicts or handle them explicitly
 +	maps.Merge(vars, opts.Variables)
 +	envMap, err := godotenv.Read(filepath.Join(projectPath, ".env"))
 +	if err != nil && !errors.Is(err, os.ErrNotExist) {
 +		return nil, err
 +	}
 +	maps.Merge(vars, envMap)  # Handle potential key conflicts during merge
 ```

rill-dev, line 192: **Scalability Issue**: The use of godotenv.Read to load environment variables from a .env file during runtime can lead to scalability issues. This approach does not support horizontal scaling well because each instance would need to manage its own set of environment variables independently, potentially leading to inconsistencies across instances. <br> **Fix**: Implement a centralized configuration management service to handle environment variables. This service can distribute configuration updates in real-time to all instances, ensuring consistency across the application's environment. <br> **Code Suggestion**: 
 ```
 -	envMap, err := godotenv.Read(filepath.Join(projectPath, ".env"))
 -	if err != nil && !errors.Is(err, os.ErrNotExist) {
 -		return nil, err
 -	}
 -	maps.Copy(vars, envMap)
 +	configService, err := config.NewCentralConfigService()
 +	if err != nil {
 +		return nil, fmt.Errorf("failed to initialize configuration service: %w", err)
 +	}
 +	err = configService.LoadEnvironmentVariables(projectPath, vars)
 +	if err != nil {
 +		return nil, fmt.Errorf("failed to load environment variables: %w", err)
 +	}
 ```

rill-dev, line 192: **Optimization Issue**: The use of godotenv.Read to load environment variables from a file during the execution of NewApp introduces potential I/O blocking and inefficiency, especially if the .env file is large or the disk I/O is slow. This can degrade the performance of the application startup. <br> **Fix**: To optimize the loading of environment variables and reduce I/O blocking, consider asynchronously loading the .env file or using a non-blocking approach. Additionally, evaluate if environment variables can be loaded once at the application startup and reused, rather than reading from the file system repeatedly. <br> **Code Suggestion**: 
 ```
 -	envMap, err := godotenv.Read(filepath.Join(projectPath, ".env"))
 -	if err != nil && !errors.Is(err, os.ErrNotExist) {
 -		return nil, err
 +	go func() {
 +		envMap, err := godotenv.Read(filepath.Join(projectPath, ".env"))
 +		if err != nil and !errors.Is(err, os.ErrNotExist) {
 +			log.Error("Failed to read .env file", err)
 +		}
 +		maps.Copy(vars, envMap)
 +	}()
 ```

runtime/registry.go
```diff
@@ -540,6 +545,48 @@ func (r *registryCache) emitHeartbeatForInstance(inst *drivers.Instance) {
...
+	for _, v := range rillYAML.Variables {
+		inst.ProjectVariables[v.Name] = v.Default
+	}
+
+	dotEnv, err := rillv1.ParseDotEnv(iwc.ctx, repo, iwc.instanceID)
+	if err != nil {
+		return err
+	}
+	maps.Copy(inst.ProjectVariables, dotEnv)
+	return r.rt.EditInstance(iwc.ctx, inst, false)
```
begelundmuller, line 587: Would it make sense to we extract this part into a util function directly on the `Runtime` object so it can be called from both the `ProjectParserReconciler` and here? Like `rt.UpdateInstanceWithRillYAML(ctx, instanceID, rillYAML, dotEnv)`.

It's probably best to centralize it one place because this is actually already missing handling of a property added a few hours back: https://github.com/rilldata/rill/blob/d0cba69b5d7ec3f08ca914666cfbae16d13ed0a4/runtime/reconcilers/project_parser.go#L358

begelundmuller (changes requested):

runtime/registry.go
```diff
@@ -417,6 +418,9 @@ func (r *registryCache) restartController(iwc *instanceWithController) {
 			}
 
 			// Start controller
+			if err := r.updateProjectConfig(iwc); err != nil {
+				iwc.logger.Warn("failed to update variables", zap.Error(err))
```
begelundmuller, line 422:
```suggestion
				iwc.logger.Warn("failed to parse and update the project config before starting the controller", zap.Error(err))
```

begelundmuller (approved):

PR #4744, opened 2024-04-29
UI changes for updated Dashboard Component specs

AdityaHegde:
Integrates the updated Dashboard component specs in the UI. Note that embedded components breaks in this PR. Should be handled separately.


AdityaHegde:
Merging since this is to another branch

PR #4753, opened 2024-04-30
Fix flicker for inline custom dashboard sub components

AdityaHegde:
While editing a custom dashboard with inline components there is a flicker. This is due to inline component mapping to the same file. In this PR we are not setting the `name` for any sub resource.


briangregoryholmes (approved):

briangregoryholmes:
Intercepting the review as this is needed for an upcoming PR of mine.

PR #4754, opened 2024-04-30
Update custom dashboard integration to new spec

briangregoryholmes:
This PR tackles the most critical updates required to bring the Custom Dashboard up to date with the new spec.

### DIFF
--- a/web-common/src/features/custom-dashboards/Component.svelte
+++ b/web-common/src/features/custom-dashboards/Component.svelte
@@ -1,12 +1,14 @@
 <script lang="ts" context="module">
   import { onMount } from "svelte";
-  import { writable } from "svelte/store";
   import type ResizeHandle from "./ResizeHandle.svelte";
   import type { ComponentType } from "svelte";
-
   import { builderActions, getAttrs, type Builder } from "bits-ui";
-
-  const zIndex = writable(0);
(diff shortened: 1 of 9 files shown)


web-common/src/features/custom-dashboards/Component.svelte
```diff
@@ -26,16 +29,25 @@
...
   export let width: number;
   export let height: number;
-  export let i: number;
   export let localZIndex = 0;
   export let chartView = false;
+  export let componentName: string;
+  export let instanceId: string;
+
+  $: resourceQuery = useResource(
+    instanceId,
```
AdityaHegde, line 38: Whats the idea of getting the instance id in the params? We have the global to avoid prop drilling.

AdityaHegde (approved):
Just a small comment. Looks good otherwise

web-common/src/features/custom-dashboards/Component.svelte
briangregoryholmes: Nothing hard and fast here. Just an instinct to get implementation details out of what are essentially display components. The runtime global is likely to change structure in the future. It also just means that only one component is subscribing to the store as opposed to however many dashboard elements there are. Not critical, though.

PR #4762, opened 2024-04-30
Fix admin repo startup deadlock because of rill.yaml read

AdityaHegde:


runtime/drivers/admin/admin.go
```diff
@@ -262,19 +262,19 @@ func (h *Handle) cloneOrPull(ctx context.Context) error {
...
 
+	// Read rill.yaml and fill in `ignore_paths`
+	rawYaml, err := h.Get(context.Background(), "/rill.yaml")
+	if err == nil {
+		yml := &rillYAML{}
+		err = yaml.Unmarshal([]byte(rawYaml), yml)
+		if err == nil {
+			h.ignorePaths = yml.IgnorePaths
+		}
+	}
```
begelundmuller, line 276: This will read `rill.yaml` and mutate `ignorePaths` without holding `repoMu`.

I think putting it back in the singleflight and using `os.ReadFile` directly (or an internal get that doesn't take a lock) would be a simpler approach

begelundmuller (changes requested):

begelundmuller (approved):

PR #4765, opened 2024-04-30
Add deployment teardown error logs

kaspersjo:


admin/deployments.go
```diff
@@ -353,38 +354,37 @@ func (s *Service) HibernateDeployments(ctx context.Context) error {
...
+		s.Logger.Error("failed to open runtime client", zap.String("deployment_id", depl.ID), zap.String("runtime_instance_id", depl.RuntimeInstanceID), zap.Error(err), observability.ZapCtx(ctx))
+	} else {
+		_, err = rt.DeleteInstance(ctx, &runtimev1.DeleteInstanceRequest{
+			InstanceId: depl.RuntimeInstanceID,
+		})
+		if err != nil {
+			s.Logger.Error("failed to delete instance", zap.String("deployment_id", depl.ID), zap.String("runtime_instance_id", depl.RuntimeInstanceID), zap.Error(err), observability.ZapCtx(ctx))
+		}
 	}
+	defer rt.Close()
```
begelundmuller, line 375: This can cause a panic if `err != nil` – it will try to call `.Close()` on `rt` even if it is `nil`.

Maybe move it to the top of the `else` clause? (`defer` is scoped by function, not by inner scopes like if/else)

begelundmuller (changes requested):

admin/deployments.go
kaspersjo: Fixed - thanks

begelundmuller (approved):

PR #4778, opened 2024-04-30
Fix: complex values as tooltip key

djbarnwal:
Stacked area tooltips were breaking for values having special characters in them (ex- dot, dash etc.)

This PR sanitizes the value before passing them to be used in the spec.

### DIFF
--- a/web-common/src/features/charts/templates/utils.ts
+++ b/web-common/src/features/charts/templates/utils.ts
@@ -22,3 +22,12 @@ export function multiLayerBaseSpec() {
 
   return baseSpec;
 }
+
+export function sanitzeValueForVega(value: string | null) {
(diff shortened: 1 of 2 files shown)


web-common/src/features/charts/templates/utils.ts
```diff
@@ -19,6 +19,14 @@ export function multiLayerBaseSpec() {
     data: { name: "table" },
     layer: [],
   };
-
   return baseSpec;
 }
+
+export function sanitzeValueForVega(value: string | null) {
```
ericpgreen2, line 25: Typo: should be `sanitizeValueForVega`

ericpgreen2 (approved):

PR #4780, opened 2024-05-01
UI: Use `type:` instead of `kind:` for resource types

ericpgreen2:
Follows https://github.com/rilldata/rill/pull/4774 with the relevant UI changes:
- New files are generated with `type` not `kind`
- The toast message that appears when moving a file out of its native folder uses "type" not "kind"
- The project status page uses "type" not "kind"
- Updates the client-side YAML parser in `file-content-utils.ts`
- Updates tests

### DIFF
--- a/web-common/src/features/entity-management/file-content-utils.spec.ts
+++ b/web-common/src/features/entity-management/file-content-utils.spec.ts
@@ -31,9 +31,21 @@ describe("parseKindAndNameFromFile", () => {
       { kind: ResourceKind.MetricsView, name: "AdBids_dashboard_name" },
     ],
     [
+      "explicit kind and name for sql",
+      "sources/AdBids_model.sql",
+      `\n\n-- @type : model\n--@name:AdBids_model_name\nselect * from AdBids`,
+      { kind: ResourceKind.Model, name: "AdBids_model_name" },
+    ],
+    [
       "explicit invalid kind for yaml",
       "sources/AdBids_dashboard.yaml",
       `\n\ntype : invalid\nmodel: AdBids_model\nmeasures: []\ndimensions: []`,
+      undefined,
+    ],
+    [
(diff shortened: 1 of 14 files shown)


web-common/src/features/entity-management/file-content-utils.spec.ts
```diff
@@ -27,25 +27,13 @@ describe("parseKindAndNameFromFile", () => {
     [
       "explicit kind and name for yaml",
       "sources/AdBids_dashboard.yaml",
-      `\n\nkind : metrics_view\nname:AdBids_dashboard_name\nmodel: AdBids_model\nmeasures: []\ndimensions: []`,
+      `\n\ntype : metrics_view\nname:AdBids_dashboard_name\nmodel: AdBids_model\nmeasures: []\ndimensions: []`,
       { kind: ResourceKind.MetricsView, name: "AdBids_dashboard_name" },
     ],
-    [
-      "explicit kind and name for sql",
```
AdityaHegde, line 34: Why are we removing tests for .sql?

AdityaHegde (approved):
Just a small question. Looks good otherwise

web-common/src/features/entity-management/file-content-utils.spec.ts
ericpgreen2: Because we're now assuming all SQL files are models and we aren't documenting `--@kind: model` anywhere, I figured there's no need to support it in these tests.

AdityaHegde: Shouldnt we remove the code as well then? Also there is a test for `@name` as well in there. IMO we should keep this since backend didnt remove the support for it.

ericpgreen2: Makes sense, just restored.

PR #4788, opened 2024-05-01
File explorer - Invalidation issues when source's table query has errored or is deleted

AdityaHegde:
- [x] When a source's table query errors (can be simulated by deleting the uploaded file and refreshed) we are not properly invalidating the list of olap tables. This was because we were not checking for chance in `table` field before invalidating. We should do a refactor here after the release to decrease complexity.
- [x] When a source (or any entity for that matter) is deleted and re-added, name collision occurs. This was because we were not handling delete event from resource watcher and there could be a race condition with file watcher leading to the artifact being re-added.


ericpgreen2 (approved):
I agree that this is getting more and more complex and that we should revisit this after the release.

PR #4789, opened 2024-05-01
File Explorer - Handle spaces in files and folders

AdityaHegde:
If there is a space in a folder or file name, application crashes since the file path will be url encoded.

We need to handle this in 2 places,
- [x] UI should make sure all references to the file name is unescaped.
- [x] Backend API that take the path in the url should unescape before consuming.
- [ ] Test a bit to make sure there are no missed case with spaces in the path.

### DIFF
--- a/runtime/server/repos.go
+++ b/runtime/server/repos.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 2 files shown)


AdityaHegde:
Closing in favour of #4790

PR #4791, opened 2024-05-01
Fix redirect on folder rename

AdityaHegde:
Seems like a regression where renaming a folder was not redirecting properly.


ericpgreen2 (approved):
Elsewhere we've found that `/files${path}` is risky because some client-side functions can accidentally return paths without leading slashes. So, I think `/files/${removeLeadingSlash(path)}` is generally safer.

AdityaHegde:
Good point. Have updated to use `removeLeadingSlash`. Lets revisit this and make it consistant

PR #4793, opened 2024-05-01
Aggregation comparison API

egor-ryashin:


runtime/queries/metricsview_aggregation.go
```diff
@@ -199,6 +224,25 @@ func (q *MetricsViewAggregation) Resolve(ctx context.Context, rt *runtime.Runtim
 	return q.pivotDruid(ctx, rows, mv, cfg.PivotCellLimit)
 }
 
+func (q *MetricsViewAggregation) executeComparisonAggregation(ctx context.Context, olap drivers.OLAPStore, priority int, mv *runtimev1.MetricsViewSpec, dialect drivers.Dialect, security *runtime.ResolvedMetricsViewSecurity) error {
+	sqlString, args, err := q.buildMetricsComparisonAggregationSQL(ctx, olap, priority, mv, dialect, security, false)
+	if err != nil {
+		return fmt.Errorf("error building query: %w", err)
+	}
+	fmt.Println(sqlString, args)
```
begelundmuller, line 232: Leftover from debugging?

proto/rill/runtime/v1/queries.proto
```diff
@@ -323,9 +323,11 @@ message MetricsViewAggregationRequest {
   // Required
   repeated MetricsViewAggregationMeasure measures = 4;
   // Optional. Defaults to unsorted
-  repeated MetricsViewAggregationSort sort = 5;
+  repeated MetricsViewComparisonSort sort = 5;
```
begelundmuller, line 326: Is there a reason you didn't add configuration for the comparison in the `MetricsViewAggregationMeasure` type as discussed here? https://www.notion.so/rilldata/Pivot-APIs-improvements-8ee3db49882e4093af85de68198f5984

What should the frontend put in `measures` to request a comparison value?

Also, if keeping this, we should rename the type to `MetricsViewAggregationSort` to align with the API.

runtime/queries/metricsview_aggregation_test.go
```diff
@@ -2023,6 +2024,277 @@ measures:
...
+	}
+	fmt.Printf("\n")
+
+	for i, row := range q.Result.Data {
+		for _, sf := range q.Result.Schema.Fields {
+			fmt.Printf("%v ", row.Fields[sf.Name].AsInterface())
+		}
+		fmt.Printf(" %d \n", i)
+
+	}
```
begelundmuller, line 2142: Avoid prints if possible

runtime/queries/metricsview_aggregation_test.go
```diff
@@ -2023,6 +2024,277 @@ measures:
...
```
begelundmuller, line 2269: Also prints

proto/rill/runtime/v1/queries.proto
AdityaHegde: +1 to this. There is no way to add having filter on comparison otherwise. So might need to build and pass `aliases` to `buildExpression` for having clause.

egor-ryashin: yes, it's the next step to adapt it to the new proposal

begelundmuller:
@egor-ryashin Can you fix the merge conflicts here?

begelundmuller (approved):

PR #4800, opened 2024-05-02
Use GetInstance to load feature flags instead of rill.yaml

### LINKED ISSUE #4763: Fix: Do not request rill.yaml in rill cloud
For feature flags we are fetching `rill.yaml` to get the defaults. We cannot fetch this in cloud since read file access will only be given to admins.

AdityaHegde:
closes #4763

We cannot use repo APIs in cloud to get rill.yaml. Feature flags was moved to GetInstance. This PR updates to make use of GetInstance.

### DIFF
--- a/web-common/src/features/feature-flags.ts
+++ b/web-common/src/features/feature-flags.ts
@@ -1,15 +1,12 @@
 import type { BeforeNavigate } from "@sveltejs/kit";
-import type { Readable } from "svelte/store";
 import { writable } from "svelte/store";
-import { parse } from "yaml";
-import { createRuntimeServiceGetFile } from "../runtime-client";
+import {
+  createRuntimeServiceGetInstance,
+  V1InstanceFeatureFlags,
+} from "../runtime-client";
 import { runtime } from "../runtime-client/runtime-store";
 import { debounce } from "../lib/create-debouncer";
 import { queryClient } from "@rilldata/web-common/lib/svelte-query/globalQueryClient";
-
-export type DerivedReadables<T> = {
-  [K in keyof T]: Readable<T[K]>;
-};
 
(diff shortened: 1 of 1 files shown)


briangregoryholmes:
If this PR is not a release blocker, I think it would be a good time to solidify our precedence going forward.

I'm of the opinion that we should remove setting feature flags via localStorage or the URL. If we need to maintain that for some reason, then we need to change the feature flag implementation so that certain flags cannot be "overridden". For instance, you might set `exports` to false in a `rill.yaml` and a user should not be able to force this to true by appending something to the URL. But that may not be true of every feature flag. Currently, there's this idea of "system" vs "user", but I think we should change to this `scope` idea to `rill` vs `user` and then add an additional property `forceable` that defaults to false.

@ericpgreen2

AdityaHegde:
Good point. Ideally we should have a check in the backend as well. Because it is trivial to force change it on client side and still enable export for example.

begelundmuller:
> @begelundmuller Since the issue was with an incomplete release, do you think this is a blocker still?

I don't think it should be a release blocker

ericpgreen2:
> If this PR is not a release blocker, I think it would be a good time to solidify our precedence for different feature flag methods going forward.
> 
> I'm of the opinion that we should remove setting feature flags via localStorage or the URL. 
> 
> @ericpgreen2

I'm game for this. Both setting flags via local storage and via URL preceded our ability to set flags in `rill.yaml`. It'll be simpler to have just one way of operating. Both ourselves and customers can handle `rill.yaml`.

ericpgreen2 (approved):
Approving, since removing URL & local storage feature flag support can always be done in a separate PR.

PR #4801, opened 2024-05-02
Change feature flag yaml structure to map

briangregoryholmes:
We've had a few requests to allow for disabling certain features and the current feature flag syntax for `rill.yaml` would make this cumbersome. This PR proposes changing the structure to a `string/bool` map (which is how it's treated after parsing now anyway) so that customers can easily set certain features to `false`

#4800 would need to be merged first

### DIFF
--- a/runtime/compilers/rillv1/parse_rillyaml.go
+++ b/runtime/compilers/rillv1/parse_rillyaml.go
@@ -63,8 +63,8 @@ type rillYAML struct {
 	Dashboards yaml.Node `yaml:"dashboards"`
 	// Default YAML values for migrations
 	Migrations yaml.Node `yaml:"migrations"`
-	// Feature flags
-	Features map[string]bool `yaml:"features"`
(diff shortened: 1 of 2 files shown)


runtime/compilers/rillv1/parse_rillyaml.go
```diff
@@ -64,7 +64,7 @@ type rillYAML struct {
 	// Default YAML values for migrations
 	Migrations yaml.Node `yaml:"migrations"`
 	// Feature flags
-	Features []string `yaml:"features"`
+	Features map[string]bool `yaml:"features"`
```
begelundmuller, line 67: Do we have users using a feature flag today? Because this is a breaking change, and a parse error in `rill.yaml` can break the entire project.

For backwards compatibility, you could use `yaml.Node` and try parsing it as both a map and an array :)

begelundmuller: Agreed with Brian that we'll make this change to maintain backwards compatibility

begelundmuller (approved):

PR #4804, opened 2024-05-02
Disable stacked charts for no comparison

djbarnwal:
Disabled stacked area and stacked bar chart when there is no active comparison.


ericpgreen2:
Nice, I like the functionality. 

UXQA:
- The color contrast makes it hard to tell that the button is disabled (CC @jkhwu).  
- A tooltip would be nice to explain what's going on, e.g. "Add a comparison to enable the stacked bar chart."

image

ericpgreen2 (approved):

jkhwu:
Try this for the disabled state:
Screenshot 2024-05-02 at 12 48 07 PM
https://www.figma.com/file/Qt6EyotCBS3V6O31jVhMQ7/RILL?type=design&node-id=12107-811732&mode=design&t=DfzCKfa8fhFEwo90-11

PR #4807, opened 2024-05-02
Increase provisioner await retry time

kaspersjo:


begelundmuller (approved):
Is this correct – in the worst case, it will still only retry for 6 seconds? Maybe we should go higher?

kaspersjo:
> Is this correct – in the worst case, it will still only retry for 6 seconds? Maybe we should go higher?

No - 6 seconds is the maximum wait time between retries, so first it waits 2 s, then ~4 s and then finally 6 s.

PR #4816, opened 2024-05-03
Runtime: Fix multiple controller restart due to duplicate watcher events for rill.yaml and .env

k-anshul:
The watcher sends multiple events for a single edit in a file. This leads to multiple restarts of the controller due to edit in `rill.yaml` or `.env` file. 
We now take a diff of the latest file with the one present in parser state and only trigger restart if the contents are actually changed.


runtime/registry.go
```diff
@@ -556,6 +557,9 @@ func (r *registryCache) updateProjectConfig(iwc *instanceWithController) error {
 
 	rillYAML, err := rillv1.ParseRillYAML(iwc.ctx, repo, iwc.instanceID)
 	if err != nil {
+		if strings.Contains(err.Error(), "rill.yaml not found") { // empty project
+			return nil
+		}
```
begelundmuller, line 562: Since this is in the same language/repo, it's better to declare `rillv1.ErrRillYAMLNotFound` and use `errors.Is` to check here

runtime/compilers/rillv1/parser.go
```diff
@@ -389,8 +398,17 @@ func (p *Parser) reparseExceptRillYAML(ctx context.Context, paths []string) (*Di
...
+			if err == nil {
+				modifiedDotEnv = !maps.Equal(p.DotEnv, oldDotEnv)
+			} else {
+				// any error means .env is under change
+				modifiedDotEnv = true
+			}
+			if modifiedDotEnv {
+				p.DotEnv = nil
+			}
 		}
```
begelundmuller, line 412: 1. I think it shouldn't be added to `parsePaths` if it is not modified? Could move it up to around line 377 and call `continue` if it's not modified.
2. Add a comment about why we're checking this (otherwise in some future refactor, someone may not know the context and revert it)

runtime/compilers/rillv1/parser.go
```diff
@@ -242,11 +243,19 @@ func Parse(ctx context.Context, repo drivers.RepoStore, instanceID, environment,
...
+		oldRillYAML := p.RillYAML
+		err := p.parseRillYAML(ctx, path)
+		if err == nil {
+			changedRillYAML = !equalRillYAML(oldRillYAML, p.RillYAML)
+		} else {
+			// any error including parse error means rill.yaml changed
 			changedRillYAML = true
-			break
 		}
+		break
```
begelundmuller, line 258: Add a comment about why this is necessary

runtime/compilers/rillv1/parse_rillyaml.go
```diff
@@ -178,3 +181,76 @@ func (p *Parser) parseRillYAML(ctx context.Context, path string) error {
 	p.RillYAML = res
 	return nil
 }
+
+func equalRillYAML(r, other *RillYAML) bool {
```
begelundmuller, line 185: 1. Consider adding as a member function: `(r *RillYAML) equals(other *RillYAML) bool`
2. Did you check if using `reflect.DeepEquals` would do the job? We don't need a very good comparison function here, just avoiding reparses for the CHMOD (where the YAML is exactly the same) is fine. We could also consider just storing an MD5 hash of the YAML contents in the struct and comparing that.

begelundmuller (commented):
One other comment: I noticed when reading this that in `runtime/reconcilers/project_parser.go` on line 295, if `diff == nil`, then it doesn't update the project config at all. I guess it should still update the project config, but without triggering a restart and returning?

runtime/compilers/rillv1/parse_rillyaml.go
k-anshul: No I did not evaluate `reflect.DeepEquals` earlier. 
Checked and seems to work fine.

k-anshul:
> One other comment: I noticed when reading this that in `runtime/reconcilers/project_parser.go` on line 295, if `diff == nil`, then it doesn't update the project config at all. I guess it should still update the project config, but without triggering a restart and returning?

Yes it was a miss. Didn't see any issues in testing because `ProjectVariables` are already updated in `registry.go` before starting controller but will be an issue for other configs like `FeatureFlags`. Fixed it.

runtime/reconcilers/project_parser.go
```diff
@@ -291,6 +291,13 @@ func (r *ProjectParserReconciler) reconcileParser(ctx context.Context, inst *dri
...
+	if diff == nil {
+		err := r.reconcileProjectConfig(ctx, parser, false)
+		if err != nil {
+			return err
+		}
+	}
+
 	restartController := diff != nil && (diff.ModifiedDotEnv || diff.Reloaded)
 	if restartController {
 		return r.reconcileProjectConfig(ctx, parser, true)
```
begelundmuller, line 303: If you don't realize the dependency between `diff == nil` and `restartController`, it looks like `reconcileProjectConfig` can be called twice in a row here. This might help make things clearer:

```suggestion
	updateConfig := diff == nil || diff.ModifiedDotEnv || diff.Reloaded
	if updateConfig {
		restartController := diff != nil
		err := r.reconcileProjectConfig(ctx, parser, restartController)
		if err != nil {
			return err
		}
		if restartController {
			return nil
		}		
```

begelundmuller (approved):

runtime/reconcilers/project_parser.go
k-anshul: Yeah looks more cleaner.

PR #4818, opened 2024-05-03
feat: Comparison based alerts

### LINKED ISSUE #4544: Feat: Support comparison-based alerts in the UI
https://www.notion.so/rilldata/Context-metrics-in-Alerts-089a5ced9126434e831d72201d0e1857?pvs=4

AdityaHegde:
closes #4544

Uses #4793 to add comparison based alerts on the MetricsViewAggregation API

- [x] Add support for comparison in create/edit alert forms.
- [x] Extract comparison during edit alert.
- [x] Add comparison to readonly pills
- [x] Update formatting for MetricsAggregation for comparison.
- [x] Fix MetricsAggregation in alerts when comparison is enabled.
- [x] Refactor based on new mocks.


begelundmuller (commented):
Backend changes look good :)

ericpgreen2:
UXQA (some of which is out-of-scope of this PR, but want to document) –

- The "Comparing with" pill looks like it's missing something. In this case, shouldn't it say it's "Comparing with the Previous Period"?
image
- On the first page, using the keyboard's "escape" key triggers the "Close without saving" dialog (while the "X" and "Cancel" buttons correctly do not trigger it)
image
- The table header's text (here "Number of contributors") spills across the header's bottom border
image
- I'm seeing "% change from _undefined_" in the measure extension dropdown
image
- When a criteria does not have a measure selected (see Criteria 2), the Next button is currently unresponsive. It should either be disabled or clicking on it should show an error.
image
- The email recipients should not be initialized with an extra blank email address.
image
- This criteria pill should say "[Measure label] % change > -10%", not "[Measure name] % change _from_ > -10%"
image

ericpgreen2:
Code – So far I've just skimmed it, but I'm seeing a few leftover console logs FYI.

AdityaHegde:
Thanks for the review. Have fixed 1,4 and 7. The duration and label mapping was incorrect (delta was incorrectly 0).
Rest are out of scope.
2. I have disabled escape for now. There is a challenge here where escape is handled by a wrapper `GuardedDialog`, but `isTouched` is deeper inside.
3. Couldnt reproduce this. It is increasing the height of the header for me. Either way this is using the table components. I can see the issue in dimension table now and then as well.
5. Forms lib is not populating the error properly. It just has `Required` for `criteria` and no indication of which index the error is for. I have set the measure to a valid value for now.
6. Hmm this has been there since we added it for the 1st time. Are we sure we dont want an empty entry there? (It gets removed before submitting)

ericokuma:
**UX/Product Check on @ericpgreen2's first pass**
1. Comparing with pill is now displaying previous time period
Screenshot 2024-05-30 at 3 35 31 PM

2. Keyboard "Escape" key doesn't trigger "close without saving" dialog anymore

3. Looks like the column spillover is still there
Screenshot 2024-05-30 at 3 42 10 PM

4. % change from _undefined_ is now fixed

5. Adding a new criteria now preloads with a measure (instead of displaying the **measure** label). Works but not sure if this is intended behavior? A question I have is: would there ever be more than one measure available to select from the dropdown?

6.Email recipients still initialized with extra blank email input field


7.  Criteria readout looks good now
Screenshot 2024-05-30 at 3 57 39 PM

**New UXQA**
1. For long measure names, looks like there's spillover in the criteria dropdown box
Screenshot 2024-05-30 at 3 45 26 PM

2. Selecting first measure from Alert data dropdown will autopopulate measure selection in Criteria step. Selecting other measures will not autopopulate measure selection 
Screenshot 2024-05-30 at 3 51 13 PM

Screenshot 2024-05-30 at 3 51 25 PM

3. [URGENT] You have to click **next** button twice on Criteria step to get to Delivery step

4. Shouldn't say "% change from from previous day". Remove one of the "from"
Screenshot 2024-05-30 at 3 56 45 PM
Screenshot 2024-05-30 at 3 56 55 PM

ericpgreen2 (approved):
Approved 👍. For those UXQA items that are not in scope of this PR, @ericokuma is adding to the [Alerts UXQA Notion doc](https://www.notion.so/rilldata/UXQA-Alerts-fd2c84406c5247e4868ccef18b13f5ff).

AdityaHegde:
Fixed the new items. Found the issue for incorrect column height. Fixed it when the column is not in the 1st place. For the case where it is in 1st place, the code is used in dimension table as well, so will do a follow up after confirming why we have that in dimension table.

PR #4823, opened 2024-05-04
Fix bug in conditional for OLAP instructions

ericpgreen2:
Closes https://github.com/rilldata/rill-private-issues/issues/343


medriscoll (approved):
lgtm

PR #4827, opened 2024-05-05
File Explorer: Add e2e tests

ericpgreen2:
- Improves the HTML semantics and accessibility of folders in the file explorer
- Adds two end-to-end tests for common actions in the file explorer


runtime/drivers/file/watcher.go
```diff
@@ -181,15 +181,15 @@ func (w *watcher) runInner() error {
 				continue
 			}
 
-			if e.Has(fsnotify.Create) {
+			if we.Type == runtimev1.FileEvent_FILE_EVENT_WRITE {
```
begelundmuller, line 184: The `write` and `chmod` events are much more frequent, so calling `os.Stat` every time might be expensive if there are many files or streaming writes (e.g. a file being exported). But maybe it's not a problem.

(Note that this is before the batching in `flush`, so the same file may get checked very frequently here.)

runtime/drivers/file/watcher.go
```diff
@@ -181,15 +181,15 @@ func (w *watcher) runInner() error {
...
 				info, err := os.Stat(e.Name)
 				we.Dir = err == nil && info.IsDir()
 			}
 
 			w.buffer[path] = we
 
 			// Calling addDir after appending to w.buffer, to sequence events correctly
-			if we.Dir && e.Has(fsnotify.Create) {
+			if we.Dir && we.Type == runtimev1.FileEvent_FILE_EVENT_WRITE {
 				err = w.addDir(e.Name, true, false)
```
begelundmuller, line 193: I'm not sure if it's safe to call `addDir` multiple times for the same directory, you'll need to check the behavior in detail. At best, it will read all nested directories again (which might be expensive). At worst, it will double-track the directory, so the same events are emitted multiple times (which could snowball).

runtime/drivers/file/watcher.go
AdityaHegde: Ya I think we can merge events with other events for the path within a batch. That should fix the issue with renamed directories not getting `Dir=true`.

runtime/drivers/file/watcher.go
AdityaHegde: Makes sense. Lemme try if merging events within a batch is enough here.

AdityaHegde (approved):
Approving the UI changes

runtime/drivers/file/watcher.go
```diff
@@ -175,27 +194,20 @@ func (w *watcher) runInner() error {
...
 			}
 
-			if e.Has(fsnotify.Create) {
-				info, err := os.Stat(e.Name)
-				we.Dir = err == nil && info.IsDir()
+			existing, ok := w.buffer[path]
+			if ok && existing.IsCreate {
+				// copy over `IsCreate` within the batch for a path
+				we.IsCreate = existing.IsCreate
 			}
```
begelundmuller, line 208: What if the file is rapidly created and deleted? Putting `IsCreate` on a delete event seems odd.

runtime/drivers/repo.go
```diff
@@ -32,8 +32,12 @@ type WatchCallback func(event []WatchEvent)
...
 type WatchEvent struct {
 	Type runtimev1.FileEvent
+	// Path relative to the repo root
 	Path string
-	Dir  bool
+	// FullPath used during flush
+	FullPath string
+	Dir      bool
+	IsCreate bool
 }
```
begelundmuller, line 41: Feels like this is leaking implementation details in the interface. Consider using an internal wrapper instead.

runtime/drivers/file/watcher.go
```diff
@@ -119,6 +119,24 @@ func (w *watcher) flush() {
 		return
 	}
 
+	for p, event := range w.buffer {
+		if !event.IsCreate {
+			continue
+		}
```
begelundmuller, line 125: If it skips the non-create case, what is it exactly this solves? I thought the issue was some directories not having create events when moved?

AdityaHegde: The order of events is unreliable. Merging them was what fixed it earlier. This change moves it to flush.

runtime/drivers/file/watcher.go
AdityaHegde: The issue was older events overriding `Dir`. But calling `addDir` multiple times would be suboptimal. That is why we are doing it this way.

AdityaHegde: Ya ignoring DELETE events. Newer delete events should overwrite older events.

Issue #4688, opened 2024-04-23
Refactor(file explorer): Use one common Editor for all files/resources


ericpgreen2:
One canonical editor will:
- Make it easier to add proper workspaces for yet-to-be-handled resources, like APIs, Themes, and Metrics
- Improve performance when switching between files of different resource kinds
- Reduce code duplication and maintenance costs 

The editor should take a list of CodeMirror extensions as a prop.

Related to https://github.com/rilldata/rill/issues/4675

PR #4830, opened 2024-05-06
Add autosave toggle to fallback `Editor` component

ericpgreen2:
This PR takes a few steps forward in refactoring the `Editor` components (https://github.com/rilldata/rill/issues/4688). This PR brings the fallback `Editor` more into parity with the other editors (`YAMLEditor`, `ModelEditor`), so that we can ultimately use this `Editor` as the one canonical editor, regardless of file/resource type.

Specifically, this PR:
- Adds the autosave toggle to the fallback `Editor`
- Hides the "autosave" toggle for `rill.yaml` and `.env` files (given edits to these files can trigger source re-ingestion)
- Moves the debounce logic into the `Editor` component, so the "autosave" is debounced but explicit save actions (the "Save" button; `CMD` + `s`) are not.
- Adds the gray wrapper/padding to the fallback `Editor`
- Aligns terminology across the different editors


web-common/src/layout/workspace/workspace-stores.ts
```diff
@@ -40,7 +42,9 @@ class WorkspaceLayoutStore {
         parsed?.table?.height ?? DEFAULT_PREVIEW_TABLE_HEIGHT,
       );
       this.tableVisible.set(parsed?.table?.visible ?? true);
-      this.autoSave.set(parsed?.editor?.autoSave ?? true);
+      this.autoSave.set(
+        parsed?.editor?.autoSave ?? this.getAutoSaveDefault(key),
+      );
```
briangregoryholmes, line 47: Nit: Looks like you can override this. I can change `false` to `true` in localStorage and it will still auto-save. The thing that's really controlling autosave is the Editor component (for better or for worse), so I think it'd be fine to either import FILES_WITHOUT_AUTOSAVE there or use the hideAutoSave prop you added. Could leave this file as is, then.

web-common/src/features/editor/Editor.svelte
```diff
@@ -34,6 +46,30 @@
...
+          // any extensions passed as props
+          ...newExtensions,
+          EditorView.updateListener.of((v) => {
+            if (v.focusChanged && v.view.hasFocus) {
+              dispatch("receive-focus");
+            }
+            if (v.docChanged) {
+              latest = v.state.doc.toString();
+
+              if (autoSave) debounceSave();
```
briangregoryholmes, line 65: Add !hideAutoSave or !FILES_WITHOUT_AUTOSAVE conditional check here, perhaps?

briangregoryholmes (approved):

web-common/src/features/editor/Editor.svelte
ericpgreen2: Yeah, that's simpler, good call

PR #4832, opened 2024-05-06
CLI: Uninstall command

begelundmuller:
This PR adds a `rill uninstall` command.

It also makes two other minor changes to improve the CLI's use of the install script:
- Adds support for the `--version` flag on `rill upgrade`
- Embeds the install script directly in the binary

### DIFF
--- a/cli/pkg/installscript/installscript.go
+++ b/cli/pkg/installscript/installscript.go
@@ -44,6 +44,7 @@ func createScriptFile() (string, error) {
 	if err != nil {
 		return "", fmt.Errorf("install script not embedded (is this a dev build?): %w", err)
 	}
+	defer in.Close()
 
 	// Write the install script to a temporary file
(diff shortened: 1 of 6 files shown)


nishantmonu51 (approved):

cli/pkg/installscript/installscript.go
```diff
@@ -0,0 +1,63 @@
...
+func execScript(ctx context.Context, args ...string) error {
+	script, err := createScriptFile()
+	if err != nil {
+		return err
+	}
+	defer os.Remove(script)
+
+	// Execute the script with bash
+	args = append([]string{script}, args...)
+	cmd := exec.CommandContext(ctx, "/bin/bash", args...)
```
kaspersjo, line 35: I suggest we change this to `/bin/sh`

kaspersjo (changes requested):

PR #4843, opened 2024-05-06
runtime-local auth support

pjain1:
- Support login using PKCE auth flow https://www.oauth.com/oauth2-servers/pkce/ without need for any secrets at local runtime
- After login access token is persisted locally in `~/.rill/credentials.yaml`
- Next up will be to use this token to deploy project using local UI

### DIFF
--- a/admin/database/database.go
+++ b/admin/database/database.go
@@ -515,6 +515,7 @@ const (
 	AuthClientIDRillWeb     = "12345678-0000-0000-0000-000000000001"
 	AuthClientIDRillCLI     = "12345678-0000-0000-0000-000000000002"
 	AuthClientIDRillSupport = "12345678-0000-0000-0000-000000000003"
+	AuthClientIDRillLocal   = "12345678-0000-0000-0000-000000000004"
 )
 
 // DeviceAuthCodeState is an enum representing the approval state of a DeviceAuthCode
--- a/admin/database/postgres/migrations/0028.sql
+++ b/admin/database/postgres/migrations/0028.sql
@@ -1,0 +1,3 @@
+-- Hard-coded first-party auth clients
+INSERT INTO auth_clients (id, display_name)
+VALUES ('12345678-0000-0000-0000-000000000004', 'Rill Local');
--- a/admin/server/auth/handlers.go
+++ b/admin/server/auth/handlers.go
@@ -5,6 +5,7 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
+	"io"
 	"net/http"
 	"net/url"
 	"strconv"
@@ -56,7 +57,8 @@ func (a *Authenticator) RegisterEndpoints(mux *http.ServeMux, limiter ratelimit.
 	observability.MuxHandle(inner, "/auth/logout", middleware.Check(checkLimit("/auth/logout"), http.HandlerFunc(a.authLogout)))
 	observability.MuxHandle(inner, "/auth/logout/callback", middleware.Check(checkLimit("/auth/logout/callback"), http.HandlerFunc(a.authLogoutCallback)))
 	observability.MuxHandle(inner, "/auth/oauth/device_authorization", middleware.Check(checkLimit("/auth/oauth/device_authorization"), http.HandlerFunc(a.handleDeviceCodeRequest)))
-	observability.MuxHandle(inner, "/auth/oauth/device", a.HTTPMiddleware(middleware.Check(checkLimit("/auth/oauth/device"), http.HandlerFunc(a.handleUserCodeConfirmation)))) // NOTE: Uses auth middleware
+	observability.MuxHandle(inner, "/auth/oauth/device", a.HTTPMiddleware(middleware.Check(checkLimit("/auth/oauth/device"), http.HandlerFunc(a.handleUserCodeConfirmation))))   // NOTE: Uses auth middleware
+	observability.MuxHandle(inner, "/auth/oauth/authorize", a.HTTPMiddleware(middleware.Check(checkLimit("/auth/oauth/authorize"), http.HandlerFunc(a.handleAuthorizeRequest)))) // NOTE: Uses auth middleware
 	observability.MuxHandle(inner, "/auth/oauth/token", middleware.Check(checkLimit("/auth/oauth/token"), http.HandlerFunc(a.getAccessToken)))
 	mux.Handle("/auth/", observability.Middleware("admin", a.logger, inner))
 }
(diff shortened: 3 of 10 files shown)


admin/database/database.go
```diff
@@ -515,6 +515,7 @@ const (
 	AuthClientIDRillWeb     = "12345678-0000-0000-0000-000000000001"
 	AuthClientIDRillCLI     = "12345678-0000-0000-0000-000000000002"
 	AuthClientIDRillSupport = "12345678-0000-0000-0000-000000000003"
+	AuthClientIDRillLocal   = "12345678-0000-0000-0000-000000000004"
```
begelundmuller, line 518: nit: maybe `WebLocal` to differentiate from `CLI` and `Web` (admin)?

admin/server/auth/handlers.go
```diff
@@ -355,3 +357,84 @@ func (a *Authenticator) authLogoutCallback(w http.ResponseWriter, r *http.Reques
...
+// and generates an authorization code while associating the code challenge.
+func (a *Authenticator) handleAuthorizeRequest(w http.ResponseWriter, r *http.Request) {
+	claims := GetClaims(r.Context())
+	if claims == nil {
+		internalServerError(w, fmt.Errorf("did not find any claims, %w", errors.New("server error")))
+		return
+	}
+	if claims.OwnerType() == OwnerTypeAnon {
+		// not logged in, redirect to login
+		// TODO how to choose between login and signup?
```
begelundmuller, line 371: Probably we need a way for the caller to indicate a preference (based on whether a login or signup button is pressed). Maybe we could support the `screen_hint=signup` URL arg like Auth0 does? See here: https://github.com/rilldata/rill/blob/b1e5563681aa8bff0894a1eb978527a8d320e623/admin/server/auth/handlers.go#L109

admin/server/auth/handlers.go
```diff
@@ -355,3 +357,84 @@ func (a *Authenticator) authLogoutCallback(w http.ResponseWriter, r *http.Reques
...
+		}
+		handlePKCE(w, r, clientID, userID, codeChallenge, codeChallengeMethod, redirectURI)
+	} else {
+		http.Error(w, "only PKCE based authorization code flow is supported", http.StatusBadRequest)
+		return
+	}
+}
+
+// getAccessToken verifies the device code and returns an access token if the request is approved
+func (a *Authenticator) getAccessToken(w http.ResponseWriter, r *http.Request) {
```
begelundmuller, line 412: nit: comment needs to be updated to indicate that it supports both device code and PKCE

admin/server/auth/pkce.go
```diff
@@ -0,0 +1,178 @@
...
+	"time"
+
+	"github.com/rilldata/rill/cli/pkg/auth"
+)
+
+const authorizationCodeGrantType = "authorization_code"
+
+// in-memory storage for authorization codes
+// TODO use persistent storage to handle admin server restarts, but since codes expire in a minute, is it really necessary, user can just re-initiate login ?
+var authCodeDB = make(map[string]AuthorizationCode)
```
begelundmuller, line 21: It's necessary because we have multiple replicas of the admin service – so the next request may go to another machine

admin/server/auth/pkce.go
```diff
@@ -0,0 +1,178 @@
...
+		ClientID:            clientID,
+		RedirectURI:         redirectURI,
+		UserID:              userID,
+		Expiration:          expiration,
+		CodeChallenge:       codeChallenge,
+		CodeChallengeMethod: codeChallengeMethod,
+	}
+
+	// Build the redirection URI with the authorization code as per OAuth2 spec
+	redirectWithCode := fmt.Sprintf("%s?code=%s&state=%s", redirectURI, code, r.URL.Query().Get("state"))
```
begelundmuller, line 56: Could there ever be non-URL safe characters in the state? If yes, then consider using the `url` package to parse, add to and reconstruct the url

admin/server/auth/pkce.go
```diff
@@ -0,0 +1,178 @@
...
+		http.Error(w, "invalid code verifier", http.StatusBadRequest)
+		return
+	}
+
+	// Issue an access token
+	authToken, err := a.admin.IssueUserAuthToken(r.Context(), userID, authCode.ClientID, "", nil, nil)
+	if err != nil {
+		internalServerError(w, fmt.Errorf("failed to issue access token, %w", err))
+		return
+	}
```
begelundmuller, line 137: If `errors.Is(err, r.Context().Err())`, then we should not return an internal error

admin/server/auth/pkce.go
```diff
@@ -0,0 +1,178 @@
...
+
+	resp := auth.OAuthTokenResponse{
+		AccessToken: authToken.Token().String(),
+		TokenType:   "Bearer",
+		ExpiresIn:   time.UnixMilli(0).Unix(), // never expires
```
begelundmuller, line 142: Could this just be `0` or is there a reason for the unix two-way conversion?

admin/server/auth/pkce.go
```diff
@@ -0,0 +1,178 @@
...
+	"crypto/sha256"
+	"encoding/base64"
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/url"
+	"time"
+
+	"github.com/rilldata/rill/cli/pkg/auth"
```
begelundmuller, line 14: nit: we currently only have imports from cli -> admin, not from admin -> cli. Can we move this to `admin/pkg` instead?

cli/pkg/auth/utils.go
```diff
@@ -0,0 +1,14 @@
+package auth
+
+const (
+	FormMediaType = "application/x-www-form-urlencoded"
+	JSONMediaType = "application/json"
+)
+
+// OAuthTokenResponse contains the information returned after fetching an access token from the OAuth server.
+type OAuthTokenResponse struct {
```
begelundmuller, line 9: nit: Consider changing the package name to `oauth`. Then you could remove the `OAuth` prefix and the type name could just be `oauth.TokenResponse`.

cli/pkg/local/app.go
```diff
@@ -578,6 +588,79 @@ func (a *App) emitStartEvent(ctx context.Context) error {
...
+			return
+		}
+
+		// Exchange the code for an access token
+		token, err := authenticator.ExchangeCodeForToken(code)
+		if err != nil {
+			http.Error(w, "failed to exchange code for token", http.StatusInternalServerError)
+			return
+		}
+		// save token and redirect to home screen or TODO url provided in initial request i.e. initiateAuthFlow
```
begelundmuller, line 654: Yes it would be nice if we could support a `redirect` parameter in `initiateAuthFlow` like we do in the admin auth endpoints.

cli/pkg/local/app.go
```diff
@@ -578,6 +588,79 @@ func (a *App) emitStartEvent(ctx context.Context) error {
...
+		err = dotrill.SetAccessToken(token)
+		if err != nil {
+			http.Error(w, "failed to save access token", http.StatusInternalServerError)
+			return
+		}
```
begelundmuller, line 659: In addition to updating `dotrill`, we may also need to set `ch.AdminTokenDefault = token` (this is used in the current process for subsequent requests – `.rill` is only read at start time). But not sure if `ch` is in scope here (from the CLI), we may need to pass it in.

cli/pkg/pkce/authenticator.go
```diff
@@ -0,0 +1,176 @@
...
+
+	// Check if the response is an error
+	if resp.StatusCode != http.StatusOK {
+		return "", fmt.Errorf("unexpected status code: %d", resp.StatusCode)
+	}
+
+	// TODO move OAuthTokenResponse to a common place
+	tokenResponse := &auth.OAuthTokenResponse{}
+	// Decode the response into the tokenResponse struct
+	if err := json.NewDecoder(resp.Body).Decode(&tokenResponse); err != nil {
```
begelundmuller, line 103: Seems like this `TODO` has already been addressed?

begelundmuller (commented):
Overall looks very good!

cli/pkg/local/app.go
begelundmuller: Can we call the parameter `redirect=https://...` instead of `origin`? Just to align naming with the admin's `/auth/login` endpoint: https://github.com/rilldata/rill/blob/e14e6feb9a1d196a13ee526e33239f72ef0630e8/admin/server/auth/handlers.go#L94

admin/server/auth/pkce.go
begelundmuller: Okay, can we also add a small comment here to note that both the state and code are URL safe then? Just to save the next reader the time to trace how they are generated

admin/server/auth/handlers.go
begelundmuller: Not sure what the UI will do, just thought we might maintain optionality and let the UI decide. But probably you're right it will go into a single flow (and our Auth0 login screen allows switching to signup), so let's just not do this, and see if anyone requests it.

admin/database/database.go
```diff
@@ -540,6 +546,19 @@ type DeviceAuthCode struct {
 	UpdatedOn     time.Time           `db:"updated_on"`
 }
 
+type AuthorizationCode struct {
```
begelundmuller, line 549: nit: Add a small comment clarifying it's usage (i.e. that it is used for the OAuth PKCE flow)

admin/database/database.go
```diff
@@ -515,6 +520,7 @@ const (
 	AuthClientIDRillWeb     = "12345678-0000-0000-0000-000000000001"
 	AuthClientIDRillCLI     = "12345678-0000-0000-0000-000000000002"
 	AuthClientIDRillSupport = "12345678-0000-0000-0000-000000000003"
+	AuthClientIDWebLocal    = "12345678-0000-0000-0000-000000000004"
```
begelundmuller, line 523: nit: `AuthClientIDWebLocal` -> `AuthClientIDRillWebLocal`

admin/database/postgres/migrations/0028.sql
```diff
@@ -0,0 +1,20 @@
+-- Hard-coded first-party auth clients
+INSERT INTO auth_clients (id, display_name)
+VALUES ('12345678-0000-0000-0000-000000000004', 'Web Local');
```
begelundmuller, line 3: Since the display_name may be user-facing, would suggest `'Rill Localhost'`

cli/pkg/local/app.go
pjain1: sure I can change the query param but in the Authenticator it needs to be called this as there is already a redirectURL field for auth flow

begelundmuller (approved):

PR #4844, opened 2024-05-07
CLI: Misc. `rill env` fixes

begelundmuller:
Changes:

- In `rill env configure`, prompt for credentials for all connectors specified under `connectors:` in `rill.yaml`
- In `rill env {show,set,rm}`, infer the project name from the current Git repository if available
- Remove redeploy prompt from `rill env configure` (not needed anymore, new variables are automatically used)
- Fix typo in `rill env set` example in the docs
- Regenerate CLI docs


esevastyanov (approved):
Looks good!

PR #4829, opened 2024-05-06
feat: NOT support in alerts

AdityaHegde:
## Checklist
- [x] Manual verification
- [ ] Unit test coverage
- [ ] E2E test coverage
- [x] Needs manual QA?

#### Issue addressed: 
We have had a few requests to add a NOT support in alert criteria.

#### Details:
1. Adding a `NOT` operation to our expressions.
2. Adding a `NOT` toggle per group.
3. Adding a global `NOT` toggle for the entire list of criteria. (Need design for polish)
4. Adding `NOT` indication to the alert metadata. (Need design for polish)

## Steps to Verify
1. Start cloud dev
2. Create a project with a valid dashboard
3. Create an alert
4. `NOT` toggle should show up and it should apply to the data preview in criteria tab.
5. After creating `!` should show up in the alert metadata

PR #4854, opened 2024-05-08
Cleanup measure filter usage in alerts

AdityaHegde:
This is the common set of cleanup used in #4829 , #4818 and fix for https://github.com/rilldata/rill-private-issues/issues/319


ericpgreen2 (approved):
Nice cleanup!

PR #4875, opened 2024-05-10
feat: Update alert criteria options to have a simpler list

AdityaHegde:
We are updating the list of options available in alert criteria to be a simpler and smaller list.
- `greater than or equals`
- `less than`
- `equals`
- `does not equals`

Older alerts will still work and will show the correct criteria in alert page. But during edit it wont be available anymore.

### DIFF
--- a/web-common/src/features/alerts/criteria-tab/operations.ts
+++ b/web-common/src/features/alerts/criteria-tab/operations.ts
@@ -20,7 +20,7 @@ export const CriteriaOperationOptions = [
   },
   {
     value: MeasureFilterOperation.NotEquals,
-    label: "does not equals",
+    label: "does not equal",
     shortLabel: "!=",
   },
 ];
--- a/web-common/src/features/alerts/criteria-tab/operations.ts
+++ b/web-common/src/features/alerts/criteria-tab/operations.ts
@@ -1,26 +1,27 @@
+import { MeasureFilterOperation } from "@rilldata/web-common/features/dashboards/filters/measure-filters/measure-filter-options";
 import { V1Operation } from "@rilldata/web-common/runtime-client";
 
 // TODO: should match measure filter. remove this once that is merged to main
 export const CriteriaOperationOptions = [
   {
-    value: V1Operation.OPERATION_LT,
-    label: "Less Than",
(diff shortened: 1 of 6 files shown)


web-common/src/features/alerts/criteria-tab/operations.ts
```diff
@@ -1,26 +1,27 @@
...
+    value: MeasureFilterOperation.Equals,
+    label: "equals",
+    shortLabel: "=",
   },
   {
-    value: V1Operation.OPERATION_GTE,
-    label: "Greater Than Or Equals",
-    shortLabel: ">=",
+    value: MeasureFilterOperation.NotEquals,
+    label: "does not equals",
```
ericpgreen2, line 23: Should read "does not equal" (no "s" at the end)

web-common/src/features/dashboards/filters/measure-filters/measure-filter-options.ts
```diff
@@ -73,3 +77,17 @@ export const MeasureFilterOptions: MeasureFilterOption[] = [
     shortLabel: "",
   },
 ];
+// Full list with options not supported in filter pills just yet.
```
ericpgreen2, line 80: Mind creating a GitHub issue to track this work?

ericpgreen2 (approved):

web-common/src/features/dashboards/filters/measure-filters/measure-filter-options.ts
AdityaHegde: I am not sure if this is the end goal. But @mindspank is aware of this, dont think it makes sense to create an issue until we know what we are doing.

ericpgreen2: In the Criteria section of the Alert Metadata, I see that the Criterion pill doesn't handle e.g. the "Does not equal" operator. In the pic below, rather than seeing "New vaccinations != [some number]", it just says "New vaccinations". 

image

I assumed the code comment referred to supporting all operators in the Criterion pills, so the Alert Metadata can showcase all kinds of Criteria.

PR #4876, opened 2024-05-10
Setup gRPC service for the local app

begelundmuller:
So far, the local app has not had a dedicated gRPC service because it had only a few, simple endpoints (`/local/config`, `/local/version`, `/local/track`). As we're looking to support deployments directly from the local UI, we'll be adding some new endpoints for checking the deploy status and triggering deploys. These endpoints will have more complex types and so warrant auto-generated frontend and backend types using protocol buffers.

Notably, this PR uses [Connect](https://connectrpc.com) for generating the gRPC service, instead of the gRPC-gateway+OpenAPI approach we currently use for the runtime and admin services. Connect has several advantages, notably being able to serve gRPC and HTTP on the same port and generating cleaner types. If this integration is successful, we may consider refactoring the runtime and admin service to use it as well (so we can finally get rid of hacks like [this](https://github.com/rilldata/rill/blob/64a38d07522a1d49e21136d5360e8b734b2c2d35/cli/cmd/auth/login.go#L55)).

Changes:
- Adds a new `rill.local.v1.LocalService` gRPC service defined in `proto/rill/local/v1`
- Adds a `GetMetadata` RPC that is equivalent to the current `/local/config` endpoint
- Adds a `GetVersion` RPC that is equivalent to the current `/local/version` endpoint
- Moves the non-gRPC auth/tracking/UI-serving handlers to a new `local.Server` type
- Configures and implements the Connect service in `cli/pkg/local`
- Generates frontend types as a [Connect](https://connectrpc.com) service in `web-common/src/proto/gen/rill/local/v1`
  - See this tutorial for how to query using Connect from TypeScript: https://connectrpc.com/docs/web/getting-started/


proto/rill/local/v1/api.proto
```diff
@@ -0,0 +1,44 @@
...
+
+message PingResponse {
+  google.protobuf.Timestamp time = 1;
+}
+
+message GetMetadataRequest {}
+
+message GetMetadataResponse {
+  string instance_id = 1;
+  string project_path = 4;
```
pjain1, line 27: just curious about the order number here ?

pjain1 (approved):

proto/rill/local/v1/api.proto
begelundmuller: Fixed :)

PR #4885, opened 2024-05-13
Runtime: add a config to enable clickhouse cache

k-anshul:


runtime/drivers/clickhouse/clickhouse.go
```diff
@@ -49,6 +49,8 @@ type driver struct{}
 type configProperties struct {
 	// DSN is the connection string
 	DSN string `mapstructure:"dsn"`
+	// EnableClickhouseCache controls whether to enable cache for Clickhouse queries.
+	EnableClickhouseCache bool `mapstructure:"enable_clickhouse_cache"`
```
pjain1, line 53: this is already part of clickhouse config, can it be just called `enable_cache`

k-anshul: Need to pass set variable `connector.clickhouse.enable_clickhouse_cache=true`

pjain1 (approved):

PR #4889, opened 2024-05-14
Update message and icon for empty pivot with no measure

djbarnwal:
The pivot table is empty when you 
- Only have a column dimension 
- Have a column dimension and a row dimension 

This PR updates the message and icon for those cases.


briangregoryholmes (approved):
Can you just add `noreferrer noopener` to the doc links?

djbarnwal:
Makes sense to add `noopener` Wondering if we should add `noreferrer` as well? We would want to track how many users are coming to the docs from the empty pivot view?

briangregoryholmes:
Totally. I wasn't considering that these were internal links.

PR #4901, opened 2024-05-14
Refactor: Consolidate Editors + FileArtifact updates

### LINKED ISSUE #4688: Refactor(file explorer): Use one common Editor for all files/resources

briangregoryholmes:
This PR consolidates editors across the application to use the single generic Editor component.

Additionally, it moves the file mutation functions and local/remote content caches to the FileArtifact class.

Because of this update, it is no longer necessary to warn a user that unsaved changes will be lost when navigating. While product/design input is necessary to make this fully featured, this PR adjusts the FileExplorer to display unsaved files with 50% opacity and adds a save all key binding.

- Removes the queryHighlight writable and instead uses the global event bus to achieve the same functionality
- Resolves some quirks around asynchronous updates the Editor by using `queueMicrotask`. May address this in a future PR by trying to integrate our linting with CodeMirror in a more canonical way. This is skipped when running unit tests for the component.
- Adds a confirmation popup when a user has unsaved local changes but then receives an update to the remote content
- Adds a 500ms timeout the `updateCodeEditor` helper function used during testing

Closes #4688 

Unsaved file indicators:
Screenshot 2024-06-13 at 10 21 20 PM

File sync warnings:
Screenshot 2024-06-13 at 10 21 01 PM


rill-dev:
## Code Review Agent Run Status

- **Limitations and other issues:** Failure - Whoops! The AI Code Review Agent did not run because the source or target branch includes the main branch. Please contact support@bito.ai to change this.

rill-dev:

web-local/src/routes/(application)/custom-dashboards/[name]/+page.svelte
```diff
@@ -148,6 +120,7 @@
       dimensions: Vector;
     }>,
   ) {
+    const parsedDocument = parseDocument($localContent ?? $remoteContent ?? "");
```
Comment (rill-dev) [inline web-local/src/routes/(application)/custom-dashboards/[name]/+page.svelte:123]: **Issue**: The use of nullish coalescing with empty string fallback in 'parseDocument' might lead to unexpected behavior if both '$localContent' and '$remoteContent' are null or undefined. This could result in parsing an empty string which might not be the intended behavior and could lead to errors downstream if the parsed document structure is expected to have specific properties or structure. <br> **Fix**: Ensure that 'parseDocument' is called with valid data. Consider handling cases where both '$localContent' and '$remoteContent' are null or undefined more explicitly, possibly by not calling 'parseDocument' at all, or by initializing '$localContent' and '$remoteContent' to a valid default state elsewhere in your application. <br> **Code Suggestion**: 
 ```
 - const parsedDocument = parseDocument($localContent ?? $remoteContent ?? "");
 + if (!$localContent && !$remoteContent) return; // Handle or log error appropriately
 + const parsedDocument = parseDocument($localContent ?? $remoteContent);
 ```

web-common/src/components/editor/line-status/index.ts
```diff
@@ -13,14 +13,26 @@ import {
  * bindable prop and which tends to be accessible from edit events dispatched via
  * the dispatch-events extension in the editor component directory.
  */
-export function setLineStatuses(lineStatuses: LineStatus[], view: EditorView) {
+export function setLineStatuses(
+  lineStatuses: LineStatus[],
+  view: EditorView,
+  wait = true,
```
rill-dev, line 19: **Issue**: The default parameter 'wait = true' in the 'setLineStatuses' function can lead to unexpected behavior when not explicitly set by all callers, potentially causing race conditions or UI inconsistencies due to the use of 'queueMicrotask'. <br> **Fix**: Remove the default parameter and enforce explicit parameter passing to ensure that callers are aware of the behavior they are invoking. This change will make the control flow clearer and prevent potential bugs related to task scheduling. <br> **Code Suggestion**: 
 ```
 -  wait = true,
 +  wait,
 ```

web-common/src/features/models/workspace/inspector/References.svelte
```diff
@@ -84,8 +74,12 @@
...
                 class="ui-copy-muted grid justify-between gap-x-2 {query_reference_trigger} pl-4 pr-4"
                 style:grid-template-columns="auto max-content"
-                on:focus={focus(reference.reference)}
-                on:mouseover={focus(reference.reference)}
+                on:focus={() => {
+                  eventBus.emit("highlightSelection", [reference.reference]);
+                }}
+                on:mouseover={() => {
+                  eventBus.emit("highlightSelection", [reference.reference]);
+                }}
```
rill-dev, line 82: **Issue**: The new event handlers for focus and mouseover events emit a "highlightSelection" event without checking if the reference is valid. This could potentially lead to emitting events with undefined or null values, which might cause runtime errors or unexpected behavior in other parts of the application that listen to this event. <br> **Fix**: Add null checks before emitting the event to ensure that the reference is valid. This will prevent potential errors in other parts of the application. <br> **Code Suggestion**: 
 ```
 @@ -77,5 +77,5 @@
 -                on:focus={() => {
 -                  eventBus.emit("highlightSelection", [reference.reference]);
 -                }}
 -                on:mouseover={() => {
 -                  eventBus.emit("highlightSelection", [reference.reference]);
 +                on:focus={() => reference && eventBus.emit("highlightSelection", [reference.reference])}
 +                on:mouseover={() => reference && eventBus.emit("highlightSelection", [reference.reference])}
 ```

web-common/src/features/editor/Codespace.svelte
```diff
@@ -0,0 +1,177 @@
...
+      state: EditorState.create({
+        doc: $localContent ?? $remoteContent ?? undefined,
+        extensions: [
+          baseExtensions(),
+          extensionCompartment.of([]),
+          EditorView.updateListener.of(({ docChanged, state: { doc } }) => {
+            if (editor?.hasFocus && docChanged) {
+              updateLocalContent(doc.toString());
+
+              if (!disableAutoSave && autoSave) debounceSave();
```
rill-dev, line 56: **Issue**: The debounceSave function is called without any debouncing mechanism, which can lead to excessive and unnecessary save operations that may impact performance and cause race conditions. <br> **Fix**: Implement a debouncing mechanism to limit the rate at which save operations are performed. This can be achieved by using a timeout to delay the execution of the save operation and ensure that it is not called more frequently than necessary. <br> **Code Suggestion**: 
 ```
 -              if (!disableAutoSave && autoSave) debounceSave();
 +              if (!disableAutoSave && autoSave) {
 +                clearTimeout(debounceTimer);
 +                debounceTimer = setTimeout(debounceSave, 300);
 +              }
 ```

web-common/src/features/models/workspace/ModelEditor.svelte
```diff
@@ -141,230 +85,29 @@
...
   function updateAutocompleteSources(
     schema: { [table: string]: string[] },
     defaultTable?: string,
   ) {
     if (editor) {
-      editor.dispatch({
-        effects: autocompleteCompartment.reconfigure(
-          makeAutocompleteConfig(schema, defaultTable),
-        ),
+      queueMicrotask(() => {
```
rill-dev, line 93: **Issue**: Using 'queueMicrotask' to dispatch editor updates can lead to uncontrolled execution timing, potentially causing race conditions or UI inconsistencies, especially in a complex UI environment where multiple tasks might be queued simultaneously. <br> **Fix**: Replace 'queueMicrotask' with a more controlled asynchronous handling mechanism such as 'setTimeout' with a defined delay, or integrate a state management solution that handles such updates more predictably. <br> **Code Suggestion**: 
 ```
 - queueMicrotask(() => {
 + setTimeout(() => {
 ```

web-common/src/features/metrics-views/workspace/editor/MetricsEditor.svelte
```diff
@@ -1,51 +1,66 @@
...
-    bind:view
-    content={yaml}
-    extensions={[placeholderElements.extension, yamlSchema(metricsJsonSchema)]}
-    whenFocused
-    on:save={updateMetrics}
+<MetricsEditorContainer error={$remoteContent ? mainError : undefined}>
+  <Editor
+    bind:autoSave
+    bind:editor
+    onSave={(content) => {
```
rill-dev, line 49: **Issue**: The onSave callback function does not handle asynchronous operations properly, which can lead to race conditions or unhandled promise rejections if the operations inside it fail. <br> **Fix**: Modify the onSave callback to properly handle asynchronous operations using async/await to ensure that errors are caught and handled properly. <br> **Code Suggestion**: 
 ```
 @@ -49,7 +49,11 @@
 onSave={(content) => {
 -  metricsExplorerStore.remove(metricViewName);
 -  createPersistentDashboardStore(metricViewName).reset();
 +  (async () => {
 +    try {
 +      await metricsExplorerStore.remove(metricViewName);
 +      await createPersistentDashboardStore(metricViewName).reset();
 +    } catch (error) {
 +      console.error('Failed to save metrics:', error);
 +    }
 +  })();
   if (!content?.length) {
     setLineStatuses([], editor);
   }
 }}
 ```

web-common/src/features/editor/Codespace.svelte
briangregoryholmes: 😒

ericpgreen2 (changes requested):
This cleanup will be a huge help!

UXQA –
1. Because we also gray-out dotfiles, it’s hard to tell the semantic difference between a dotfile and an unsaved file. Maybe we could use a gray dot in the sidebar, like we do in the Workspace Header? 
image
image
2. The dashboard workspace header has no unsaved file indicator, like the other workspaces do
3. Here’s a dashboard YAML that’s errored because it doesn’t have any dimensions. The editor is not showing a message with the error (and the file name in the sidebar is not red, and the preview button in the top-right is not disabled). image
4. I ran into this type error:
image

briangregoryholmes:
Point 3 is true on `main` as well. Looks like this isn't considered a parse error. Not really sure if it should be. @begelundmuller

web-common/src/features/metrics-views/workspace/editor/MetricsEditor.svelte
briangregoryholmes: These functions are not async.

web-common/src/features/models/workspace/inspector/References.svelte
briangregoryholmes: The types don't suggest that reference can be undefined. I think this is being thrown off by the fact that the `?` operator is being used unnecessarily in other parts of this file.

ericpgreen2:
UXQA: the workspace header for files like `rill.yaml`, `.env`, and `.gitignore` do not have a gray dot unsaved-file indicator, like other files do.

Design: I like the italics to indicate unsaved files. Like you mentioned, I agree it could also be nice to show a number somewhere that indicates the number of unsaved files, like VSCode. If so (up to you), this could be a place to add a tooltip to surface the "save all files" keyboard shortcut.

ericpgreen2 (approved):

web-common/src/features/workspaces/MetricsWorkspace.svelte
```diff
@@ -117,49 +77,51 @@
...
-        />
-      </div>
-    </WorkspaceHeader>
-
-    <MetricsEditor slot="body" {yaml} {filePath} {allErrors} {metricViewName} />
-
-    <MetricsInspector {filePath} slot="inspector" />
-  </WorkspaceContainer>
-{/if}
+<svelte:window on:focus={() => fileArtifact.fetchContent(true)} />
```
ericpgreen2, line 87: Looks like this should apply to all workspaces?

ericpgreen2 (commented):
I navigated to a `data/Adbids.csv` file, and I got this "File not found" error:

image

web-common/src/features/entity-management/file-artifact.ts
```diff
@@ -0,0 +1,297 @@
...
+
+    const queryFn: QueryFunction<
+      Awaited<ReturnType<typeof runtimeServiceGetFile>>
+    > = ({ signal }) => runtimeServiceGetFile(instanceId, queryParams, signal);
+
+    const { blob } = await queryClient.fetchQuery({
+      queryKey,
+      queryFn,
+      staleTime: Infinity,
+    });
```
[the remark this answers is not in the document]
briangregoryholmes: I believe this is correct. It's definitely still on my mind to figure out a way to make this cleaner, but it should be operating as intended.

`invalidateQueries` doesn't _really_ refetch anything. It marks queries as stale and, if they're being actively rendered using `createQuery`, then they'll be refetched and subscribers will be updated.

However, with this PR, there are no longer components that are subscribed to any file queries via `createRuntimeServiceGetFile`. This is mostly CodeMirror's "fault", whose API does not mesh well with the Svelte approach of reactive functions or derived queries.

If you try to comment out the invalidateQueries line or early return immediately after it, you'll see that the network request won't actually happen when you make a change in your IDE. You could set `{ refetchType: "all" }` on the invalidation and early return, which would fire the network request, but it wouldn't propagate to the editor because `updateRemoteContent`, which handles the callbacks and writable update, wasn't called.

web-common/src/features/workspaces/MetricsWorkspace.svelte
briangregoryholmes: I don't really have a preference, but, just to confirm, this would be a behavior change. If  you look at the codebase now, the only place where `refetchOnWindowFocus` is true is the dashboard editor page. I was thinking about moving this to the Editor component and having it be configurable via a prop. Is there any file type this shouldn't apply to?

web-common/src/features/entity-management/file-artifact.ts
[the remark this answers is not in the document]
ericpgreen2: Ok, thanks for the detail! Makes sense that this is needed after removing component-native Query Observers.

web-common/src/features/workspaces/MetricsWorkspace.svelte
ericpgreen2: Oh, got it. Seems like it should apply to all workspaces, as long as unchanged files don't trigger needless ingest and profiling on the Source and Model workspaces.

djbarnwal (approved):

AdityaHegde (approved):

PR #4905, opened 2024-05-15
feat: Global dimension search in metrics view in both rill dev and cloud

### LINKED ISSUE #4882: Feat: Global Dimension Search
https://www.notion.so/rilldata/Global-Dimension-Search-f1281d7380e648468364c9adb00d0dfd?pvs=4

AdityaHegde:
closes #4882

This adds a search button at the top right corner for both rill dev and cloud.
- [x] Add the search button to make a call to batch API.
- [x] Cache results of the search.
- [x] Polish UX and match mocks.


ericpgreen2 (changes requested):
Overall it looks pretty good! 

UXQA –
1. If a search query errored, we should say so. Here I received "context deadline exceeded" when searching against a Druid-backed dashboard.
image
2. There's very poor performance when searching Druid-backed dashboards. Mind opening up a conversation with Platform to confirm the batch query approach, or if they recommend a different query pattern?
3. It'd be nice to submit the form after a pause in typing. If so, subsequent typing should cancel an outstanding request.
4. On an Adbids dashboard in Rill Developer, I searched for "facebook" and the search never finished searching the `Domain` dimension. I see the below error in my browser console.
image
image
5. When a dimension value is already set as a dashboard filter, it'd be nice if there were an indicative checkmark in the search results.
image
6. For parity with the Enterprise application, the previous search input text should be retained when re-opening the search box.
7. In Rill Developer, the search button should be to the left of the Deploy & Edit Metrics buttons.
8. Nit: When there are no results yet, there shouldn't be the extra empty bottom section below the `DropdownMenuSeparator`.
image
9. Nit: After using the down arrow key to focus on the first menu item, I'd like the up arrow key to re-focus on the search input.

ericpgreen2:
As mentioned in the demo, let's add a feature flag for this so that we can iterate on the Druid performance in a follow-on PR.

ericpgreen2 (approved):
Looks good. One more UXQA finding:
- The "x" button shouldn't have this background color:
image

package-lock.json
ericpgreen2, line 1: Confirming this `package-lock.json` modification is intended? I'm not seeing any `package.json` changes in this PR.

AdityaHegde: Nope. Will revert

AdityaHegde:
Merging this behind a feature flag: `dimensionSearch`

PR #4909, opened 2024-05-15
runtime-local apis for project deployment

pjain1:
This PR adds 3 new APIs to support auto-deploy from UI.

1. `DeployValidation` - It does basic validation and returns a response which can indicate further action to be taken by the client 
    1. `is_authenticated` - If false then client should redirect to `/auth` to get user signed in, other fields in the response should be ignored, after login the validation should be done again to get latest info. A `redirect` query can be set with url that the runtime should redirect to after login
    2. `is_github_connected` - User does have not rill git app installed, so redirect user to `git_grant_access_url` which is the next field.
    3. `git_grant_access_url` - git access url
    4. `git_user_name` - git user name, also the default git org for the user.
    5. `git_user_orgs` - git orgs that user is part of. Currently not used, we use the default org which is equivalent to git username
    6. `is_git_repo` - is the rill project folder a git repo ? meaning does it have .git folder. If this is true, then in first version we should just ask to continue in cli.
    7. `git_url` - first git remote of the git repo, only if is_git_repo is true.
    8. `uncommited_changes` - if its already a git repo then are there any uncommitted changes that are not pushed to remote git. In first version this is not used as we only allow deploying from folder that are not git repo
    9. `rill_org_exists_as_git_user_name` - If a rill org already exists as the git_user_name, this is useful because if no orgs exists for user then we auto create an org with git_user_name, so if this is true then ask user for a new org name. This is applicable only for new user those who don’t have any org. For existing user see next option.
    10. `rill_user_orgs` - User can choose a rill org to deploy to.
    11. `local_project_name` - Folder name of local project from where `rill start` is done
    
2. `PushToGit` - This is only used when `is_git_repo` is false and all other validation passes. This will create a git repo in the existing project folder from where `rill start` is done, auto commit all changes and push the project to git org named `git_user_name` in a project named the folder name of the local project.
3. `Deploy` - Should be done when all DeployValidation passes and it accepts two parameters `rill_org` and `rill_project_name` . We will ask user to enter these, can be auto populated as per DeployValidation response.


pjain1:
added 3 more fields to `DeployValidationResponse` -
1. `user_app_permission` - if user has provided read or write access to the rill git app on their personal account
2. `git_remote_found` - if `is_git_repo` is true then if there exists a github remote 
3. `is_repo_access_granted` - when the local project has git remote meaning `is_git_repo` is true and `git_remote_found` is also true then user should have give access to repo by authorizing app for the repo, if not redirect to `git_grant_access_url`
4. `organizations_with_app`- List of orgs on which user has installed the rill app along with permission given. If user choses to push to one of these or deploy to one of this then we should - first have app installed and write permission for push and read permission for deploying. This list will not contain users' personal account name, `git_user_name` and `user_app_permission` can be used for this.
 
`uncommitted_changes` is change to enum and now have `UNCOMMITTED_CHANGES_UNSPECIFIED` to indicate we do not know the status so user should ensure repo is in sync with remote.

pjain1:
One issue I am seeing currently is that `gitClient.Organizations.List(ctx, user.GithubUsername, nil)` does not list all my orgs however when I explicitly do `s.admin.Github.AppClient().Apps.FindOrganizationInstallation(ctx, "<org>")` for that org it works.

proto/rill/admin/v1/api.proto
```diff
@@ -1127,6 +1127,14 @@ message GetGithubUserStatusResponse {
...
   repeated string organizations = 5;
+  bool is_user_app_installed = 6;
+  string user_installation_permission =7;
+  repeated OrganizationWithApp organizations_with_app = 8;
+}
+
+message OrganizationWithApp {
+  string org = 1;
+  string permission = 2;
 }
```
begelundmuller, line 1138: What do you think about using `map<string, string> organization_installation_permissions` (doesn't require a separate type)?

proto/rill/admin/v1/api.proto
```diff
@@ -1127,6 +1127,14 @@ message GetGithubUserStatusResponse {
   string access_token = 3;
   string account = 4;
   repeated string organizations = 5;
+  bool is_user_app_installed = 6;
+  string user_installation_permission =7;
```
begelundmuller, line 1131: Could `is_user_app_installed` be removed? Seems like `user_installation_permissions` already indicates it (i.e. by not having any permissions if not installed). Or is there a difference?

proto/rill/local/v1/api.proto
```diff
@@ -42,3 +51,56 @@ message GetVersionResponse {
   string current = 1;
   string latest = 2;
 }
+
+message DeployValidationRequest {}
+
+message DeployValidationResponse {
+  bool is_authenticated = 1; // if true below fields are relevant after login
+  bool is_github_connected = 2; // if true below fields are relevant after git install
+  string git_grant_access_url = 3; // redirect to this if is_github_connected or is_repo_access_granted is false
```
begelundmuller, line 60: Change to `github_grant_access_url` since it's not native to Git itself

proto/rill/local/v1/api.proto
```diff
@@ -42,3 +51,56 @@ message GetVersionResponse {
...
+  string git_user_name = 4;
```
begelundmuller, line 61: Is this the Git username or Github username? If the latter, use `github_username`

proto/rill/local/v1/api.proto
```diff
@@ -42,3 +51,56 @@ message GetVersionResponse {
...
+  bool is_user_app_installed = 5; // if rill git app is installed on user account
+  string user_app_permission = 6;
```
begelundmuller, line 63: 1. Same question as in the admin protos
2. Since it's Github-related, put a ` github_` prefix in the name
3. Since it seems related to `organizations_with_app` (both dealing in Github access permissions), can we move that up next to this and have similar names between the two?

proto/rill/local/v1/api.proto
```diff
@@ -42,3 +51,56 @@ message GetVersionResponse {
...
+  repeated string git_user_orgs = 7;
```
begelundmuller, line 64: 1. How is this different from `organizations_with_app`?
2. Also, "Github" not "Git" :)

proto/rill/local/v1/api.proto
```diff
@@ -42,3 +51,56 @@ message GetVersionResponse {
...
+  bool is_git_repo = 8;
+  bool git_remote_found = 9; // only applicable when is_git_repo is true
+  bool is_repo_access_granted = 10; // relevant only when is_git_repo is true and remote found, if false redirect to git_grant_access_url
```
begelundmuller, line 67: Since it's github-specific, put `github` in the name. Also consider moving it next to the Github-related properties

proto/rill/local/v1/api.proto
```diff
@@ -42,3 +51,56 @@ message GetVersionResponse {
...
+  repeated string rill_user_orgs = 14;
+  string local_project_name = 15;
+  repeated OrganizationWithApp organizations_with_app = 16;
+}
+
+enum UncommittedChanges {
+  UNCOMMITTED_CHANGES_UNSPECIFIED = 0;
+  UNCOMMITTED_CHANGES_YES = 1;
+  UNCOMMITTED_CHANGES_NO = 2;
+}
```
begelundmuller, line 80: Can be replaced by an optional bool, e.g. `optional bool has_uncommitted_changes`

proto/rill/local/v1/api.proto
```diff
@@ -42,3 +51,56 @@ message GetVersionResponse {
...
+  string permission = 2;
+}
+
+message PushToGitRequest {
+  string org = 1;
+  string repo = 2;
+}
+
+message PushToGitResponse {
+  string git_url = 1;
```
begelundmuller, line 93: Is this a Git URL (ends in `.git`) or a Github URL (for opening in a browser)? If the second, change to `github_`

proto/rill/local/v1/api.proto
```diff
@@ -42,3 +51,56 @@ message GetVersionResponse {
...
+}
+
+message DeployRequest {
+  string rill_org = 1; // creates the org if does not exists, no validation
+  string rill_project_name = 2; // no validation, assuming user provided correct name as per the prompts
```
begelundmuller, line 98: Can remove `rill` prefix – it's implied since this is a Rill API

admin/server/github.go
```diff
@@ -86,32 +87,82 @@ func (s *Server) GetGithubUserStatus(ctx context.Context, req *adminv1.GetGithub
...
+	userInstallationPermission := ""
 	installation, _, err := s.admin.Github.AppClient().Apps.FindUserInstallation(ctx, user.GithubUsername)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get user installation: %w", err)
+		if !strings.Contains(err.Error(), "404") {
+			return nil, fmt.Errorf("failed to get user installation: %w", err)
+		}
+	} else {
+		isUserAppInstalled = true
+		userInstallationPermission = "read"
```
begelundmuller, line 99: Can we add an enum like `GithubPermission` in the proto for this instead of a string?

admin/server/github.go
```diff
@@ -86,32 +87,82 @@ func (s *Server) GetGithubUserStatus(ctx context.Context, req *adminv1.GetGithub
...
+		}
+		permission := "read"
+		// older git app would ask for Contents=read permission whereas new one asks for Contents=write and && Administration=write
+		if i.Permissions != nil && i.Permissions.Contents != nil {
+			if i.Permissions.Administration != nil && strings.EqualFold(*i.Permissions.Administration, "write") && strings.EqualFold(*i.Permissions.Contents, "write") {
+				permission = "write"
+			}
+		} else {
+			return nil, status.Errorf(codes.Internal, "failed to get organization permissions for org %s", org.GetLogin())
+		}
```
begelundmuller, line 150: Can get rid of a level of nesting by reverting the `if` expression and erroring early

begelundmuller (changes requested):

proto/rill/admin/v1/api.proto
pjain1: yeah it can be removed, wanted to keep it separate to keep things clean but can be removed and if `user_installation_permissions` is empty string then app is not installed on user github account ?

proto/rill/local/v1/api.proto
pjain1: github

proto/rill/local/v1/api.proto
pjain1: This list all the users orgs where as `organizations_with_app` lists only the orgs on which Rill github app is installed. not sure what would be the use of `git_user_orgs`, may be we can merge both the lists and orgs on which app is not installed we set the permission as empty string.

proto/rill/local/v1/api.proto
pjain1: I thought using optional is not recommended

cli/cmd/deploy/deploy.go
```diff
@@ -460,15 +460,39 @@ func createGithubRepoFlow(ctx context.Context, ch *cmdutil.Helper, localGitPath
...
 			return err
 		}
+		if !ok {
+			ch.PrintfWarn("\nIf you want to deploy to another Github account, Please visit %q to grant access.\n", pollRes.GrantAccessUrl)
+			return nil
+		}
+	} else {
+		repoOwner, err = cmdutil.SelectPrompt("Rill has write access to following Github accounts. Please select one", candidateOrgs, candidateOrgs[0])
+		if err != nil {
+			ch.PrintfWarn("\nIf you want to deploy to another Github account, Please visit %q to grant access.\n", pollRes.GrantAccessUrl)
```
begelundmuller, line 491:
```suggestion
			ch.PrintfWarn("\nIf you want to deploy to another Github account, visit this URL to grant access: %s\n", pollRes.GrantAccessUrl)
```

cli/cmd/deploy/deploy.go
```diff
@@ -460,15 +460,39 @@ func createGithubRepoFlow(ctx context.Context, ch *cmdutil.Helper, localGitPath
...
+		ok, err = cmdutil.ConfirmPrompt(fmt.Sprintf("Rill has write access to %q Github account. Do you want to use this account?", repoOwner), "", true)
 		if err != nil {
 			return err
 		}
+		if !ok {
+			ch.PrintfWarn("\nIf you want to deploy to another Github account, Please visit %q to grant access.\n", pollRes.GrantAccessUrl)
+			return nil
+		}
+	} else {
+		repoOwner, err = cmdutil.SelectPrompt("Rill has write access to following Github accounts. Please select one", candidateOrgs, candidateOrgs[0])
```
begelundmuller, line 489:
```suggestion
		repoOwner, err = cmdutil.SelectPrompt("Select a Github account for the new repository", candidateOrgs, candidateOrgs[0])
```

cli/pkg/local/server.go
```diff
@@ -90,6 +103,394 @@ func (s *Server) GetVersion(ctx context.Context, r *connect.Request[localv1.GetV
...
+	resp, err := c.ListOrganizations(ctx, &adminv1.ListOrganizationsRequest{})
+	if err != nil {
+		return nil, err
+	}
+
+	userOrgs := make([]string, 0, len(resp.Organizations))
+	for _, org := range resp.Organizations {
+		userOrgs = append(userOrgs, org.Name)
+	}
+	// TODO if len(userOrgs) > 0 then check if any project in these orgs already deploys from ghUrl, however I think this is not enough, probably should be checked globally ?
```
begelundmuller, line 229: Is this important to address?

cli/pkg/local/server.go
```diff
@@ -90,6 +103,394 @@ func (s *Server) GetVersion(ctx context.Context, r *connect.Request[localv1.GetV
...
+		if err != nil {
+			if !errors.Is(err, database.ErrNotFound) {
+				return nil, err
+			}
+		} else {
+			rillOrgExistsAsGitUserName = true
+		}
+	}
+
+	// TODO check if any project exists with name localProjectName in any of the users rill org
```
begelundmuller, line 247: I don't think this is important, but what do you think?

cli/pkg/local/server.go
```diff
@@ -90,6 +103,394 @@ func (s *Server) GetVersion(ctx context.Context, r *connect.Request[localv1.GetV
...
+		repoName = r.Msg.Repo
+	}
+	githubClient := github.NewTokenClient(ctx, gitStatus.AccessToken)
+	defaultBranch := "main"
+
+	// create remote repo
+	githubRepo, _, err := githubClient.Repositories.Create(ctx, githubOrg, &github.Repository{Name: &repoName, DefaultBranch: &defaultBranch})
+	if err != nil {
+		if !strings.Contains(err.Error(), "name already exists") {
+			// TODO should we retry with new name, append a number to repoName like -1, -2 etc
```
begelundmuller, line 340: Yes, I think that was actually also requested by Nishant

cli/pkg/local/server.go
```diff
@@ -90,6 +103,394 @@ func (s *Server) GetVersion(ctx context.Context, r *connect.Request[localv1.GetV
...
+		ProdOlapDriver:   "",
+		ProdOlapDsn:      "",
+		ProdSlots:        2,
+		Subpath:          "",
+		ProdBranch:       repoStatus.DefaultBranch,
+		Public:           false,
+		GithubUrl:        ghURL,
+	})
+	if err != nil {
+		// TODO if project with same name exists - should we retry with new name, append a number to repoName like -1, -2 etc
```
begelundmuller, line 482: Yeah, same as comment above

proto/rill/admin/v1/api.proto
```diff
@@ -1126,7 +1126,16 @@ message GetGithubUserStatusResponse {
   string grant_access_url = 2;
   string access_token = 3;
   string account = 4;
-  repeated string organizations = 5;
+  // DEPRECATED: Use organization_installation_permissions instead.
+  repeated string organizations = 5 [deprecated = true];
+  GithubPermission user_installation_permission = 6;
+  map<string, GithubPermission> organization_installation_permissions = 7;
```
begelundmuller, line 1132:
```suggestion
  GithubPermission user_installation_permission = 6;
  map<string, GithubPermission> organization_installation_permissions = 7;
  // DEPRECATED: Use organization_installation_permissions instead.
  repeated string organizations = 5 [deprecated = true];
```

proto/rill/local/v1/api.proto
```diff
@@ -42,3 +52,44 @@ message GetVersionResponse {
...
+message DeployValidationRequest {}
+
+message DeployValidationResponse {
+  bool is_authenticated = 1; // if true below fields are relevant after login
+  string login_url = 2; // redirect to this if is_authenticated is false
+  bool is_github_connected = 3; // if true below fields are relevant after github install
+  string github_grant_access_url = 4; // redirect to this if is_github_connected or is_github_repo_access_granted is false
+  string github_user_name = 5;
+  rill.admin.v1.GithubPermission github_app_user_permission = 6; // if unspecified then github app not installed on user account
+  map<string, rill.admin.v1.GithubPermission> github_organization_installation_permissions = 7;
```
begelundmuller, line 65: Is there a reason to use "app" in the first name and "installation" in the second? Otherwise, could consider making the names more similar, e.g. `github_user_permissions` and `github_organization_permissions`.

cli/pkg/local/server.go
pjain1: cli flow checks this [here](https://github.com/rilldata/rill/blob/main/cli/cmd/deploy/deploy.go#L285) but it only check in the current org, may be its fine as current user might not have access to another org. There 3 actions that user can take, 2 should be taken from cli and one is continue deployment with another project name, so was not sure how this should be surfaced so did not added this validation. Can be a follow up PR, may be just add flag to indicate this and UI can just say - `Another project deploy from this repo, to abort press abort or continue with another name` ?

cli/pkg/local/server.go
pjain1: once we add retires with suffixes in case of conflict this is not that important

cli/pkg/local/server.go
pjain1: handled, retry 3 times

cli/pkg/local/server.go
pjain1: handled, retry 3 times

begelundmuller (approved):
LGTM

PR #4919, opened 2024-05-16
Embed DuckDB extensions in the Rill binary

### LINKED ISSUE #4574: Embed DuckDB extensions in the Rill binary
At present we download the binaries when starting rill, which can fail if duckDB CDN is having issues. 
```
HTTP Error: Failed to download extension "json" at URL "http://nightly-extensions.duckdb.org/v0.9.2/osx_arm64/json.duckdb_extension.gz" Extension "json" is an existing extension.
```

We would like to avoid downloading the extensions at runtime.

esevastyanov:
Closes #4574

### DIFF
--- a/runtime/drivers/duckdb/extensions.go
+++ b/runtime/drivers/duckdb/extensions.go
@@ -1,0 +1,134 @@
+package duckdb
+
+import (
+	"compress/gzip"
+	"database/sql"
+	"embed"
+	"errors"
+	"fmt"
+	"io"
+	"io/fs"
+	"log"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+// Since DuckDB is called from multiple packages, the extensions are installed in the init function
+func init() {
+	err := installExtensions()
+	if err != nil {
+		// If extensions cannot be installed, log the error and continue as the extensions can be downloaded
+		// Should it be fatal in order to notice the issue prior to a release?
+		log.Printf("Error preparing DuckDB extensions: %v", err)
+	}
+}
+
+//go:embed embed/extensions/**/*
+var embeddedFiles embed.FS
+
+// installExtensions copies embedded DuckDB extensions to DuckDB's extensions directory to avoid downloading them
+func installExtensions() error {
+	// Connect to DuckDB and get the version
+	db, err := sql.Open("duckdb", "")
+	if err != nil {
+		return err
+	}
+	defer db.Close()
+
+	var duckdbVersion string
+	err = db.QueryRow("SELECT version();").Scan(&duckdbVersion)
+	if err != nil {
+		return err
+	}
+
+	// Get the platform name using PRAGMA
+	var platformName string
+	err = db.QueryRow("PRAGMA platform;").Scan(&platformName)
+	if err != nil {
+		return err
+	}
+
+	// Define source and destination paths
+	embedPath := fmt.Sprintf("embed/extensions/%s/%s", duckdbVersion, platformName)
+	duckdbExtensionsPath := filepath.Join(os.Getenv("HOME"), ".duckdb", "extensions", duckdbVersion, platformName)
+
+	// Create the destination directory if it doesn't exist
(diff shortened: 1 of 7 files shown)


runtime/drivers/duckdb/extensions.go
```diff
@@ -0,0 +1,134 @@
...
+	// Get the platform name using PRAGMA
+	var platformName string
+	err = db.QueryRow("PRAGMA platform;").Scan(&platformName)
+	if err != nil {
+		return err
+	}
+
+	// Define source and destination paths
+	embedPath := fmt.Sprintf("embed/extensions/%s/%s", duckdbVersion, platformName)
+	duckdbExtensionsPath := filepath.Join(os.Getenv("HOME"), ".duckdb", "extensions", duckdbVersion, platformName)
```
begelundmuller, line 54: Could we use `os.UserHomeDir()` instead of `Getenv`?

runtime/drivers/duckdb/extensions.go
```diff
@@ -0,0 +1,134 @@
...
+	defer gzipReader.Close()
+
+	output, err := os.Create(dest)
+	if err != nil {
+		return err
+	}
+	defer output.Close()
+
+	for {
+		_, err = io.CopyN(output, gzipReader, 1024) // CopyN is used to prevent a warning (G110: Potential DoS vulnerability...)
```
begelundmuller, line 124: If you prefer `io.Copy`, adding a `// nolint` is also okay IMO, since we trust the source of the data

runtime/drivers/duckdb/extensions.go
```diff
@@ -0,0 +1,134 @@
...
+
+// Since DuckDB is called from multiple packages, the extensions are installed in the init function
+func init() {
+	err := installExtensions()
+	if err != nil {
+		// If extensions cannot be installed, log the error and continue as the extensions can be downloaded
+		// Should it be fatal in order to notice the issue prior to a release?
+		log.Printf("Error preparing DuckDB extensions: %v", err)
+	}
+}
```
begelundmuller, line 25: I think this might delay process start. Can we move this to lazy evaluation, e.g. using a `sync.Once` global that is called when the first DuckDB handle is opened?

runtime/drivers/duckdb/extensions.go
```diff
@@ -0,0 +1,134 @@
...
+
+//go:embed embed/extensions/*
```
begelundmuller, line 27: To avoid accidental commits, can we add the extension files here to `gitignore`? Might need a nested pattern like `runtime/drivers/duckdb/embed/extensions/*/*` or something like that to still maintain the `.gitkeep` file.

runtime/drivers/duckdb/extensions.go
esevastyanov: Do you expect something similar to this?
```
var once sync.Once
func (c *connection) prepareExtensions() error {
	var err error
	once.Do(func() {
		err = installExtensions()
	})
	return err
}
```
However DuckDB is also used in `duckdbsql` package (`query` [function](https://github.com/rilldata/rill/blob/main/runtime/pkg/duckdbsql/duckdbsql.go#L42)). This function is called on app start by `emitStartEvent`.

begelundmuller: Yeah something like that. You could make it a public function, like `duckdb.PrepareExtensions`, and call it from the `duckdbsql` package before loading an extension.

I understand in practice it gets loaded fast for `rill start`, but there are other CLI commands like `rill user add` etc. that don't need DuckDB, and also some tests that don't need it but might still load the DuckDB driver.

esevastyanov: Moved to separate package to resolve a circular dependency and also I decided to ignore the case if no extensions are embedded at all as this might be noisy in dev runs and tests

begelundmuller (approved):
Looks great!

PR #4883, opened 2024-05-13
fix: Reduce redundant network calls in rill developer

AdityaHegde:
We have multiple redundant network calls in rill developer.
- [x] During initial load we make a call to all resource and all files names. Separately we also make a call to all resources, this can be filled in using all resources call.
- [x] WatchResourceClient receives the updated resource, but we still make another call to it.
- [x] We refresh list of all file names for every file write. We can only do this when a new file is added or file is deleted.
- [x] We make a call to ListResources with `kind` in addition to the ListResource with no params. We can filter from the full list instead. (Dashboard list is pending since it used in cloud where we might not always make a call to list resources)

This also attempts to move any metadata needed for these watch clients to those files instead of the global store FileArtifacts.
- [x] Add WatchFileClient to store any metadata needed only during watching files.
- [x] Add WatchResourceClient to store metadata needed only during watching resources.
- [ ] Finally fix E2E and make sure there are no regressions.

PR #4923, opened 2024-05-17
fix: Clearing a dashboard yaml stops updating errors

### LINKED ISSUE #4918: SQL Editor: latest reconcile error is sometimes not shown in editor
After various SQL editing, I've run into the following state:
- In Rill Developer, I've previously run `select * from "Adbds"`, and I see its accompanying preview table
- In Rill Developer, I've edited my query to `select abc from "Adbids"`, and the prior query's preview table remains. I should see an error that "column `abc` does not exist", but I don't.
- In VSCode, I see the same file as displayed in Rill Developer. 
- In my terminal, I see the correct reconcile error.
- Refreshing the page in Rill Developer, I see the correct reconcile error.

image

image

image

AdityaHegde:
Completely clearing a dashboard (cmd/ctrl + a and delete) and adding back something will stop showing any successive errors. This is reset after a refresh or opening some other file and coming back.

I think this is because we remove the resource query. Even resetting/refetching after that doesnt bring it back. A proper fix will follow but this is a hotfix for the release.

Might fix #4918


nishantmonu51 (approved):

AdityaHegde:
Pull some temporary changes from #4883

PR #4924, opened 2024-05-17
Runtime: Component renderer for "template:"

begelundmuller:
- Adds support for a `template:` renderer in resources of `type: component`.
- Resources will have `spec.renderer == "template"` and `spec.rendererProperties` will contain the supplied properties.
- The template properties are validated using the JSON schema in `runtime/compilers/rillv1/data/component-template-v1.json`. It currently has validation stubs for `line_chart` and `bar_chart` templates.

Example:
```yaml
type: component

template:
  name: line_chart
  x: foo
  y: bar
```

### DIFF
--- a/runtime/compilers/rillv1/parse_component.go
+++ b/runtime/compilers/rillv1/parse_component.go
@@ -17,6 +17,11 @@ var vegaLiteSpec string
 
 var vegaLiteSchema = jsonschema.MustCompileString("https://vega.github.io/schema/vega-lite/v5.json", vegaLiteSpec)
 
+//go:embed data/component-template-v1.json
+var componentTemplateSpec string
+
+var componentTemplateSchema = jsonschema.MustCompileString("https://github.com/rilldata/rill/runtime/compilers/rillv1/data/component-template-v1.json", componentTemplateSpec)
+
 type ComponentYAML struct {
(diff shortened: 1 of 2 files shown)


runtime/compilers/rillv1/parse_component.go
```diff
@@ -17,6 +17,11 @@ var vegaLiteSpec string
 
 var vegaLiteSchema = jsonschema.MustCompileString("https://vega.github.io/schema/vega-lite/v5.json", vegaLiteSpec)
 
+//go:embed data/component-template-v1.json
+var componentTemplateSpec string
+
+var componentTemplateSchema = jsonschema.MustCompileString("https://github.com/rilldata/rill/runtime/compilers/rillv1/data/component-template-v1.json", componentTemplateSpec)
```
djbarnwal, line 23: Is there a way we can use a relative path instead of an URL. We might want to support Rill developer without network access as much as possible.

begelundmuller: It's confusing, but this doesn't actually make a network request. It's just JSON schema expecting a URL that identifies the particular schema (it seems to include it in some error messages). 

For example, we load the Vega spec like this a few lines above, but that also doesn't do a network request:
```go
var vegaLiteSchema = jsonschema.MustCompileString("https://vega.github.io/schema/vega-lite/v5.json", vegaLiteSpec)
```

djbarnwal (approved):

PR #386, opened 2022-06-10
Improve autocomplete and disable filesystem materialization

ericpgreen2:
This is an alternate option to #385. Both improve autocomplete. We should pick one of the two:
- #385: Enables filesystem sync and includes debouncing (editor feels ever-so-slightly sluggish).
- #386: Disables filesystem sync and removes debouncing (editor feels snappy).

PR #4932, opened 2024-05-20
chore: enrich scaling log

Kavinjsir:
### Description

This pull request refactors the log output for successful slot scaling operations to include the following fields:
- `project_name`
- `updated_slots`
- `prev_slots`
- `org_name`

### Motivation

This update addresses the needs outlined in [Issue #386](https://github.com/rilldata/rill-private-issues/issues/386).

### DIFF
--- a/admin/worker/run_autoscaler.go
+++ b/admin/worker/run_autoscaler.go
@@ -38,7 +38,13 @@ func (w *Worker) runAutoscaler(ctx context.Context) error {
 
 		targetProject, err := w.admin.DB.FindProject(ctx, rec.ProjectID)
 		if err != nil {
-			w.logger.Debug("failed to find project:", zap.String("project_id", rec.ProjectID), zap.Error(err))
+			w.logger.Debug("failed to find project", zap.String("project_id", rec.ProjectID), zap.Error(err))
+			continue
+		}
+
+		projectOrg, err := w.admin.DB.FindOrganization(ctx, targetProject.OrganizationID)
+		if err != nil {
+			w.logger.Error("failed to find org for the project", zap.String("project_name", targetProject.Name), zap.String("org_id", targetProject.OrganizationID), zap.Error(err))
 			continue
 		}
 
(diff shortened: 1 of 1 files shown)


admin/worker/run_autoscaler.go
```diff
@@ -72,7 +72,20 @@ func (w *Worker) runAutoscaler(ctx context.Context) error {
...
+			zap.Int("updated_slots", updatedProject.ProdSlots),
+			zap.Int("prev_slots", targetProject.ProdSlots),
+		}
+
+		updatedProjectOrg, err := w.admin.DB.FindOrganization(ctx, updatedProject.OrganizationID)
+		if err != nil {
+			w.logger.Error("failed to find org for the project", zap.String("project_name", updatedProject.Name), zap.String("org_id", updatedProject.OrganizationID), zap.Error(err))
+		} else {
+			logInfo = append(logInfo, zap.String("org_name", updatedProjectOrg.Name))
+		}
```
begelundmuller, line 86: Move finding of the org up to after the call to `FindProject`, before calling `UpdateProject`. That way you can just call `continue` if there's an `err`, and the code will be cleaner.

admin/worker/run_autoscaler.go
```diff
@@ -72,7 +72,20 @@ func (w *Worker) runAutoscaler(ctx context.Context) error {
...
+
+		w.logger.Info("succeeded in autoscaling:", logInfo...)
```
begelundmuller, line 88: nit: remove the trailing `:` in the message – this is structured logging, so formatting is already taken care of

begelundmuller (changes requested):

admin/worker/run_autoscaler.go
Kavinjsir: Sounds good!

admin/worker/run_autoscaler.go
Kavinjsir: Thx for letting me know 😃  Updated other places also correspondingly

Kavinjsir:
@begelundmuller Thx for reviewing with suggestions! Just updated code, can you help with another review :) 🏓

begelundmuller (approved):

PR #4938, opened 2024-05-21
Feat: Pivot tooltip functionality

briangregoryholmes:
This PR adds tooltip and click to copy functionality to the pivot table. To do so, it pulls the tooltip implementation in the Virtual table into its own component and reuses this in the PivotTable. It also pulls the modifiedClick action from this PR https://github.com/rilldata/rill/pull/4516.


djbarnwal (commented):
Couple of UX related feedback - 

1. It would be great if we can disable tooltips when the table is being scrolled. Otherwise, the tooltip for a particular value seems stuck for the duration of the scroll.

2. The "Copy" text in the tooltip in our other surfaces has an animation on Shift+Click (try copying a value from the leaderboard). Would be great if we can also have that here. This is not a blocker but nice to have.

djbarnwal (approved):

PR #4947, opened 2024-05-23
Refactor: Add content cache and mutation functions to FileArtifact class

briangregoryholmes:
Prereq/Subset of #4901:

- Moves the file mutation functions and local/remote content caches to the FileArtifact class
- Does not yet implement them in the application


briangregoryholmes:
Closing. Needs to be reviewed in context.

PR #4951, opened 2024-05-23
refactor(dashboard): move markdown to template

mindspank:
Removes the markdown and image parser.
Moves the markdown to template.
Adds css styling support to the markdown component.


web-common/src/features/custom-dashboards/Markdown.svelte
```diff
@@ -2,21 +2,25 @@
   import DOMPurify from "dompurify";
   import { marked } from "marked";
 
-  export let markdown: string;
-  export let fontSize: number;
+  export let markdown: string = "";
   export let editing = false;
+  export let css = {};
```
mindspank, line 7: We should probably have some dummy interface here I guess

PR #4952, opened 2024-05-23
OpenAI: Upgrade to GPT4o

ericpgreen2:
To be A/B tested against GPT3.5 Turbo before merging!

### DIFF
--- a/admin/ai/openai.go
+++ b/admin/ai/openai.go
@@ -34,7 +34,7 @@ func (c *openAI) Complete(ctx context.Context, msgs []*adminv1.CompletionMessage
 	}
 
 	res, err := c.client.CreateChatCompletion(ctx, openai.ChatCompletionRequest{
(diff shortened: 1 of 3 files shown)


admin/ai/openai.go
```diff
@@ -34,7 +34,7 @@ func (c *openAI) Complete(ctx context.Context, msgs []*adminv1.CompletionMessage
 	}
 
 	res, err := c.client.CreateChatCompletion(ctx, openai.ChatCompletionRequest{
-		Model:       openai.GPT3Dot5Turbo,
+		Model:       openai.GPT4o,
```
nishantmonu51, line 37: When using GPT4o in chats, I have noticed that its unavailable many times. I wonder if we should have a list of models for fallback ? i.e. If gpt4o isn't available fallback to 3.5 turbo ?
@begelundmuller : Any thoughts ?

nishantmonu51 (approved):

Issue #4944, opened 2024-05-23
Source Inspector is sometimes not showing reconciling spinner


ericpgreen2:
![image](https://github.com/rilldata/rill/assets/14206386/1a11e70d-d545-4b98-934f-3143601d5777)

PR #4953, opened 2024-05-24
fix: Updating errors for reconciling resources

### LINKED ISSUE #4943: Rill Developer: previous error remains during ingest
Here's a Source that errored:
image

I fixed the Source definition, then waited as the data was ingested. During ingest, the error remained. I'd expect to see a loading spinner.
image

AdityaHegde:
closes #4943 (potentially #4944 as well?)

This is a patch to make sure reconciling resources are updated and the reconciling spinner is shown. This might have regression where the left nav entry will flicker. We need to have a proper fix with debounce. We can do that after the patch release.

### DIFF
--- a/web-common/src/features/entity-management/file-artifacts.ts
+++ b/web-common/src/features/entity-management/file-artifacts.ts
@@ -95,15 +95,7 @@ export class FileArtifact {
         this.getResource(queryClient, instanceId),
       ],
       ([name, projectParser, resource]) => {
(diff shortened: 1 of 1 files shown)


ericpgreen2:
I'll approve & merge, but, yes, this introduces a flicker in the left nav. I see this does not fix #4944.

ericpgreen2 (approved):

PR #4956, opened 2024-05-24
Dedicated connectors

### LINKED ISSUE #4967: Dedicated connectors

esevastyanov:
Closes #4967


proto/rill/runtime/v1/resources.proto
```diff
@@ -499,3 +500,17 @@ message ExecutionError {
 message CharLocation {
   uint32 line = 1;
 }
+
+message ConnectorSpec {
+  string name = 1;
```
begelundmuller, line 505: It doesn't need a `name` field here because every resource already has a name (in `meta.name.name`)

runtime/compilers/rillv1/connectors_test.go
```diff
@@ -68,6 +73,12 @@ notify:
     - channel-test
     users:
     - user@example.com
+`,
+		// Dedicated S3 connector
+		"/connectors/s3-dedicated.yaml": `
+driver: s3
+name: s3-dedicated
```
begelundmuller, line 80: The name defaults to the filename without extensions (for all resources), so not necessary to set explicitly

runtime/registry.go
```diff
@@ -561,11 +561,25 @@ func (r *registryCache) updateProjectConfig(iwc *instanceWithController) error {
...
 		return err
 	}
-	dotEnv, err := rillv1.ParseDotEnv(iwc.ctx, repo, iwc.instanceID)
+
+	instance, err := r.get(iwc.instanceID)
+	if err != nil {
+		return err
+	}
+
+	p, err := rillv1.Parse(iwc.ctx, repo, iwc.instanceID, instance.Environment, instance.OLAPConnector)
```
begelundmuller, line 570: There's a call to `ParseRillYAML` right above that I think is also redundant now that it's anyway doing a full parse. Should be able to just check if `rill.yaml` is missing by accessing `p`.

begelundmuller (changes requested):

runtime/registry.go
esevastyanov: Yes, it is redundant. I also simplified a definition of `UpdateInstanceWithRillYAML` that is called next

runtime/compilers/rillv1/parse_connector.go
```diff
@@ -17,20 +17,19 @@ func (p *Parser) parseConnector(node *Node) error {
 		return err
 	}
 
+	name := tmp.Name
+	if name == "" {
+		name = node.Name
+	}
```
begelundmuller, line 23: Can be removed because it's already done in `parseStem`: https://github.com/rilldata/rill/blob/b46a80ad941905f99176c3d3c22d445a51472b16/runtime/compilers/rillv1/parse_node.go#L252

begelundmuller (approved):

PR #4961, opened 2024-05-25
Runtime: Advanced measures

begelundmuller:
This PR introduces support for advanced measures. To accomplish that, it introduces a completely refactored metrics query generation package.

**High-level changes:**

- Introduces three different measure types:
  - `simple`: regular aggregations against the underlying raw data.
  - `derived`: composed of other measures of type `simple`.
  - `time_comparison`: similar to `derived`, but supports referencing data from a comparison time range.
- Adds support for adding a `window` on any measure definition.
- Completely refactors metrics query generation. The new logic incrementally builds and optimizes an AST that is eventually traversed to generate the SQL query(s) to run against the DB.
- Moves metrics-related logic such as generating queries and validating measures to a new `runtime/metricsview` package.
- Adds a `metrics:` resolver for querying metrics (accessible in custom APIs, dashboard components, alerts, reports)
- Adds backwards compatibility that enables querying advanced measures with the `MetricsViewAggregation` and `MetricsViewTimeSeries` RPCs.
- Adds scaffolding for a `per` option enabling multi-level aggregation.
- Adds scaffolding for sub-queries in filters, enabling correct measure filters on dimensions not included in the present query.

**Documentation for advanced measures:**

This PR adds three new YAML properties for a measure definition:
- `measures[*].type` – allowed values:
  - `simple`: for a normal aggregation measure (e.g. `COUNT(*)`). This is the default if `requires` does not contain any measure names.
  - `derived`: for non-aggregation measures that reference other measures (e.g. dividing a measure by another). This is the default if `requires` contains one or more measure names.
  - `time_comparison`: similar to `derived`, but with access to values for both the base and comparison time range. The measure expression can use the table names `base` and `comparison` to reference the two values (e.g. `base.total_sales - comparison.total_sales`). This measure type only works if a comparison time range is passed at query time. 
- `measures[*].requires` – a list of dimension and measure names required when querying the measure. Any dimensions listed here must be included in API calls that query the measure. Any measure names added here will be accessible in the measure's `expression` (but need not be added separately when querying the measure).
- `measures[*].window` – allowed values:
  - `all` (string): adds a window that is neither partitioned nor ordered and without a frame. Equivalent to the SQL `OVER ()`.
  - `time` (string): adds a window partitioned by all non-time dimensions present when queried, and ordered by all time dimensions present when queried, and with a frame up to the current row. The metrics view's time dimension will implicitly be added to the measure's `requires` list. Equivalent to the SQL `OVER (PARTITION BY <non time dimensions> ORDER BY <time dimensions ascending> ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW`.
  - A nested object with three options:
    - `order:` list of dimension names to order by. Can optionally use nested objects with `name:` and `time_grain:` to limit to certain time granularities. Any dimensions listed here will implicitly be added to the measure's `requires` list.
    - `partition`: `false` or `true` (default). If `true`, the window will have a `PARTITION BY` clause that includes all non-time dimensions present in the current query.
    - `frame`: a literal SQL frame to apply to the window (e.g. `RANGE BETWEEN INTERVAL 7 DAYS PRECEDING AND CURRENT ROW`).

**Example advanced measure definitions:**

```yaml
measures:
  - name: total_cost
    expression: "SUM(cost)"

  - name: total_cost_7day_rolling_avg
    expression: AVG(total_cost)
    requires: [total_cost]
    window:
      order: __time
      frame: RANGE BETWEEN INTERVAL 7 DAYS PRECEDING AND CURRENT ROW

  - name: total_cost_delta
    expression: "SUM(cost)-LAG(SUM(cost))"
    window: time

  - name: total_cost_delta_alt
    expression: "total_cost-LAG(total_cost)"
    requires: total_cost
    window:
      order:
        - name: __time
           time_grain: day
    
  - name: total_cost_cumulative
    expression: SUM(total_cost)
    requires: [total_cost]
    window:
      order: __time

  - name: total_cost_pct_total
    expression: total_cost/sum(total_cost)
    window: all
    requires: [total_cost]

  - name: fancy
    type: time_comparison
    expression: 2 * (base.total_cost_pct_total - comparison.total_cost_pct_total)
    requires: [total_cost_pct_total]

  - name: very_fancy
    expression: fancy/sum(fancy)
    window:
      partition: false
    requires: [fancy]
```

**UI integration notes:**

- The measures in a `MetricsView` spec has a new property `required_dimensions`. This property lists dimensions that must be present in the `dimensions` parameter of a `MetricsViewAggregation` request when querying the measure.
- A required dimension consists of a `name` and optional `time_grain`. If the `time_grain` is present, the measure can only be queried if the request includes the dimension name truncated to that specific time grain.
- To ease support for rolling time series windows, this PR will also support querying advanced measures through the `MetricsViewTimeseries` RPC (as long as the measure's `required_dimensions` is empty or only contains the metrics view's `time_dimension`).

**Known issues**

The following known issues will be addressed in follow-up PRs:
- Advanced measures are not validated when the metrics view is parsed (will fail at query time instead)
- Pivoted exports containing advanced measures do not work
- Measures of `type: time_comparison` break the timeseries charts in the dashboard
- Comparison queries are not optimized for Druid
- Query generation has not been extensively tested against Druid or ClickHouse


begelundmuller:
The metrics can be queried using an API on `POST http://localhost:9009/v1/instances/default/api/metrics`. The API is created with the new resolvers and added as a built-in "custom API" so it will show up in auto-generated API docs when we add support for that.

The queries should be passed as a JSON body in the API request.

Example resolver queries:
```json
{
	"metrics_view": "customer_margin",
	"dimensions": [
		{
			"name": "company"
		},
		{
			"name": "time__day",
			"compute": {
				"time_floor": {
					"dimension": "__time",
					"grain": "day"
				}
			}
		}
	],
	"measures": [
		{
			"name": "total_cost"
		},
		{
			"name": "total_cost_delta"
		},
		{
			"name": "total_cost_delta_alt"
		},
		{
			"name": "total_cost_cumulative"
		},
		{
			"name": "total_cost_7day_rolling_avg"
		}
	],
	"time_range": {
		"iso_duration": "P2W"
	},
	"sort": [
		{
			"name": "company"
		},
		{
			"name": "time__day"
		}
	]
}
```

```json
{
	"metrics_view": "customer_margin",
	"dimensions": [
		{
			"name": "company"
		}
	],
	"measures": [
		{
			"name": "total_cost"
		},
		{
			"name": "total_cost_pct_total"
		},
		{
			"name": "very_fancy"
		}
	],
	"time_range": {
		"iso_duration": "P2W"
	},
	"comparison_time_range": {
		"iso_duration": "P2W",
		"iso_offset": "P2W"
	},
	"sort": [
		{
			"name": "total_cost",
			"desc": true
		}
	],
	"having": {
		"cond": {
			"op": "gte",
			"exprs": [{"name": "very_fancy"}, {"val": 1}]
		}
	}
}
```

PR #4980, opened 2024-05-29
fix: First source import modal opening before ingestion is completed

### LINKED ISSUE #4976: Rill Developer: "Source imported successfully" message shows before ingest completes
**To reproduce**
1. Start a new project from scratch
2. From the root page `http://localhost:9009/`, add a new data source. Here's a GCS data source that can take ~50s to ingest: gs://pkg.rilldata.com/example-data/sf311.parquet
3. Immediately see the "Source imported successfully" message. This message appears _before_ the source has completed ingest.
image

**Expected behavior**
- Submission of the Add Source modal should take me to the Source YAML page.
- During ingest, the Table Preview and the Source Inspector should both show an "Ingesting [source]" spinner.
- A non-blocking toast message notification should be displayed when the source completes ingest. During long-running ingest, a user should be able to proceed with other tasks and not be interrupted by a blocking CTA when the ingest completes asynchronously.

AdityaHegde:
closes #4976

The code checking for source ingestion was not looking for a successful creation of table. Updated the code to be specific to source and it now checks for either a table is created or an error is encountered.


web-common/src/features/sources/source-imported-utils.ts
```diff
@@ -13,12 +14,75 @@ export async function checkSourceImported(
...
+      instanceId,
+      filePath,
+    ).subscribe(({ done, errored }) => {
+      if (!done) return;
 
-  if (!success) return;
-  sourceImportedPath.set(filePath);
+      debouncer(() => {
+        end(!errored);
+      }, 500);
```
ericpgreen2, line 49: Why is the debounce necessary here?

web-common/src/features/sources/source-imported-utils.ts
```diff
@@ -13,12 +14,75 @@ export async function checkSourceImported(
...
   if (lastUpdatedOn) return; // For now only show for fresh sources
 
-  const success = await waitForResourceUpdate(
-    queryClient,
-    get(runtime).instanceId,
-    filePath,
-  );
+  waitForResourceUpdate(queryClient, get(runtime).instanceId, filePath)
+    .then((success) => {
+      console.log("s", success);
```
ericpgreen2, line 19: Leftover console log

web-common/src/features/sources/source-imported-utils.ts
```diff
@@ -13,12 +14,75 @@ export async function checkSourceImported(
...
+      if (!success) return;
+      sourceImportedPath.set(filePath);
```
ericpgreen2, line 21: (In the future) this could be a good use case for the Event Bus. Showing an overlay in response to an event is very analogous to showing a toast notification in response to an event. @briangregoryholmes has recently cleaned up the `NotificationCenter.svelte` – I'd recommend checking out that new code.

web-common/src/features/sources/source-imported-utils.ts
AdityaHegde: The source went into an IDLE state sometimes. With table check this is not be needed anymore.

web-common/src/features/sources/source-imported-utils.ts
AdityaHegde: Hmmm good point. Will do a follow up after the realse.

PR #4990, opened 2024-05-31
Connectors: Break DSN into constituent parts

ericpgreen2:
image


k-anshul:
I think we should also allow user to specify DSN since it allows user to set lot more connection parameters. One option could be to flip the screen with a button. See superset's db connector for reference:

image

image

k-anshul:
May be we can skip `SSL` parameter and let user specify schema in the hostname ?

nishantmonu51:
👍  to Anshul's comment about keeping DSN as and advanced property as well.

nishantmonu51 (approved):

PR #4992, opened 2024-05-31
Update instance connectors on a connector reconcile

esevastyanov:
The aim is to update project connectors on connector update


runtime/reconcilers/connector.go
```diff
@@ -59,6 +60,35 @@ func (r *ConnectorReconciler) Reconcile(ctx context.Context, n *runtimev1.Resour
...
+	parser, err := rillv1.Parse(ctx, repo, r.C.InstanceID, inst.Environment, inst.OLAPConnector)
+	if err != nil {
+		return runtime.ReconcileResult{Err: fmt.Errorf("failed to parse: %w", err)}
+	}
+
+	// Update instance connectors
+	err = r.C.Runtime.UpdateInstanceWithRillYAML(ctx, inst.ID, parser, false)
+	if err != nil {
+		return runtime.ReconcileResult{Err: fmt.Errorf("failed to update instance: %w", err)}
+	}
```
begelundmuller, line 90: The file watcher already parsed the file and created the `Connector` resource (returned from `self.GetConnector()` above). Doing a repo sync (clone or pull) and full project parse again here is redundant, and potentially quite slow if there are many connectors. 

Instead, you will probably want to add new functions, separate from `UpdateInstanceWithRillYAML`, for upserting/deleting connectors for an instance. They should also take care only to update in the DB (invalidating various caches) if the connector actually changed. (Since the reconciler can trigger even if there are no changes, e.g. on process restart.)

The delete case needs to be handled in the `if self.Meta.DeletedOn != nil {` case below.

begelundmuller (changes requested):

runtime/controller_test.go
```diff
@@ -1390,7 +1390,8 @@ name: my-gcs
 		},
 		Resource: &runtimev1.Resource_Connector{
 			Connector: &runtimev1.ConnectorV2{
-				Spec: &runtimev1.ConnectorSpec{Driver: "s3", Properties: map[string]string{"region": "us-west-2"}},
+				Spec:  &runtimev1.ConnectorSpec{Driver: "s3", Properties: map[string]string{"region": "us-west-2"}},
+				State: &runtimev1.ConnectorState{SpecHash: "76b3819c673b0a4e4755418614d8cb62"},
```
begelundmuller, line 1394: You can `nil` the field here to omit it from tests: https://github.com/rilldata/rill/blob/abb730b9e8843b7f912e7010c4edd3e42ace22f6/runtime/testruntime/reconcile.go#L155

runtime/reconcilers/connector.go
```diff
@@ -61,8 +64,68 @@ func (r *ConnectorReconciler) Reconcile(ctx context.Context, n *runtimev1.Resour
...
+
+	// Update instance connectors
+	err = r.C.Runtime.UpdateInstanceConnector(ctx, r.C.InstanceID, self.Meta.Name.Name, t.Spec)
+	if err != nil {
+		return runtime.ReconcileResult{Err: err}
+	}
+
+	if t.State == nil {
+		t.State = &runtimev1.ConnectorState{}
+	}
```
begelundmuller, line 92: If you update `func (r *ConnectorReconciler) ResetState(res *runtimev1.Resource) error {` to set an empty state, then this isn't needed. We do that in most other reconcilers, but actually looks like there's some inconsistency, so I understand the confusion. There might be some other code that assumes the state is not nil, so it's probably best to set it there.

begelundmuller (approved):
Have two minor comments, but overall this looks good!

PR #4995, opened 2024-05-31
Runtime : connector props from env

k-anshul:


proto/rill/runtime/v1/api.proto
```diff
@@ -251,6 +251,8 @@ message Connector {
   string name = 2;
   // Config for the connector
   map<string, string> config = 3;
+  // config_from_variables stores a copy of configs whose value is a variable. The key is the property and value is the variable. This is only set for configs obtained from `connector.yaml`
+  map<string, string> config_from_variables = 4;
```
k-anshul, line 255: This is not very great that we are storing a field for internal usage in API object. May be we can add another struct to store data in DB in future and remove this field from API struct.

runtime/compilers/rillv1/parse_connector.go
```diff
@@ -26,5 +28,26 @@ func (p *Parser) parseConnector(node *Node) error {
 
 	r.ConnectorSpec.Driver = tmp.Driver
 	r.ConnectorSpec.Properties = tmp.Defaults
```
begelundmuller, line 30: We should exclude entries that end up in `PropertiesFromVariables` from here

runtime/compilers/rillv1/parse_connector.go
```diff
@@ -26,5 +28,26 @@ func (p *Parser) parseConnector(node *Node) error {
...
+	r.ConnectorSpec.PropertiesFromVariables, err = propertiesFromVariables(tmp.Defaults)
+	if err != nil {
+		return err
+	}
```
begelundmuller, line 34: You need to do the parsing and handle the error earlier in this function call – before the part that says:
```
```

runtime/compilers/rillv1/parse_connector.go
```diff
@@ -26,5 +28,26 @@ func (p *Parser) parseConnector(node *Node) error {
...
+		if err != nil {
+			return nil, err
+		}
+		// assume that only one variable will be set
+		for _, k := range meta.Variables {
+			if after, found := strings.CutPrefix(k, "vars."); found {
+				res[key] = after
+			}
+		}
+	}
```
begelundmuller, line 51: Can we return an error if there are non-templated contents or multiple variables (without making the code too ugly)? It would be a very hard issue for the user to debug otherwise.

runtime/connections.go
```diff
@@ -199,11 +199,20 @@ func (r *Runtime) ConnectorConfig(ctx context.Context, instanceID, name string)
 		}
 	}
 
+	vars := inst.ResolveVariables()
 	// Search for connector definition in rill.yaml
 	for _, c := range inst.ProjectConnectors {
```
begelundmuller, line 204: nit: don't put a line of code above a comment that covers a whole section of code (put the code under the comment or add a newline above the comment)

begelundmuller (changes requested):

proto/rill/runtime/v1/api.proto
begelundmuller: Agree. Might make sense to leverage the new `ConnectorSpec` internally instead.

runtime/compilers/rillv1/parse_connector.go
k-anshul: The downstream code handles the logic that `PropertiesFromVariables` from `Properties` is a copy of `Properties` whose value is a variable. 
Will change it.

proto/rill/runtime/v1/api.proto
k-anshul: I can submit a follow up PR for this change. 
Also we need to introduce `name` field in `ConnectorSpec` to make it compatible with `connector` which will be somewhat unintuitive. Ideally spec is same for all names.

runtime/compilers/rillv1/parse_connector.go
```diff
@@ -25,6 +39,33 @@ func (p *Parser) parseConnector(node *Node) error {
...
+		if err != nil {
+			return nil, nil, err
+		}
+		if len(meta.Variables) == 0 { // property does not use any variables
+			props[key] = val
+			continue
+		}
+		// property uses variables
+		if len(meta.Variables) > 1 {
+			return nil, nil, fmt.Errorf("connector property should contain atmost one variable. Property %q contains %v", key, len(meta.Variables))
```
begelundmuller, line 61:
```suggestion
			return nil, nil, fmt.Errorf("invalid property %q: can contain at most one variable, found %d variables", key, len(meta.Variables))
```

runtime/compilers/rillv1/parse_connector.go
```diff
@@ -25,6 +39,33 @@ func (p *Parser) parseConnector(node *Node) error {
...
+		}
+		if !envVarRegex.MatchString(val) {
+			return nil, nil, fmt.Errorf("invalid property %q. When accessing a variable in a connector property, the value should match the form `{{ .vars.variable }}`", key)
```
begelundmuller, line 64:
```suggestion
			return nil, nil, fmt.Errorf(`invalid property %q: variable references should match the format "{{ .vars.variable_name }}"`, key)
```

begelundmuller (approved):

nishantmonu51 (approved):

PR #4997, opened 2024-05-31
pivot: resize column ux

briangregoryholmes:
Updates the UX for resizing columns in the pivot table by extending the handle vertically across the entire table.


briangregoryholmes:
@jkhwu

ericokuma:
Works well! 

Just a minor nit, it seems to not be horizontally aligned: 
Screenshot 2024-07-09 at 9 51 50 AM

djbarnwal:
The resizing is a bit buggy for me. After expanding or contracting the column width, it takes the default width and is not respecting the user defined width

https://github.com/user-attachments/assets/320abfc8-f442-4c2f-bb7c-e3a3df93a45a

briangregoryholmes:
This was being caused by the behavior and placement of the `removeActiveCell` function.

1. I've moved this active cell cancellation behavior to the `PivotToolbar` component, which is the only place it can practically be. Its use with the `clickOutside` action meant that it was being triggered on mouse up when dragging. Click outside does not really capture the intended cancellation behavior as clicking things like the panel toggle, export button or search bar would clear the active cell. It's now only cancelled when clicking the empty space above the pivot table (between the panel toggle and export buttons).
2. Toggling the active cell causes the entire config object to be re-derived, which in turn causes the table to re-render and the column widths to be reset. I've changed the implementation of the `measureLengths` so that this is avoided, but I think we should isolate the activeCell state so that this isn't happening in the background.

djbarnwal (approved):
Looks good! 

I still have some resizing bugs (mentioned in the video above) on firefox but resizing works smoothly on chrome.

Firefox support is probably not crucial but good to have. Will leave the choice to you if it's a small fix otherwise this should be good to be merged

PR #5003, opened 2024-06-03
fix(advanced measures): Exclude advanced measures from pivot and exclude comparison measures

AdityaHegde:
Removed certain advanced measures from pivot and alerts. Also updates the dashboard state with any incorrect selections. EG: when in a TDD with a measure depending on a specific time grain and time grain is changed, active measure is cleared.

Sample dashboard,
```
# Dashboard YAML

type: metrics_view
title: Adbids Model
model: AdBids_model
timeseries: timestamp
dimensions:
  - label: Publisher
    column: publisher
  - label: Domain
    column: domain

measures:
  - name: total_records
    label: Total records
    expression: COUNT(*)
    format_preset: humanize
    valid_percent_of_total: true

  - name: bid_price
    label: Sum of Bid Price
    expression: SUM(bid_price)
    format_preset: humanize
    valid_percent_of_total: true

  - name: bid_price_lag
    label: Bid Price Lag
    expression: "bid_price - LAG(bid_price)"
    window: true
    requires: [bid_price, timestamp]
    valid_percent_of_total: true

  - name: bid_price_percent_total
    label: Bid Price Percent of Total
    expression: bid_price/SUM(bid_price)
    window:
      partition: false
    requires: [bid_price]
    format_preset: percentage
    valid_percent_of_total: true

  - name: bid_price_percent_total_diff
    type: time_comparison
    expression: (base.bid_price_percent_total - comparison.bid_price_percent_total)
    requires: [bid_price_percent_total]

  - name: bid_price_7day_rolling_avg
    expression: AVG(bid_price)
    window:
      frame: RANGE BETWEEN INTERVAL 7 DAYS PRECEDING AND CURRENT ROW
    requires:
      - name: timestamp
        time_grain: day
      - name: bid_price
```

### DIFF
--- a/web-common/src/features/dashboards/state-managers/selectors/measures.ts
+++ b/web-common/src/features/dashboards/state-managers/selectors/measures.ts
@@ -52,26 +52,24 @@ export const isMeasureValidPercentOfTotal = ({
   };
 };
 
-export const filteredBasicMeasures = ({
+export const filteredSimpleMeasures = ({
   metricsSpecQueryResult,
-  dashboard,
 }: DashboardDataSources) => {
-  return (includeComparisonMeasure: boolean) => {
-    includeComparisonMeasure =
-      includeComparisonMeasure &&
-      !!dashboard.showTimeComparison &&
-      !!dashboard.selectedComparisonTimeRange;
+  return () => {
     return (
       metricsSpecQueryResult.data?.measures?.filter(
         (m) =>
           !m.window &&
-          (includeComparisonMeasure ||
-            m.type !== MetricsViewSpecMeasureType.MEASURE_TYPE_TIME_COMPARISON),
+          m.type !== MetricsViewSpecMeasureType.MEASURE_TYPE_TIME_COMPARISON,
       ) ?? []
     );
   };
 };
 
+/**
+ * Selects measure valid for current dashboard selections.
+ * Also includes additional dimensions needed for any advanced measures.
+ */
 export const getFilteredMeasuresAndDimensions = ({
   dashboard,
 }: Pick<DashboardDataSources, "dashboard">) => {
@@ -169,7 +167,7 @@ export const measureSelectors = {
    */
   isMeasureValidPercentOfTotal,
 
-  filteredBasicMeasures,
+  filteredSimpleMeasures,
 
   getFilteredMeasuresAndDimensions,
 };
--- a/web-common/src/features/dashboards/state-managers/selectors/measures.ts
+++ b/web-common/src/features/dashboards/state-managers/selectors/measures.ts
@@ -91,6 +91,11 @@ export const getFilteredMeasuresAndDimensions = ({
       if (
         !measure ||
         measure.type === MetricsViewSpecMeasureType.MEASURE_TYPE_TIME_COMPARISON
+        // TODO: we need to send a single query for this support
+        // (measure.type ===
+        //   MetricsViewSpecMeasureType.MEASURE_TYPE_TIME_COMPARISON &&
+        //   (!dashboard.showTimeComparison ||
+        //     !dashboard.selectedComparisonTimeRange))
       )
         return;
 
--- a/web-common/src/features/dashboards/state-managers/selectors/measures.ts
+++ b/web-common/src/features/dashboards/state-managers/selectors/measures.ts
@@ -1,5 +1,6 @@
 import {
   MetricsViewSpecDimensionSelector,
+  MetricsViewSpecMeasureType,
   MetricsViewSpecMeasureV2,
   V1MetricsViewSpec,
   V1TimeGrain,
@@ -51,11 +52,27 @@ export const isMeasureValidPercentOfTotal = ({
   };
 };
 
-export const filterBasicMeasures = (
-  measures: MetricsViewSpecMeasureV2[] | undefined,
-) => measures?.filter((m) => !m.window) ?? [];
+export const filteredBasicMeasures = ({
+  metricsSpecQueryResult,
+  dashboard,
+}: DashboardDataSources) => {
+  return (includeComparisonMeasure: boolean) => {
+    includeComparisonMeasure =
+      includeComparisonMeasure &&
+      !!dashboard.showTimeComparison &&
+      !!dashboard.selectedComparisonTimeRange;
+    return (
+      metricsSpecQueryResult.data?.measures?.filter(
+        (m) =>
+          !m.window &&
+          (includeComparisonMeasure ||
+            m.type !== MetricsViewSpecMeasureType.MEASURE_TYPE_TIME_COMPARISON),
+      ) ?? []
+    );
+  };
+};
 
-export const getMeasuresAndDimensions = ({
+export const getFilteredMeasuresAndDimensions = ({
   dashboard,
 }: Pick<DashboardDataSources, "dashboard">) => {
   return (
@@ -71,7 +88,11 @@ export const getMeasuresAndDimensions = ({
       const measure = metricsViewSpec.measures?.find(
         (m) => m.name === measureName,
       );
-      if (!measure) return;
+      if (
+        !measure ||
+        measure.type === MetricsViewSpecMeasureType.MEASURE_TYPE_TIME_COMPARISON
+      )
+        return;
 
       let skipMeasure = false;
       measure.requiredDimensions?.forEach((reqDim) => {
@@ -143,5 +164,7 @@ export const measureSelectors = {
    */
   isMeasureValidPercentOfTotal,
 
-  getMeasuresAndDimensions,
+  filteredBasicMeasures,
+
+  getFilteredMeasuresAndDimensions,
 };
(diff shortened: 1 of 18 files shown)


rill-dev:

web-common/src/features/dashboards/time-series/MeasureChart.svelte
```diff
@@ -76,7 +77,7 @@
   $: numberKind = numberKindForMeasure(measure);
 
   const tweenProps = { duration: 400, easing: cubicOut };
-  const xScale = getContext(contexts.scale("x")) as ScaleStore;
+  const xScale = getContext(contexts.scale("x"));
```
rill-dev, line 80: **Issue**: The getContext function is used without type assertion, which may lead to runtime errors if the context does not match the expected type. <br> **Fix**: Reinstate the type assertion for getContext to ensure type safety. <br> **Code Suggestion**: 
 ```
 -  const xScale = getContext(contexts.scale("x"));
 +  const xScale = getContext(contexts.scale("x")) as ScaleStore;
 ```

web-common/src/features/dashboards/stores/dashboard-stores.ts
```diff
@@ -451,10 +453,12 @@ const metricViewReducers = {
   setSelectedComparisonRange(
     name: string,
     comparisonTimeRange: DashboardTimeControls,
+    metricsViewSpec: V1MetricsViewSpec,
   ) {
     updateMetricsExplorerByName(name, (metricsExplorer) => {
       setDisplayComparison(metricsExplorer, true);
       metricsExplorer.selectedComparisonTimeRange = comparisonTimeRange;
+      AdvancedMeasureCorrector.correct(metricsExplorer, metricsViewSpec);
```
rill-dev, line 461: **Issue**: The method 'AdvancedMeasureCorrector.correct' is called without any null checks on 'metricsExplorer' or 'metricsViewSpec'. This could lead to a runtime error if either is null. <br> **Fix**: Add null checks before calling the 'AdvancedMeasureCorrector.correct' method to ensure stability. <br> **Code Suggestion**: 
 ```
 @@ -460,3 +460,5 @@
      metricsExplorer.selectedComparisonTimeRange = comparisonTimeRange;
 +    if (metricsExplorer && metricsViewSpec) {
 +      AdvancedMeasureCorrector.correct(metricsExplorer, metricsViewSpec);
      }
 -    AdvancedMeasureCorrector.correct(metricsExplorer, metricsViewSpec);
 ```

web-common/src/features/dashboards/stores/AdvancedMeasureCorrector.ts
```diff
@@ -0,0 +1,153 @@
...
+        return true;
+
+      // for measures with window operations, if the component doesnt support it
+      case !!measure.window && !supportsWindowedMeasure:
+        return true;
+
+      default:
+        return false;
+    }
+  }
```
rill-dev, line 152: **Issue**: The method 'measureMismatch' checks for various conditions to determine if there is a mismatch in the measure's requirements versus the dashboard's state. However, the method returns 'true' immediately if the measure is not found in the 'measuresMap', which could lead to unintended behavior where measures are incorrectly assumed to be mismatched if they are simply not present in the map. <br> **Fix**: Modify the 'measureMismatch' method to handle the case where a measure is not found more gracefully. This could involve logging a warning or error, or handling the case where the measure might be dynamically added later. <br> **Code Suggestion**: 
 ```
 @@ -122,1 +122,5 @@
 -    if (!measure) return true;
 +    if (!measure) {
 +      console.error('Measure not found: ${measureName}');
 +      return false; // or handle according to the application's needs
 +    }
 ```

web-common/src/features/dashboards/state-managers/selectors/measures.ts
```diff
@@ -51,11 +52,27 @@ export const isMeasureValidPercentOfTotal = ({
...
+    return (
+      metricsSpecQueryResult.data?.measures?.filter(
+        (m) =>
+          !m.window &&
+          (includeComparisonMeasure ||
+            m.type !== MetricsViewSpecMeasureType.MEASURE_TYPE_TIME_COMPARISON),
+      ) ?? []
+    );
+  };
+};
```
rill-dev, line 73: **Issue**: The function 'filteredBasicMeasures' introduces a significant performance issue by potentially executing the filter operation every time it is invoked with a different 'includeComparisonMeasure' value. This can lead to unnecessary computations, especially in a reactive Svelte context where it might be re-evaluated frequently due to state changes. <br> **Fix**: To optimize performance, refactor the function to memoize the result based on the 'includeComparisonMeasure' and 'metricsSpecQueryResult' parameters. This can be achieved by using a memoization library or implementing a simple caching mechanism within the function. <br> **Code Suggestion**: 
 ```
 @@ -55,3 +73,10 @@
 +  const cache = new Map();
 +  return (includeComparison: boolean) => {
 +    const cacheKey = '${includeComparison}';
 +    if (cache.has(cacheKey)) {
 +      return cache.get(cacheKey);
 +    }
 +    const result = (metricsSpecQueryResult.data?.measures?.filter(
 +      (m) => !m.window && (includeComparison || m.type !== MetricsViewSpecMeasureType.MEASURE_TYPE_TIME_COMPARISON),
 +    ) ?? []);
 +    cache.set(cacheKey, result);
 +    return result;
 +  };
 ```

web-common/src/features/dashboards/state-managers/selectors/pivot.ts
```diff
@@ -1,13 +1,14 @@
...
 import { PivotChipType } from "../../pivot/types";
 
 export const pivotSelectors = {
   showPivot: ({ dashboard }: DashboardDataSources) => dashboard.pivot.active,
   rows: ({ dashboard }: DashboardDataSources) => dashboard.pivot.rows,
   columns: ({ dashboard }: DashboardDataSources) => dashboard.pivot.columns,
-  measures: ({ metricsSpecQueryResult, dashboard }: DashboardDataSources) => {
-    const measures = metricsSpecQueryResult.data?.measures ?? [];
-    const columns = dashboard.pivot.columns;
+  measures: (dashData: DashboardDataSources) => {
```
rill-dev, line 9: **Issue**: The refactoring of the 'measures' selector function changes the function signature, which could lead to issues if other parts of the application rely on the original signature and expected behaviors. <br> **Fix**: Ensure that all references to the 'measures' selector across the application are updated to pass the correct type of argument, or revert to the original function signature to maintain compatibility. <br> **Code Suggestion**: 
 ```
 - measures: ({ metricsSpecQueryResult, dashboard }: DashboardDataSources) => {
 + measures: (dashData: DashboardDataSources) => {
 ```

web-common/src/features/dashboards/state-managers/selectors/measures.ts
```diff
@@ -51,11 +52,27 @@ export const isMeasureValidPercentOfTotal = ({
...
 };
 
-export const filterBasicMeasures = (
-  measures: MetricsViewSpecMeasureV2[] | undefined,
-) => measures?.filter((m) => !m.window) ?? [];
+export const filteredBasicMeasures = ({
+  metricsSpecQueryResult,
+  dashboard,
+}: DashboardDataSources) => {
+  return (includeComparisonMeasure: boolean) => {
```
ericpgreen2, line 59: 1. It's quite hard to understand this function where it's invoked like `filteredBasicMeasures(dashData)(false)`. To get rid of the second function call, maybe we can split it up into two functions – one for comparisons, one without?
2. Rather than introducing a new term "basic" measure, can we use existing terms, like "simple" measure or "windowless" measure?

web-common/src/features/dashboards/state-managers/selectors/measures.ts
```diff
@@ -51,11 +52,27 @@ export const isMeasureValidPercentOfTotal = ({
...
+          !m.window &&
+          (includeComparisonMeasure ||
+            m.type !== MetricsViewSpecMeasureType.MEASURE_TYPE_TIME_COMPARISON),
+      ) ?? []
+    );
+  };
+};
 
-export const getMeasuresAndDimensions = ({
+export const getFilteredMeasuresAndDimensions = ({
```
ericpgreen2, line 75: Can you add a comment at the top of this function to explain what the function filters for? One has to read the function implementation to understand what's going on, and it's quite complicated.

web-common/src/features/dashboards/stores/AdvancedMeasureCorrector.ts
```diff
@@ -0,0 +1,153 @@
...
+      (s) =>
+        !this.measuresMap.has(s.id) || !this.measureMismatch(s.id, true, false),
+    );
+  }
+
+  private measureMismatch(
+    measureName: string,
+    supportsComparisonMeasure: boolean,
+    supportsWindowedMeasure: boolean,
+  ) {
```
ericpgreen2, line 121: This function signature is hard to understand. Mind making the function name more descriptive? And maybe add a function-level comment?

ericpgreen2 (approved):

web-common/src/features/dashboards/state-managers/selectors/measures.ts
AdityaHegde: The cost of converting `metricsSpecQueryResult` into a key for the cache is higher than iterating through the measures.

web-common/src/features/dashboards/state-managers/selectors/measures.ts
AdityaHegde: Sounds good. Will just remove the boolean for now since it is always false.

PR #5008, opened 2024-06-03
Runtime: Use system tables and settings override

### LINKED ISSUE #5006: Runtime: Use `system.tables` instead of `information_schema.tables` for clickhouse OLAP queries
Clickhouse playground allows access to `system.tables` but blocks access to `information_schema.tables`. We can route our information_schema queries to `system.tables` instead to make our integration work with clickhouse playground. 

**Current Query**:
```
SELECT
	T.table_schema AS SCHEMA,
	T.table_schema = currentDatabase() AS is_default_schema,
	T.table_name AS NAME,
	T.table_type AS TABLE_TYPE, 
	C.column_name AS COLUMNS,
	C.data_type AS COLUMN_TYPE,
	C.ordinal_position as ORDINAL_POSITION
FROM information_schema.tables T 
JOIN information_schema.columns C ON T.table_schema = C.table_schema AND T.table_name = C.table_name
WHERE lower(T.table_schema) NOT IN ('information_schema', 'system')
ORDER BY SCHEMA, NAME, TABLE_TYPE, ORDINAL_POSITION
```

**Updated Query**:
```
SELECT 
    T.database AS SCHEMA,
    T.database = currentDatabase() AS is_default_schema,
    T.name AS NAME,
    if(lower(T.engine) like '%view%', 'VIEW', 'TABLE') AS TABLE_TYPE,
    C.name AS COLUMNS,
    C.type AS COLUMN_TYPE,
    C.position AS ORDINAL_POSITION
FROM system.tables AS T
INNER JOIN system.columns AS C ON (T.database = C.database) AND (T.name = C.table)
WHERE lower(T.database) NOT IN ('information_schema', 'system') 
ORDER BY
    SCHEMA ASC,
    NAME ASC,
    TABLE_TYPE ASC,
    ORDINAL_POSITION ASC
```

**NOTE** : 
1. It is possible that some other hosting or clickhouse playground in future blocks access to `system.tables` as well. Relevant configs : https://github.com/ClickHouse/ClickHouse/pull/38970
2. It is also possible that some users give access to `information_schema` but blocks access to `system`.
3. The type for both queries are not equivalent. The table types can have multiple values for `information_schema.tables` like `BASE TABLE`, `FOREIGN TABLE` etc. For simplicity I just kept `TABLE` and `VIEW` in new query.

k-anshul:
- closes #5006
- Adds a `settings_override` to set `readonly = 1` and disable other settings in case of read only user.

### DIFF
--- a/runtime/drivers/clickhouse/olap.go
+++ b/runtime/drivers/clickhouse/olap.go
@@ -107,9 +107,13 @@ func (c *connection) Execute(ctx context.Context, stmt *drivers.Statement) (res 
 		return nil, err
 	}
 
-	stmt.Query += "\n SETTINGS cast_keep_nullable = 1, join_use_nulls = 1, session_timezone = 'UTC'"
-	if c.config.EnableCache {
-		stmt.Query += ", use_query_cache = 1"
+	if c.config.SettingsOverride != "" {
+		stmt.Query += "\n SETTINGS " + c.config.SettingsOverride
+	} else {
+		stmt.Query += "\n SETTINGS cast_keep_nullable = 1, join_use_nulls = 1, session_timezone = 'UTC'"
+		if c.config.EnableCache {
+			stmt.Query += ", use_query_cache = 1"
+		}
 	}
 
 	// Gather metrics only for actual queries
--- a/runtime/drivers/clickhouse/clickhouse.go
+++ b/runtime/drivers/clickhouse/clickhouse.go
@@ -93,6 +93,8 @@ type configProperties struct {
 	EnableCache bool `mapstructure:"enable_cache"`
 	// LogQueries controls whether to log the raw SQL passed to OLAP.Execute.
 	LogQueries bool `mapstructure:"log_queries"`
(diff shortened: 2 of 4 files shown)


runtime/drivers/clickhouse/olap.go
```diff
@@ -107,9 +107,13 @@ func (c *connection) Execute(ctx context.Context, stmt *drivers.Statement) (res
 		return nil, err
 	}
 
-	stmt.Query += "\n SETTINGS cast_keep_nullable = 1, join_use_nulls = 1, session_timezone = 'UTC'"
-	if c.config.EnableCache {
-		stmt.Query += ", use_query_cache = 1"
+	if c.config.SettingsOverride != "" {
+		stmt.Query += "\n SETTINGS " + c.config.SettingsOverride
```
nishantmonu51, line 111: Do we need a way to mention achieve - do not override any settings i.e. No settings to be configured ?

nishantmonu51 (approved):

PR #5015, opened 2024-06-04
Connectors: Handle arbitrary names

ericpgreen2:
This PR is a small refactor to handle arbitrary connector names in the UI. This sets us up to allow >1 connector per connector type/driver.

For illustration, here's a connector not named the same as its driver:

image

### DIFF
--- a/web-local/src/routes/(application)/connector/clickhouse/[name]/[database]/[table]/+page.svelte
+++ b/web-local/src/routes/(application)/connector/clickhouse/[name]/[database]/[table]/+page.svelte
@@ -7,8 +7,10 @@
 
   const { readOnly } = featureFlags;
 
+  $: name = $page.params.name;
   $: database = $page.params.database;
-  $: databaseSchema = $page.params.schema;
+  // ClickHouse does not have a database "schema" concept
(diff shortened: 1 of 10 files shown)


web-local/src/routes/(application)/connector/clickhouse/[name]/[database]/[table]/+page.svelte
```diff
@@ -23,8 +24,4 @@
   <title>Rill Developer | {table}</title>
 </svelte:head>
 
-<TablePreviewWorkspace
```
Comment (AdityaHegde) [inline web-local/src/routes/(application)/connector/clickhouse/[name]/[database]/[table]/+page.svelte:26]: Any reason to keep individual pages for different connectors?

AdityaHegde (approved):
Just a small question

web-local/src/routes/(application)/connector/clickhouse/[name]/[database]/[table]/+page.svelte
Comment (ericpgreen2) [reply web-local/src/routes/(application)/connector/clickhouse/[name]/[database]/[table]/+page.svelte:26]: I originally did it because each database has its own specific hierarchy (see below), but actually we might be able to use a `[...catchall]` route. I'll try that after we get the release out.

image

PR #5016, opened 2024-06-04
refactor: project initialization check

### LINKED ISSUE #5013: Avoid flash when starting a new project and redirecting to Welcome page
When starting a fresh project, the application briefly opens `http://localhost:9009/`, assesses whether or not the project directory includes a `rill.yaml` file, then routes to `http://localhost:9009/welcome`. We should not flash `http://localhost:9009/`.

https://github.com/rilldata/rill/assets/14206386/bb4d5aa8-68aa-4a97-ab12-0a745264e421

briangregoryholmes:
This PR reworks the `isProjectInitialized` check to prevent unnecessary page flashing and, hopefully, simplify the process overall.

Essentially, it consolidates the responsibility for reroutes to the top level layout load function. When the application first loads, it calls `isProjectInitialized` and will reroute or create a new project depending on what page the user is on.

Once the application has started, it uses the `WatchFilesClient` to invalidate this load function if the `rill.yaml` file is deleted or added.

Components that previously used the `useIsProjectInitialized` query store now just call the same `isProjectInitialized` function asynchronously.

Because we're using the url parameter in the load function, SvelteKit will automatically re-run this function whenever the url changes. We don't want this, but don't have the ability to `untrack` things until we update to SvelteKit 2.0. As such, this PR includes a `firstLoad` store that allows us to control this behavior. After upgrading, this can simply be removed. (Note:  This PR initially took a different approach to avoid this workaround, but I now think the temporary workaround is worth it.)

It also creates a new [`hooks.client.ts`](https://kit.svelte.dev/docs/hooks) file to capture things that are needed to initialize the application, but may not necessarily be appropriate for a load function.

Closes #5013


web-local/src/hooks.client.ts
ericpgreen2, line 1: I'm not sure if it's an issue, but I see this warning when building the application: 
image

web-local/src/hooks.client.ts
```diff
@@ -0,0 +1,20 @@
...
+const INSTANCE_ID = "default";
+
+const runtimeInit = {
+  host: HOST,
+  instanceId: INSTANCE_ID,
+};
+
+runtime.set(runtimeInit);
+
+projectInitialized.init().catch(console.error);
```
ericpgreen2, line 20: This (maybe counter-intuitively) leads to the wrong behavior. Run `./rill start new-project` to see. We want a new project to open to the `/welcome` page.

ericpgreen2 (changes requested):

web-local/src/hooks.client.ts
briangregoryholmes: This was somewhat unrelated. The list files query returns a 400 response if there's no project folder and our `fetchWrapper` rejects the promise when that happens. Wrapping the query in a try catch block solved the issue, though I would argue that we should change the `fetchWrapper` behavior since this is kind of in conflict with Svelte Query's stateful error approach.

Issue #4893, opened 2024-05-14
Runtime: Support evaluating measure filters with subqueries


begelundmuller:
https://www.notion.so/rilldata/Measure-Filters-Beyond-250-Values-22d2f3558ce444358cf1d4fa742bbc86?pvs=4

PR #5023, opened 2024-06-05
Runtime: Support subqueries in "where" and "having" expressions

begelundmuller:
Supports subqueries with limited capabilities (e.g. can only return one dimension) in metrics query expressions.

A current limitation is that it will never match for `NULL` values (due to SQL semantics around `NULL` not being equal to anything).

Closes https://github.com/rilldata/rill/issues/4893


AdityaHegde (approved):
Would be good to add some tests for this, but not a blocker.

PR #5037, opened 2024-06-06
cd: enable add chart button

briangregoryholmes:
This PR enables the `Create new chart` option in the `Add Chart` dropdown present in the Custom Dashboard workspace. It also temporarily removes the `Generate chart with AI` option as the sequencing of that item is still being determined.


web-common/src/features/custom-dashboards/AddChartMenu.svelte
```diff
@@ -23,6 +25,16 @@
   );
   $: chartFileNames =
     $chartsQuery.data?.map((c) => c.meta?.name?.name ?? "") ?? [];
+
+  async function handleAddChart() {
+    const newRoute = await handleEntityCreate(ResourceKind.Component);
+
+    const chartName = newRoute?.split("/").pop()?.split(".")[0];
```
djbarnwal, line 32: Probably better to use `getNameFromFile` util function here?

djbarnwal (approved):

PR #5038, opened 2024-06-06
Pivot and export support for advanced measures

begelundmuller:
- Support building `PIVOT` queries in the new `runtime/metricsview` package
- Support exports in the `PIVOT` package, using native DuckDB exports when possible
- Rewrites exports to use `drivers.ModelExecutor`, simulating an `olap-to-file` model. This has several advantages:
  - Enables implementing export logic in the drivers
  - Drivers can provide custom export overrides, e.g. DuckDB uses `COPY TO` for faster exports
  - Enables re-using export logic for models that `output:` to a file
  - Makes it easier to support exports to an object store in the future
- Retrofits the new exports support into `MetricsViewAggregation`


runtime/pkg/jsonval/jsonval.go
```diff
@@ -0,0 +1,382 @@
...
+	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
+)
+
+// ToString is similar to ToValue, but finally converts the value to a JSON string.
+func ToString(v any, t *runtimev1.Type) (string, error) {
+	v, err := ToValue(v, t)
+	if err != nil {
+		return "", err
+	}
+	if v == nil {
```
k-anshul, line 24: Is it fine to return an empty string when v is nil in all cases ?

runtime/pkg/jsonval/jsonval.go
```diff
@@ -0,0 +1,382 @@
...
+			return uint64(*v), nil
+		}
+		return nil, nil
+	case *uint32:
+		if v != nil {
+			return uint64(*v), nil
+		}
+		return nil, nil
+	case *uint64:
+		if v != nil {
```
k-anshul, line 182: I think we should serialize it as string instead of uint64 to avoid loss of precision. Some information here : https://groups.google.com/g/protobuf/c/4-BY-k-Lk-g

k-anshul (changes requested):

runtime/pkg/jsonval/jsonval.go
begelundmuller: "All cases" are a little hard to think through here, but it was needed for the current case – serializing the value to a CSV record. It would be incorrect to output `nil` since that could be misinterpreted as a real string value.

I think you're right this doesn't seem generic. Since it's only currently used for CSV serialization, I removed this function and inlined it in the CSV serializer instead.

runtime/pkg/jsonval/jsonval.go
begelundmuller: Yeah that would be nice, but the frontend can't handle numbers as strings yet. That's also why we have this for all the numeric types:
```go
// Evil cast to float until frontend can deal with bigs:
```

k-anshul: Does UI parse the CSV/JSON exports ?
I thought this is only for CSV/JSON exports.

k-anshul (approved):

runtime/pkg/jsonval/jsonval.go
begelundmuller: Regarding frontend – this will also be integrated for the resolvers and custom APIs, so will eventually also be exposed there.

PR #5046, opened 2024-06-10
Add instance annotations to data_dir_size_bytes metric

esevastyanov:


runtime/registry.go
```diff
@@ -537,10 +537,16 @@ func (r *registryCache) emitHeartbeats() {
...
+	attrs := make([]attribute.KeyValue, 0, len(inst.Annotations)+2)
+	attrs = append(attrs,
 		attribute.String("instance_id", inst.ID),
-		attribute.String("updated_on", inst.UpdatedOn.Format(time.RFC3339)),
-	)
+		attribute.String("updated_on", inst.UpdatedOn.Format(time.RFC3339)))
+	// Add instance annotations as attributes to pass organization id, project id, etc.
+	for k, v := range inst.Annotations {
+		attrs = append(attrs, attribute.String(k, v))
+	}
```
begelundmuller, line 547: Consider using `instanceAnnotationsToAttribs` to avoid duplicated logic

runtime/registry.go
```diff
@@ -537,10 +537,16 @@ func (r *registryCache) emitHeartbeats() {
 func (r *registryCache) emitHeartbeatForInstance(inst *drivers.Instance) {
 	dataDir := filepath.Join(r.rt.opts.DataDir, inst.ID)
 
-	r.activity.RecordMetric(context.Background(), "data_dir_size_bytes", float64(sizeOfDir(dataDir)),
+	attrs := make([]attribute.KeyValue, 0, len(inst.Annotations)+2)
+	attrs = append(attrs,
 		attribute.String("instance_id", inst.ID),
-		attribute.String("updated_on", inst.UpdatedOn.Format(time.RFC3339)),
-	)
+		attribute.String("updated_on", inst.UpdatedOn.Format(time.RFC3339)))
```
begelundmuller, line 543: Let's omit this – it made sense for the wide events, but not really for metrics events (would be a separate metric if anything)

begelundmuller (changes requested):

begelundmuller (approved):

PR #5049, opened 2024-06-10
Connectors: various UX cleanups

ericpgreen2:
- When there are no connectors in a project, adds a light "Add data to get started" CTA
- Masks secrets (e.g. a ClickHouse `password`) in form inputs
- Removes quotes around numbers (e.g. a ClickHouse `port`) in generated Connector YAML files
- Cleans-up the ClickHouse types displayed in the Table Schema
- Autogenerates models with sufficiently qualified table names (i.e. specifies when a database or database schema are not the default value)
- Adds `{{ if dev }} limit 100000 {{ end }}` to models generated from tables


briangregoryholmes:
This looks good in general, though I think it's best if we resolve some Input related tech debt and provide the show/toggle UX now that it's being used for passwords. I've started that here: https://github.com/rilldata/rill/pull/5059

briangregoryholmes (approved):

PR #5050, opened 2024-06-10
cloud billing customer creation and reporting

pjain1:
This includes - Creating a customer in external billing system on new org creation and assigning a billing plan to it. Plan metadata configured in the billing system should have quotas that will be fetched to be enforced as well reportable metric names. These metrics name will directly corresponds to the apis in the metrics project to fetch the corresponding usage. Usage reporter will use the apis to report usage to the billing system.

Next -
- [ ] Enforce usage limits especially managed data bytes, will go into reconciler
- [x] Create metrics APIs in metrics project that will be used to fetch usage
- [x] APIs for UI to get list of public plans, current subscriptions for a customer and change plan for a customer
- [x] Rill cli cmds for the same
- [x] Handling of trail period and subscription end in web-admin, APIs for this, will need product input


admin/worker/billing_reporter.go
```diff
@@ -0,0 +1,247 @@
+package worker
```
begelundmuller, line 1: I think it's worth going over this file once more with an aim of simplifying the code where possible. There are some places with pretty deep nesting and if/else scenarios.

Also, have you considered syncing events in bulk for all orgs/projecst in one query (going over all events in the bucket) instead of querying and uploading per org/project? It seems the Orb APIs would support that and could reduce the latency and number of queries by a lot.

begelundmuller (changes requested):

admin/database/database.go
```diff
@@ -238,6 +239,9 @@ type Organization struct {
 	QuotaSlotsTotal         int       `db:"quota_slots_total"`
 	QuotaSlotsPerDeployment int       `db:"quota_slots_per_deployment"`
 	QuotaOutstandingInvites int       `db:"quota_outstanding_invites"`
+	QuotaNumUsers           int       `db:"quota_num_users"`
```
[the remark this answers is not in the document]
pjain1: It was part of rill cloud billing notion doc but not in mocks lets just remove it

admin/database/database.go
```diff
@@ -260,32 +267,36 @@ type UpdateOrganizationOptions struct {
...
+	ProdVariables          map[string]string `db:"prod_variables"`
+	ProdOLAPDriver         string            `db:"prod_olap_driver"`
+	ProdOLAPDSN            string            `db:"prod_olap_dsn"`
+	ProdSlots              int               `db:"prod_slots"`
+	ProdTTLSeconds         *int64            `db:"prod_ttl_seconds"`
+	ProdDeploymentID       *string           `db:"prod_deployment_id"`
+	Annotations            map[string]string `db:"annotations"`
+	CreatedOn              time.Time         `db:"created_on"`
+	UpdatedOn              time.Time         `db:"updated_on"`
+	NextUsageReportingTime time.Time         `db:"next_usage_reporting_time"`
```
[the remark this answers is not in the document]
pjain1: `UsageReportedOn` seems like the time on which last usage was reported but In the code this is used as the reporting start time for next reporting window. This window may be bigger than one delta as reporting may be delayed because of data unavailability or other number of reasons. Do you still think this makes sense, so I will make the change

admin/server/organizations.go
```diff
@@ -844,6 +876,122 @@ func (s *Server) SudoUpdateOrganizationQuotas(ctx context.Context, req *adminv1.
...
+					}
+					cancelDate = req.SubscriptionChangeDate.AsTime()
+				default:
+					return nil, status.Error(codes.InvalidArgument, "invalid subscription change effective")
+				}
+			}
+			err = s.admin.Biller.CancelSubscriptionsForCustomer(ctx, org.BillingCustomerID, cancelOption, cancelDate)
+			if err != nil {
+				return nil, status.Error(codes.Internal, err.Error())
+			}
```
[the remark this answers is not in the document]
pjain1: Yeah I see there is an api to schedule plan change, will use that. Thanks

admin/server/organizations.go
```diff
@@ -844,6 +876,122 @@ func (s *Server) SudoUpdateOrganizationQuotas(ctx context.Context, req *adminv1.
 	}, nil
 }
 
+func (s *Server) SudoUpdateOrganizationBilling(ctx context.Context, req *adminv1.SudoUpdateOrganizationBillingRequest) (*adminv1.SudoUpdateOrganizationBillingResponse, error) {
```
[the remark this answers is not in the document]
pjain1: I have split this into two APIs - one sudo api for just assigning a customer id to the rill org, will be useful in case customer is already created or manually created in Orb. Second is for self serve plan change, this will be used for upgrading plans by customers, currently there is no check but that would require having a billing method.

proto/rill/admin/v1/api.proto
```diff
@@ -557,6 +565,8 @@ message GetOrganizationResponse {
 message CreateOrganizationRequest {
   string name = 2 [(validate.rules).string.min_len = 1];
   string description = 3;
+  optional string rill_plan = 4;
+  optional string biller_plan = 5;
 }
```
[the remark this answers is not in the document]
pjain1: makes sense, removed plan from org creation request

admin/billing/biller.go
```diff
@@ -0,0 +1,99 @@
...
+	// GetPublicPlans for listing purposes
+	GetPublicPlans(ctx context.Context) ([]*Plan, error)
+	// GetPlan returns the plan with the given Rill plan ID or biller plan ID.
+	GetPlan(ctx context.Context, rillPlanID string, billerPlanID string) (*Plan, error)
+
+	// CreateCustomer creates a customer for the given organization in the billing system and returns the external customer ID.
+	CreateCustomer(ctx context.Context, organization *database.Organization) (string, error)
+
+	// CreateSubscription creates a subscription for the given organization.
+	// The subscription starts immediately. TODO - support starting at a future date.
```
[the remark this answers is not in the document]
pjain1: removed this support, Orb supports it but I guess we don't have a use for it now

admin/billing/orb.go
```diff
@@ -0,0 +1,378 @@
...
+}
+
+func (o *Orb) ReportUsage(ctx context.Context, customerID string, usage []*Usage) error {
+	var orbUsage []orb.EventIngestParamsEvent
+	for _, u := range usage {
+		eventName := u.MetricName + "_" + string(u.ReportingGran)
+		// use end time minus 1 second to make sure the event is attributed to the current time bucket
+		eventTime := u.EndTime.Add(-1 * time.Second)
+		// generate idempotency key using customer id, timestamp, event name and metadata
+		idempotencyKey := fmt.Sprintf("%s_%d_%s_%v", customerID, eventTime.UnixMilli(), eventName, u.Metadata)
```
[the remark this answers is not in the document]
pjain1: but the final usage event might not have 1:1 mapping with event_id because the api may may do some aggregation, key is used to make sure duplicate usage is not reported for an interval.

admin/billing/orb.go
```diff
@@ -0,0 +1,378 @@
...
+		}
+	}
+	return nil, nil
+}
+
+func (o *Orb) CreateCustomer(ctx context.Context, organization *database.Organization) (string, error) {
+	customer, err := o.client.Customers.New(ctx, orb.CustomerNewParams{
+		Email:              orb.String(SupportEmail),
+		Name:               orb.String(organization.Name),
+		ExternalCustomerID: orb.String(organization.Name),
```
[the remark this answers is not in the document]
pjain1: using org Id now

admin/worker/billing_reporter.go
```diff
@@ -0,0 +1,247 @@
...
+		// add grace period buffer to lower and upper bound to ensure correct comparison
+		validationLowerTimeBound = validationLowerTimeBound.Add(-validationGracePeriod)
+		validationUpperTimeBound := workerTimeBucketEnd.Add(validationGracePeriod)
+
+		// Get availability
+		availability, err := client.GetProjectUsageAvailability(ctx, org.Name, reportableProjects, validationLowerTimeBound, validationUpperTimeBound)
+		if err != nil {
+			w.logger.Error("failed to report usage: unable to get availability", zap.String("org", org.Name), zap.Error(err))
+			continue
+		}
```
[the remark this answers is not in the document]
pjain1: using id now

admin/server/organizations.go
```diff
@@ -123,7 +132,18 @@ func (s *Server) CreateOrganization(ctx context.Context, req *adminv1.CreateOrga
 		return nil, status.Errorf(codes.FailedPrecondition, "quota exceeded: you can only create %d single-user orgs", user.QuotaSingleuserOrgs)
 	}
 
-	org, err := s.admin.CreateOrganizationForUser(ctx, user.ID, req.Name, req.Description)
+	var plan *billing.Plan
+	if req.RillPlan != nil || req.BillerPlan != nil {
+		rillPlan := valOrDefault(req.RillPlan, "")
+		billerPlan := valOrDefault(req.BillerPlan, "")
+		// support assigning plan by Rill ID or biller ID
+		plan, err = s.admin.Biller.GetPlan(ctx, rillPlan, billerPlan)
```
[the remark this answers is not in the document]
pjain1: We can have a mapping in client or have a postgres table for this, will just need to be kept in sync with Orb changes, also we don't have plans defined in Orb yet. Having postgres table looks more convenient than changing code to keep things in sync, if we want to do that.

admin/worker/billing_reporter.go
pjain1: Yeah its pretty busy. Tried simplifying a bit by breaking logic into functions to handle org and then its projects. Also changed reporting logic to first collect usage for all projects in an org and reporting all at once. Two more places to look into -
- Reporting usage to Orb for all orgs, not sure if we would hit issues with big payload 
- Usage availability is fetched for all projects of an org in one go from rill-cloud-metrics but fetching usage is still done per project per bucket per metric. Each project might have different buckets to report and metrics to report may change as per plan so creating a single api to fetch usage per plan can be hard to keep in sync.

admin/server/organizations.go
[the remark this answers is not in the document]
pjain1: Or I can just fetch all the plans and match against the external Id set in Orb.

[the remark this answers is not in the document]
pjain1: Another issue is that existing plans do not have an external id defined so if we want to use those and keep a single id here then we will have to use Orb's defined random alphanumeric ids. Or we just choose to not use them and create all plan afresh in Orb with an external meaningful id

admin/database/postgres/migrations/0030.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE orgs ADD COLUMN billing_customer_id TEXT NOT NULL DEFAULT '';
+
+CREATE UNIQUE INDEX orgs_billing_customer_id_idx ON orgs (billing_customer_id) WHERE billing_customer_id <> '';
+
+ALTER TABLE orgs ADD COLUMN quota_storage_limit_bytes_per_deployment BIGINT NOT NULL DEFAULT -1;
+UPDATE orgs SET quota_storage_limit_bytes_per_deployment = 5368709120;
+
+ALTER TABLE projects ADD COLUMN next_usage_reporting_time TIMESTAMP DEFAULT '0001-01-01 00:00:00';
```
begelundmuller, line 8: Can we have the default be `null` instead?

admin/metrics/client.go
```diff
@@ -98,3 +98,158 @@ func (c *Client) AutoscalerSlotsRecommendations(ctx context.Context, limit, offs
...
+
+	uri, err := url.Parse(runtimeHost)
+	if err != nil {
+		return nil, err
+	}
+
+	var usage []Usage
+	for _, metric := range metricNames {
+		// metric name is the api name
+		uri.Path = path.Join("/v1/instances", c.InstanceID, fmt.Sprintf("/api/%s", metric+"-usage"))
```
begelundmuller, line 203: Can we refactor to have a 1:1 mapping between function calls and network requests? Makes it easier to think about concurrency, and the loop can be easily handled at the caller

admin/server/organizations.go
```diff
@@ -200,6 +212,204 @@ func (s *Server) UpdateOrganization(ctx context.Context, req *adminv1.UpdateOrga
...
+
+	org, err = s.admin.DB.UpdateOrganization(ctx, org.ID, &database.UpdateOrganizationOptions{
+		Name:                                org.Name,
+		Description:                         org.Description,
+		QuotaProjects:                       quotaProjects,
+		QuotaDeployments:                    quotaDeployments,
+		QuotaSlotsTotal:                     quotaSlotsTotal,
+		QuotaSlotsPerDeployment:             quotaSlotsPerDeployment,
+		QuotaOutstandingInvites:             quotaOutstandingInvites,
+		QuotaStorageLimitBytesPerDeployment: quotaStorageLimitBytesPerDeployment,
```
begelundmuller, line 318: Can be inlined without declaring separate variables

admin/user.go
```diff
@@ -211,7 +219,72 @@ func (s *Service) CreateOrganizationForUser(ctx context.Context, userID, orgName
 
 	s.Logger.Info("created org", zap.String("name", orgName), zap.String("user_id", userID))
 
-	return org, nil
+	// return early if billing is not configured
+	if s.Biller.Name() == "noop" {
+		return org, nil
+	}
+
+	var updatedOrg *database.Organization
```
begelundmuller, line 227: Unnecessary declaration

admin/user.go
```diff
@@ -211,7 +219,72 @@ func (s *Service) CreateOrganizationForUser(ctx context.Context, userID, orgName
 
 	s.Logger.Info("created org", zap.String("name", orgName), zap.String("user_id", userID))
 
-	return org, nil
+	// return early if billing is not configured
+	if s.Biller.Name() == "noop" {
+		return org, nil
+	}
```
begelundmuller, line 225: Not really useful to have a `noop` biller if it needs to be manually checked. Might as well remove it then and have `s.Biller == nil` checks instead.

Consider if the interface or calls can be changes such that the no-op logger accepts the same calls as the real logger. It would make test and prod code flows more similar. For example, maybe `CreateCustomer` should be refactored to automatically create a subscription with the default plan?

admin/worker/billing_reporter.go
begelundmuller: > Each project might have different buckets to report and metrics to report may change as per plan so creating a single api to fetch usage per plan can be hard to keep in sync.

My hope would be that we could simplify usage reporting a lot by not trying to do such a granular breakdown. It should also lead to fewer, more efficient queries to the DB. Unless Orb charges a lot of money for redundant metrics, it seems simpler to just ship all non-zero billable metrics to them and have the logic of which metrics to use implemented there.

Of course we would probably need to batch the uploads to Orb to not upload huge payloads.

I'm imagining a query like this on our side (pseudo-code):
```sql
SELECT instance_id, <date trunc> as event_time, metric_name, <agg> as value
FROM metrics
WHERE event_time > <from time> AND event_time < <to time>
GROUP BY ALL
ORDER BY 
```

Then in the worker, it just does a lookup and caches each `instance_id`, enriches the events with info about the instance's org, and does a bulk upload to Orb of e.g. every 1k events.

admin/billing/biller.go
```diff
@@ -0,0 +1,101 @@
...
+
+import (
+	"context"
+	"time"
+
+	"github.com/rilldata/rill/admin/database"
+)
+
+const (
+	DefaultPlanID   = "Starter"
```
begelundmuller, line 11: This looks more like a name than an ID?

admin/billing/biller.go
```diff
@@ -0,0 +1,101 @@
...
+)
+
+type Biller interface {
+	Name() string
+	GetDefaultPlan(ctx context.Context) (*Plan, error)
+	GetPlans(ctx context.Context) ([]*Plan, error)
+	// GetPublicPlans for listing purposes
+	GetPublicPlans(ctx context.Context) ([]*Plan, error)
+	// GetPlan returns the plan with the given Rill plan ID or biller plan ID.
+	GetPlan(ctx context.Context, rillPlanID string, billerPlanID string) (*Plan, error)
```
begelundmuller, line 23: If it's necessary to do lookups by both IDs, this should be broken into two functions – e.g. similar to how `database` has `FindOrganization(ctx, id)` and `FindOrganizationByName(ctx, name)`

admin/billing/biller.go
```diff
@@ -0,0 +1,101 @@
...
+	ReportUsage(ctx context.Context, customerID string, usage []*Usage) error
+
+	GetReportingGranularity() UsageReportingGranularity
+	GetReportingWorkerCron() string
+}
+
+type Plan struct {
+	BillerID          string // ID of the plan in the external billing system
+	RillID            string // ID of the plan in Rill, can be empty if biller does not support it
+	Name              string
```
begelundmuller, line 47: I would suggest aligning the terminology here with the one we have for orgs, projects, etc. – i.e. using "ID" for non-user facing IDs, "name" for unique user-facing names (e.g. for use in the CLI or URL slugs), and "display name" for friendly non-unique user facing names (like UI titles).

```go
type Plan struct {
  ID string // ID from Orb, e.g. "6753a734-e327-43c4-a46a-55f8d00721f5"
  Name string // Unique name we assign, e.g. "starter-plan"
  DisplayName string // Friendly name, e.g. "Starter Plan"
  ...
}

File: admin/billing/orb.go [hunk 20]
```diff
@@ -0,0 +1,422 @@
...
+	"github.com/orbcorp/orb-go/option"
+	"github.com/rilldata/rill/admin/database"
+)
+
+const (
+	requestMaxLimit = 500
+	requestTimeout  = 10 * time.Second
+)
+
+var ErrNotFound = errors.New("not found")
```
begelundmuller, line 22: nit: Move to `biller.go` since this in theory could be returned by multiple billing implementations

begelundmuller (changes requested):

admin/billing/orb.go
```diff
@@ -0,0 +1,377 @@
...
+	}
+
+	var billingPlans []*Plan
+	for i := 0; i < len(plans.Data); i++ {
+		billingPlan, err := getBillingPlanFromOrbPlan(&plans.Data[i])
+		if err != nil {
+			return nil, err
+		}
+		billingPlans = append(billingPlans, billingPlan)
+	}
```
begelundmuller, line 320: Should we maybe filter out those plans where `Name == ""`? In case we forget to set it in Orb, then could someone send a manual API call to upgrade to `""` and then get access to some internal plan? Or is there something that protects against that?

admin/metrics/client.go
```diff
@@ -98,3 +98,101 @@ func (c *Client) AutoscalerSlotsRecommendations(ctx context.Context, limit, offs
...
 	return recommendations, nil
 }
+
+type Usage struct {
+	OrgID      string    `json:"org_id"`
+	ProjectID  string    `json:"project_id"`
+	StartTime  time.Time `json:"start_time"`
+	EndTime    time.Time `json:"end_time"`
+	MetricName string    `json:"metric_name"`
+	Amount     float64   `json:"amount"`
```
begelundmuller, line 108: Should we maybe call this `value` or `max_value` instead?

admin/worker/billing_reporter.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	if !ok {
+		w.logger.Debug("skipping usage reporting: no metrics project configured")
+		return nil
+	}
+
+	u, err := client.GetUsageMetrics(ctx, startTime, endTime, sqlGrainIdentifier)
+	if err != nil {
+		return fmt.Errorf("failed to get usage metrics: %w", err)
+	}
```
begelundmuller, line 61: Can we have some pagination for this? E.g. if we add other billable metrics, this might get very large and not fit in a single API request (or maybe even in memory).

For example, we could order by event time and paginate using it? And keep paging until we reach a page that is sized below the limit. (And call `UpdateBillingUsageReportedOn` for each page so if it crashes, it won't keep reprocessing the same pages.)

admin/worker/billing_reporter.go
```diff
@@ -0,0 +1,115 @@
...
+	err = w.admin.Biller.ReportUsage(ctx, usage)
+	if err != nil {
+		return fmt.Errorf("failed to report usage: %w", err)
+	}
+
+	// update last reporting time to endTime as it is the max processed event time
+	err = w.admin.DB.UpdateBillingUsageReportedOn(ctx, endTime)
+	if err != nil {
+		return fmt.Errorf("failed to update last usage reporting time: %w", err)
+	}
```
begelundmuller, line 92: In case of data delays in intake into the metrics project, this might advance the watermark too far. Should we use the max `m.EndTime` observed in the results instead?

admin/database/database.go
```diff
@@ -213,6 +213,12 @@ type DB interface {
 	UpsertVirtualFile(ctx context.Context, opts *InsertVirtualFileOptions) error
 	UpdateVirtualFileDeleted(ctx context.Context, projectID, branch, path string) error
 	DeleteExpiredVirtualFiles(ctx context.Context, retention time.Duration) error
+
+	FindBillingOrganizationIDs(ctx context.Context) ([]string, error)
```
begelundmuller, line 217: I first thought this returned billing customer IDs. Can we call with `FindOrganizationIDsWithBilling` or something like that?

admin/worker/billing_reporter.go
```diff
@@ -0,0 +1,115 @@
...
+			count, err := w.admin.DB.CountBillingProjectsForOrganization(ctx, org, endTime)
+			if err != nil {
+				w.logger.Warn("failed to validate active projects for org", zap.String("org", org), zap.Error(err))
+				continue
+			}
+			if count > 0 {
+				w.logger.Warn("skipping usage reporting for org: no usage data available", zap.String("org", org), zap.Time("start_time", startTime), zap.Time("end_time", endTime))
+			}
+		}
+	}
```
begelundmuller, line 113: (Not needed for this PR) To avoid loading all orgs into memory at once, and to speed up the job, we could consider pulling this logic into a separate job. The new job could be a kind of "billing usage health check job" that maybe even just calls Orb directly and checks if usage data is available for within the last few hours for each org with billing projects.

proto/rill/admin/v1/api.proto
```diff
@@ -1646,3 +1735,29 @@ message AlertOptions {
   // base64 state url of the dashboard when it was created
   string web_open_state = 12;
 }
+
+message ListPublicBillingPlansRequest {}
+
+message ListPublicBillingPlansResponse {
+  repeated BillingPlan plans = 1;
+}
```
begelundmuller, line 1743: Should be placed in the correct relative location in the file (the style guide says first service defs, then request/response messages, then other messages)

begelundmuller (changes requested):

admin/billing/orb.go
pjain1: Good catch but it cannot happen currently, in the API where `GetPlanByName` is used, a check exists to disallow empty string. Also there is option to add plan in the cli cmd using biller plan id as well (not sure if we need that), so filtering would cause that to fail. 

But to be more defensive I can return err from `GetPlanByName` if name is empty ?

admin/metrics/client.go
pjain1: changed to value as agg func can change as per the metric

admin/worker/billing_reporter.go
pjain1: thanks updated

admin/billing/orb.go
begelundmuller: Sounds good with the additional defensive option

proto/rill/admin/v1/api.proto
```diff
@@ -45,6 +45,24 @@ service AdminService {
     };
   }
 
+  // UpdateOrganizationBillingPlan updates the billing plan for the organization
+  rpc UpdateOrganizationBillingPlan(UpdateOrganizationBillingPlanRequest) returns (UpdateOrganizationBillingPlanResponse) {
+    option (google.api.http) = {
+      patch: "/v1/organizations/{org_name}/billing/plan",
+      body: "*"
+    };
+  }
```
begelundmuller, line 54: Should we rename this to `UpdateOrganizationBillingSubscription` to align with the naming of `GetOrganizationBillingSubscription`?

proto/rill/admin/v1/api.proto
```diff
@@ -45,6 +45,24 @@ service AdminService {
...
+
+  // GetOrganizationBillingSubscription lists the subscription for the organization
+  rpc GetOrganizationBillingSubscription(GetOrganizationBillingSubscriptionRequest) returns (GetOrganizationBillingSubscriptionResponse) {
+    option (google.api.http) = {get: "/v1/organizations/{org_name}/billing/subscriptions"};
+  }
+
+  // DeleteOrganizationSubscription deletes the given subscription for the organization
+  rpc DeleteOrganizationBillingSubscription(DeleteOrganizationBillingSubscriptionRequest) returns (DeleteOrganizationBillingSubscriptionResponse) {
+    option (google.api.http) = {delete: "/v1/organizations/{org_name}/billing/subscriptions/{subscription_id}"};
+  }
```
begelundmuller, line 64: Maybe we already discussed this – but why is it we need the ability to delete a subscription? Won't it put the org into a weird state? If you want to stop a paid subscription, shouldn't you just call `UpdateOrganizationBillingPlan` to change to a free plan?

proto/rill/admin/v1/api.proto
```diff
@@ -606,6 +637,40 @@ message UpdateOrganizationResponse {
   Organization organization = 1;
 }
 
+message UpdateOrganizationBillingPlanRequest {
+  string org_name = 1;
+  optional string plan_name = 2;
+  optional string biller_plan_id = 3;
```
begelundmuller, line 643: Can we decide on just one way of doing this?

begelundmuller (commented):
Also check out the merge conflicts

proto/rill/admin/v1/api.proto
pjain1: Yeah my original intention was to allow for fix for issues that could arise from manual assignment, also don't see any free plan in the mocks, I just see Starter plan which has 30 days trial period and then its all paid. In that case as well if no subscprition then org should be hibernated, not sure but its a separate issue. Anyways removed.

proto/rill/admin/v1/api.proto
pjain1: moved to using plan name only as its a more friendly option for cli but also enforces that all plans defined in Orb have an external plan id defined.

begelundmuller (approved):

PR #5054, opened 2024-06-11
Add sorting and full filter support for expanded pivot sub tables

djbarnwal:
Add sorting and full filter support for expanded pivot sub tables. Also fixes https://github.com/rilldata/rill-private-issues/issues/407


AdityaHegde (approved):
Approving this, but we should probably find a way to simplify this through either the API or code splitting.

Also update the comment at the top of `createPivotDataStore`

PR #5059, opened 2024-06-11
Refactor: Unify Input component

briangregoryholmes:
This PR consolidates the `Input` and `InputV2` components and adds a `secret` property that allows for the toggling of password visibility.


ericpgreen2 (approved):
UXQA – 
1. In the Add Source dialog, we autogenerate a Source Name based on certain form inputs. For example, in the GCS form, we use the GS URI to generate a Source Name. This doesn't seem to be working on this branch.
2. The Bookmark form has too much vertical space between the Input labels and the Input elements. It looks like the `Input` component should have a wrapper that combines the label and the input?
image
3. A nice-to-have (but expanded scope, so not needed for this PR), would be to leverage this `Input.svelte` code in the `InputArray.svelte` component. I noticed that the styling diverges between the two components in the Alert and Scheduled Report forms.

PR #5063, opened 2024-06-11
Static assets endpoint

egor-ryashin:


runtime/compilers/rillv1/parse_rillyaml.go
```diff
@@ -67,7 +68,8 @@ type rillYAML struct {
 	// Default YAML values for migrations
 	Migrations yaml.Node `yaml:"migrations"`
 	// Feature flags (preferably a map[string]bool, but can also be a []string for backwards compatibility)
-	Features yaml.Node `yaml:"features"`
+	Features    yaml.Node `yaml:"features"`
+	PublicPaths []string  `yaml:"public_paths"`
```
begelundmuller, line 72:
```suggestion
	Features yaml.Node `yaml:"features"`
	// Paths to expose over HTTP (defaults to ./public)
	PublicPaths []string  `yaml:"public_paths"`
```

runtime/drivers/admin/admin.go
```diff
@@ -279,9 +283,45 @@ func (h *Handle) cloneOrPull(ctx context.Context) error {
...
+					}
+					cache[strings.TrimLeft(rel, "/")] = b
+				}
+				return nil
+			})
+			if err != nil {
+				return nil, err
+			}
+		}
+		h.assetsCache = cache
```
begelundmuller, line 323: This is not a useful cache. The place where caching is nice is to be able to serve assets immediately when the runtime starts, even if Github takes a few minutes to clone. This means the cached files would need to be stored on the persistent disk, and be readable before the first call to `cloneOrPull`.

Also, we should not eagerly cache the assets in memory – it can lead to excessive memory usage if there are larger files. Relying on the OS file system page cache is more than adequate for our performance requirements here.

runtime/drivers/file/repo.go
```diff
@@ -159,6 +179,41 @@ func (c *connection) Delete(ctx context.Context, filePath string, force bool) er
...
+			}
+			return nil
+		})
+		if err != nil {
+			return err
+		}
+	}
+	c.cacheMutex.Lock()
+	defer c.cacheMutex.Unlock()
+	c.assetsCache = cache
```
begelundmuller, line 216: The `file` driver does not have a cloning latency like Github has, so there is no need for caching assets in it.

runtime/reconcilers/project_parser.go
```diff
@@ -290,6 +290,17 @@ func (r *ProjectParserReconciler) reconcileParser(ctx context.Context, inst *dri
...
+	paths := parser.RillYAML.PublicPaths
+	if len(paths) == 0 {
+		paths = []string{"public"}
+	}
+	repo, release, err := r.C.Runtime.Repo(ctx, parser.InstanceID)
+	if err != nil {
+		return err
+	}
+	defer release()
+	repo.SetCachedPaths(paths)
```
begelundmuller, line 302: This should be handled inside `UpdateInstanceWithRillYAML`

runtime/server/assets.go
```diff
@@ -0,0 +1,52 @@
...
+
+	repo, release, err := s.runtime.Repo(ctx, instanceID)
+	if err != nil {
+		return err
+	}
+	defer release()
+
+	paths := repo.GetCachedPaths()
+	allowed := false
+	for _, p := range paths {
```
begelundmuller, line 36: It's too implicit to rely on the repo's cached paths as an allow list. It only works as long as that feature is only used for `PublicPaths`. 

Instead, I would suggest explicitly adding `PublicPaths` as a property on `Instance`, assigning it in `UpdateInstanceWithRillYAML`, and checking it here.

runtime/server/assets.go
```diff
@@ -0,0 +1,52 @@
...
+		if strings.HasPrefix(strings.TrimLeft(path, ","), strings.TrimLeft(p, ",")) {
```
begelundmuller, line 37: 1. This is not a safe way to check the path. For example, if `/public` is the public directory, it will match `/publications/secret.txt`.
2. Why is it necessary to trim commas from the paths?

runtime/server/assets.go
```diff
@@ -0,0 +1,52 @@
...
+			allowed = true
+			break
+		}
+	}
+	if !allowed {
+		return fmt.Errorf("path is not allowed")
```
begelundmuller, line 43: Should return a `httputil.Error` with an appropriate status code

begelundmuller (changes requested):

begelundmuller:
I would suggest completing this PR without caching support and optionally taking it up in a follow up PR.

runtime/server/assets.go
egor-ryashin: so you suggest adding an instance reference to repo, right? because there's no instance in the params list 
image

runtime/drivers/admin/admin.go
egor-ryashin: I wonder how a user can get into a state where there's cache on disk but no cloned repo? Because cache can be created from a clone repo.

And why would user like to delete a cloned repo leaving just the cache on disk?

begelundmuller: The repo is currently cloned into a temp directory which will not exist after a process restart: https://github.com/rilldata/rill/blob/1983fb1194d9f358eb5321534b15a4a824b0910c/runtime/drivers/admin/admin.go#L356.

It might be fine to cache the full repo on the persistent disk – that just isn't what it's doing now.

runtime/server/assets.go
begelundmuller: No, it could just pass the paths to cache, no need to pass the full instance. But as discussed above, it might be fine just to cache the whole repo on a persistent disk.

begelundmuller: I would still suggest to omit caching from this PR since it's not so important

runtime/drivers/registry.go
```diff
@@ -60,6 +60,8 @@ type Instance struct {
 	EmbedCatalog bool `db:"embed_catalog"`
 	// WatchRepo indicates whether to watch the repo for file changes and reconcile them automatically.
 	WatchRepo bool `db:"watch_repo"`
+	// Paths to expose over HTTP (defaults to ./public)
+	PublicPaths []string `db:"public_paths"`
```
begelundmuller, line 64: Need to add migrations and handling for it in `runtime/drivers/sqlite` (can search for `feature_flags` to find locations)

runtime/reconcilers/project_parser.go
```diff
@@ -290,6 +290,10 @@ func (r *ProjectParserReconciler) reconcileParser(ctx context.Context, inst *dri
 		return parseErrsErr
 	}
 
+	if len(parser.RillYAML.PublicPaths) == 0 {
+		parser.RillYAML.PublicPaths = []string{"public"}
+	}
+
```
begelundmuller, line 296: Can be removed since the default is assigned in `parse_rillyaml.go`

runtime/drivers/file/file.go
```diff
@@ -58,6 +58,7 @@ type configProperties struct {
 // a smaller subset of relevant parts of rill.yaml
 type rillYAML struct {
 	IgnorePaths []string `yaml:"ignore_paths"`
+	CachedPaths []string `yaml:"public_paths"`
```
begelundmuller, line 61:
```suggestion
	PublicPaths []string `yaml:"public_paths"`
```

begelundmuller (changes requested):

begelundmuller:
@egor-ryashin After this PR merges, can you also update in #team-applications about the new endpoint? So that they can wire it up to the custom dashboard components.

begelundmuller (approved):

PR #5066, opened 2024-06-12
DataYAML support in alerts and reports

esevastyanov:


runtime/reconcilers/alert.go
```diff
@@ -561,30 +584,40 @@ func (r *AlertReconciler) executeSingleWrapped(ctx context.Context, self *runtim
...
-	err = r.C.Runtime.Query(ctx, r.C.InstanceID, q, alertQueryPriority)
+	res, err := r.C.Runtime.Resolve(ctx, &runtime.ResolveOptions{
+		InstanceID: r.C.InstanceID,
+		Resolver:   "legacy_metrics",
+		ResolverProperties: map[string]any{
+			"query_name":      a.Spec.QueryName,
+			"query_args_json": a.Spec.QueryArgsJson,
+			"is_formatted":    true,
+			"row_count":       1,
+		},
```
begelundmuller, line 595: How about moving this logic into `parse_alert.go` and completely removing `Spec.QueryName` and `Spec.QueryArgsJSON` from the spec? I.e. if the YAML has a `Query`, it just sets the spec's resolver to to `legacy_metrics` right away?

Each time the runtime is restarted, it re-parses all YAML files and overwrites any resources that have been changed, so this should be safe (there might be a small race condition on the restart around whether an alert is checked before the re-parse is complete – so just need to maintain the `if a.Spec.Resolver != "" {` in here to avoid crashes).

runtime/resolvers/legacy_metrics.go
```diff
@@ -0,0 +1,329 @@
...
+	rowCount        int
+	isFormatted     bool
+	logger          *zap.Logger
+}
+
+type legacyMetricsResolverProps struct {
+	QueryName     string `mapstructure:"query_name"`
+	QueryArgsJSON string `mapstructure:"query_args_json"`
+	RowCount      int    `mapstructure:"row_count"`
+	IsFormatted   bool   `mapstructure:"is_formatted"`
```
begelundmuller, line 40: Maybe call it `Format: true`? Since it's something the resolver should do, not something that "is" already done?

runtime/resolvers/legacy_metrics.go
```diff
@@ -0,0 +1,329 @@
...
+	metricsViewName string
+	rowCount        int
+	isFormatted     bool
+	logger          *zap.Logger
+}
+
+type legacyMetricsResolverProps struct {
+	QueryName     string `mapstructure:"query_name"`
+	QueryArgsJSON string `mapstructure:"query_args_json"`
+	RowCount      int    `mapstructure:"row_count"`
```
begelundmuller, line 39: Maybe call this `Limit`? I also wonder if it makes sense to move this to `ResolverOptions` and implementing it generically for all resolvers? Since for any resolver, the alert only needs the first row. But maybe that is over engineering.

runtime/compilers/rillv1/parse_alert.go
```diff
@@ -116,65 +117,79 @@ func (p *Parser) parseAlert(node *Node) error {
...
+			return errors.New(`missing query args (must set either "query.args" or "query.args_json")`)
 		}
-	}
-	if n > 1 {
-		return fmt.Errorf(`only one of "query.for.user_id", "query.for.user_email", or "query.for.attributes" may be set`)
-	}
 
-	if len(tmp.Email.Recipients) > 0 && len(tmp.Notify.Email.Recipients) > 0 {
-		return errors.New(`cannot set both "email.recipients" and "notify.email.recipients"`)
+		// Query for: validate only one of user_id, user_email, or attributes is set
```
begelundmuller, line 158: We need to also support `Query.For` for resolvers. The resolved attributes can be passed in `ResolveOptions.UserAttributes` (needed to enforce security policies).

Maybe we should just support `For` at the top-level of the yaml? Like:
```
type: alert
data:
  ...
for:
  ...
notify:
  ...

File: admin/server/alerts.go [hunk 5]
```diff
@@ -577,6 +599,13 @@ type alertYAML struct {
 	Intervals struct {
 		Duration string `yaml:"duration"`
 	} `yaml:"intervals"`
+	Data struct {
+		Connector  string         `yaml:"connector"`
+		SQL        string         `yaml:"sql"`
+		MetricsSQL string         `yaml:"metrics_sql"`
+		API        string         `yaml:"api"`
+		Args       map[string]any `yaml:"args"`
+	} `yaml:"data"`
```
begelundmuller, line 608: You could also consider just having `Data map[string]any` and relying on the frontend to provide a valid resolver? It's a bit less safe, but reduces duplication of the resolver definition.

begelundmuller (changes requested):

proto/rill/admin/v1/api.proto
```diff
@@ -1677,4 +1677,5 @@ message AlertOptions {
   repeated string slack_webhooks = 11;
   // base64 state url of the dashboard when it was created
   string web_open_state = 12;
+  google.protobuf.Struct data_props = 13;
```
begelundmuller, line 1680: 1. Split into `resolver` and `resolver_properties` (to be serialized into `data:` in the YAML as `data[resolver] = resolverProperties`)
2. Move up above `query_name` and put `// Deprecated: ` after `query_name` and `query_args_json`

proto/rill/runtime/v1/resources.proto
```diff
@@ -350,6 +348,9 @@ message AlertSpec {
   uint32 renotify_after_seconds = 19;
   repeated Notifier notifiers = 21;
   map<string, string> annotations = 20;
+  string resolver = 22;
+  google.protobuf.Struct resolver_properties = 23;
+  google.protobuf.Struct data_properties = 24;
```
begelundmuller, line 353: 1. Move `resolver` and `resolver_properties` up where `query_name` was (after `timeout_seconds`) – the field order in code doesn't need to match the field numbers (which are just for wire compatibility).
2. Should be able to remove the `data_properties` field – that's a YAML thing which is parsed into `resolver` and `resolver_properties`

runtime/compilers/rillv1/parse_alert.go
```diff
@@ -25,8 +26,14 @@ type AlertYAML struct {
 		Limit         uint   `yaml:"limit"`
 		CheckUnclosed bool   `yaml:"check_unclosed"`
 	} `yaml:"intervals"`
-	Timeout string `yaml:"timeout"`
-	Query   struct {
+	Timeout string         `yaml:"timeout"`
+	Data    map[string]any `yaml:"data"` // will be parsed as DataYAML, is defined as raw map to save the original content
```
begelundmuller, line 30: We should use `DataYAML` here as in the other resources that use it. We should be able to avoid needing to capture the raw contents of it.

runtime/resolvers/legacy_metrics.go
```diff
@@ -0,0 +1,346 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse query: %w", err)
+	}
+
+	metricsViewName, err := queries.MetricsViewFromQuery(props.QueryName, props.QueryArgsJSON)
+	if err != nil {
+		return nil, fmt.Errorf("failed extract metrics view name from query: %w", err)
+	}
+
+	q, err := queries.ProtoToQuery(qpb, opts.UserAttributes)
```
begelundmuller, line 68: After the introduction of shareable URLs recently, this now also needs to incorporate `opts.SecurityPolicy` (a security policy provided in the JWT instead of in the resource itself)

begelundmuller (changes requested):

begelundmuller (approved):

PR #5068, opened 2024-06-12
Add Superforms, update RenameAssetModal

briangregoryholmes:
This PR begins the process of replacing [svelte-forms-lib](https://github.com/tjinauyeung/svelte-forms-lib), which is no longer supported, with [Superforms](https://github.com/ciscoheat/sveltekit-superforms).

It reworks the `RenameAssetModal` to use the new API. Follow up PRs will replace other uses of the old form library.


web-common/src/features/entity-management/RenameAssetModal.svelte
```diff
@@ -20,36 +18,46 @@
...
 
-  $: existingDirectories = useDirectoryNamesInDirectory(
-    runtimeInstanceId,
-    folder,
-  );
+  const validationSchema = object({
+    newName: string()
+      .required("Enter a name!")
+      .matches(VALID_NAME_PATTERN, INVALID_NAME_MESSAGE),
+  });
```
ericpgreen2, line 37: I can't seem to trigger this validation message. When I leave the form empty, then try to submit it, I don't see the message "Enter a name!" pop up anywhere.

ericpgreen2 (approved):

web-common/src/features/entity-management/RenameAssetModal.svelte
briangregoryholmes: Unifying the Input components meant bringing over the conditionally shown error behavior from InputV2, I believe. One of the conditionals is that there exists a value in the input. Obviously that's in conflict with an error for a missing value... Turned on the alwaysShowError prop for now, though I may consider removing this conditional behavior as I convert more forms.

Issue #4911, opened 2024-05-16
Nit: Clicking dashboard title in preview should open switcher


mindspank:
<img width="305" alt="Screenshot 2024-05-16 at 14 01 37" src="https://github.com/rilldata/rill/assets/1181922/b6cae8bf-7898-46bc-8125-e349d04137e0">

We have the Dashboard title indicating that it's clickable but it doesn't do anything. 
I would expect it would open the dashboard switcher dropdown when being clicked.

mindspank:
Either don't have a hover pointer or we can display the dropdown on click. Both would be acceptable.

briangregoryholmes:
So, what do we think about the following PR: https://github.com/rilldata/rill/pull/5079

This does three things:

1) Changes the styling of the dropdown arrow to shift background color rather than "drop" on hover so that the separation between the "link" and the menu trigger is clearer.
2) Adds the project title back to the Rill Developer breadcrumbs so that the font weight makes it clearer that there are "active" sections
3) When you click the breadcrumb link of the current page, it actively refreshes (rather than what seems like a no-op) and, in the case of Rill Cloud, takes you back to the "main" tab, which, in our case is `Dashboards`.

I worry that having different behavior for this component in different places would be similarly confusing. IMO, the title should always be a link and the arrow should always be a trigger.

Here's my reasoning:

<img width="491" alt="Screenshot 2024-06-14 at 4 29 20 PM" src="https://github.com/rilldata/rill/assets/120223836/4105ae27-d21a-494f-84dd-7301b2a9944a">

On Cloud, I think the behavior makes a lot more sense. I think the font weight and separators make it clear that there are different elements with different behaviors and that clicking the active segment shouldn't really "do" anything

<img width="502" alt="Screenshot 2024-06-14 at 4 30 00 PM" src="https://github.com/rilldata/rill/assets/120223836/fb6c1356-e170-41cd-9380-1d56a3cfec1c">

This matches Vercel's behavior. Clicking the project title has doesn't do anything unless you're on a tab other than `Overview`, in which case it takes you back there. It's just a link to the root page, which you may already be on.

<img width="231" alt="Screenshot 2024-06-14 at 4 29 33 PM" src="https://github.com/rilldata/rill/assets/120223836/d9d75acd-4b64-49dd-baca-d4a02ddad41c">

[...]

PR #5079, opened 2024-06-14
breadcrumb tweak

briangregoryholmes:
See discussion here: https://github.com/rilldata/rill/issues/4911

### DIFF
--- a/web-local/src/routes/(viz)/+layout.svelte
+++ b/web-local/src/routes/(viz)/+layout.svelte
@@ -33,7 +33,7 @@
     return map;
   }, new Map<string, PathOption>());
 
-  $: huh = <PathOption>{
+  $: projectPath = <PathOption>{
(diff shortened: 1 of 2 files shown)


web-local/src/routes/(viz)/+layout.svelte
```diff
@@ -29,9 +33,19 @@
     return map;
   }, new Map<string, PathOption>());
 
-  $: pathParts = [dashboardOptions];
+  $: huh = <PathOption>{
```
ericpgreen2, line 36: `huh` is from Brian's dev mode, right...?

ericpgreen2 (approved):

web-local/src/routes/(viz)/+layout.svelte
briangregoryholmes: 🫠

PR #5080, opened 2024-06-15
Cloud UI: Fix 401 error on the Project page due to stale JWT 

ericpgreen2:
Currently, re-visiting a Project page can trigger a 401 error. In the error case, a stale JWT sits in the `GetProject` Query Cache and is used in a runtime request. The 401 triggers a full-screen error page.

This PR does a few things to mitigate this, notably:
1. Adds `refetchOnMount`, `refetchOnWindowFocus`, and `refetchOnReconnect` to refetch the JWT when the user re-visits the page
2. Explicitly adds `cacheTime` to garbage-collect old JWTs
3. Removes the full-screen error page for runtime 401s on the Project page

[Bug report in Slack](https://rilldata.slack.com/archives/C01190F1R1D/p1718237032223209?thread_ts=1718235808.839709&cid=C01190F1R1D)

### DIFF
--- a/web-admin/src/features/errors/error-utils.ts
+++ b/web-admin/src/features/errors/error-utils.ts
@@ -52,18 +52,27 @@ export function createGlobalErrorCallback(queryClient: QueryClient) {
 
     // Special handling for some errors on the Project page
     const onProjectPage = isProjectPage(get(page));
-    if (onProjectPage && error.response?.status === 400) {
-      // If "repository not found", ignore the error and show the page
-      if (
-        (error.response.data as RpcStatus).message === "repository not found"
-      ) {
-        return;
+    if (onProjectPage) {
+      if (error.response?.status === 400) {
+        // If "repository not found", ignore the error and show the page
+        if (
+          (error.response.data as RpcStatus).message === "repository not found"
+        ) {
+          return;
+        }
+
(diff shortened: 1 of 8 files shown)


web-admin/src/features/errors/error-utils.ts
```diff
@@ -52,18 +52,27 @@ export function createGlobalErrorCallback(queryClient: QueryClient) {
...
+          return;
+        }
       }
 
-      // This error is the error:`driver.ErrNotFound` thrown while looking up an instance in the runtime.
-      if ((error.response.data as RpcStatus).message === "driver: not found") {
-        const [, org, proj] = get(page).url.pathname.split("/");
-        void queryClient.resetQueries(getProjectRuntimeQueryKey(org, proj));
+      // If the runtime throws a 401, it's likely due to a stale JWT that will soon be refreshed
+      if (isRuntimeQuery(query) && error.response?.status === 401) {
```
AdityaHegde, line 75: What will happen now when the deployment is not available? (I believe `driver: not found` comes mainly from that use case)

AdityaHegde (approved):
Just left one comment. Looks good otherwise

web-admin/src/features/errors/error-utils.ts
ericpgreen2: There shouldn't be any change of behavior there. That case was just moved to lines 64-71.

PR #5083, opened 2024-06-17
Runtime: Better clickhouse modelling support

k-anshul:
Example model that incrementally fetches data from a postgres server:

```
type: model
sql: SELECT * except(password) FROM postgresql('localhost:5432', 'postgres', 'new_users', 'postgres', 'postgres') {{ if incremental }} WHERE created_at > parseDateTime64BestEffort('{{ .state.max_created_at }}') {{ end }}

output:
    columns: (id UInt32, username varchar, email varchar, created_at datetime, INDEX idx1 username TYPE set(100) GRANULARITY 3) # also adds an index
  incremental_strategy: append
  materialize: true
  engine: ReplacingMergeTree(id)
  order_by: (id, created_at)

incremental: true
state:
  sql: SELECT MAX(created_at) AS max_created_at FROM new_users

refresh:
  cron: '* * * * *'
 ```

### CODE REVIEW

File: runtime/drivers/clickhouse/clickhouse.go [hunk 1]
```diff
@@ -324,15 +324,6 @@ func (c *connection) AsModelManager(instanceID string) (drivers.ModelManager, bo
...
 func (c *connection) AsTransporter(from, to drivers.Handle) (drivers.Transporter, bool) {
-	olap, _ := to.(*connection)
-	if c == to {
-		switch from.Driver() {
-		case "s3":
-			return NewS3Transporter(from, olap, c.logger), true
-		case "https":
-			return NewHTTPTransporter(from, olap, c.logger), true
-		}
-	}
```
begelundmuller, line 335: Do you know if anyone were using these? I think not, but if yes we might need a shim. The long term plan is to emit `type: source` as models from the parser, so we don't have to handle them differently at all, but we can't do that until all the transporters have been migrated.

runtime/drivers/clickhouse/model_manager.go
```diff
@@ -21,8 +21,11 @@ func (p *ModelInputProperties) Validate() error {
 }
 
 type ModelOutputProperties struct {
-	Table       string `mapstructure:"table"`
-	Materialize *bool  `mapstructure:"materialize"`
+	Table               string                      `mapstructure:"table"`
+	Materialize         *bool                       `mapstructure:"materialize"`
+	UniqueKey           []string                    `mapstructure:"unique_key"`
+	IncrementalStrategy drivers.IncrementalStrategy `mapstructure:"incremental_strategy"`
+	// can have other properties like order_by clause, primary key clause etc
```
begelundmuller, line 28: Can you list those out here? And then parse `tableOpts` to `ModelOutputProperties` in `CreateTableAsSelect` instead? Will make it easier to see (and maybe eventually generate docs for) the supported properties.

runtime/drivers/clickhouse/olap.go
```diff
@@ -206,22 +208,124 @@ func (c *connection) AlterTableColumn(ctx context.Context, tableName, columnName
...
+				Query: fmt.Sprintf("DELETE FROM %s WHERE (%s) IN (SELECT %s FROM %s)", safeSQLName(name), cols, cols, safeSQLName(tmp)),
+			})
+			if err != nil {
+				return err
+			}
+
+			// Insert the new data into the target table
+			return c.Exec(ctx, &drivers.Statement{
+				Query: fmt.Sprintf("INSERT INTO %s SELECT * FROM %s", safeSQLName(name), safeSQLName(tmp)),
+			})
```
begelundmuller, line 324: I'm fine with keeping this if you think it has use cases, but just FYI we don't need to support all strategies for all engines. I'm wondering if given ClickHouse has `ReplacingMergeTree` (even with its quirks of eventual consistency), we should just recommend people use that instead? Because ClickHouse doesn't have transactions, right? So there might be queries in between deleting the old data and inserting the new?

runtime/drivers/clickhouse/olap.go
```diff
@@ -206,22 +208,124 @@ func (c *connection) AlterTableColumn(ctx context.Context, tableName, columnName
...
 			Priority: 100,
 		})
 	}
+	var create strings.Builder
+	create.WriteString("CREATE OR REPLACE TABLE ")
+	create.WriteString(safeSQLName(name))
+	// see if there is any column override
+	if prop, ok := tableOpts["columns"].(string); ok {
+		create.WriteString(prop)
+	}
```
begelundmuller, line 224: See other comment – can we parse into `ModelOutputProperties` here to avoid these type casts?

begelundmuller (changes requested):

runtime/drivers/clickhouse/clickhouse.go
k-anshul: I don't think anybody is using them. Most of the current use cases rely on data being populated offline.
These implementations were pretty bad and had following issues:
1. Credentials have to be set on clickhouse server
2. If credentials are set on clickhouse server we can't use any anonymous sources. 
3. Can not have any column override which is required in mostly all cases.
4. Can not set any sorting keys or primary keys which means clickhouse need to scan entire table for every query,

The current approach handles all these cases plus we can also support all possible integrations. Example model that fetches data from s3 with current implementation.
```
type: model
sql: SELECT * s3(...)

output:
  columns: (id UInt32, username varchar, email varchar, created_at datetime, INDEX idx1 username TYPE set(100) GRANULARITY 3) # also adds an index
  materialize: true
  order_by: (id, created_at)

refresh:
  cron: '* * * * *'
```

The clickhouse transporters were added only for the POC and I believe we should remove them given the shortcomings listed above.

runtime/drivers/clickhouse/olap.go
k-anshul: Okay. I thought we want to keep consistency with respect to supported strategies. In the other case I think its better to remove this since lightweight deletes are also experimental in clickhouse.

> So there might be queries in between deleting the old data and inserting the new

From what I understand these lightweight deletes are stored separately and table is not changed immediately. For select queries the deleted rows are handled on the fly and eventually a background migration fixes the table atomically. But given these deletes happen in background there are lot of considerations to think of like these can also fail and clickhouse keeps complaining about a failed migration on subsequent queries. I think its best to remove them for now.

runtime/drivers/clickhouse/olap.go
k-anshul: Do you mean we should change `	CreateTableAsSelect(... tableOpts map[string]any) error` to `CreateTableAsSelect(... tableOpts map[string]string) error` because then only we can avoid these type casts ?

And on the caller side generate `tableOpts` from `ModelOutputProperties` ?
```
tableOpts := make(map[string]string)
if err := mapstructure.WeakDecode(outputProps, tableOpts); err != nil {
	return nil, fmt.Errorf("failed to generate tableOpts: %w", err)
}
```

runtime/drivers/clickhouse/clickhouse.go
begelundmuller: Good, let's not worry about this then

runtime/drivers/clickhouse/olap.go
begelundmuller: No, I just mean that if you add the props in `ModelOutputProperties` as this comment says: https://github.com/rilldata/rill/pull/5083#discussion_r1648170276.

Then in this func you can do something like:
```go
props := &ModelOutputProperties{}
err := mapstructure.WeakDecode(tableOpts, props)
// use props.Columns instead of tableOpts["columns"].(string)
```

runtime/drivers/clickhouse/olap.go
begelundmuller: Okay, let's remove it then. With both deletes and inserts having eventual consistency, there could be cases where a query is missing data.

begelundmuller (approved):

PR #5090, opened 2024-06-17
Comparison aggregation unnest testing

egor-ryashin:


runtime/reconcilers/metrics_view.go
```diff
@@ -76,12 +77,16 @@ func (r *MetricsViewReconciler) Reconcile(ctx context.Context, n *runtimev1.Reso
...
+
 	if ctx.Err() != nil {
 		return runtime.ReconcileResult{Err: errors.Join(validateErr, ctx.Err())}
 	}
 	if validateErr == nil {
 		mv.State.ValidSpec = mv.Spec
 	} else {
+		if os.Getenv("DEBUG") == "1" {
+			fmt.Println("Reconciler error", validateErr.Error())
+		}
```
begelundmuller, line 89: Remove

runtime/reconcilers/project_parser.go
```diff
@@ -264,6 +265,10 @@ func (r *ProjectParserReconciler) reconcileParser(ctx context.Context, inst *dri
 				continue
 			}
 			r.C.Logger.Warn("Parser error", zap.String("path", e.FilePath), zap.String("error", e.Message))
+			debug := os.Getenv("DEBUG")
+			if debug == "1" {
+				fmt.Println("Parser error", e.Message)
+			}
```
begelundmuller, line 271: Remove

runtime/testruntime/testruntime.go
```diff
@@ -244,7 +260,7 @@ func NewInstanceForDruidProject(t TestingT) (*runtime.Runtime, string) {
 			{
 				Type:   "druid",
 				Name:   "druid",
-				Config: map[string]string{"dsn": "http://localhost:8888/druid/v2/sql"},
+				Config: map[string]string{"dsn": fmt.Sprintf("https://%s:%s@druid.metrics.in.rilldata.io/druid/v2/sql", splits[0], splits[1])},
```
begelundmuller, line 263: Remove – we should use test containers instead of running against a prod cluster in testing. If we really need to run against an external cluster, it should be a dedicated cluster and generically configured

runtime/queries/metricsview.go
```diff
@@ -76,7 +76,7 @@ func lookupMetricsView(ctx context.Context, rt *runtime.Runtime, instanceID, nam
 
 	res, err := ctrl.Get(ctx, &runtimev1.ResourceName{Kind: runtime.ResourceKindMetricsView, Name: name}, false)
 	if err != nil {
-		return nil, nil, status.Error(codes.InvalidArgument, err.Error())
+		return nil, nil, status.Error(codes.InvalidArgument, fmt.Sprintf("%s, %s", err.Error(), name))
```
begelundmuller, line 79:
```suggestion
		return nil, nil, status.Error(codes.InvalidArgument, fmt.Sprintf("error getting metrics view %q: %s", name, err.Error()))
```

runtime/metricsview/executor.go
```diff
@@ -189,6 +189,7 @@ func (e *Executor) Query(ctx context.Context, qry *Query, executionTime *time.Ti
 		if err != nil {
 			return nil, false, err
 		}
+		fmt.Println(sql)
```
begelundmuller, line 192: Remove

runtime/metricsview/astsql.go
```diff
@@ -74,7 +74,8 @@ func (b *sqlBuilder) writeSelectWithLabels(n *SelectNode) error {
 	return nil
 }
 
-func (b *sqlBuilder) writeSelect(n *SelectNode) error {
+func (b *sqlBuilder) writeSelect(n *SelectNode, level int) error {
+	b.out.WriteString(strings.Repeat("\t", level))
```
begelundmuller, line 78: Can we remove the pretty printing here? It's redundant extra length and allocations. If it's for debugging, there are many SQL formatters that can prettify a query (I use the `adpyke.vscode-sql-formatter` extension in VS Code)

begelundmuller (changes requested):

runtime/reconcilers/metrics_view.go
egor-ryashin: It took me several hours of debugging without those log messages, I don't understand. Reconciler and validation errors aren't shown if tests fail

runtime/testruntime/testruntime.go
egor-ryashin: I think tests are better than no tests

runtime/metricsview/astsql.go
egor-ryashin: Meaning additional manual work in exchange for nano-optimisation

runtime/reconcilers/metrics_view.go
begelundmuller: You just need to uncomment this line to get reconciler logs from tests: https://github.com/rilldata/rill/blob/1983fb1194d9f358eb5321534b15a4a824b0910c/runtime/testruntime/testruntime.go#L68

begelundmuller: Unless it's a very special circumstance, we should only read ENV variables at the entrypoint (e.g. CLI command or test function) and then pass config explicitly to internal packages, and we should use the Zap logger consistently instead of `fmt.Println`.

runtime/metricsview/astsql.go
begelundmuller: I'm more worried about the need for understanding, maintaining and debugging formatting code for no reason. There are plenty of decent SQL formatter libraries for Go if we have cases where we need to print approximately formatted SQL.

runtime/reconcilers/project_parser.go
egor-ryashin: I wonder if uncommenting ` // logger, err := zap.NewDevelopment() ` will show this error too?

runtime/testruntime/testruntime.go
egor-ryashin: `and it looks bad to open source contributors who don't get a chance to run or modify these tests` I will look even worse if get production Druid-based errors
`This both has the potential to overload a production cluster we rely on for more important use cases, makes it more difficult for folks internally to configure the tests` @nishantmonu51 I wonder if we should start a dedicated cluster for CI tests?

runtime/reconcilers/metrics_view.go
egor-ryashin: `Unless it's a very special circumstance, we should only read ENV variables at the entrypoint (e.g. CLI command or test function)` I doubt that, in terms of speed of development I would rather add everywhere 
```
if os.Getenv("DEBUG") == "1") {
fmt.Println(sql)
}
```
and forgot about adding/removing and checking `fmt.Println(sql)` all the time.

runtime/testruntime/testruntime.go
egor-ryashin: There's a dedicated container implemented but it was switched off for performance reason. 
"Regarding this PR, I think bigger changes to Druid testing improvements should be done in a dedicated PR."
I expected it will be an easy addition but it turns we want to postpone Druid tests again

runtime/reconcilers/metrics_view.go
egor-ryashin: No, I can keep DEBUG=1 in config on my local machine.

runtime/testruntime/testruntime.go
egor-ryashin: Does changing `metrics-in` to `master-in` cluster resolve the issue?

egor-ryashin:
CI tests part is moved here https://github.com/rilldata/rill/pull/5184

PR #5095, opened 2024-06-18
feat: Add annotations to identify dashboard settings in reports and alerts

AdityaHegde:
Some dashboard settings are not always clear from the query request.
1. For reports we need to know the active page it was created from. Queries from time dimension detail, dimension detail and pivot can be same. Adding `dashboard_active_page` to identify this.
2. For alerts measure filter added from the criteria section in the form will be added to the same place as measure filters selected in the dashboard. Adding `dashboard_measure_filter_indices` to identify which filters are from dashboard.


admin/server/alerts.go
```diff
@@ -471,6 +471,7 @@ func (s *Server) yamlForManagedAlert(opts *adminv1.AlertOptions, ownerUserID str
 	res.Annotations.AdminOwnerUserID = ownerUserID
 	res.Annotations.AdminManaged = true
 	res.Annotations.AdminNonce = time.Now().Format(time.RFC3339Nano)
+	res.Annotations.DashboardMeasureFilterIndices = opts.DashboardMeasureFilterIndices
```
begelundmuller, line 474: 1. Maybe use `MetricsView` as the prefix instead of `Dashboard` (now that we're introducing custom dashboards as well)
2. What will a value for this field look like? A little unclear on the purpose

admin/server/reports.go
```diff
@@ -452,6 +452,7 @@ func (s *Server) yamlForManagedReport(opts *adminv1.ReportOptions, ownerUserID s
 	res.Annotations.AdminManaged = true
 	res.Annotations.AdminNonce = time.Now().Format(time.RFC3339Nano)
 	res.Annotations.WebOpenProjectSubpath = opts.OpenProjectSubpath
+	res.Annotations.DashboardActivePage = opts.DashboardActivePage
```
begelundmuller, line 455: 1. Consider using `Web` prefix instead of `Dashboard` – e.g. maybe in the future, we have reports for non-dashboard pages in the UI
2. What value will this have? If it's an URL, could it be incorporated into `WebOpenProjectSubpath`?

AdityaHegde: Ya we could say `WebActivePage`. It is sadly not part of the subpath but stored in the base64 blob. I would assume the type MetricsView / Dashboard / Chart would be based on the query somehow. If not we could add that later on and `nil` would default to `MetricsView`

begelundmuller: Okay what about calling it `WebShowPage` or something like that then?

AdityaHegde: Ya that sounds good

begelundmuller (approved):

PR #5097, opened 2024-06-18
Admin: Fix prodVersion passed to UpdateDeployment and increase project timeouts

kaspersjo:


himadrisingh (approved):

admin/projects.go
```diff
@@ -185,7 +185,6 @@ func (s *Service) UpdateProject(ctx context.Context, proj *database.Project, opt
 	// It needs to be refactored when implementing preview deploys.
 	for _, d := range ds {
 		err := s.UpdateDeployment(ctx, d, &UpdateDeploymentOptions{
-			Version:         opts.ProdVersion,
```
begelundmuller, line 188: This feels a little weird. Can we maybe put `Version: ""` instead with a comment about why it's omitted? Or alternatively set `Version: d.Version` to clarify that it's being kept the same?

We've previously had bugs around unchanged values being mistaken for empty values. Explicitly listing all values in an `XXXOptions` type makes it easier to audit for such issues.

begelundmuller (changes requested):

admin/projects.go
kaspersjo: Sure - It's already omitted in the same way in `UpdateOrgDeploymentAnnotations`, so I'll update both places to use `Version: d.RuntimeVersion,` to clarify the intent.

begelundmuller (approved):
LGTM

PR #5103, opened 2024-06-19
Runtime: Fix session usage for clickhouse http interface

k-anshul:
Clickhouse http interface requires a `session_id` to be passed to tie queries to a certain session.

### DIFF
--- a/runtime/drivers/clickhouse/clickhouse.go
+++ b/runtime/drivers/clickhouse/clickhouse.go
@@ -166,6 +166,7 @@ func (d driver) Open(instanceID string, config map[string]any, client *activity.
 		if conf.SSL {
 			opts.Protocol = clickhouse.HTTP
 			opts.TLS = &tls.Config{
+				MinVersion:         tls.VersionTLS12,
 				InsecureSkipVerify: false,
 			}
 		} else {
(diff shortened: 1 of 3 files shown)


runtime/drivers/clickhouse/clickhouse.go
```diff
@@ -149,38 +147,44 @@ func (d driver) Open(instanceID string, config map[string]any, client *activity.
...
 	} else if conf.Host != "" {
-		var dsnURL url.URL
-		dsnURL.Host = conf.Host
+		opts = &clickhouse.Options{}
+
+		// address
+		host := conf.Host
 		if conf.Port != 0 {
-			dsnURL.Host = fmt.Sprintf("%v:%v", dsnURL.Host, conf.Port)
+			host = fmt.Sprintf("%v:%v", conf.Host, conf.Port)
```
begelundmuller, line 163: nit: `"%s:%d"`

runtime/drivers/clickhouse/clickhouse.go
```diff
@@ -149,38 +147,44 @@ func (d driver) Open(instanceID string, config map[string]any, client *activity.
...
 		}
+		opts.Addr = append(opts.Addr, host)
```
begelundmuller, line 165: Why `append` instead of `=`?

begelundmuller (approved):

runtime/drivers/clickhouse/clickhouse.go
k-anshul: Yes that can be used as well. I copied this from the underlying driver to build options so used the same append code.

PR #5110, opened 2024-06-20
limit preferredMinStreamCount equal to max connection pool size

k-anshul:
BigQuery changes : https://github.com/k-anshul/google-cloud-go/pull/5/files


esevastyanov (approved):
Looks good!

PR #5106, opened 2024-06-19
feat: login / signup in rill developer

AdityaHegde:
- [x] Add `GetCurrentUser` endpoint to get info similar to cloud.
- [x] Add `/auth/logout` endpoint for logout and to clear the token stored locally. Also logs out of cloud ui.
- [x] Add `LocalAvatarButton` that calls the above endpoints and is shown in dashboard (will be moved to use global header in a future PR)

This PR doesnt add pylon to rill developer, this will be in a separate PR

PR #5111, opened 2024-06-20
feat: Adding tanstack query wrapper around LocalService

AdityaHegde:
Splitting out tanstack wrapper around LocalService from #5106

### DIFF
--- a/web-common/src/runtime-client/local-service.ts
+++ b/web-common/src/runtime-client/local-service.ts
@@ -1,10 +1,6 @@
 import type { PartialMessage } from "@bufbuild/protobuf";
-import { createPromiseClient } from "@connectrpc/connect";
+import { type ConnectError, createPromiseClient } from "@connectrpc/connect";
 import { createConnectTransport } from "@connectrpc/connect-web";
-import type {
-  CreateMutationOptions,
-  CreateQueryOptions,
-} from "@rilldata/svelte-query";
 import { LocalService } from "@rilldata/web-common/proto/gen/rill/local/v1/api_connect";
 import {
   DeployRequest,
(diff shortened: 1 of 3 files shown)


web-common/src/runtime-client/local-service.ts
```diff
@@ -0,0 +1,163 @@
...
+import { LocalService } from "@rilldata/web-common/proto/gen/rill/local/v1/api_connect";
+import {
+  DeployRequest,
+  DeployValidationRequest,
+  GetMetadataRequest,
+  GetVersionRequest,
+  PushToGithubRequest,
+} from "@rilldata/web-common/proto/gen/rill/local/v1/api_pb";
+import type { RpcStatus } from "@rilldata/web-common/runtime-client/gen/index.schemas";
+import type { ErrorType } from "@rilldata/web-common/runtime-client/http-client";
```
begelundmuller, line 17: Shouldn't it use the Connect error types instead of the ones from Orval? See https://connectrpc.com/docs/web/errors

web-common/src/runtime-client/local-service.ts
```diff
@@ -0,0 +1,163 @@
+import type { PartialMessage } from "@bufbuild/protobuf";
+import { createPromiseClient } from "@connectrpc/connect";
+import { createConnectTransport } from "@connectrpc/connect-web";
+import type {
+  CreateMutationOptions,
+  CreateQueryOptions,
+} from "@rilldata/svelte-query";
```
begelundmuller, line 7: Shouldn't it use `@tanstack/svelte-query`? That is used below for `import { createMutation, createQuery }`

begelundmuller (commented):
I think it only works for React out of the box, but did you look into whether there's anything useful in https://github.com/connectrpc/connect-query-es?

begelundmuller (approved):

PR #5117, opened 2024-06-20
fix: reveal sub row dimension values in pivot without measures

### LINKED ISSUE #4637: Fix(Pivot): expanding a row does not reveal sub-rows when there are no columns
To reproduce:
1. Go to [this pivot table](https://ui.rilldata.com/demo/rill-openrtb-prog-ads/auction?state=CgUKA1AxVxgFKghyZXF1ZXN0czgASAFYAWAEagdFdGMvVVRDeAKAAQGYAf%2F%2F%2F%2F%2F%2F%2F%2F%2F%2F%2FwGiAQQaAggIsAEBugEBBMIBDWFwcF9zaXRlX25hbWXwAQH4AQKAAgSKAgdkZWZhdWx0)
2. Click to expand any row, like 00:00 or 01:00
3. See a skeleton loading state that doesn't resolve

image

djbarnwal:
The challenge with revealing sub-rows when there are no measures is that we really don't know what should come in those sub-rows. 
Example -
Dimension 1: Country
Values: India, USA, UK

Dimension 2: Cities
Values: Delhi, New York, London

Displaying a "cross-product" to get the sub-rows would give wrong results.

In case we use the first measure as a reference, depending on the measure type avg/sum/etc we can get confusing results.

The best solution probably is to use the inline `count(*)` as a measure when there is no measure. This PR implements this approach.

Closes #4637

### DIFF
--- a/web-common/src/features/dashboards/pivot/PivotTable.svelte
+++ b/web-common/src/features/dashboards/pivot/PivotTable.svelte
@@ -1,11 +1,14 @@
 <script lang="ts">
   import ArrowDown from "@rilldata/web-common/components/icons/ArrowDown.svelte";
+  import VirtualTooltip from "@rilldata/web-common/components/virtualized-table/VirtualTooltip.svelte";
   import { extractSamples } from "@rilldata/web-common/components/virtualized-table/init-widths";
   import { getMeasureColumnProps } from "@rilldata/web-common/features/dashboards/pivot/pivot-column-definition";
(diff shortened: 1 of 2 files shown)


ericpgreen2:
Functionally looks good. I like the `count(*)` approach.

Though, when I reveal a sub-row, I'm seeing two `Aggregation` requests. The second seems unnecessary?

First –
image

Second –
image

djbarnwal:
@ericpgreen2 This is ready for another pass

ericpgreen2 (approved):

PR #5121, opened 2024-06-20
fix: pretty AI generated yaml

mindspank:
* Omits ValidPercentOfTotal and FormatPreset for Expressions. FE uses humanize as default
* Omits Description for Dimensions
* Omits AvailableTimeRanges and TimeZones - FE has sensible defaults for TimeRanges.
* Inserts new lines before `title`, `dimensions` and `measures` (Semi-brittle as we rely on positioning)
* Aligns the order of properties so that dimension and measure items share similar order


mindspank:
Sample output

```yaml
# Dashboard YAML
# This file was generated using AI.

type: metrics_view

title: Suggested Metrics for Bids Data Model
model: bids_data_model
timeseries: __time

dimensions:
    - label: Is Beta
      column: is_beta
    - label: Ad Position
      column: ad_position

measures:
    - name: average_bid_price_usd
      label: Average Bid Price in USD
      expression: AVG(bid_price_usd)
      description: Calculating the average bid price in USD across all bids
    - name: total_bids
      label: Total Number of Bids
      expression: COUNT(bid_cnt)
      description: Counting the total number of bids made


```

runtime/server/generate_metrics_view.go
```diff
@@ -362,23 +358,19 @@ func generateMetricsViewYAMLSimpleDimensions(schema *runtimev1.StructType) []*me
...
 	for _, f := range schema.Fields {
 		switch f.Type.Code {
 		case runtimev1.Type_CODE_FLOAT32, runtimev1.Type_CODE_FLOAT64:
 			measures = append(measures, &metricsViewMeasureYAML{
-				Name:                f.Name,
-				Label:               fmt.Sprintf("Sum of %s", identifierToTitle(f.Name)),
-				Expression:          fmt.Sprintf("SUM(%s)", safeSQLName(f.Name)),
-				Description:         "",
-				FormatPreset:        "humanize",
-				ValidPercentOfTotal: true,
```
begelundmuller, line 381: I think it makes sense not to include `valid_percent_of_total` by default, but wondering if we should make it evaluate to `true` by default? Since it's true for most cases, and only slightly confusing for the cases where it's not.

If yes, we can make it such that it becomes `false` only if explicitly set to false in the YAML (and `true` if not present in the YAML at all)

runtime/server/generate_metrics_view.go
```diff
@@ -403,55 +395,32 @@ type metricsViewYAML struct {
...
-	Name                string
-	Label               string
-	Expression          string
-	Description         string
-	FormatPreset        string `yaml:"format_preset"`
-	ValidPercentOfTotal any    `yaml:"valid_percent_of_total"`
+	Label       string
+	Expression  string
+	Name        string
+	Description string
```
begelundmuller, line 406: Subjective, but I really like having `name` first – kind of like a variable declaration (name = props) or `id` attribute in HTML

runtime/server/generate_metrics_view.go
```diff
@@ -403,55 +395,32 @@ type metricsViewYAML struct {
...
-		"rill-TD",
-		"rill-WTD",
-		"rill-MTD",
-		"rill-QTD",
-		"rill-YTD",
-		"rill-PDC",
-		"rill-PWC",
-		"rill-PMC",
-		"rill-PQC",
-		"rill-PYC",
```
begelundmuller, line 452: Are these already the defaults in the UI? Or should we update the UI to use these as defaults if they are unset in the YAML?

runtime/server/generate_metrics_view.go
```diff
@@ -461,15 +430,16 @@ func marshalMetricsViewYAML(doc *metricsViewYAML, aiPowered bool) (string, error
 	}
 	buf.WriteString("\n")
 
-	enc := yaml.NewEncoder(buf)
-	enc.SetIndent(2)
```
begelundmuller, line 465: Can we keep two-space indents? It's also what we use for most (all?) YAML files in our own repo

runtime/server/generate_metrics_view.go
mindspank: Ye agree, I'll move both column and name back to first position

runtime/server/generate_metrics_view.go
mindspank: image
Default ranges

runtime/server/generate_metrics_view.go
nishantmonu51: Its hard to assume its true for all metrics by default as for avg it would give wrong information on the dashboards. We can apply a regex for count and sum on simple metrics to set it to true otherwise treat it as false. Lets do it on the platform side in a separate PR, where we also have the capability to parse the expression and determine true/false.

nishantmonu51 (approved):

Issue #4424, opened 2024-03-26
Feat(runtime): Support OpenAPI for custom APIs


begelundmuller:
Support specifying parameters and response types using OpenAPI, and expose the combined OpenAPI spec on `/v1/instances/{instance_id}/api/openapi`

nishantmonu51:
@esevastyanov : Lets make sure we get to this in this sprint. Atleast the openAPI part one. we can scope out the export and error handling ones.

PR #5134, opened 2024-06-24
Support OpenAPI for custom APIs

esevastyanov:
Closes https://github.com/rilldata/rill/issues/4424

Example of OpenAPI spec

```yaml
openapi:
  summary: "Hello API"
  request:
    parameters:
      - name: item_id
        in: query
        schema:
          type: integer
      - name: limit
        in: query
        schema:
          type: integer
      - name: api_key
        in: query
        schema:
          type: string
  response:
    schema:
      type: object
      properties:
        id:
          type: integer
        name:
          type: string
        price:
          type: number
          format: float
        in_stock:
          type: boolean
```


runtime/compilers/rillv1/parse_api.go
```diff
@@ -1,14 +1,64 @@
...
 )
 
 // APIYAML is the raw structure of a API resource defined in YAML (does not include common fields)
 type APIYAML struct {
 	DataYAML `yaml:",inline" mapstructure:",squash"`
+	OpenAPI  *OpenAPIYAML `yaml:"openapi"`
+}
+
+type OpenAPIYAML struct {
+	Info     map[string]any `yaml:"info"`
```
begelundmuller, line 19: Appears unused, can be removed

runtime/compilers/rillv1/parse_api.go
```diff
@@ -1,14 +1,64 @@
...
+
+type ReqSpecYAML struct {
+	Summary    string           `yaml:"summary"`
+	Parameters []map[string]any `yaml:"parameters"`
+}
+
+type RespSpecYAML struct {
+	Description string         `yaml:"description"`
+	Schema      map[string]any `yaml:"schema"`
+}
```
begelundmuller, line 32: I would consider flattening or reworking this slightly – a few comments:
1. The `summary` is not related to the request itself, but to the endpoint as a whole
2. We might consider omitting the response `description`. Unlike the full OpenAPI spec, the user here can only define a single response schema, so we should be able to autopopulate the description with something like "Successful resolver response".
3. The `ReqSpecYAML` and `RespSpecYAML` type names should maybe be prefixed by `OpenAPI` since the package is quite large, to avoid polluting the scope. Or maybe just be inline structs (see `AlertYAML` for an example)?

runtime/compilers/rillv1/parse_api.go
```diff
@@ -1,14 +1,64 @@
...
+	jsonData, err := json.Marshal(schema)
+	if err != nil {
+		return nil, err
+	}
+	err = specSchema.UnmarshalJSON(jsonData)
+	if err != nil {
+		return nil, err
+	}
+	return &specSchema, nil
 }
```
begelundmuller, line 62: Can we pull these into a separate OpenAPI utils package (e.g. `runtime/pkg/openapiutil`)? And maybe name them something like `MapToParameters` and `MapToSchema`. Alternatively, if kept in this package, they should be prefixed with `OpenAPI` to indicate that they're OpenAPI related (this is a big package).

proto/rill/runtime/v1/resources.proto
```diff
@@ -495,10 +495,18 @@ message API {
 message APISpec {
   string resolver = 1;
   google.protobuf.Struct resolver_properties = 2;
+  OpenAPISpec open_api_spec = 3;
```
begelundmuller, line 498: Consider just inlining the values here, for example:
```
string openapi_summary = 3;
repeated google.protobuf.Struct openapi_parameters = 4;
repeated google.protobuf.Struct openapi_response_schema = 5;
```

Nested types are nice in YAML, but less so in protos. And especially the name `OpenAPISpec` could be confused for a resource named "OpenAPI", as opposed to a sub-type of `APISpec`.

runtime/server/api.go
```diff
@@ -78,3 +84,204 @@ func (s *Server) apiHandler(w http.ResponseWriter, req *http.Request) error {
...
+		return httputil.Error(http.StatusInternalServerError, err)
+	}
+
+	return nil
+}
+
+func (s *Server) generateOpenAPISpec(instanceID string, apis map[string]*runtimev1.API) (*spec.Swagger, error) {
+	baseSpec := &spec.Swagger{
+		SwaggerProps: spec.SwaggerProps{
+			Swagger:  "2.0",
```
begelundmuller, line 172: Any reason for using v2 instead of v3?

runtime/server/api.go
```diff
@@ -78,3 +84,204 @@ func (s *Server) apiHandler(w http.ResponseWriter, req *http.Request) error {
...
+			Produces: []string{"application/json"},
+			Info: &spec.Info{
+				InfoProps: spec.InfoProps{
+					Title:   "Rill custom API of instance " + instanceID,
```
begelundmuller, line 176: Could we pull the org/project name from the instance annotations and incorporate those here? And if not found, default to just `Rill project API` or something like that?

runtime/server/api.go
```diff
@@ -78,3 +84,204 @@ func (s *Server) apiHandler(w http.ResponseWriter, req *http.Request) error {
...
+					Version: "1.0.0",
+				},
+			},
+			BasePath: "/",
```
begelundmuller, line 180: Could we put `/v1/instances/%s/api` here and then omit it from the individual routes?

runtime/server/api.go
```diff
@@ -78,3 +84,204 @@ func (s *Server) apiHandler(w http.ResponseWriter, req *http.Request) error {
...
+	pathItem := spec.PathItem{
+		PathItemProps: spec.PathItemProps{
+			Get: &spec.Operation{
+				OperationProps: spec.OperationProps{
+					Summary:    reqSummary,
+					Parameters: params,
+					Responses: &spec.Responses{
+						ResponsesProps: spec.ResponsesProps{
+							StatusCodeResponses: map[int]spec.Response{
+								200: {
```
begelundmuller, line 265: Maybe we should add the 400 case we return when a resolver fails here also?

begelundmuller (changes requested):

runtime/server/api.go
esevastyanov: Majorly, to align with project swagger specs and also there is no support of v3 in [go-openapi/spec](https://github.com/go-openapi/spec?tab=readme-ov-file#faq)
There is [go-openapi/spec3](https://github.com/go-openapi/spec3) but it is incomplete

begelundmuller: I think we should go for OpenAPI 3 if possible because it has some useful new features and has been out for a very long time. These libraries might be worth looking into:
- https://github.com/getkin/kin-openapi (used by https://github.com/oapi-codegen/oapi-codegen)
- https://github.com/ogen-go/ogen/tree/main/openapi (subpackage of https://github.com/ogen-go/ogen/tree/main)

begelundmuller (commented):
I think this PR looks great now, I think we should use OpenAPI v3 though (see my response to the last unresolved comment)

runtime/server/api.go
esevastyanov: Migrated to kin-openapi

begelundmuller (approved):

Issue #4490, opened 2024-04-02
Feat: Copy dimension values in TDD and Pivot


rillsidd:
**Describe the bug**
In time dimension detail view or in a pivot table, if i am comparing by any dimension, i can not hit shift+click a value to copy the dimension value to clipboard

**To Reproduce**
1. Go to https://ui.rilldata.com/demo/rill-cost-monitoring/customer_margin_dash
2. Click on 'Total Cost' metric to expand to time dimension detail view
3. Compare by 'Cost by App Name'
4. Try to copy any of the table values to clipboard (cannot do).

**Expected behavior**
User should be able to shift+click value to copy to clipboard. This is possible on any dimension table in the main dashboard view, just not in the time dimension detail.

mindspank:
Adding Pivot table to scope

mindspank:
@briangregoryholmes @nishantmonu51 looks like this slipped between sprints, moving it up to current sprint again so we dont miss it.

PR #5138, opened 2024-06-24
feat: add copy/tooltips to TDD table

briangregoryholmes:
This PR is a stopgap solution to enable copying of values via the TDD table. A larger refactor is necessary to remove the use of `RegularTable`, but this is a significant lift that is not currently prioritized.

This PR moves the responsibility for formatting from the data preparation utilities to the display components.

Tooltips are displayed using the native HTML `title` attribute rather than our bespoke `Tooltip` component as using the latter is not practical with Regular Table and the tooltip approach throughout the app is in flux.

Closes: #4490

### DIFF
--- a/web-common/src/features/dashboards/pivot/RegularTable.svelte
+++ b/web-common/src/features/dashboards/pivot/RegularTable.svelte
@@ -5,6 +5,7 @@
   import "./regular-table-style.css";
   import type { PivotPos, PivotRenderCallback } from "./types";
   import { isEmptyPos, range } from "./util";
+  import type { createMeasureValueFormatter } from "@rilldata/web-common/lib/number-formatting/format-measure-value";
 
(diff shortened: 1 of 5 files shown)


djbarnwal (commented):
Looks good for most parts. 

Getting a weird tooltip for null values 
image

web-common/src/features/dashboards/pivot/RegularTable.svelte
```diff
@@ -147,6 +143,8 @@
     const value = meta?.value;
     td.setAttribute("__col", String(x));
     td.setAttribute("__row", String(y));
+    console.log({ value });
```
djbarnwal, line 146: Stale log

djbarnwal (approved):
Looks good. 
There's a stale log in the changes. Approving so that you can merge post fixing that.

Issue #4969, opened 2024-05-28
Feat(journey): Deploy path without github connectivity for first time users


nishantmonu51:
https://www.notion.so/rilldata/First-time-browser-based-deploy-from-Rill-Developer-to-Rill-Cloud-with-deferred-Github-connection-d8f37e57723345f4a4c064ce65c37c1a

PR #5142, opened 2024-06-25
feat(deploy): UI based deploy flow

### LINKED ISSUE #5216: Push credentials from local to cloud as part of deploy flow
For connectors whose credentials are explicitly specified as part of the connector configuration in the .env file. We would like to push those credentials as part of the project deploy, In order to avoid any errors after the deployment. 

The admin would have the option to change the credentials via `rill env configure` and via admin UI in the future. 

This is a subtask of the bigger credential management.

AdityaHegde:
closes [#4969](https://github.com/rilldata/rill/issues/4969) closes #5216

Testing steps,
1. Create cli with this: `make cli`
2. Start cloud (needs docker running): `./rill devtool start cloud`
3. (Optional) Copy cli to home `cp ./rill ~/rill` and remove quarantine: `sudo xattr -d com.apple.quarantine ~/rill`
4. Run this within a new folder or existing folder: `~/rill start`

This will work for non-github and github based projects. Redeploy wont show for projects deployed before this.


ericokuma:
Hey Aditya! When I click on Deploy, I get a 500 error
Screenshot 2024-07-02 at 9 23 28 AM

cli/pkg/local/server.go
```diff
@@ -144,8 +145,63 @@ func (s *Server) DeployValidation(ctx context.Context, r *connect.Request[localv
...
+	repoAccess := false
+	var remote *gitutil.Remote
+	var ghURL string
+
+	if !hasProject || projectIsGithub {
+		// check if project is a git repo only if there is no project already, or it is connected to github
+		remote, ghURL, err = gitutil.ExtractGitRemote(s.app.ProjectPath, "", false)
+		if err != nil {
+			if errors.Is(err, git.ErrRepositoryNotExists) {
+				isGithubRepo = false
```
k-anshul, line 196: We can end up in a state where `isGithubRepo` is false but `githubRemoteFound` is true which doesn't seem correct.

cli/pkg/local/server.go
```diff
@@ -144,8 +145,63 @@ func (s *Server) DeployValidation(ctx context.Context, r *connect.Request[localv
 		return nil, err
 	}
 	var deployedProjectID string
+	hasProject := false
```
k-anshul, line 148: I think these two flags can be avoided by keeping a `var p *Project` directly ?

proto/rill/local/v1/api.proto
```diff
@@ -28,6 +28,8 @@ service LocalService {
 
   // User returns the locally logged in user
   rpc GetCurrentUser(GetCurrentUserRequest) returns (GetCurrentUserResponse) {}
+
+  rpc CheckOrgName(CheckOrgNameRequest) returns (CheckOrgNameResponse) {}
```
k-anshul, line 32: I don't know the full e2e flow from UI but for CLI we are mostly creating first and if conflict exist then prompting for project/org names. The org name can still be taken if this check passes plus it avoids extra network call.

k-anshul (changes requested):

cli/pkg/local/server.go
AdityaHegde: Hmm this code was moved around. But I agree it doesnt make sense.

proto/rill/local/v1/api.proto
AdityaHegde: What could be reason for org name to exist even if this passes? Race condition?

We do not have a flow to prompt user in UI. It has to automatically select the next available by adding `-1`, `-2` suffixes.

k-anshul: > What could be reason for org name to exist even if this passes? Race condition?

Yes

> It has to automatically select the next available by adding -1, -2 suffixes.

Then I would suggest lets create first and if it fails due to name mismatches then select the next available ?

web-local/src/routes/(misc)/deploy/+page.svelte
```diff
@@ -0,0 +1,60 @@
...
+      <CTAHeader variant="bold">
+        Hang tight! We're deploying your project...
+      </CTAHeader>
+    {:else if $deployerStatus.error}
+      <div class="h-36">
+        <CancelCircleInverse size="7rem" className="text-gray-200" />
+      </div>
+      <CTAHeader variant="bold">Oops! An error occurred</CTAHeader>
+      <CTAMessage>{$deployerStatus.error}</CTAMessage>
+      <CTAButton variant="secondary" on:click={onContactUs}>
```
ericpgreen2, line 54: Here too we should use `href` not `on:click` for links

web-common/src/features/project/ProjectDeployer.ts
```diff
@@ -0,0 +1,177 @@
...
+    await waitUntil(() => !get(this.validation).isFetching);
+    validation = get(this.validation).data as DeployValidationResponse;
+
+    if (!validation.isAuthenticated) {
+      void behaviourEvent?.fireDeployEvent(BehaviourEventAction.LoginStart);
+      window.open(
+        `${validation.loginUrl}/?redirect=${get(page).url.toString()}`,
+        "_self",
+      );
+      return false;
```
ericpgreen2, line 83: It initially surprised me to see the `validate()` function opens up a log-in page as a side effect. Might be more clear if you moved this to the `deploy()` function.

web-admin/vite.config.ts
```diff
@@ -16,9 +17,22 @@ const config: UserConfig = {
...
+function readPublicEmailDomains() {
+  const contents = readFileSync(
+    __dirname + "/../admin/pkg/publicemail/public_email_providers_list",
+  ).toString();
+  return contents
+    .split("\n")
+    .map((l) => l.trim())
+    .filter((l) => !l.startsWith("#"));
+}
+
```
ericpgreen2, line 37: Can we move this function elsewhere?

web-admin/src/routes/[organization]/[project]/-/invite/+page.svelte
```diff
@@ -13,6 +17,7 @@
 
   let allowDomain = false;
   $: userDomain = getUserDomain();
+  $: isPublicDomain = userDomainIsPublic();
   const addToAllowlist = createAdminServiceCreateProjectWhitelistedDomain();
```
Comment (ericpgreen2) [inline web-admin/src/routes/[organization]/[project]/-/invite/+page.svelte:21]: Missing error handling for these queries & mutation

web-common/src/features/project/ProjectDeployer.ts
```diff
@@ -0,0 +1,177 @@
...
+            this.promptOrgSelection.set(true);
+            return;
+          } else {
+            const userResp = await queryClient.fetchQuery({
+              queryKey: getLocalServiceGetCurrentUserQueryKey(),
+              queryFn: localServiceGetCurrentUser,
+            });
+            org = userResp.user!.displayName.replace(/ /g, "");
+            checkNextOrg = true;
+          }
```
ericpgreen2, line 129: With the `try/catch` and 3 nested `if` statements, this is pretty deep indentation. Consider if there's a way to avoid such nesting, e.g. with guard clauses or factoring out a `getOrg()` function.

ericpgreen2 (changes requested):
UXQA: the `/deploy` page's loading spinner flashes with a "Need help" message. See [Jam video](https://jam.dev/c/c01b71c3-74ba-4988-9c99-4487951ad0b8).

k-anshul (approved):

web-common/src/features/project/deploy-errors.ts
```diff
@@ -0,0 +1,27 @@
...
+    };
+  }
+  const match = RPCErrorExtractor.exec(error.message);
+  if (!match) {
+    if (error.message.includes("EntityTooLarge")) {
+      return {
+        message:
+          "It looks like you have more than 10GB. Contact us to finish deployment.",
+      };
+    }
```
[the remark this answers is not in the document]
AdityaHegde: Not right now. We can do a follow up to handle this in backend and send a consistent error upfront. I believe there is a way to pre calculate the size.

web-local/src/routes/(misc)/deploy/+page.svelte
AdityaHegde: This will open pylon similar to AvatarButton. Need pylon integration to local before this is merged.

ericpgreen2:
We should avoid this parse error, and anyways we should probably not show the `.rillcloud` directory in the file explorer. Can we add `.rillcloud` to the default `ignore_paths`? Or will that prevent the runtime from writing to the directory?

image

image

web-common/src/features/dashboards/workspace/DeployDashboardCTA.svelte
```diff
@@ -1,40 +1,95 @@
...
+  {:else}
+    <Button
+      loading={$deployValidation.isLoading}
+      on:click={onShowDeploy}
+      type="primary"
+      href={deployPageUrl}
+      target="_blank"
+    >
+      Redeploy
+    </Button>
```
ericpgreen2, line 58: These are reversed

web-admin/src/lib/vite-utils.ts
ericpgreen2, line 1: Sorry, I could have been more clear – this function should live with the "feature" that it relates to. Looks like it relates to the code in `features/projects/user-invite`

web-local/src/routes/(misc)/deploy/+page.svelte
ericpgreen2: Oh now I see. Given the event handler isn't yet implemented, we should hide the button.

web-common/src/features/project/OrgSelector.svelte
```diff
@@ -0,0 +1,32 @@
...
+  function selectHandler() {
+    open = false;
+    onSelect(selectedOrg);
+  }
+</script>
+
+<div class="text-xl">Select an organization</div>
+<div class="text-base text-gray-500">
+  Choose an organization to deploy this project to. <a
+    href="https://docs.rilldata.com/reference/cli/org/"
```
ericpgreen2, line 20: Nit: trailing slash

web-admin/src/routes/[organization]/[project]/-/invite/+page.svelte
```diff
@@ -1,59 +1,97 @@
...
+          type: "error",
+          message:
+            (e as AxiosError<RpcStatus>).response.data.message ?? e.message,
+          options: {
+            persisted: true,
+          },
+        });
+      }
     }
     return goto(`/${organization}/${project}/-/status`);
```
Comment (ericpgreen2) [inline web-admin/src/routes/[organization]/[project]/-/invite/+page.svelte:49]: We should only continue to the `/status` page if there's no error. Rather than show an ephemeral notification, we should show the error in red text below the "Continue" button.

web-local/src/routes/(misc)/deploy/+page.svelte
```diff
@@ -0,0 +1,61 @@
...
+    {:else if $deployerStatus.isLoading}
+      <div class="h-36">
+        <Spinner status={EntityStatus.Running} size="7rem" duration={725} />
+      </div>
+      <CTAHeader variant="bold">
+        Hang tight! We're deploying your project...
+      </CTAHeader>
+      <CTANeedHelp />
+    {:else if $deployerStatus.error}
+      <div class="h-36">
```
ericpgreen2, line 50: No need for this `h-36` – it adds funky extra space in this case

web-admin/src/routes/[organization]/[project]/-/invite/+page.svelte
Comment (AdityaHegde) [reply web-admin/src/routes/[organization]/[project]/-/invite/+page.svelte:49]: I disagree, error while adding a domain whitelist should not block the user from going forward. This can barely happen and there are no mocks for this. Lets do a follow up for this.

Comment (ericpgreen2) [reply web-admin/src/routes/[organization]/[project]/-/invite/+page.svelte:49]: Sounds good, I won't block on this. @ericokuma for resolution.

web-common/src/features/project/ProjectDeployer.ts
```diff
@@ -0,0 +1,179 @@
...
+        // since after successful deploy the same page is redirected to cloud we wont have an issue of the final loaded state
+        return {
+          isLoading: true,
+          error: undefined,
+        };
+      },
+    );
+  }
+
+  public async validate() {
```
ericpgreen2, line 64: I think this is a private function, not called externally.

ericpgreen2 (approved):
Approving from the frontend perspective. Great feature!

web-common/src/features/project/ProjectDeployer.ts
AdityaHegde: Ah the validate method is meant to read through DeployValidation and make sure it passes.

ericokuma (approved):
looks good

PR #5146, opened 2024-06-25
feat(deploy): Add .rillcloud/project.yaml to store deployed info

AdityaHegde:
We are storing these in `.rillcloud/project.yaml` to identify the deployed project,
```yaml
project_id: <uuid>
```

This will be used in UI deploy flow to correctly identify a redeploy.


cli/pkg/dotrillcloud/dotrillcloud.go
```diff
@@ -0,0 +1,45 @@
...
+	conf := &RillCloud{}
+	err = yaml.Unmarshal(data, conf)
+	if err != nil {
+		return nil, false, err
+	}
+	return conf, false, nil
+}
+
+func SetAll(localProjectPath string, conf *RillCloud) error {
+	err := os.MkdirAll(path.Join(localProjectPath, ".rillcloud"), os.ModePerm)
```
k-anshul, line 34: Should use `filepath.Join` with files.

cli/pkg/dotrillcloud/dotrillcloud.go
```diff
@@ -0,0 +1,45 @@
...
+	if err != nil {
+		return err
+	}
+
+	data, err := yaml.Marshal(conf)
+	if err != nil {
+		return err
+	}
+
+	return os.WriteFile(path.Join(localProjectPath, confPath), data, 0o644)
```
k-anshul, line 44: Same as above

cli/pkg/local/server.go
```diff
@@ -163,6 +164,16 @@ func (s *Server) DeployValidation(ctx context.Context, r *connect.Request[localv
 		}), nil
 	}
 
+	rc, ok, err := dotrillcloud.GetAll(s.app.ProjectPath)
+	if err != nil {
+		return nil, err
+	}
+	isDeployed := ok
+	deployedProjectID := ""
```
k-anshul, line 172: Can use `var deployedProjectID string` to avoid string allocation

cli/pkg/local/server.go
```diff
@@ -561,7 +583,22 @@ func (s *Server) RedeployProject(ctx context.Context, r *connect.Request[localv1
 		}
 		defer release()
 
-		assetID, err := cmdutil.UploadRepo(ctx, repo, s.app.ch, r.Msg.Org, r.Msg.ProjectName)
+		var projName, orgName string
+		if r.Msg.ProjectId != "" {
+			proj, err := c.GetProjectByID(ctx, &adminv1.GetProjectByIDRequest{
+				Id: projName,
```
k-anshul, line 589: This should be `r.Msg.ProjectId` and not `projName` ?

cli/pkg/dotrillcloud/dotrillcloud.go
```diff
@@ -0,0 +1,45 @@
...
+	"gopkg.in/yaml.v3"
+)
+
+type RillCloud struct {
+	ProjectID string `json:"project_id"`
+}
+
+const confPath = ".rillcloud/project.yaml"
+
+func GetAll(localProjectPath string) (*RillCloud, bool, error) {
```
k-anshul, line 16: The value of returned `bool` is always false. Is that expected ?
Also why do we need the bool ?

proto/rill/local/v1/api.proto
```diff
@@ -74,6 +74,8 @@ message DeployValidationResponse {
   bool rill_org_exists_as_github_user_name = 13; // only applicable when user does not have any orgs
   repeated string rill_user_orgs = 14;
   string local_project_name = 15;
+  bool is_deployed = 16;
+  string deployed_project_id = 17;
```
k-anshul, line 78: Could there be a case where `deployed_project_id` is set and yet `is_deployed` is false? 
Just trying to understand why `is_deployed` is separately required.

proto/rill/local/v1/api.proto
```diff
@@ -103,6 +105,7 @@ message DeployProjectResponse {
 message RedeployProjectRequest {
```
k-anshul, line 105: Given we now have project_id we should remove `org` and `project_name` ?

k-anshul (changes requested):

cli/pkg/local/server.go
AdityaHegde: Ah good catch

cli/pkg/dotrillcloud/dotrillcloud.go
AdityaHegde: No this is a remnant of some old code. Removing it

proto/rill/local/v1/api.proto
AdityaHegde: Hmmm. This was an attempt to keep it consistent with other fields. Not necessary since it is just a single value now.

proto/rill/local/v1/api.proto
AdityaHegde: True. I assumed it was used elsewhere. I guess it is not.

cli/cmd/deploy/deploy.go
```diff
@@ -293,6 +294,13 @@ func DeployFlow(ctx context.Context, ch *cmdutil.Helper, opts *Options) error {
 		return fmt.Errorf("create project failed with error %w", err)
 	}
 
+	err = dotrillcloud.SetAll(localProjectPath, ch.AdminURL, &dotrillcloud.RillCloud{
+		ProjectID: res.Project.Id,
+	})
+	if err != nil {
+		return err
+	}
```
begelundmuller, line 302: This might not be safe. I believe `localProjectPath` can be empty when deploying a Github remote directly. Maybe need to check `isLocalGitPath` before calling this?

cli/cmd/project/delete.go
```diff
@@ -33,8 +34,27 @@ func DeleteCmd(ch *cmdutil.Helper) *cobra.Command {
...
+					return fmt.Errorf("please provide a valid project name. Run `rill project list` to see the available projects")
+				}
+
+				proj, err := client.GetProjectByID(cmd.Context(), &adminv1.GetProjectByIDRequest{
+					Id: deployedID,
+				})
+				if err != nil {
+					return err
+				}
+				name = proj.Project.Name
```
begelundmuller, line 57: This can be reverted because `GetProjectByID` is already called in `InferProjectName` above

cli/pkg/dotrillcloud/dotrillcloud.go
```diff
@@ -0,0 +1,83 @@
...
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/rilldata/rill/admin/pkg/adminenv"
+	"gopkg.in/yaml.v3"
+)
+
+type RillCloud struct {
```
begelundmuller, line 12: nit: consider renaming to `dotrillcloud.Config`

admin/pkg/adminenv/admin_env.go
```diff
@@ -0,0 +1,29 @@
+package adminenv
```
begelundmuller, line 1: Can we move this package to `cli/pkg`? Since it relates to CLI configuration. (The admin service is, in theory, generic.)

begelundmuller (changes requested):

begelundmuller (approved):

k-anshul (approved):

PR #5148, opened 2024-06-25
Runtime: Retrofit metricsview package to MetricsView RPCs

begelundmuller:
Migrates `MetricsViewComparison`, `MetricsViewTotals`, `MetricsViewTimeSeries` and `MetricsViewToplist` to use `runtime/metricsview` for generating queries. This ensures they all support advanced measures and correct enforcement of filters embedded in security policies.


runtime/metricsview/astexpr.go
```diff
@@ -408,10 +408,14 @@ func (b *sqlExprBuilder) writeInCondition(left, right *Expression, leftOverride
 }
 
 func (b *sqlExprBuilder) writeInConditionForValues(left *Expression, leftOverride string, vals []any, not bool) error {
```
k-anshul, line 410: It will be great if we can include examples as unit tests for various if else cases handled here

runtime/queries/metricsview_aggregation.go
```diff
@@ -166,7 +164,7 @@ func (q *MetricsViewAggregation) Export(ctx context.Context, rt *runtime.Runtime
 	return nil
 }
 
-func (q *MetricsViewAggregation) rewriteToMetricsViewQuery() (*metricsview.Query, error) {
+func (q *MetricsViewAggregation) rewriteToMetricsViewQuery(export bool) (*metricsview.Query, error) {
```
k-anshul, line 167: Seemed okay to depend on `q.Export`?
Under what condition can `q.Exporting` be true now. I see it being set only from unit tests.

runtime/queries/metricsview_comparison_toplist.go
```diff
@@ -87,1207 +84,395 @@ func (q *MetricsViewComparison) Resolve(ctx context.Context, rt *runtime.Runtime
...
-// * the first query fetches only sorted dimensions
-// * the second query isn't run if the topN already less than the limit
-func (q *MetricsViewComparison) executeDruidApproximateToplist(ctx context.Context, olap drivers.OLAPStore, mv *runtimev1.MetricsViewSpec, priority int, security *runtime.ResolvedMetricsViewSecurity) error {
-	originalMeasures := q.Measures
-	if len(q.Measures) >= 5 {
-		originalMeasures = q.removeNoSortMeasures()
-	}
-
-	err := q.executeToplist(ctx, olap, mv, priority, security)
+	// Attempt to route to metricsview executor
```
k-anshul, line 92: Given we return error if that attempt fails we should remove the word Attempt

runtime/queries/metricsview_totals.go
```diff
@@ -111,59 +106,36 @@ func (q *MetricsViewTotals) Export(ctx context.Context, rt *runtime.Runtime, ins
...
-			mv:      mv,
-			dialect: dialect,
-		}
-		clause, clauseArgs, err := builder.buildExpression(q.Where)
-		if err != nil {
-			return "", nil, err
-		}
-		if strings.TrimSpace(clause) != "" {
-			whereClause += fmt.Sprintf(" AND (%s)", clause)
+	if q.Filter != nil { // backwards backwards compatibility
```
k-anshul, line 127: duplicate backwards ?

runtime/queries/metricsview_toplist.go
```diff
@@ -67,229 +65,167 @@ func (q *MetricsViewToplist) UnmarshalResult(v any) error {
...
 	case runtimev1.ExportFormat_EXPORT_FORMAT_PARQUET:
-		return WriteParquet(q.Result.Meta, q.Result.Data, w)
+		format = "parquet"
+	default:
+		return fmt.Errorf("unsupported format: %s", opts.Format.String())
 	}
 
-	return nil
-}
+	path, err := e.Export(ctx, qry, nil, format)
```
k-anshul, line 143: Why not  directly use `runtimev1.ExportFormat` as `parameter` type ?

k-anshul (approved):

runtime/queries/metricsview_aggregation.go
begelundmuller: Marked `q.Exporting` as deprecated (you're right, it's only used in unit tests to enable testing export formats – would have been nicer to actually call `Export` from the unit test).

runtime/queries/metricsview_comparison_toplist.go
begelundmuller: Fixed

runtime/queries/metricsview_totals.go
begelundmuller: Fixed

runtime/queries/metricsview_toplist.go
begelundmuller: It's an API message so would be slightly weird to pass into a model's `OutputProperties`. Just added a designated `drivers.FileFormat` type to use instead.

PR #5152, opened 2024-06-26
feat: Asset cleanup

### LINKED ISSUE #5144: Add a background task to cleanup assets bucket

k-anshul:
closes #5144

### DIFF
--- a/admin/admin.go
+++ b/admin/admin.go
@@ -43,7 +43,7 @@ type Service struct {
 	AutoscalerCron   string
 }
 
-func New(ctx context.Context, opts *Options, logger *zap.Logger, issuer *auth.Issuer, emailClient *email.Client, github Github, aiClient ai.Client, assetsHandle *storage.BucketHandle) (*Service, error) {
+func New(ctx context.Context, opts *Options, logger *zap.Logger, issuer *auth.Issuer, emailClient *email.Client, github Github, aiClient ai.Client, assetsBucket *storage.BucketHandle) (*Service, error) {
 	// Init db
 	db, err := database.Open(opts.DatabaseDriver, opts.DatabaseDSN)
 	if err != nil {
@@ -105,7 +105,7 @@ func New(ctx context.Context, opts *Options, logger *zap.Logger, issuer *auth.Is
 		VersionCommit:    opts.VersionCommit,
 		metricsProjectID: metricsProjectID,
 		AutoscalerCron:   opts.AutoscalerCron,
-		AssetsBucket:     assetsHandle,
+		AssetsBucket:     assetsBucket,
 	}, nil
 }
 
--- a/admin/admin.go
+++ b/admin/admin.go
@@ -27,17 +27,12 @@ type Options struct {
 }
 
 type Service struct {
-	DB             database.DB
-	ProvisionerSet map[string]provisioner.Provisioner
-	Email          *email.Client
-	Github         Github
-	AI             ai.Client
-	// Assets as reduced surface of storage.BucketHandle
-	// enables us to use a mock in tests.
-	Assets interface {
-		SignedURL(object string, opts *storage.SignedURLOptions) (string, error)
-		Object(path string) *storage.ObjectHandle
-	}
+	DB               database.DB
+	ProvisionerSet   map[string]provisioner.Provisioner
+	Email            *email.Client
+	Github           Github
+	AI               ai.Client
+	Assets           *storage.BucketHandle
 	Used             *usedFlusher
 	Logger           *zap.Logger
 	opts             *Options
--- a/admin/admin.go
+++ b/admin/admin.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 
+	"cloud.google.com/go/storage"
 	"github.com/rilldata/rill/admin/ai"
 	"github.com/rilldata/rill/admin/database"
 	"github.com/rilldata/rill/admin/provisioner"
@@ -31,6 +32,7 @@ type Service struct {
 	Email            *email.Client
 	Github           Github
 	AI               ai.Client
+	AssetsBucket     *storage.BucketHandle
 	Used             *usedFlusher
 	Logger           *zap.Logger
 	opts             *Options
@@ -41,7 +43,7 @@ type Service struct {
 	AutoscalerCron   string
 }
 
-func New(ctx context.Context, opts *Options, logger *zap.Logger, issuer *auth.Issuer, emailClient *email.Client, github Github, aiClient ai.Client) (*Service, error) {
+func New(ctx context.Context, opts *Options, logger *zap.Logger, issuer *auth.Issuer, emailClient *email.Client, github Github, aiClient ai.Client, assetsHandle *storage.BucketHandle) (*Service, error) {
 	// Init db
 	db, err := database.Open(opts.DatabaseDriver, opts.DatabaseDSN)
 	if err != nil {
@@ -103,6 +105,7 @@ func New(ctx context.Context, opts *Options, logger *zap.Logger, issuer *auth.Is
 		VersionCommit:    opts.VersionCommit,
 		metricsProjectID: metricsProjectID,
 		AutoscalerCron:   opts.AutoscalerCron,
+		AssetsBucket:     assetsHandle,
 	}, nil
 }
 
--- a/admin/admin.go
+++ b/admin/admin.go
@@ -102,6 +102,7 @@ func New(ctx context.Context, opts *Options, logger *zap.Logger, issuer *auth.Is
 		Email:            emailClient,
 		Github:           github,
 		AI:               aiClient,
+		Assets:           assets,
 		Used:             newUsedFlusher(logger, db),
 		Logger:           logger,
 		opts:             opts,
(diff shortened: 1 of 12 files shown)


admin/database/postgres/postgres.go
```diff
@@ -1580,6 +1580,46 @@ func (c *connection) InsertAsset(ctx context.Context, organizationID, path, owne
...
+	ON a.id = p.archive_asset_id 
+	WHERE p.archive_asset_id IS NULL`
+	var args []any
+	if createdBefore.IsZero() {
+		query += " ORDER BY a.created_on DESC LIMIT $1"
+		args = []any{limit}
+	} else {
+		query += " AND a.created_on < $1 ORDER BY a.created_on DESC LIMIT $2"
+		args = []any{createdBefore, limit}
+	}
```
begelundmuller, line 1595: 1. Consider dropping the `if` condition since `createdBefore` is always passed and is not a pointer
2. Consider using `WHERE NOT EXISTS (SELECT ...` instead of the `LEFT JOIN`. It seems slightly weird to join on a field that you have `IS NULL` on.

admin/database/postgres/migrations/0031.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX assets_created_on_idx ON assets (created_on);
```
begelundmuller, line 1: Of course always hard to say, but I don't think this index will be useful – indexes mostly only help if they significantly reduce the number of rows to scan. In this case:
1. Looking for assets not present in `projects` will likely lead to a full table scan anyway
2. Most assets will have been created before the `createdBefore`, so doesn't reduce rows scanned significantly

admin/worker/delete_unsued_assets.go
```diff
@@ -0,0 +1,80 @@
...
+	"net/url"
+	"strings"
+	"time"
+
+	"cloud.google.com/go/storage"
+	"go.uber.org/zap"
+	"golang.org/x/sync/errgroup"
+)
+
+const _defaultPageSize = 20
```
begelundmuller, line 15: 1. Maybe 100? 
2. Also, rename to something like `_unusedAssetsPageSize` – it is not a default (since never overridden), and the scope covers multiple files that do pagination for different things (remember Go's scope is not per file, but per package)

admin/worker/delete_unsued_assets.go
```diff
@@ -0,0 +1,80 @@
...
+
+func (w *Worker) deleteUnusedAssets(ctx context.Context) error {
+	// We skip unused assets created in last 15 minutes to prevent race condition
+	// where somebody just created an asset but is yet to use it
+	createdBefore := time.Now().Add(-15 * time.Minute)
```
begelundmuller, line 20: This is cutting it a little close (clocks can diverge, or we may change the upstream value and forget to change here). Maybe something like 6 hours?

admin/worker/delete_unsued_assets.go
```diff
@@ -0,0 +1,80 @@
...
+		// 1. Fetch unused assets
+		assets, err := w.admin.DB.FindUnusedAssets(ctx, createdBefore, _defaultPageSize)
+		if err != nil {
+			return err
+		}
+		if len(assets) == 0 {
+			return nil
+		}
+		createdBefore = assets[len(assets)-1].CreatedOn
+		ids := make([]string, len(assets))
```
begelundmuller, line 31: I would suggest simplifying the logic to always delete the full page fetched. And if any file delete fails with a different error than "not exists", can just return an error from the job (we will be alerted on it, and if the error was ephemeral, it will just be retried on next job invocation).

That avoids pagination by created on and will avoid values sticking around in the DB for too long.

begelundmuller (changes requested):

admin/admin.go
```diff
@@ -31,6 +32,7 @@ type Service struct {
 	Email            *email.Client
 	Github           Github
 	AI               ai.Client
+	AssetsBucket     *storage.BucketHandle
```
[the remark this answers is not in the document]
k-anshul: I can change it here to keep it consistent with other variable names. 
Overall I am skeptical on this practice. 
One argument given by Rob Pike to use capital names for exported variable in golang is that I don't need to look at variable's definition to determine whether its an exported or local. I would extend the argument to here that I don't need to look at the variable definition to determine what it does. Sadly `Email` or `AI` provides no such information.
The variable names should be somewhat self explanatory when used in such bigger contexts like `Service`.

admin/admin.go
```diff
@@ -103,6 +105,7 @@ func New(ctx context.Context, opts *Options, logger *zap.Logger, issuer *auth.Is
 		VersionCommit:    opts.VersionCommit,
 		metricsProjectID: metricsProjectID,
 		AutoscalerCron:   opts.AutoscalerCron,
+		AssetsBucket:     assetsBucket,
 	}, nil
```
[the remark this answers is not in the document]
k-anshul: I couldn't find any.

admin/database/postgres/postgres.go
k-anshul: 1. 
2. The `SELECT a.* FROM tblA  a JOIN tblB b ON a.id = b.id WHERE b.id IS NULL` is a very standard SQL statement to do `ANTI JOIN` IMO. However I see people seem to be preferring `WHERE NOT EXISTS` over `LEFT JOIN WHERE id IS NULL`. I can make the change.

admin/database/postgres/migrations/0031.sql
k-anshul: Good point.
1. Don't think it is true. When I did an explain on the query(`left join with id = null`), it was filtering the rows first and then joining the tables. 
2. Yes makes sense.

admin/worker/delete_unsued_assets.go
k-anshul: Sure. My thinking was that a few continuous failures can end up building lot of unused assets in the table and it will be good to keep deleting at least the successful deletes. 
When issuing 100 network calls it feels somewhat weird to only take further action when all 100 pass.

> will avoid values sticking around in the DB for too long.

No right if we return without deleting anything than all values stick in the DB till the next successful run.

admin/server/admin_rbac_test.go
```diff
@@ -59,6 +64,7 @@ func TestAdmin_RBAC(t *testing.T) {
 		emailClient,
 		github,
 		ai.NewNoop(),
+		client.Bucket("mock"),
```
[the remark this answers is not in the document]
k-anshul: The `storage.BucketHandle` is not exposed as interface. We need to create a custom interface that wraps `storage.BucketHandle` and add all used functions.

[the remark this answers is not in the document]
k-anshul: I evaluated `go cdk` but it has a gap that it requires passing a signing function when using signed URL to write a header and the default implementation only works when the sdk fetches creds from environment and not from service account json like we are doing for dev environment.
Also the CDK provides no extra benefit over a simple blob wrt signing URLs. The `mem` blob does not implement it and the file blob requires passing a signing implementation similar to a mock in this case.    

So I took some inspiration from the library wrt mocks and added an anonymous interface which can be mocked later when we are actually using the handle in tests.

admin/admin.go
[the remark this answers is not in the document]
begelundmuller: I think you may have misunderstood that argument. See the section on [variable names](https://google.github.io/styleguide/go/decisions#variable-names) in the style guide. It states:
> Omit types and type-like words from most variable names.
With the argument that:
> The compiler always knows the type of a variable, and in most cases it is also clear to the reader what type a variable is by how it is used.

IMO it makes sense and is in line with ideas of domain driven design. Basically, you name things for what they are used for, not their implementation details. In this case, it's a variable that gives you access to `Assets` – whether it's a bucket, or a directory handle, or an interface, does not need to be reflected in the name unless there is ambiguity.

admin/worker/delete_unsued_assets.go
begelundmuller: > When issuing 100 network calls it feels somewhat weird to only take further action when all 100 pass.

Normally I would agree, but in this case all the calls have the same type and go to the same service, so any other error than "not exists" should be a service-wide error, not a resource-specific error. So it is okay to keep it simple and just retry the whole batch at a later point.

admin/admin.go
```diff
@@ -26,11 +27,17 @@ type Options struct {
...
+	ProvisionerSet map[string]provisioner.Provisioner
+	Email          *email.Client
+	Github         Github
+	AI             ai.Client
+	// Assets as reduced surface of storage.BucketHandle
+	// enables us to use a mock in tests.
+	Assets interface {
+		SignedURL(object string, opts *storage.SignedURLOptions) (string, error)
+		Object(path string) *storage.ObjectHandle
+	}
```
begelundmuller, line 40: If you don't actually mock it in the test, let's just keep this as `*storage.BucketHandle`. If you add mocks, we should use a separate interface instead of an inline interface here.

begelundmuller (approved):

admin/admin.go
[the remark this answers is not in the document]
k-anshul: >  Basically, you name things for what they are used for, not their implementation details.

`AssetHandle` or `AssetsBucket` is not an implementation detail IMO. `AssetsGCSHandle` would be.
It gives a subtle hint about what they are used for. 
With just `Assets` I can't possible guess what it is used for. Is it a cache ?

Issue #5105, opened 2024-06-19
Runtime: Security policies refactors


begelundmuller:
- [x] Support "include all" like semantics for certain users ([thread](https://rilldata.slack.com/archives/CTZ8XBQ85/p1719395096051299?thread_ts=1719336471.788239&cid=CTZ8XBQ85))
- [x] Change security policy to `repeated SecurityRule rules`
- [x] Have rules like `AllowKind`, `AllowIf`, `AllowResource`, `TrimMetricsViewFields`, etc.
- [x] In `ResolvedMetricsViewSecurity`, the `QueryFilter` should be of type `metricsview.Expression`, not `runtimev1.Expression`
- [x] For non-admins, add security policy limiting access to exclude access to e.g. sources and models
- [x] Change alert/report authorization code to use security policies

PR #5157, opened 2024-06-27
Security policy rules

begelundmuller:
This PR introduces a general notion of "security rules" that can be evaluated not just for metrics views, but also for other resources. Initially, it supports three rules:
- `type: access`: conditional rule for whether access to a resource is allowed or denied
- `type: field_access`: conditional list of fields that are allowed or denied (initially only applied on metrics views)
- `type: row_filter`: conditional expression to add to SQL queries (initially only applied for metrics views)

Rules have the following properties:
- Have a `condition` expression that determines if the rule should be applied or not
- The `access` and `field_access` rules follow allow/deny semantics where explicit deny takes precedence over explicit allow. If any rule rule of the given type is present, the default is implicit deny, otherwise it is implicit allow.

The change to a chain of generic rules enables some much needed refactors, namely:
- Generic handling of security rules injected in the JWT token itself (as used in shareable URLs)
- Excluding certain resources from the APIs for viewer roles (e.g. sources, models, connectors)
- Using security policies for UI-managed alerts/reports instead of having hard-coded logic

**Other minor changes:**

- Adds support for `names: '*'` to indicate all names.
- Adds support for mixing include and exclude policies where excludes take precedence over includes.

**Example:**

This metrics view security policy:

```yaml
security:
  access: '{{ .user.domain }}' != 'mallory.com'
  row_filter: {{ if eq .user.domain 'rilldata.com' }} 1=1 {{ else }} domain = '{{ .user.domain }}' {{ end }}
  include:
    - names: '*'
  exclude:
    - if: '{{ .user.domain }}' != 'rilldata.com'
      names: ['domain', 'profit']
  ```

Will internally be equivalent to this chain of security rules:

```yaml
security:  
  rules:
    - type: access
      if: '{{ .user.domain }}' = 'mallory.com'
      action: deny
    - type: row_filter
      if: '{{ .user.domain }}' != 'rilldata.com'
      expression: domain = '{{ .user.domain }}'
    - type: field_access
      action: allow
      all: true
    - type: field_access
      if: '{{ .user.domain }}' != 'rilldata.com'
      action: deny
      names: ['domain', 'profit']
```

Closes https://github.com/rilldata/rill/issues/5105


proto/rill/runtime/v1/resources.proto
```diff
@@ -262,6 +244,46 @@ message MetricsViewSpec {
...
+  Expression expression = 3;
+}
+
+message SecurityCondition {
+  oneof condition {
+    string expression = 1;
+    string resource_kind = 2;
+    ResourceName resource_name = 3;
+  }
+}
```
begelundmuller, line 279: Just expression

proto/rill/runtime/v1/resources.proto
```diff
@@ -262,6 +244,46 @@ message MetricsViewSpec {
...
+message SecurityRuleAccess {
+  SecurityCondition condition = 1;
+  SecurityAction action = 2;
+}
+
+message SecurityRuleFieldAccess {
+  SecurityCondition condition = 1;
+  SecurityAction action = 2;
+  repeated string fields = 3;
+  bool all_except = 4;
```
begelundmuller, line 264:
```proto3
repeated string fields = 3;
bool all = 4;
bool exclude = 5;
```

runtime/compilers/rillv1/parse_metrics_view.go
```diff
@@ -284,6 +272,265 @@ func (f *MetricsViewMeasureWindow) UnmarshalYAML(v *yaml.Node) error {
...
+	}
+
+	switch r.Type {
+	case "access":
+		if allow == nil {
+			if r.If == "" {
+				return nil, fmt.Errorf("invalid security rule of type %q: must specify at least an action or a condition", r.Type)
+			}
+			tmp := true
+			allow = &tmp
```
pjain1, line 482: should we enforce specifying an action instead of assuming if not specified then allow is true ?

runtime/security.go
```diff
@@ -166,172 +103,199 @@ func (p *securityEngine) resolveMetricsViewSecurity(instanceID, environment stri
...
+				res, err := rillv1.EvaluateBoolExpression(expr)
+				if err != nil {
+					return nil, err
+				}
+				apply = res
 			}
-			incCond, err := rillv1.EvaluateBoolExpression(cond)
-			if err != nil {
-				return nil, err
+			if !apply {
```
pjain1, line 167: if condition is `false` or empty then fieldAccess will be empty and all fields will be allowed, should we explicitly ask what action they want here allow or deny ? also I think this is not backwards compatible behaviour. May be we enforce specifying an `action` in the YAML then we use the `allow` field here, I hope I am not mixing things up.

admin/server/projects.go
```diff
@@ -146,12 +147,30 @@ func (s *Server) GetProject(ctx context.Context, req *adminv1.GetProjectRequest)
...
+	var rules []*runtimev1.SecurityRule
 	if claims.OwnerType() == auth.OwnerTypeUser {
 		attr, err = s.jwtAttributesForUser(ctx, claims.OwnerID(), proj.OrganizationID, permissions)
 		if err != nil {
 			return nil, status.Error(codes.Internal, err.Error())
 		}
+
+		// Non-admins should be denied access to certain resource kinds
+		if !permissions.ManageProject {
+			rules = append(rules, &runtimev1.SecurityRule{
```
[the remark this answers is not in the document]
begelundmuller: They get dashboard access, so they can see if it is in an error state. But yeah, they can't see source or model details (which might contain privileged information – e.g. for a partner-filtered dashboard, there may be some confidential info in the underlying model SQL).

proto/rill/runtime/v1/resources.proto
```diff
@@ -262,6 +244,32 @@ message MetricsViewSpec {
...
+  string condition = 1;
+  bool allow = 2;
+  repeated string fields = 3;
+  bool all_fields = 4;
+}
+
+message SecurityRuleRowFilter {
+  string condition = 1;
+  string sql = 2; // Raw SQL expression to apply to the underlying table
+  Expression expression = 3; // Regular query expression referencing dimension names
```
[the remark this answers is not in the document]
begelundmuller: Yeah the problem is that for the shareable URLs, the UI constructs the filter (and uses dimension names in the filter, not the underlying column expressions).

runtime/compilers/rillv1/parse_metrics_view.go
begelundmuller: Good call, changed it to be required

runtime/security.go
begelundmuller: Actually it is the reverse, as soon as a field access rule is seen, even if it is false, all fields are implicitly denied until explicitly allowed. The only case for implicit allow is if there are no field access rules at all (i.e. no include/exclude properties). I believe this should be backwards compatible?

See this code: https://github.com/rilldata/rill/blob/36573348772a602dabf3ef5b0101323ce4a5393c/runtime/security.go#L148-L153

pjain1: ok got it, makes sense.

pjain1 (approved):

PR #5163, opened 2024-07-01
Cloud UI: Add page for shareable link via magic token

ericpgreen2:
This PR adds a new route `ui.rilldata.com/[organization]/[project]/-/share/[token]` to render a dashboard with the provided "magic" token (see https://github.com/rilldata/rill/pull/5071 for details on these magic tokens). The magic token grants the recipient of the link the ability to open a dashboard without needing to be an official member of the project.

Specifically, this PR:
- Centralizes logic for fetching & setting a project's runtime credentials in `[organization]/[project]/+layout.svelte`. Now, this `+layout.svelte` file sets credentials for logged-in users, magic link visitors, and mocked users (for whom we previously had a disjoint code path).
- Edits the `TopNavigationBar` for the magic-token shared links. We show only the dashboard title, not the full breadcrumbs.
- Switches out the Home icon for the Rill Logo icon on all pages. When users are not logged-in (which will often be the case for recipients of magic-token shared links), clicking on the Rill Logo will navigate to the Rill marketing site (`https://www.rilldata.com`). As discussed with @jkhwu and @ericokuma, this is an opportunity to give anonymous users (likely new users) more context about Rill.
- Adds a `Banner` component and a corresponding `BannerCenter` to manage its rendering.
- Uses a banner to direct project members to the "original" dashboard at `ui.rilldata.com/[organization]/[project]/[dashboard]`. The original dashboard has alerts, reports, and bookmarks available. The magic link's dashboard does not.

Here's what an anonymous user will see when visiting a magic-token shared link:
image

Here's what a user with project access will see when visiting one of these links:
image


ericokuma:
Looks good!

web-admin/src/routes/[organization]/[project]/+layout.svelte
```diff
@@ -1,31 +1,142 @@
+<script context="module" lang="ts">
+  const PollTimeWhenProjectDeploymentPending = 1000;
+  const PollTimeWhenProjectDeploymentError = 5000;
+  const PollTimeWhenProjectDeploymentOk = RUNTIME_ACCESS_TOKEN_DEFAULT_TTL / 2; // Proactively refetch the JWT before it expires
+
+  const baseGetProjectQueryOptions: QueryObserverOptions<
```
Comment (AdityaHegde) [inline web-admin/src/routes/[organization]/[project]/+layout.svelte:6]: Any reason to no move this to a .ts file?

AdityaHegde (commented):
Are magic link users supposed to have access to all resources? Same with schema for the current dashboard. @begelundmuller FYI
But even though I can see other dimensions here, I couldnt query them.
Screenshot 2024-07-03 at 12 03 06 PM
Here it is also exposing the rill managed project link.
Screenshot 2024-07-03 at 12 02 32 PM

Because of the list of resources returning all resources the title of the dashboard is incorrect in my case, since it takes the 1st entry in ListResources API.

begelundmuller:
> Are magic link users supposed to have access to all resources? Same with schema for the current dashboard. 

No, it's a general problem that security policies only apply to metrics views, e.g. viewers can also see non-dashboard related resources. I'm working on a refactor to security policies to fix this.

> @ericpgreen2 Because of the list of resources returning all resources the title of the dashboard is incorrect in my case, since it takes the 1st entry in ListResources API.

I think you just need to pass `kind` as metrics view to that API since the security policies are enforced for metrics view listings.

ericpgreen2:
> I think you just need to pass `kind` as metrics view to that API since the security policies are enforced for metrics view listings.

Oh, I found the bug: in the top bar, I was pulling the name from the list of visualizations, which included both metrics views and custom dashboards. Fixed.

web-admin/src/routes/[organization]/[project]/+layout.svelte
Comment (ericpgreen2) [reply web-admin/src/routes/[organization]/[project]/+layout.svelte:6]: I don't feel too strongly. It only relates to the Query Observers in this `+layout.svelte` file, so this felt like a middle-ground between code co-location & factoring it out.

AdityaHegde (approved):

PR #5167, opened 2024-07-01
Update slot definition and remove DuckDB driver cpuThreadRatio

kaspersjo:


runtime/drivers/duckdb/config.go
```diff
@@ -104,7 +103,7 @@ func newConfig(cfgMap map[string]any) (*config, error) {
 	if cfg.ThreadsOverride != 0 {
 		threads = cfg.ThreadsOverride
 	} else if cfg.CPU > 0 {
-		threads = int(cpuThreadRatio * float64(cfg.CPU))
+		threads = cfg.CPU
 		if threads <= 0 {
```
k-anshul, line 107: Can also remove this check fractional CPU is no longer possible now which would get translated to 0 previously ?

kaspersjo: True - removed

himadrisingh:
Lets keep it 1:4 for standard nodes. Highmem is costlier and will require more efforts in order to change the slots of all the current projects.

admin/provisioner/kubernetes.go
```diff
@@ -123,7 +123,7 @@ func (p *KubernetesProvisioner) Provision(ctx context.Context, opts *ProvisionOp
 		Names:        names,
 		Host:         strings.Split(host, "//")[1], // Remove protocol
 		CPU:          1 * opts.Slots,
-		MemoryGB:     2 * opts.Slots,
+		MemoryGB:     8 * opts.Slots,
```
himadrisingh, line 126:
```suggestion
		MemoryGB:     4 * opts.Slots,
```

admin/provisioner/static.go
```diff
@@ -72,7 +72,7 @@ func (p *StaticProvisioner) Provision(ctx context.Context, opts *ProvisionOption
 		Host:         target.Host,
 		Audience:     target.Audience,
 		CPU:          1 * opts.Slots,
-		MemoryGB:     2 * opts.Slots,
+		MemoryGB:     8 * opts.Slots,
```
himadrisingh, line 75:
```suggestion
		MemoryGB:     4 * opts.Slots,
```

himadrisingh (approved):

PR #5169, opened 2024-07-01
Runtime : Better export limits

k-anshul:
closes https://github.com/rilldata/rill-private-issues/issues/432

The excel export limit is not fully optimised. The entire export data is written into temporary files by the underlying library and the limit is only checked when writing data from these temporary files to output files. I checked and could not find any way to control these temporary files.

### DIFF
--- a/runtime/drivers/duckdb/model_executor_self_file.go
+++ b/runtime/drivers/duckdb/model_executor_self_file.go
@@ -71,8 +71,8 @@ func (e *selfToFileExecutor) Execute(ctx context.Context) (*drivers.ModelResult,
 						continue
 					}
 					if f.Size() > outputProps.FileSizeLimitBytes {
-						cancel()
 						overLimit.Store(true)
+						cancel()
 					}
 				}
 			}
@@ -94,12 +94,14 @@ func (e *selfToFileExecutor) Execute(ctx context.Context) (*drivers.ModelResult,
 	// check the size again since duckdb writes data with high throughput
 	// and it is possible that the entire file is written
 	// before we check size in background goroutine
-	f, err := os.Stat(outputProps.Path)
-	if err != nil {
-		return nil, err
-	}
-	if f.Size() > outputProps.FileSizeLimitBytes {
-		return nil, fmt.Errorf("file exceeds size limit %q", datasize.ByteSize(outputProps.FileSizeLimitBytes).HumanReadable())
+	if outputProps.FileSizeLimitBytes > 0 {
+		f, err := os.Stat(outputProps.Path)
+		if err != nil {
+			return nil, err
+		}
+		if f.Size() > outputProps.FileSizeLimitBytes {
+			return nil, fmt.Errorf("file exceeds size limit %q", datasize.ByteSize(outputProps.FileSizeLimitBytes).HumanReadable())
+		}
 	}
 
 	// Build result props
--- a/runtime/drivers/duckdb/model_executor_self_file.go
+++ b/runtime/drivers/duckdb/model_executor_self_file.go
@@ -3,7 +3,11 @@ package duckdb
 import (
 	"context"
 	"fmt"
+	"os"
+	"sync/atomic"
+	"time"
 
+	"github.com/c2h5oh/datasize"
 	"github.com/mitchellh/mapstructure"
 	"github.com/rilldata/rill/runtime/drivers"
 	"github.com/rilldata/rill/runtime/drivers/file"
@@ -47,13 +51,54 @@ func (e *selfToFileExecutor) Execute(ctx context.Context) (*drivers.ModelResult,
 		return nil, err
 	}
 
+	overLimit := atomic.Bool{}
+	if outputProps.FileSizeLimitBytes > 0 {
+		var cancel context.CancelFunc
+		// override the parent context
+		ctx, cancel = context.WithCancel(ctx)
+		defer cancel()
+		go func() {
+			ticker := time.NewTicker(time.Millisecond * 500)
+			defer ticker.Stop()
+			for {
+				select {
+				case <-ctx.Done():
+					return
+				case <-ticker.C:
+					f, err := os.Stat(outputProps.Path)
+					if err != nil { // ignore error since file may not be created yet
+						continue
+					}
+					if f.Size() > outputProps.FileSizeLimitBytes {
+						cancel()
+						overLimit.Store(true)
+					}
+				}
+			}
+		}()
+	}
+
 	err = olap.Exec(ctx, &drivers.Statement{
 		Query:    sql,
 		Args:     inputProps.Args,
 		Priority: e.opts.Priority,
 	})
 	if err != nil {
+		if overLimit.Load() {
+			return nil, fmt.Errorf("file exceeds size limit %q", datasize.ByteSize(outputProps.FileSizeLimitBytes).HumanReadable())
+		}
 		return nil, fmt.Errorf("failed to execute query: %w", err)
+	}
+
+	// check the size again since duckdb writes data with high throughput
+	// and it is possible that the entire file is written
+	// before we check size in background goroutine
+	f, err := os.Stat(outputProps.Path)
+	if err != nil {
+		return nil, err
+	}
+	if f.Size() > outputProps.FileSizeLimitBytes {
+		return nil, fmt.Errorf("file exceeds size limit %q", datasize.ByteSize(outputProps.FileSizeLimitBytes).HumanReadable())
 	}
 
 	// Build result props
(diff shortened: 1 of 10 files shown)


runtime/metricsview/executor.go
```diff
@@ -59,6 +60,7 @@ func NewExecutor(ctx context.Context, rt *runtime.Runtime, instanceID string, mv
 // Close releases the resources held by the Executor.
 func (e *Executor) Close() {
 	e.olapRelease()
+	_ = os.RemoveAll(e.tempDir)
```
begelundmuller, line 63: Unfortunately this is not safe because the executor might be closed before the caller is done with the export file. The `Export` function states `The caller is responsible for cleaning up the file.`

We probably need to handle deletion of the file in the individual error cases.

runtime/drivers/duckdb/model_executor_self_file.go
```diff
@@ -47,15 +51,56 @@ func (e *selfToFileExecutor) Execute(ctx context.Context) (*drivers.ModelResult,
...
+				case <-ctx.Done():
+					return
+				case <-ticker.C:
+					f, err := os.Stat(outputProps.Path)
+					if err != nil { // ignore error since file may not be created yet
+						continue
+					}
+					if f.Size() > outputProps.FileSizeLimitBytes {
+						cancel()
+						overLimit.Store(true)
```
begelundmuller, line 74: Swap these statements to reduce chance of a race condition where the `cancel()` call leads to `overLimit` being checked before it is stored.

runtime/drivers/duckdb/model_executor_self_file.go
```diff
@@ -47,15 +51,56 @@ func (e *selfToFileExecutor) Execute(ctx context.Context) (*drivers.ModelResult,
 		return nil, err
 	}
 
+	overLimit := atomic.Bool{}
```
begelundmuller, line 54:
```suggestion

	// Check the output file size does not exceed the configured limit.
	overLimit := atomic.Bool{}
```

runtime/drivers/duckdb/model_executor_self_file.go
```diff
@@ -47,15 +51,56 @@ func (e *selfToFileExecutor) Execute(ctx context.Context) (*drivers.ModelResult,
...
+	// check the size again since duckdb writes data with high throughput
+	// and it is possible that the entire file is written
+	// before we check size in background goroutine
+	f, err := os.Stat(outputProps.Path)
+	if err != nil {
+		return nil, err
+	}
+	if f.Size() > outputProps.FileSizeLimitBytes {
+		return nil, fmt.Errorf("file exceeds size limit %q", datasize.ByteSize(outputProps.FileSizeLimitBytes).HumanReadable())
+	}
```
begelundmuller, line 102: Only do this is `if outputProps.FileSizeLimitBytes > 0`

runtime/drivers/file/model_executor_olap_self.go
```diff
@@ -6,20 +6,24 @@ import (
...
 	"github.com/apache/arrow/go/v14/parquet/pqarrow"
+	"github.com/c2h5oh/datasize"
 	"github.com/mitchellh/mapstructure"
 	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
 	"github.com/rilldata/rill/runtime/drivers"
 	"github.com/rilldata/rill/runtime/pkg/jsonval"
 	"github.com/xuri/excelize/v2"
 )
 
+const maxParquetRowGroupLen = 4 * 1024 * 1024
```
begelundmuller, line 25: The Parquet docs recommend 512 MB, so this is probably too small: https://parquet.apache.org/docs/file-format/configurations/

runtime/drivers/file/model_executor_olap_self.go
```diff
@@ -328,22 +339,54 @@ func writeParquet(res *drivers.Result, path string) error {
...
-	return parquetwriter.Write(recordBuilder.NewRecord())
+func (l *limitedWriter) Write(p []byte) (n int, err error) {
+	if l.N <= 0 {
+		return 0, io.ErrShortWrite
+	}
+	truncated := false
+	if int64(len(p)) > l.N {
+		p = p[0:l.N]
+		truncated = true
+	}
```
begelundmuller, line 385: Why does it do truncation? Shouldn't it just error without writing at all?

runtime/drivers/file/model_executor_olap_self.go
```diff
@@ -328,22 +339,54 @@ func writeParquet(res *drivers.Result, path string) error {
 				recordBuilder.Field(i).(*array.BinaryBuilder).Append(v)
 			}
 		}
+		rows++
+		// We write the data to the underlying file to not buffer entire data in memory till it is written to parquet file.
+		// Since the check is not on memory but on number of rows this can still consume lot of memory
+		// so we set the limit to a much aggressive value than the default which is 64M.
+		if rows == maxParquetRowGroupLen {
```
begelundmuller, line 346: Re comment above – I think this is too aggressive. Have you looked into whether there are any settings/stats on the options on `NewFileWriter` and/or on `parquetwriter` that can be used to solve this? Maybe `RowGroupTotalCompressedBytes`? 

It's okay to go a row group size over in terms of memory usage. As long as we enforce the right limit on the export file, and we don't use unbounded memory.

begelundmuller (changes requested):

runtime/metricsview/executor.go
k-anshul: I had checked that in both cases i.e. downloads and pivot the file is consumed before executor is closed. 

The problem is that we also need to clear the directory path (`metrics_export*`) and not just the file otherwise we are left with many empty paths in `temp_dir`. Given the temp_dir only gets cleared on server restarts this can add up to a lot of empty directories which is not ideal. 
I can see following solutions to handle such situations:
1. The solution presented here. We can change the comment but I agree it is not ideal.
2. When the creator does not delete the file then the files should be added directly into the `temp_dir` or a unique directory across all calls and not a random child directory.
3. The caller to `executor.Export` passes a path. 

Please let me know what do you think. I think we should not leave empty directories. We can go with option 2.

runtime/drivers/file/model_executor_olap_self.go
k-anshul: The underlying writer uses a new row group if the number of rows is higher than 64M. I reduced it by a factor of 16 to 4M. 
The default for clickhouse exports is 1M and that for duckdb is just 122,880.

runtime/drivers/file/model_executor_olap_self.go
k-anshul: I copied it from k8s source code and I could think of two reasons why they write the remaining bytes first and then fail. 
1. To confirm to the definition that the `limitedWriter` limits the data written to N bytes.
2. To give it a chance to underlying writer to return error if any till N bytes.

runtime/drivers/file/model_executor_olap_self.go
k-anshul: The problem is not with the `parquetwriter`. The problem is that the entire data is kept in-memory in form of `arrow.Record` till it is written to the file using the writer. 
If you feel that our limits are too aggressive (though it is still higher than the defaults use by clickhouse and duckdb) we can use the following approach :
1. Write data to the file in batches of 1000 using the `WriteBuffered` API which keeps the data in-memory (I **think** there should not be a copy because it should use the underlying `arrow.Records` which is the main USP of arrow :) )
2. Keep checking the `RowGroupTotalCompressedBytes` to see what is the size of the data written so far and if it exceeds the limit throw an error.
3. If the limit is not exceed but the total bytes exceed limit of `512MB` create a new row group.

runtime/drivers/file/model_executor_olap_self.go
begelundmuller: Maybe those guarantees are important for how it is used in Kubernetes. Up to you, but since it is as a util writer in this specific package, I would remove it and save the extra writes as soon as we know the limit is breached.

runtime/drivers/file/model_executor_olap_self.go
begelundmuller: Got it, so this is the number of rows, not bytes

runtime/metricsview/executor.go
begelundmuller: By option 2, you basically mean outputting to `<temp dir>/metrics_export<rand>.<format>` instead of `<temp dir>/metrics_export<rand>/name.<format>`? I think that sounds good.

runtime/drivers/file/model_executor_olap_self.go
begelundmuller: I'm not too worried about the aggressive limit, though it would be nice to stick with the defaults, I'm more worried about the potential issues relating to `Since the check is not on memory but on number of rows this can still consume lot of memory`. We're moving away from row-based limits to byte-based limits due to those issues. Some people have very wide rows (especially in a pivoted response) that might have large binary fields.

k-anshul: Cool. In that case I think the solution proposed above can fix that issue.

begelundmuller (approved):

Issue #5045, opened 2024-06-10
Feat: UI-Based User Invitation


mindspank:
https://www.figma.com/board/EQiPmBWnt7zVzaywP0lzZx/UI-Based-User-Invitation?node-id=1-765&t=pja18tAWQQLkWBFA-1

https://www.notion.so/rilldata/UI-Based-User-Project-Invitation-Dashboard-Sharing-826f1d3615f14941935c1d57cf7d8df3?pvs=4

mindspank:
Product owner @ericokuma

AdityaHegde:
Related PRs: #5176 #5198

PR #5176, opened 2024-07-02
feat(deploy): UI based user invite

AdityaHegde:
closes [#5045](https://github.com/rilldata/rill/issues/5045)

This PR adds a popup for adding users.
Also adds basic code for invite after a deploy. The actual implementation might change when integrating with deploy flow.

Note that this doesnt support identifying public/private domains of the user.
This also doesnt add the `Request Access` flow or the sharable link flow through `copy link`


ericokuma:
Looks good though, this popover is appearing on the wrong page :)

Needs to appear as a popover from the Project Page e.g. https://ui.rilldata.com/demo/my-rill-project
Screenshot 2024-07-03 at 10 13 08 AM

web-admin/src/features/projects/user-invite/UserInviteForm.svelte
```diff
@@ -0,0 +1,73 @@
+<script lang="ts">
+  import { createAdminServiceAddProjectMember } from "@rilldata/web-admin/client";
+  import UserRoleSelect from "@rilldata/web-admin/features/projects/user-invite/UserRoleSelect.svelte";
+  import { Button } from "@rilldata/web-common/components/button";
+  import MultiInput from "@rilldata/web-common/components/forms/MultiInput.svelte";
+  import { createForm } from "svelte-forms-lib";
```
ericpgreen2, line 6: Now that @briangregoryholmes added [SuperForms](https://superforms.rocks/), we should migrate off `svelte-forms-lib` toward `sveltekit-superforms`. It can be a gradual migration. Currently, `RenameAssetModal` and `CreateShareableURLForm` both use SuperForms and can be inspiration.

AdityaHegde:
@ericokuma Thanks for the UX pass. I will update the location of the share button and the popover. Regarding the dropdown for role, where are the designs for it? I am referring to https://www.figma.com/design/Qt6EyotCBS3V6O31jVhMQ7/RILL?node-id=6801-232076&t=Dp2SPeOU2RW3lVMh-0 and they dont have the dropdown icon and they say `can view` (I extrapolated `can edit`)

web-admin/src/features/projects/user-invite/UserInviteForm.svelte
AdityaHegde: Updated to use SuperForms

ericokuma:
Ah @jkhwu, would you kindly update the mocks to include a version where there is a dropdown selector for role in the invite by email feature? This feature is only available for `project_admins` on accounts that have Connected to Github 

By Pills, I mean this:

Screenshot 2024-07-04 at 9 27 12 AM

jkhwu:
@AdityaHegde cc @ericokuma here I've shown the dropdown next to "can view". See the last variant and click into the component to see the hover and active states: https://www.figma.com/design/Qt6EyotCBS3V6O31jVhMQ7/RILL?node-id=13326-670655&t=jQKpt7XXDqz0oS2K-11

ericokuma:
1. The role should appear once user enters at least 1 character
Screenshot 2024-07-09 at 9 56 22 AM

2. Looks like there's some spacing issue from the pill the role dropdown selector
Screenshot 2024-07-09 at 9 57 11 AM

AdityaHegde:
Yes copy link needs this #5198 for it to fully work

web-admin/src/features/navigation/TopNavigationBar.svelte
```diff
@@ -168,7 +169,11 @@
       <ViewAsUserChip />
     {/if}
     {#if onProjectPage}
-      <ShareProjectButton {organization} {project} />
+      <ProjectAccessControls {organization} {project}>
+        <svelte:fragment slot="manage-project">
+          <UserInviteButton />
+        </svelte:fragment>
+      </ProjectAccessControls>
```
ericpgreen2, line 176: For your awareness, @briangregoryholmes kicked off another way to manage permissions in [this PR](https://github.com/rilldata/rill/pull/5193/files). It could be more scalable than this `slot` approach – one reason being `AND`/`OR` permission logic is tricky to do with these single-permission slots.

ericpgreen2:
UXQA –
1. Nothing happens when I submit an invalid email address. I'd expect to see an input validation warning.<br>image

2. The input box should not need to expand vertically after adding 1 character.

3. The input box should not be taller than the "Invite" button<br>image

4. > @jkhwu, the "Share this project" font is much larger than any other font on the page. Should we consider shrinking?

Agree.

5. This page does not match the Figma mocks.<br>image

AdityaHegde:
Addressed 1, 2 & 3. For 5, the page is just placeholder. Will update once i integrate with deploy flow.

ericokuma:
The only feedback here is that when a user clicks on "Invite", toast notification is missing.

Here's the requirement: https://www.notion.so/rilldata/RD-RC-1-5-UI-Based-User-Project-Invitation-Dashboard-Sharing-826f1d3615f14941935c1d57cf7d8df3?pvs=4#d442dd5e5c8e472081f075311338cf53

ericpgreen2:
UXQA –

1. When the form submission errors, we should show it in the dialog
    image
2. Emails are not included in the payload unless they've been submitted & converted to a pill. CC @ericokuma and @jkhwu for desired behavior. One solution (which we should do regardless) would be to disable the "Invite" button until at least 1 email has been officially added.

    These work:
    image
    image

    These don't work:
    image
    image

ericokuma (approved):

PR #5183, opened 2024-07-03
Admin: Fix lint error in assets job

begelundmuller:


k-anshul (approved):
Given this issue if fixed in latest go versions I wonder if we can add no lint error here till it is fixed in the linter ?

PR #5184, opened 2024-07-03
Druid CI tests

egor-ryashin:


runtime/testruntime/testruntime.go
```diff
@@ -224,11 +229,26 @@ func NewInstanceForProject(t TestingT, name string) (*runtime.Runtime, string) {
...
 
 	_, currentFile, _, _ := goruntime.Caller(0)
 	projectPath := filepath.Join(currentFile, "..", "testdata", "ad_bids_druid")
+	var err error
+	creds := os.Getenv("METRICS_CREDS")
+	address := os.Getenv("METRICS_ADDRESS")
+	data, err := base64.StdEncoding.DecodeString(creds)
+	if err != nil {
+		return nil, "", err
+	}
```
begelundmuller, line 243: Can we replace this with a single env var `RILL_RUNTIME_DRUID_TEST_DSN`? And save it to our shared `.env` file so everyone on the team is able to run it (see ` rill devtool dotenv -h`)

begelundmuller (changes requested):

runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go
```diff
@@ -40,187 +55,286 @@ type sqlConnection struct {
...
+var _ retrier.AdditionalTest = &coordinatorHTTPCheck{}
+
+// if Coordinator is down for a transient reason it's not a hard failure
+func (chc *coordinatorHTTPCheck) IsHardFailure(ctx context.Context) (bool, error) {
+	dr := newDruidRequest("SELECT * FROM sys.segments LIMIT 1", nil)
 	b, err := json.Marshal(dr)
 	if err != nil {
-		return nil, err
+		return false, err
 	}
```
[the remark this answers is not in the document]
egor-ryashin: Only on errors, added a more elaborate description.

runtime/drivers/druid/olap.go
```diff
@@ -196,7 +196,15 @@ func (i informationSchema) All(ctx context.Context) ([]*drivers.Table, error) {
...
 
 func (i informationSchema) Lookup(ctx context.Context, db, schema, name string) (*drivers.Table, error) {
-	q := `
+	// ensure Coordinator is ready
+	q := "SELECT * FROM sys.segments LIMIT 1"
+	rows, err := i.c.db.QueryxContext(ctx, q, name)
+	if err != nil {
+		return nil, err
+	}
+	rows.Close()
```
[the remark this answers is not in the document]
egor-ryashin: It's only on Lookup query and it's negligible compared to the whole amount of other requests.
The issues is that the request
```
		SELECT
			T.TABLE_SCHEMA AS SCHEMA,
			T.TABLE_NAME AS NAME,
			T.TABLE_TYPE AS TABLE_TYPE, 
			C.COLUMN_NAME AS COLUMN_NAME,
			C.DATA_TYPE AS COLUMN_TYPE,
			C.IS_NULLABLE = 'YES' AS IS_NULLABLE
		FROM INFORMATION_SCHEMA.TABLES T 
		JOIN INFORMATION_SCHEMA.COLUMNS C ON T.TABLE_SCHEMA = C.TABLE_SCHEMA AND T.TABLE_NAME = C.TABLE_NAME
		WHERE T.TABLE_SCHEMA = 'druid' AND T.TABLE_NAME = ?
		ORDER BY SCHEMA, NAME, TABLE_TYPE, C.ORDINAL_POSITION
```
returns false-negative if the Coordinator is being restarted. Retrier is a more abstract class and it doesn't check if SQL tries to retreive dynamic schema and the will be no error from Druid Router (because if dynamic schema is empty - it's considered OK by the cluster).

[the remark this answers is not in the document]
egor-ryashin: added a description.

runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go
```diff
@@ -40,187 +55,286 @@ type sqlConnection struct {
...
-		}
-
-		resp, err := c.client.Do(r)
-		if err != nil {
-			return
-		}
-		resp.Body.Close()
-	})
-	req, err := http.NewRequestWithContext(ctx, http.MethodPost, c.dsn, bodyReader)
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, chc.c.dsn, bodyReader)
```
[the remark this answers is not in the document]
egor-ryashin: let's do that in a separate PR

runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go
```diff
@@ -40,187 +55,286 @@ type sqlConnection struct {
...
+
+			return nil, retrier.Retry, err
+		}
+
+		switch resp.StatusCode {
+		case http.StatusTooManyRequests:
+			return nil, retrier.Retry, fmt.Errorf("Too many requests")
+		case http.StatusUnauthorized, http.StatusForbidden:
+			return nil, retrier.Fail, fmt.Errorf("Unauthorized request")
+		}
```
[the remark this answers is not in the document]
egor-ryashin: I've added `defer` in the other request. Here the body should go to a handler that will parse it outside of the function so that we can support `drivers.Rows.Next()` semantics.

runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go
```diff
@@ -40,187 +55,286 @@ type sqlConnection struct {
...
 	if err != nil {
-		return nil, err
+		return false, err
+	}
+
+	switch resp.StatusCode {
+	case http.StatusTooManyRequests:
+		return false, nil
+	case http.StatusUnauthorized, http.StatusForbidden:
+		return true, fmt.Errorf("Unauthorized request")
```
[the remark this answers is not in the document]
egor-ryashin: thanks

runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go
```diff
@@ -40,187 +55,286 @@ type sqlConnection struct {
...
-func (c *sqlConnection) QueryContext(ctx context.Context, query string, args []driver.NamedValue) (driver.Rows, error) {
-	dr := newDruidRequest(query, args)
+type coordinatorHTTPCheck struct {
+	c *sqlConnection
+}
+
+var _ retrier.AdditionalTest = &coordinatorHTTPCheck{}
+
+// if Coordinator is down for a transient reason it's not a hard failure
+func (chc *coordinatorHTTPCheck) IsHardFailure(ctx context.Context) (bool, error) {
```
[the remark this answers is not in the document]
egor-ryashin: `coordinatorHTTPCheck` is used right below

runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go
```diff
@@ -40,193 +55,285 @@ type sqlConnection struct {
 
 var _ driver.QueryerContext = &sqlConnection{}
 
-func (c *sqlConnection) QueryContext(ctx context.Context, query string, args []driver.NamedValue) (driver.Rows, error) {
-	dr := newDruidRequest(query, args)
+type coordinatorHTTPCheck struct {
+	c *sqlConnection
+}
+
+var _ retrier.AdditionalTest = &coordinatorHTTPCheck{}
```
begelundmuller, line 62: Move to later in the file – style guide says to sort functions in call order with exported functionality first (i.e. `QueryContext` first)

runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go
```diff
@@ -40,193 +55,285 @@ type sqlConnection struct {
...
-					case string:
-						return strconv.ParseFloat(v, 64)
-					default:
-						return v, nil
-					}
+		req.Header.Add("Content-Type", "application/json")
+		resp, err := c.client.Do(req)
+		if err != nil {
+			// nolint:errorlint // there's no wrapping
+			if v, ok := err.(*url.Error); ok {
```
begelundmuller, line 168: We don't know how the library will change in the future. Better to use `errors.As` which is created for this.

runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go
```diff
@@ -40,193 +55,285 @@ type sqlConnection struct {
...
+				if errTooManyRedirects.MatchString(v.Error()) {
+					resp.Body.Close()
```
begelundmuller, line 170: This may panic – accessing the body when the request failed. 

It's a problem multiple places, I would serialize consider if this can be refactored to use only `defer`.

runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go
```diff
@@ -40,193 +55,285 @@ type sqlConnection struct {
...
+
+			types := toStringArray(obj.([]any))
+
+			transformers := make([]func(any) (any, error), len(columns))
+			for i, c := range types {
+				transformers[i] = createTransformer(c)
+			}
+
+			druidRows := &druidRows{
+				closer:        resp.Body,
```
begelundmuller, line 247: Since `dec.Decode` has already been called on a JSON decoder that uses the body, won't it be fully consumed at this point? So it can just be closed immediately.

runtime/testruntime/testruntime.go
```diff
@@ -224,11 +228,21 @@ func NewInstanceForProject(t TestingT, name string) (*runtime.Runtime, string) {
...
 	rt := New(t)
 
 	_, currentFile, _, _ := goruntime.Caller(0)
 	projectPath := filepath.Join(currentFile, "..", "testdata", "ad_bids_druid")
+	var err error
+	b64 := os.Getenv("RILL_RUNTIME_DRUID_TEST_DSN")
+	dsn, err := base64.StdEncoding.DecodeString(b64)
+	if err != nil {
+		return nil, "", err
+	}
```
begelundmuller, line 245: Why does it need to be base64 encoded? We have other env variables containing DSNs that are not.

begelundmuller (changes requested):

runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go
egor-ryashin: > Decode reads the next JSON-encoded value from its input and stores it in the value pointed to by v.

It's NDJSON format, ie:
```
{"name": "Leandro","lastName": "Shokida"}
{"name": "Mariano","lastName": "De Achaval"}
```

runtime/testruntime/testruntime.go
egor-ryashin: It contains password in format https://username:password@hostname...
Bash has special symbols that the password can contain, to reduce the mental burden remembering which ones, I just coded it into base64.

runtime/testruntime/testruntime.go
```diff
@@ -224,11 +228,21 @@ func NewInstanceForProject(t TestingT, name string) (*runtime.Runtime, string) {
 	return rt, inst.ID
 }
 
-func NewInstanceForDruidProject(t TestingT) (*runtime.Runtime, string) {
+func NewInstanceForDruidProject(t *testing.T) (*runtime.Runtime, string, error) {
+	if os.Getenv("RILL_RUNTIME_DRUID_TEST_DSN") == "" {
+		t.Skip("skipping the test without the test instance")
+	}
```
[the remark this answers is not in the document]
egor-ryashin: The test are run when I commit the code if they pass it's correct.

runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go
egor-ryashin: Thanks. But `defer` won't work - see next comment.

runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go
egor-ryashin: It didn't complain when CI tests were run. Shouldn't we skip it - it's looks good only from one side but it contradicts the rule - place dependent parts closer together to avoid scrolling back and forth.

begelundmuller: It's not a lint rule, but described in the style guide here: https://github.com/uber-go/guide/blob/master/style.md#function-grouping-and-ordering.

I think it's a good rule to sort declarations by call order – it doesn't feel intuitive to look at a Druid driver and the first 100+ lines of code are internal network retrying logic.

runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go
begelundmuller: Got it, makes sense then

runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go
begelundmuller: Okay, got it. In that case, it's worth considering if things can be re-organized in such a way that `resp.Body.Close` needs to be manually called in as few cases as possible

runtime/testruntime/testruntime.go
begelundmuller: I thought bash/env variables can contain any ASCII string? Just need to escape it or use single quotes to enter it. 

Given that all other environment variables in the project do not use base64, I think we should avoid it here as well.

begelundmuller (commented):
I left some follow-up comments on various previous review comments. Please take a look as some of them require some more changes.

runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go
egor-ryashin: But otherwise you'll see the struct you don't know and have to scroll.

runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go
egor-ryashin: I mean `error.As` returns a bool but then after that I have to get the error message.
I agree that the library implementation can change but it will be incompatible changes for all clients (which is rare and a bad practice by the library). But anyway, if they change they can wrap it double time and this code fails again. So I should Unwrap recursively - which looks weird and clumsy for error handling.
So for example what can happen if the error is wrapped double time:

```go
type Er1 struct {
}

func (*Er1) Error() string {
	return "er1"
}

type wrapError struct {
	msg string
	err error
}

func (w *wrapError) Error() string {
	return w.msg
}

func (w *wrapError) Unwrap() error {
	return w.err
}

func ret() error {
	var e error
	e = &Er1{}
	e = &wrapError{
		msg: "extra info",
		err: e,
	}
	e = &wrapError{
		msg: "superextra info",
		err: e,
	}
	return e
}

var _ error = &Er1{}

func TestError(t *testing.T) {
	e := ret()
	e0 := &Er1{}
	require.True(t, errors.As(e, &e0))
	require.Equal(t, "er1", errors.Unwrap(e).Error())
}
```
```
        	            	expected: "er1"
        	            	actual  : "extra info"
```

runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go
egor-ryashin: Let's think about it in another PR.

runtime/testruntime/testruntime.go
egor-ryashin: Let's try it.

runtime/testruntime/testruntime.go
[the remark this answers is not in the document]
egor-ryashin: So do you propose to `godotenv.Load() ` before a test run?

runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go
egor-ryashin: Moreover I believe if the auto-formatter cannot do it then it's a maintenance burden.

egor-ryashin: Changed.

runtime/testruntime/testruntime.go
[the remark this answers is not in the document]
egor-ryashin: Updated `.env`

runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go
begelundmuller: `errors.As` doesn't just return a bool, it also assigns the error – it handles the recursive unwrapping automatically.

runtime/testruntime/testruntime.go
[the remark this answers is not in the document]
begelundmuller: Cool, did you check that it loads from the repo root and not the `testruntime` folder? Might need to change some code to have `godotenv` load from the right directory

runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go
egor-ryashin: Changed.

runtime/testruntime/testruntime.go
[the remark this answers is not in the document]
egor-ryashin: Not sure what you mean, do you suggest to load .env when `NewInstanceForDruidProject` is called?

runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go
```diff
@@ -40,193 +56,223 @@ type sqlConnection struct {
...
+	return nil, fmt.Errorf("unsupported")
+}
 
-		resp, err := c.client.Do(r)
+func (c *sqlConnection) QueryContext(ctx context.Context, query string, args []driver.NamedValue) (driver.Rows, error) {
+	// total sum is 126 seconds (sum(2*2^x) from 0 to 5 inclusive)
+	re := retrier.NewRetrier(6, 2*time.Second, &coordinatorHTTPCheck{
+		c: c,
+	})
+	return re.RunCtx(ctx, func(ctx2 context.Context) (driver.Rows, retrier.Action, error) {
```
begelundmuller, line 79: Is it on purpose that most calls use `ctx` instead of `ctx2`? What is the implication of using one vs. the other?

runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go
```diff
@@ -40,193 +56,223 @@ type sqlConnection struct {
...
-				transformers[i] = func(v any) (any, error) {
-					var l map[string]any
-					err := json.Unmarshal([]byte(v.(string)), &l)
-					if err != nil {
-						return nil, err
-					}
-					return l, nil
+			}
+
+			return nil, retrier.Retry, err
```
begelundmuller, line 131: Does it make sense to retry when we don't even get a response back? For example, might just be the DSN that's broken or a local firewall, it's not great to have 126 seconds of retrying just to discover that.

Should we maybe opt for just retrying on the retryable HTTP status codes?

begelundmuller (changes requested):

runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go
egor-ryashin: Makes sense.

runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go
egor-ryashin: Removed.

runtime/testruntime/testruntime.go
```diff
@@ -224,11 +230,20 @@ func NewInstanceForProject(t TestingT, name string) (*runtime.Runtime, string) {
 	return rt, inst.ID
 }
 
-func NewInstanceForDruidProject(t TestingT) (*runtime.Runtime, string) {
+func NewInstanceForDruidProject(t *testing.T) (*runtime.Runtime, string, error) {
+	_, err := os.Stat(".env")
```
begelundmuller, line 234: I believe Go tests always evaluate relative to the test file, not the repo root. So you probably need to do something similar to this to get the path of the root `.env` file: https://github.com/rilldata/rill/blob/d66a28ff94b55273d102cc2112faf8d05465d254/runtime/testruntime/testruntime.go#L181-L182

Please test and confirm that the approach works

runtime/testruntime/testruntime.go
```diff
@@ -224,11 +230,20 @@ func NewInstanceForProject(t TestingT, name string) (*runtime.Runtime, string) {
...
+	if !errors.Is(err, fs.ErrNotExist) {
```
begelundmuller, line 235: There can be other errors than file not existing – it would probably be better to do `if err == nil`, which is the only way to know that the file actually exists

runtime/testruntime/testruntime.go
```diff
@@ -224,11 +230,20 @@ func NewInstanceForProject(t TestingT, name string) (*runtime.Runtime, string) {
...
+		require.NoError(t, godotenv.Load())
```
begelundmuller, line 236: Related to the comment above – you may need to pass the repo root path here for this to work

begelundmuller (changes requested):

egor-ryashin:
Done.

runtime/testruntime/testruntime.go
```diff
@@ -224,11 +228,22 @@ func NewInstanceForProject(t TestingT, name string) (*runtime.Runtime, string) {
 	return rt, inst.ID
 }
 
-func NewInstanceForDruidProject(t TestingT) (*runtime.Runtime, string) {
+func NewInstanceForDruidProject(t *testing.T) (*runtime.Runtime, string, error) {
+	_, currentFile, _, _ := goruntime.Caller(0)
+	envPath := filepath.Join(currentFile, "..", "..", ".env")
```
begelundmuller, line 233: Won't this resolve to `<repo>/runtime/.env`, not `<repo>/.env`? Can you test any changes to confirm?

begelundmuller (changes requested):

runtime/testruntime/testruntime.go
egor-ryashin: Fixed.

begelundmuller (approved):

PR #5185, opened 2024-07-03
Runtime : health check rpc api

k-anshul:


proto/rill/runtime/v1/api.proto
```diff
@@ -15,6 +15,10 @@ service RuntimeService {
     option (google.api.http) = {get: "/v1/ping"};
   }
 
+  rpc Healthz(HealthzRequest) returns (HealthzResponse) {
+    option (google.api.http) = {get: "/v1/healthz"};
+  }
```
begelundmuller, line 20: Could it be called just `Health`?

proto/rill/runtime/v1/api.proto
```diff
@@ -215,6 +219,24 @@ message PingResponse {
   google.protobuf.Timestamp time = 2;
 }
 
+message HealthzRequest {}
```
begelundmuller, line 222: Wondering if we should either:
- Add support for an optional `instance_id` here to limit checks to only one instance
- Or extract instance checks to a separate `InstanceHealth` RPC

Thinking we might want to expose the instance health check in `rill project status` as well.

proto/rill/runtime/v1/api.proto
```diff
@@ -215,6 +219,24 @@ message PingResponse {
...
+  string metastore = 3;
+  string network = 4;
+  map<string, InstanceHealth> instances = 5;
+}
+
+message InstanceHealth {
+  string controller = 1;
+  string repo_sync = 2;
+  string admin_connect = 3;
+}
```
begelundmuller, line 238: Subjective, but I would consider putting an `_error` suffix on the string types here since the message names do not themselves imply that these are errors.

runtime/connection_cache.go
```diff
@@ -25,6 +25,11 @@ var (
 	connCacheCloseLatencyMS = observability.Must(meter.Int64Histogram("connnection_cache.close_latency", metric.WithUnit("ms")))
 )
 
+type connectionCache struct {
+	conncache.Cache
+	hasHungConn bool
+}
```
begelundmuller, line 31: Could this be moved to a func on `conncache.Cache`? It could be called `HangingErr()` and return an error containing the name of the first connection that is hanging (cached from the hanging check). It would also enable auto-recover in case the connection actually stops hanging.

runtime/drivers/duckdb/catalogv2.go
```diff
@@ -11,6 +11,12 @@ import (
 // NOTE: Not using RETURNING, indexes, transactions or any other fancy features to avoid DuckDB bugs.
 // The DuckDB database lock and acquireMetaConn anyway ensure we don't need to worry about concurrency.
 
+func (c *connection) Ping(ctx context.Context) error {
+	var r int
+	err := c.db.QueryRowxContext(ctx, "SELECT 1").Scan(&r)
+	return err
+}
```
begelundmuller, line 18: Shouldn't it use `database/sql`'s native `PingContext` for this?

runtime/drivers/registry.go
```diff
@@ -12,6 +12,9 @@ import (
 
 // RegistryStore is implemented by drivers capable of storing and looking up instances and repos.
 type RegistryStore interface {
+	// Ping returns a non nil error if RegistryStore is not in a healthy state. It is used for health checks.
+	Ping(ctx context.Context) error
+
```
begelundmuller, line 17: Any reason not to put `Ping` on the root `Handle`? The idea at least is that the different interfaces exposed by `Handle` all apply to the same underlying connector.

runtime/drivers/repo.go
```diff
@@ -28,6 +30,11 @@ type RepoStore interface {
 	Watch(ctx context.Context, cb WatchCallback) error
 }
 
+type RepoHealth struct {
+	Sync         error
+	AdminConnect error
+}
```
begelundmuller, line 36: Feels like this leaks a specific connector (`admin`) to all the other connectors. I would suggest keeping it simple by relying on a `Ping(ctx) error` on the handle itself. The `admin` connector can use `errors.Join` to combine the sync and ping error.

From the health check endpoint side, it's okay just to have that one error for the admin connector.

proto/rill/runtime/v1/api.proto
k-anshul: It seems like a convention that google [follows](https://stackoverflow.com/questions/43380939/where-does-the-convention-of-using-healthz-for-application-health-checks-come-f).
I can change it to `Health` as well.

proto/rill/runtime/v1/api.proto
k-anshul: Extracting to `InstanceHealth` seems better. Otherwise need to ping registry and other common stores for every instance.

runtime/drivers/registry.go
k-anshul: Yeah I originally planned to add `Ping` to the `Handle` but later moved to individual stores since the health API seemed a function of the underlying Store. 
Also some drivers may want to expose extra information like the `RepoHealth`.
For example duckdb implementing a catalog store would just do a `Ping` but duckdb implementing OLAP might do additional checks checking if a connection has been acquired for longer time. 

TBH I was split on this and went ahead with the other approach.

runtime/drivers/repo.go
k-anshul: The idea behind not combining errors was that admin service(or manual) can take decision what to do based on different parameters. 
A `sync` error may just require triggering a reconcile while a `connect` error will require `reset`/`redeploy`. I am wondering what action will be taken in case we combine these errors ?

k-anshul: > Feels like this leaks a specific connector (admin) to all the other connectors. 

Only to connectors implementing `RepoStore`. Can also be changed to just `Connect`.

proto/rill/runtime/v1/api.proto
```diff
@@ -215,6 +223,25 @@ message PingResponse {
   google.protobuf.Timestamp time = 2;
 }
 
+message HealthRequest {}
```
begelundmuller, line 226: I like that `InstanceHealth` is separated out and it would be great to incorporate that into `rill project status`. But it would also be nice for the runtime-wide `Health` request to be able to check indexes – since a runtime isn't really healthy if its instances aren't, and it will make the health check easier to call (which is important to maximize it's utility).

We could potentially add a `bool skip_instances` option here in the `HealthRequest` for cases where the caller doesn't want the instance-level checks.

runtime/server/health.go
```diff
@@ -0,0 +1,52 @@
...
+	h, err := s.runtime.InstanceHealth(ctx, req.InstanceId)
+	if err != nil {
+		return nil, err
+	}
+	return &runtimev1.InstanceHealthResponse{
+		ControllerError: h.Controller.Error(),
+		RepoError:       h.Repo.Error(),
+		OlapError:       h.OLAP.Error(),
+	}, nil
+}
```
begelundmuller, line 42: Need permission checks (since these will be public)

begelundmuller (changes requested):

runtime/drivers/athena/athena.go
```diff
@@ -113,6 +113,11 @@ type Connection struct {
 
 var _ drivers.Handle = &Connection{}
 
+// Ping implements drivers.Handle.
+func (c *Connection) Ping(ctx context.Context) error {
+	return drivers.ErrNotImplemented
+}
```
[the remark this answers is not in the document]
k-anshul: But if we return `nil` then it means HealthCheck is implemented and the connection is healthy which in this case is not. 
Once we start using these `Pings` we should actually implement this.

runtime/pkg/conncache/conncache.go
```diff
@@ -433,15 +445,19 @@ func (c *cacheImpl) periodicallyCheckHangingConnections() {
 		select {
 		case <-ticker.C:
 			c.mu.Lock()
+			var hangingConnErr error
 			for k := range c.singleflight {
 				e := c.entries[k]
 				if c.opts.OpenTimeout != 0 && e.status == entryStatusOpening && time.Since(e.since) > c.opts.OpenTimeout {
+					hangingConnErr = fmt.Errorf("a connection has been in opening state for too long")
```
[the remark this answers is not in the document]
k-anshul: The `handle` will not be populated if the connection is in `entryStatusOpening` state. I could not think of any other way to get the driver name.

runtime/registry.go
```diff
@@ -243,6 +243,39 @@ func (r *registryCache) list() ([]*drivers.Instance, error) {
...
 }
 
+func (r *registryCache) health(ctx context.Context, instanceID string) (InstanceHealth, error) {
+	r.mu.Lock()
+	inst, ok := r.instances[instanceID]
+	if !ok {
+		r.mu.Unlock()
+		return InstanceHealth{}, drivers.ErrNotFound
+	}
+	r.mu.Unlock()
```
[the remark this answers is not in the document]
k-anshul: Okay now I have a point that why I prefer names like `ReadLock` over `RLock`.
I wonder if there is a psychological theory to explain this (apart from lack of concentration :) ).

proto/rill/runtime/v1/api.proto
k-anshul: Okay my understanding based on the previous comment was that we want to extract instance health checks from the health check API. I can add instance health checks as well.

k-anshul: The `skip_instances` can be avoided as of now.

runtime/server/health.go
```diff
@@ -21,23 +28,43 @@ func (s *Server) Health(ctx context.Context, req *runtimev1.HealthRequest) (*run
...
 // InstanceHealth implements RuntimeService
 func (s *Server) InstanceHealth(ctx context.Context, req *runtimev1.InstanceHealthRequest) (*runtimev1.InstanceHealthResponse, error) {
+	observability.AddRequestAttributes(ctx,
+		attribute.String("args.instance_id", req.InstanceId),
+	)
+	s.addInstanceRequestAttributes(ctx, req.InstanceId)
+
+	if !auth.GetClaims(ctx).Can(auth.ManageInstances) {
+		return nil, ErrForbidden
+	}
```
begelundmuller, line 60:
```suggestion
	if !auth.GetClaims(ctx).CanInstance(req.InstanceId, auth.ReadInstance) {
		return nil, ErrForbidden
	}
```

begelundmuller (approved):

PR #5188, opened 2024-07-03
User groups

esevastyanov:


admin/database/database.go
```diff
@@ -115,9 +116,14 @@ type DB interface {
...
 
 	InsertUsergroup(ctx context.Context, opts *InsertUsergroupOptions) (*Usergroup, error)
+	UpdateUsergroupName(ctx context.Context, name, groupID, orgID string) (*Usergroup, error)
+	DeleteUsergroup(ctx context.Context, groupID string) error
+	FindUsergroupByName(ctx context.Context, name, orgID string) (*Usergroup, error)
 	FindUsergroupsForUser(ctx context.Context, userID, orgID string) ([]*Usergroup, error)
 	InsertUsergroupMember(ctx context.Context, groupID, userID string) error
+	FindUsergroupMembers(ctx context.Context, groupID string) ([]*UsergroupMember, error)
 	DeleteUsergroupMember(ctx context.Context, groupID, userID string) error
+	DeleteUsergroupsMember(ctx context.Context, userID, orgID string) error
```
begelundmuller, line 126: Can it return the `Member` type instead of `UsergroupMember`? And maybe be renamed to `FindUsergroupMemberUsers`, `InsertUsergroupMemberUser`, etc.

Then it would align completely with the related functions for org and project membership (e.g. `FindOrganizationMemberUsers`, which returns `Member`). We should consider renaming the `type Member` to `type MemberUser` to clarify that it is a member that is a user.

admin/database/database.go
```diff
@@ -179,10 +185,18 @@ type DB interface {
...
 	UpdateProjectMemberUserRole(ctx context.Context, projectID, userID, roleID string) error
+	FindOrganizationUsergroups(ctx context.Context, orgID, afterName string, limit int) ([]*Usergroup, error)
+	FindAllUsergroupOrganizationRoles(ctx context.Context, orgID string) ([]*UsergroupOrgRole, error)
+	FindUsergroupOrganizationRole(ctx context.Context, groupID, orgID string) (*UsergroupOrgRole, error)
+	FindUsergroupAllProjectOrganizationRoles(ctx context.Context, orgID string) ([]*UsergroupProjectRole, error)
+	FindUsergroupProjectRoles(ctx context.Context, groupID, orgID string) ([]*UsergroupProjectRole, error)
+	UpsertOrganizationUsergroup(ctx context.Context, groupID, orgID, roleID string) error
+	DeleteOrganizationUsergroup(ctx context.Context, groupID, orgID string) error
+	UpsertProjectUsergroup(ctx context.Context, groupID, projectID, roleID string) error
+	DeleteProjectUsergroup(ctx context.Context, projectID, groupID string) error
```
begelundmuller, line 199: It would be nice if we could lean into the "member" terminology used for orgs and projects. But we have to be careful since a usergroup can both *have* members and *be* a member of something. I would propose using `MemberUsergroup` to consistently denote a "member that is a usergroup", and naming these functions `FindOrganizationMemberUsergroups`, `InsertOrganizationMemberUsergroup`, etc. Then they will align with the user-related functions `FindOrganizationMemberUser`, `InsertOrganizationMemberUser`, etc.

admin/database/database.go
```diff
@@ -465,9 +479,34 @@ type UpdateServiceOptions struct {
...
+	RoleName    string `db:"role_name"`
+}
+
+// UsergroupProjectRole represents a project role of Usergroup
+type UsergroupProjectRole struct {
+	UsergroupID string `db:"usergroup_id"`
+	ProjectID   string `db:"project_id"`
+	ProjectName string `db:"project_name"`
+	RoleName    string `db:"role_name"`
+}
```
begelundmuller, line 503: Similar to how `Member` is used for both org and project member users (we should rename the type to `MemberUser`, but it's okay not to do it in this PR to keep things simple), it would be nice if we could have one `MemberUsergroup` type used to represent a "member of type usergroup".

admin/database/database.go
```diff
@@ -465,9 +479,34 @@ type UpdateServiceOptions struct {
...
+
+type UsergroupMember struct {
+	UsergroupID string `db:"usergroup_id"`
+	UserID      string `db:"user_id"`
+	Email       string `db:"email"`
+	DisplayName string `db:"display_name"`
 }
```
begelundmuller, line 510: In line with other comments, I would propose removing this type and using the `Member` type, as is used for org/project member users.

admin/database/postgres/migrations/0032.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE usergroups ADD COLUMN description TEXT;
```
begelundmuller, line 1: Consider `NOT NULL DEFAULT ''` (since it makes the code simpler and there is no meaningful difference between a `null` value and empty string in this case).

admin/database/database.go
```diff
@@ -465,9 +479,34 @@ type UpdateServiceOptions struct {
 
 // Usergroup represents a group of org members
 type Usergroup struct {
-	ID    string `db:"id"`
-	OrgID string `db:"org_id"`
-	Name  string `db:"name" validate:"slug"`
+	ID          string         `db:"id"`
+	OrgID       string         `db:"org_id"`
+	Name        string         `db:"name" validate:"slug"`
+	Description sql.NullString `db:"description"`
```
begelundmuller, line 485: Can be changed to a simple `string` if the migration is changed slightly

cli/cmd/usergroup/list_user.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/rilldata/rill/cli/pkg/cmdutil"
+	adminv1 "github.com/rilldata/rill/proto/gen/rill/admin/v1"
+	"github.com/spf13/cobra"
+)
+
+func ListUserCmd(ch *cmdutil.Helper) *cobra.Command {
+	var group string
+
+	addCmd := &cobra.Command{
+		Use:   "list-user <name>",
```
begelundmuller, line 13: Similar comment – torn on this, but if we want just one way of doing things, this might better fit in `rill user list --group NAME`, similar to `rill user list --project NAME`.

proto/rill/admin/v1/api.proto
```diff
@@ -211,6 +211,83 @@ service AdminService {
...
+      body: "*"
+    };
+  }
+
+  // RevokeProjectUsergroupRole revokes the project-level role for the user group
+  rpc RevokeProjectUsergroupRole(RevokeProjectUsergroupRoleRequest) returns (RevokeProjectUsergroupRoleResponse) {
+    option (google.api.http) = {
+      delete: "/v1/organizations/{organization}/projects/{project}/usergroups/{usergroup}/roles",
+    };
+  }
```
begelundmuller, line 271: It would be nice to align these names with the ones for managing org and project member users, like `AddProjectMember`, `SetProjectMemberRole`, etc. As discussed previously, we might rename `Member` to `MemberUser` in those cases. Then we might get:

- `ListOrganizationMembers` -> `ListOrganizationMemberUsers` 
- `AddOrganizationMember` -> `AddOrganizationMemberUser`
- `RemoveOrganizationMember` -> `RemoveOrganizationMemberUser`
- `SetOrganizationMemberRole` -> `SetOrganizationMemberUserRole`
- `ListProjectMembers` -> `ListProjectMemberUsers`
- `AddProjectMember` -> `AddProjectMemberUser`
- `RemoveProjectMember` -> `RemoveProjectMemberUser`
- `SetProjectMemberRole` -> `SetProjectMemberUserRole`
- `ListOrganizationMemberUsergroups` 
- `AddOrganizationMemberUsergroup`
- `RemoveOrganizationMemberUsergroup`
- `SetOrganizationMemberUsergroupRole`
- `ListProjectMemberUsergroups`
- `AddProjectMemberUsergroup`
- `RemoveProjectMemberUsergroup`
- `SetProjectMemberUsergroupRole`

admin/database/database.go
```diff
@@ -465,9 +480,27 @@ type UpdateServiceOptions struct {
...
+	RoleName    string `db:"role_name"`
+}
+
+// ProjectMemberUsergroupRole represents a project role of Usergroup
+type ProjectMemberUsergroupRole struct {
+	UsergroupID string `db:"usergroup_id"`
+	ProjectID   string `db:"project_id"`
+	ProjectName string `db:"project_name"`
+	RoleName    string `db:"role_name"`
 }
```
begelundmuller, line 504: I think these types could be replaced with one type `MemberUsergroup` to mirror `MemberUser`, which is used for both org and project memberships

admin/server/usergroups.go
```diff
@@ -0,0 +1,696 @@
...
+	}
+
+	org, err := s.admin.DB.FindOrganization(ctx, proj.OrganizationID)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	if org.AllUsergroupID != nil && usergroup.ID == *org.AllUsergroupID {
+		return nil, status.Error(codes.InvalidArgument, "cannot set role for all-users group")
+	}
```
begelundmuller, line 369: For projects, it's actually okay to be able to remove/edit/add the `all-users` access permissions. This will enable creating a project in an org that all of the org's users don't have access to.

admin/server/usergroups.go
```diff
@@ -0,0 +1,696 @@
...
```
begelundmuller, line 414: Same comment – okay to edit the `all-users` permission on projects

admin/server/usergroups.go
```diff
@@ -0,0 +1,696 @@
...
+
+	org, err := s.admin.DB.FindOrganization(ctx, proj.OrganizationID)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	if org.AllUsergroupID != nil && usergroup.ID == *org.AllUsergroupID {
+		return nil, status.Error(codes.InvalidArgument, "cannot revoke role from all-users group")
+	}
```
begelundmuller, line 453: Same comment – okay to edit the `all-users` permission on projects

admin/server/usergroups.go
```diff
@@ -0,0 +1,696 @@
...
+	projectRoles, err := s.admin.DB.FindProjectMemberUsergroupRoles(ctx, group.ID, org.ID)
+	if err != nil && !errors.Is(err, database.ErrNotFound) {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	for _, role := range projectRoles {
+		if !claims.ProjectPermissions(ctx, org.ID, role.ProjectID).ReadProjectMembers {
+			return nil, status.Error(codes.PermissionDenied, "not allowed to list usergroup members")
+		}
+	}
```
begelundmuller, line 561: Might be missing something, but I think listing a group's members should just require `ReadOrgMembers` on the group's org?

admin/server/usergroups.go
```diff
@@ -0,0 +1,696 @@
...
+	projectRoles, err := s.admin.DB.FindProjectMemberUsergroupRoles(ctx, group.ID, org.ID)
+	if err != nil && !errors.Is(err, database.ErrNotFound) {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	for _, role := range projectRoles {
+		if !claims.ProjectPermissions(ctx, org.ID, role.ProjectID).ManageProjectMembers {
+			return nil, status.Error(codes.PermissionDenied, "not allowed to remove usergroup members")
+		}
+	}
```
begelundmuller, line 627: Similarly, removing a user should just require `ReadOrgMembers` on the group's org

admin/database/database.go
```diff
@@ -115,9 +115,14 @@ type DB interface {
 	CheckUserIsAProjectMember(ctx context.Context, userID, projectID string) (bool, error)
 
 	InsertUsergroup(ctx context.Context, opts *InsertUsergroupOptions) (*Usergroup, error)
+	UpdateUsergroupName(ctx context.Context, name, groupID string) (*Usergroup, error)
+	DeleteUsergroup(ctx context.Context, groupID string) error
+	FindUsergroupByName(ctx context.Context, name, orgID string) (*Usergroup, error)
```
begelundmuller, line 120: Consider changing this to `FindUsergroupByName(ctx, orgName, name)`. This would be similar to `FindProjectByName(ctx, orgName, name)`. As far as I can tell, most places it is used, that would save a separate lookup for the org. The usergroup returned here will contain the org ID to use for permission checks.

begelundmuller (changes requested):

admin/database/database.go
```diff
@@ -170,19 +175,29 @@ type DB interface {
 	ResolveOrganizationRolesForUser(ctx context.Context, userID, orgID string) ([]*OrganizationRole, error)
```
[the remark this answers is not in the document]
esevastyanov: it is already there so that a collection of roles is returned

admin/database/database.go
```diff
@@ -170,19 +175,29 @@ type DB interface {
...
 	CountSingleuserOrganizationsForMemberUser(ctx context.Context, userID string) (int, error)
 
-	FindProjectMemberUsers(ctx context.Context, projectID, afterEmail string, limit int) ([]*Member, error)
+	FindProjectMemberUsers(ctx context.Context, projectID, afterEmail string, limit int) ([]*MemberUser, error)
 	InsertProjectMemberUser(ctx context.Context, projectID, userID, roleID string) error
-	InsertProjectMemberUsergroup(ctx context.Context, groupID, projectID, roleID string) error
 	DeleteProjectMemberUser(ctx context.Context, projectID, userID string) error
 	DeleteAllProjectMemberUserForOrganization(ctx context.Context, orgID, userID string) error
 	UpdateProjectMemberUserRole(ctx context.Context, projectID, userID, roleID string) error
+	FindOrganizationMemberUsergroups(ctx context.Context, orgID, afterName string, limit int) ([]*Usergroup, error)
```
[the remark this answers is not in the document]
esevastyanov: Going to update to `FindOrganizationUsergroups` but it returns a collection of MemberUsergroup to reduce the number of lookups
```
// FindOrganizationUsergroups returns org user groups as collection of MemberUsergroup.
// If a user group has no org role then RoleName is empty.
func (c *connection) FindOrganizationUsergroups(ctx context.Context, orgID, afterName string, limit int) ([]*database.MemberUsergroup, error) {
```

admin/database/database.go
```diff
@@ -115,9 +115,15 @@ type DB interface {
...
 	InsertUsergroup(ctx context.Context, opts *InsertUsergroupOptions) (*Usergroup, error)
+	UpdateUsergroupName(ctx context.Context, name, groupID string) (*Usergroup, error)
+	UpdateUsergroupDescription(ctx context.Context, description, groupID string) (*Usergroup, error)
+	DeleteUsergroup(ctx context.Context, groupID string) error
+	FindUsergroupByName(ctx context.Context, orgName, name string) (*Usergroup, error)
 	FindUsergroupsForUser(ctx context.Context, userID, orgID string) ([]*Usergroup, error)
-	InsertUsergroupMember(ctx context.Context, groupID, userID string) error
-	DeleteUsergroupMember(ctx context.Context, groupID, userID string) error
+	InsertUsergroupMemberUser(ctx context.Context, groupID, userID string) error
+	FindUsergroupMembersUsers(ctx context.Context, groupID string) ([]*MemberUser, error)
```
begelundmuller, line 124: 1. Should have pagination similar to `FindOrganizationMemberUsers`
2. nit: remove the "s" from "members": `FindUsergroupMemberUsers`

admin/database/database.go
```diff
@@ -170,20 +176,29 @@ type DB interface {
...
 
-	FindProjectMemberUsers(ctx context.Context, projectID, afterEmail string, limit int) ([]*Member, error)
+	FindProjectMemberUsers(ctx context.Context, projectID, afterEmail string, limit int) ([]*MemberUser, error)
 	InsertProjectMemberUser(ctx context.Context, projectID, userID, roleID string) error
-	InsertProjectMemberUsergroup(ctx context.Context, groupID, projectID, roleID string) error
 	DeleteProjectMemberUser(ctx context.Context, projectID, userID string) error
 	DeleteAllProjectMemberUserForOrganization(ctx context.Context, orgID, userID string) error
 	UpdateProjectMemberUserRole(ctx context.Context, projectID, userID, roleID string) error
 
+	FindOrganizationUsergroups(ctx context.Context, orgID, afterName string, limit int) ([]*MemberUsergroup, error)
```
begelundmuller, line 192: Name `FindOrganizationMemberUsergroups` to align with the similar `FindXXXMemberYYY` functions? I realize this one is slightly different due to the left joins though, so maybe not?

Question – did you consider having separate `FindUsergroups(ctx, org)` (for listing usergroups belonging to an org) and `FindOrganizationMemberUsergroups(ctx, org)` (for listing usergroup memberships, i.e. usergroups having a role on an org)? It might be a cleaner separation, but on the other hand, maybe makes downstream logic unnecessarily complex?

cli/pkg/printer/resources.go
```diff
@@ -388,3 +388,71 @@ type plan struct {
...
+		UpdatedOn: ug.UpdatedOn.AsTime().Local().Format(time.DateTime),
+		Role:      role,
+	}
+}
+
+type memberUsergroup struct {
+	Name      string `header:"name" json:"name"`
+	CreatedOn string `header:"created_on,timestamp(ms|utc|human)" json:"created_at"`
+	UpdatedOn string `header:"updated_on,timestamp(ms|utc|human)" json:"updated_at"`
+	Role      string `header:"role" json:"role"`
```
begelundmuller, line 426:
```suggestion
	Name      string `header:"name" json:"name"`
	Role      string `header:"role" json:"role"`
	CreatedOn string `header:"created_on,timestamp(ms|utc|human)" json:"created_at"`
	UpdatedOn string `header:"updated_on,timestamp(ms|utc|human)" json:"updated_at"`
```

admin/database/database.go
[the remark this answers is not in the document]
begelundmuller: Ah perfect

begelundmuller (approved):

PR #5189, opened 2024-07-04
stripe integration

pjain1:


admin/database/database.go
```diff
@@ -224,6 +224,10 @@ type DB interface {
 	CountBillingProjectsForOrganization(ctx context.Context, orgID string, createdBefore time.Time) (int, error)
 	FindBillingUsageReportedOn(ctx context.Context) (time.Time, error)
 	UpdateBillingUsageReportedOn(ctx context.Context, usageReportedOn time.Time) error
+
+	FindBillingRepairedOn(ctx context.Context) (time.Time, error)
+	UpdateBillingRepairedOn(ctx context.Context, repairedOn time.Time) error
+	FindOrganizationIDsWithoutPaymentCreatedOnOrAfter(ctx context.Context, createdAfter time.Time) ([]string, error)
```
begelundmuller, line 230: Is there a reason that an org won't have a payment provider ID? If not, we could remove the `createdAfter` part here and just do `FindOrganizationIDsWithoutPaymentCustomerID`

admin/server/organizations.go
```diff
@@ -240,79 +249,69 @@ func (s *Server) UpdateOrganizationBillingSubscription(ctx context.Context, req
...
-						return nil, status.Error(codes.Internal, err.Error())
-					}
-				}
-
-				// create new subscription
-				_, err = s.admin.Biller.CreateSubscription(ctx, org.BillingCustomerID, plan)
+		} else {
+			// multiple subscriptions, cancel them first immediately and assign new plan
+			// should not happen unless externally assigned multiple subscriptions to the same org in the billing system
+			for _, sub := range subs {
```
begelundmuller, line 283: Is this needed now that it calls `RepairOrgBilling` above?

admin/worker/billing_repair.go
```diff
@@ -0,0 +1,42 @@
...
+)
+
+func (w *Worker) repairOrgBilling(ctx context.Context) error {
+	startTime := time.Now().UTC()
+	t, err := w.admin.DB.FindBillingRepairedOn(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get last billing repair time: %w", err)
+	}
+
+	ids, err := w.admin.DB.FindOrganizationIDsWithoutPaymentCreatedOnOrAfter(ctx, t)
```
begelundmuller, line 18: Could we just try for all orgs without a payment ID each time the job runs?

proto/rill/admin/v1/api.proto
```diff
@@ -58,6 +58,11 @@ service AdminService {
     option (google.api.http) = {get: "/v1/organizations/{org_name}/billing/subscriptions"};
   }
 
+  // GetOrganizationBillingSessionURL returns the URL for the billing session to collect payment method
+  rpc GetOrganizationBillingSessionURL(GetOrganizationBillingSessionURLRequest) returns (GetOrganizationBillingSessionURLResponse) {
+    option (google.api.http) = {get: "/v1/organizations/{org_name}/billing/payments/session-url"};
+  }
```
begelundmuller, line 64: Can we merge this with `GetOrganizationBillingSubscriptionResponse`, i.e. just return the payment portal URL alongside the billing portal URL?

begelundmuller (changes requested):

admin/billing/payment/stripe.go
```diff
@@ -0,0 +1,107 @@
...
+	return &Customer{
+		ID:                 c.ID,
+		Name:               c.Name,
+		Email:              c.Email,
+		ValidPaymentMethod: i.Next(), // very basic check if the customer has a payment method // TODO improve this
+	}, nil
+}
+
+func (s *Stripe) FindCustomerForOrg(ctx context.Context, organization *database.Organization) (*Customer, error) {
+	// TODO once we capture billing email then we can use that directly to find the customer
```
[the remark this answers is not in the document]
pjain1: there is no external ID field as I can see [here](https://github.com/stripe/stripe-go/blob/master/customer.go), the customer id returned from stripe after creation of customer is the only unique identifier. Even the email is not unique there can be more than one customer having same email.
Screenshot 2024-07-10 at 11 20 04 PM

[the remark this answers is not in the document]
pjain1: This method is only used by repair logic to find if we created customer in stripe but failed to update the `organization.PaymentCustomerID` in the database. If we have valid `organization.PaymentCustomerID` then this is not used but `FindCustomer(ctx context.Context, customerID string)` is used which does lookup using `PaymentCustomerID`

admin/database/database.go
pjain1: an org won't have payment provider id when there was any failure before updating the database with it or we failed in creating customer in stripe (may be some network issue or stripe server issue) etc. as stripe, orb customer creation and org updation with ids cannot be done in a transaction. If any or all step fails, repair job takes care of all scenarios. `createdAfter` is just an optimization to not look at orgs that were created before start time of last repair job.

admin/server/organizations.go
pjain1: So repair org does not fix multiple subscription issue, as I am not sure which subscription to cancel and which to keep, I can add a warn or error log in this case and hook it up so that we get notified properly. However this API is plan change API so it can just cancel all older subscription and create a new one with new plan.

admin/worker/billing_repair.go
pjain1: sure that can also be done, will be simpler as well.

proto/rill/admin/v1/api.proto
pjain1: The reason this is a separate API because billing customer portal URL is fixed (for Orb atleast) and does not change where as this  payment session url expires and needs to be created on-demand using stripe API. I wanted to unnecessarily create this every time we fetch subscription details which can be done multiple times for various use cases.

admin/database/database.go
pjain1: changed this to `FindOrganizationWithoutPaymentCustomerID` to get all orgs without valid payment customer id

admin/billing/payment/stripe.go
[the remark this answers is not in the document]
begelundmuller: Okay, did you consider using the search API? If it has some problems, can you:
1. Add a docstring about being cautious with using this function
2. Maybe change the name to `FindCustomersCreatedAfter` so someone doesn't accidentally use it to lookup payment methods for an org
3. Limit pagination so it doesn't accidentally page over potentially 1000s of rows?

admin/database/database.go
```diff
@@ -224,6 +224,8 @@ type DB interface {
 	CountBillingProjectsForOrganization(ctx context.Context, orgID string, createdBefore time.Time) (int, error)
 	FindBillingUsageReportedOn(ctx context.Context) (time.Time, error)
 	UpdateBillingUsageReportedOn(ctx context.Context, usageReportedOn time.Time) error
+
+	FindOrganizationWithoutPaymentCustomerID(ctx context.Context) ([]*Organization, error)
```
begelundmuller, line 228: nit: organizations in plural: `FindOrganizationsWithoutPaymentCustomerID`

admin/server/organizations.go
begelundmuller: Got it. Might be worth adding a comment about.

proto/rill/admin/v1/api.proto
begelundmuller: Okay, can we then rename to clarify the difference – e.g. `GetPaymentsPortalURL` ("session URL" does not seem very descriptive).

And on reflection, I don't think we need to prefix these with calls with "organization". Should we also change:
- `GetOrganizationBillingSubscription` -> `GetBillingSubscription`
- `UpdateOrganizationBillingSubscription` -> `UpdateBillingSubscription`

proto/rill/admin/v1/api.proto
```diff
@@ -654,6 +659,17 @@ message GetOrganizationBillingSubscriptionRequest {
 message GetOrganizationBillingSubscriptionResponse {
   Organization organization = 1;
   Subscription subscription = 2;
+  string billing_portal_url = 3;
+  bool has_payment_method = 4;
+}
+
+message GetOrganizationBillingSessionURLRequest {
+  string org_name = 1;
+  string return_url = 2;
```
begelundmuller, line 668: Consider `redirect_url` or `redirect` – I think that's what we call redirect URLs other places

pjain1: so this is a redirect url actually, page does not redirect after a payment method is entered, this is shown as a simple hyperlink which user has to click to return to our application. See the screenshot below, its the `Return to New Business` link.
Screenshot 2024-07-11 at 5 02 13 PM

begelundmuller: Okay, then `return_url` makes sense

admin/billing/payment/stripe.go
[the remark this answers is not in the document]
begelundmuller: Yeah limiting to a few mins sounds good

begelundmuller (approved):
Note there's a merge conflict around the DB migrations to be aware of

pjain1:
resolved conflicts with main

Issue #4851, opened 2024-05-07
Feat(dashboard): Table Component


mindspank:
https://www.notion.so/rilldata/Table-Component-7f5a4cac15f3467cb44c8be4a5f64342?pvs=4

PR #5191, opened 2024-07-05
feat: table component for custom dashboard

djbarnwal:
Closes [#4851](https://github.com/rilldata/rill/issues/4851)

Spec 

```
kind: component

table:
  metric_view: Bids_Sample_Dash
  time_range: P7D
  measures:
    - measure_2
  row_dimensions:
    - player_size
  col_dimensions: 
    - app_or_site
```

image


briangregoryholmes:
Application crashed for me on this uncaught error. Just copied from your spec, but, with auto save, didn't get a chance to change the metric view and dimension/measure names to ones that were valid for my project.

Screenshot 2024-07-09 at 11 01 57 AM

briangregoryholmes:
More specific error when the measure is invalid:

Screenshot 2024-07-09 at 11 08 34 AM

nishantmonu51:
@djbarnwal : Is this ready to be merged ?

briangregoryholmes:
https://jam.dev/c/f81d00f5-834b-4488-8241-a3a1873ff3ea

Still having a few issues. Changing the dimension name to another valid dimension doesn't update the table when saving. In this recording, it required two refreshes, though that isn't the case in every browser.

With auto-save turned on, changing something like metric view name to something that isn't valid no longer hangs the application, but it does require a refresh of the page once you change it back to something valid.

runtime/compilers/rillv1/data/component-template-v1.json
```diff
@@ -145,10 +148,8 @@
         }
       },
       "required": [
-        "col_dimensions",
         "measures",
         "metric_view",
```
ericpgreen2, line 152: Nit: we should consolidate on "metrics_view" throughout the codebase, not "metric_view"

web-common/src/features/templates/table/TableTemplate.svelte
```diff
@@ -0,0 +1,89 @@
...
+  $: if (isValidSchema) {
+    const stateManagerContext = createStateManagers({
+      queryClient,
+      metricsViewName: tableProperties.metric_view,
+      extraKeyPrefix: TABLE_PREFIX,
+    });
+
+    pivotConfig = getTableConfig(instanceId, tableProperties, $pivotState);
+    pivotDataStore = createPivotDataStore(stateManagerContext, pivotConfig);
+  }
```
ericpgreen2, line 74: Writing down what we discussed [in Slack](https://rilldata.slack.com/archives/C04AGHL77PS/p1722346431806149?thread_ts=1722321321.527189&cid=C04AGHL77PS) – we likely shouldn't re-create the `pivotDataStore` every time the `tableProperties` change, but rather we should update the store's values.

ericpgreen2 (approved):
The table editing experience is good! Great component to have.

(UXQA) The table component renders oddly in the Custom Dashboard workspace. It's small and doesn't fill the space of the container:
image

web-common/src/features/templates/table/TableTemplate.svelte
djbarnwal: Will address in follow up

djbarnwal:
> (UXQA) The table component renders oddly in the Custom Dashboard workspace. It's small and doesn't fill the space of the container

Sizing in general is an issue which needs to be worked out for all components. It's tricky because we don't want to always increase the font size based on container size. This needs more discussion offline.

runtime/compilers/rillv1/data/component-template-v1.json
djbarnwal: This is a common change for all components. Will create a separate PR.

PR #5192, opened 2024-07-05
feat: http health api and admin job for health rpc

k-anshul:


admin/database/database.go
```diff
@@ -100,6 +100,9 @@ type DB interface {
 
 	ResolveRuntimeSlotsUsed(ctx context.Context) ([]*RuntimeSlotsUsed, error)
 
+	// FindAllocatedRuntimes returns runtimes on which atleast one instance is allocated
+	FindAllocatedRuntimes(ctx context.Context, afterRuntime string, limit int) ([]*AllocatedRuntime, error)
```
begelundmuller, line 104: nit: `afterHost` (assuming `host` is the pagination field used here?)

admin/worker/worker.go
```diff
@@ -67,6 +67,9 @@ func (w *Worker) Run(ctx context.Context) error {
 	group.Go(func() error {
 		return w.schedule(ctx, "delete_unused_assets", w.deleteUnusedAssets, 6*time.Hour)
 	})
+	group.Go(func() error {
+		return w.schedule(ctx, "runtime_health_check", w.runtimeHealthCheck, 10*time.Minute)
```
begelundmuller, line 71: Would it be more correct to call it "deployments health check" (`deployments_health_check`) instead?

runtime/drivers/admin/admin.go
```diff
@@ -143,7 +143,9 @@ func (h *Handle) Ping(ctx context.Context) error {
 	// check connectivity with admin service
 	_, err := h.admin.Ping(ctx, &adminv1.PingRequest{})
 
-	_ = h.repoMu.RLock(ctx)
+	if lockErr := h.repoMu.RLock(ctx); lockErr != nil {
+		return lockErr
+	}
```
begelundmuller, line 148:
```suggestion
	if err := h.repoMu.RLock(ctx); err != nil {
		return err
	}
```

admin/worker/runtime_health_check.go
```diff
@@ -0,0 +1,73 @@
...
+	"context"
+	"fmt"
+
+	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
+	"go.uber.org/zap"
+	"golang.org/x/sync/errgroup"
+	"google.golang.org/protobuf/encoding/protojson"
+)
+
+const _allocatedRuntimesPageSize = 20
```
begelundmuller, line 13: Since this is just fetching relatively short host URLs, consider increasing this to maybe 500 or 1k

admin/worker/runtime_health_check.go
```diff
@@ -0,0 +1,73 @@
...
+
+func (w *Worker) runtimeHealthCheck(ctx context.Context) error {
+	lastRuntime := ""
```
begelundmuller, line 16: nit: `afterHost`

admin/worker/runtime_health_check.go
```diff
@@ -0,0 +1,73 @@
...
+		group.SetLimit(8)
+		for _, rt := range runtimes {
+			rt := rt
+			group.Go(func() error {
+				client, err := w.admin.OpenRuntimeClient(rt.Host, rt.Audience)
+				if err != nil {
+					w.logger.Error("runtimeHealthCheck: failed to open runtime client", zap.String("host", rt.Host), zap.Error(err))
+					return nil
+				}
+
```
begelundmuller, line 37: Use `defer client.Close()`?

runtime/drivers/admin/admin.go
k-anshul: There is already a `err` declared above this line. I just feel better to use different names to avoid introducing bugs.

admin/worker/worker.go
k-anshul: It is both IMO since we also check healths of runtime so pretty subjective TBH. I can change it to `deployments_health_check`.

kaspersjo: I also prefer `deployments_health_check` and `FindAllocatedDeployments` etc.
Mainly since we might add more components to a deployment in the near future, so the runtime is really only one aspect of the more general deployment health check (but currently the only one).

admin/worker/runtime_health_check.go
```diff
@@ -0,0 +1,73 @@
...
+				if err != nil {
+					client.Close()
+					w.logger.Error("runtimeHealthCheck: health check call failed", zap.String("host", rt.Host), zap.Error(err))
+					return nil
+				}
+
+				if !isRuntimeHealthy(resp) {
+					s, _ := protojson.Marshal(resp)
+					w.logger.Error("runtimeHealthCheck: runtime is unhealthy", zap.String("host", rt.Host), zap.ByteString("health_response", s))
+				}
```
kaspersjo, line 48: Should we consider updating the deployment status if there's an error and clear the error status if it's healthy?

Can be done with: `w.admin.DB.UpdateDeploymentStatus(ctx, depl.ID, database.DeploymentStatusError, err.Error())`

kaspersjo (changes requested):

admin/database/postgres/postgres.go
```diff
@@ -441,6 +441,15 @@ func (c *connection) DeleteProjectWhitelistedDomain(ctx context.Context, id stri
 	return checkDeleteRow("project whitelist domain", res, err)
 }
 
+func (c *connection) FindDeployments(ctx context.Context, createdAfter time.Time, limit int) ([]*database.Deployment, error) {
+	var res []*database.Deployment
+	err := c.getDB(ctx).SelectContext(ctx, &res, "SELECT d.* FROM deployments d WHERE d.created_on > $1 ORDER BY d.created_on LIMIT $2", createdAfter, limit)
```
begelundmuller, line 446: Would be better to order with `id` and paginate using  `afterID` since that's the primary key and indexed

runtime/drivers/admin/admin.go
begelundmuller: Hadn't seen, makes sense

admin/database/postgres/postgres.go
k-anshul: My understanding was that given our page size selects a significant number of rows in the table the index will not be useful and postgres will sort entire table(https://www.postgresql.org/docs/current/indexes-ordering.html). Given UUID is not naturally sortable it is better to sort on created_on. 
I could also confirm both by running explain on production table. Though deployments are supposed to increase but was not sure what that ratio would be.

But as you mentioned creation time is not unique so it is indispensable to paginate on id now. I have made the relevant changes.

begelundmuller (approved):

kaspersjo (approved):

PR #5193, opened 2024-07-05
fix: hide magic link tab when disallowed

briangregoryholmes:


web-admin/src/routes/+layout.ts
```diff
@@ -4,3 +4,41 @@ app in production. Here, we are setting server-side rendering (SSR) to false to
...
+    });
+
+    const { projectPermissions } = response;
+
+    return {
+      projectPermissions,
+    };
+  } catch (e) {
+    console.error(e);
+    throw error(404, "Unable to find project");
```
ericpgreen2, line 42: This leads to a 404 on magic links – the user doesn't have access to the project via the typical cookie-based authentication (but does have access via the URL's token)

briangregoryholmes: Adjusted.

ericpgreen2 (approved):

PR #5198, opened 2024-07-08
feat: Request project access flow

AdityaHegde:
Adds support for requesting access to a project.
1. Adds new APIs,
    a. `RequestProjectAccess`: For a given `organization` and `project` requests access for the logged in user. Will send an email with `approve` or `deny` CTAs to all admins. Fails if there is already an entry for access request for the user to the project.
    b. `GetProjectAccess`: Gets information about a request. Is mainly used in UI to show approve/deny pages.
    c. `ApproveProjectAccess`: Approves the access request with a role override. Adds the user and sends a confirmation email to the requester. Deletes the entry for access request.
    d. `DenyProjectAccess`: Denies access request and sends an email to the requester. Deletes the entry for access request.
2. If a user visits a project page that the user doesnt have access to, they are redirected to a `request access` page. Calls `RequestProjectAccess` in this case.
3. Adds pages for `Approve` and `Deny`

Notion: https://www.notion.so/rilldata/RD-RC-1-5-UI-Based-User-Project-Invitation-Dashboard-Sharing-826f1d3615f14941935c1d57cf7d8df3?pvs=4#690ec49f17184a3ab72d7af076e5cb7e
Figma: https://www.figma.com/design/Qt6EyotCBS3V6O31jVhMQ7/RILL?node-id=13427-640601&t=ySKcJi0KjtifPXtG-0

**Note** This doesnt redirect to request access page for all links within a project yet. Will do a follow up for this.
For testing: `make cli && ./rill devtool start cloud`


admin/database/postgres/migrations/0034.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE project_access_request (
```
begelundmuller, line 1: nit: `project_access_requests` since our table names are plural

admin/database/postgres/migrations/0034.sql
```diff
@@ -0,0 +1,8 @@
...
+    id UUID NOT NULL DEFAULT uuid_generate_v4() PRIMARY KEY,
+    email TEXT NOT NULL,
```
begelundmuller, line 3: I know this might have UX implications, but it would be great if we could require a user to log in before requesting access, and then store the `user_id` here instead.

By requiring a user to log in, we:
- Avoid sending lots of spam access request emails
- Can better mitigate data harvesting attacks
- Safeguard against pages like "Request access to Customer X on Rill" showing up in search engine results.

admin/server/projects.go
```diff
@@ -974,6 +974,228 @@ func (s *Server) GetCloneCredentials(ctx context.Context, req *adminv1.GetCloneC
...
+
+	adminRole, err := s.admin.DB.FindOrganizationRole(ctx, "admin")
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	admins, err := s.admin.DB.FindOrganizationMemberUsersByRole(ctx, proj.OrganizationID, adminRole.ID)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
```
begelundmuller, line 1032: Instead of hard-coding the role name, It would be better to search for members with the `ManageOrgMembers` permission.

proto/rill/admin/v1/api.proto
```diff
@@ -690,6 +690,31 @@ service AdminService {
...
+      body: "*"
+    };
+  }
+
+  rpc DenyProjectAccess(DenyProjectAccessRequest) returns (DenyProjectAccessResponse) {
+    option (google.api.http) = {
+      post: "/v1/organizations/{organization}/projects/{project}/access-request/{id}/deny",
+      body: "*"
+    };
+  }
```
begelundmuller, line 717: I would suggest just using the ID of the project access request here (without passing org/project names), since it already has a foreign key to the project ID. This also avoids nullifying all access requests if an org/project is renamed.

proto/rill/admin/v1/api.proto
```diff
@@ -690,6 +690,31 @@ service AdminService {
...
+  rpc RequestProjectAccess(RequestProjectAccessRequest) returns (RequestProjectAccessResponse) {
+    option (google.api.http) = {
+      post: "/v1/organizations/{organization}/projects/{project}/access-request",
+      body: "*"
+    };
+  }
+
+  rpc GetProjectAccess(GetProjectAccessRequest) returns (GetProjectAccessResponse) {
+    option (google.api.http) = {get: "/v1/organizations/{organization}/projects/{project}/access-request/{id}"};
+  }
```
begelundmuller, line 703: I know it stutters, but I think this needs to be called `GetProjectAccessRequest(GetProjectAccessRequestRequest)` because right now it sounds like the API returns info about a given user's access permissions.

begelundmuller (changes requested):

admin/database/postgres/migrations/0034.sql
AdityaHegde: We already require login.

admin/server/projects.go
AdityaHegde: Updated

proto/rill/admin/v1/api.proto
AdityaHegde: Got it. Updated

admin/database/postgres/postgres.go
```diff
@@ -1582,6 +1597,56 @@ func (c *connection) UpdateProjectInviteRole(ctx context.Context, id, roleID str
 	return checkUpdateRow("project invite", res, err)
 }
 
+func (c *connection) FindProjectAccessRequests(ctx context.Context, projectID, afterID string, limit int) ([]*database.ProjectAccessRequest, error) {
+	var res []*database.ProjectAccessRequest
+	err := c.getDB(ctx).SelectContext(ctx, &res, `
+			SELECT par.user_id
+			FROM project_access_requests par
+			WHERE par.project_id = $1 AND par.user_id > lower($2)
```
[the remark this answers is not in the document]
AdityaHegde: 1. Missed removing it when i switched to `user_id`
2. Yes it does fail. This method is not used so never tested it. Lemme remove that argument if `afterID == ""`

begelundmuller (approved):

PR #5204, opened 2024-07-08
Runtime: Fix a bug where shareable URLs don't work for dashboards that only admins can access

begelundmuller:


pjain1:
is it a valid use case where an admin dashboard will need to publicly shared ?

pjain1 (approved):

PR #5208, opened 2024-07-08
Public URL follow-ups

ericpgreen2:
This PR:
- Automatically hides any filtered dimensions
- Only specifies `metricsViewFields` if a subset of fields is selected
- Shows the actual API error message (e.g. "metrics view filter size exceeds limit (got 13108 bytes, but the limit is 1024 bytes)")
- Upon successfully creating a Public URL, automatically copies the URL to clipboard
- Switches terminology from "Public Link" to "Public URL" for copy consistency
- Uses `semibold` not `bold` for the header text

Addresses:
- [Slack request to minimize the size of the resultant JWT](https://rilldata.slack.com/archives/C02T907FEUB/p1720444728396409)
- [Notion UXQA doc](https://www.notion.so/rilldata/5c9758e637e54bf883afd81c4b575803?v=80eab730be874d208f2f3ea3d18b8e57)


web-admin/src/features/dashboards/share/ShareDashboardButton.svelte
```diff
@@ -50,3 +50,9 @@
     </Tabs>
   </PopoverContent>
 </Popover>
+
+<style lang="postcss">
+  h3 {
```
AdityaHegde, line 55: why not tailwind style use `font-semibold` ?

AdityaHegde (approved):

web-admin/src/features/dashboards/share/ShareDashboardButton.svelte
ericpgreen2: This is applying that Tailwind class. It's just separating the styles from the markup, which we've started to do across the app & IMO is a bit cleaner. But you're right that in this component, the rest of the styling is embedded in the markup, so it's not necessary.

PR #5209, opened 2024-07-08
Fix groups checks for mock users

begelundmuller:
Changes `IssueDevJWT` to `POST` so it can use JSON to correctly pass a list of groups (with `GET`, Orval serializes them as a comma-separated list)


pjain1 (approved):
proto changes look good, not sure if any more web change is needed as CI is failing

PR #5210, opened 2024-07-08
Docs: Update security policy docs with usergroup and wildcard examples

begelundmuller:


AndrewRTsao (changes requested):
Made a 2nd commit with the proposed cleanup and adding an example for showing `access`, which should be the most common use case for groups with security policies. @begelundmuller Could you validate quickly and then I'll approve?

AndrewRTsao (approved):

PR #5221, opened 2024-07-09
Add validate_deployments worker job and add ValidateConfig to provisioner interface

kaspersjo:


admin/provisioner/provisioner.go
```diff
@@ -15,6 +15,8 @@ type Provisioner interface {
 	AwaitReady(ctx context.Context, provisionID string) error
 	Update(ctx context.Context, provisionID string, newVersion string) error
 	CheckCapacity(ctx context.Context) error
+	ValidateConfig(ctx context.Context, provisionID string) (bool, error)
+	GetType() string
```
begelundmuller, line 19: nit: can it just be named `Type() string`?

admin/worker/reconcile_deployments.go
```diff
@@ -0,0 +1,127 @@
...
+
+	"github.com/rilldata/rill/admin"
+	"github.com/rilldata/rill/admin/database"
+	"github.com/rilldata/rill/runtime/pkg/observability"
+	"go.uber.org/zap"
+)
+
+const reconcileAllDeploymentsForProjectTimeout = 5 * time.Minute
+
+func (w *Worker) reconcileDeployments(ctx context.Context) error {
```
begelundmuller, line 16: Since "reconcile" means something else in the context of our runtime deployments, could we find another name for it here? Maybe "checkDeployments" or "validateDeployments" or something like that? If you want to stick with "reconcile", then add a docstring clarifying the difference.

begelundmuller (approved):

admin/provisioner/provisioner.go
kaspersjo: Make sense - renamed

admin/worker/reconcile_deployments.go
kaspersjo: Agree, that's an undesirable naming collision, I've changed it to `validate_deployments`

PR #5222, opened 2024-07-09
close response body on error

pjain1:


runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go
```diff
@@ -75,6 +75,7 @@ func (c *sqlConnection) QueryContext(ctx context.Context, query string, args []d
 	}
 
 	if resp.StatusCode != http.StatusOK {
+		resp.Body.Close()
```
pjain1, line 78: I wonder if we should add the response body to the return err as well, we can always look at druid logs for details so was not sure.

begelundmuller (approved):

PR #5223, opened 2024-07-09
feat: add image component

mindspank:
Adds a image component template

```yaml
type: component

image:
  url: https://pbs.twimg.com/profile_images/1753125727107706880/nVQlSOS8_400x400.jpg
  css:
    object-fit: "contain"
    opacity: 1
    saturation: 0
    blur: 5

### DIFF
--- a/web-common/src/features/templates/image/Image.svelte
+++ b/web-common/src/features/templates/image/Image.svelte
@@ -1,0 +1,39 @@
+<script lang="ts">
+  import {
+    ImageProperties,
+    ImageAdjust,
+  } from "@rilldata/web-common/features/templates/types";
+  import { V1ComponentSpecRendererProperties } from "@rilldata/web-common/runtime-client";
+
+  export let rendererProperties: V1ComponentSpecRendererProperties;
+
+  // Define default values for image properties
+  const defaultImageProperties: ImageProperties = {
+    url: "",
+    size: "contain",
+    adjust: {
+      opacity: 1,
+      blur: 0,
+      saturation: 1,
+    },
+  };
+
+  // Merge defaults with provided rendererProperties
+  $: imageProperties = {
+    ...defaultImageProperties,
+    ...rendererProperties,
+    adjust: {
+      ...defaultImageProperties.adjust,
+      ...rendererProperties.adjust,
+    },
+  } as ImageProperties;
+</script>
+
+<img
+  src={imageProperties.url}
+  style={`
+    object-fit: ${imageProperties.size}; 
+    opacity: ${imageProperties.adjust?.opacity}; 
+    filter: blur(${imageProperties.adjust?.blur}px) saturate(${imageProperties.adjust?.saturation});
+    `}
+/>
--- a/web-common/src/features/templates/image/Image.svelte
+++ b/web-common/src/features/templates/image/Image.svelte
@@ -29,6 +29,7 @@
 <img
   src={imageProperties.url}
   alt={"Dashboard Image"}
+  draggable="false"
   style={`
     object-fit: ${imageProperties.size}; 
     opacity: ${imageProperties.adjust?.opacity}; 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: web-common/src/features/templates/types.ts [hunk 1]
```diff
@@ -50,10 +50,26 @@ export interface MarkdownTemplateT {
...
 }
 
+export interface ImageAdjust {
+  opacity?: number;
+  blur?: number;
+  saturation?: number;
+}
+export interface ImageProperties {
+  url: string;
+  size?: string;
```
djbarnwal, line 60: Instead of this being a generic `string` type. Better to mention the strings that can be used explicitly here.
` fill | contain | cover | none | scale-down `

web-common/src/features/templates/image/Image.svelte
```diff
@@ -0,0 +1,38 @@
...
+    adjust: {
+      ...defaultImageProperties.adjust,
+      ...rendererProperties.adjust,
+    },
+  } as ImageProperties;
+</script>
+
+<img
+  src={imageProperties.url}
+  alt={"Dashboard Image"}
```
djbarnwal, line 31: Nit: Can directly use `"Dashboard Image"`

web-common/src/features/templates/types.ts
djbarnwal: The above typing will also help with autocomplete and type hints when we add that.

web-common/src/features/templates/image/Image.svelte
```diff
@@ -0,0 +1,38 @@
...
+  export let rendererProperties: V1ComponentSpecRendererProperties;
+
+  // Define default values for image properties
+  const defaultImageProperties: ImageProperties = {
+    url: "",
+    size: "contain",
+    adjust: {
+      opacity: 1,
+      blur: 0,
+      saturation: 1,
```
djbarnwal, line 14: We can also have something like this 

``` 
 interface ImageProperties {
    url: string;
    style?: Partial<CSSStyleDeclaration>;
  }

  const DEFAULT_IMAGE_PROPERTIES: ImageProperties = {
    url: "",
    style: {
      objectFit: "contain",
      opacity: "1",
      filter: "blur(0px) saturate(1)",
    },
  };
```
This way users can mention any style possible.

djbarnwal: `imageProperties` then can be updated as 

```
  $: imageProperties = {
    ...DEFAULT_IMAGE_PROPERTIES,
    ...rendererProperties,
    style: {
      ...DEFAULT_IMAGE_PROPERTIES.style,
      ...rendererProperties.style,
    },
  } as ImageProperties;

  $: imageStyle = Object.entries(imageProperties.style || {})
    .map(([key, value]) => `${kebabCase(key)}: ${value};`)
    .join(' ');
```

web-common/src/features/templates/image/Image.svelte
mindspank: Without a expression a1y compiler will complain sadly

web-common/src/features/templates/types.ts
```diff
@@ -50,10 +50,25 @@ export interface MarkdownTemplateT {
   markdown: MarkdownProperties;
 }
 
+export interface ImageAdjust {
+  opacity?: number;
+  blur?: number;
+  saturation?: number;
+}
```
djbarnwal, line 57: Looks like this can be removed?

web-common/src/features/templates/image/Image.svelte
```diff
@@ -0,0 +1,96 @@
...
+
+  export let rendererProperties: V1ComponentSpecRendererProperties;
+
+  const instanceId = get(runtime).instanceId;
+  const DEFAULT_IMAGE_PROPERTIES: ImageProperties = {
+    url: "",
+    css: {
+      "object-fit": "contain",
+      opacity: "1",
+      filter: "blur(0px) saturate(1)",
```
djbarnwal, line 16: Don't have to add this now as these are CSS defaults

djbarnwal (approved):
Looks good, left some nit comments. Approving so that you can merge post fixing those issues.

PR #5234, opened 2024-07-10
druid ingest executor

rohithreddykota:
## Summary

This PR introduces a druidIndexExecutor which is responsible for executing Druid index tasks using configurations from YAML files. The executor handles the ingestion of data from an object store into Druid, calculating necessary intervals, generating dynamic index JSON, and performing the ingestion process based on the model input and output configurations.

Example:
```yaml
kind: model

refresh:
  cron: 2 * * * *

connector: gcs
path: gs://hitech.rilldata.com/data-export/etl/bids/monthly/
pattern: 'yyyy=2006/mm=01/dd=02/HH=15'
gran: 1h
format: parquet
file_pattern: '.*\.parquet'
retry_period: 15m
max_retires: 3

incremental: true
output:
  connector: druid
  dataSource: demand_log_qa
  initial_look_back_period: 3h
  period_before: 1h
  quiet_period: 1h
  catchup: false
  max_work: 2h # maximum interval to index at a time
  coordinator_url: https://druid.ws1.hitech.rilldata.com/druid/coordinator/v1/datasources
  datasource_name: demand_log
  spec_json: >
    {
      "type": "index_parallel",
      "spec": {
        "ioConfig": {
          "type": "index_parallel",
          "inputSource": {
            "type": "google",
            "prefixes": {{ .prefixes | toJson }},
          },
          "inputFormat": {
            "type": "csv",
            "findColumnsFromHeader": true
          }
        },
        "tuningConfig": {
          "type": "index_parallel",
          "partitionsSpec": {
            "type": "dynamic"
          }
        },
        "dataSchema": {
          "dataSource": "%s",
          "timestampSpec": {
            "column": "timestamp",
            "format": "iso"
          },
          "transformSpec": {},
          "dimensionsSpec": {
            "dimensions": [
              {"type": "long", "name": "id"},
              "publisher",
              "domain",
              {"type": "double", "name": "bid_price"}
            ]
          },
          "granularitySpec": {
            "queryGranularity": "none",
            "rollup": false,
            "segmentGranularity": "day",
            "intervals": {{ .intervals | toJson }}
          }
        }
      }
    }
```


runtime/drivers/druid/druid.go
```diff
@@ -256,12 +257,19 @@ func (c *connection) AsObjectStore() (drivers.ObjectStore, bool) {
 
 // AsModelExecutor implements drivers.Handle.
 func (c *connection) AsModelExecutor(instanceID string, opts *drivers.ModelExecutorOptions) (drivers.ModelExecutor, bool) {
+	if opts.OutputHandle == c && opts.InputConnector == "gcs" {
```
begelundmuller, line 260: Connector names can be aliased (e.g. if you connect with two different service accounts for different buckets). So instead of checking `InputConnector` it should check `InputHandle.Driver()` instead.

runtime/drivers/druid/druid.go
```diff
@@ -182,9 +182,10 @@ func (d *driver) TertiarySourceConnectors(ctx context.Context, src map[string]an
 }
 
 type connection struct {
-	db     *sqlx.DB
-	config *configProperties
-	logger *zap.Logger
+	db         *sqlx.DB
+	config     *configProperties
+	logger     *zap.Logger
+	instanceID string
```
begelundmuller, line 188: This is used, but not assigned anywhere

runtime/drivers/druid/model_executor_index.go
```diff
@@ -0,0 +1,228 @@
...
+
+	startTime := calculateStartTime(previousExecutionTime, previousIntervalEndTime, periodBefore, outputProperties.Catchup, quietPeriod)
+	endTime := calculateEndTime(startTime, maxWork, quietPeriod)
+
+	if endTime.Before(startTime) {
+		return nil, fmt.Errorf("end time is before start time")
+	}
+
+	fmt.Println("==>inputProperties", inputProperties)
+	fmt.Println("==>outputProperties", outputProperties)
```
begelundmuller, line 102: Use `e.connection.logger`

runtime/drivers/druid/model_manager.go
```diff
@@ -0,0 +1,105 @@
...
+		return fmt.Errorf("missing property 'specJson'")
+	}
+	return nil
+
+}
+
+type ModelResultProperties struct {
+	DataSource              string `mapstructure:"datasource"`
+	PreviousExecutionTime   string `mapstructure:"previous_execution_time"`
+	PreviousIntervalEndTime string `mapstructure:"previous_interval_end_time"`
```
begelundmuller, line 74: Could it use `time.Time` (it serializes/deserializes naturally in JSON)?

runtime/drivers/druid/model_manager.go
```diff
@@ -0,0 +1,105 @@
...
+
+type ModelInputProperties struct {
+	Path          string `mapstructure:"path"`
+	Pattern       string `mapstructure:"pattern"`
+	Granularity   string `mapstructure:"gran"`
+	Format        string `mapstructure:"format"`
+	FilePattern   string `mapstructure:"file_pattern"`
+	RetriesPeriod string `mapstructure:"retry_period"`
+	MaxRetries    int    `mapstructure:"max_retries"`
+}
```
begelundmuller, line 20: So these are actually properties for an object store connector, not for the Druid connector. Ideally the name `druid.ModelInputProperties` would be reserved for input properties for Druid (i.e. models where Druid is the input and something else is the output – obviously we don't support that now, but could make sense in the future for e.g. models that export to S3).

Ideally the input properties for GCS would be defined in `gcs.ModelInputProperties` (or maybe `drivers.ObjectStoreModelInputProperties` if shared across multiple object store drivers), but I realize these are quite specific to this Druid driver. So maybe having it be an internal `gcsModelInputProperties` would make more sense.

begelundmuller (commented):
Some high-level questions:
1. I didn't go deep on the time manipulation logic here, but it seems like it's basically trying to emulate splits/partitions for incremental ingestion? If yes, it would be better to support that with native splits or something similar – and then the actual input/output properties could just template in values from the split currently being executed. Does that make sense? And do you have any inputs or new discoveries for things to think about here?
2. Did you look into using Druid's `INSERT` or `REPLACE` SQL commands instead? I know it's less mature, but would be sweet if we could get away with only supporting the SQL interface.

rohithreddykota:
> Some high-level questions:
> 
> 1. I didn't go deep on the time manipulation logic here, but it seems like it's basically trying to emulate splits/partitions for incremental ingestion? If yes, it would be better to support that with native splits or something similar – and then the actual input/output properties could just template in values from the split currently being executed. Does that make sense? And do you have any inputs or new discoveries for things to think about here?
> 2. Did you look into using Druid's `INSERT` or `REPLACE` SQL commands instead? I know it's less mature, but would be sweet if we could get away with only supporting the SQL interface.

1. You are right. It is kind of emulating splits for the incremental ingestion. Having native splits would definitely make things lot easier. From my understanding, splits should be able to serve all the options with a single query.
2. My initial idea was to add support for both sql based ingestion and specJSON based ingestion. The reason I want to add specJson based ingestion is that I can port the existing implementations into rill cloud. And another reasons are the [limitations](https://arc.net/l/quote/qnpchbzi).

rohithreddykota:
Lot changed since then. So, closing this. Will raise a new PR.

PR #5241, opened 2024-07-10
fix: Order of comparison measures in dimension table export

### LINKED ISSUE #5238: Order of columns in comparison exports does not match the UI
If I enable time comparisons and go to the dimension detail view, the comparison values for the current measure comes right after the current measure column. But in the exports, they come at the very end. When there are many measures, this feels like the comparison data is missing.

It seems like this will require two changes:
1. The UI uses the Aggregation API, but it uses the Comparison API for the exports. It should use the Aggregation API both places since it supports supports giving a total column order.
2. The UI currently puts the comparison measures last in the list requested in the Aggregation API. It should order the requested measures with the same order as they are shown in the UI.

AdityaHegde:
closes #5238

### DIFF
--- a/web-common/src/features/dashboards/dimension-table/ExportDimensionTableDataButton.svelte
+++ b/web-common/src/features/dashboards/dimension-table/ExportDimensionTableDataButton.svelte
@@ -115,7 +115,7 @@
 {#if includeScheduledReport && CreateScheduledReportDialog && showScheduledReportDialog && $scheduledReportsQueryArgs}
   <svelte:component
     this={CreateScheduledReportDialog}
-    queryName="MetricsViewComparison"
+    queryName="MetricsViewAggregation"
     queryArgs={$scheduledReportsQueryArgs}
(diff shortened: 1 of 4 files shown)


ericpgreen2:
Hmm, my export doesn't match the column order of the dimension table:

image

AdityaHegde:
Ah didnt realise backend doesnt honour the order either.

ericpgreen2:
To match the UI, the comparisons should come _after_ their associated column.

image

web-common/src/features/dashboards/dimension-table/ExportDimensionTableDataButton.svelte
```diff
@@ -43,6 +43,7 @@
   });
 
   $: scheduledReportsQueryArgs = getDimensionTableExportArgs(ctx);
+  $: console.log($scheduledReportsQueryArgs);
```
ericpgreen2, line 46: Leftover log

PR #5250, opened 2024-07-11
fix: handle nil claims in sql resolver

k-anshul:
IMO no claims will be set when resolving incremental model state so we should handle nil claims


esevastyanov (approved):
looks good!

k-anshul:
Thanks @esevastyanov 
Benjamin and I are also discussing on an alternative solution here : https://github.com/rilldata/rill/pull/5251

begelundmuller:
Closing in favor of https://github.com/rilldata/rill/pull/5251

PR #5251, opened 2024-07-11
Runtime: Ensure resolver claims are set

begelundmuller:


runtime/resolver.go
```diff
@@ -156,19 +156,26 @@ type ResolveResult struct {
...
 // Resolve resolves a query using the given options.
 func (r *Runtime) Resolve(ctx context.Context, opts *ResolveOptions) (ResolveResult, error) {
-	// Initialize the resolver
-	initializer, ok := ResolverInitializers[opts.Resolver]
-	if !ok {
-		return ResolveResult{}, fmt.Errorf("no resolver found for name %q", opts.Resolver)
+	// Prepare options
+	claims := opts.Claims
+	if claims == nil {
+		claims = &SecurityClaims{SkipChecks: true}
```
k-anshul, line 162: May be this should be set by the caller and this should return an error if `SecurityClaims` is nil.
This leaves a hole that in new integration we can forget to set it and all security checks will be skipped.

begelundmuller: Good idea

k-anshul (approved):

PR #5258, opened 2024-07-11
drop not null constraint on asset

k-anshul:


begelundmuller (commented):
Did you test the code that reads assets for rows that have nulls? Sometimes with `database/sql`, those need a `*string` or `sql.NullString` instead

begelundmuller (approved):

PR #5261, opened 2024-07-12
feat: API to get connected github repos

AdityaHegde:
To support `Connect to github` we need to know the connected repos so that user can select one of them.

Adding an API to fetch connected github repos: `ListGithubUserRepos`
It has to be called after `GetGithubUserStatus` with `HasAccess=true`


admin/server/github.go
```diff
@@ -237,6 +237,95 @@ func (s *Server) GetGithubRepoStatus(ctx context.Context, req *adminv1.GetGithub
...
+	for _, installation := range installations {
+		instToken, _, err := s.admin.Github.AppClient().Apps.CreateInstallationToken(ctx, *installation.ID, nil)
+		if err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+
+		// use the installation token to fetch the repos accessible to it
+		instClient := github.NewTokenClient(ctx, *instToken.Token)
+
+		req, err := instClient.NewRequest("GET", *installation.RepositoriesURL, nil)
```
k-anshul, line 297: Can directly use `instClient.Apps.ListUserRepos(ctx, *installation.ID, &github.ListOptions{})`

admin/server/github.go
```diff
@@ -237,6 +237,95 @@ func (s *Server) GetGithubRepoStatus(ctx context.Context, req *adminv1.GetGithub
...
+		PreferenceTimeZone:  user.PreferenceTimeZone,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to update user: %w", err)
+	}
+
+	client := github.NewTokenClient(ctx, token)
+
+	// use a client with user's token to get installations
+	installations, _, err := client.Apps.ListUserInstallations(ctx, &github.ListOptions{
```
k-anshul, line 279: 1. Can skip setting 0
2. Just reconfirming that `Page: 0` disables pagination because usually `Page` starts from 1 in the github APIs.

k-anshul: IIRC the github APIs also silently limit page_size to a max value as well.

proto/rill/admin/v1/api.proto
```diff
@@ -349,6 +349,10 @@ service AdminService {
     option (google.api.http) = {get: "/v1/github/user"};
   }
 
+  rpc ListGithubUserRepos(ListGithubUserReposRequest) returns (ListGithubUserReposResponse) {
+    option (google.api.http) = {get: "/v1/github/repositories/list"};
```
k-anshul, line 353: I think list can be skipped and the path can be changed to `/v1/github/repositories` to reflect that the repositories are for a user.

k-anshul (changes requested):

proto/rill/admin/v1/api.proto
AdityaHegde: `GetGithubRepoStatus` has the endpoint `/v1/github/repositories` already. Perhaps we should move that to `v1/github/repositories/status`

admin/server/github.go
AdityaHegde: I tried that 1st. Got `403 Resource not accessible by integration`

admin/server/github.go
AdityaHegde: Got it it was an initial autogenerated code that i didnt remove.

proto/rill/admin/v1/api.proto
k-anshul: Sorry I missed the `user` part. It should be `/v1/github/user/repositories` IMO

admin/server/github.go
k-anshul: Okay just realised that it works with [user access token](https://docs.github.com/en/rest/apps/installations?apiVersion=2022-11-28#list-repositories-accessible-to-the-user-access-token--fine-grained-access-tokens). In that case the above `client` itself will work and we don't need to generate a new client.

admin/server/github.go
```diff
@@ -237,6 +237,92 @@ func (s *Server) GetGithubRepoStatus(ctx context.Context, req *adminv1.GetGithub
...
+		PreferenceTimeZone:  user.PreferenceTimeZone,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to update user: %w", err)
+	}
+
+	client := github.NewTokenClient(ctx, token)
+
+	// use a client with user's token to get installations
+	installations, _, err := client.Apps.ListUserInstallations(ctx, nil)
```
k-anshul, line 279: Is it a conscious decision that we don't want to fetch the next page ?

admin/server/github.go
k-anshul: Also shouldn't the other request `instClient.NewRequest("GET", *installation.RepositoriesURL, nil)`  also respond with `403` unless the underlying API does something else.

AdityaHegde: Ah i was missing something earlier i think. I get the response now with `client`

admin/server/github.go
AdityaHegde: Added pagination. We should do the same in other places as well in a future PR

k-anshul (approved):
LGTM

PR #5263, opened 2024-07-12
feat: Add and edit github connection

AdityaHegde:
Notion: https://www.notion.so/rilldata/RD-RC-1-3-Project-Status-Page-updates-dd5daf216f5a4c05886b98bc20d4f8f5


web-admin/src/features/projects/github/GithubRepoUpdater.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  public readonly status = derived(
+    [this.userStatus, this.userRepos],
+    ([userStatus, userRepos]) => {
+      if (userStatus.isFetching || userRepos.isFetching) {
+        return {
+          isLoading: true,
+          error: undefined,
+        };
+      }
```
ericpgreen2, line 36: We should be consistent with terminology. If reading `isFetching`, then we should return `isFetching`. If reading `isLoading`, then we should return `isLoading`.

web-admin/src/features/projects/github/GithubRepoSelectionDialog.svelte
```diff
@@ -0,0 +1,135 @@
...
+    AlertDialogContent,
+    AlertDialogDescription,
+    AlertDialogFooter,
+    AlertDialogHeader,
+    AlertDialogTitle,
+    AlertDialogTrigger,
+  } from "@rilldata/web-common/components/alert-dialog";
+  import { Button } from "@rilldata/web-common/components/button";
+  import Github from "@rilldata/web-common/components/icons/Github.svelte";
+  import Select from "@rilldata/web-common/components/forms/Select.svelte";
```
ericpgreen2, line 19: A little out-of-scope of this PR, but it'd be nice to deprecate this in favor of [ShadCN's Select component](https://www.shadcn-svelte.com/docs/components/select)

web-admin/src/features/projects/github/GithubConnection.ts
ericpgreen2, line 1: IMO this extra layer of abstraction isn't necessary & adds complexity. 

Is there something wrong with using vanilla TanStack Query in the `ProjectGitHubConnection.svelte` component? That's the pattern we're using most frequently throughout the codebase.

web-admin/src/features/projects/github/GithubRepoUpdater.ts
ericpgreen2, line 1: Similarly, is there a reason to be adding this abstraction on top of vanilla TanStack Query? To update the GitHub repo seems like a vanilla mutation and shouldn't require too much complexity. I wonder if the bugginess I shared via Jam is a derivative of this complexity.

_If_ we're to go down this route, I'd expect a "GithubRepoUpdater" to include a method to actually update the Github repo, but that looks to be happening outside of this abstraction in the `GithubRepoSelectionDialog.svelte` component.

ericpgreen2:
Some initial UXQA –

1. Editing the GitHub repo is buggy. The GitHub repo rendered in `ProjectGitHubConnection.svelte` never updates. And the listed resources appear to be linked to the previous (t-1) GitHub repo, not the current (t) GitHub repo. I'll send a Jam via Slack.
2. The dropdown should be initialized with the current GitHub repository (in this case, `rill-examples`)
    image
3. @ericokuma, it looks like we need to add a field to configure the `subpath`. FWIW, here's how Vercel handles selecting a "root directory":
    image

ericpgreen2 (changes requested):

web-admin/src/features/projects/github/GithubRepoUpdater.ts
AdityaHegde: This endpoint can be slow if there are a lot of repos and orgs connected. I want to avoid refetching it on window focus all the time, but only when `connect to github` or `choose other repos` is clicked.

Both have a check to only refetch if it was explicitly asked for (basically the `connecting` check).

Naming this `GithubRepoUpdater` is just a remnant of some old code.

web-admin/src/features/projects/github/GithubRepoSelectionDialog.svelte
```diff
@@ -0,0 +1,140 @@
...
+    void invalidateRuntimeQueries(
+      queryClient,
+      $projectQuery.data.prodDeployment.runtimeInstanceId,
+    );
+    open = false;
+  }
+
+  function handleVisibilityChange() {
+    if (document.visibilityState !== "visible") return;
+    void githubReposConnection.focused();
```
ericpgreen2, line 72: I had to read this function implementation to know what it was doing. If we put the Query Observer in this component, we could use `$githubUserReposQuery.refetch()`, which would be more explicit.

web-admin/src/features/projects/github/GithubRepoUpdater.ts
ericpgreen2: Gotcha. The vanilla TanStack approach would be to use `$githubQuery.refetch()` in the component itself. Might be simpler?

ericpgreen2 (changes requested):
Some more UXQA:
- All user-facing text that says "github" or "Github" should be "GitHub"
- After triggering the connect-to-GitHub flow via the UI, it looks like this page should say "... continue setup in your original tab."
    image
- I get back 150 repos from the `ListGithubUserRepos` API, but the `Select` component only shows me the first 100.

web-admin/src/features/projects/github/ProjectGithubConnection.svelte
```diff
@@ -0,0 +1,128 @@
...
+  });
+  const userStatus = githubConnection.userStatus;
+
+  function connectToGithub() {
+    confirmDialogOpen = true;
+  }
+
+  function confirmConnectToGithub() {
+    confirmDialogOpen = false;
+    void githubConnection.check();
```
ericpgreen2, line 42: Here as well, I had to read the function implementation to understand what was going on. What about just 
```
const githubUserStatusQuery = createAdminServiceGetGithubUserStatus();

...

window.open(
  $githubUserStatusQuery.data?.grantAccessUrl,
  "_blank",
)
```

web-admin/src/features/projects/github/GithubRepoUpdater.ts
AdityaHegde: We are refetching the query right? The idea is to hide as much code outside of components as possible. We still need the check if we popped a page to select github repos or not.

This will also hopefully allow us to abstract such code out. But it will have to be follow ups

ericpgreen2:
Writing down what we discussed live:
- The motivation behind the `GitHubConnection` classes is to track the state of the connect-to-GitHub flow (which happens in a new tab). We want to be smart about when we refetch `GetGitHubUserStatus` and `ListGitHubUserRepos`, each of which can take a couple seconds. The refetches should happen after the connect-to-GitHub flow on page re-focus. However, these `GitHubConnection` classes are hard to read.
- We should know why updating the GitHub repo leads to a page flash. We figured out it's the result of a `GetProject` request that omits cookies & returns a 401 error. But why does the request not include cookies?

admin/server/github.go
```diff
@@ -836,11 +836,18 @@ func (s *Server) fetchReposForInstallation(ctx context.Context, client *github.C
 			if repo.Owner != nil {
 				owner = fromStringPtr(repo.Owner.Login)
 			}
+			var branch string
+			if repo.DefaultBranch != nil {
```
k-anshul, line 840: Just confirming, did you check if either one of them is always set ? The github APIs though document same repository schema but I have seen usually they miss field in one API or other.

k-anshul (approved):

k-anshul:
Approving from backend POV

admin/server/github.go
AdityaHegde: Good question, just checked [the docs](https://docs.github.com/en/rest/apps/installations?apiVersion=2022-11-28#list-repositories-accessible-to-the-user-access-token) for this API. `default_branch` is in required section.

ericpgreen2:
FYI, @jkhwu updated the designs to remove a couple cases of indentation. E.g., see these new mocks:
image
image

web-admin/src/features/projects/github/ConnectToGithubButton.svelte
```diff
@@ -0,0 +1,86 @@
...
+
+  export let open = false;
+  export let loading = false;
+  export let onContinue: () => void;
+
+  let githubRepoCreated = false;
+
+  function openGithubRepoCreator() {
+    // we need a popup window so we cannot use href
+    window.open("https://github.com/new", "", "width=1024,height=600");
```
ericpgreen2, line 22: Nit: we can name the window (2nd parameter) `githubWindow`

web-admin/src/features/projects/github/GithubData.ts
```diff
@@ -0,0 +1,144 @@
...
+    await waitUntil(() => !get(this.userStatus).isFetching);
+    const userStatus = get(this.userStatus).data;
+    if (!userStatus?.grantAccessUrl) {
+      return;
+    }
+
+    this.openGithubConnectWindow(userStatus.grantAccessUrl);
+  }
+
+  public async refetch() {
```
ericpgreen2, line 93: I think this function is only used internally – so it could be `private`?

ericpgreen2 (commented):
UXQA: 
1. The "Are you sure you want to connect?" dialog flashes & proceeds without waiting for confirmation from the user. [See Jam](https://jam.dev/c/9353aa1d-1e70-49c9-89db-8fb8da8de07b)
2. After going through the "choose other repos" flow, I get brought to this page that references the CLI.
    image
    Possibly related, I've been wondering – when the connect-to-GitHub flow is complete, can we somehow communicate this to the `/status` page, so it knows exactly when to refetch the list of repositories? Polling doesn't feel exactly right, (but the way you've contained the polling looks good).

web-admin/src/features/projects/github/GithubData.ts
```diff
@@ -0,0 +1,144 @@
...
+    if (this.windowCheckTimer) clearInterval(this.windowCheckTimer);
+    this.userPromptWindow = window.open(url, "", "width=1024,height=600");
+
+    // periodically check if the new window was closed
+    this.windowCheckTimer = setInterval(() => {
+      if (!this.userPromptWindow.closed) return;
+      clearInterval(this.windowCheckTimer);
+      this.windowCheckTimer = undefined;
+      this.userPromptWindow = undefined;
+      void this.refetch();
```
ericpgreen2, line 133: This looks like it's refetching when the `userPromptWindow` is _open_. Shouldn't refetch only once the `userPromptWindow` is _closed_?

ericpgreen2:
UXQA –
1. After the "Choose another repo" flow, I see this page that references the CLI (cc @ericokuma):
    image
2. I created a blank repo, then I hit a 404 when I selected it and clicked "continue" in the UI:
    image
    image

admin/server/github.go
```diff
@@ -286,6 +305,99 @@ func (s *Server) ListGithubUserRepos(ctx context.Context, req *adminv1.ListGithu
...
+
+	// Find project
+	proj, err := s.admin.DB.FindProjectByName(ctx, req.Organization, req.Project)
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+
+	claims := auth.GetClaims(ctx)
+	if !claims.ProjectPermissions(ctx, proj.OrganizationID, proj.ID).ManageProject {
+		return nil, status.Error(codes.PermissionDenied, "does not have permission to delete project")
```
k-anshul, line 326: Nit: Incorrect error msg `delete project`

admin/pkg/archive/archive.go
```diff
@@ -0,0 +1,126 @@
+package archive
```
k-anshul, line 1: Nit: Should this be part of `runtime/pkg` since this is also used in runtime ?

admin/server/github.go
```diff
@@ -853,9 +972,294 @@ func (s *Server) fetchReposForInstallation(ctx context.Context, client *github.C
...
+	var ghRepo *git.Repository
+	empty := false
+	ghRepo, err = git.PlainClone(gitPath, false, &git.CloneOptions{
+		URL:           repo,
+		Auth:          gitAuth,
+		ReferenceName: plumbing.NewBranchReferenceName(branch),
+		SingleBranch:  true,
+	})
+	if err != nil {
+		if errors.Is(err, transport.ErrEmptyRemoteRepository) {
```
k-anshul, line 1009: Nit : Can simplify the if-else block by switching the if condition

admin/server/github.go
```diff
@@ -853,9 +972,294 @@ func (s *Server) fetchReposForInstallation(ctx context.Context, client *github.C
...
+
+	// generate a temp dir to extract the archive
+	dir, err := os.MkdirTemp(os.TempDir(), "extracted_archives")
+	if err != nil {
+		return err
+	}
+	defer os.RemoveAll(dir)
+
+	// use a subfolder for working with git
+	gitPath := filepath.Join(dir, "proj")
```
k-anshul, line 987: Can this be simplified using different directories for downloading archive and project directory ?
In that case we don't need to pass `dir` to `copyData` as well and it can just download it in any temp directory.

admin/server/github.go
```diff
@@ -853,9 +972,294 @@ func (s *Server) fetchReposForInstallation(ctx context.Context, client *github.C
...
+			return err
+		}
+	}
+
+	err = copyData(dir, projPath)
+	if err != nil {
+		return fmt.Errorf("failed to copy data: %w", err)
+	}
+
+	// add back the older gitignore contents if present
```
k-anshul, line 1055: If we are anyways overwriting all other files then I think it should be okay to also overwrite `.gitignore` file ?

k-anshul (changes requested):

admin/server/github.go
AdityaHegde: This approach of extracting to the intended target avoids copy after extracting. With any temp directory we would have additional copy to the target directory (either root of the github repo or a selected subpath).

admin/server/github.go
AdityaHegde: This was a requirement to not completely replace the existing file. But rather append like we do with `rill init`

admin/server/github.go
k-anshul: May be I am missing something. 
1) For archive case the src is a zipped archive and destination is a project directory. We need to unzip the file and put it into its corresponding location in project directory. So it should be same whether src and destination are different directories or not.
2) For git case there is an explicit copy step from `tmp/extracted_archives/src` to `tmp/extracted_archives/proj`

2 got me thinking. For monorepos we seem to be copying `tmp/extracted_archives/src/subpath` to `tmp/extracted_archives/proj/subpath` (projPath) and then we do `git push` from this path. Will this work if rest of the contents especially `.git` at the root is not copied ?

admin/server/github.go
k-anshul: IIRC in `rill init` we did not delete existing files so it made sense to not remove existing `.gitignore`.
Here we are deleting all existing files as well.

AdityaHegde: @nishantmonu51 @ericokuma  Thoughts on the above?

admin/server/github.go
```diff
@@ -853,9 +1000,298 @@ func (s *Server) fetchReposForInstallation(ctx context.Context, client *github.C
...
+
+	// generate a temp dir to extract the archive
+	dir, err := os.MkdirTemp(os.TempDir(), "dest_git_repos")
+	if err != nil {
+		return err
+	}
+	defer os.RemoveAll(dir)
+
+	// use a subfolder for working with git
+	gitPath := filepath.Join(dir, "proj")
```
k-anshul, line 1015: The subfolder should not be required now. We can directly use the `temp_dir`.

admin/server/github.go
```diff
@@ -853,9 +1000,298 @@ func (s *Server) fetchReposForInstallation(ctx context.Context, client *github.C
 	return repos, nil
 }
 
+func (s *Server) pushToGit(ctx context.Context, copyData func(projPath string) error, repo, branch, subpath, token string, force bool) error {
+	ctx, cancel := context.WithTimeout(ctx, archivePullTimeout)
+	defer cancel()
+
+	// generate a temp dir to extract the archive
+	dir, err := os.MkdirTemp(os.TempDir(), "dest_git_repos")
```
k-anshul, line 1008: Given this is used for both upload and git flow so we can just change the name to `project` or similar?

admin/server/github.go
```diff
@@ -853,9 +1000,298 @@ func (s *Server) fetchReposForInstallation(ctx context.Context, client *github.C
...
+
+// copyFromSrcGit clones a repo, branch and a subpath and copies the content to the projPath
+// used to switch a project to a new github repo connection
+func copyFromSrcGit(projPath, repo, branch, subpath, token string) error {
+	srcGitPath, err := os.MkdirTemp(os.TempDir(), "src_git_repos")
+	if err != nil {
+		return err
+	}
+
+	// projPath is the target for extracting the archive
```
k-anshul, line 1207: misplaced comment ?

k-anshul (changes requested):
Looks good from backend POV apart from few nits.

k-anshul (approved):
Approving from backend POV

web-admin/src/features/projects/github/GithubConnectionUpdater.ts
```diff
@@ -0,0 +1,121 @@
...
+  private defaultBranch = "";
+
+  public constructor() {}
+
+  public init(githubUrl: string, subpath: string, branch: string) {
+    this.githubUrl.set(githubUrl);
+    this.subpath.set(subpath);
+    this.branch.set(branch);
+    this.isConnected = !!githubUrl;
+  }
```
ericpgreen2, line 33: It'd be more idiomatic to implement the constructor, rather than create a separate `init` function

web-admin/src/features/projects/github/GithubData.ts
ericpgreen2, line 1: I feel like the separation of concerns between the `GitHubData` and `GitHubConnectionUpdater` classes is blurry. It might be a matter of naming & comments, or maybe all the "update" methods (`startRepoSelection`, `reselectRepos`, `openGitHubConnectWindow`) should live on the "Updater" class?

web-admin/src/features/projects/github/GithubRepoSelectionDialog.svelte
```diff
@@ -0,0 +1,197 @@
...
+
+  async function updateGithubUrl(force: boolean) {
+    if (
+      !(await githubConnectionUpdater.update({
+        organization,
+        project,
+        force,
+        instanceId: $projectQuery.data?.prodDeployment?.runtimeInstanceId ?? "",
+      }))
+    ) {
```
ericpgreen2, line 90: Nit: Embedding the mutation into the conditional check makes this hard to read. I'd recommend declaring a self-describing boolean, like `didUpdate`.

web-admin/src/features/projects/github/ProjectGithubConnection.svelte
```diff
@@ -0,0 +1,123 @@
...
+    getRepoNameFromGithubUrl($proj.data.project.githubUrl);
+  $: subpath = $proj.data?.project?.subpath;
+  $: githubLastSynced = useDashboardsLastUpdated(
+    $runtime.instanceId,
+    organization,
+    project,
+  );
+
+  const githubData = new GithubData();
+  setGithubData(githubData);
```
ericpgreen2, line 34: Should `setGitHubData` happen in the constructor?

web-admin/src/features/projects/github/GithubConnectionUpdater.ts
AdityaHegde: Ya that was ideal earlier. But it is simpler to do this than create a new instance everytime.

web-admin/src/features/projects/github/GithubData.ts
AdityaHegde: `GitHubData` is meant to be a wrapper around the user status and repos APIs. `GitHubConnectionUpdater` is only around the mutation to update. Both basically move code out of components as much as possible.

ericpgreen2 (approved):
Approving from the frontend POV 👍

PR #5265, opened 2024-07-12
Left pad parsed auth token bytes

begelundmuller:
Access tokens in Rill follow the format `rill_<type>_<base62(payload)>` where the payload contains the token's ID (16 bytes) and randomly generated secret (24 bytes). 

The `base62` implementation drops leading `0x00` bytes, which means that on average 1 out of every 256 tokens can't be parsed. This PR fixes this issue by left-padding the parsed payload with `0x00`.

It also adds a test that generates 100000 random tokens and tests that they parse correctly to prevent such errors from occurring again.

### DIFF
--- a/admin/pkg/authtoken/authtoken.go
+++ b/admin/pkg/authtoken/authtoken.go
@@ -82,8 +82,10 @@ func FromString(s string) (*Token, error) {
 		return nil, ErrMalformed
 	}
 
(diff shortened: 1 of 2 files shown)


pjain1 (commented):
Great catch!

pjain1 (approved):

PR #5272, opened 2024-07-15
Runtime: Errored instance log per instance

k-anshul:


admin/worker/deployments_health_check.go
```diff
@@ -88,21 +87,46 @@ func (w *Worker) deploymentHealthCheck(ctx context.Context, d *database.Deployme
...
+		if resp.ConnCacheError != "" {
+			f = append(f, zap.String("conn_cache_error", resp.ConnCacheError))
+		}
+		if resp.MetastoreError != "" {
+			f = append(f, zap.String("metastore_error", resp.MetastoreError))
+		}
+		if resp.NetworkError != "" {
+			f = append(f, zap.String("network_error", resp.NetworkError))
+		}
+		w.logger.Error("deploymentsHealthCheck: runtime is unhealthy", f...)
```
begelundmuller, line 104: nit: can we put `deployment health check: ...` instead of `deploymentsHealthCheck` in the logs? The other jobs do that

begelundmuller (approved):

PR #5285, opened 2024-07-16
CLI: Add `sudo project hibernate` command

begelundmuller:


kaspersjo (approved):

nishantmonu51 (approved):
How would an admin/user unhibernate the project ?

begelundmuller:
> How would an admin/user unhibernate the project ?

They need to run `rill project reconcile` – instructions are shown on the project status page when the project is hibernated, so should be easy for them to resolve.

PR #5286, opened 2024-07-16
fix(dimension table): Re-enable comparison values on selected measure when sorting by dimension

AdityaHegde:


AdityaHegde:
This is intended behaviour

PR #5292, opened 2024-07-16
feat: multiselect alert measure

mindspank:
Allows for multi measure alerts.

Todo:
* Alert Preview queries for the Data and Criteria tab still fires when there is no valid measure or incomplete criteria


web-common/src/components/forms/Multiselectcombobox.svelte
ericpgreen2, line 1: A few things given this will be the first instance of this component in the application:
- Nit: This is a vanilla multi-select, not a "combobox", because it doesn't include a freeform text input & autocomplete. So we should rename this file to `MultiSelect.svelte`
- ShadCN Svelte doesn't have a multi-select or a multi-select combobox, but it does recommend [this guide](https://formsnap.dev/docs/recipes/bits-ui-select#multiple-select) for how to do a multi-select with BitsUI (which ShadCN Svelte is built upon).
- We should use the ShadCN `DropdownMenu` in place of the `WithTogglableFloatingElement` and `Menu` components.

mindspank: Gotcha, it technically does have free form input and autocomplete if the main input is not set to readonly though. Don't think we ever used it anywhere in the app though so happy to get rid of it

ericpgreen2 (approved):
Code looks good. Approving so you can merge at your discretion.

web-common/src/features/alerts/criteria-tab/CriteriaGroup.svelte
```diff
@@ -19,7 +19,6 @@
   function handleAddCriteria() {
     $form["criteria"] = $form["criteria"].concat({
       ...getEmptyMeasureFilterEntry(),
```
AdityaHegde, line 21: Can just be `$form["criteria"].concat(getEmptyMeasureFilterEntry())`

web-common/src/features/alerts/extract-alert-form-values.ts
```diff
@@ -66,7 +66,14 @@ export function extractAlertFormValues(
...
 
   return {
-    measure: measures[0]?.name ?? "",
+    measures: measures
+      .filter((m) => !m.comparisonDelta && !m.comparisonRatio)
+      .map((m) => {
+        const mappedMeasure = metricsViewSpec.measures
+          ? metricsViewSpec.measures.find((mm) => mm.name === m.name)
+          : null;
+        return { value: m.name, label: mappedMeasure?.label };
```
AdityaHegde, line 75: Should this fallback to expression when label is empty?

web-common/src/features/alerts/BaseAlertForm.svelte
```diff
@@ -73,12 +73,6 @@
     if (!name) return;
     $form.name = name;
   }
-
-  $: measure = $form.measure;
```
AdityaHegde, line 77: We should still have this i feel. Right now if you deselect a measure with criteria, the criteria form shows empty value but i think the `$form.criteria` still has the old measure value. So I can still create the alert with a measure filter not actually selected. Which leads to a query error,
```
Alert check failed: failed to resolve alert: failed to execute query: failed to compile 'having': name "total_records" in expression is not a dimension or measure available in the current context
```

Perhaps we should unset the measure name in the criteria if it is not selected in the list of measures?

mindspank:
Closing in favour of new PR that also includes multi dimensional selects

PR #5294, opened 2024-07-16
Runtime : Snowflake connector for clickhouse

k-anshul:
Sample model:

```
refresh:
  cron: "0 * * * *"

type: model
connector: snowflake
sql: >
  select * 
  from SNOWFLAKE_SAMPLE_DATA.TPCH_SF1.CUSTOMER 
  {{ if incremental }} WHERE C_CUSTKEY > '{{ .state.max_c_custkey }}' {{ end }}
  order by c_custkey asc 
  limit 1001

incremental: true
state:
  sql: SELECT MAX(C_CUSTKEY) as max_c_custkey FROM model


stage:
  connector: s3
  path: s3://<bucket>/snow

output:
  connector: clickhouse
  incremental_strategy: append
  materialize: true
  
  ```

closes https://github.com/rilldata/rill-private-issues/issues/480

### CODE REVIEW

**Comment by k-anshul** [comment]
closes https://github.com/rilldata/rill-private-issues/issues/480

File: runtime/drivers/models.go [hunk 1]
```diff
@@ -32,6 +32,8 @@ type ModelExecutorOptions struct {
 	InputHandle      Handle
 	InputConnector   string
 	InputProperties  map[string]any
+	StageConnector   string
+	StageProperties  map[string]any
```
begelundmuller, line 36: Can we omit these here? Since the use of the staging connector is contained in the reconciler, and the individual model executor should only need to deal with moving data from the provided input connector to the output connector. 

Maybe it can't be avoided eventually, but it would be nice to remove it to keep model executor implementation simpler and not give an indication of too many special cases that needs handling.

runtime/reconcilers/model.go
```diff
@@ -598,6 +610,9 @@ func (r *ModelReconciler) resolveIncrementalState(ctx context.Context, mdl *runt
 // acquireExecutor acquires a ModelExecutor capable of executing a model with the given execution options.
 // It handles acquiring and setting opts.InputHandle and opts.OutputHandle.
 func (r *ModelReconciler) acquireExecutor(ctx context.Context, opts *drivers.ModelExecutorOptions) (string, drivers.ModelExecutor, func(), error) {
+	if opts.StageConnector != "" {
+		return r.acquireExecutorWithStage(ctx, opts)
+	}
```
begelundmuller, line 615: It would be nice to break this out from `acquireExecutor` because `acquireExecutorWithStage` calls `acquireExecutor`, which makes the logic more convoluted than necessary.

runtime/reconcilers/model.go
```diff
@@ -276,6 +281,13 @@ func (r *ModelReconciler) Reconcile(ctx context.Context, n *runtimev1.ResourceNa
...
 
+	// Open executor for the new output
+	executorConnector, executor, release, err := r.acquireExecutor(ctx, opts)
+	if err != nil {
+		return runtime.ReconcileResult{Err: err}
+	}
+	defer release()
+
 	// Build the output
 	execRes, execErr := executor.Execute(ctx)
```
begelundmuller, line 292: Maybe combine these two steps into a `r.execute` function that handles the staging and non-staging cases?

runtime/reconcilers/model.go
```diff
@@ -644,6 +659,94 @@ func (r *ModelReconciler) acquireExecutor(ctx context.Context, opts *drivers.Mod
 	return executorName, e, release, nil
 }
 
+func (r *ModelReconciler) acquireExecutorWithStage(ctx context.Context, opts *drivers.ModelExecutorOptions) (string, drivers.ModelExecutor, func(), error) {
```
begelundmuller, line 662: Since this function actually does execution, not just acquisition (which sounds "light"), may be worth refactoring the name

runtime/reconcilers/model.go
```diff
@@ -644,6 +659,94 @@ func (r *ModelReconciler) acquireExecutor(ctx context.Context, opts *drivers.Mod
...
+		rel()
+		return "", nil, nil, err
+	}
+	rel()
+
+	rc, rr, err := r.C.AcquireConn(ctx, res.Connector)
+	if err != nil {
+		return "", nil, nil, err
+	}
+	defer rr()
```
begelundmuller, line 706: Maybe name it `sc` since it's gonna be the staging connector?

runtime/reconcilers/model.go
```diff
@@ -644,6 +659,94 @@ func (r *ModelReconciler) acquireExecutor(ctx context.Context, opts *drivers.Mod
...
+		}
+		defer rr()
+		if mm, ok := rc.AsModelManager(r.C.InstanceID); ok {
+			// May block reconcile. May be add a timeout and run in background ?
+			err = mm.Delete(context.Background(), res)
+			if err != nil {
+				r.C.Logger.Warn("failed to clean up stage output", zap.Error(err))
+			}
+		}
+	}
```
begelundmuller, line 746: Yeah the use of `context.Background()` in a reconciler is quite dangerous and something we need to avoid because the controller assumes that reconcilers are "cooperative" i.e. when a `ctx` is cancelled (for example because the resource YAML is edited, or the process is shutting down), it waits for all currently running reconciles to finish gracefully.

In order to keep things simple (a timeout can also be problematic, especially for large data where cleanup can take a while), could we maybe just let garbage stay in the staging connector in the case of cancels – and instead clean it up on the next ingestion run? (I.e. also clear the staging location before writing to it.)

runtime/drivers/warehouse.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"context"
+	"net/url"
+)
+
+type Warehouse interface {
+	// QueryAsFiles downloads results into files and returns an iterator to iterate over them
+	QueryAsFiles(ctx context.Context, props map[string]any, opt *QueryOption) (FileIterator, error)
+	Export(ctx context.Context, props map[string]any, store ObjectStore, outputLocation string) (*ExportResult, error)
```
begelundmuller, line 11: Maybe I'm missing a detail, but the `Export` function here looks very close to the model executor capabilities. 

Can it be removed and implemented as an `objectStoreToSelf` executor on the warehouse?

begelundmuller (changes requested):

runtime/drivers/clickhouse/model_executor_s3_self.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+	sb.WriteString(")")
+	return sb.String(), nil
+}
+
+type s3ConfigProperties struct {
+	AccessKeyID     string `mapstructure:"aws_access_key_id"`
+	SecretAccessKey string `mapstructure:"aws_secret_access_key"`
+	SessionToken    string `mapstructure:"aws_access_token"`
+}
```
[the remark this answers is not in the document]
k-anshul: I also thought so. But that will introduce dependency between s3 and other drivers that needs it so refrained from adding it. 
Do you think it is okay to add that dependency ?

runtime/reconcilers/model.go
k-anshul: Thanks for clarifying the issues with `context.Background()`.
There are some issues with cleaning in next ingestion run.
1. The warehouses write to a temporary directory within staging path. The idea is to avoid conflicts in case the same staging path is used by another connector. Snowflake also recommends writing to an empty directory.
2. We can consider storing the temporary path in model state and clear it but it will add extra latency for the next ingestion.

How about we leave the data as it is during cancelation and recommend setting a short data retention on the bucket? IIRC athena has similar constructs as well where user needs to cleanup `output_location` manually.

runtime/drivers/warehouse.go
k-anshul: So we have these options : 
1. Implement `selfToObjectStore` (I think it should be called `selfToObjectStore` and not `objectStoreToSelf`). This has to be done for every warehouse.
2. Or implement one `warehouseToSelfExecutor` in `S3` and implement `Export` method on each of the `Warehouse` which is how it is done right now.
Both seems equivalent to me. 
I can implement 1 if we want to emphasise on the usage of `ModelExecutors`. What do you suggest ?
Also in that case I think we can also refactor `QueryAsFiles` to `selfToFile`.

runtime/reconcilers/model.go
k-anshul: Okay slightly refactored the code so that we can call this `executeWithStage`

runtime/drivers/clickhouse/model_executor_s3_self.go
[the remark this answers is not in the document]
k-anshul: We can specify this in drivers pkg as well or a `drivers/config` pkg which is imported everywhere to avoid direct dependency

runtime/reconcilers/model.go
begelundmuller: Makes sense and yeah that sounds like a good idea

runtime/drivers/clickhouse/model_executor_s3_self.go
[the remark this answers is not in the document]
begelundmuller: I think it's okay for some drivers to import other drivers – after all, the `runtime/drivers/s3` package is supposed to be our place for S3-related logic, so if ClickHouse has some S3-specific logic it makes sense that it imports our package for S3 logic.

The only risk is if you get into cyclic dependencies, but hopefully that doesn't happen (e.g. Clickhouse uses S3, but S3 doesn't use ClickHouse) and if it does we can handle it on a case-by-case basis.

runtime/drivers/warehouse.go
begelundmuller: I think option 1 sounds good. My thinking is that it is a nice principle to have drivers/executors be a one-way relationship – for example, currently, warehouses are aware of S3 and GCS because they are handled in the `Export` function *and* S3/GCS are aware of warehouses because they have an executor for them. But if we make this refactor, then warehouses will be aware of S3/GCS in the executor, but S3/GCS will not be aware of warehouses at all.

Additionally, the `Export` implementation already looks very close to an executor so should be an easy change to make.

Regarding `QueryAsFiles`, yeah that might make sense, though let's keep that for a follow up if it makes sense.

runtime/drivers/clickhouse/model_executor_s3_self.go
```diff
@@ -0,0 +1,82 @@
...
+		ModelName:        e.opts.ModelName,
+		InputConnector:   e.opts.InputConnector,
+		InputProperties:  propsMap,
+		OutputConnector:  e.opts.OutputConnector,
+		OutputProperties: e.opts.OutputProperties,
+		Priority:         e.opts.Priority,
+		Incremental:      e.opts.Incremental,
+		IncrementalRun:   e.opts.IncrementalRun,
+	}
+	executor := &selfToSelfExecutor{c: e.c, opts: opts}
```
begelundmuller, line 60: Since it rewrites to self-to-self, shouldn't the rewritten options have `InputConnector: e.opts.OutputConnector`?

runtime/drivers/s3/model_manager.go
```diff
@@ -0,0 +1,113 @@
...
+	p := &ModelResultProperties{}
+	if err := mapstructure.Decode(res.Properties, p); err != nil {
+		return err
+	}
+	u, err := url.Parse(p.Path)
+	if err != nil {
+		return err
+	}
+	base, _ := doublestar.SplitPattern(strings.TrimPrefix(u.Path, "/"))
+	return deleteObjectsInPrefix(ctx, cfg, u.Host, base)
```
begelundmuller, line 45: Not sure, but can the `ModelResultProperties.Path` be a fixed path to a directory, file, or list of files (instead of a glob pattern)? Since glob patterns are for selecting files to ingest, not files to output. Also, by deleting the full prefix here it might delete files that actually don't match the full glob pattern, which seems weird (even if it doesn't matter in practice).

runtime/drivers/s3/model_manager.go
```diff
@@ -0,0 +1,113 @@
...
+			continuationToken = out.NextContinuationToken
+		} else {
+			break
+		}
+	}
+
+	return nil
+}
+
+func (c *Connection) awsConfig(ctx context.Context, awsRegion string) (aws.Config, error) {
```
begelundmuller, line 95: Can it use `getAwsSessionConfig(...).Config` instead so we don't have duplicate functions for initializing AWS configs? (Note the func is renamed to `newSessionForBucket` in my glob PR.)

runtime/drivers/snowflake/model_executor_self_objectstore.go
```diff
@@ -0,0 +1,123 @@
...
+	} else if e.c.configProperties.DSN != "" { // get from driver configs
+		dsn = e.c.configProperties.DSN
+	} else {
+		return "", fmt.Errorf("the property 'dsn' is required for Snowflake. Provide 'dsn' in the YAML properties or pass '--var connector.snowflake.dsn=...' to 'rill start'")
+	}
+
+	db, err := sql.Open("snowflake", dsn)
+	if err != nil {
+		return "", err
+	}
```
begelundmuller, line 67: Missing `defer db.Close()`?

runtime/drivers/snowflake/model_executor_self_objectstore.go
```diff
@@ -0,0 +1,123 @@
...
+	case "gcs":
+		return "", fmt.Errorf("snowflake connector can't export to connector 'gcs'. Use s3 compatibility.")
+	default:
+		return "", fmt.Errorf("snowflake connector can't export to connector %q", h.Driver())
+	}
+}
+
+type modelOutputProperties struct {
+	Path string `mapstructure:"path"`
+}
```
begelundmuller, line 112: Since this is not a snowflake model output, but an object store model output, can it either be moved to `s3.ModelOutputProperties` or `drivers.ObjectStoreModelOutputProperties`? 

Since it is the object store's `ModelManager` that will be handling the output properties later, the type should be managed in a place that that `ModelManager` is aware of.

runtime/drivers/snowflake/model_executor_self_objectstore.go
```diff
@@ -0,0 +1,123 @@
...
+		FROM (%s) 
+		CREDENTIALS = %s
+		HEADER = TRUE
+		MAX_FILE_SIZE = 536870912 
+		FILE_FORMAT = (TYPE='PARQUET' COMPRESSION = 'SNAPPY')`, outputLocation, conf.SQL, creds)
+	_, err = db.ExecContext(ctx, query)
+	if err != nil {
+		return "", err
+	}
+	return url.JoinPath(outputLocation, "*."+string(drivers.FileFormatParquet))
```
begelundmuller, line 91: So regarding the earlier comment about output properties, I think it would be nicer to return just `outputLocation` here to clearly indicate that the full directory is the model output.

runtime/reconcilers/model.go
```diff
@@ -276,10 +273,27 @@ func (r *ModelReconciler) Reconcile(ctx context.Context, n *runtimev1.ResourceNa
...
 	}
 
-	// Build the output
-	execRes, execErr := executor.Execute(ctx)
+	var (
+		executorConnector string
+		execRes           *drivers.ModelResult
+		execErr           error
+	)
+	// Open executor for the new output and build the output
```
begelundmuller, line 281: nit: prefer comments that precede a full code section

```suggestion
	// Open executor for the new output and build the output
	var (
		executorConnector string
		execRes           *drivers.ModelResult
		execErr           error
	)
```

begelundmuller (changes requested):

runtime/drivers/clickhouse/model_executor_s3_self.go
k-anshul: Yes it should from correctness POV though not used here.

runtime/drivers/snowflake/model_executor_self_objectstore.go
k-anshul: The idea was to make sure the `Format` is not fixed and is also returned. With the earlier implementation of `Export` it explicitly returned a `FileFormat`. 
The problem in returning just the `outputLocation` is that clickhouse's integration also requires a glob pattern or file name to work properly. 
Also the `path` parameter for s3 sourceproperties expects it to be a glob.

runtime/drivers/s3/model_manager.go
k-anshul: Hmm. I copied it from athena and planned to refactor all of these into `s3` pkg with the athena/redshift connector changes.

runtime/drivers/s3/model_manager.go
k-anshul: If it is a fixed path to a directory then it is essentially the same as what we are doing here. 
To generate a list of files we will need to do a `ListObjects` call in `snowflake->object_store` connector. 

To ingest files using clickhouse's s3 integration we need a glob_pattern. Since a user can submit a s3->clickhouse model directly we need to make sure that the `input_properties` cover all cases.

begelundmuller (approved):

PR #5296, opened 2024-07-17
Fixed a few typos in connectors.md

royendo:


docs/docs/reference/olap-engines/duckdb.md
```diff
@@ -25,4 +25,6 @@ Well now you can! For more details, see our page on [Using Multiple OLAP Engines
 
 ## Additional Notes
 
-- For dashboards powered by DuckDB, [measure definitions](/build/dashboards/dashboards.md#measures) are required to follow standard [DuckDB SQL](https://duckdb.org/docs/sql/introduction) syntax.
\ No newline at end of file
+- For dashboards powered by DuckDB, [measure definitions](/build/dashboards/dashboards.md#measures) are required to follow standard [DuckDB SQL](https://duckdb.org/docs/sql/introduction) syntax.
+- Known issue around creating a DuckDB via the UI, you will need to create the YAML file manually.
```
AndrewRTsao, line 29: Probably would clarify this is a DuckDB source

docs/docs/reference/olap-engines/duckdb.md
```diff
@@ -25,4 +25,6 @@ Well now you can! For more details, see our page on [Using Multiple OLAP Engines
...
+- Note that local running instances of DuckDB (outside of the default one) is mainly for development and local use. When uploading to Rill Cloud, it is recommended to use a `source` type instead
```
AndrewRTsao, line 30: Should the above two points go into this page or the "MotherDuck / DuckDB" connector page since this is more about DuckDB as a source vs an OLAP engine (perhaps we may need to further think about how we organize connectors vs OLAP engine docs)?

docs/docs/reference/project-files/connectors.md
```diff
@@ -28,7 +28,7 @@ Select + Add -> Data -> Connect an OLAP engine
 
 :::tip A note on OLAP engines
 
-You can have multiple OLAP engines in a single project and have each dashboard run on a different engine using the olap_connector parameter directly on a dashboard.
+You can have multiple OLAP engines in a single project and have each dashboard run on a different engine using the `connector` parameter directly on a dashboard.
```
AndrewRTsao, line 31: Rereading this again, I'm thinking we should also say "using the connector parameter directly within the dashboard YAML" with a link to the dashboard YAML reference page? WDYT?

AndrewRTsao (approved):
LGTM but added a few comments to get your input on before we did a final merge

docs/docs/reference/olap-engines/duckdb.md
royendo: This is for connecting to DuckDB as an OLAP engine... 
It's a bit confusing in general in the UI, is it a source or an engine?
Discussion on where this should be is already being discussed :)

docs/docs/reference/olap-engines/duckdb.md
royendo: Same as the previous comment, DuckDB should only be added as a source even though its an OLAP engine but then how do we make this less confusing for our users...

docs/docs/reference/project-files/connectors.md
royendo: I agree, i should've checked these before merging, will make the changes tomorrow

PR #5297, opened 2024-07-17
Skip DuckDB extension download if exists

### LINKED ISSUE #5277: Update DuckDB extensions embed script to skip download if possible
Since we run `make cli` quite often in development, it would be nice if the extension embed script only downloads extensions if they haven't already been downloaded.

esevastyanov:
Closes #5277


begelundmuller (approved):
Awesome, thanks!

PR #5299, opened 2024-07-17
rewrite comparison query to use CTE if possible

pjain1:
Resolves task 1 from https://github.com/rilldata/rill-private-issues/issues/493


runtime/drivers/registry.go
```diff
@@ -92,6 +92,8 @@ type InstanceConfig struct {
 	// AlertStreamingRefDefaultRefreshCron sets a default cron expression for refreshing alerts with streaming refs.
 	// Namely, this is used to check alerts against external tables (e.g. in Druid) where new data may be added at any time (i.e. is considered "streaming").
 	AlertsDefaultStreamingRefreshCron string `mapstructure:"rill.alerts.default_streaming_refresh_cron"`
+	// MetricsComparisonsCTE indicates whether to rewrite metrics comparison queries to use CTEs.
+	MetricsComparisonsCTE bool `mapstructure:"rill.metrics.comparisons_cte"`
```
begelundmuller, line 96: I would be good with just making it part of the `rill.metrics.approximate_comparisons` option for now – not sure if people will need the ability to opt out of this

runtime/metricsview/astsql.go
```diff
@@ -104,8 +117,21 @@ func (b *sqlBuilder) writeSelect(n *SelectNode) error {
...
+				b.out.WriteString(u)
+				b.out.WriteString(" ON TRUE")
+			} else {
+				b.out.WriteString(", ")
+				b.out.WriteString(u)
+			}
+		}
+
+		if n.JoinComparisonSelect != nil {
+			err := b.writeJoin(n.JoinComparisonType, n.CTEs[0], n.JoinComparisonSelect, nil)
```
begelundmuller, line 131: Feels unsafe to reference `n.CTEs[0]` like this – maybe in the future, another rewriter adds to `CTEs`.

runtime/metricsview/astsql.go
```diff
@@ -198,16 +231,26 @@ func (b *sqlBuilder) writeSelect(n *SelectNode) error {
 	return nil
 }
 
-func (b *sqlBuilder) writeJoin(joinType JoinType, baseSelect, joinSelect *SelectNode) error {
+func (b *sqlBuilder) writeJoin(joinType JoinType, baseSelect, joinSelect *SelectNode, joinTable *string) error {
```
begelundmuller, line 234: A little worried if this is becoming too hard to understand. I think the basic problem is that `SelectNode` is a fixed type, but that also keeps the code simple many other places, so changing it to an interface would have separate pains. 

What do you think about adding a `IsCTE bool` field to `SelectNode`? And then if it's set, then the node's alias is templated into the query instead of expanding the full `SELECT` subquery (everywhere except in the `WITH ...` part of course)?

runtime/metricsview/executor_rewrite_comparisons_cte.go
```diff
@@ -0,0 +1,94 @@
...
+func (e *Executor) rewriteComparisonsCTE(ast *AST) {
+	if !e.instanceCfg.MetricsComparisonsCTE {
+		return
+	}
+	if ast.Root.JoinComparisonSelect != nil {
+		_ = e.rewriteComparisonsCTENode(ast, ast.Root)
+	}
+}
+
+func (e *Executor) rewriteComparisonsCTENode(a *AST, n *SelectNode) bool {
```
begelundmuller, line 16: Since the logic is quite similar to the approx comparisons here, I wonder if we should just add this rewrite to that one instead of having a separate rewriter?

begelundmuller (changes requested):

runtime/metricsview/executor_rewrite_approx_comparisons.go
```diff
@@ -92,6 +97,23 @@ func (e *Executor) rewriteApproxComparisonNode(a *AST, n *SelectNode) bool {
...
+		// now change the JoinComparisonSelect WHERE clause to use selected dim values from CTE
+		for _, dim := range cte.DimFields {
+			var dimExpr string
+			if dim.Expr != "" {
+				dimExpr = dim.Expr
+			} else {
+				dimExpr = dim.Name
+			}
+			n.JoinComparisonSelect.Where = n.JoinComparisonSelect.Where.and(fmt.Sprintf("%[1]s IS NULL OR %[1]s IN (SELECT %[2]q.%[3]q FROM %[2]q)", dimExpr, cte.Alias, dim.Name), nil)
+		}
```
begelundmuller, line 116: Might be worth constructing an `*Expression` and using `sqlForExpression` to build the SQL here. There are two things to think about:
- Correct SQL escaping of the names
- Correct handling if the dim is unnested (not completely sure if it's needed here or not)

https://github.com/rilldata/rill/blob/4edcc574cb32a6201597069e1e51db96604b67ef/runtime/metricsview/astexpr.go#L10-L13

runtime/metricsview/astsql.go
```diff
@@ -102,42 +116,60 @@ func (b *sqlBuilder) writeSelect(n *SelectNode) error {
...
+				return fmt.Errorf("internal: FromSelect is a CTE but JoinComparisonSelect is nil, this is used only for comparison queries")
+			}
+			// find the CTE for FromSelect
+			var cte *SelectNode
+			for _, c := range n.CTEs {
+				if c.Alias == n.FromSelect.Alias {
+					cte = c
+					break
+				}
+			}
```
begelundmuller, line 137: Can't it just use `n.FromSelect` directly? Won't it still be the same pointer, just with `IsCTE == true`?

runtime/metricsview/executor_rewrite_approx_comparisons.go
pjain1: you mean create an expression for `"%[1]s IS NULL OR %[1]s IN (SELECT %[2]q.%[3]q FROM %[2]q)"` first in the same place then use `sqlForExpression` to get expr string instead of `fmt.Sprintf`. Not sure what difference will it make as at this point SelectNode contains ExprNode for where clause which is already deserialized form of the expression, also the dims are also deserialized into FieldNode

begelundmuller: Just for context, an expression filtering on an unnested dimension which isn't in `DimFields` requires some extra handling, such as this: https://github.com/rilldata/rill/blob/46bf4dec7b6a69d95a6fe24d1af2f00681ec4655/runtime/metricsview/astexpr.go#L238

However, in this case you're only targeting dimensions in `DimFields`, so you're right that that concern does not apply.

However, the concern of escaping still applies. You can fix it here by incorporating the right escape functions (`dialect.EscapeIdentifier` for the dim name and parentheses around the dim expressions).

pjain1: I understand what you mean but my only concern is `Expression` is created at an upper level in `NewAST` before optimizations so its a little odd to go back to that format here.

pjain1: I tried using expression but there its not fully supported yet for this use case
1. `IS` operator is not supported.
2. Writing `NULL` value is not supported.
3. `subquery` in the Expression does not support writing plain (non nested) select stmt referring any table or a CTE.

For now, I have changed to escape dim names and parenthesize dim expr, adding support for 3 will require some thought.

runtime/metricsview/astsql.go
```diff
@@ -102,42 +118,51 @@ func (b *sqlBuilder) writeSelect(n *SelectNode) error {
...
+				err := b.writeJoin(n.JoinComparisonType, n.FromSelect, n.JoinComparisonSelect)
+				if err != nil {
+					return err
+				}
 			}
-		}
-
-		if n.SpineSelect != nil {
-			err := b.writeJoin("RIGHT", n.FromSelect, n.SpineSelect)
+		} else {
```
begelundmuller, line 135: Shouldn't the `if` statement only cover the `FromSelect`? As it is right now, it duplicates the join select handling in both the `if` and `else` case, and it doesn't cover the left and spine joins in the `if` case.

runtime/metricsview/executor_rewrite_approx_comparisons.go
```diff
@@ -92,6 +99,19 @@ func (e *Executor) rewriteApproxComparisonNode(a *AST, n *SelectNode) bool {
 		n.FromSelect.OrderBy = order
 		n.FromSelect.Limit = a.Root.Limit
 		n.FromSelect.Offset = a.Root.Offset
+
+		// ---- CTE Optimization ---- //
+		// make FromSelect a CTE and set IsCTE flag on FromSelect
+		n.FromSelect.IsCTE = true
+		cte := n.FromSelect
+		a.CTEs = append(a.CTEs, cte)
```
begelundmuller, line 107: In theory, other rewriters might also add to the CTEs, and the same node could end up being added to `a.CTEs` multiple times. To protect agains that, maybe add a something like a `ast.convertToCTE(n *SelectNode)` util func that sets `IsCTE` and only adds to `a.CTEs` if `IsCTE` was false.

begelundmuller (changes requested):

begelundmuller (approved):

Issue #5231, opened 2024-07-09
Custom Dashboards: Variables and Filtering


ericpgreen2:
https://www.notion.so/rilldata/Passing-arguments-in-Custom-Dashboard-88a274c7c48c44dc8980c00e24d64ad7

mindspank:
Here is a POC branch: https://github.com/rilldata/rill/tree/feat/dynamicfilters

I would expect that we need new runtime endpoints. The POC takes a fairly naive approach to resolving a resource based on some input params.

The things that were outstanding in POC were wiring up the custom dash to dispatch only required params to components as well as setting up some store to manage it client side.

djbarnwal:
Variables closed via #5301
Filtering being tracked in #5448

PR #5301, opened 2024-07-17
feat: variable, input and output support for custom dashboards

AdityaHegde:
#5231

### Remaining Items
- [ ] Lint fixes
- [ ] Support special characters as arguments in Data query
- [ ] Support platform validation for Vega specs with args inside them
- [ ] Change type for `show` to string and link with variables store

Spec for select component

```
# Chart YAML
# Reference documentation: https://docs.rilldata.com/reference/project-files/charts
    
type: component

data:
  metrics_sql: |
    SELECT advertiser_name, measure_2
    FROM Bids_Sample_Dash
    LIMIT 250

output: 
  name: selection
  value: Leafly

select:
  placeholder: No selection
  valueField: advertiser_name
  label: Advertiser name
```

Sample Spec for Bar template with variable

```
kind: component

input:
 - name: selection
   type: string
   value: Instacart

data:
  metrics_sql: |
    SELECT
    time_grain_hour,
    measure_2
    FROM
        Bids_Sample_Dash
    WHERE
        __time >= '2023-03-06T00:30:00.000Z' 
        AND __time < '2023-03-08T00:30:00.000Z'
        AND advertiser_name = '{{ .args.selection }}'
    ORDER BY
        time_grain_hour DESC
    LIMIT 200
    OFFSET 0

bar_chart:
  x: time_grain_hour
  y: measure_2
```

Sample dashboard spec

```
type: dashboard
columns: 20
gap: 1

variables:
  - name: selection
    type: string
    value: Instacart

items:
  - component: horizontal_bar_chart
    height: 6
    width: 11
    x: 4
    y: 12
  - component: kpi_template
    height: 2
    width: 5
    x: 4
    y: 2
  - component: kpi_template
    height: 2
    width: 6
    x: 9
    y: 2
  - component: selection
    height: 1
    width: 5
    x: 5
    y: 1
  - component: stacked_area_chart
    height: 4
    width: 11
    x: 4
    y: 8
  - component: image
    height: 1
    width: 1
    x: 4
    y: 1
  - component: bar_template
    height: 4
    width: 11
    x: 4
    y: 4
```


runtime/server/components.go
```diff
@@ -10,14 +10,15 @@ import (
...
+	"github.com/rilldata/rill/runtime/compilers/rillv1"
 	"github.com/rilldata/rill/runtime/drivers"
 	"github.com/rilldata/rill/runtime/pkg/httputil"
 	"github.com/rilldata/rill/runtime/pkg/observability"
 	"github.com/rilldata/rill/runtime/server/auth"
 	"go.opentelemetry.io/otel/attribute"
 )
 
-func (s *Server) componentDataHandler(w http.ResponseWriter, req *http.Request) error {
+func (s *Server) parseTemplate(w http.ResponseWriter, req *http.Request) error {
```
begelundmuller, line 21: Since the response here is structured and pretty small, it would probably be better to define an RPC with protocol buffers for it

runtime/server/components.go
```diff
@@ -101,3 +145,37 @@ func (s *Server) getComponent(ctx context.Context, instanceID, name string) (*ru
 
 	return spec, nil
 }
+
+func (s *Server) getParsedComponent(ctx context.Context, instanceID, name string, args map[string]any) *string {
+	ctrl, err := s.runtime.Controller(ctx, instanceID)
+	if err != nil {
+		return nil
+	}
+	fmt.Println(name, args)
```
begelundmuller, line 154: Are these `fmt.Println`s leftover from debugging?

runtime/server/components.go
```diff
@@ -101,3 +145,37 @@ func (s *Server) getComponent(ctx context.Context, instanceID, name string) (*ru
...
+
+	res, err := ctrl.Get(ctx, &runtimev1.ResourceName{Kind: runtime.ResourceKindComponent, Name: name}, false)
+	if err != nil {
+		return nil
+	}
+
+	blob, _, err := s.runtime.GetFile(ctx, instanceID, res.Meta.GetFilePaths()[0])
+	if err != nil {
+		return nil
+	}
```
begelundmuller, line 164: It's not safe to rely on file loading from APIs that are used for dashboard serving because the files may be unavailable (for example, if the Github clone takes a long time or Github is unavailable). Viewer roles also are not allowed to have any way of loading files for security reasons.

So it should only use fields in `res` (which are available for querying as soon as the server is up).

begelundmuller (dismissed):

begelundmuller:
Here's a PR that adds an RPC that addresses the above issues with the `parseTemplate` and `componentDataHandler` APIs: https://github.com/rilldata/rill/pull/5362.

PR #5302, opened 2024-07-17
fix : external duckdb source detach fix

k-anshul:
partially closes https://github.com/rilldata/rill-private-issues/issues/479

### DIFF
--- a/runtime/drivers/duckdb/transporter_duckDB_to_duckDB.go
+++ b/runtime/drivers/duckdb/transporter_duckDB_to_duckDB.go
@@ -145,11 +145,13 @@ func (t *duckDBToDuckDB) transferFromExternalDB(ctx context.Context, srcProps *d
 
 		// make sure that dbName is not updated in any code below
 		cleanupFunc = func() {
+			// we don't want to run any detach db without `tx` lock
+			// tx=true will reopen duckdb handle which will detach the attached external db as well
 			err := t.to.WithConnection(context.Background(), 100, false, true, func(wrappedCtx, ensuredCtx context.Context, conn *sql.Conn) error {
-				return t.to.Exec(ensuredCtx, &drivers.Statement{Query: fmt.Sprintf("DETACH %s;", safeSQLName(dbName))})
+				return nil
 			})
 			if err != nil {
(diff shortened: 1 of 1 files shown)


begelundmuller (changes requested):
Looks like there's a test failure related to detaching

k-anshul:
The issue is that `tx=true` does not reopen db if it is in-memory db and the test use in-memory db.

k-anshul:
Moved the test to use db file. In real world scenarios in-memory db is unlikely and that too for a case when using external duckdb src. The other solution is to check for in-memory db explicitly.

We can't just do detach here and let it fail. Since the `detach` happens in another `WithConnection` block it is possible that by that time `detach` is executed another db with same name is attached. This is unlikely in case of external db storage since db names change.

runtime/drivers/duckdb/transporter_duckDB_to_duckDB_test.go
```diff
@@ -31,7 +31,7 @@ func TestDuckDBToDuckDBTransfer(t *testing.T) {
 	require.NoError(t, err)
 	require.NoError(t, conn.Close())
 
-	to, err := Driver{}.Open("default", map[string]any{"dsn": ":memory:"}, activity.NewNoopClient(), zap.NewNop())
+	to, err := Driver{}.Open("default", map[string]any{"path": filepath.Join(tempDir, "main.db")}, activity.NewNoopClient(), zap.NewNop())
```
begelundmuller, line 34: Should it be `tempDir, "tmp", "main.db"`? Worried this might cause `main.db` to be read in repo listings or trigger watches in the project parser, which could lead to some overhead or weird triggers

k-anshul: So we are not starting any runtime in this test. It directly opens a duckdb handle and executes transfer so it should be okay here ?

begelundmuller: Ah okay, I see, then it's fine!

begelundmuller (approved):

PR #5305, opened 2024-07-17
Runtime: Support listing files in an ObjectStore

begelundmuller:


runtime/resolvers/legacy_metrics.go
```diff
@@ -327,20 +331,33 @@ func (r *legacyMetricsResolver) formatValue(f formatter.Formatter, v any) any {
 type legacyResolverResult struct {
 	data   []byte
 	schema *runtimev1.StructType
+
+	parsed []map[string]any
```
k-anshul, line 335: Nit: `rows` to match with other resolver results ?

runtime/resolvers/glob.go
```diff
@@ -0,0 +1,376 @@
...
+		if entry.IsDir {
+			continue
+		}
+
+		dir := path.Dir(entry.Path)
+
+		row := rows[dir]
+		if row == nil {
+			// Init a new row
+			row = map[string]any{
```
k-anshul, line 234: May be add a struct for `row` and convert it to `map[string]any` while returning to avoid type assertions in below code ?

k-anshul (approved):

runtime/resolvers/glob.go
begelundmuller: I actually did that initially, but the problem is that `parseHivePartitions` can add arbitrary keys

runtime/resolvers/legacy_metrics.go
begelundmuller: Fixed

PR #5312, opened 2024-07-18
org billing email, restrict plan downgrades, card expiry check

pjain1:


admin/server/organizations.go
```diff
@@ -301,63 +314,66 @@ func (s *Server) UpdateBillingSubscription(ctx context.Context, req *adminv1.Upd
...
+		_, err = s.admin.Biller.CreateSubscription(ctx, org.BillingCustomerID, plan)
 		if err != nil {
 			return nil, status.Error(codes.Internal, err.Error())
 		}
 	}
 
+	org, err = s.admin.DB.UpdateOrganization(ctx, org.ID, &database.UpdateOrganizationOptions{
+		Name:                                org.Name,
+		Description:                         org.Description,
+		QuotaProjects:                       valOrDefault(plan.Quotas.NumProjects, org.QuotaProjects),
```
pjain1, line 364: I am thinking that plan quotas should always have a value, if we forgot to set the metadata key for a quota in Orb then we use the default values 
```
const (
	DefaultQuotaProjects                       = 5
	DefaultQuotaDeployments                    = 10
	DefaultQuotaSlotsTotal                     = 20
	DefaultQuotaSlotsPerDeployment             = 5
	DefaultQuotaOutstandingInvites             = 200
	DefaultQuotaSingleuserOrgs                 = 3
	DefaultQuotaStorageLimitBytesPerDeployment = int64(5368709120)
)
```

admin/server/organizations.go
```diff
@@ -1194,3 +1211,40 @@ func whitelistedDomainToPB(a *database.OrganizationWhitelistedDomainWithJoinedRo
...
+		return true
+	}
+	if comparableInt64(newPan.Quotas.StorageLimitBytesPerDeployment) < comparableInt64(&org.QuotaStorageLimitBytesPerDeployment) {
+		return true
+	}
+	return false
+}
+
+func comparableInt(v *int) int {
+	if v == nil || *v < 0 {
```
pjain1, line 1239: if the quotas is not set (nil) the we should use the default value instead of unlimited. can just change the plan creation logic to use default value instead of nil in case a quota is not set

begelundmuller: Not sure, but I think I'm fine with having unlimited when a quota is not explicitly set – it reduces user pain if we mess up, and should help in dev and testing. What do you think?

admin/billing/payment/stripe.go
```diff
@@ -54,19 +54,32 @@ func (s *Stripe) FindCustomer(ctx context.Context, customerID string) (*Customer
...
+	var isCardValid *bool
+	// very basic check if the customer has a payment method and if it's a card then is not expired
+	for i.Next() {
+		hasPaymentMethod = true
+		pm := i.PaymentMethod()
+		if pm.Type == stripe.PaymentMethodTypeCard {
+			isCardValid = new(bool)
+			*isCardValid = int(pm.Card.ExpYear) > time.Now().Year() || (int(pm.Card.ExpYear) == time.Now().Year() && int(pm.Card.ExpMonth) >= int(time.Now().Month()))
+		}
+	}
```
begelundmuller, line 67: Does Stripe have an API for this? There are other factors than just expiry, and Stripe might know about them (like probably they track cards they have detected to be blocked)

admin/billing/payment/stripe.go
```diff
@@ -80,11 +93,24 @@ func (s *Stripe) FindCustomerForOrg(ctx context.Context, organization *database.
...
+			var isCardValid *bool
+			// very basic check if the customer has a payment method and if it's a card then is not expired
+			for it.Next() {
+				hasPaymentMethod = true
+				pm := it.PaymentMethod()
+				isCardValid = new(bool)
+				if pm.Type == stripe.PaymentMethodTypeCard {
+					*isCardValid = int(pm.Card.ExpYear) > time.Now().Year() || (int(pm.Card.ExpYear) == time.Now().Year() && int(pm.Card.ExpMonth) >= int(time.Now().Month()))
+				}
+			}
```
begelundmuller, line 106: Since this is duplicated twice and non-trivial, can we pull it into a util func?

admin/server/organizations.go
```diff
@@ -301,63 +314,66 @@ func (s *Server) UpdateBillingSubscription(ctx context.Context, req *adminv1.Upd
...
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	if !c.HasPaymentMethod {
+		return nil, status.Errorf(codes.FailedPrecondition, "no valid payment method found for the organization")
+	}
+
+	// don't allow plan downgrades, only superuser can downgrade
+	if planDowngrade(plan, org) && !claims.Superuser(ctx) {
+		return nil, status.Errorf(codes.FailedPrecondition, "plan downgrade not allowed")
+	}
```
begelundmuller, line 336: Why do we not allow downgrades? What if someone doesn't use a bigger plan and wants to switch to a smaller one?

proto/rill/admin/v1/api.proto
```diff
@@ -1839,6 +1840,14 @@ message GetBillingSubscriptionResponse {
...
   bool has_payment_method = 4;
+  PaymentCardStatus payment_card_status = 5;
+
+}
+
+enum PaymentCardStatus {
+  PAYMENT_CARD_STATUS_UNSPECIFIED = 0;
+  PAYMENT_CARD_STATUS_OK = 1;
+  PAYMENT_CARD_STATUS_EXPIRED = 2;
 }
```
begelundmuller, line 1851: What is the goal of exposing this? Just wondering if we could just track whether they have a valid payment method or not, and direct them to the Stripe portal if they want more details?

As stated elsewhere, I'm also guessing there are many other conditions that could cause a card to be invalid than just expiry.

admin/billing/payment/stripe.go
pjain1: I could not find any such API but added specifically this check because Eric O mentioned that we want to show a banner if the card expires.

proto/rill/admin/v1/api.proto
pjain1: using the api we can just check if there exists a payment method or not for the customer, does explicitly say if its a valid or not.

admin/server/organizations.go
pjain1: @nishantmonu51 suggested we should not allow downgrades, for this they should contact us. Downgrades can be tricky if they are already over some quotas.

proto/rill/admin/v1/api.proto
pjain1: The goal is Eric O mentioned that we want to show a banner if the card expires.

begelundmuller: Okay, maybe not completely related, but do you think it would make sense to have generic `billing_warning` and `billing_error` fields on an org, and then show those in banners instead? Then we could populate `billing_warning` if it looks like the card will expire within the next 3 months or so, and set `billing_error` if we get a webhook from Stripe that tells us that a payment actually failed or a card has become invalid for some reason.

admin/server/organizations.go
begelundmuller: I wonder if it would be better to let people downgrade, and then in the beginning just have a job that checks/alerts if they don't get within the lower quotas within a day or so. Then we can just contact or hibernate non-compliant projects without causing trouble for people who want to downgrade for legitimate reasons.

proto/rill/admin/v1/api.proto
pjain1: sure that makes sense, however we will need some background jobs to clear out these warnings/errors once they are resolved.

pjain1: or we just rely on webhooks to clear these warnings like payment card updated for expired card warning or payment succeeded for payment failed error. Looks fragile though because webhook delivery may fail or our service may be unavailable at that point etc.

pjain1: ok I see webhooks are retried upto 3 days by stripe but we may need async queue to process the webhooks internally and prevent processing duplicate webhooks

pjain1: should webhook support be added in separate PR ?

begelundmuller: Yeah we can add it in a separate PR, but I think we should do it sooner rather than later.

Webhooks should be very stable, as you say they are usually retried until ack'ed, but if you worry about inconsistencies you can also have a background job once per day or so that checks and syncs statuses in case a webhook was dropped.

admin/server/organizations.go
pjain1: removed the restriction, just logging in case of downgrade

proto/rill/admin/v1/api.proto
pjain1: Removed checking of card expiry, just keeping check of presence of a payment method so that we can disallow plan change if no payment is present. This is all internal and not exposed in any API so that once we have billing warnings and errors, will just make sure no billing errors present instead of this payment method check.

admin/server/organizations.go
nishantmonu51: Yes, we intentionally wanted to keep the scope limited here and not allow downgrade, as downgrade would require checking if the quotas for the lower plan are met and only allow to downgrade. The expected flow here for a user would be to create a support ticket, FE team will manually verify the quotas and then downgrade directly in orb. 

Other reasons  for not allowing downgrade is for enterprise users who have signed a multi-year contract and we don't want them to downgrade unless explicitly approved. Since its not going to be a common case and in many such cases we want to have manual intervention. We do not allow downgrade at present.

admin/database/database.go
```diff
@@ -313,6 +315,7 @@ type UpdateOrganizationOptions struct {
 	QuotaStorageLimitBytesPerDeployment int64
 	BillingCustomerID                   string
 	PaymentCustomerID                   string
+	BillingEmail                        string
```
k-anshul, line 318: This is missing update in a few places.

k-anshul (changes requested):

admin/database/database.go
pjain1: I fixed them here - https://github.com/rilldata/rill/pull/5312/commits/de29f94cca5628533311fb0541f9fb9ed3fbb0ce

k-anshul: Sorry missed latest changes :)

k-anshul (approved):

PR #5316, opened 2024-07-19
trial end check worker

pjain1:
Checks every midnight for trial period end and logs it

### DIFF
--- a/admin/billing/noop.go
+++ b/admin/billing/noop.go
@@ -71,7 +71,7 @@ func (n noop) CancelSubscriptionsForCustomer(ctx context.Context, customerID str
 }
 
 func (n noop) FindSubscriptionsPastTrialPeriod(ctx context.Context) ([]*Subscription, error) {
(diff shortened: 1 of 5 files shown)


admin/billing/noop.go
```diff
@@ -70,6 +70,10 @@ func (n noop) CancelSubscriptionsForCustomer(ctx context.Context, customerID str
 	return nil
 }
 
+func (n noop) FindSubscriptionsPastTrialPeriod(ctx context.Context) ([]*Subscription, error) {
+	return []*Subscription{{Customer: &Customer{}, Plan: &Plan{Quotas: Quotas{}}}}, nil
+}
```
begelundmuller, line 75: Can it just return an empty list? Might be unexpected to get a value with nil fields

begelundmuller (approved):

begelundmuller (approved):
Looks good, but note the CI failure

Issue #5218, opened 2024-07-09
Runtime: Setup test fixtures for resolvers and migrate existing metrics tests


begelundmuller:
The idea here is to setup a test file abstractions for `runtime/resolvers`, where test queries for any resolver, security claims, project files, and backend can be written in a nice syntax.

The idea here is to draw inspiration from `sqllogictest` (also see [this example](https://github.com/duckdb/duckdb/blob/main/test/sql/attach/attach_export_import.test) from DuckDB's test files). For example, this could be YAML test files in `runtime/resolvers/testdata`, which are loaded, parsed and executed. 

It would also be nice to support a way to automatically update the expected test output, e.g. using an `--update` flag as described in this [blog post](https://ieftimov.com/posts/testing-in-go-golden-files/).

Lastly, we should start migrating our existing metrics tests to the new files.

begelundmuller:
The idea is that each file could define a set of project files and OLAP connectors to target, and then a list of tests to run and expected results. We might then have many of these files, such as `resolvers/testdata/metrics_unnest.yaml`, `resolvers/testdata/metrics_measure_filters.yaml`, etc.

Some of the goals here are:
- Having the test files be self-contained instead of requiring the project files to be defined separately in `testruntime` (which has the problem that making a tweak for one test causes a lot of unrelated tests other places to fail)
- Being able to optionally test the export output instead of the query output
- Being able to selectively test against multiple connectors (DuckDB and ClickHouse can be started with `testcontainers`, Druid will run only if a test cluster is defined)
- Being able to test queries for different security claims

I'm sure this can be made much cleaner, but here's a draft YAML spec for a test file.

```yaml
files:
  model.sql: SELECT ...
  metrics_view.yaml:
    type: metrics_view
    dimensions:
      - ...
    measures:
      - ...
connectors:
  - duckdb
  - clickhouse
tests:
  test_name:
    resolver: metrics
    properties:
      query:
        metrics_view: mv
        ...
    result:
      rows:
        - ...
      csv: >
        xxx,xxx,xxx
  other_test_name:
    variables:
      rill.metrics.approximate_comparisons: true
    claims:
      attributes:
      - ...
      rules:
      - ...
    resolver: metrics_sql
    properties:
      sql: SELECT ...
    result:
      rows:
```

egor-ryashin:
```
files:
  model.sql: SELECT ...
  metrics_view.yaml:
    type: metrics_view
    dimensions:
      - ...
    measures:
      - ...
[...]

PR #5317, opened 2024-07-20
Setup test fixtures for resolvers

egor-ryashin:


runtime/resolvers/resolvers_test.go
```diff
@@ -0,0 +1,124 @@
...
+					o.ResolverProperties = api.Spec.ResolverProperties.AsMap()
+
+					res, err := rt.Resolve(context.Background(), &o)
+					require.NoError(t, err)
+					var rows []map[string]interface{}
+					require.NoError(t, json.Unmarshal(res.Data, &rows), string(res.Data))
+					assert := td.Assert(t)
+					if !assert.Cmp(rows, test.Result) {
+						t.Log("Failure") // breakpoint can be placed here
+					}
```
begelundmuller, line 118: Can we use the `testify/require` package we use elsewhere? It also has deep comparison functions

runtime/security.go
```diff
@@ -23,7 +23,7 @@ var ErrForbidden = errors.New("action not allowed")
 // SecurityClaims represents contextual information for the enforcement of security rules.
 type SecurityClaims struct {
 	// UserAttributes about the current user (or service account). Usually exposed through templating as {{ .user }}.
-	UserAttributes map[string]any
+	UserAttributes map[string]any `yaml:"user_attributes"`
 	// AdditionalRules are optional security rules to apply *in addition* to the built-in rules and the rules defined on the requested resource.
 	// These are currently leveraged by the admin service to enforce restrictions for magic auth tokens.
 	AdditionalRules []*runtimev1.SecurityRule
```
begelundmuller, line 29: Can we remove the `yaml` parsing from this type (can be handled with a separate type in the tests)? It makes it look like this type can be serialized/deserialized to YAML, which it can't (because the proto type `runtimev1.SecurityRule` is not YAML compatible)

runtime/resolvers/resolvers_test.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v3"
+)
+
+type Project struct {
+	Sources    map[string]yaml.Node
+	Models     map[string]yaml.Node
+	Dashboards map[string]yaml.Node
+	APIs       map[string]yaml.Node
+}
```
begelundmuller, line 25: Instead of hard-coding the types, can we just use full paths as key types? We support quite a lot more resource types than just these four now. See this comment: 
https://github.com/rilldata/rill/issues/5218#issuecomment-2239087187

runtime/resolvers/resolvers_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+type Test struct {
+	Resolver string
+	Options  runtime.ResolveOptions
+	Result   []map[string]any
+}
+
+func TestResolvers(t *testing.T) {
+	yamlFile, err := os.ReadFile("resolvers_test.yaml")
```
begelundmuller, line 39: Should also support multiple test files (maybe list all YAML files in the dir?)

runtime/resolvers/resolvers_test.go
```diff
@@ -0,0 +1,124 @@
...
+	Project    Project
+	Connectors map[string]testruntime.InstanceOptions
+	Tests      map[string]*Test
+}
+
+type Test struct {
+	Resolver string
+	Options  runtime.ResolveOptions
+	Result   []map[string]any
+}
```
begelundmuller, line 36: Instead of turning internal types like `testruntime.InstanceOptions` and `runtime.ResolveOptions` into YAML serializable types, I would recommend explicitly adding the relevant fields here and then creating those internal types separately in the test initializer

runtime/resolvers/resolvers_test.yaml
```diff
@@ -0,0 +1,102 @@
...
+    result:
+      - domain: yahoo.com
+        publisher: YAHOO
+        "total volume": ~
+        "total impressions": 3.0
+  msn:
+    resolver: mv_sql_policy_api 
+    options:
+      claims:
+        user_attributes:
```
begelundmuller, line 87: For the sake of cleaner tests, I think we should flatten the nesting here and just have `user_attributes` as a root property of the test type

begelundmuller (changes requested):
Some other comments:

1. Put `.yaml` and `.csv` files into a separate `runtime/resolvers/testdata` directory
2. Add support for ClickHouse and Druid tests
3. Do you think it makes sense to have some kind of `--update` flag that will write out resolver results to the test files? To make it easier to author new tests. E.g. see this for an example: https://ieftimov.com/posts/testing-in-go-golden-files/

nishantmonu51:
@egor-ryashin : Any updates on this PR ?

runtime/resolvers/resolvers_test.go
egor-ryashin: I wonder why replacing `assert` - it allows deep comparison of structures.

runtime/resolvers/resolvers_test.yaml
egor-ryashin: It relates to the previous argument If you like to have more test structures then it's less production code coverage.

egor-ryashin:
@nishantmonu51  Most of comments I'll resolve in 1d, the first one "updates" can take longer.

runtime/security.go
egor-ryashin: I wonder why we cannot do it YAML compatible, more copies of structs - more structs to maintain - harder to maintain when 2 structs are almost equal (finding the distinction is harder) - more error-prone.

runtime/resolvers/resolvers_test.go
egor-ryashin: I personally don't see much difference:
```
sources:
  1.yaml
  2.yaml
dashboards:
  1.yaml
  2.yaml
```

```
- sources/1.yaml:
- sources/2.yaml:
- dashboards/1.yaml:
- dashboards/2.yaml:
```

We can easily add a new field to Project struct if more resources is required. 
That could make sense if you'd like to process all files equally with a single for-loop and it can save a few lines of code but you'll have to repeat yourself if you need multiple similar-type entries in the future, ie:
```
- sources/1.yaml:
- sources/2.yaml:
- dashboards/1.yaml:
- dashboards/2.yaml:
```

runtime/resolvers/resolvers_test.go
egor-ryashin: Done.

egor-ryashin:
>1. Put .yaml and .csv files into a separate runtime/resolvers/testdata directory
>1. Add support for ClickHouse and Druid tests
>1. Do you think it makes sense to have some kind of --update flag that will write out resolver results to the test files? To make it easier to author new tests. E.g. see this for an example: https://ieftimov.com/posts/testing-in-go-golden-files/

1. Done (replaced with the public `https` source - no need to commit data).
2. Done.

runtime/resolvers/resolvers_test.go
egor-ryashin: Done.

egor-ryashin:
Merge conflicts resolved.

runtime/resolvers/resolvers_test.yaml
egor-ryashin: I understand that we would like to see a more concise schema but it's already have a diverse structure, simplifying it is short term goal - eventually we will need to test more production fields anyway.

egor-ryashin:
Updates having this issue https://github.com/go-yaml/yaml/issues/430
Floats are overwritten with integers breaking unit tests consequently, trying to workaround.

egor-ryashin:
Updates done.

runtime/resolvers/clickhouse_resolvers_test.yaml
```diff
@@ -0,0 +1,91 @@
+project:
```
begelundmuller, line 1: Please move the `.yaml` test fixtures into a separate directory at `runtime/resolvers/testdata` since the Go docs recommend putting non-Go test files in a separate `testdata` directory

runtime/resolvers/resolvers_test.go
```diff
@@ -0,0 +1,158 @@
...
+var update = flag.Bool("update", false, "Update test results")
+
+func TestMain(m *testing.M) {
+	flag.Parse()
+	os.Exit(m.Run())
+}
+
+func TestResolvers(t *testing.T) {
+
+	entries, err := os.ReadDir("./")
```
begelundmuller, line 48: nit: redundant newline

runtime/resolvers/resolvers_test.go
```diff
@@ -0,0 +1,158 @@
...
+	require.NoError(t, err)
+	var reg = regexp.MustCompile(`^(.*)_resolvers_test.yaml$`)
+	for _, e := range entries {
+		if reg.MatchString(e.Name()) {
+			t.Log("Running with", e.Name())
```
begelundmuller, line 53: I believe `t.Run` can be nested any number of times, so maybe use `t.Run` with the filename, and then do a nested call to `t.Run` for the test names. 

Using `t.Run` automatically prints test run status and success/failure info

runtime/resolvers/resolvers_test.go
```diff
@@ -0,0 +1,158 @@
...
+					opts.OLAPDriver = "clickhouse"
+				}
+
+				maps.Copy(opts.Files, files)
+				rt, instanceID := testruntime.NewInstanceForResolvers(t, *opts)
+				for testName, test := range r.Tests {
+					t.Run(testName, func(t *testing.T) {
+						t.Log("======================")
+						t.Log("Running ", testName, "with", e.Name(), "and", connector)
+						testruntime.RequireParseErrors(t, rt, instanceID, nil)
```
begelundmuller, line 109: We should also check that there aren't any reconcile errors. Consider using `RequireReconcileState` and patching it to skip checks when the desired length is `-1`. Then you could call `RequireReconcileState(t, rt, instanceID, -1, 0, 0)`.

runtime/resolvers/resolvers_test.go
```diff
@@ -0,0 +1,158 @@
...
+						api, err := rt.APIForName(context.Background(), instanceID, test.Resolver)
+						require.NoError(t, err)
+
+						o := test.Options
+						o.InstanceID = instanceID
+						o.Resolver = api.Spec.Resolver
+						o.ResolverProperties = api.Spec.ResolverProperties.AsMap()
+
+						res, err := rt.Resolve(context.Background(), &o)
```
begelundmuller, line 118: It shouldn't invoke the resolver through an API (not all resolvers can be accessed through APIs). It should just call it directly, for example like in the alerts checks: https://github.com/rilldata/rill/blob/d66a28ff94b55273d102cc2112faf8d05465d254/runtime/reconcilers/alert.go#L561

runtime/resolvers/resolvers_test.go
```diff
@@ -0,0 +1,158 @@
...
+						require.NoError(t, err)
```
begelundmuller, line 119: It should be possible to specify a desired error in the test fixture file. Something like `error_matches` or `error_contains`.

runtime/resolvers/resolvers_test.go
```diff
@@ -0,0 +1,158 @@
...
+									switch val := v.(type) {
+									case float32:
+										node.Value = strconv.FormatFloat(float64(val), 'f', 2, 32)
+										m[k] = &node
+									case float64:
+										node.Value = strconv.FormatFloat(val, 'f', 2, 64)
+										m[k] = &node
+									}
+								}
+							}
```
begelundmuller, line 139: Instead of trying to force YAML to deserialize a value into a specific type (it is a dynamically typed format after all), I think it would be simpler to just serialize the desired result from the YAML to JSON and comparing that to the resolver result's JSON (Go's serialized JSON output is mostly deterministic IIRC).

runtime/resolvers/duckdb_resolvers_test.yaml
```diff
@@ -0,0 +1,99 @@
...
+    resolver: mv_sql_policy_api 
+    options:
+      claims:
+        user_attributes:
+          domain: yahoo.com 
+          email: user@yahoo.com
+    result:
+      - domain: yahoo.com
+        publisher: YAHOO
+        "total impressions": 3.0
```
begelundmuller, line 80: If there are many output values, this might get tedious to type – do you think it would make sense to optionally support a `result_csv:` key to represent multi-row output more concisely?

runtime/testruntime/testruntime.go
```diff
@@ -84,6 +86,13 @@ type InstanceOptions struct {
 	StageChanges bool
 }
 
+type InstanceOptionsForResolvers struct {
+	InstanceOptions
+	OLAPDriver string
+	OLAPDSN    string
+	TempDir    string
```
begelundmuller, line 93: This appears unused

runtime/security.go
begelundmuller: As written above, the `*runtimev1.SecurityRule` proto-generated type is not safe for use with YAML, so we don't want to give the impression that the `SecurityClaims` type is YAML-serializable. Additionally, it's just a bad practice to hack an internal type that used in many places for a single specific use case (in this case, test fixtures) – it creates unnecessary coupling.

In this case, `UserAttributes` is the only field we need to be able to populate from the tests, so it is simple to parse it from the test fixture and create a `SecurityClaims` value for it.

runtime/resolvers/resolvers_test.go
begelundmuller: IMO reducing nesting one level for all the test files definitely improves readability here. The very slight repetition of the directory name does not hurt readabilty IMO, and anyway we don't require specific directory names anymore, so the files can just be created at the root level.

This also has the additional benefit of simplifying the test fixture parsing code.

runtime/resolvers/resolvers_test.yaml
begelundmuller: This doesn't impact production code coverage at all – it just makes the test fixture simpler to read/write. Internally the test fixture parser would create a `SecurityClaims` value, so the production code coverage is exactly the same.

runtime/resolvers/resolvers_test.go
```diff
@@ -0,0 +1,158 @@
...
+	Project    Project
+	Connectors map[string]*testruntime.InstanceOptionsForResolvers
+	Tests      map[string]*Test
+}
+
+type Test struct {
+	Resolver string
+	Options  runtime.ResolveOptions
+	Result   []map[string]any
+}
```
begelundmuller, line 37: I strongly believe it would simplify the code and readability here to combine this into a single struct with anonymous sub-structs to represent the YAML structure – and avoid using internal types.

This is also what we do for the resource YAML parsing, and it serves us well. Look at the `AlertYAML` type here as an example: https://github.com/rilldata/rill/blob/d66a28ff94b55273d102cc2112faf8d05465d254/runtime/compilers/rillv1/parse_alert.go#L18

runtime/resolvers/resolvers_test.go
egor-ryashin: It won't run in vscode https://github.com/golang/vscode-go/issues/3022

go.mod
```diff
@@ -261,6 +261,7 @@ require (
 	github.com/mattn/go-isatty v0.0.20 // indirect
 	github.com/mattn/go-runewidth v0.0.15 // indirect
 	github.com/mattn/go-sqlite3 v2.0.3+incompatible // indirect
+	github.com/maxatome/go-testdeep v1.14.0 // indirect
```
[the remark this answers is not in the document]
egor-ryashin: Fixed.

runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go
```diff
@@ -65,12 +66,18 @@ func (c *sqlConnection) QueryContext(ctx context.Context, query string, args []d
 	})
 	req, err := http.NewRequestWithContext(ctx, http.MethodPost, c.dsn, bodyReader)
 	if err != nil {
+		if strings.Contains(err.Error(), c.dsn) { // avoid returning the actual DSN with the password which will be logged
+			return nil, fmt.Errorf("invalid dsn")
+		}
```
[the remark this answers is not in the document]
egor-ryashin: Done.

runtime/resolvers/clickhouse_resolvers_test.yaml
egor-ryashin: Done.

runtime/resolvers/resolvers_test.go
```diff
@@ -0,0 +1,158 @@
...
+	os.Exit(m.Run())
+}
+
+func TestResolvers(t *testing.T) {
+
+	entries, err := os.ReadDir("./")
+	require.NoError(t, err)
+	var reg = regexp.MustCompile(`^(.*)_resolvers_test.yaml$`)
+	for _, e := range entries {
+		if reg.MatchString(e.Name()) {
```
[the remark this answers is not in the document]
egor-ryashin: Done.

runtime/resolvers/resolvers_test.go
egor-ryashin: Done.

runtime/resolvers/resolvers_test.go
egor-ryashin: Done.

runtime/resolvers/resolvers_test.go
egor-ryashin: OK, but I'll need to copy all the data to ResolveOptions anyway here to call a method:
```
					res, err := rt.Resolve(context.Background(), &ropts)
```

egor-ryashin: `SecurityClaims` have a `oneof` field, now it makes sense.

runtime/resolvers/duckdb_resolvers_test.yaml
egor-ryashin: You cannot specify the type `number` or `string` with CSV.

egor-ryashin: Done.

runtime/testruntime/testruntime.go
egor-ryashin: Removed

runtime/resolvers/resolvers_test.go
egor-ryashin: "serialize the desired result from the YAML to JSON" puzzled here

runtime/resolvers/resolvers_test.go
egor-ryashin: Done.

runtime/resolvers/resolvers_test.go
egor-ryashin: Done.

runtime/resolvers/resolvers_test.go
egor-ryashin: Done.

runtime/resolvers/resolvers_test.yaml
egor-ryashin: Done.

runtime/resolvers/resolvers_test.go
egor-ryashin: Done.

egor-ryashin:
All done.

begelundmuller:
Can you take a look at the CI failures?

egor-ryashin:
There's a weird error currently 
```
        	            	rill.runtime.v1.Model/ad_bids_mini: failed to create model: code: 27, message: Cannot parse input: expected ',' before: 'Z,,msn.com,2,1,2,cars,,iphone\n2,2022-01-03T14:49:50.459Z,,msn.com,2.5,10,1,cars,,iphone\n1,2022-01-02T11:58:12.475Z,Yahoo,yahoo.com,2,100,1,cars,1,\n3,2022-01-04T': (at row 1)
```
It's not reproduced locally with the same docker image.

egor-ryashin:
There's a new exception emerged after the merge: 
```
ORDER BY or PRIMARY KEY clause is missing. Consider using extended storage definition syntax with ORDER BY or PRIMARY KEY clause. With deprecated old syntax
```
I'm figuring that out.

egor-ryashin:
Solved here https://github.com/rilldata/rill/pull/5591

begelundmuller (approved):

PR #5340, opened 2024-07-25
feat: human readable dimension filters in URL

briangregoryholmes:
https://www.notion.so/rilldata/Passing-Filters-as-part-of-URL-ac2972fcd3f64de3af20838ec1c0231b?pvs=4

The API is largely taken from https://learn.microsoft.com/en-us/power-bi/collaborate-share/service-url-filters , but see the notion doc above for greater detail.

Single value, no parenthesis:
`"dimension_0 eq 'Zoom'"`
`dimension_0 eq 'Zoom'`

With parentheses: 
`"dimension_0 eq ('Zoom')"`
`dimension_0 eq ('Zoom')`
`"dimension_0 eq ('Zoom', 'Zapier')"`
`dimension_0 eq ('Zoom', 'Zapier')`

Exclude:
`dimension_0 ne ('Zoom')`
`dimension_0 ne 'Zoom'`

Multiple filters:
`dimension_0 eq ('Zoom', 'Zapier') and dimension_1 eq 'Standard Plan'`

Angled versions of single and double quotes (if copied from a text editor that automatically changes the character) should also work:
`”dimension_0 eq (‘Zoom’, ‘Zapier’)”`


nishantmonu51:
@briangregoryholmes : Can you resolve the conflicts here ? Is this ready to review ?

ericpgreen2:
It looks like you need to add support for this feature to Rill Cloud. On opening a Cloud dashboard, I hit this error:
image

And while we currently don't need support for the other instances of the `<Dashboard />` component (`MetricsExplorerEmbed.svelte`, `[org]/[project]/-/share/[token]/+page.svelte`), I'm guessing they'll need to be protected from the above error.

ericpgreen2:
On an Adbids dashboard, I've tried navigating to this URL: `http://localhost:3001/dashboard/Adbids_dashboard?filter="publisher eq Facebook"`

The "F" is getting clipped from "Facebook" to "acebook":

image


Can you please update the PR description with the API that you've implemented and some examples?

briangregoryholmes:
@ericpgreen2 According to the Notion, dimension values are required to be surrounded by some form of single quotes. I updated the description with a variety of examples.

Unfortunately, I don't think it's practical to validate whether a dimension value is valid, though that's certainly something we can discuss. I've added some guard rails and an updated warning if we're able to detect that something should be quoted that isn't.

Added cloud support.

ericpgreen2:
> @ericpgreen2 According to the Notion, dimension values are required to be surrounded by some form of single quotes. I updated the description with a variety of examples.

Oh, got it now, thanks. 

[I commented in the Notion doc](https://www.notion.so/Passing-Filters-as-part-of-URL-ac2972fcd3f64de3af20838ec1c0231b?d=b6702fad2d34442cb1cfa1642ddb1211&pvs=4#656d6084828246ee81c55862f4250516), but IMO the ephemeral nature of these filters makes for a frustrating UX. Often, I find myself getting some part of the filter syntax incorrect. I would like to iterate on the filter / edit it after submission. But when the filter is immediately transformed into its protobuf representation, I cannot edit what I input. Let's wait for feedback from @mindspank and @nishantmonu51.

PR #5342, opened 2024-07-28
Fix: Protect against table names with spaces in generated models

ericpgreen2:
Closes https://github.com/rilldata/rill-private-issues/issues/404

Additionally, this PR removes the `createModelFromSource` function in favor of `createModelFromTable`.


web-common/src/features/connectors/olap/createModel.ts
```diff
@@ -75,16 +77,22 @@ export async function createModelFromTable(
   const topComments =
     "-- Model SQL\n-- Reference documentation: https://docs.rilldata.com/reference/project-files/models";
   const connectorLine = `-- @connector: ${connector}`;
-  const selectStatement = `select * from ${sufficientlyQualifiedTableName}`;
+  const selectStatement = hasSpaces(sufficientlyQualifiedTableName)
```
AdityaHegde, line 80: Why not always surround with quotes? Would be convenient when the name is changed later to include spaces.

AdityaHegde (approved):

web-common/src/features/connectors/olap/createModel.ts
ericpgreen2: Fair, it's subjective but I think it's a little cleaner to omit the quotes if possible

PR #5346, opened 2024-07-30
Runtime : Parse metrics_sql to `metricsview.query`

k-anshul:


go.mod
```diff
@@ -237,6 +237,7 @@ require (
 	github.com/imdario/mergo v0.3.16 // indirect
 	github.com/inconshreveable/mousetrap v1.1.0 // indirect
 	github.com/invopop/yaml v0.3.1 // indirect
+	github.com/itlightning/dateparse v0.2.0 // indirect
```
begelundmuller, line 240: Seems like it changed `go.mod` without any change to the direct dependencies? Might be worth running `go mod tidy`

runtime/metricsview/astexpr.go
```diff
@@ -121,6 +121,8 @@ func (b *sqlExprBuilder) writeCondition(cond *Condition) error {
 		return b.writeJoinedExpressions(cond.Expressions, " OR ")
 	case OperatorAnd:
 		return b.writeJoinedExpressions(cond.Expressions, " AND ")
+	case OperatorEqNull, OperatorNeqNull:
+		return b.writeNullCondition(cond.Expressions, cond.Operator == OperatorEqNull)
```
begelundmuller, line 125: Parag actually just added handling of this here: 
https://github.com/rilldata/rill/blob/a370792226082e832932733ce352c7b2941578ef/runtime/metricsview/astexpr.go#L292-L296

It's in a slightly different way – just checking if `Eq` is used with a `nil` on one side. That avoids a new operator.

But looks like he didn't handle the `NOT` case, so you might want to add that :)

runtime/resolvers/metrics_sql.go
```diff
@@ -41,8 +39,8 @@ func newMetricsSQL(ctx context.Context, opts *runtime.ResolverOptions) (runtime.
 		return nil, err
 	}
 
-	var finalRefs []*runtimev1.ResourceName
-	props.SQL, finalRefs, err = resolveTemplate(props.SQL, opts.Args, instance, opts.Claims.UserAttributes, opts.ForExport)
+	// todo handle refs
+	props.SQL, _, err = resolveTemplate(props.SQL, opts.Args, instance, opts.Claims.UserAttributes, opts.ForExport)
```
begelundmuller, line 43: I think we need to handle that to ensure correct cache invalidation when refs are updated?

begelundmuller (approved):

runtime/resolvers/metrics_sql.go
k-anshul: So WRT `metrics_sql` the only ref as of now is `metrics_view` which is already handled in `metricsResolver` so it should be okay to ignore this for now ?

begelundmuller: Oh yeah true, should be fine then

PR #5347, opened 2024-07-30
ClickHouse as a subprocess

### LINKED ISSUE #5207: ClickHouse: Support starting ClickHouse as a sub-process
- When no other DSN is configured
- Should only do it on local

esevastyanov:
Closes #5207


cli/cmd/start/start.go
```diff
@@ -143,6 +143,14 @@ func StartCmd(ch *cmdutil.Helper) *cobra.Command {
 			}
 			localURL := fmt.Sprintf("%s://localhost:%d", scheme, httpPort)
 
+			// if olapDriver is clickhouse and no olapDSN is specified, run ClickHouse as a subprocess
+			if olapDriver == "clickhouse" && olapDSN == local.DefaultOLAPDSN {
+				olapDSN, err = startClickHouse(cmd.Context(), projectPath)
+				if err != nil {
+					return err
+				}
+			}
```
begelundmuller, line 152: Unfortunately this is not a reliable way of discovering if ClickHouse is the OLAP engine because these flags are deprecated: https://github.com/rilldata/rill/blob/b67474d7b4fd3770f21999f0c3e96a3e86c216b7/cli/cmd/start/start.go#L196-L199

We now support having one project connect to multiple OLAP engines simultaneously by configuring `connector: name` in the individual sources and models. To avoid repetition, a default engine can be configured in `rill.yaml` using `olap_connector: ...`.

I think the only way to handle this is to move the handling of starting the ClickHouse subprocess into `runtime/drivers/clickhouse` in the `driver.Open` function. This has some nice other advantages:
- Provides easy access to the `DataDir` and `TempDir` (they are automatically made available as connector properties [here](https://github.com/rilldata/rill/blob/b67474d7b4fd3770f21999f0c3e96a3e86c216b7/runtime/connections.go#L281-L285)).
- Allows overriding parameters like the port using variables, maybe something like `--var connector.clickhouse.embed_port`.
- Centralizes ClickHouse logic in the ClickHouse driver package.
- The `Close` function on the driver handle is automatically called when interrupt signals are received, so the subprocess doesn't need to listen for OS signals directly. It can just respond to `Close` calls.

begelundmuller (changes requested):

cli/cmd/start/clickhouse.go
```diff
@@ -0,0 +1,259 @@
...
+		if err != nil {
+			return "", fmt.Errorf("error creating ClickHouse directory: %w", err)
+		}
+	}
+
+	if _, err := os.Stat(destPath); err == nil {
+		// ClickHouse binary already exists
+		// TODO: Check compatibility in case the binary is outdated.
+		return destPath, nil
+	}
```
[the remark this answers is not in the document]
esevastyanov: I haven't found a way to determine the version of the Clickhouse binary without running and querying it.

cli/cmd/start/clickhouse.go
```diff
@@ -0,0 +1,259 @@
...
+// ClickHouse requires a config file with a minimal set of properties to be passed:
+// https://github.com/ClickHouse/ClickHouse/blob/master/programs/server/embedded.xml
+// Full list of properties can be found here:
+// https://github.com/ClickHouse/ClickHouse/blob/master/programs/server/config.xml
+func clickHouseConfigContent(projectPath string) []byte {
+	config := []byte(fmt.Sprintf(`<clickhouse>
+    <logger>
+        <level>trace</level>
+        <console>true</console>
+    </logger>
```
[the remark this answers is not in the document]
esevastyanov: at least `information` level is required to determine the start status

cli/cmd/start/clickhouse.go
```diff
@@ -0,0 +1,259 @@
...
+</clickhouse>`, projectPath, projectPath, projectPath))
+	return config
+}
+
+// newCmd initializes an exec.Cmd that sends SIGINT instead of SIGKILL when the ctx is canceled.
+func newCmd(ctx context.Context, name string, args ...string) *exec.Cmd {
+	cmd := exec.CommandContext(ctx, name, args...)
+	cmd.Cancel = func() error {
+		return cmd.Process.Signal(os.Interrupt)
+	}
```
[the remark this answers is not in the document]
esevastyanov: `ctx` isn't required as `driver.Open` doesn't have it

cli/cmd/start/clickhouse.go
```diff
@@ -0,0 +1,259 @@
...
+		if err == nil {
+			conn.Close()
+			return nil
+		}
+		if time.Since(start) > timeout {
+			return fmt.Errorf("timed out waiting for ClickHouse to be ready")
+		}
+		time.Sleep(1 * time.Second)
+	}
+}
```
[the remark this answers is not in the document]
esevastyanov: this check might be removed as log monitoring is enough

cli/cmd/start/clickhouse.go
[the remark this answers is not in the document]
begelundmuller: What if we pin to a specific version when downloading the binary? Like we pin to a specific DuckDB version. Then we would know it, right?

cli/cmd/start/clickhouse.go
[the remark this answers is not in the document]
begelundmuller: I meant are these shown in the user's CLI logs? We may need them internally to determine if it started successfully, but I just don't want to clutter the Rill logs.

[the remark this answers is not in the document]
esevastyanov: I left only clickhouse error logs to be proxied into the zap error logs until clickhouse is up. Once clickhouse is up, there won't be any logs proxied (shown in the user's cli logs)

begelundmuller (changes requested):
Caught up with Eugene and he mentioned the following todos are still outstanding:
- [ ] Download the proper binary per OS and architecture
- [x] Pin to a specific ClickHouse version and cache it by version name in `~/.rill`
- [ ] Debug the timeout errors that occasionally occur

runtime/drivers/clickhouse/clickhouse.go
```diff
@@ -112,6 +117,7 @@ func (d driver) Open(instanceID string, config map[string]any, client *activity.
 		return nil, err
 	}
 
+	var embed *embedClickHouse
 	// build clickhouse options
 	var opts *clickhouse.Options
```
begelundmuller, line 122: nit: Prefer section-level comments

```suggestion
	// build clickhouse options
	var opts *clickhouse.Options
	var embed *embedClickHouse
```

runtime/drivers/clickhouse/clickhouse.go
```diff
@@ -238,7 +252,15 @@ func (c *connection) Config() map[string]any {
...
-	return c.db.Close()
+	return errors.Join(
+		c.db.Close(),
+		func() error {
+			if c.embed != nil {
+				return c.embed.stop()
+			}
+			return nil
+		}(),
+	)
```
begelundmuller, line 263: Since these calls won't be parallelized, and it's best to close the DB connection first, how about flattening the code like this?

```suggestion
	errDB := c.db.Close()

	var errEmbed error
	if c.embed != nil {
		errEmbed = c.embed.stop()
	}

	return errors.Join(errDB, errEmbed)
```

runtime/drivers/clickhouse/embed.go
```diff
@@ -0,0 +1,377 @@
...
+		err := e.startAndWaitUntilReady()
+		ready <- err
+		if err != nil && e.cmd != nil && e.cmd.Process != nil {
+			_ = e.cmd.Process.Kill()
+			return
+		}
+
+		if err := e.cmd.Wait(); err != nil {
+			e.logger.Error("clickhouse server exited with an error", zap.Error(err))
+		}
```
begelundmuller, line 77: What about when there's a graceful shutdown (i.e. `stop` is called) – will ClickHouse shut down with status 0 or will there be some kind of error? If there's an error, we should probably skip it here.

runtime/drivers/clickhouse/embed.go
```diff
@@ -0,0 +1,377 @@
...
+	}()
+
+	if err := <-ready; err != nil {
+		return nil, err
+	}
+
+	addr := net.JoinHostPort("localhost", fmt.Sprintf("%d", e.tcpPort))
+	e.logger.Info("ClickHouse server: " + "clickhouse://" + addr)
```
begelundmuller, line 85:
```suggestion
	e.logger.Info("Running an embedded ClickHouse server", zap.String("addr", addr))
```

runtime/drivers/clickhouse/embed.go
```diff
@@ -0,0 +1,377 @@
...
+
+	return &clickhouse.Options{
+		Protocol: clickhouse.Native,
+		Addr:     []string{addr},
+	}, nil
+}
+
+func (e *embedClickHouse) stop() error {
+	e.logger.Info("Stopping ClickHouse server: " + fmt.Sprintf("localhost:%d", e.tcpPort))
```
begelundmuller, line 94: Should use structured logging and same format as in `start`:

```suggestion
	addr := net.JoinHostPort("localhost", fmt.Sprintf("%d", e.tcpPort))
	e.logger.Info("Stopping embedded ClickHouse server", zap.String("addr", addr))
```

runtime/drivers/clickhouse/embed.go
```diff
@@ -0,0 +1,377 @@
...
+	case "darwin":
+		fileName := ""
+		switch goarch {
+		case "amd64":
+			fileName = "clickhouse-macos"
+		case "arm64":
+			fileName = "clickhouse-macos-aarch64"
+		}
+		url := "https://github.com/ClickHouse/ClickHouse/releases/download/" + release + "/" + fileName
+		logger.Info(fmt.Sprintf("Will download %s into %s\n", url, destPath))
```
begelundmuller, line 139:
```suggestion
		logger.Info("Downloading ClickHouse binary", zap.String("url", url), zap.String("dst", destPath))
```

runtime/drivers/clickhouse/embed.go
```diff
@@ -0,0 +1,377 @@
...
+		fileName := ""
+		switch goarch {
+		case "amd64":
+			fileName = "clickhouse-common-static-24.7.4.51-amd64.tgz"
+		case "arm64":
+			fileName = "clickhouse-common-static-24.7.4.51-arm64.tgz"
+		}
+		url := "https://github.com/ClickHouse/ClickHouse/releases/download/" + release + "/" + fileName
+		destTgzPath := filepath.Join(destDir, release, fileName)
+		logger.Info(fmt.Sprintf("Will download %s into %s\n", url, destTgzPath))
```
begelundmuller, line 153:
```suggestion
		logger.Info("Downloading ClickHouse binary", zap.String("url", url), zap.String("dst", destTgzPath))
```

runtime/drivers/clickhouse/embed.go
```diff
@@ -0,0 +1,377 @@
...
+}
+
+func newEmbedClickHouse(tcpPort int, dataDir, tempDir string, logger *zap.Logger) *embedClickHouse {
+	return &embedClickHouse{tcpPort: tcpPort, dataDir: dataDir, tempDir: tempDir, logger: logger}
+}
+
+// start installs (depending on OS and platform) and starts ClickHouse server.
+// The destination directory for the ClickHouse binary is .rill/clickhouse.
+// The function returns the DSN for the ClickHouse server and close function.
+func (e *embedClickHouse) start() (*clickhouse.Options, error) {
```
begelundmuller, line 40: Adding a todo here since this is important to handle in a follow up (will assign someone after this merges):

```suggestion
// The function returns the DSN for the ClickHouse server and close function.
// 
// TODO: Since this can be a long-running process, we should accept a `ctx`,
// but the `drivers.Open` function currently doesn't propagate that.
func (e *embedClickHouse) start() (*clickhouse.Options, error) {
```

runtime/drivers/clickhouse/embed.go
esevastyanov: There is no error in case of a graceful shutdown

begelundmuller (approved):

PR #5348, opened 2024-07-30
feat: Add support for percent of totals in MetricsViewAggregation

AdityaHegde:
To add support for percent of totals based filter criteria in alerts, we need to add percent of totals as a variant of a measure.

This PR adds that support to MetricsViewAggregation by calculating totals and rewriting percent of totals with `measure/<calculated_total>`


proto/rill/runtime/v1/queries.proto
```diff
@@ -389,6 +389,7 @@ message MetricsViewAggregationMeasure {
     MetricsViewAggregationMeasureComputeComparisonValue comparison_value = 7;
     MetricsViewAggregationMeasureComputeComparisonDelta comparison_delta = 8;
     MetricsViewAggregationMeasureComputeComparisonRatio comparison_ratio = 9;
+    MetricsViewAggregationMeasurePercentOfTotal percent_of_total = 10;
```
begelundmuller, line 392: nit: put "compute" in the name like for the other computed measure types: `MetricsViewAggregationMeasureComputePercentOfTotal`

begelundmuller:
On reflection, I think this might be simpler to solve with an additional query. 

I think there might actually be a simple way to achieve that. The new `metricsview` package supports `Query`-level and `AST`-level rewrites. This might be achieved with a `Query`-level rewrite that looks for percent-of-total measures, and if it finds any, computes the total and rewrites the compute measure to contain the actual float value to divide by.

For inspiration, you can see this rewriter which turns relative time ranges into fixed time ranges, sometimes making a separate query to get the min/max timestamps of the metrics view: https://github.com/rilldata/rill/blob/b67474d7b4fd3770f21999f0c3e96a3e86c216b7/runtime/metricsview/executor_rewrite_time.go#L14-L16

runtime/metricsview/executor.go
```diff
@@ -184,6 +184,10 @@ func (e *Executor) Query(ctx context.Context, qry *Query, executionTime *time.Ti
 		return nil, err
 	}
 
+	if err := e.rewritePercentOfTotals(ctx, qry, e.metricsView, e.security); err != nil {
```
begelundmuller, line 187: This rewrite should be moved up above the `rewriteQueryDruidExactify` rewrite (since I think the exactify rewrite may try computing the measure)

runtime/metricsview/ast.go
```diff
@@ -372,6 +372,21 @@ func (a *AST) resolveMeasure(qm Measure, visible bool) (*runtimev1.MetricsViewSp
 		}, nil
 	}
 
+	if qm.Compute.PercentOfTotal != nil {
+		m, err := a.lookupMeasure(qm.Compute.PercentOfTotal.Measure, visible)
```
begelundmuller, line 376: For safety, let's add a check/error that `qm.Compute.PercentOfTotal.Total` is not `nil` at this point

runtime/metricsview/query.go
```diff
@@ -94,6 +98,11 @@ type MeasureComputeComparisonRatio struct {
 	Measure string `mapstructure:"measure"`
 }
 
+type MeasureComputePercentOfTotal struct {
+	Measure string  `mapstructure:"measure"`
+	Total   float64 `mapstructure:"total"`
```
begelundmuller, line 103: Change to `*float64` to distinguish between cases where the total was not computed, and where the total is `0`

runtime/metricsview/executor.go
```diff
@@ -291,6 +295,10 @@ func (e *Executor) Export(ctx context.Context, qry *Query, executionTime *time.T
 		return "", err
 	}
 
+	if err := e.rewritePercentOfTotals(ctx, qry, e.metricsView, e.security); err != nil {
```
begelundmuller, line 298: Also needs to be moved up above `rewriteQueryDruidExactify` here

runtime/metricsview/executor_rewrite_percent_of_totals.go
```diff
@@ -0,0 +1,111 @@
...
+	if err != nil {
+		return err
+	}
+	defer e.Close()
+
+	res, err := e.Query(ctx, totalsQry, nil)
+	if err != nil {
+		return err
+	}
+	defer res.Close()
```
begelundmuller, line 48: The exactify rewriter does a couple things differently that I think we should also do here:

1. Explicitly set all members of the `Query{}` struct to reduce chance of forgetting to handle one if it's changed in the future
2. Explicitly call `NewAST` and `e.olap.Execute` to a) prevent any chance infinite recursion, b) only apply the rewrites that make sense in this context (I think that would be only `rewriteApproxComparisons` and `rewriteDruidGroups`)

runtime/metricsview/executor_rewrite_percent_of_totals.go
```diff
@@ -0,0 +1,111 @@
...
+
+	for i, measure := range measures {
+		t, ok := rowMap[measure.Name]
+		if !ok {
+			return fmt.Errorf("measure %q didnt return data", measure.Name)
+		}
+		tf, ok := numberLikeToFloat64(t)
+		if !ok {
+			return fmt.Errorf("%q is not a number", measure.Name)
+		}
```
begelundmuller, line 68: If you create a slice of floats and pass it to `Scan`, you can lean into the `database/sql` drivers ability to scan numbers into a `float64`. Similar to this logic in the Parquet exporter: https://github.com/rilldata/rill/blob/36d9921ea0f2f13ebc3b262412853c834eff13a2/runtime/drivers/file/model_executor_olap_self.go#L295

It might look something like this:
```go
vals := make([]*float64, len(measures))
for i := range vals {
  vals[i] = new(float64)
}

for res.Next() {
  err := rows.Scan(vals...)
  // ...
}
```

begelundmuller (changes requested):

runtime/metricsview/ast.go
```diff
@@ -372,6 +372,21 @@ func (a *AST) resolveMeasure(qm Measure, visible bool) (*runtimev1.MetricsViewSp
...
+		if err != nil {
+			return nil, err
+		}
+
+		return &runtimev1.MetricsViewSpec_MeasureV2{
+			Name:               qm.Name,
+			Expression:         fmt.Sprintf("%s*100/%f", a.dialect.EscapeIdentifier(m.Name), qm.Compute.PercentOfTotal.Total),
+			Type:               runtimev1.MetricsViewSpec_MEASURE_TYPE_DERIVED,
```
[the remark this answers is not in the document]
AdityaHegde: 1. We are calculating the measure value itself in a subquery since it might be used for comparison columns as well. So adding the expression here will recompute it wont it?
2. Update to be safer.

runtime/metricsview/executor_rewrite_percent_of_totals.go
AdityaHegde: Will this not add some complexity? Any new rewrite need to be added in multiple places and the author needs to be aware of it. Not sure if it can be solved as is right now.

runtime/metricsview/ast.go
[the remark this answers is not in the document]
begelundmuller: Okay, if you think it will mostly be queried in requests that also query the plain measure value, then I agree the derived approach is better

runtime/metricsview/executor_rewrite_percent_of_totals.go
begelundmuller: Yeah it's a little annoying, but I think it's best to explicitly add the rewrites each time at this point. To avoid recursion dangers and because the rewrites are a very mixed bag (some require a specific order of other rewrites, some are expensive to run, some don't make sense in the context of certain calls).

If we end up adding more rewrites, it may make sense to refactor how they're configured to avoid the repetition.

begelundmuller (approved):

runtime/metricsview/ast.go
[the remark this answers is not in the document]
AdityaHegde: Ya. I would assume all instances of totals in dashboard right now can be replaced with this.

Issue #1567, opened 2023-01-04
Add the global status spinner back to the application


magorlick:
**Issue**
We have a regression where there is no longer a global status spinner to indicate the state of the server.  

**Importance**
This spinner is important because it helps users know if they have the right information in front of them that relates to the current query, not the prior query.

**Ask**
A spinner should be present at the top of the modeling view that spins when the server is actively processing data to surface for insights.

hamilton:
discussed solution: add some check to the svelte query middleware that updates a "global status" store

PR #5351, opened 2024-07-30
Delay loading spinner

### LINKED ISSUE #5345: Delay the loading spinner
As discussed, frequently the loading spinner flashes quickly before content is rendered. To avoid this flash, we should delay the loading spinner by ~300ms before showing it. We should delay by default, but allow specific instances to configure their own delay.

lovincyrus:
Reviewers - Check out this branch locally, and please let me know if any of the hot path's spinners shouldn't be delayed.

The goal is to eliminate the flashing/flickering spinners on page load. To do that, we only show the spinner if the loading state takes longer than the specified delay.

I've added two new helpers, `DelayedSpinner.svelte` ~~and `spinner-utils.ts`~~, to efficiently manage different spinner states. ~~Use `spinner-utils.ts` for managing loading states in useQuery.~~ `DelayedSpinner.svelte` works best for various other conditions, especially within inline tertiary conditionals.

The delayed spinner is handy for preventing the loading state from flickering. We've added a default of 300ms delay before actually displaying the spinner.

closes #5345

ref https://github.com/rilldata/rill/issues/1567

- [x] revise existing `<Spinner/>`

Before

https://github.com/user-attachments/assets/6ddeb861-1256-42b3-b7ab-aa66ae590884

After

https://github.com/user-attachments/assets/83156c49-eb42-4a14-9b46-14cb445df85a

Also notably snappy big number (no flickering spinner and no data text)

https://github.com/user-attachments/assets/c639a5c4-7b57-47bd-90aa-7238b73e4e5a


web-admin/src/features/alerts/listing/AlertsTable.svelte
```diff
@@ -11,12 +11,17 @@
...
+  import { isLoadingWithTimeout } from "@rilldata/web-common/features/entity-management/spinner-utils";
 
   export let organization: string;
   export let project: string;
 
   $: alerts = useAlerts($runtime.instanceId);
 
+  const loadingWithTimeout = isLoadingWithTimeout();
+  $: loadingWithTimeout.setLoading($alerts.isLoading);
+  $: loadingAlerts = $loadingWithTimeout;
```
ericpgreen2, line 23: Thoughts on updating the `isLoadingWithTimeout()` function to make this a one-liner?

lovincyrus: sure, I can do that. This helps us remove the boilerplate-y code.

`$: loading = isLoadingWithTimeout(isLoading)` is pretty straight forward.

lovincyrus:
In response to https://github.com/rilldata/rill/pull/5351#discussion_r1700997760, I tried simplifying the code into a one-liner without success. However, I managed to boil it down to `<DelayedSpinner />` so we only have one abstraction going forward.

We could integrate the delayed spinner within nested conditional statements more efficiently:

```svelte
{#if isLoading}
  <DelayedSpinner isLoading={isLoading} />
{:else if $isError}
  <Error />
{:else if $isSuccess}
  {#if $data.length === 0}
    <Empty />
  {:else}
    <Data />
  {/if}
{/if}
```

Curious for your thoughts @ericpgreen2

web-admin/src/features/alerts/listing/AlertsTable.svelte
lovincyrus: See: https://github.com/rilldata/rill/pull/5351#issuecomment-2266283116

jkhwu:
This looks like an improvement to me! Thank you.

web-common/src/features/entity-management/DelayedSpinner.svelte
```diff
@@ -0,0 +1,42 @@
...
+    }
+  });
+
+  onDestroy(() => {
+    clearTimeout(timeoutId);
+  });
+</script>
+
+{#if $showSpinner}
+  <Spinner status={EntityStatus.Running} {duration} {size} />
```
AdityaHegde, line 41: We should get rid of `EntityStatus`, it was a concept when we had nodejs backend. Perhaps copy over the contents of `Spinner.svelte` in here as a start?

web-common/src/layout/BlockingOverlayContainer.svelte
```diff
@@ -32,7 +32,10 @@
         style:font-size="48px"
       >
         <div class="on" style="--length: {2000 + Math.random() * 5000}ms;">
-          <Spinner {status} duration={300 + Math.random() * 200} />
+          <DelayedSpinner
```
AdityaHegde, line 35: Does it make sense here? If a blocking overlay is present it should always have a spinner no?

web-common/src/features/entity-management/DelayedSpinner.svelte
```diff
@@ -0,0 +1,42 @@
...
+    } else {
+      showSpinner.set(false);
+    }
+  }
+
+  onMount(() => {
+    if (isLoading) {
+      timeoutId = setTimeout(() => showSpinner.set(true), delay);
+    }
+  });
```
ericpgreen2, line 33: Looks like this is redundant to the reactive statement, so it can be removed

ericpgreen2 (approved):
Sweet 👍. I agree with Aditya that we should deprecate the `EntityStatus` concept, but that's slightly wider in scope, so perhaps that merits a standalone follow-up PR. Approving so you can merge at your discretion.

web-common/src/features/entity-management/DelayedSpinner.svelte
lovincyrus: Will deprecate the `EntityStatus` in a follow-up.

web-common/src/layout/BlockingOverlayContainer.svelte
lovincyrus: https://github.com/rilldata/rill/pull/5351/commits/e352da7bde1f34e6c06100ed614ac52892aa069c

PR #5358, opened 2024-07-31
feat: Add dashboard state to error telemetry to better debug errors

AdityaHegde:
closes https://github.com/rilldata/rill-private-issues/issues/511

### DIFF
--- a/web-common/src/metrics/ErrorEventHandler.ts
+++ b/web-common/src/metrics/ErrorEventHandler.ts
@@ -33,7 +33,7 @@ export class ErrorEventHandler {
         error.status ?? "",
         error.message ?? "unknown error",
         screenName,
-        this.getDashboardState(),
+        get(page).url.toString(),
       );
       return;
     } else {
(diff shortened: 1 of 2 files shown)


web-common/src/metrics/ErrorEventHandler.ts
```diff
@@ -133,6 +142,12 @@ export class ErrorEventHandler {
       screenName,
       stack,
       message,
+      dashboardState,
     ]);
   }
+
+  private getDashboardState() {
+    // TODO: we need to also save other params that get added after
```
ericpgreen2, line 150: What about just storing the whole URL? The main goal is to make it easier to reproduce bugs, so it'd be great if we can just copy/paste the URL into our browser.

AdityaHegde: Ya i thought we save org & project. I guess not, so it makes sense to do this

ericpgreen2 (approved):
Good from the frontend POV. @esevastyanov, can you please augment the data pipeline with this new `pageURL` field?

nishantmonu51 (approved):

nishantmonu51:
merging it, Lets create a separate ticket for the pipeline work.

esevastyanov:
Created a [ticket](https://github.com/rilldata/rill-private-issues/issues/525)

PR #5359, opened 2024-07-31
feat: Add pylon to rill developer

AdityaHegde:
With the new login/logout within rill developer we also can support pylon chat support in there.

This PR adds it to the avatar dropdown and deploy screen when there is a failure.


web-common/src/features/help/initPylonChat.ts
```diff
@@ -1,4 +1,4 @@
-import type { V1User } from "../../client";
+import type { V1User } from "@rilldata/web-admin/client";
```
ericpgreen2, line 1: `web-common` should not import anything from `web-local` or `web-admin`. Now that Rill Developer has more interaction with the admin server, I wonder if we should move the `web-admin/client` to `web-common` and rename `web-admin` to `web-cloud` for clarity. (Though this should certainly happen in a separate PR.)

ericpgreen2 (commented):
UXQA –
1. When you're logged-out of Rill Developer, the "Contact Rill Support" button doesn't work because there's no email address to use to initialize the Pylon chat. Like the Cloud UI, we should hide it for now. Unless @ericokuma @mindspank @AndrewRTsao @jkhwu have something else in mind.
2. I notice we have duplication between the legacy Rill Developer footer & the menu in the Local Avatar Button. @ericokuma @mindspank, can we remove "Documentation" and "Ask a question" from the legacy footer? And can we move "Report an issue" from the legacy footer to the avatar menu?

web-common/src/features/help/initPylonChat.ts
AndrewRTsao: Agreed we should hide Pylon if they're logged out. Two options that come to mind:
1. We simply hide the Contact Rill Support button (and/or use the previous link to community)
2. We keep the Contact Rill Support button but let them email support@rilldata.com instead (we could trigger an email form while requiring them to share an email address)

If trivial, I think option 2 might be better to help build engagement (and we would still retain attribution) but will let Product make the final call.

AdityaHegde: @ericpgreen2 Ya makes sense. Update to not use admin's model. For moving clients to common, I disagree. We can only use those clients on cloud. Any communication has to go through the local server.

@AndrewRTsao Hiding the button for now. We can do a follow up to send an email in cloud as well.

mindspank: It's fine to remove the duplicated entries in the footer

web-common/src/features/help/initPylonChat.ts
```diff
@@ -1,11 +1,15 @@
-import type { V1User } from "../../client";
-
 const appId = import.meta.env.RILL_UI_PUBLIC_PYLON_APP_ID as string;
 
+export type UserLike = {
+  email: string;
+  displayName: string;
+  photoUrl: string;
+};
+
```
ericpgreen2, line 8: Nice solution to the import problem. Mind just adding a comment to explain the situation?

ericpgreen2 (approved):

PR #5364, opened 2024-07-31
Disable big number tooltip in expanded time series metrics

### LINKED ISSUE #5352: Flickering big number tooltip when switching views
**Describe the bug**
The big number tooltip flickers when switching between Overview and Measure views. See recording.

**To Reproduce**
1. Go to a project
2. Import https://gist.githubusercontent.com/nicosuave/1f2ce1bf82d6461151e9531f0f0a5206/raw/73438edfc469d90a7d8e56d6564905be2f51890a/trips.csv
3. Click on the measure
4. Now click on ← Overview in the measure view
5. Once you're back in the overview, repeat step 2 and step 3
6. Pay attention to the top left, the tooltip is anchored to the top left of the page.

**Expected behavior**
The tooltip shouldn't anchor to the top left of the page when switching between Overview and Measure view.

**Recording**

https://github.com/user-attachments/assets/6612b07b-b7f6-4431-a0b1-dd7e51716945

lovincyrus:
Disable the big number tooltip when switching between Overview and Measure views. Fixes #5352

See before: https://github.com/rilldata/rill/issues/5352#issue-2438372614

After

https://github.com/user-attachments/assets/725aa77e-46bf-4f9c-80d5-dc889e52e40e

### DIFF
--- a/web-common/src/features/dashboards/time-series/MetricsTimeSeriesCharts.svelte
+++ b/web-common/src/features/dashboards/time-series/MetricsTimeSeriesCharts.svelte
@@ -372,7 +372,6 @@
           <MeasureBigNumber
             {measure}
             value={bigNum}
-            isMeasureExpanded={isInTimeDimensionView}
             {showComparison}
             {comparisonValue}
             errorMessage={$timeSeriesDataStore?.error?.totals}
--- a/web-common/src/features/dashboards/time-series/MetricsTimeSeriesCharts.svelte
+++ b/web-common/src/features/dashboards/time-series/MetricsTimeSeriesCharts.svelte
@@ -372,6 +372,7 @@
           <MeasureBigNumber
             {measure}
             value={bigNum}
+            isMeasureExpanded={isInTimeDimensionView}
             {showComparison}
             {comparisonValue}
             errorMessage={$timeSeriesDataStore?.error?.totals}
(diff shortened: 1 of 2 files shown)


web-common/src/features/dashboards/time-series/MetricsTimeSeriesCharts.svelte
```diff
@@ -372,7 +372,6 @@
           <MeasureBigNumber
             {measure}
             value={bigNum}
-            isMeasureExpanded={isInTimeDimensionView}
```
djbarnwal, line 375: We need this prop for the "Big Number" component because the component is a button in the explore screen and a div in the Time dimension detail screen. It's also styled a bit differently (tile vs plain). Removing this also removes that behavior.

djbarnwal (changes requested):

web-common/src/features/dashboards/time-series/MetricsTimeSeriesCharts.svelte
lovincyrus: Ah, I noticed the style changes when the prop is removed. Passing isInTimeDimensionView to BigMeasureNumber causes the tooltip to flash briefly during transitions from Overview to Measure view. Ideally, we should disable the tooltip during these page transitions.

I'll explore other ways to remove the tooltip flashes and keep the div/button style.

lovincyrus: Addressed

https://github.com/user-attachments/assets/15252cab-7256-4224-865a-6f17ca4e4c0d

lovincyrus:
Demo:

https://github.com/user-attachments/assets/15252cab-7256-4224-865a-6f17ca4e4c0d

djbarnwal (approved):
Looks good!

Issue #5039, opened 2024-06-06
readonly mode should not allow arbitrary file listing or listing sensitive project variables/env values


dweinstein:
**Is your feature request related to a problem? Please describe.**
rill CLI currently starts a server that allows listing files (e.g., `.env` and allows retrieval of (potentially) sensitive project variables, like postgresql URL with username/passwords)

**Describe the solution you'd like**
Readonly mode should only display dashboards and allow basic export data.

**Describe alternatives you've considered**

**Additional context**
A developer could browser to a website, while using rill developer CLI, (`rill start`) that makes an XHR to the developer's machine via `fetch('http://localhost:9009/v1/instances/default?sensitive=true')` to grab the values and possibly gain access to sensitive information. This could leak sensitive information

mindspank:
@begelundmuller related to what we discussed previously.

begelundmuller:
So there are two problems here:
1. The regular experience does not apply CORS and CSP policies that guard against XHR to localhost from other websites. Changing it to only allow requests from `localhost` may break the experience for people who self-host local Rill today.
2. The `--readonly` flag does not block access to endpoints that expose access to variables or the local file system.

begelundmuller:
Since the local Rill experience is not really designed for self-hosting (even thought it's possible), we will prioritize problem 2 separately. However, we should fix the CORS and CSP issue immediately.

@Kavinjsir The problem here is basically that the current CORS policy allows calls from all origins. The changes needed are:
- Only allow calls from the HTTP host, which is `localhost:9009` by default (replacing port 9009 with the port passed to `--port`).
- Support additional CORS origins with a `rill start --allowed-origins host1,host2,...` flag to not break self-hosted instances of Rill.
- Investigate and maybe add a CSP policy for the local UI. We have one for the cloud UI, but not the local one. I'm not entirely sure what if anything it should contain – maybe browser defaults are sufficient.

Here are some pointers for code to edit:

- The entrypoint for the `rill start` command is here: https://github.com/rilldata/rill/blob/b67474d7b4fd3770f21999f0c3e96a3e86c216b7/cli/cmd/start/start.go#L25
- When running `rill start`, it initializes and starts the runtime's HTTP server with a few additional local endpoints mounted. It currently configures the allowed origins to `*` here: https://github.com/rilldata/rill/blob/b67474d7b4fd3770f21999f0c3e96a3e86c216b7/cli/pkg/local/app.go#L380
[...]

PR #5365, opened 2024-08-01
refactor: apply cors

Kavinjsir:
### Description

1. Replace `cors allowed-origins` with default localHttpUrl
2. Add a flag `--allowed-origins` to input additional `allowed-origins`

### Motivation

Resolve https://github.com/rilldata/rill/issues/5039

### DIFF
--- a/cli/cmd/start/start.go
+++ b/cli/cmd/start/start.go
@@ -36,6 +36,7 @@ func StartCmd(ch *cmdutil.Helper) *cobra.Command {
 	var logFormat string
 	var env []string
 	var vars []string
+	var allowedOrigins []string
 	var tlsCertPath string
 	var tlsKeyPath string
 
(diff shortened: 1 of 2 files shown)


Kavinjsir:
I think this PR is ready for review.
One concern left is `CSP`.
It looks the current one has well defined:
https://github.com/rilldata/rill/blob/v0.47.5/netlify.toml#L6-L14

Perhaps the only concern is ever want to drop `unsafe-eval` or `unsafe-inline`.
cc: @himadrisingh would you have some suggestions?

Kavinjsir:
> > We have third party dependencies which require unsafe-eval, unless we get them compliant with cso it would be difficult.
> 
> @himadrisingh if vega is the mentioned one, it seems they have a resolution, I tried opening a PR for discussion at: ~#4784~ #5376

As @djbarnwal has already made great contribution and there has been an upstream PR for vega to review, I think we can deal with CSP later.

himadrisingh:
Did we test it locally and see if it fixes the issue.

himadrisingh (approved):

Kavinjsir:
> Did we test it locally and see if it fixes the issue.

@himadrisingh Yeah, here’s what I tested locally:

1. **HOST A**:  ran the compiled `rill` CLI based on this PR with the following command:

```shell
go run ./cli start --verbose --debug <local-rill-project-path> --allowed-origins="http://localhost:3000" --allowed-origins="http://localhost:4000"
```

This sets the allowed origins to `["localhost:3000", "localhost:4000"]`.

2. **HOST B**: set up a simple HTTP server using:

```shell
python3 -m http.server 4003
```

With an HTML file like this:

```html
<!DOCTYPE html>
<html>

<head>
  <title>
    CORS Test
  </title>
</head>

<body>
  <script>
    fetch("http://<HOST-A-IP>:9009/v1/instances/default?sensitive=true")
      .then(res => res.json())
      .then(data => console.log(data))
      .catch(err => console.error("Error: ", err));
  </script>
</body>

</html>
```

When the port of **HOST B** is set to `3000` or `4000`, the script inside the HTML file successfully fetches results from the endpoint. For any other port value, it fails as expected.

PR #5366, opened 2024-08-01
Public URLs Table UI

### LINKED ISSUE #5328: Public URLs: Add a page to list/manage
[Figma mock](https://www.figma.com/design/Qt6EyotCBS3V6O31jVhMQ7/RILL?node-id=15019-622855&t=TVF5Fzm10rNjF5Ik-0)

Requirements
- Create a page at `ui.rilldata.com/[org]/[project]/-/public-urls`
- List all Public URLs in a table
- Include:
    - dashboard to which the URL applies
    - expiration time (if any)
    - a delete action

Notes by Cyrus
- Ref: https://github.com/rilldata/rill/pull/5283
- Ref: https://www.notion.so/rilldata/Support-for-shared-filtered-dashboard-URLs-6490b207cb8e43278b9992ebf11d6195#c4bbda6182514673a6857500bd2be820
- Ref: https://docs.rilldata.com/notes/0.47#public-links
- Ref: https://rilldata.slack.com/archives/C04JP568RR6/p1722288927983179
- Can ping benjamin for platform support

lovincyrus:
Closes #5328

This PR adds a new public URLs table to the project settings page.

- [x] Rename CreateShareableURLForm.svelte references
- [x] Add project settings tab (only accessible by admin)
- [x] Create /[org]/[project]/-/settings
- [x] Check table columns
- [x] Style settings page (left side nav, content)
- [x] Once a URL is created, a toast notification should appear saying, "Go to Public URLs."
- [x] Check table actions for deletion
- [x] Enable `/${organization}/-/settings`
- [x] Refetch queries after creating/deleting public URL
- [x] Rerender magic auth tokens after deletion
- [x] Blocked by https://github.com/rilldata/rill-private-issues/issues/642
- [x] Row click to open public URL
- [x] Support [forward cursor-based pagination](https://github.com/rilldata/rill/pull/5366/commits/4acb8b7d10425270ce4b19f58d7e52b156b47b96)
- [x] Make column name clickable
- [x] Convert actions row to dot icon menu
- [x] Add public URL name edit to the creation flow
- [x] Add new Public URL name column
- [x] https://www.figma.com/design/Qt6EyotCBS3V6O31jVhMQ7/RILL?node-id=13077-805393&node-type=frame&t=A7iL1ItEjIm3473q-0
- [x] Handle no token, no URL case
- [x] Confirm deletion
- [x] Address feedback
- [x] Check pagination
- [x] Rename title → name in magic auth token in the API and database (https://github.com/rilldata/rill/pull/5366#issuecomment-2345049467)
- [x] Use dashboard name for "Dashboard name" col
- [x] Revisit pagination controls based on https://www.figma.com/design/Qt6EyotCBS3V6O31jVhMQ7/RILL?node-id=15019-622855&node-type=frame&t=ZKlKJUwaeOWIUQdh-0
- [x] actions menu https://www.figma.com/design/Qt6EyotCBS3V6O31jVhMQ7/RILL?node-id=16418-896007&node-type=symbol&t=y9nsmi8GgKDf9iL2-0
- [x] ~~add `previousPageToken`~~
- [x] change 'name' back to 'title' in the magic auth token https://github.com/rilldata/rill/pull/5366#issuecomment-2352568775
- [x] [delete confirmation dialog](https://rilldata.slack.com/files/U04S15TUQJG/F07MGU13KHA/screenshot_2024-09-13_at_4.03.54___pm.png)
- [x] support scrolling
- [x] consider `createInfiniteQuery`
- [x] revisit to remove createdOn if possible
- [x] sticky?

Ability to edit name to the public URL creation
CleanShot 2024-09-11 at 14 12 04@2x

Support title in magic auth token to display Public URL name in the management table
https://github.com/rilldata/rill/pull/5366/commits/34b41718edad96633c0acbd0133a38c9550d864a
```diff
export interface V1MagicAuthToken {
  id?: string;
  projectId?: string;
  url?: string;
  token?: string;
  createdOn?: string;
  expiresOn?: string;
  usedOn?: string;
  createdByUserId?: string;
  createdByUserEmail?: string;
  attributes?: V1MagicAuthTokenAttributes;
  metricsView?: string;
  metricsViewFilter?: V1Expression;
  metricsViewFields?: string[];
  state?: string;
+ title?: string;
}
```
CleanShot 2024-09-11 at 17 30 06@2x


begelundmuller:
@lovincyrus This PR has now added support for accessing the full URLs of previously created shareable URLs: https://github.com/rilldata/rill/pull/5621. You can access the URLs in the `url` property in the objects returned from the `ListMagicAuthTokensRequest` RPC.

lovincyrus:
Demo:

https://github.com/user-attachments/assets/86375bc0-8030-4d8a-a45a-3c395b4b2b39

lovincyrus:
> the `url` will only be available for newly created shareable URLs, not for ones created before this PR. So the UI will need to handle that case gracefully.

Sanity check, @jkhwu - since this is a fairly new feature, I think it's okay to say that we don't need to handle the case for the previously created public URLs. If we do want to handle the case, here's what the empty case looks like:

CleanShot 2024-09-10 at 10 16 52@2x

jkhwu:
Watched the demo and it looks good to me @lovincyrus . The empty state that you proposed looks fine to me, good copy decisions.

ericpgreen2 (changes requested):
UXQA –
- [x] I have a table of previously-created Public URLs. It's confusing that the row has a "pointer" cursor, yet clicking on the row does nothing. [See Jam](https://jam.dev/c/f10233f1-837d-499f-9f90-099379679a25).
- [x] We should use the dashboard `title`, if one exists, else fallback to the `name`
- [x] It's unclear to me how these URLs are sorted. After creating a URL, I wasn't sure where I'd find it in the list. (CC @jkhwu)
    image
- [x] I deleted a bunch of URLs, then the "Public URL deleted" notification hung on the screen:
    image
- [x] When there are more than 20 URLs, the table doesn't offer pagination or somehow indicate that there more results not shown. (CC @jkhwu)
- [x] It'd be nice if we could use an `<a>` tag to navigate the URL, so that on:hover the browser shows the URL in the bottom-left corner. Rather than the whole row being clickable, I wonder if we should have a column with a button that says "Copy URL". (CC @jkhwu)

lovincyrus:
> 1. I have a table of previously-created Public URLs.

Oh - we need to handle this for sure.

lovincyrus:
Tagging @begelundmuller and @pjain1 for a review - I've added `title` support to the magic auth token for displaying the Public URL name in the management table. Ref: https://github.com/rilldata/rill/commit/34b41718edad96633c0acbd0133a38c9550d864a

```diff
export interface V1MagicAuthToken {
  id?: string;
  projectId?: string;
  url?: string;
  token?: string;
  createdOn?: string;
  expiresOn?: string;
  usedOn?: string;
  createdByUserId?: string;
  createdByUserEmail?: string;
  attributes?: V1MagicAuthTokenAttributes;
  metricsView?: string;
  metricsViewFilter?: V1Expression;
  metricsViewFields?: string[];
  state?: string;
+ title?: string;
}
```

Originally, I added `title` for the dashboard title, but after discussions with Janet and Eric O, we also need an additional column for 'Public URL name'.

Should I instead add two more fields to support those requirements?

```diff
export interface V1MagicAuthToken {
  id?: string;
  projectId?: string;
  url?: string;
  token?: string;
  createdOn?: string;
  expiresOn?: string;
  usedOn?: string;
  createdByUserId?: string;
  createdByUserEmail?: string;
  attributes?: V1MagicAuthTokenAttributes;
  metricsView?: string;
  metricsViewFilter?: V1Expression;
  metricsViewFields?: string[];
  state?: string;
+ urlName?: string;
+ dashboardName?: string;
}
```

CleanShot 2024-09-11 at 17 52 49@2x

begelundmuller:
The changes here look good to me. About this:
> Originally, I added title for the dashboard title, but after discussions with Janet and Eric O, we also need an additional column for 'Public URL name'.
> 
> Should I instead add two more fields to support those requirements?

It makes sense to give the public URL a name, but I don't think we should add extra fields for capturing the dashboard title. It may be edited and would get out of sync.

You can either do a separate request at render time to get the dashboard _title_ (slow for large lists), or would have to limit to showing just the public URL _title_ and dashboard _name_.

lovincyrus:
Latest demo:

https://github.com/user-attachments/assets/abec9730-ab7b-4134-8d1a-e1d3ae9c6c7d

lovincyrus:
@ericpgreen2 Please report any anomalies if you find any in the previously generated public URLs.

ericpgreen2:
> @ericpgreen2 Please report any anomalies if you find any in the previously generated public URLs.

Here are what the previously-created URLs look like:
image

No _anomalies_ per se, but there's no an obvious link available to click. I supposed I was envisioning an explicit "Copy link" button being in-line, not hidden behind a dropdown menu.

lovincyrus:
> > @ericpgreen2 Please report any anomalies if you find any in the previously generated public URLs.
> 
> Here are what the previously-created URLs look like: image
> 
> No _anomalies_ per se, but there's no an obvious link available to click. I supposed I was envisioning an explicit "Copy link" button being in-line, not hidden behind a dropdown menu.

Sweet! I've hidden the ability to copy URL based on the availability of the constructed `url`. The previously created public urls will have undefined url, as per https://github.com/rilldata/rill/pull/5366#issuecomment-2340039866.

Quick note - if url is available, the public url name cell will be clickable and there is a Copy menu item in the actions dropdown.

Re: "not hidden behind a dropdown menu." — I've already synced with @jkhwu and @ericokuma extensively on the consistency of actions ui in our table components. We will use 3 dots icon in the public URL management table.

web-admin/src/features/public-urls/PublicURLsTable.svelte
```diff
@@ -0,0 +1,268 @@
...
+      header: "Public URL name",
+      cell: ({ row }) =>
+        flexRender(DashboardLink, {
+          href: row.original.url,
+          title: row.original.name,
+        }),
+    },
+    {
+      accessorFn: (row) => row.dashboardTitle,
+      header: "Dashboard name",
```
ericpgreen2, line 60: We should label this "Dashboard title" because "name" means "unique identifier" in Rill resources.

ericpgreen2:
Some more UXQA & design feedback:
1. It's confusing what the text input element represents. IMO there should be a label "Public URL name" maybe with a helper description/tooltip e.g. "Give your URL a name to help you identify it later" (CC @jkhwu)
    image
2. When I clicked through the sorting states & landed on the "unsorted" state, I was confused how the URLs were sorted (they have to be sorted by _something_, right?). [Jam](https://jam.dev/c/484a76e4-e058-4f9f-b78f-00955299588b). (CC @jkhwu)
3. There's no "Previous" button to get back to the first page of the table.
4. I'd expect the table sort to apply across all pages, not just the current one.
5. It'd be nice to store the table page & sort state in the URL

web-admin/src/routes/[organization]/[project]/-/settings/public-urls/+page.svelte
```diff
@@ -0,0 +1,102 @@
...
+  $: currentPageTokens =
+    $magicAuthTokens.data && $dashboards.data
+      ? $magicAuthTokens.data.tokens.map((token) => {
+          const dashboard = $dashboards.data.find(
+            (d) => d.resource.meta.name.name === token.metricsView,
+          );
+          return {
+            ...token,
+            dashboardTitle:
+              dashboard?.resource.metricsView.spec.title || "Unknown Dashboard",
```
Comment (ericpgreen2) [inline web-admin/src/routes/[organization]/[project]/-/settings/public-urls/+page.svelte:43]: 1. We don't currently do it throughout the codebase, but FYSA that we _should_ reference the Metrics View's `state.validSpec` rather than it's `spec`
2. When there's no `title` we can fallback to the `name`, which is required: `dashboard?.resource.metricsView.spec.title || dashboard?.resource.meta.name.name`

web-admin/src/features/public-urls/PublicURLsTable.svelte
```diff
@@ -0,0 +1,268 @@
...
+      day: "numeric",
+      hour: "numeric",
+      minute: "numeric",
+    });
+  }
+
+  const columns: ColumnDef<MagicAuthTokenProps, any>[] = [
+    {
+      accessorKey: "name",
+      header: "Public URL name",
```
ericpgreen2, line 51: As below, because this "name" isn't a unique identifier, perhaps we should call it "Public URL label"

web-admin/src/features/public-urls/DashboardLink.svelte
```diff
@@ -0,0 +1,12 @@
+<script lang="ts">
+  export let href: string;
+  export let title: string;
+</script>
+
+{#if title}
+  <a {href} class="text-blue-600 hover:underline">
```
ericpgreen2, line 7: Nit: maybe we should open these links in a new tab?

jkhwu:
Syncing with @ericokuma and reading through this comment thread. Responses:

1. @ericpgreen2 @lovincyrus the presentation of the URL creation dialog is intended to look like this: https://www.figma.com/design/Qt6EyotCBS3V6O31jVhMQ7/RILL?node-id=13077-805393&node-type=frame&t=p88wi9zSyvDbXoMO-11

Screenshot 2024-09-12 at 1 55 37 PM

2. It appears that standard practice is for tables to go back to page 1 when changing sort type. Even better would be to have enumerated page numbers, but we can incorporate that later if that's not supported yet.

ericokuma:
public URL title and dashboard name is what we would want :)

Is there a way to provide  `previousPageToken. Feels weird and broken to only have next pagination

jkhwu:
Also please note that the "Share" button is designed to have an active state for when this dialog is open underneath: https://www.figma.com/design/Qt6EyotCBS3V6O31jVhMQ7/RILL?node-id=13058-552865&node-type=frame&t=p88wi9zSyvDbXoMO-11
Screenshot 2024-09-12 at 2 01 22 PM

jkhwu:
Re: the previous URLs with "-" for their name. Could we just make the "-" dashes into clickable links?

lovincyrus:
> Re: the previous URLs with "-" for their name. Could we just make the "-" dashes into clickable links?

There are no exposed URLs associated with the previously generated URLs, https://github.com/rilldata/rill/pull/5366#issuecomment-2340039866

lovincyrus:
> Is there a way to provide `previousPageToken`. Feels weird and broken to only have next pagination

We would need a separate ticket for this to support `previousPageToken`. cc: @begelundmuller 

Nice to have on the API response is the `totalSize` as well. https://www.figma.com/design/Qt6EyotCBS3V6O31jVhMQ7?node-id=15019-622855#947870789

To my knowledge, many of our existing list endpoints only supply `nextPageToken` https://github.com/rilldata/rill/blob/main/web-admin/src/client/gen/index.schemas.ts#L895-L898

web-admin/src/routes/[organization]/[project]/-/settings/public-urls/+page.svelte
Comment (lovincyrus) [reply web-admin/src/routes/[organization]/[project]/-/settings/public-urls/+page.svelte:43]: Okay, will use validSpec's title

```
dashboardTitle: dashboard?.resource.metricsView.state?.validSpec?.title || dashboard?.resource.meta.name.name,
```

lovincyrus:
> When I clicked through the sorting states & landed on the "unsorted" state, I was confused how the URLs were sorted (they have to be sorted by something, right?). [Jam](https://jam.dev/c/484a76e4-e058-4f9f-b78f-00955299588b). (CC @jkhwu)

They are initially sorted by `createdOn` and whatever tokens we get from the results based on the `pageSize` (16). The subsequent results are fetched using the `nextPageToken`.

https://github.com/user-attachments/assets/ac38afaa-7efb-4bce-ae54-6a7cb5135808

lovincyrus:
Given that we have `nextPageToken` in the API response, I'll switch from previous/next to infinite scrolling. Synced with @jkhwu offline and agreed that we should just scroll the page.

```
export interface V1ListMagicAuthTokensResponse {
  tokens?: V1MagicAuthToken[];
  nextPageToken?: string;
}
```

begelundmuller:
> Is there a way to provide `previousPageToken`. Feels weird and broken to only have next pagination

It's common to support pagination only in one direction, see for example Google's API style guide, which we try to follow: https://cloud.google.com/apis/design/design_patterns#list_pagination.

The way this is usually represented in a UI is using infinite scroll/paging, and then after a page refresh (which drops browser state), showing a "go back to top" button.

If needed, we can support backwards pagination for certain endpoints (it depends on what the underlying queries look like and how much data is being scanned), but it's not a quick change.

> @begelundmuller wait...so how would user today share the URL with their clients/partners?

Until recently, we created a URL and showed it to the user (so they could copy and share it), but only persisted a secure hash of the URL's token in the database. This is a common security practice and similar to how we store other access tokens (like user sessions and service tokens).

Based on the requirements here, we made an exception for public URL tokens and started storing the actual token in the database as well, so URLs created going forward can be listed and re-accessed. But we cannot recover the original token for the previously created public URLs.

lovincyrus:
> It's common to support pagination only in one direction, see for example Google's API style guide, which we try to follow: https://cloud.google.com/apis/design/design_patterns#list_pagination.
> 
> I believe the way this is usually represented in a UI is using infinite scroll/paging, and then after a page refresh (which drops browser state), showing a "go back to top" button.

Makes sense. I was going to switch to infinite scrolling anyway. Hmm, this PR would definitely benefit from an endpoint that allows `createInfiniteQuery` and a server-side `sortBy` and `sortOrder`. @begelundmuller 

Something like this perhaps.
```diff
interface ListMagicAuthTokensParams {
  pageSize?: number;
  pageToken?: string;
+ sortBy?: string;
+ sortOrder?: 'asc' | 'desc';
}
```

ref: https://github.com/TanStack/virtual/blob/main/examples/svelte/infinite-scroll/src/InfiniteRows.svelte#L7
ref: https://github.com/TanStack/query/blob/dac5da5416b82b0be38a8fb34dde1fc6670f0a59/packages/svelte-query/src/createInfiniteQuery.ts#L16

lovincyrus:
Added `createAdminServiceListMagicAuthTokensInfiniteQuery` to support `createInfiniteQuery`. Also, removed previously initial sorted by `createdOn` because it will arbitrary sort the data as the user scrolls down the page, this causes item shifting around (not good UX).

Demo:

https://github.com/user-attachments/assets/41967783-bfdf-4560-9826-8336ebc818fe

begelundmuller:
> Makes sense. I was going to switch to infinite scrolling anyway. Hmm, this PR would definitely benefit from ~an endpoint that allows `createInfiniteQuery`~ (I'll explore this) and a server-side [`sortBy`](https://cloud.google.com/apis/design/design_patterns#sorting_order), `sortOrder`, and `query`. @begelundmuller

Yeah that should be possible, there's some good inspiration for that in the Google API style guide as well:
- https://cloud.google.com/apis/design/design_patterns#sorting_order
- https://google.aip.dev/160

Is this high priority? It's not a quick change (probably a day or so of effort).

lovincyrus:
> > Makes sense. I was going to switch to infinite scrolling anyway. Hmm, this PR would definitely benefit from ~an endpoint that allows `createInfiniteQuery`~ (I'll explore this) and a server-side [`sortBy`](https://cloud.google.com/apis/design/design_patterns#sorting_order), `sortOrder`, and `query`. @begelundmuller
> 
> Yeah that should be possible, there's some good inspiration for that in the Google API style guide as well:
> 
> * https://cloud.google.com/apis/design/design_patterns#sorting_order
> * https://google.aip.dev/160
> 
> Is this high priority? It's not a quick change (probably a day or so of effort).

Not a high priority. For context, I would say these features will be needed in the project settings subpages (public URLs, environment variables, user management, etc.) as we expand our coverage for table component.

web-admin/src/features/public-urls/CreatePublicURLForm.svelte
```diff
@@ -86,6 +97,18 @@
...
+            getAdminServiceListMagicAuthTokensQueryKey(organization, project),
+          );
+
+          eventBus.emit("notification", {
+            message: "Public URL created",
+            link: {
+              href: `/${organization}/${project}/-/settings/public-urls`,
+              text: "Go to public URLs",
+            },
+          });
```
ericpgreen2, line 111: UX feedback: I'd prefer we didn't show this notification. After I submit the "Create and copy URL" button, my next desired action is either 1) to open a new tab and paste in the URL to double-check it, or 2) paste the URL into a messaging app. Seeing this prompt to navigate to the management page makes me feel conflicted.

web-admin/src/features/public-urls/CreatePublicURLForm.svelte
```diff
@@ -86,6 +97,18 @@
             `${window.location.origin}/${organization}/${project}/-/share/${token}`,
             "URL copied to clipboard",
           );
+
+          void queryClient.refetchQueries(
+            getAdminServiceListMagicAuthTokensQueryKey(organization, project),
+          );
```
ericpgreen2, line 103: `invalidateQueries` could be better, so we don't refetch it if we don't have to

web-admin/src/routes/[organization]/[project]/-/settings/public-urls/+page.svelte
```diff
@@ -0,0 +1,103 @@
...
+        message: "Error deleting public URL",
+        type: "error",
+      });
+    }
+  }
+</script>
+
+<div class="flex flex-col w-full">
+  <div class="flex md:flex-row flex-col gap-6">
+    <div class="w-full">
```
Comment (ericpgreen2) [inline web-admin/src/routes/[organization]/[project]/-/settings/public-urls/+page.svelte:80]: Looks like some extra `div` wrappers

ericpgreen2 (approved):
This is great! Nice work 🙌

ericokuma:
Looks great, good work here @lovincyrus !

PR #5368, opened 2024-08-01
fix: backwards compatibility for alerts

AdityaHegde:
Alerts created before 0.48 cannot be opened anymore. Also dashboard name is `undefined` even when created in the new release.


proto/buf.gen.ui.yaml
```diff
@@ -4,6 +4,6 @@ managed:
   go_package_prefix:
     default: github.com/rilldata/rill
 plugins:
-  - plugin: buf.build/bufbuild/es
+  - plugin: buf.build/bufbuild/es:v1.10.0
```
esevastyanov, line 7: This change should not be passed as you've already passed generated proto sources.

proto/buf.gen.yaml
```diff
@@ -7,7 +7,7 @@ managed:
       - buf.build/googleapis/googleapis
       - buf.build/envoyproxy/protoc-gen-validate
 plugins:
-  - plugin: buf.build/protocolbuffers/go
+  - plugin: buf.build/protocolbuffers/go:v1.34.2
```
esevastyanov, line 10: This change should not be passed too

esevastyanov (changes requested):

proto/buf.gen.ui.yaml
AdityaHegde: Anyone else generating will face the same issue if we dont pin the version.

nishantmonu51 (approved):
It's ok to pin versions for now. 
As a separate effort, we need to look into how would we upgrade proto versions in future without hitting the same issues again. I believe there should be some ways of adding library versions in the serialized proto objects, that the newer version can check to deserialize. 
@AdityaHegde : Can you help create an issue for tracking the upgrade effort ?

esevastyanov (approved):

proto/rill/runtime/v1/resources.proto
```diff
@@ -365,6 +365,9 @@ message AlertSpec {
   int32 intervals_limit = 6;
   bool intervals_check_unclosed = 7;
   uint32 timeout_seconds = 8;
+  // for alerts that have not been edited since resolver and resolver_properties have been added
```
begelundmuller, line 368: When the runtime restarts, it should parse all YAML files and update any resources that output a different proto object. So I don't think this should be necessary? Is the issue perhaps in some other place?

begelundmuller: Here's the code that triggers a comparison of all resources when the parser first runs: https://github.com/rilldata/rill/blob/main/runtime/reconcilers/project_parser.go#L147

And here's the code that does the comparison for an alert object specifically: https://github.com/rilldata/rill/blob/main/runtime/reconcilers/project_parser.go#L520

AdityaHegde: Good point. There were issues with stage runtimes not coming up. The root cause could have been that. (This PR was raised before it was identified)

PR #5372, opened 2024-08-01
fix: onboarding screen being shown again even after creating a project

AdityaHegde:
We seem to have a fetch to files as a `queryClient.fetchQuery`. If this fails for any reason we redirect the user back to onboarding screen. In some cases the query is being cancelled which also triggers this flow. But adding a `retry` doesnt seem to be honoured when the query is cancelled.

Wrapping the fetch with a retry to fix this.


ericpgreen2:
I think we should address the root problem – the query cancelation – rather than brute force it with a retry. The determination of whether or not the project is initialized is too important to have a bug lingering beneath the surface.

AdityaHegde:
Closing this to take a look at the root cause

PR #5373, opened 2024-08-01
Runtime : Implement security policy filters in MetricsViewSearch

k-anshul:
closes https://github.com/rilldata/rill-private-issues/issues/506

The refactoring from `metricsview.AST` to a generic AST can be done in a followup PRs.


runtime/drivers/druid/native_query.go
```diff
@@ -61,7 +71,9 @@ type QueryPlan struct {
 	} `json:"query"`
 }
 
-func (n *NativeQuery) Do(ctx context.Context, dr, res interface{}, queryID string) error {
+func (n *NativeQuery) Do(ctx context.Context, dr, res interface{}, queryID string, logger *zap.Logger) error {
+	logger = logger.With(zap.String("query_id", queryID))
+	logger.Debug("Executing native query", zap.Any("request", dr))
```
begelundmuller, line 76: It should probably only do these logs if `connector.druid.log_queries` is `true`

runtime/metricsview/executor.go
```diff
@@ -320,3 +323,91 @@ func (e *Executor) Export(ctx context.Context, qry *Query, executionTime *time.T
 		"args": args,
 	})
 }
+
+// SearchQuery executes the provided query against the metrics view.
+func (e *Executor) SearchQuery(ctx context.Context, searchQry *SearchQuery, executionTime *time.Time) ([]SearchResult, error) {
```
begelundmuller, line 328:
```suggestion
// SearchQuery executes the provided query against the metrics view.
func (e *Executor) Search(ctx context.Context, qry *SearchQuery, executionTime *time.Time) ([]SearchResult, error) {
```

runtime/metricsview/executor.go
```diff
@@ -320,3 +323,91 @@ func (e *Executor) Export(ctx context.Context, qry *Query, executionTime *time.T
...
+	if !e.security.CanAccess() {
+		return nil, runtime.ErrForbidden
+	}
+	logger := e.rt.Logger.With(zap.String("instance_id", e.instanceID))
```
begelundmuller, line 332: Can use `e.rt.InstanceLogger(ctx, e.instanceID)`

runtime/metricsview/executor.go
```diff
@@ -320,3 +323,91 @@ func (e *Executor) Export(ctx context.Context, qry *Query, executionTime *time.T
...
+		if err == nil || !errors.Is(err, errDruidNativeSearchUnimplemented) {
+			return res, err
+		}
+	}
+
+	sql, args, err := searchSQL(ast, searchQry.Search)
+	if err != nil {
+		return nil, err
+	}
+	logger.Debug("executing search query", zap.String("sql", sql), zap.String("metrics_view", searchQry.MetricsView), zap.String("instance_id", e.instanceID), observability.ZapCtx(ctx))
```
begelundmuller, line 379: I would recommend removing this and using request-level logs and traces for search observability, and DB-level logs/traces for query logging

runtime/metricsview/executor.go
```diff
@@ -320,3 +323,91 @@ func (e *Executor) Export(ctx context.Context, qry *Query, executionTime *time.T
...
+			return nil, err
+		}
+		dim, ok := row["dimension"].(string)
+		if !ok {
+			return nil, fmt.Errorf("unknown result dimension: %q", dim)
+		}
+		searchResult = append(searchResult, SearchResult{
+			Dimension: dim,
+			Value:     row["value"],
+		})
```
begelundmuller, line 407: Can't it just scan into the search result? Something like:
```go
var row SearchResult
err := res.Scan(&row.Dimension, &row.Value)
...
```

runtime/metricsview/search_query.go
```diff
@@ -0,0 +1,292 @@
...
+	if a.Root.FromSelect != nil {
+		// This means either the dimension uses an unnest or measure filters which are not directly supported by native search.
+		// This can be supported in future using query datasource in future if performance turns out to be a concern.
+		return nil, errDruidNativeSearchUnimplemented
+	}
+	var query map[string]interface{}
+	if a.Root.Where != nil {
+		// NOTE :: this does not work for measure filters.
+		// The query planner resolves them to joins instead of filters.
+		rows, err := olap.Execute(ctx, &drivers.Statement{
```
begelundmuller, line 43: The `rows` are not closed

runtime/metricsview/search_query.go
```diff
@@ -0,0 +1,292 @@
...
+	}
+	req := druid.NewNativeSearchQueryRequest(trimQuotes(*a.Root.FromTable), q.Search, dims, virtualCols, limit, q.TimeRange.Start, q.TimeRange.End, query) // TODO: timestamps may be nil!
+
+	// Execute the native query
+	var res druid.NativeSearchQueryResponse
+	nq := druid.NewNativeQuery(druidSQLDSN.ReplaceAllString(dsn, "/v2/"))
+	err = nq.Do(ctx, req, &res, req.Context.QueryID, logger)
+	if err != nil {
+		return nil, err
+	}
```
begelundmuller, line 126: I think this should be pushed into the Druid driver, so we don't have DSN and similar parsing logic here. Could use a similar pattern like the DuckDB appender, for example:
```go
h := druid.NewNativeClientFromHandle(olap.(drivers.Handle))
res, err := h.Search(a.Root.FromTable, q.Search, ...)
```

runtime/metricsview/search_query.go
```diff
@@ -0,0 +1,292 @@
...
+				Type:       "expression",
+				Name:       fmt.Sprintf("%v_virtual_native", f.Name), // The name of the virtual column should not clash with actual column
+				Expression: dim.Expression,
+			})
+			dims = append(dims, fmt.Sprintf("%v_virtual_native", f.Name))
+		} else {
+			dims = append(dims, trimQuotes(f.Expr))
+		}
+	}
+	req := druid.NewNativeSearchQueryRequest(trimQuotes(*a.Root.FromTable), q.Search, dims, virtualCols, limit, q.TimeRange.Start, q.TimeRange.End, query) // TODO: timestamps may be nil!
```
begelundmuller, line 118: Maybe use `e.metricsView.Table` instead of `trimQuotes...`?

runtime/metricsview/search_query.go
```diff
@@ -0,0 +1,292 @@
...
+	err := b.writeSelect(ast.Root)
+	if err != nil {
+		return "", nil, err
+	}
+
+	return b.out.String(), b.args, nil
+}
+
+func (b *searchSQLBuilder) writeSelect(n *SelectNode) error {
+	for i, f := range n.DimFields {
```
begelundmuller, line 164: Looking at this, I wonder if it could be simplified by creating a `Query` and calling `NewAST` and `.SQL()` separately for *each* dimension and combining the outputs with `UNION ALL`?

That might help avoid needing to access the AST internals here (and also remove some duplicated SQL output logic).

begelundmuller (changes requested):

runtime/metricsview/search_query.go
k-anshul: The `metricsviewSpec` is not available here and seems wasteful to pass it till here in `SearchQuery`

runtime/metricsview/search_query.go
k-anshul: Actually that's what I had thought initially but the current AST sql lacks two features:
1. It can't add an alias to a plain dimension select.
2. It can't select constants
These are required to generate following query:
```
SELECT 
  column1 AS Value,
  'dimension_name1' AS dimension
FROM model
GROUP BY Value
WHERE ....
UNION ALL 
SELECT 
  column2 AS Value,
  'dimension_name2' AS dimension
FROM model
WHERE ...
GROUP BY Value
...
```

Now we can circumnavigate this limitation by querying each dimension separately as well but that will lead to one query per dimension.

We can also add the missing functionality to the AST.
Please let me know what do you think.

begelundmuller: Got it, I can think of two ways to work around this:
1. Wrap the generated SQL for each dimension with `SELECT 'dim name 1' as dim, dim1 as val FROM (<generated SQL>)`
2. After calling `NewAST`, you can manually override the ASTs fields like this:
```go
ast.Root.DimFields = append(ast.Root.DimFields, FieldNode{
  Name: "dimension",
  Expr: ast.Root.DimFields[0].Name,
})
ast.Root.DimFields[0].Name = "value"
```

runtime/metricsview/executor.go
```diff
@@ -320,3 +322,150 @@ func (e *Executor) Export(ctx context.Context, qry *Query, executionTime *time.T
...
+			dimensions[i] = Dimension{Name: d}
+		}
+		q := &Query{
+			MetricsView: qry.MetricsView,
+			Dimensions:  dimensions,
+			Where:       qry.Where,
+			Having:      qry.Having,
+			TimeRange:   qry.TimeRange,
+			Limit:       qry.Limit,
+		}
```
begelundmuller, line 348: nit: Can you add all fields in the struct, even the `nil` ones? Since it would be quite bad if we add a new field and don't add it everywhere, doing this will reduce the chance of errors. Btw, I'm looking into setting up this package to force setting struct members for specific structs: https://github.com/GaijinEntertainment/go-exhaustruct

runtime/metricsview/executor.go
```diff
@@ -320,3 +322,150 @@ func (e *Executor) Export(ctx context.Context, qry *Query, executionTime *time.T
...
+
+		if err := e.rewriteLimitsIntoSubqueries(ast); err != nil {
+			return nil, err
+		}
+
+		// native search
+		res, err := qry.executeSearchInDruid(ctx, e.olap, ast)
+		if err == nil || !errors.Is(err, errDruidNativeSearchUnimplemented) {
+			return res, err
+		}
```
begelundmuller, line 367: Since this function does execution, I think it might make more sense to put it on the `Executor` type – then you could even move the AST stuff above into the function, and just have something like `e.trySearchInDruid(ctx, qry)` (and you wouldn't need the hack to unquote the table name)

runtime/metricsview/executor.go
```diff
@@ -320,3 +322,150 @@ func (e *Executor) Export(ctx context.Context, qry *Query, executionTime *time.T
...
+		if i > 0 {
+			finalSQL.WriteString(" UNION ALL ")
+		}
+		q := &Query{
+			MetricsView: qry.MetricsView,
+			Dimensions:  []Dimension{{Name: d}},
+			Having:      qry.Having,
+			TimeRange:   qry.TimeRange,
+			Limit:       qry.Limit,
+		}
```
begelundmuller, line 386: Same note about specifying all struct members

runtime/metricsview/executor.go
```diff
@@ -320,3 +322,150 @@ func (e *Executor) Export(ctx context.Context, qry *Query, executionTime *time.T
...
+
+		if err := e.rewriteLimitsIntoSubqueries(ast); err != nil {
+			return nil, err
+		}
+
+		sql, args, err := ast.SQL()
+		if err != nil {
+			return nil, err
+		}
+		finalSQL.WriteString(fmt.Sprintf("SELECT '%s' AS dimension, %s AS value FROM (%s)", d, e.olap.Dialect().EscapeIdentifier(d), sql))
```
begelundmuller, line 411: What if `d` contains a single quote? I think the dialect type might have a func for escaping a string, else we may need to add it (or put it in the args if possible to do at the right place)

begelundmuller (changes requested):

runtime/metricsview/executor.go
k-anshul: Seems somewhat painful to specify nil fields if there are more than few. 
But could be an easy solution to make sure fields are not missed.

runtime/metricsview/executor.go
```diff
@@ -328,3 +332,288 @@ func (e *Executor) Export(ctx context.Context, qry *Query, executionTime *time.T
...
+
+		ast, err := NewAST(e.metricsView, e.security, q, e.olap.Dialect())
+		if err != nil {
+			return nil, err
+		}
+
+		if err := e.rewriteLimitsIntoSubqueries(ast); err != nil {
+			return nil, err
+		}
+
```
begelundmuller, line 395: This could be moved into `executeSearchInDruid`

runtime/metricsview/executor.go
```diff
@@ -328,3 +332,288 @@ func (e *Executor) Export(ctx context.Context, qry *Query, executionTime *time.T
...
+		}
+		searchResult = append(searchResult, row)
+	}
+	if res.Err() != nil {
+		return nil, res.Err()
+	}
+	return searchResult, nil
+}
+
+func (e *Executor) executeSearchInDruid(ctx context.Context, olap drivers.OLAPStore, a *AST, search string) ([]SearchResult, error) {
```
begelundmuller, line 484: `olap` is a redundant arg, can just access `e.olap`

runtime/metricsview/executor.go
```diff
@@ -328,3 +332,288 @@ func (e *Executor) Export(ctx context.Context, qry *Query, executionTime *time.T
...
+	if a.Root.Where != nil {
+		// NOTE :: this does not work for measure filters.
+		// The query planner resolves them to joins instead of filters.
+		rows, err := olap.Execute(ctx, &drivers.Statement{
+			Query:            fmt.Sprintf("EXPLAIN PLAN FOR SELECT 1 FROM %s WHERE %s", *a.Root.FromTable, a.Root.Where.Expr),
+			Args:             a.Root.Where.Args,
+			DryRun:           false,
+			Priority:         0,
+			LongRunning:      false,
+			ExecutionTimeout: 0,
```
begelundmuller, line 500: Should these use the same args as above (in the `Search` implementation)? Otherwise, it's fine to omit them – my suggestion around including empty fields in `Query` only applies to the few structs where it's critical to think about all fields (similar to the project insert/update structs); in normal cases it's totally fine to omit empty fields

runtime/metricsview/executor.go
```diff
@@ -328,3 +332,288 @@ func (e *Executor) Export(ctx context.Context, qry *Query, executionTime *time.T
...
+				Type:       "expression",
+				Name:       fmt.Sprintf("%v_virtual_native", f.Name), // The name of the virtual column should not clash with actual column
+				Expression: dim.Expression,
+			})
+			dims = append(dims, fmt.Sprintf("%v_virtual_native", f.Name))
+		} else {
+			dims = append(dims, f.Name)
+		}
+	}
+	req := druid.NewNativeSearchQueryRequest(e.metricsView.Table, search, dims, virtualCols, limit, a.query.TimeRange.Start, a.query.TimeRange.End, query) // TODO: timestamps may be nil!
```
begelundmuller, line 563: Should this TODO be handled by returning `errDruidNativeSearchUnimplemented` if they are empty?

runtime/queries/metricsview_search.go
```diff
@@ -68,27 +64,10 @@ func (q *MetricsViewSearch) Resolve(ctx context.Context, rt *runtime.Runtime, in
...
-
-	if mv.TimeDimension == "" && !isTimeRangeNil(q.TimeRange) {
-		return fmt.Errorf("metrics view '%s' does not have a time dimension", mv)
-	}
-
 	if !isTimeRangeNil(q.TimeRange) {
+		if mv.TimeDimension == "" {
+			return fmt.Errorf("metrics view '%s' does not have a time dimension", mv)
+		}
 		start, end, err := ResolveTimeRange(q.TimeRange, mv)
```
begelundmuller, line 71: Can this be removed now? I think the `rewriteQueryTimeRanges` rewriter should resolve time ranges

begelundmuller (approved):

runtime/metricsview/executor.go
begelundmuller: Not thinking about adding it for all structs, that would be too painful, only for critical ones (like project update) – it has `//exhaustruct:enforce` to selectively add to specific structs

runtime/metricsview/executor.go
k-anshul: Sure. I just refactored this code and didn't realise that some of these args can be skipped.

runtime/metricsview/executor.go
k-anshul: I am not sure. I kept the existing todo as it is. I think this is not possible right now since UI sends time range in all cases. But yeah returning an error seems to be a good idea.

PR #5377, opened 2024-08-02
fix: fix flaky metrics view sql test

k-anshul:


pjain1 (approved):
why was it flake ?

k-anshul:
So there were two publishers `NULL` and `Microsoft` and I am assuming that in absence of an order by clause it was indeterministic which one is the first row.

PR #5390, opened 2024-08-05
Runtime: metrics view refer model connector

### LINKED ISSUE #5085: Runtime: Use connector of underlying model for metrics views

k-anshul:
closes #5085


begelundmuller (changes requested):
> The change works but conceptually speaking I am not sure of it.
> We are changing the spec during reconcile which is something we haven't done before.
> Also considered storing this in mv's state but technically speaking the model connector is a spec attribute.

Yeah editing the current resource's spec is something we want to be careful with. In this particular case, I think one of these options would be preferable:

1. (Simple and mostly correct option) Apply this PR to the `mv.State.ValidSpec` instead. It's used everywhere to actually render/query the metrics view.
2. (Complex but maybe more correct option) If `model:` is set in the YAML, capture it as a separate property in the spec (e.g. `Spec.Model` instead of `Spec.Table`). Then at query time, if the metrics view is based on a model, look it up to find the underlying connector+table to query.

k-anshul:
>(Simple and mostly correct option) Apply this PR to the mv.State.ValidSpec instead. It's used everywhere to actually render/query the metrics view.

So `mv.Spec` is what is being set to `mv.State.ValidSpec` so setting one or other should be equivalent?

>(Complex but maybe more correct option) If model: is set in the YAML, capture it as a separate property in the spec (e.g. Spec.Model instead of Spec.Table). Then at query time, if the metrics view is based on a model, look it up to find the underlying connector+table to query.

I thought about it but not sure if we should worry about the underlying model being changed and metricsview not reconciled yet leading to inconsistencies.

begelundmuller:
> So mv.Spec is what is being set to mv.State.ValidSpec so setting one or other should be equivalent?

The `ValidSpec` is only updated if the `Spec` passes validation, otherwise it stays as the previous correct spec (only when `StageChanges` is true). In general, the idea is that a resource's "spec" is its *desired* state, and the "state" represents its *actual* state. So a creator (usually the YAML parser) will emit a spec, then the reconciler will update the state based on the spec, and then consumers should *only* consume from the state.

So I think for this PR we could extend this idea and say that the `ValidSpec` is not just a validated version of the spec, but a validated *and enriched* version of the spec.

k-anshul:
> The `ValidSpec` is only updated if the `Spec` passes validation, otherwise it stays as the previous correct spec (only when `StageChanges` is true). In general, the idea is that a resource's "spec" is its _desired_ state, and the "state" represents its _actual_ state. So a creator (usually the YAML parser) will emit a spec, then the reconciler will update the state based on the spec, and then consumers should _only_ consume from the state.
> 
> So I think for this PR we could extend this idea and say that the `ValidSpec` is not just a validated version of the spec, but a validated _and enriched_ version of the spec.

Sure. Sounds good.

runtime/reconcilers/metrics_view.go
```diff
@@ -65,14 +65,22 @@ func (r *MetricsViewReconciler) Reconcile(ctx context.Context, n *runtimev1.Reso
 		return runtime.ReconcileResult{}
 	}
 
+	var modelConnector string
+	res, err := r.C.Get(ctx, &runtimev1.ResourceName{Name: mv.Spec.Table, Kind: runtime.ResourceKindModel}, false)
+	if err == nil && res.GetModel().State.ResultConnector != "" { // ignore error since dashboard can directly reference a table as well
+		modelConnector = res.GetModel().State.ResultConnector
+	} else {
+		modelConnector = mv.Spec.Connector
+	}
```
begelundmuller, line 74: It should only try to infer the connector if `mv.Spec.Connector == ""` (i.e. not explicitly set)

runtime/reconcilers/metrics_view.go
```diff
@@ -65,14 +65,22 @@ func (r *MetricsViewReconciler) Reconcile(ctx context.Context, n *runtimev1.Reso
...
+
 	// NOTE: In other reconcilers, state like spec_hash and refreshed_on is used to avoid redundant reconciles.
 	// We don't do that here because none of the operations below are particularly expensive.
 	// So it doesn't really matter if they run a bit more often than necessary ¯\_(ツ)_/¯.
 
 	// NOTE: Not checking refs for errors since they may still be valid even if they have errors. Instead, we just validate the metrics view against the table name.
 
 	// Validate the metrics view and update ValidSpec
-	validateResult, validateErr := r.C.Runtime.ValidateMetricsView(ctx, r.C.InstanceID, mv.Spec)
+	validateResult, validateErr := r.C.Runtime.ValidateMetricsView(ctx, r.C.InstanceID, mv.Spec, modelConnector)
```
begelundmuller, line 83: There's already a `Connector` property on the spec, so better to update the spec before passing it to `ValidateMetricsView`. I.e. something like:
```go
// Enrich spec before validating it
spec := mv.Spec
if spec.Connector == "" {
  // Attempt to infer spec from a model
}

// ...
validateResult, validateErr := r.C.Runtime.ValidateMetricsView(ctx, r.C.InstanceID, spec)

**Review by begelundmuller — CHANGES_REQUESTED** [review]

File: runtime/reconcilers/metrics_view.go (hunk 2 above)
Comment (k-anshul) [reply runtime/reconcilers/metrics_view.go:83]: Isn't this equivalent to the solution that we had initially ? :)

File: runtime/reconcilers/metrics_view.go (hunk 1 above)
Comment (k-anshul) [reply runtime/reconcilers/metrics_view.go:74]: The `mv.Spec.Connector` is never unset : https://github.com/rilldata/rill/blob/fcd9e75b450d01cc930f2e95c669c23365c91a7d/runtime/compilers/rillv1/parse_node.go#L295-L296

Comment (k-anshul) [reply runtime/reconcilers/metrics_view.go:74]: Okay I think we can leverage `ConnectorInferred` to only set it if user provides it. 
For metrics_view referring to models it will be set by newly introduced logic.
For external tables it anyways fallbacks to default OLAP connector when acquiring OLAP connection.
What do you think ?

Comment (begelundmuller) [reply runtime/reconcilers/metrics_view.go]: Yeah that sounds good

File: runtime/compilers/rillv1/parse_metrics_view.go [hunk 3]
```diff
@@ -555,15 +555,18 @@ func (p *Parser) parseMetricsView(node *Node) error {
...
-	if table == "" {
+	var (
+		table string
+		model string
+	)
+	if tmp.Table != "" {
+		table = tmp.Table
+	} else if tmp.Model != "" {
+		model = tmp.Model
+	} else {
```
begelundmuller, line 569:
```suggestion
	if tmp.Table == "" && tmp.Model == "" {
```

runtime/compilers/rillv1/parse_metrics_view.go
```diff
@@ -815,7 +818,11 @@ func (p *Parser) parseMetricsView(node *Node) error {
 		return err
 	}
 
-	node.Refs = append(node.Refs, ResourceName{Name: table})
+	if table != "" {
+		node.Refs = append(node.Refs, ResourceName{Name: table})
+	} else {
+		node.Refs = append(node.Refs, ResourceName{Name: model, Kind: ResourceKindModel})
+	}
```
begelundmuller, line 825:
```suggestion
	if tmp.Model != "" {
		// Not setting Kind because for backwards compatibility, it may actually be a source or an external table.
		node.Refs = append(node.Refs, ResourceName{Name: tmp.Model})
	}
```

runtime/compilers/rillv1/parse_metrics_view.go
```diff
@@ -831,6 +838,7 @@ func (p *Parser) parseMetricsView(node *Node) error {
 	spec.Database = tmp.Database
 	spec.DatabaseSchema = tmp.DatabaseSchema
 	spec.Table = table
+	spec.Model = model
```
begelundmuller, line 841:
```suggestion
	spec.Table = tmp.Table
	spec.Model = tmp.Model
```

runtime/reconcilers/metrics_view.go
```diff
@@ -65,14 +65,32 @@ func (r *MetricsViewReconciler) Reconcile(ctx context.Context, n *runtimev1.Reso
...
+			validateErr = fmt.Errorf("failed to get model %q: %w", mv.Spec.Model, err)
+		} else {
+			if res.GetModel().State.ResultTable != "" {
+				mv.Spec.Table = res.GetModel().State.ResultTable
+				mv.Spec.Connector = res.GetModel().State.ResultConnector
+			} else {
+				mv.Spec.Table = mv.Spec.Model
+			}
+		}
+	}
```
begelundmuller, line 81:
```suggestion
	// If the spec references a model, try resolving it to a table before validating it.
	// For backwards compatibility, the model may actually be a source or external table. 
	// So if a model is not found, we optimistically use the model name as the table and proceed to validation
	if mv.Spec.Model != "" {
		res, err := r.C.Get(ctx, &runtimev1.ResourceName{Name: mv.Spec.Model, Kind: runtime.ResourceKindModel}, false)
		if err == nil && res.GetModel().State.ResultTable != "" {
			mv.Spec.Table = res.GetModel().State.ResultTable
			mv.Spec.Connector = res.GetModel().State.ResultConnector
		} else {
			mv.Spec.Table = mv.Spec.Model
		}
	}
```

runtime/reconcilers/metrics_view.go
```diff
@@ -65,14 +65,32 @@ func (r *MetricsViewReconciler) Reconcile(ctx context.Context, n *runtimev1.Reso
...
 	// So it doesn't really matter if they run a bit more often than necessary ¯\_(ツ)_/¯.
 
 	// NOTE: Not checking refs for errors since they may still be valid even if they have errors. Instead, we just validate the metrics view against the table name.
 
 	// Validate the metrics view and update ValidSpec
-	validateResult, validateErr := r.C.Runtime.ValidateMetricsView(ctx, r.C.InstanceID, mv.Spec)
+	var validateResult *runtime.ValidateMetricsViewResult
+	if validateErr == nil {
+		validateResult, validateErr = r.C.Runtime.ValidateMetricsView(ctx, r.C.InstanceID, mv.Spec)
+	}
```
begelundmuller, line 93:
```suggestion
	validateResult, validateErr := r.C.Runtime.ValidateMetricsView(ctx, r.C.InstanceID, mv.Spec)
```

begelundmuller (approved):

Issue #5375, opened 2024-08-01
Public URLs: Support encrypted secret in magic auth tokens


lovincyrus:
The public URLs we create are currently ephemeral. We can store the encrypted magic token in the database so that we can query and list them in the public URLs table.

This is blocking https://github.com/rilldata/rill/pull/5366

Related: https://github.com/rilldata/rill/pull/5071, https://github.com/rilldata/rill/pull/5175, https://github.com/rilldata/rill/pull/5163

#### NOW
- [ ] Add `EncryptedToken` or `Secret` in `magic_auth_tokens` table
- [ ] Store encrypted token upon `IssueMagicAuthToken`
- [ ] https://github.com/rilldata/rill/pull/5402#discussion_r1705353488

#### References

`magic_auth_tokens`
| encrypted_secret  |
| ------------- | 
| 1Vc74Wk790Ck0p27u8cJdGvts1eYivSWlZAeCCdRLuMEvFZWdFLJE2  |

Construct http://localhost:3000/{org}/{project}/-/share/{magic_token}

List magic tokens `/v1/organizations/${organization}/projects/${project}/tokens/magic`

```go
// admin.swagger.yaml
// /v1/organizations/{organization}/projects/{project}/tokens/magic
// lists all the magic auth tokens for a specific project.
ListMagicAuthTokens

// admin-service.ts
// IssueMagicAuthToken creates a "magic" auth token that provides limited access to a specific filtered dashboard in a specific project.
adminServiceListMagicAuthTokens

// admin-service.ts
getAdminServiceListMagicAuthTokensQueryKey

// auth_token.go
// IssueMagicAuthToken generates and persists a new magic auth token for a project.
IssueMagicAuthToken

// database.go
FindMagicAuthTokensWithUser(ctx context.Context, projectID string, createdByUserID *string, afterID string, limit int) ([]*MagicAuthTokenWithUser, error)
FindMagicAuthToken(ctx context.Context, id string) (*MagicAuthToken, error)
FindMagicAuthTokenWithUser(ctx context.Context, id string) (*MagicAuthTokenWithUser, error)
InsertMagicAuthToken(ctx context.Context, opts *InsertMagicAuthTokenOptions) (*MagicAuthToken, error)
UpdateMagicAuthTokenUsedOn(ctx context.Context, ids []string) error
DeleteMagicAuthToken(ctx context.Context, id string) error
DeleteExpiredMagicAuthTokens(ctx context.Context, retention time.Duration) error
```

[...]

PR #5402, opened 2024-08-05
Support encrypted secret for magic auth tokens

lovincyrus:
- [x] Add `EncryptedSecret` to `MagicAuthToken`
- [x] Add a new pg migration SQL `0038.sql`
- [x] Make sure `V1ListMagicAuthTokensResponse` has encrypted secret
- [ ] Run migration script (added new column)
- [ ] Test public link creation flow

Solves #5375 and unblocks #5366

Reference to https://www.figma.com/design/Qt6EyotCBS3V6O31jVhMQ7?node-id=15019-622855#888924088

Similar to https://github.com/rilldata/rill/pull/5339


admin/auth_token.go
```diff
@@ -187,9 +187,17 @@ func (s *Service) IssueMagicAuthToken(ctx context.Context, opts *IssueMagicAuthT
 		expiresOn = &t
 	}
 
+	// Encrypt the secret
+	// TODO: revisit this
+	encryptedSecret, err := tkn.EncryptSecret([]byte(tkn.ID.String()))
```
begelundmuller, line 192: Good job getting started on this issue! About column-level encryption, there's a few things to think through. 

We also need column-level encryption for credentials, which is something we have been putting off, but have committed to building soon. So ideally we take an approach that we can also leverage there (to keep only one set of encryption keys for column-level encryption).

For the encryption keys, we would need those to be externally configured (e.g. with an environment variable or key file) and we need to support key rotation for them, which may entail persisting a key ID alongside the encrypted value.

There is also a question of where in the stack we handle this – one option is where the value is used, as you have done in this PR, but another is to handle it when values are read/written to the DB in the `admin/database/postgres` package, or using something like the `pgcrypto` Postgres extension and handling it in the DB itself.

lovincyrus: Sgtm! Closing this PR for now.

PR #5403, opened 2024-08-06
log num rows downloaded to tmp file from bq

pjain1:
We have TotalRows field on iterator but also see that in some cases it may be 0 until next is called so doing this so that we log actual rows written.


runtime/drivers/bigquery/warehouse.go
```diff
@@ -263,6 +264,7 @@ func (f *fileIterator) Next() ([]string, error) {
 			if err := writer.WriteBuffered(rec); err != nil {
 				return nil, err
 			}
+			rows++
```
k-anshul, line 267: One arrow record can have multiple rows as well.

pjain1: ok so we can use `rec.NumRows()` to add to `rows`, should we do this or just use `TotalRecords` field on `FileIterator` ?

pjain1: I think lets calculate rows ourselves so that if we miss something due to any bug we can see that

k-anshul: Sure sounds good.
Lets log the number of rows from Bigquery(TotalRecords) as well as based on row group records(rec.NumRows()) since there could be a bug in parquet generation as well. 
This will rule out near impossible case that `BigQuery` itself is returning less rows.

k-anshul: We can also make this and the other log `Info` since I don't think we capture debug logs.

pjain1:
I also wonder why we don't file size check when downloading as json, is it a miss ?

runtime/drivers/bigquery/warehouse.go
```diff
@@ -242,19 +242,34 @@ func (f *fileIterator) Next() ([]string, error) {
...
 	ticker := time.NewTicker(time.Second * 5)
 	defer ticker.Stop()
+
+	limitExceeded := make(chan struct{})
+	defer close(limitExceeded)
+
+	go func() {
+		for {
+			select {
+			case <-limitExceeded:
```
k-anshul, line 252: I think this is an issue that same channel is received by this goroutine as well as the main goroutine. One of these will not receive it.
It is also not idiomatic to send and receive on a channel from the same goroutine. 
Also I think in cases when limits are not exceeded(or main goroutine receives the value) this goroutine will leak waiting on both channels (unless `ticker.Stop` closes the channel).
I think we can override the parent context with a cancellable context and cancel the context on limit exceeds. This goroutine can listen on `ctx.Done()` and exit.

runtime/drivers/bigquery/warehouse.go
```diff
@@ -310,7 +330,7 @@ func (f *fileIterator) downloadAsJSONFile() (string, error) {
 	f.downloaded = true
 
 	init := false
-	rows := 0
+	rows := int64(0)
```
k-anshul, line 333: I think this is not being incremented.

k-anshul (changes requested):

runtime/drivers/bigquery/warehouse.go
k-anshul: Okay I see it is incremented in the existing code itself.

k-anshul:
> I also wonder why we don't file size check when downloading as json, is it a miss ?

We don't allow downloading large datasets in JSON format so that check is not there : https://github.com/rilldata/rill/blob/156c8d22a26aba81f691f830f0f873e80284122b/runtime/drivers/bigquery/warehouse.go#L368-L369

runtime/drivers/bigquery/warehouse.go
pjain1: thanks fixed.

runtime/drivers/bigquery/warehouse.go
```diff
@@ -271,11 +292,15 @@ func (f *fileIterator) Next() ([]string, error) {
...
+	if uint64(rows) < f.bqIter.TotalRows {
+		f.logger.Error("not all rows written to parquet file", zap.Int64("rows_written", rows), zap.Uint64("total_rows", f.bqIter.TotalRows), observability.ZapCtx(f.ctx))
+	}
+
 	fileInfo, err := os.Stat(fw.Name())
 	if err != nil {
 		return nil, err
 	}
-	f.logger.Debug("size of file", zap.String("size", datasize.ByteSize(fileInfo.Size()).HumanReadable()), observability.ZapCtx(f.ctx))
+	f.logger.Debug("parquet file written", zap.String("size", datasize.ByteSize(fileInfo.Size()).HumanReadable()), zap.Int64("rows", rows), observability.ZapCtx(f.ctx))
```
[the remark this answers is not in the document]
pjain1: don't feel strongly about it but if you think so lets change

PR #5404, opened 2024-08-06
fix: Block user from deploying when deployed project is connected to github

AdityaHegde:
Requirement reference: https://www.notion.so/rilldata/Redeployment-flow-updates-bb6e4d00c0e34b45aa6eb5368eda7e3e


k-anshul:
LGTM.
Did an experience pass on UI as well. 

For this popup 
image

I would suggest to skip showing the command and rewording.
a) This is not the only command which is required to push changes.
b) This will not work in cases when tracking branch is not set.

k-anshul (approved):

AdityaHegde:
Good point. I have brought it up with @ericokuma on slack. Will work with him to clean it up.

ericokuma:
Looks good on my end as well!

PR #5405, opened 2024-08-06
docs: release notes 0.48

AndrewRTsao:


AndrewRTsao:
@royendo Small nit but we should also be consistent in our punctuation (either end every line with a period or without one if they are all single-sentence phrases).

mindspank (approved):

PR #5412, opened 2024-08-07
sql expression filters for dimensions only

k-anshul:


runtime/pkg/metricssql/filter_parser.go
```diff
@@ -0,0 +1,350 @@
...
+		expr, err := parseValueExpr(node.Args[1]) // handling of x
+		if err != nil {
+			return nil, err
+		}
+		amt, err := strconv.Atoi(expr)
+		if err != nil {
+			return nil, fmt.Errorf("metrics sql: expected integer value in date_add/date_sub function")
+		}
+
+		// TODO :: check this
```
begelundmuller, line 329: Does it need to be handled before merging?

runtime/pkg/metricssql/filter_parser_test.go
```diff
@@ -0,0 +1,116 @@
...
+					Expressions: []*metricsview.Expression{
+						{
+							Name: "dim",
+						},
+						{
+							Value: "helllo",
+						},
+						{
+							Value: "world",
+						},
```
begelundmuller, line 52: Since `IN` is a binary operator, can we change this to output two sub-expressions where the second one is a slice of values instead? I.e. `[{"name": "dim"}, {"value": ["helllo", "world"]}]`? 

I know the flattened approach is also supported (because the UI does it), but that's just for backwards compatibility (and I hope we can eventually deprecate it entirely): https://github.com/rilldata/rill/blob/d66a28ff94b55273d102cc2112faf8d05465d254/runtime/metricsview/astexpr.go#L157-L175

runtime/pkg/metricssql/time_range_parser.go
```diff
@@ -67,7 +67,43 @@ func (q *query) parseTimeRangeEnd(ctx context.Context, node *ast.FuncCallExpr) (
 	}, nil
 }
 
-func (q *query) parseTimeRangeArgs(args []ast.ExprNode) (duration.Duration, int, string, error) {
+func (q *query) getWatermark(ctx context.Context, colName string) (watermark time.Time, err error) {
```
begelundmuller, line 70: Would it be possible to leverage the `Watermark` function from the `metricsview` package? 

https://github.com/rilldata/rill/blob/d66a28ff94b55273d102cc2112faf8d05465d254/runtime/metricsview/executor.go#L82-L86

runtime/queries/metricsview_aggregation.go
```diff
@@ -317,3 +323,16 @@ func (q *MetricsViewAggregation) rewriteToMetricsViewQuery(export bool) (*metric
...
+func metricViewExpression(expr *runtimev1.Expression, sql string) (*metricsview.Expression, error) {
+	if expr != nil && sql != "" {
+		return nil, fmt.Errorf("both expression and sql are set")
+	}
+	if expr != nil {
+		return metricsview.NewExpressionFromProto(expr), nil
+	}
+	if sql != "" {
+		return metricssqlparser.ParseSQLFilter(sql)
+	}
```
begelundmuller, line 336: For consistency, can we also add a `metricsview.NewExpressionFromSQL` function that proxies to the `metricssqlparser` internally?

begelundmuller (approved):

runtime/pkg/metricssql/filter_parser.go
k-anshul: I had actually checked this. Forgot to remove todo :)

runtime/pkg/metricssql/time_range_parser.go
k-anshul: Actually in the current time range functions we also allow user to provide an optional column whereas in the metricview only time dimension can be used. 
We can restrict the time range functions as well to work only with `timedimension` and then we can leverage the pkg function itself.

runtime/queries/metricsview_aggregation.go
k-anshul: This is leading to an import cycle since `metricssqlparser` imports and generates a `metricsview.Query`.

PR #5416, opened 2024-08-08
feat: Disconnect project from github

AdityaHegde:
- [x] Add API to move a project from github to uploaded asset.
- [x] Add delete icon and a confirmation before deleting the connection. 
- [x] Update dialog based on mocks.


runtime/pkg/archive/archive.go
```diff
@@ -59,6 +66,72 @@ func Download(ctx context.Context, downloadURL, downloadDst, projPath string, cl
...
+		fullPath := filepath.Join(root, entry.Path)
+		info, err := os.Lstat(fullPath)
+		if err != nil {
+			return fmt.Errorf("%s: %w", fullPath, err)
+		}
+		if info.Mode()&os.ModeSymlink != 0 {
+			return fmt.Errorf("%s: repo contains symlinks", entry.Path)
+		}
+
+		fmt.Println("createTar", entry.Path)
```
k-anshul, line 108: Can be removed ?

admin/server/assets.go
```diff
@@ -96,3 +196,47 @@ func (s *Server) assetPath(object string) (string, error) {
 	uploadPath.Path = object
 	return uploadPath.String(), nil
 }
+
+func gitToFilesList(gitPath, repo, branch, subpath, token string) ([]drivers.DirEntry, error) {
```
k-anshul, line 200: Did you consider directly using the download as tar API : https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28#download-a-repository-archive-tar

k-anshul (changes requested):

admin/server/assets.go
AdityaHegde: It doesnt seem to support subpath.

k-anshul (approved):
Approving from backend POV

k-anshul (approved):
Approving from backend POV

ericpgreen2:
I hit the following error. Can we also show API errors in the dialog in red error text?

image

web-admin/src/features/projects/github/github-errors.ts
```diff
@@ -26,3 +26,29 @@ export function extractGithubConnectError(err: AxiosError<RpcStatus>) {
     notEmpty: false,
   };
 }
+
+export function extractGithubDisconnectError(err: AxiosError<RpcStatus>) {
+  console.log(err);
```
ericpgreen2, line 31: Leftover?

web-admin/src/features/projects/github/ProjectGithubConnection.svelte
```diff
@@ -58,6 +65,11 @@
       },
     );
   }
+
+  function disconnectGithubConnect() {
+    void githubData.promptGithubLogin();
```
ericpgreen2, line 70: Based on the behavior, given the user isn't always shown a GitHub Login page, it feels like this function name should read something like "ensureLoggedInToGitHub".

ericpgreen2 (approved):
Here's UXQA, which admittedly spans a little more than the "disconnect" feature. Feel free to address in this PR or another one.

1. I ran into this state when adding a GitHub connection, selecting a non-empty repository, and specifying a subdirectory: 
    image
2. These two dialogs have different vertical orientation. Because the latter follows the former, it's a bit jarring.
    This one is higher on the page: image
    This one is lower on the page:  
    image
3. Extra trailing slash at the end of the repo name:
    image
4. When I click on "Yes, disconnect", the dialog's vertical height shrinks slightly. [See Jam](https://jam.dev/c/076fd334-bdbe-46a7-86af-edbfb1133502).

web-admin/src/features/projects/github/DisconnectProjectConfirmDialog.svelte
```diff
@@ -0,0 +1,90 @@
...
+        <div class="text-red-500 text-sm py-px">
+          {parsedError.message}
+        </div>
+      {/if}
+    </AlertDialogHeader>
+    <AlertDialogFooter>
+      <Button type="secondary" on:click={() => (open = false)}>Cancel</Button>
+      <Button type="primary" on:click={onDisconnect} loading={isLoading}>
+        Yes, disconnect
+      </Button>
```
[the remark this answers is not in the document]
AdityaHegde: Disabling just here for now. We should think about this for all dialogs I fell.

web-admin/src/features/projects/github/ProjectGithubConnection.svelte
AdityaHegde: `ensureGithubAccess` makes more sense since we are checking and using `grantAccessUrl`

AdityaHegde:
Thanks @ericpgreen2. Couldnt reproduce 1. 2 and 3 are unrelated, will raise separate PRs. Fixed 4

PR #5417, opened 2024-08-08
adjust comparison pill disabled styles

briangregoryholmes:
Before:
Screenshot 2024-08-08 at 12 24 59 PM
After:
Screenshot 2024-08-08 at 12 24 33 PM


ericokuma (approved):
UI/UX looks good!

Issue #5076, opened 2024-06-14
Runtime: Don't cache connection errors indefinitely


begelundmuller:

egor-ryashin:
Assuming (simplified), conncache has a map of http.Clients structs, those clients don't have a state - if there's a transient error then all users of http-clients will be retrying/failing, once the error is gone they all get successful results again. Besides we have timeouts for Druid requests here https://github.com/rilldata/rill/blob/34b8e4a960db6a678294ac98ea946c7346521a90/runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go#L128
 Could you elaborate?

begelundmuller:
If opening the connection fails, the returned error is cached here:

https://github.com/rilldata/rill/blob/c4ce5c35d3c0ac9c5004dc4c2464eba0ec0d1c48/runtime/pkg/conncache/conncache.go#L240

Next time something tries to open the connection, the cached error will be returned immediately:

https://github.com/rilldata/rill/blob/c4ce5c35d3c0ac9c5004dc4c2464eba0ec0d1c48/runtime/pkg/conncache/conncache.go#L166

The goal of this issue is to have some kind of TTL on the error caching before trying to connect again.

egor-ryashin:
I wonder if we have a state-machine diagram for conncache logic somewhere?

begelundmuller:
Unfortunately we don't. Feel free to add one in a docstring or `README.md` file in the `conncache` package directory.

begelundmuller:
I guess the previous issue we saw came from when Druid used the Avatica driver.

begelundmuller:
I still think this PR would be nice though, since it may be a problem for other connectors (like ClickHouse). We might also consider adding a ping when opening the Druid connector to be able to show credentials errors earlier.

PR #5418, opened 2024-08-08
conncache - skip caching errors

egor-ryashin:
Depends on this https://github.com/rilldata/rill/issues/5218


begelundmuller (changes requested):
Seems like this just skips caching errors entirely? Having a TTL as described in the issue might be preferable: https://github.com/rilldata/rill/issues/5076

Also, there are CI failures.

egor-ryashin:
I guess, you proposing something like this:
```go
	go func() {
		for true {
			time.Sleep(5 * time.Second)
			go func() {
				c.mu.Lock()
				defer c.mu.Unlock()
				for k, v := range c.entries {
					if v.err != nil && time.Now().Sub(v.previousErrTime) > 100*time.Second {
						c.beginClose(k, v)
					}
				}
			}()
		}
	}()
```

egor-ryashin:
CI fixed, need a design review.

runtime/pkg/conncache/conncache.go
```diff
@@ -62,6 +62,8 @@ type Options struct {
 	HangingFunc func(cfg any, open bool)
 	// Metrics are optional instruments for observability.
 	Metrics Metrics
+	// The maximum time the error stays in cache
+	ErrorTimeout time.Duration
```
begelundmuller, line 66: 1. It looks like this is not set anywhere?
2. Consider calling it `ErrorTTL` instead since it's not a timeout the same way that `OpenTimeout` and `CloseTimeout` are

runtime/pkg/conncache/conncache.go
```diff
@@ -136,6 +138,25 @@ func New(opts Options) Cache {
 		go c.periodicallyCheckHangingConnections()
 	}
 
+	go func() {
+		for {
+			time.Sleep(5 * time.Second)
+			go func() {
```
begelundmuller, line 144: This is missing cancellation from `Close`. Instead of having a background loop (which adds concurrency complexity), how about just checking the TTL when accessing a cached error in `Acquire`? That should still ensure correctness, and cleanup for connections that are not acquired again anyway happens automatically by the LRU cache

begelundmuller (changes requested):

runtime/pkg/conncache/conncache.go
egor-ryashin: Done.

runtime/pkg/conncache/conncache.go
egor-ryashin: Done.

egor-ryashin:
Tests passed.

PR #5422, opened 2024-08-09
S3 to Clickhouse model executor should default to `materialize: true`

k-anshul:


runtime/drivers/clickhouse/model_executor_s3_self.go
```diff
@@ -80,6 +80,22 @@ func (e *s3ToSelfExecutor) Execute(ctx context.Context, opts *drivers.ModelExecu
...
+	}
+	if out.Materialize == nil {
+		materialize := true
+		out.Materialize = &materialize
+	} else if !*out.Materialize {
+		return nil, fmt.Errorf("s3 to clickhouse executor only supports materialized models. Set `materialize` to true")
+	}
+	if err := mapstructure.WeakDecode(out, &newOpts.OutputProperties); err != nil {
+		return nil, fmt.Errorf("failed to parse output properties: %w", err)
+	}
```
begelundmuller, line 96: The DuckDB executors apply defaults when consuming the output properties (when the `Validate` func is called). We could do the same here? 

See: https://github.com/rilldata/rill/blob/d66a28ff94b55273d102cc2112faf8d05465d254/runtime/drivers/duckdb/model_executor_self.go#L42-L44

And: https://github.com/rilldata/rill/blob/d66a28ff94b55273d102cc2112faf8d05465d254/runtime/drivers/duckdb/model_manager.go#L31-L44

begelundmuller (changes requested):

runtime/drivers/clickhouse/model_executor_s3_self.go
k-anshul: So this validation is only specific to `s3_clickhouse` model and not to all.
Added a specific property that embeds and doing validation on that.

begelundmuller: What do you mean it's only specific to that executor? Seems to me like the default value for `materialize` should always be `true` when the output is a known physical destination (like S3 or incremental table), like in DuckDB?

begelundmuller: I don't understand why we can't apply the same defaults as DuckDB in all the cases? I'm not thinking about always setting it to `true`, I'm thinking about these defaults linked above (i.e. defaulting to true for incremental models, split models, and multi-connector models): https://github.com/rilldata/rill/blob/d66a28ff94b55273d102cc2112faf8d05465d254/runtime/drivers/duckdb/model_manager.go#L31-L44

k-anshul: Ahh okay. I understand now. Yeah I think this makes sense.

runtime/drivers/clickhouse/model_executor_s3_self.go
```diff
@@ -26,6 +26,21 @@ func (p *inputProps) Validate() error {
...
+func (p *outputProperties) Validate() error {
+	// set output props with defaults
+	if p.Materialize == nil {
+		materialize := true
+		p.Materialize = &materialize
+	} else if !*p.Materialize {
+		return fmt.Errorf("s3 to clickhouse executor only supports materialized models. Set `materialize` to true")
+	}
+	return nil
+}
```
begelundmuller, line 42: So now these changes aren't needed, right?

k-anshul: This is still needed.
In the model_executor it doesn't have context of s3. The input and output connector are both clickhouse in that case and it is not necessary that incremental and split are true so we need to make sure that materialize is true by default.
The `model_executor_s3_self` just calls `model_executor_self` with a SQL `SELECT * FROM S3(...`

begelundmuller: Okay I see – it's not very clear from the context, so it would be nice with a comment and maybe to combine it into one section of code, especially given the output properties aren't accessed or used for other purposes here. What do you think of my suggestion below?

runtime/drivers/clickhouse/model_executor_s3_self.go
```diff
@@ -75,11 +98,19 @@ func (e *s3ToSelfExecutor) Execute(ctx context.Context, opts *drivers.ModelExecu
...
 	clone := *opts
 	clone.InputProperties = propsMap
 	newOpts := &clone
 
+	var outputPropsMap map[string]any
+	err = mapstructure.WeakDecode(outputProps, &outputPropsMap)
+	if err != nil {
+		return nil, fmt.Errorf("failed to set output properties: %w", err)
+	}
+	newOpts.OutputProperties = outputPropsMap
```
begelundmuller, line 111:
```suggestion
	// Ensure materialize is true because the selfToSelfExecutor is not able to infer it independently.
	outputProps := &outputProperties{}
	err = mapstructure.WeakDecode(opts.OutputProperties, &outputProps)
	if err != nil {
		return nil, fmt.Errorf("failed to parse output properties: %w", err)
	}
	truth := true
	outputProps = &truth
	err = mapstructure.WeakDecode(outputProps, &newOpts.OutputProperties)
	if err != nil {
		return nil, fmt.Errorf("failed to parse output properties: %w", err)
	}
```

k-anshul: Sorry this is unclear to me. 
Do you mean we handle the validation(that materialize should not be false) in the `validate` function and handle setting the default here ?

In that case I think we can avoid the `validate` function as well and directly throw error here. 
We can also avoid create a separate `outputProperties` struct and directly parse it `ModelOutputProperties` ?
This is what I have in my mind : https://github.com/rilldata/rill/pull/5422/commits/43b0af324158b72a53cc81980bfeff21e221507e

k-anshul: Okay I need `outputProperties` for a separate change so we can keep it and its validation function.

begelundmuller (approved):

Issue #5242, opened 2024-07-10
Feat: Add Percent of Total as valid Alert criteria


mindspank:
If a measure is a valid percent of total it should be able to be used as a alert criteria.

AC:
- For a measure that is a valid percent of total a user should be able to select that as a criteria in the Criteria tab for alerts
- The input should accept percentages from 0 to 100
- Percent of total is independent of comparison and should always be able available if the measure is a valid percent of total
- The label in the dropdown displays aw "% of total"

PR #5429, opened 2024-08-12
feat: Add percent of total support in alerts

AdityaHegde:
closes [#5242](https://github.com/rilldata/rill/issues/5242)


web-common/src/features/alerts/criteria-tab/CriteriaForm.svelte
```diff
@@ -35,28 +36,36 @@
...
+          return {
+            ...o,
+            label: `${o.label} ${comparisonLabel}`,
+          };
+        })
+      : MeasureFilterBaseTypeOptions),
+    ...(selectedMeasure?.validPercentOfTotal
+      ? [MeasureFilterPercentOfTotalOption]
+      : []),
+  ];
```
ericpgreen2, line 66: This is pretty hairy. A function `makeTypeOptions(...)` could be warranted.

ericpgreen2 (commented):
UXQA –
1. Per Alex's requirements, we should validate that the input is within 0 and 100, otherwise show a validation error message.
    image
2. I don't think it makes sense to show "% of total" when the measure is not split by a dimension. It'll always be 100%, right? @mindspank, you say in the requirements "Percent of total is independent of comparison and should always be able available". Am I missing something?
    image

web-common/src/features/alerts/criteria-tab/parseCriteriaError.ts
```diff
@@ -6,5 +6,8 @@ export function parseCriteriaError(errStr: string, index: number): string {
   const match = criteriaParserRegex.exec(errStr);
   if (!match) return "";
   const [, matchedIndex, matchedErr] = match;
-  return Number(matchedIndex) === index ? matchedErr : "";
+  return Number(matchedIndex) === index
+    ? // `value1` in error is not user friendly. replacing with `criteria value`
```
ericpgreen2, line 10: Food for thought for later: I was confused myself when I saw "value1". I wonder if there's a way we can restructure the `MeasureFilterEntry` type to make it more self-explanatory.

ericpgreen2 (approved):

web-common/src/features/alerts/criteria-tab/parseCriteriaError.ts
AdityaHegde: Ya. When we redesign measure filters in dashboard we can get rid of `value1` and `value2`

PR #5436, opened 2024-08-13
Added note on --allowed-origins flag

royendo:
Added under get-started where we mention start a new rill project as a note,

added the flag to references/cli/start.md

Based on Issue: https://github.com/rilldata/rill/pull/5365
and Slack: https://rilldata.slack.com/archives/C01UFNH4W3B/p1723481208803969?thread_ts=1723461905.169519&cid=C01UFNH4W3B

### DIFF
--- a/docs/docs/home/get-started.md
+++ b/docs/docs/home/get-started.md
@@ -38,6 +38,11 @@ rill start my-rill-project
 
 The Rill web app runs locally at `http://localhost:9009` and will create code files in the `my-rill-project` directory.
 
+:::note
+Starting from v0.48, we have added the ability to self host the local Rill Developer application on a non-localhost endpoint. Please refer to the [rill start reference page](https://docs.rilldata.com/reference/cli/start) for more details on using the `--allowed-origins` flag. This method is not officially supported, please proceed at your own risk.
+:::
+
+
 ## Load and transform data
 
 On the welcome screen, initialize an example project or load up Rill with your own data (use local files, cloud storage and/or database connections)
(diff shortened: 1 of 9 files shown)


docs/docs/home/get-started.md
```diff
@@ -38,6 +38,11 @@ rill start my-rill-project
 
 The Rill web app runs locally at `http://localhost:9009` and will create code files in the `my-rill-project` directory.
 
+:::note
+Starting from v0.48, we have added the ability to self host the local Rill Developer application on a non-localhost endpoint. Please refer to the [rill start reference page](https://docs.rilldata.com/reference/cli/start) for more details on using the `--allowed-origins` flag. This method is not officially supported, please proceed at your own risk.
+:::
```
begelundmuller, line 43: I think we should take this out – we have *not* added the ability to self-host, it is still very much discouraged 😅  The `--allowed-origins` flag is just to provide an escape hatch for the people who do it despite the poor "support" for it.

begelundmuller:
I have pushed a commit that improves the CLI help note for `--allowed-origins`, and regenerated the CLI docs. Seems like regenerating the docs added a few unrelated changes that had been missed earlier.

docs/docs/home/get-started.md
AndrewRTsao: We can say that starting in v0.48, we've made changes to the CORS policy to only allow calls from the default HTTP URL by default for security policies. Therefore, if you are hosting Rill Developer (not officially supported), this means that you will need to pass in additional CORS origins via the `--allowed-origins` flag when starting Rill.

royendo: made the above changes!

AndrewRTsao (approved):

PR #5437, opened 2024-08-13
use correctly aliased dim expr in comparison query

pjain1:
`dimExpr` contains alias as well so it may refer to the other query dimension instead of the current query. For example if base query is a CTE then in the comparison query where clause it may use `t1.dim IN (select ... from base)`

Its strange when I tested I did not encounter this case.

### DIFF
--- a/runtime/metricsview/executor_rewrite_approx_comparisons.go
+++ b/runtime/metricsview/executor_rewrite_approx_comparisons.go
@@ -90,19 +90,6 @@ func (e *Executor) rewriteApproxComparisonNode(a *AST, n *SelectNode) bool {
 	}
 	order := []OrderFieldNode{sortField}
 
-	cteOptimization := true
-	if len(n.FromSelect.DimFields) != len(n.JoinComparisonSelect.DimFields) {
-		// generally this should not happen
-		cteOptimization = false
-	}
-	for i := range n.FromSelect.DimFields {
-		if n.FromSelect.DimFields[i].Name != n.JoinComparisonSelect.DimFields[i].Name {
-			// generally this should not happen
-			cteOptimization = false
-			break
-		}
-	}
-
 	// Note: All these cases are approximations in different ways.
 	if sortBase {
 		// We're sorting by a measure in FromSelect. We do a LEFT JOIN and push down the order/limit to it.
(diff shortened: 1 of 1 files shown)


runtime/metricsview/executor_rewrite_approx_comparisons.go
```diff
@@ -90,6 +90,19 @@ func (e *Executor) rewriteApproxComparisonNode(a *AST, n *SelectNode) bool {
...
+		// generally this should not happen
+		cteOptimization = false
+	}
+	for i := range n.FromSelect.DimFields {
+		if n.FromSelect.DimFields[i].Name != n.JoinComparisonSelect.DimFields[i].Name {
+			// generally this should not happen
+			cteOptimization = false
+			break
+		}
+	}
```
begelundmuller, line 104: Did you find cases where the `DimFields` are not equal for a comparison join? In general, it should not be possible, and it is a requirement that sub-nodes always have all the dimension fields that a parent node has (it may have more for rollups, but never fewer).

So this might better be replaced with a comment or assertion because it should never happen.

runtime/metricsview/executor_rewrite_approx_comparisons.go
```diff
@@ -100,15 +113,17 @@ func (e *Executor) rewriteApproxComparisonNode(a *AST, n *SelectNode) bool {
...
+			// make FromSelect a CTE
+			a.convertToCTE(n.FromSelect)
 
-		// now change the JoinComparisonSelect WHERE clause to use selected dim values from CTE
-		for _, dim := range n.FromSelect.DimFields {
-			dimName := a.dialect.EscapeIdentifier(dim.Name)
-			dimExpr := "(" + dim.Expr + ")" // wrap in parentheses to handle expressions
-			n.JoinComparisonSelect.Where = n.JoinComparisonSelect.Where.and(fmt.Sprintf("%[1]s IS NULL OR %[1]s IN (SELECT %[2]q.%[3]s FROM %[2]q)", dimExpr, n.FromSelect.Alias, dimName), nil)
+			// now change the JoinComparisonSelect WHERE clause to use selected dim values from CTE
+			// this necessitates that order of dims in from and comparison select are same
```
begelundmuller, line 121: Agian, this should be guaranteed, though I notice it's not very clear from the comments. Did you find a palce where this is not guaranteed?

runtime/metricsview/executor_rewrite_approx_comparisons.go
```diff
@@ -100,15 +113,17 @@ func (e *Executor) rewriteApproxComparisonNode(a *AST, n *SelectNode) bool {
...
+			for i := range n.FromSelect.DimFields {
+				dimName := a.dialect.EscapeIdentifier(n.FromSelect.DimFields[i].Name)
```
begelundmuller, line 123: Why access by `i` instead of `for _, dim := ...` and then using `dim.Name` like before?

begelundmuller (changes requested):
The fix looks good, but seems to me it can be solved by just replacing `for _, dim := range n.FromSelect.DimFields` with `for _, dim := range n.JoinComparisonSelect.DimFields` and vice versa? Since the `dim.Name` is guaranteed to be the same on both sides

runtime/metricsview/executor_rewrite_approx_comparisons.go
pjain1: no I did not but this is just a defensive check, should I replace it with assertion or just remove these checks ?

runtime/metricsview/executor_rewrite_approx_comparisons.go
pjain1: because I want to access corresponding dim from `ComparisonSelect` in the next line. So I want to use dim name and alias from base query for `IN (SELECT %[2]q.%[3]s FROM %[2]q)` and `dimExpr` from comparison to filter on using this condition and `IS NULL` condition. If dimName remains same across base and comparison then we may just use `for _, dim := range n.JoinComparisonSelect.DimFields` instead of  `n.FromSelect.DimFields` as you suggested.

runtime/metricsview/executor_rewrite_approx_comparisons.go
begelundmuller: Let's remove the checks, if the `DimFields` don't match, it would cause issues several other places as well. I'll send a PR describing the guarantees better

runtime/metricsview/executor_rewrite_approx_comparisons.go
begelundmuller: Yeah the dimension name is guaranteed to be the same – it is a fundamental guarantee that `SelectNode`s used in  joins must have the same `DimFields`

begelundmuller (approved):

PR #5440, opened 2024-08-13
billing errors and warnings, stripe and orb webhook, river worker for async background jobs

pjain1:
Fixes https://github.com/rilldata/rill/issues/5467, https://github.com/rilldata/rill/issues/5466 and https://github.com/rilldata/rill-private-issues/issues/527


admin/server/server.go
```diff
@@ -235,6 +235,24 @@ func (s *Server) HTTPHandler(ctx context.Context) (http.Handler, error) {
 	// Add Github-related endpoints (not gRPC handlers, just regular endpoints on /github/*)
 	s.registerGithubEndpoints(mux)
 
+	// Add biller webhook handler if any
+	if s.admin.Biller != nil {
+		handler := s.admin.Biller.WebhookHandlerFunc(ctx)
+		if handler != nil {
+			// TODO add observability middleware and rate limiter
```
pjain1, line 242: 1. May be we don't add observability to this endpoint as Orb for example cannot selectively deliver events so it might be too much logging
2. Should we move webhook endpoints to worker server ?

admin/billing/biller.go
```diff
@@ -32,21 +34,32 @@ type Biller interface {
...
+	IsInvoiceValid(ctx context.Context, invoice *Invoice) bool
+	IsInvoicePaid(ctx context.Context, invoice *Invoice) bool
+
 	ReportUsage(ctx context.Context, usage []*Usage) error
 
 	GetReportingGranularity() UsageReportingGranularity
 	GetReportingWorkerCron() string
+
+	// WebhookHandlerFunc returns a http.HandlerFunc that can be used to handle incoming webhooks from the payment provider. Return nil if you don't want to register any webhook handlers.
+	WebhookHandlerFunc(ctx context.Context) func(w http.ResponseWriter, r *http.Request)
```
begelundmuller, line 59: Can use `http.HandleFunc` for the return type

admin/billing/biller.go
```diff
@@ -32,21 +34,32 @@ type Biller interface {
...
+
+	// SetJobsClient needs to be explicitly set because of circular dependency, see admin start method
+	SetJobsClient(jobs jobs.Client)
```
begelundmuller, line 62: Not sure, but you could consider merging this into `WebhookHandlerFunc(ctx, jobs)` – to clarify that the webhooks simply dispatch jobs. That would also clarify that the billing client will only submit jobs when and as long as the webhooks are served. Right now, I wondered if some other parts of the billing client might send jobs (and was wondering if it could somehow lead to circular job invocations.)

admin/billing/payment/payment.go
```diff
@@ -16,6 +18,12 @@ type Provider interface {
 	DeleteCustomer(ctx context.Context, customerID string) error
 	// GetBillingPortalURL returns the payment portal URL to collect payment information from the customer.
 	GetBillingPortalURL(ctx context.Context, customerID, returnURL string) (string, error)
+
+	// WebhookHandlerFunc returns a http.HandlerFunc that can be used to handle incoming webhooks from the payment provider. Return nil if you don't want to register any webhook handlers.
+	WebhookHandlerFunc(ctx context.Context) func(w http.ResponseWriter, r *http.Request)
+
+	// SetJobsClient needs to be explicitly set because of circular dependency, see admin start method
+	SetJobsClient(jobs jobs.Client)
```
begelundmuller, line 26: The same comments as on `billing.Client` about `http.HandleFunc` and passing `jobs.Client` to the handler apply here

admin/billing/payment/stripe_webhook.go
```diff
@@ -0,0 +1,139 @@
...
+	"time"
+
+	"github.com/stripe/stripe-go/v79"
+	"github.com/stripe/stripe-go/v79/webhook"
+	"go.uber.org/zap"
+)
+
+// handleWebhook handles incoming webhook events from Stripe
+func (s *Stripe) handleWebhook(w http.ResponseWriter, r *http.Request) {
+	const MaxBodyBytes = int64(65536)
```
begelundmuller, line 18: Why make it exported? Also consider putting consts at the top of the file instead of inside a function.

admin/billing/payment/stripe_webhook.go
```diff
@@ -0,0 +1,139 @@
...
+	r.Body = http.MaxBytesReader(w, r.Body, MaxBodyBytes)
+	payload, err := io.ReadAll(r.Body)
+	if err != nil {
+		s.logger.Error("error reading request body", zap.Error(err))
+		http.Error(w, "error reading request body", http.StatusServiceUnavailable)
+		return
+	}
```
begelundmuller, line 25: Consider using `httputil.Handler` and `httputil.Errorf` for returning errors.

Also consider using `observability.Middleware` for logging HTTP requests and errors instead of duplicating the error messages as logs (as happens in several cases here)

admin/billing/payment/stripe_webhook.go
```diff
@@ -0,0 +1,139 @@
...
+	endpointSecret := s.webhookSecret
+	sigHeader := r.Header.Get("Stripe-Signature")
+	event, err := webhook.ConstructEventWithOptions(payload, sigHeader, endpointSecret, webhook.ConstructEventOptions{IgnoreAPIVersionMismatch: true})
+	if err != nil {
+		s.logger.Error("error verifying webhook signature", zap.Error(err))
+		http.Error(w, "error verifying webhook signature", http.StatusBadRequest)
+		return
+	}
+
+	s.logger.Debug(fmt.Sprintf("got event : %s\n", event.Data))
```
begelundmuller, line 36: Mistakenly committed? It's not clear this relates to stripe

admin/database/postgres/migrations/0044.sql
```diff
@@ -0,0 +1,21 @@
...
+    org_id UUID NOT NULL,
+    type INTEGER NOT NULL,
+    metadata JSONB NOT NULL DEFAULT '{}'::JSONB,
+    event_time TIMESTAMPTZ NOT NULL DEFAULT now(),
+    created_on TIMESTAMPTZ NOT NULL DEFAULT now(),
+    FOREIGN KEY (org_id) REFERENCES orgs (id) ON DELETE CASCADE,
+    CONSTRAINT billing_errors_org_id_type_unique UNIQUE (org_id, type)
+);
+
+CREATE TABLE billing_warnings (
```
begelundmuller, line 12: Since the schema is the same, would it make sense to merge this with `billing_errors` and using a column to determine if it's a warning – for example by incorporating the warning types in the `type` enum or adding something like a `level` or `warning BOOL` field instead? Presumably that would also make consuming the errors/warnings simpler downstream?

admin/jobs/river/biller_event_handlers.go
```diff
@@ -0,0 +1,286 @@
...
+	err = w.admin.Email.SendInformational(&email.Informational{
+		ToEmail: org.BillingEmail,
+		ToName:  org.Name,
+		Subject: "Your invoice payment failed",
+		Title:   "",
+		Body:    template.HTML(fmt.Sprintf("Your invoice payment failed, please visit the billing portal to fix issues or contact support to continue using Rill. Your projects will be hibernated on %s UTC, if invoice still not paid.", gracePeriodEndDate.Format("2006-01-02"))),
+	})
+	if err != nil {
+		return fmt.Errorf("failed to send invoice payment failed email for org %q: %w", org.Name, err)
+	}
```
begelundmuller, line 100: This should be wrapped in a function in the `runtime/pkg/email` package, something like `Email.SendPaymentFailed` or something like that. The goal is that all email text is contained in the `email` package so that it's easier for the product folks to audit and edit.

admin/jobs/river/biller_event_handlers.go
```diff
@@ -0,0 +1,286 @@
...
+		err = w.admin.Email.SendInformational(&email.Informational{
+			ToEmail: org.BillingEmail,
+			ToName:  org.Name,
+			Subject: "Invoice payment default",
+			Title:   "",
+			Body:    "We did not receive any payment for the unpaid invoice, your projects have been hibernated. Please visit the billing portal to complete payment.",
+		})
+		if err != nil {
+			return fmt.Errorf("failed to send payment method expired email for org %q: %w", org.Name, err)
+		}
```
begelundmuller, line 281: Same comment as above

admin/jobs/river/biller_event_handlers.go
```diff
@@ -0,0 +1,286 @@
...
+					return fmt.Errorf("failed to hibernate project %q: %w", proj.Name, err)
+				}
+				afterProjectName = proj.Name
+			}
+
+			if len(projs) < limit {
+				break
+			}
+		}
+		w.admin.Logger.Info("Projects hibernated due to no valid payment method", zap.String("org_id", org.ID), zap.String("org_name", org.Name), zap.String("billing_customer_id", org.BillingCustomerID))
```
begelundmuller, line 269: Is there any way we can namespace all these billing-related logs? Either with a `Named` logger or with a `billing: ...` prefix in the messages. It might be nice to setup a Slack channel of all billing-related logs.

admin/jobs/river/river.go
```diff
@@ -149,7 +463,7 @@ func (h *ErrorHandler) HandlePanic(ctx context.Context, job *rivertype.JobRow, p
 	var args string
 	_ = json.Unmarshal(job.EncodedArgs, &args) // ignore parse errors
 	h.logger.Error("Job panicked", zap.Int64("job_id", job.ID), zap.String("kind", job.Kind), zap.String("args", args), zap.Any("panic_val", panicVal), zap.String("trace", trace))
-	// Set the job to be immediately cancelled
+	// Set the job to be immediately cancelled TODO review if we should retry or cancel the job
```
begelundmuller, line 466: Does this need to be handled now?

admin/server/organizations.go
```diff
@@ -1198,6 +1298,264 @@ func (s *Server) ListPublicBillingPlans(ctx context.Context, req *adminv1.ListPu
 	}, nil
 }
 
+func (s *Server) ListOrganizationBillingErrors(ctx context.Context, req *adminv1.ListOrganizationBillingErrorsRequest) (*adminv1.ListOrganizationBillingErrorsResponse, error) {
```
begelundmuller, line 1301: Since this file is getting massive, can we break the billing-related RPCs into a separate `admin/server/billing.go` file?

admin/server/server.go
begelundmuller: 1. How many events are we talking about? In general, it would be nice to have logging for all requests as it also helps us understand load and where errors are occurring.
2. No the worker server is really only a fallback for health checks, it's not supposed to be exposed on the public internet. This is the service's external server, so this is the place to capture webhooks.

cli/cmd/billing/list_errors.go
```diff
@@ -0,0 +1,56 @@
...
+				orgNames, err := org.OrgNames(ctx, ch)
+				if err != nil {
+					return err
+				}
+
+				orgName, err = cmdutil.SelectPrompt("Select org to list billing errors", orgNames, ch.Org)
+				if err != nil {
+					return err
+				}
+			}
```
begelundmuller, line 35: This and similar prompts in the other commands can be removed. Try searching for `&ch.Org` in the code. Our convention is that it falls back to using the org configured in `~/.rill` unless you explicitly pass `--org`. 

Also consider setting `--og` as a persistent flag for all `rill billing` commands if possible, like for example for `rill project`: https://github.com/rilldata/rill/blob/75b38efbcd4dc109c4a9b624ddbd20aef623614d/cli/cmd/project/project.go#L19

cli/cmd/billing/subscription/cancel.go
```diff
@@ -0,0 +1,74 @@
...
+
+			if subResp.Subscription == nil {
+				ch.PrintfWarn("No subscriptions found for organization %q\n", orgName)
+				return nil
+			}
+
+			ch.PrintfBold("Organization has following subscription\n")
+			ch.PrintSubscriptions([]*adminv1.Subscription{subResp.Subscription})
+
+			ok, err := cmdutil.ConfirmPrompt("\nAfter end of the current billing cycle, you will loose access to your projects. Do you want to Continue ?\n", "", false)
```
begelundmuller, line 51:
```suggestion
			ch.PrintfWarn("\nAt the end of the current billing cycle, you will lose access to %q and all its projects.", ch.Org)
			ok, err := cmdutil.ConfirmPrompt("Do you want to continue?", "", false)
```

cli/pkg/printer/resources.go
```diff
@@ -503,3 +504,79 @@ type modelSplit struct {
...
+		CreatedAt:    e.CreatedOn.AsTime().Local().Format(time.DateTime),
+	}
+}
+
+type billingError struct {
+	Organization string `header:"organization" json:"organization"`
+	Type         string `header:"type" json:"type"`
+	Metadata     string `header:"metadata" json:"metadata"`
+	EventTime    string `header:"event_time,timestamp(ms|utc|human)" json:"event_time"`
+	CreatedAt    string `header:"created_at,timestamp(ms|utc|human)" json:"created_at"`
```
begelundmuller, line 543: Can `created_at` be omited? Seems like `event_time` should be enough for users?

cli/pkg/printer/resources.go
```diff
@@ -503,3 +504,79 @@ type modelSplit struct {
...
+		CreatedAt:    w.CreatedOn.AsTime().Local().Format(time.DateTime),
+	}
+}
+
+type billingWarning struct {
+	Organization string `header:"organization" json:"organization"`
+	Type         string `header:"type" json:"type"`
+	Metadata     string `header:"metadata" json:"metadata"`
+	EventTime    string `header:"event_time,timestamp(ms|utc|human)" json:"event_time"`
+	CreatedAt    string `header:"created_at,timestamp(ms|utc|human)" json:"created_at"`
```
begelundmuller, line 581: Same comment

proto/rill/admin/v1/api.proto
```diff
@@ -503,6 +503,18 @@ service AdminService {
     };
   }
 
+  rpc SudoDeleteOrganizationBillingError(SudoDeleteOrganizationBillingErrorRequest) returns (SudoDeleteOrganizationBillingErrorResponse) {
+    option (google.api.http) = {
+      delete: "/v1/superuser/organizations/{organization}/billing/errors/{type}",
+    };
+  }
+
+  rpc SudoDeleteOrganizationBillingWarning(SudoDeleteOrganizationBillingWarningRequest) returns (SudoDeleteOrganizationBillingWarningResponse) {
```
begelundmuller, line 512: Please add brief docstring to the RPCs

proto/rill/admin/v1/api.proto
```diff
@@ -788,6 +807,14 @@ service AdminService {
       body: "*"
     };
   }
+
+  rpc ListOrganizationBillingErrors(ListOrganizationBillingErrorsRequest) returns (ListOrganizationBillingErrorsResponse) {
+    option (google.api.http) = {get: "/v1/organizations/{organization}/billing/errors"};
+  }
+
+  rpc ListOrganizationBillingWarnings(ListOrganizationBillingWarningsRequest) returns (ListOrganizationBillingWarningsResponse) {
```
begelundmuller, line 815: Please also add docstrings here

proto/rill/admin/v1/api.proto
```diff
@@ -1274,6 +1301,34 @@ message SudoIssueRuntimeManagerTokenResponse {
...
+message SudoDeleteOrganizationBillingErrorResponse {}
+
+enum BillingErrorType {
+  BILLING_ERROR_TYPE_UNSPECIFIED = 0;
+  BILLING_ERROR_TYPE_NO_PAYMENT_METHOD = 1;
+  BILLING_ERROR_TYPE_NO_BILLABLE_ADDRESS = 2;
+  BILLING_ERROR_TYPE_INVOICE_PAYMENT_FAILED = 3;
+  BILLING_ERROR_TYPE_TRIAL_ENDED = 4;
+  BILLING_ERROR_TYPE_SUBSCRIPTION_CANCELLED = 5;
+}
```
begelundmuller, line 1318: By convention, non-request/response types are placed at the bottom of the file after all the request/response types

proto/rill/admin/v1/api.proto
```diff
@@ -1274,6 +1301,34 @@ message SudoIssueRuntimeManagerTokenResponse {
...
+  string organization = 1;
+  BillingWarningType type = 2;
+}
+
+message SudoDeleteOrganizationBillingWarningResponse {}
+
+enum BillingWarningType {
+  BILLING_WARNING_TYPE_UNSPECIFIED = 0;
+  BILLING_WARNING_TYPE_ON_TRIAL = 1;
+}
```
begelundmuller, line 1330: Same comment, place after all the request/response types

proto/rill/admin/v1/api.proto
```diff
@@ -1951,6 +2006,12 @@ message UpdateBillingSubscriptionRequest {
   string plan_name = 2;
 }
 
+message CancelBillingSubscriptionRequest {
+  string org_name = 1;
```
begelundmuller, line 2010: call it `string organization` for alignment with other RPCs

proto/rill/admin/v1/api.proto
```diff
@@ -2280,3 +2357,64 @@ message MemberUsergroup {
...
+  google.protobuf.Timestamp due_date = 6;
+}
+
+message BillingErrorMetadataTrialEnded {
+  google.protobuf.Timestamp grace_period_end_date = 1;
+}
+
+message BillingErrorMetadataSubscriptionCancelled {
+  google.protobuf.Timestamp end_date = 1;
+}
```
begelundmuller, line 2402: Just wondering from a simplicity point of view whether this info could be simplified to a string message? Or does the CLI/frontend need it destructured like this?

begelundmuller (changes requested):

admin/server/server.go
pjain1: Few events during org creation and then few events on subscription addition and changes. Then events on every invoice issue, creation, payment etc. Its just that since Orb cannot selectively deliver events, may be its ok. I can add observability.

admin/billing/biller.go
pjain1: yeah made it generic so that it can be used through out the billing code but currently only used by webhook handler so might make send to use it just there.

admin/billing/payment/stripe_webhook.go
pjain1: yeah added here because observability middleware was not added, will remove this

admin/billing/payment/stripe_webhook.go
pjain1: was looking at all events that we receive, left as it was debug level, will remove it.

admin/jobs/river/river.go
pjain1: no just wanted to discuss if we should retry jobs in case of panic or just log the panic and mark the job as complete. For now its marking the job as complete, it does so because mostly we will just run into panic again on retry.

admin/jobs/river/biller_event_handlers.go
pjain1: nice idea, added a named logger

cli/cmd/billing/list_errors.go
pjain1: Yeah I deliberately did this so that every time user perform any billing related action, they choose an org before hand even if there is single org so they are aware of it. Sometimes I feel falling back behaviour to org in `~/.rill` in case of multiple orgs not that great in case the cmd updates any resource (or we add prompts in such cases). Its fine for read operations like in this case. Let me know if you think such caution is not needed.

proto/rill/admin/v1/api.proto
pjain1: Earlier I added string messages but found that this to be would be better for UI if they want to create call to action. For example failed invoice meta contains url that can directly take them to the failed invoice page and they can pay. Or use some of the end dates to perform something.

cli/cmd/billing/list_errors.go
pjain1: May be I can add clear prompts with org name for updates/delete action and auto select for reads

proto/rill/admin/v1/api.proto
begelundmuller: For a CTA, couldn't that be managed with either just an error code and message, or message and a `cta_url`? Or does the frontend need deeper understanding of the error?

pjain1: Still feel its better to have it like this. UI can choose to create their own message. Another example on how these fields can be used is trial ending soon metadata contains trial end date or sub cancel error metadata contains sub end date so the UI can choose to display some banner or pop up on the end date or sometime before end date etc. Don't think it will be complicated for UI to use these to create flows as per product requirement since its typed metadata instead of json blobs.

admin/billing/biller.go
pjain1: ok changed

admin/jobs/river/river.go
```diff
@@ -134,6 +160,296 @@ func (c *Client) ResetAllDeployments(ctx context.Context) (*jobs.InsertResult, e
...
+	}, nil
+}
+
+func (c *Client) InvoicePaymentFailedGracePeriodCheck(ctx context.Context, orgID, invoiceID string, gracePeriodEndDate time.Time) (*jobs.InsertResult, error) {
+	res, err := c.riverClient.Insert(ctx, InvoicePaymentFailedGracePeriodCheckArgs{
+		OrgID:              orgID,
+		InvoiceID:          invoiceID,
+		GracePeriodEndDate: gracePeriodEndDate,
+	}, &river.InsertOpts{
+		ScheduledAt: gracePeriodEndDate.AddDate(0, 0, 1).Add(1 * time.Hour), // end of grace period date + 1 hour buffer
```
begelundmuller, line 293: Unfortunately I think we need to reassess the use of `ScheduledAt`. Having jobs scheduled far in the future makes them much harder to maintain because you have to think about backwards compatibility and job migrations. There's also issues of jobs going stale that needs to be cancelled or correctly ignored upon invocation.

For example, if you look at Temporal, which is a more advanced background job system, you'll see they devote a lot of docs to discussing versioning and compatibility issues for these kinds of issues, and have some advanced features for dealing with them. 

Looking at the use of `ScheduledAt`, it seems like these cases could relatively easily be moved into a cron job? For example, for this `InvoicePaymentFailedGracePeriodCheck` check, you could have an `payment_failed_grace_period_ends_on` timestamp in the `orgs` table, and have a cron job that grabs all rows where it is in the past and processes them. It is effectively the same approach, but makes the job itself stateless and easy to change.

begelundmuller (changes requested):

admin/jobs/river/river.go
pjain1: ok I see, makes sense. May be we can have cron jobs for all the billing errors/warning that requires some processing in future. When they wake up (mostly past every midnight) they look for all the orgs with that billing error and checks if its action time and takes action. We will need to maintain a flag if the billing error is processed or not to prevent reprocessing. 

The problem you mentioned about backwards compatibility and job migrations applies somewhat for immediate scheduled jobs as well, for this we can just make sure that worker does graceful shutdown during upgrades and allows current running jobs to finish.

admin/billing/payment/stripe.go
```diff
@@ -119,3 +128,11 @@ func (s *Stripe) GetBillingPortalURL(ctx context.Context, customerID, returnURL
 
 	return sess.URL, nil
 }
+
+func (s *Stripe) WebhookHandlerFunc(ctx context.Context, jc jobs.Client) httputil.Handler {
+	if s.webhookSecret == "" {
+		return nil
+	}
```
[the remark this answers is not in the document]
pjain1: > // WebhookHandlerFunc returns a http.HandlerFunc that can be used to handle incoming webhooks from the payment provider. Return nil if you don't want to register any webhook handlers.

Its documented and easy to mount a handler that just returns nil which would send http ok and webhook events will be just lost. Instead a handler like this needs to be mounted
```
return func(w http.ResponseWriter, r *http.Request) error {
    return httputil.Errorf(http.StatusBadRequest, "webhook secret not set")
}
```
Let me know if this should be done

begelundmuller (approved):
LGTM

PR #5443, opened 2024-08-13
disable cte optimization for multi value druid dims

pjain1:
hack until druid supports it

### DIFF
--- a/runtime/metricsview/executor_rewrite_approx_comparisons.go
+++ b/runtime/metricsview/executor_rewrite_approx_comparisons.go
@@ -2,6 +2,7 @@ package metricsview
 
 import (
 	"fmt"
+
 	"github.com/rilldata/rill/runtime/drivers"
 )
 
--- a/runtime/metricsview/executor_rewrite_approx_comparisons.go
+++ b/runtime/metricsview/executor_rewrite_approx_comparisons.go
@@ -46,12 +46,10 @@ func (e *Executor) rewriteApproxComparisonNode(a *AST, n *SelectNode) bool {
 	}
 	sortField := a.Root.OrderBy[0]
 
-	// if dim does not have explicit unnest, then don't optimize
-	// for example druid fails with join on cte having multi value dimension
-	for _, dim := range n.FromSelect.DimFields {
-		if dim.AutoUnnest {
-			return false
-		}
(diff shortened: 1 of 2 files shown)


runtime/metricsview/executor_rewrite_approx_comparisons.go
```diff
@@ -46,6 +46,14 @@ func (e *Executor) rewriteApproxComparisonNode(a *AST, n *SelectNode) bool {
 	}
 	sortField := a.Root.OrderBy[0]
 
+	// if dim does not have explicit unnest, then don't optimize
+	// for example druid fails with join on cte having multi value dimension
+	for _, dim := range n.FromSelect.DimFields {
+		if dim.AutoUnnest {
+			return false
+		}
+	}
```
begelundmuller, line 55: Could we just check `if len(a.unnests) > 0 { return false }` and then avoid the `AutoUnnest` field on `FieldNode`?

begelundmuller (changes requested):

runtime/metricsview/executor_rewrite_approx_comparisons.go
pjain1: ok removed, I wasn't sure if dim fields on parent select node will always be same as dim fields in FromSelect.

begelundmuller (approved):

runtime/metricsview/executor_rewrite_approx_comparisons.go
```diff
@@ -46,6 +46,12 @@ func (e *Executor) rewriteApproxComparisonNode(a *AST, n *SelectNode) bool {
 	}
 	sortField := a.Root.OrderBy[0]
 
+	// if there are unnests in the query, we can't rewrite the query
+	// for example druid fails with join on cte having multi value dimension, issue - https://github.com/apache/druid/issues/16896
+	if len(a.unnests) > 0 {
+		return false
+	}
```
begelundmuller, line 53: Actually, should it be something like `if e.olap.Dialect == drivers.DialectDruid && len(a.unnests) > 0`?

begelundmuller (changes requested):

begelundmuller (approved):

Issue #5385, opened 2024-08-02
Iteratively build query in `database.UpdateProject` based on the options set.


k-anshul:
Currently all the options need to be set wherever we are doing project update in database. This is easy to miss given we update project in many places.
We should iteratively build query in database.UpdateProject based on the options set.
I have seen this bug at least a couple of times.

PR #5444, opened 2024-08-13
Set up exhauststruct and clean up linter rules

begelundmuller:
- Adds `exhauststruct` for options types in `admin/database` and certain usages of `runtime/metricsview.Query`
- Audited and cleaned up all linter rules and fixed various issues that came up

Closes https://github.com/rilldata/rill/issues/5385


.golangci.yml
```diff
@@ -1,46 +1,28 @@
...
-  # Which files to skip: they will be analyzed, but issues from them won't be reported.
-  # Default value is empty list,
-  # but there is no need to include all autogenerated files,
-  # we confidently recognize autogenerated files.
-  # If it's not please let us know.
-  # "/" will be replaced by current OS file path separator to properly work on Windows.
-  skip-files:
-    - ".*\\.my\\.go$"
-  issues-exit-code: 1
+  tests: false # TODO: Change to true
```
k-anshul, line 3: Does this require addressing before merging ?

runtime/drivers/duckdb/utils.go
```diff
@@ -246,7 +246,7 @@ func fileSize(paths []string) int64 {
 }
 
 func quoteName(name string) string {
```
k-anshul, line 248: May be this can be removed and  we can use functions from `Dialect` itself ?

runtime/metricsview/astexpr.go
```diff
@@ -95,7 +96,7 @@ func (b *sqlExprBuilder) writeSubquery(sub *Subquery) error {
 		Offset:              nil,
 		TimeZone:            outer.TimeZone,
 		Label:               false,
-	}
+	} //exhaustruct:enforce
```
k-anshul, line 99: Why is this separately required ?
Also if you take a pull from main. This is required at few more places introduced in the `MetricsViewSearch` PR.

runtime/drivers/admin/repo.go
```diff
@@ -72,7 +72,7 @@ func (h *Handle) ListRecursive(ctx context.Context, glob string, skipDirs bool)
 		}
 
 		// Track file (p is already relative to the FS root)
-		p = filepath.Join("/", p)
+		p = path.Join("/", p)
```
k-anshul, line 75: Out of curiosity, why did we switched to `path` pkg from `filepath` pkg ?
Given these are operating system paths and not URL paths.

k-anshul (changes requested):

.golangci.yml
begelundmuller: No it was also false before and gives a lot of errors, so should be done separately. Actually I will just remove the comment and put it in the code quality burndown.

runtime/drivers/duckdb/utils.go
begelundmuller: Done

runtime/drivers/admin/repo.go
begelundmuller: I un-disabled a linter that checks if you use `/` with `filepath`, which is illegal because `filepath` requires you to use the OS path separator. In this case however, the `p` is obtained from the `doublestar` package, which always normalizes paths as Unix paths (even on Windows).

k-anshul (approved):

PR #5446, opened 2024-08-13
Prevent API requests from firing when no measures

### LINKED ISSUE #5224: Security policy without measures should not fire certain API requests
If you create a security policy that only includes dimension fields, but not any measure fields, it:
1. Makes a `/timeseries` request, which fails because there are no measure names
2. Makes `/aggregation` requests with an empty sort key
3. Makes a request for `/resource?name.kind=rill.runtime.v1.ProjectParser&name.name=parser`, which fails because restricted users don’t have access to parse errors

```
security:
  access: true
  include:
    - if: true
      names: [company, plan_name]
```

Screenshot 2024-07-08 at 18 45 24

lovincyrus:
This PR prevents the API requests from running when there are no measures mentioned in the security policy. Closes #5224 

- [x] How to check for project admins?
- [x] Check for no measures before firing the API requests
- [x] Check error banner for project admins only
- [x] Update "Error parsing dashboard" to banner instead of ErrorPage
- [x] Test error banner when there is a parser error
- [x] admin & non-admin - ErrorPage - not measures available
- [x] use the actual measures
- [x] non-admin - hide the error banner, prevent api request from running

Related: https://github.com/rilldata/rill/pull/5432

Set up this project to test this PR: https://github.com/rilldata/rill/issues/5224#issuecomment-2286370172

### DIFF
--- a/web-common/src/features/entity-management/resource-selectors.ts
+++ b/web-common/src/features/entity-management/resource-selectors.ts
@@ -74,6 +74,7 @@ export function useResource<T = V1Resource>(
   );
 }
 
+// FIXME: To remove this duplicate of `useResource` https://github.com/rilldata/rill/pull/5531#discussion_r1733027626
 /**
  * `useResourceV2` is a more flexible version of `useResource` that accepts
  *  any `queryOptions`, not just `select` and `queryClient`.
@@ -112,13 +113,18 @@ export function useResourceV2<T = V1Resource>(
   );
 }
 
-export function useProjectParser(queryClient: QueryClient, instanceId: string) {
+export function useProjectParser(
+  queryClient: QueryClient,
+  instanceId: string,
+  enabled: boolean = true,
+) {
   return useResource(
     instanceId,
     SingletonProjectParserName,
     ResourceKind.ProjectParser,
     {
       queryClient,
+      enabled,
     },
   );
 }
--- a/web-common/src/features/entity-management/resource-selectors.ts
+++ b/web-common/src/features/entity-management/resource-selectors.ts
@@ -116,7 +116,11 @@ export function useResourceV2<T = V1Resource>(
 export function useProjectParser(
   queryClient: QueryClient,
   instanceId: string,
-  enabled: boolean = true,
+  queryOptions?: CreateQueryOptions<
+    V1GetResourceResponse,
+    ErrorType<RpcStatus>,
+    V1Resource
+  >,
 ) {
   return useResource(
     instanceId,
@@ -124,7 +128,7 @@ export function useProjectParser(
     ResourceKind.ProjectParser,
     {
       queryClient,
-      enabled,
+      ...queryOptions,
     },
   );
 }
--- a/web-local/src/routes/(viz)/dashboard/[name]/+page.svelte
+++ b/web-local/src/routes/(viz)/dashboard/[name]/+page.svelte
@@ -35,10 +35,6 @@
   $: mockUserHasNoAccess =
     $selectedMockUserStore && $dashboard.error?.response?.status === 404;
 
-  $: if (mockUserHasNoAccess) {
-    eventBus.emit("banner", null);
-  }
-
   // Handle errors from dashboard YAML edits from an external IDE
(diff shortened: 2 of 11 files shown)


lovincyrus:
Recording:

https://github.com/user-attachments/assets/827f7dbf-b01b-43dc-a20a-f94d3db5d0e7

lovincyrus:
Addressed the feedback from our offline sync @ericpgreen2

web-common/src/features/entity-management/resource-selectors.ts
```diff
@@ -112,13 +113,18 @@ export function useResourceV2<T = V1Resource>(
...
   return useResource(
     instanceId,
     SingletonProjectParserName,
     ResourceKind.ProjectParser,
     {
       queryClient,
+      enabled,
     },
   );
 }
```
ericpgreen2, line 130: It'd be nice to change the function signature to accept a `queryOptions` object, then we won't need to keep modifying it to accept more individual `queryOptions` going forward. See [this instance of `useDashboard`](https://github.com/rilldata/rill/blob/cyrus/prevent-requests-no-measures/web-common/src/features/dashboards/selectors.ts#L25-L40).

web-local/src/routes/(viz)/dashboard/[name]/+page.svelte
```diff
@@ -24,26 +25,45 @@
...
+      type: "error",
+      message:
+        "Error parsing dashboard – you are viewing your last valid dashboard specification",
+      iconType: "alert",
+    });
+  }
+
+  $: if (mockUserHasNoAccess || measures.length === 0) {
+    // We want to close the banner before any of the ErrorPage is rendered
+    eventBus.emit("banner", null);
```
Comment (ericpgreen2) [inline web-local/src/routes/(viz)/dashboard/[name]/+page.svelte:54]: We should also clear the banner when navigating away from the page. (Which is logic that probably should be built into the Banner Center at some point.)

web-local/src/routes/(viz)/dashboard/[name]/+page.svelte
```diff
@@ -24,26 +25,45 @@
...
+
+  // Handle errors from dashboard YAML edits from an external IDE
+  $: if (dashboardFileHasParseError && dashboardFileHasParseError.length > 0) {
+    eventBus.emit("banner", {
+      type: "error",
+      message:
+        "Error parsing dashboard – you are viewing your last valid dashboard specification",
+      iconType: "alert",
+    });
+  }
```
Comment (ericpgreen2) [inline web-local/src/routes/(viz)/dashboard/[name]/+page.svelte:50]: So I tried this out and, apologies, I'm observing different behavior from what I originally expected. Actually, a parse error results in the resource being lost outright. There's no valid spec retained. So it seems we should revert this and show the "Error parsing dashboard" error page.

web-local/src/routes/(viz)/dashboard/[name]/+page.svelte
```diff
@@ -24,26 +25,45 @@
   $: ({ instanceId } = $runtime);
 
   $: filePaths = data.metricsView.meta?.filePaths as string[];
-  $: projectParserQuery = useProjectParser(queryClient, instanceId);
+  $: dashboard = useDashboard(instanceId, metricsViewName);
+  $: measures = $dashboard.data?.metricsView?.state?.validSpec?.measures ?? [];
+  $: projectParserQuery = useProjectParser(
+    queryClient,
+    instanceId,
+    measures.length > 0,
```
Comment (ericpgreen2) [inline web-local/src/routes/(viz)/dashboard/[name]/+page.svelte:33]: I don't think we need to prevent the Project Parser request when there are no measures. Rather we should prevent it when the mock user is not an admin.

ericpgreen2 (changes requested):

ericpgreen2 (approved):
This is good for now (pending my last 2 PR comments), though I'll note that this doesn't handle the case for Rill Cloud. We can skip Rill Cloud handling for now because:
1) The error case is more likely in Rill Developer (when you're developing your security policy) than in Rill Cloud
2) It'll be cleaner to add this logic to Rill Cloud once we move the dashboard's instantiation logic to a SvelteKit `load` function (which @AdityaHegde is considering in his "human-readable" URL state work), so IMO we can wait until then.

PR #5451, opened 2024-08-14
fix: login/logout outside of rill dev not updating until a restart

AdityaHegde:


AdityaHegde:
Closing in favour of #5460

PR #5457, opened 2024-08-16
support for `query_settings` for clickhouse models

k-anshul:


runtime/drivers/object_store.go
```diff
@@ -42,6 +42,6 @@ type ObjectStoreModelOutputProperties struct {
 }
 
 type ObjectStoreModelResultProperties struct {
-	Path   string     `mapstructure:"path"`
-	Format FileFormat `mapstructure:"format"`
+	Path   string `mapstructure:"path"`
+	Format string `mapstructure:"format"`
```
begelundmuller, line 46: Why this change?

k-anshul: When we are storing the results in catalog here : https://github.com/rilldata/rill/blob/b970338f4df225c7fc3d1acb86c27872ae52fe5c/runtime/reconcilers/model.go#L523-L524
it is not able to handle `FileFormat` type
It needs go default types only.

k-anshul: I think we do not hit this case right now but we will hit this case if we just have a model which has `snowflake` as input and `s3` as output.

begelundmuller: Got it

begelundmuller (approved):

PR #5458, opened 2024-08-16
feat: Add orgnaization tabs and settings tab skeleton

AdityaHegde:
1. Adds tabs for organization that is hidden if the available tabs <= 1
2. Adds skeleton for settings page that can be reused in project settings page.


lovincyrus (commented):
The project settings page looks great. I can definitely re-use it for the public URLs table.

Unsure about the addition of sub-navigation bar (1 in the screenshot). Ideally, we only have 1 nav bar to avoid confusion. Also asked for clarification on the design here https://www.figma.com/design/Qt6EyotCBS3V6O31jVhMQ7?node-id=144-815#915050693

CleanShot 2024-08-16 at 10 10 56@2x

Should be Settings instead of Status

CleanShot 2024-08-16 at 10 14 55@2x

jkhwu:
Yes @lovincyrus is correct, the intent was for only one tab bar to show at a time. The org tab bar when you're on an org page, the project tab bar when you're on a project page.

lovincyrus (dismissed):
Spacing nit - could we align the nav tabs with the wordmark above?

CleanShot 2024-08-19 at 11 24 00@2x


CleanShot 2024-08-19 at 11 22 59@2x

AdityaHegde:
@lovincyrus The outline is not really visible to users. The start of rill logo aligns with the tab background when hovered. You can see that in the 1st image you posted.

Side note the rill logo doesnt have a hover state in mocks but it does in UI. We should figure it out separately.

web-admin/src/features/navigation/TopNavigationBar.svelte
```diff
@@ -54,6 +55,7 @@
   $: onReportPage = !!report;
   $: onMetricsExplorerPage = isMetricsExplorerPage($page);
   $: onPublicURLPage = isPublicURLPage($page);
+  $: witinOrgPage = withinOrganization($page);
```
ericpgreen2, line 58: typo: `witin` -> `within`

web-admin/src/routes/[organization]/-/settings/+layout.svelte
```diff
@@ -0,0 +1,31 @@
...
+</script>
+
+<LeftNav
+  title="Settings"
+  {basePage}
+  baseRoute="/[organization]/-/settings"
+  {navItems}
+>
+  <slot />
+</LeftNav>
```
Comment (ericpgreen2) [inline web-admin/src/routes/[organization]/-/settings/+layout.svelte:31]: Rather than the `main` page content being a child of `LeftNav`, I think it'd be more conventional for the `main` page content and `LeftNav` component to be siblings, like:
```
<div class="layout-container">
  <LeftNav {title} {basePage} {baseRoute} {navItems} />
  <main>
    <slot />
  </main>
</div>
```

Comment (ericpgreen2) [reply web-admin/src/routes/[organization]/-/settings/+layout.svelte:31]: With this approach, it'd also make sense to pull the "Settings" header out of the `LeftNav` component.

ericpgreen2 (changes requested):
UXQA –
1. When any of the left nav tabs are highlighted, the "Settings" tab should retain its active state.
    image
2. When arriving on the settings page, the "General" tab should be active.
    image

ericpgreen2 (approved):
Sweet. Agree w/ Cyrus that we should reconsider the alignment between the Rill icon & the tabs, but can leave that to a follow-up PR that also encompasses the Project page. Approving to unblock.

Issue #5349, opened 2024-07-30
Add further usability to User groups


royendo:
**Is your feature request related to a problem? Please describe.**
The FR is two parts.
1) Adding a group assignment to a user invite.
```
rill user add --email username@domain.com --role viewer --group viewer_group
Error: user not found (InvalidArgument)

rill user add --email username@domain.com --role viewer
Invitation sent to ... as "viewer"

rill user add --group viewer_group --email username@domain.com
Error: user not found (InvalidArgument)
```
In the above case, the user has to accept the invitation before a group can be added which makes for a undesired behavior, having to wait after accepting the invite to be assigned a group.

---

2) Adding support for non-org users to be added to usergroups.
```
rill user add --email username@domain.com --project my-rill-project --role viewer

rill user add --group viewer_group --email username@domain.com
Error: user is not a member of the organization (InvalidArgument)
```
The above errors as the user is not part of the organization, only part of the single project.

**Describe the solution you'd like**
1. A user should be able to be invited with a group assignment so when accessing the environment, they immediately can start working.

2. A user that is not part of the organization, only a project, should also be able to be added to a usergroup, adding more functionality.

**Describe alternatives you've considered**
Not sure if there's any direct workarounds... 
1. Wait for accept, then set the groups afterwards.
2. Manually add the user /access to different projects

**Additional context**

![project-access](https://github.com/user-attachments/assets/540bd488-0f62-44f2-bb67-ff4b5d344da6)

begelundmuller:
Also see this Slack [thread](https://rilldata.slack.com/archives/CTZ8XBQ85/p1721781680976799) for more context. 

I think a good starting point here is to just add support for group assignment in the `org_invites` table.

pjain1:
> If trying to add a user to a group in an org they don't belong to, also add them to the org (with a prompt to confirm)

So an existing Rill user cannot be of an org usergroup without being assigned a role on org ? Then for non existing Rill user as well we make them part of org first and then assign them to the group ?

begelundmuller:
> So an existing Rill user cannot be part of an org usergroup without being assigned a role on org ? 

Yes – you can *only* be in a usergroup in an org if you are a member of the org.

PR #5461, opened 2024-08-16
support adding to usergroups directly

pjain1:
Fixes https://github.com/rilldata/rill/issues/5349

- For non existing Rill user, invite is sent and upon login user gets added to org as viewer and added to desired usergroup
- For existing Rill user, confirmation is asked to add user to org and upon confirmation, gets added to the desired usergroup


admin/server/usergroups.go
```diff
@@ -487,22 +488,99 @@ func (s *Server) AddUsergroupMemberUser(ctx context.Context, req *adminv1.AddUse
...
+				return nil, status.Error(codes.Internal, err.Error())
+			}
+
+			err = s.admin.DB.InsertUsergroupMemberUser(ctx, *org.AllUsergroupID, user.ID)
+			if err != nil {
+				return nil, status.Error(codes.Internal, err.Error())
+			}
+		} else {
+			return nil, status.Error(codes.FailedPrecondition, "user is not a member of the organization")
+		}
```
begelundmuller, line 583: You can reduce nesting by changing the case to `if !req.AddRillUserToOrg { ... }`. See the style guide: https://github.com/uber-go/guide/blob/master/style.md#reduce-nesting

cli/cmd/user/add.go
```diff
@@ -37,16 +39,44 @@ func AddCmd(ch *cmdutil.Helper) *cobra.Command {
...
+				res, err := client.AddUsergroupMemberUser(cmd.Context(), &adminv1.AddUsergroupMemberUserRequest{
+					Organization:     ch.Org,
+					Usergroup:        group,
+					Email:            email,
+					AddRillUserToOrg: false,
 				})
 				if err != nil {
+					if s, ok := status.FromError(err); ok && s.Code() == codes.FailedPrecondition && strings.Contains(s.Message(), "user is not a member of the org") {
+						// ask if user should be added to the organization
+						addToOrg, err := cmdutil.ConfirmPrompt(fmt.Sprintf("This user exists in Rill but not part of the %q organization. Do you want to add the user to the organization as viewer?", ch.Org), "", false)
```
begelundmuller, line 51: Would suggest phrasing it like this:
```
The user must be a member of %q to join one of its groups. Do you want to also invite the user to join %q?
```

cli/cmd/user/add.go
```diff
@@ -37,16 +39,44 @@ func AddCmd(ch *cmdutil.Helper) *cobra.Command {
...
+						if err != nil {
+							return err
+						}
+
+						if addToOrg {
```
begelundmuller, line 56: Again, can reduce nesting by handling the `if !addToOrg` case that exits early

cli/cmd/user/add.go
```diff
@@ -37,16 +39,44 @@ func AddCmd(ch *cmdutil.Helper) *cobra.Command {
...
+							_, err = client.AddUsergroupMemberUser(cmd.Context(), &adminv1.AddUsergroupMemberUserRequest{
+								Organization:     ch.Org,
+								Usergroup:        group,
+								Email:            email,
+								AddRillUserToOrg: true,
```
begelundmuller, line 61: You could also keep it simple by removing this flag, and just calling `AddOrganizationMemberUser` separately first. That would mean you don't need to handle the invite creation logic in multiple RPCs (and wouldn't need a `tx` in this RPC).

runtime/pkg/email/email.go
```diff
@@ -211,12 +212,21 @@ func (c *Client) SendOrganizationInvite(opts *OrganizationInvite) error {
 	queryParams.Add("redirect", mustJoinURLPath(opts.FrontendURL, opts.OrgName))
 	finalURL := mustJoinURLPath(opts.AdminURL, "/auth/signup") + "?" + queryParams.Encode()
 
+	var body string
+	if opts.RoleName != "" {
+		body = fmt.Sprintf("%s has invited you to join <b>%s</b> as a %s for their Rill account. Get started interacting with fast, exploratory dashboards by clicking the button below to sign in and accept your invitation.", opts.InvitedByName, opts.OrgName, opts.RoleName)
+	} else if opts.UsergroupName != "" {
+		body = fmt.Sprintf("%s has invited you to join <b>%s</b> user group for their <b>%s</b> Rill account. Get started interacting with fast, exploratory dashboards by clicking the button below to sign in and accept your invitation.", opts.InvitedByName, opts.UsergroupName, opts.OrgName)
```
begelundmuller, line 219:
```suggestion
		body = fmt.Sprintf("%s has invited you to join the <b>%s</b> group for their <b>%s</b> Rill account. Get started interacting with fast, exploratory dashboards by clicking the button below to sign in and accept your invitation.", opts.InvitedByName, opts.UsergroupName, opts.OrgName)
```

begelundmuller (changes requested):

admin/database/postgres/postgres.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/jackc/pgconn"
 	"github.com/jackc/pgtype"
 	"github.com/jmoiron/sqlx"
+	"github.com/lib/pq"
```
begelundmuller, line 15: This is the wrong Postgres library (we use `pgx`), so should be avoided

admin/database/postgres/postgres.go
```diff
@@ -1950,6 +1973,20 @@ func (m *magicAuthTokenWithUserDTO) AsModel() (*database.MagicAuthTokenWithUser,
 	return m.MagicAuthTokenWithUser, nil
 }
 
+type OrganizationInviteDTO struct {
```
begelundmuller, line 1976: nit: doesn't need to be exported, so should be `organizationInviteDTO`

admin/server/usergroups.go
```diff
@@ -487,22 +487,41 @@ func (s *Server) AddUsergroupMemberUser(ctx context.Context, req *adminv1.AddUse
...
+		invite, err := s.admin.DB.FindOrganizationInvite(ctx, org.ID, req.Email)
+		if err != nil {
+			if !errors.Is(err, database.ErrNotFound) {
+				return nil, status.Error(codes.Internal, err.Error())
+			}
+			// there is no pending invite return error
+			return nil, status.Error(codes.FailedPrecondition, "user is not a member of the organization")
+		}
+		// add group to the invite
+		invite.UsergroupIDs = append(invite.UsergroupIDs, group.ID)
```
begelundmuller, line 510: What if `group.ID` is already in `invite.UsergroupIDs`?

admin/user.go
```diff
@@ -82,6 +82,25 @@ func (s *Service) CreateOrUpdateUser(ctx context.Context, email, name, photoURL
...
+
+			if !exists {
+				// ignore if usergroup does not exist, might have been deleted before invite was accepted
+				continue
+			}
+
+			err = s.DB.InsertUsergroupMemberUser(ctx, usergroupID, user.ID)
+			if err != nil {
+				return nil, err
+			}
```
begelundmuller, line 101: Maybe for safety, gracefully handle if the user is already a member (I guess it would be a `database.ErrNotUnique`)? In case of duplicates in the list or retried calls

cli/cmd/user/add.go
```diff
@@ -37,12 +37,54 @@ func AddCmd(ch *cmdutil.Helper) *cobra.Command {
...
 				})
 				if err != nil {
+					if strings.Contains(err.Error(), "user is not a member of the org") {
+						// ask if user should be added to the organization
+						ok, err := cmdutil.ConfirmPrompt(fmt.Sprintf("The user must be a member of %q to join one of its groups. Do you want to invite the user to join %q?", ch.Org, ch.Org), "", false)
+						if err != nil {
+							return err
+						}
+
+						if ok {
```
begelundmuller, line 53: Can reduce nesting by handling the `if !ok` case that returns early

go.mod
```diff
@@ -38,6 +38,7 @@ require (
 	github.com/fatih/color v1.15.0
 	github.com/fsnotify/fsnotify v1.7.0
 	github.com/getkin/kin-openapi v0.126.0
+	github.com/go-git/go-billy/v5 v5.5.0
```
begelundmuller, line 41: Looks like an unrelated change to `go.mod`? Try running `go mod tidy`

begelundmuller (changes requested):

admin/server/usergroups.go
pjain1: oh yeah thought of that but forgot to dedupe

admin/user.go
pjain1: this is happening inside a transaction so in case of err tx fails, this needs to be checked explicitly. This is first time login case so If we deduplicate during adding the groups to invite, I think it should not be required to check ?

pjain1: since this is happening inside transaction then in case of retries if usergroup was updated then invite would have been deleted as well so should not be a problem

admin/server/usergroups.go
```diff
@@ -620,3 +641,12 @@ func memberUsergroupToPB(member *database.MemberUsergroup) *adminv1.MemberUsergr
...
 }
+
+func contains(slice []string, item string) bool {
+	for _, v := range slice {
+		if v == item {
+			return true
+		}
+	}
+	return false
+}
```
begelundmuller, line 652: You can use the standard library's `slices.Contains` function

admin/user.go
begelundmuller: Yeah I guess it's fine then :)

go.mod
begelundmuller: This line is still there? It's not referenced in this PR, so seems like a mistake?

begelundmuller (approved):

PR #5462, opened 2024-08-17
Nit: Make resource tables more readable

ericpgreen2:
This PR makes the resource tables a little more readable by moving the search bar out of the header and by adding padding.

The PR also factors out `ResourceHeader` and `ResourceTableEmpty` components to remove redundant code.

Before:
image

After:
image


lovincyrus (approved):
tested locally, looks great

PR #5465, opened 2024-08-19
feat: create table and then insert while creating model in clickhouse

k-anshul:
Fixed `CreateTableAsSelect` on clickhouse cloud.

**NOTE** : Creating a view based on external systems can fail sometimes but the workaround will be to specify columns manually which is also expected to be set in most cases.

### DIFF
--- a/runtime/drivers/clickhouse/model_manager.go
+++ b/runtime/drivers/clickhouse/model_manager.go
@@ -43,8 +43,8 @@ type ModelOutputProperties struct {
 	SampleBy string `mapstructure:"sample_by"`
 	// TTL sets ttl for column and table.
 	TTL string `mapstructure:"ttl"`
-	// Settings set the table specific settings.
-	Settings string `mapstructure:"settings"`
+	// TableSettings set the table specific settings.
+	TableSettings string `mapstructure:"table_settings"`
 	// QuerySettings sets the settings clause used in insert/create table as select queries.
 	QuerySettings string `mapstructure:"query_settings"`
 }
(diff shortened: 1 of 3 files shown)


runtime/drivers/clickhouse/model_manager.go
```diff
@@ -43,8 +43,8 @@ type ModelOutputProperties struct {
 	SampleBy string `mapstructure:"sample_by"`
 	// TTL sets ttl for column and table.
 	TTL string `mapstructure:"ttl"`
-	// Settings set the table specific settings.
-	Settings string `mapstructure:"settings"`
+	// TableSettings set the table specific settings.
+	TableSettings string `mapstructure:"table_settings"`
```
k-anshul, line 47: NOTE : This is a breaking change but should be fine. Don't think there are projects using `settings` property.

runtime/drivers/clickhouse/olap.go
```diff
@@ -224,8 +225,20 @@ func (c *connection) CreateTableAsSelect(ctx context.Context, name string, view
...
+	if outputProps.Columns == "" {
+		// infer columns
+		v := tempName("view")
+		err := c.Exec(ctx, &drivers.Statement{Query: fmt.Sprintf("CREATE OR REPLACE VIEW %s AS %s", v, sql)})
+		if err != nil {
+			return err
+		}
+		defer func() {
+			_ = c.Exec(context.Background(), &drivers.Statement{Query: "DROP VIEW " + v})
+		}()
```
begelundmuller, line 237: Can we use a ctx with some kind of default timeout here?

begelundmuller (approved):

runtime/drivers/clickhouse/olap.go
k-anshul: Sure. We can keep a 5 second timeout.

PR #5468, opened 2024-08-19
feat: Add a banner warning of assumed user

AdityaHegde:
If there is an assumed user, cli throws a warning for every command. Since we are adding some actions from UI as well, we need to show a warning in this case.

Screenshot 2024-08-19 at 5 52 47 PM


proto/rill/local/v1/api.proto
```diff
@@ -122,6 +98,7 @@ message GetCurrentUserRequest {}
 message GetCurrentUserResponse {
   admin.v1.User user = 1;
   repeated string rill_user_orgs = 2;
+  string assumed_user_email = 3;
```
begelundmuller, line 101: Would suggest `bool is_representing_user` because:
1. We use the word "represent" not "assume" internally, so this is more consistent
2. The value is the same as `user.email`, so better to make it a bool to avoid ambiguity

begelundmuller (changes requested):

begelundmuller (approved):

Issue #5361, opened 2024-07-31
Support comparisons with time as a dimension


begelundmuller:
Currently, comparisons are computed using a subquery for the base and comparison time range, which is joined on the dimension values using `IS NOT DISTINCT FROM` (happens [here](https://github.com/rilldata/rill/blob/8054854a54f58fc43093567e2a176f341b181a88/runtime/metricsview/astsql.go#L151)). This works well for comparing regular dimensions (e.g. `base.country IS NOT DISTINCT FROM comparison.country`), but it does not work if comparing with a time dimension (e.g. this never matches because the comparison time range is different: `base.time__day IS NOT DISTINCT FROM comparison.time__day`).

Currently, this is not a problem because the application only needs the data for the time charts, and it just makes two separate queries for the time range and computes the comparisons client-side. But they now want to support comparisons on the time dimension in the pivot table.

The way I would propose handling this is that if comparing by a time dimension, then it can look at the query's `time_range` and `comparison_time_range` and determine an offset between them. And then use `base.<time field> IS NOT DISTINCT FROM comparison.<time field> + INTERVAL '<offset>'` to do the comparison join.

It will probably be somewhat tricky to infer a correct comparison time offset. One approach would be to subtract the `start` time of the base time range from the comparison time range's, rounding to the time dimension's granularity (as set in `time_floor.grain`), and then using that as the join offset.

There may also be a possible approach that uses the `iso_duration` and `iso_offset` fields, but the logic around those and how they interplay between the base and comparison time ranges is slightly tricky, so that would need to be taken into account. They are also not always set, e.g. the UI often sends fixed start and end timestamps: https://github.com/rilldata/rill/blob/8054854a54f58fc43093567e2a176f341b181a88/runtime/metricsview/query.go#L117-L123

egor-ryashin:
FYI, there can be multiple time dimensions like timestamp_day, timestamp_year in a single query.

egor-ryashin:
BTW, it doesn't cover time comparison for a measure filter - how could we approach that?

begelundmuller:
> BTW, it doesn't cover time comparison for a measure filter - how could we approach that?

Can you provide an example of this issue? Not entirely sure what you mean.

begelundmuller:
Closed by https://github.com/rilldata/rill/pull/5475

PR #5475, opened 2024-08-19
Time dimension comparison for advanced measures

egor-ryashin:


runtime/metricsview/astsql.go
```diff
@@ -244,6 +262,15 @@ func (b *sqlBuilder) writeJoin(joinType JoinType, baseSelect, joinSelect *Select
...
 		rhs := b.ast.sqlForMember(joinSelect.Alias, f.Name)
+		if comp && f.Time {
+			intv, err := b.ast.interval()
+			if err != nil {
+				return err
+			}
+
+			// example: base.ts IS NOT DISTINCT FROM comparison.ts - (? - ?) -- <comparison-start> - <base-start>
+			rhs = fmt.Sprintf("(%s - INTERVAL %s MILLISECONDS)", rhs, intv)
+		}
```
begelundmuller, line 273: I think we can move this into the `ast.addTimeComparisonMeasure` function, where we could add the subtraction to the dimension expression in the comparison select instead, around here: https://github.com/rilldata/rill/blob/ec8712134a6c03c0aa5a1ef4b6b43ebbf4c24a77/runtime/metricsview/ast.go#L803-L807

This would allow us to keep the regular `IS NOT DISTINCT FROM` check here, and avoid accessing the `ast` too much from the SQL builder phase.

runtime/metricsview/astsql.go
```diff
@@ -217,7 +218,24 @@ func (b *sqlBuilder) writeSelect(n *SelectNode) error {
...
+		return "", fmt.Errorf("no start time for the offset")
+	}
+	if a.query.ComparisonTimeRange == nil {
+		return "", fmt.Errorf("no comparison time range for the offset")
+	}
+	if a.query.ComparisonTimeRange.Start.IsZero() {
+		return "", fmt.Errorf("no start time for the comparison time range")
+	}
+	diff := a.query.ComparisonTimeRange.Start.Sub(a.query.TimeRange.Start)
+	return fmt.Sprint(diff.Milliseconds()), nil
```
begelundmuller, line 235: Unfortunately it's not sufficient to do millisecond delta comparisons. For example, if the time granularity is monthly, the delta will be different for each month, so you need to use `INTERVAL 1 MONTH` in SQL instead of microseconds.

It may need to scan the dimensions for the biggest granularity used in `TimeFloor` and use that to determine the interval type. Also, see the issue description for some other thoughts: https://github.com/rilldata/rill/issues/5361

begelundmuller (changes requested):

runtime/metricsview/astsql.go
egor-ryashin: Done.

runtime/metricsview/ast.go
```diff
@@ -226,6 +228,14 @@ func NewAST(mv *runtimev1.MetricsViewSpec, sec *runtime.ResolvedSecurity, qry *Q
 	return ast, nil
 }
 
+func (a *AST) isTime(qd Dimension) bool {
```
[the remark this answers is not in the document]
egor-ryashin: Done.

runtime/metricsview/astsql.go
```diff
@@ -217,6 +219,33 @@ func (b *sqlBuilder) writeSelect(n *SelectNode) error {
...
+	}
+	if a.query.ComparisonTimeRange.Start.IsZero() {
+		return "", fmt.Errorf("no start time for the comparison time range")
+	}
+	start2 = a.query.ComparisonTimeRange.Start
+	if g == TimeGrainUnspecified {
+		g = TimeGrainMillisecond // todo millis won't work for druid
+		return a.dialect.DateDiff(string(g), start1, start2)
+	} else if g == mg {
+		return a.dialect.DateDiff(string(g), start1, start2)
```
begelundmuller, line 243: Should use the dialect's `ConvertToDateTruncSpecifier`

runtime/metricsview/ast.go
```diff
@@ -804,6 +833,22 @@ func (a *AST) addTimeComparisonMeasure(n *SelectNode, m *runtimev1.MetricsViewSp
 		if err != nil {
 			return err
 		}
+
+		for i, f := range csn.DimFields {
```
begelundmuller, line 837: Can you add a short comment explaining why this is needed

runtime/metricsview/ast.go
```diff
@@ -134,16 +141,30 @@ func NewAST(mv *runtimev1.MetricsViewSpec, sec *runtime.ResolvedSecurity, qry *Q
...
+				Time:      tm,
+				TimeGrain: gn,
+				MinGrain:  gn,
+			},
+		}
+		if f.TimeProps.Time {
+			timeFields = append(timeFields, f)
+			if gn.isLess(minGrain) {
+				minGrain = gn
+			}
```
begelundmuller, line 167: In the interest of keeping the `NewAST` function simple, can we move the logic of computing the grain to `addTimeComparisonMeasure`? It's very efficient and unlikely to get called a lot, so should be fine in terms of performance.

Then I think the `FieldNode.TimeProps` can also be turned into just `Time bool`.

runtime/drivers/olap.go
```diff
@@ -450,6 +450,20 @@ func (d Dialect) DateTruncExpr(dim *runtimev1.MetricsViewSpec_DimensionV2, grain
 	}
 }
 
+// unit - ie second, minute, ...
+func (d Dialect) DateDiff(unit string, t1, t2 time.Time) (string, error) {
```
begelundmuller, line 454: Might make sense to accept `runtimev1.TimeGrain` here instead and call `d.ConvertToDateTruncSpecifier` internally

runtime/metricsview/query.go
```diff
@@ -203,6 +203,10 @@ func (t TimeGrain) Valid() bool {
 	return false
 }
 
+func (t TimeGrain) isLess(tg TimeGrain) bool {
```
begelundmuller, line 206: Not sure if the verb "less" makes sense in the context of grains. Maybe is `IsSmaller` instead?

begelundmuller:
Also note the failing CI

egor-ryashin:
All done.

begelundmuller:
@egor-ryashin There are still CI failures on the PR – can you take a look at them?

egor-ryashin:
Done.

begelundmuller (changes requested):
@egor-ryashin Looking at the time adjustment logic, it's not clear to me that it handles the case where two levels of granularities are present and the base and comparison time ranges cross boundaries of both granularities.

For example, if you group by granularities `year` and `month`, and have `timeRange: {start: "2024-02-01", end: "2024-05-01"}` and `comparisonTimeRange: {start: "2023-11-01", end: "2024-02-01"}` (three months offset), I wonder if it will errorneously adjust the value `2024-01-01` in the comparison time range to `2025-01-01` in the year-granularity field? Can you try adding a test case for that and verify if that's the case or not?

egor-ryashin:
Is it different from this unit-test?
```
func TestMetricsViewsAggregation_comparison_time_dim(t *testing.T) {
	rt, instanceID := testruntime.NewInstanceForProject(t, "ad_bids")

	limit := int64(10)
	q := &queries.MetricsViewAggregation{
		MetricsViewName: "ad_bids_metrics",
		Dimensions: []*runtimev1.MetricsViewAggregationDimension{
			{
				Name: "pub",
			},
			{
				Name: "dom",
			},

			{
				Name:      "timestamp",
				TimeGrain: runtimev1.TimeGrain_TIME_GRAIN_DAY,
			},
			{
				Name:      "timestamp",
				TimeGrain: runtimev1.TimeGrain_TIME_GRAIN_YEAR,
				Alias:     "timestamp_year",
			},
		},
		Measures: []*runtimev1.MetricsViewAggregationMeasure{
			{
				Name: "measure_0",
			},
			{
				Name: "measure_1",
			},
			{
				Name: "m1",
			},
			{
				Name: "measure_0__p",
				Compute: &runtimev1.MetricsViewAggregationMeasure_ComparisonValue{
					ComparisonValue: &runtimev1.MetricsViewAggregationMeasureComputeComparisonValue{
						Measure: "measure_0",
					},
				},
			},
		},
		Where: expressionpb.AndAll(
			expressionpb.Eq("pub", "Google"),
			expressionpb.Eq("dom", "news.google.com"),
		),
		Having: expressionpb.Gt("measure_0__p", 0.0),
		Sort: []*runtimev1.MetricsViewAggregationSort{
			{
				Name: "pub",
			},
			{
				Name: "dom",
			},
			{
				Name: "timestamp",
			},
			{
				Name: "timestamp_year",
			},
			{
				Name: "measure_1",
			},
		},

		TimeRange: &runtimev1.TimeRange{
			Start: timestamppb.New(time.Date(2022, 1, 1, 0, 0, 0, 0, time.UTC)),
			End:   timestamppb.New(time.Date(2022, 1, 3, 0, 0, 0, 0, time.UTC)),
		},
		ComparisonTimeRange: &runtimev1.TimeRange{
			Start: timestamppb.New(time.Date(2022, 1, 3, 0, 0, 0, 0, time.UTC)),
			End:   timestamppb.New(time.Date(2022, 1, 5, 0, 0, 0, 0, time.UTC)),
	
```

begelundmuller:
> Is it different from this unit-test?

Yeah because the unit test doesn’t cross a year-boundary in the comparison time range case.

begelundmuller:
> I found out that we hide the reference to the timestamp column here:

The time dimension is captured in the `Compute.TimeFloor` a few lines below:
```go
		if d.TimeGrain != runtimev1.TimeGrain_TIME_GRAIN_UNSPECIFIED {
			res.Compute = &metricsview.DimensionCompute{
				TimeFloor: &metricsview.DimensionComputeTimeFloor{
					Dimension: d.Name,
					Grain:     metricsview.TimeGrainFromProto(d.TimeGrain),
				},
			}
		}
```

Isn't that sufficient to incorporate it correctly into the join?

egor-ryashin:
We can have non-primary timestamp columns (which should be in join as is) I cannot find where we distinguish them - you mean they don't have time-grains, right?

begelundmuller:
I think both of the bugs identified above can be fixed by generating a separate `ast.comparisonDimFields` property alongside `ast.dimFields` and using that for comparisons. That would allow both:
- Adding the interval to the time dimension expression before wrapping it with `date_trunc` (i.e. the expression becomes `date_trunc(<time dim> + INTERVAL ...)` instead of `date_trunc(<time dim>) + INTERVAL`. That would address the first bug identified above.
- Accessing the non-aliased time dimension field name, so it can correctly add the interval *only* for the metrics view's primary time dimension.

egor-ryashin:
Done.

runtime/metricsview/ast.go
```diff
@@ -1101,6 +1148,53 @@ func (a *AST) sqlForAnyInGroup(expr string) string {
...
+		expr := fmt.Sprintf("(%s - INTERVAL (%s) %s)", tc, dateDiff, a.dialect.ConvertToDateTruncSpecifier(mg.ToProto()))
+		dim := &runtimev1.MetricsViewSpec_DimensionV2{
+			Expression: expr,
+		}
+		expr, err = a.dialect.DateTruncExpr(dim, g.ToProto(), a.query.TimeZone, int(a.metricsView.FirstDayOfWeek), int(a.metricsView.FirstMonthOfYear))
+		if err != nil {
+			return "", fmt.Errorf(`failed to compute time floor: %w`, err)
+		}
+		return expr, nil
+	}
```
begelundmuller, line 1193: This case completely discards the input `expr`, and changes to using `a.metricsView.TimeDimension`. What if the input `expr` was not that of the `a.metricsView.TimeDimension` (like when we support multiple time dimensions, as previously discussed)? This could be fixed by moving the handling to `resolveDim` and returning a `comparisonExpression` from there or something like that. Or at least there should be a `// TODO` here explaining the inconsistency.

runtime/queries/metricsview_aggregation_test.go
```diff
@@ -2292,6 +2293,297 @@ measures:
...
+		},
+
+		TimeRange: &runtimev1.TimeRange{
+			Start: timestamppb.New(time.Date(2022, 1, 1, 23, 0, 0, 0, time.UTC)),
+			End:   timestamppb.New(time.Date(2022, 1, 2, 2, 0, 0, 0, time.UTC)),
+		},
+		ComparisonTimeRange: &runtimev1.TimeRange{
+			Start: timestamppb.New(time.Date(2022, 1, 2, 1, 0, 0, 0, time.UTC)),
+			End:   timestamppb.New(time.Date(2022, 1, 2, 3, 0, 0, 0, time.UTC)),
+		},
```
begelundmuller, line 2554: I don't think this test case addresses the concern described in this comment: https://github.com/rilldata/rill/pull/5475#pullrequestreview-2266024728?

The goal is not to test overlapping base and comparison time ranges. The goal is to test that if the comparison time range has `start: "2023-11-01"` and the base time range has `start: "2024-02-01"`, then it should correct the comparison values for e.g. `2024-01-15` to `2024-04-15`, not `2025-04-15` (accidentally adding a year, which is the year-interval datediff between the two start times).

begelundmuller (changes requested):

runtime/metricsview/ast.go
egor-ryashin: I expect `a.metricsView.TimeDimension` contains the required time column from a list of possible time columns.
From that point of view it will work.

runtime/queries/metricsview_aggregation_test.go
egor-ryashin: We fixed the bug where we didn't subtract anything from larger grain times. I abstract language:
```
Assuming format yy-mm and base and comparison values:
base: [23-12 24-01 24-02]
 base_year: [23 24 24]

comparison_month: [24-02 24-03 24-05]
  comparison_year: [24 24 24]

date_diff(23-12,24-02,month) = 2
new_comparison = comparison - date_diff
new_comparison_year = date_trunc('year', comparison - date_diff))

Subtracting date diff from comparison:
new_comparison: [23-12 24-01 24-02]
  new_comparison_year: [23 24 24]
```
The same will work for hour and day and any other combinations. Not sure I'm following.

runtime/metricsview/ast.go
begelundmuller: I think at least a comment about that would still be nice – it's pretty convoluted logic

runtime/queries/metricsview_aggregation_test.go
egor-ryashin: Oh, I see what you mean, let me check.

egor-ryashin: It's irrelevant, we don't check that record, it's a test typo - fixed.

runtime/metricsview/ast.go
egor-ryashin: Done.

begelundmuller (approved):

PR #5489, opened 2024-08-20
remove storage limit checks

k-anshul:


begelundmuller (approved):
Always nice to remove code :)

PR #5491, opened 2024-08-20
feat: General organization settings page

AdityaHegde:
- [x] Adds the `General` organization settings page.
- [x] Adds an item to edit organization name and description.
- [x] Adds an item to delete an organization.
- [x] Adds an item to add/remove domain whitelisting.
- [x] Add `displayName` to org.
- [x] Add loading and error states.


proto/rill/admin/v1/api.proto
```diff
@@ -807,6 +807,7 @@ message UpdateOrganizationRequest {
   string name = 1 [(validate.rules).string.min_len = 1];
   optional string description = 2;
   optional string new_name = 3;
+  optional string new_display_name = 5;
```
begelundmuller, line 810: This should just be `display_name` (like `description` and `billing_email`). The only reason `new_name` has a `new_` prefix is to disambiguate it from `name`, which is used to identify the org to edit.

begelundmuller (changes requested):
The other backend changes look good

web-admin/src/features/organizations/settings/UpdateOrgItem.svelte
```diff
@@ -0,0 +1,145 @@
...
+    on:submit|preventDefault={submit}
+    class="update-org-form"
+    use:enhance
+  >
+    <Input
+      bind:value={$form.name}
+      errors={$errors?.name}
+      id="name"
+      label="Name"
+      description={`Your org URL will be https://ui.rilldata.com/${sanitizeOrgName($form.name)}, to comply with our naming rules.`}
```
ericpgreen2, line 114: Nit: it might save some confusion in the dev/test/stage environments if you used the `RILL_ADMIN_FRONTEND_URL`

web-admin/src/features/organizations/settings/UserDomainAllowListItem.svelte
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    void queryClient.refetchQueries(
+      getAdminServiceListWhitelistedDomainsQueryKey(organization),
+    );
+  }
+</script>
+
+<!-- hide if user's domain is not public and no domains are added -->
+{#if $isPublicDomain.data || $allowedDomains.data?.domains?.length}
```
ericpgreen2, line 53: We should hide if the domain _is_ public, right?

ericpgreen2 (approved):
Nice! Approving with comments.

UXQA - 
1. The settings page does not accommodate a smaller screen size. The modules spill over to the right-hand-side.
    image
2. Clicking on the domain access toggle sometimes results in an extra back-and-forth flicker: [JAM](https://jam.dev/c/099916f2-4a26-4995-a166-ca223b55b989).

Naming nit: it might be cleaner to drop the "Item" from the component names: `SettingsItemContainer.svelte` (which could be `SettingContainer.svelte`), `UpdateOrgItem.svelte`, `DeleteOrgItems.svelte` (which itself shouldn't be plural "items"), `UserDomainAllowlistItem.svelte`.

web-admin/src/features/organizations/settings/UserDomainAllowListItem.svelte
AdityaHegde: Good point. The design changed to always show this part. So lemme update

web-admin/src/features/organizations/settings/UpdateOrgItem.svelte
```diff
@@ -0,0 +1,145 @@
...
+    $orgResp.data?.organization?.description !== $form.description;
+
+  $: error = parseUpdateOrgError(
+    $updateOrgMutation.error as unknown as AxiosError<RpcStatus>,
+  );
+</script>
+
+<SettingsItemContainer title="Organization">
+  <form
+    slot="description"
```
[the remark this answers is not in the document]
AdityaHegde: I will rename it to `body`. Makes more sense i guess

begelundmuller (commented):
Backend changes look good!

begelundmuller (approved):

PR #5497, opened 2024-08-21
fix: retain pivot pill order on refresh

### LINKED ISSUE #5487: Pivot: refreshing the page re-orders pills
To reproduce:
1. Go to this demo dashboard: https://ui.rilldata.com/demo/rill-openrtb-prog-ads/auction
2. Create a pivot table that looks like:
    image
3. Refresh the page
4. In the pivot's "Rows", see that the time pill has switched places with the "App Site Domain" pill:
    image

djbarnwal:
Fixes #5487

### DIFF
--- a/proto/rill/ui/v1/dashboard.proto
+++ b/proto/rill/ui/v1/dashboard.proto
@@ -72,62 +72,58 @@ message DashboardState {
   // Dimension filters applied
   rill.runtime.v1.MetricsViewFilter filters = 2;
   // Expression format for dimension filters
-  rill.runtime.v1.Expression where = 20;
+  rill.runtime.v1.Expression where = 3;
   // Expression format for measure filters
-  repeated DashboardDimensionFilter having = 21;
+  repeated DashboardDimensionFilter having = 4;
   // Selected time granularity
-  rill.runtime.v1.TimeGrain time_grain = 3;
+  rill.runtime.v1.TimeGrain time_grain = 5;
 
-  DashboardTimeRange compare_time_range = 4;
+  DashboardTimeRange compare_time_range = 6;
 
   // Determines the active page. This is cleaner than using setting `selected_dimension` or `expanded_measure`
-  ActivePage active_page = 32;
+  ActivePage active_page = 7;
 
   // Selected measure for the leaderboard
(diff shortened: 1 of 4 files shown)


proto/rill/ui/v1/dashboard.proto
```diff
@@ -72,62 +72,58 @@ message DashboardState {
   // Dimension filters applied
   rill.runtime.v1.MetricsViewFilter filters = 2;
   // Expression format for dimension filters
-  rill.runtime.v1.Expression where = 20;
+  rill.runtime.v1.Expression where = 3;
```
AdityaHegde, line 75: Changing the index will break backwards compatibility. Any specific reason to do this?

web-common/src/features/dashboards/proto-state/toProto.ts
```diff
@@ -278,20 +278,41 @@ function toPivotProto(pivotState: PivotState): PartialMessage<DashboardState> {
     };
   return {
     pivotIsActive: true,
+    pivotRowDimensions: pivotState.rows.dimension.map((d) => {
+      if (d.type === PivotChipType.Dimension) {
```
AdityaHegde, line 282: Would be good to break the mapper out reduce the complexity here, similar to how we do in `fromProto`

proto/rill/ui/v1/dashboard.proto
```diff
@@ -72,62 +72,58 @@ message DashboardState {
...
+  optional bool pivot_is_active = 24;
   // List of dimensions selected for rows
-  repeated string pivot_row_dimensions = 24;
-  // List of time dimensions selected for columns
-  repeated rill.runtime.v1.TimeGrain pivot_column_time_dimensions = 25;
-  // List of time dimensions selected for columns
-  repeated string pivot_column_dimensions = 26;
+  repeated PivotElement pivot_row_dimensions = 25;
+  // List dimensions selected for columns
+  repeated PivotElement pivot_column_dimensions = 26;
```
AdityaHegde, line 126: We should add these at the end index and add backwards compatibility for previous fields

AdityaHegde (changes requested):

AdityaHegde (approved):

Issue #5476, opened 2024-08-19
Custom domains support


begelundmuller:
See details here: https://www.notion.so/rilldata/Custom-domains-implementation-ideas-16c803f925334a299da811efd13045aa

Task list:
- [x] Add a `custom_domain` column on the `orgs` table
- [x] Add CLI commands for setting a custom domain under `rill sudo`
- [x] Add a `GetOrgForDomain` RPC for use by the frontend
- [x] Update the `authLoginCallback` handler to check if the frontend redirect URL contains a custom domain, and if yes, redirect to the `authWithToken` handler on the custom domain's API endpoint to correctly set the auth cookie (must support a redirect URL in `authWithToken` as well)
- [x] Update the UI to query the admin service on the custom domain if configured
- [x] Test in the staging environment
- [x] Audit internally generated `admin` endpoints for places where we need to prefix with `/api`

begelundmuller:
Closing as completed by the two PRs linked above

PR #5498, opened 2024-08-21
Support custom domains in the `admin` service

begelundmuller:
Behavior changes:

- Adds an RPC `GetOrganizationNameForCustomDomain` for finding the org for a custom domain. The application can use this to find the org to show when it is loaded on a non-default domain (by checking `window.origin`).
- Adds a sudo CLI command for setting custom domains: `rill sudo org set-custom-domain example-org example.com`
- Adds a sudo CLI command for getting all metadata about an org: `rill sudo org show example-org`

Internal changes:

- Adds a `custom_domain` column on the `orgs` table.
- Adds an `admin.URLs` type that centralizes external URL building in the `admin` service to make it easier to build redirect URLs with correct handling of custom domains.
- Updates the `/auth` endpoints to set auth cookies on the custom domain instead if that's where they're redirecting back to.

Contributes to https://github.com/rilldata/rill/issues/5476


kaspersjo (approved):
LGTM!

PR #5499, opened 2024-08-21
feat: Enhance banner component

AdityaHegde:
To support variants used in billing this PR enhances the banner component.
1. Add `type` that changes background and font colors.
2. Add `iconType`
3. Add `ctaText`, `ctaUrl`, `ctaCallback` Will be used to open things like pylon etc


web-common/src/lib/event-bus/events.ts
```diff
@@ -16,6 +16,17 @@ interface NotificationOptions {
...
+
+  iconType: "none" | "alert" | "check" | "sleep" | "loading";
+
+  // cta abstraction
+  ctaText?: string;
+  // if it is a direct link
+  ctaUrl?: string;
+  ctaTarget?: string;
+  // callback when we need to take action like open pylon
+  ctaCallback?: () => void;
```
ericpgreen2, line 31: At risk of over-abstraction, it might be cleaner to have a `cta` object to encapsulate its attributes, and it could make for better typing. But could go either way.

```
cta?: {
  type: "button" | "link";
  text: string;
  url?: string; // Only applicable if type is "link"
  target?: string; // Only applicable if type is "link"
  callback?: () => void; // Only applicable if type is "button"
};
```

web-common/src/components/banner/Banner.svelte
```diff
@@ -1,18 +1,42 @@
 <script lang="ts">
   import { BannerMessage } from "../../lib/event-bus/events";
   import AlertCircleIcon from "../icons/AlertCircleOutline.svelte";
+  import CheckCircleOutline from "../icons/CheckCircleOutline.svelte";
+  import LoadingCircleOutline from "../icons/LoadingCircleOutline.svelte";
+  import MoonCircleOutline from "../icons/MoonCircleOutline.svelte";
 
   export let banner: BannerMessage;
+
+  $: ctaElement = banner.ctaUrl ? "a" : "span";
```
ericpgreen2, line 10: `button` instead of `span`, right?

web-common/src/lib/event-bus/events.ts
```diff
@@ -16,6 +16,17 @@ interface NotificationOptions {
 }
 
 export interface BannerMessage {
+  type: "default" | "success" | "info" | "warning" | "error";
   message: string;
   includesHtml?: boolean;
+
+  iconType: "none" | "alert" | "check" | "sleep" | "loading";
```
ericpgreen2, line 23: I think the icon should primarily be a derivative of the `type`. For example, by default, an `info` type should have an `info` icon, and the `warning` and `error` types should have an `alert` icon. Otherwise, we'll risk inconsistency. But we could have an optional `iconOverride` to handle the special cases.

web-admin/src/routes/[organization]/[project]/-/share/[token]/+page.svelte
```diff
@@ -21,8 +21,10 @@
   $: ({ data: cookieProject } = $cookieProjectQuery);
   $: if (cookieProject) {
     eventBus.emit("banner", {
+      type: "default",
       message: `Limited view. For full access and features, visit the <a href='/${organization}/${project}/${metricsView}'>original dashboard</a>.`,
       includesHtml: true,
+      iconType: "alert",
```
Comment (ericpgreen2) [inline web-admin/src/routes/[organization]/[project]/-/share/[token]/+page.svelte:27]: This is more an informational message than a warning, so I think this should be an "info" icon. (See my other message about default icons for each `type`.)

ericpgreen2 (approved):
Approving with comments.

web-common/src/lib/event-bus/events.ts
AdityaHegde: Ya I think this will be better

web-common/src/lib/event-bus/events.ts
AdityaHegde: These are meant to map to the icon name. Alert icon is actually the one used the most. Since there is enough usages that breaks these defaults, there is no point in defaults.

web-admin/src/routes/[organization]/[project]/-/share/[token]/+page.svelte
Comment (AdityaHegde) [reply web-admin/src/routes/[organization]/[project]/-/share/[token]/+page.svelte:27]: The file had `AlertCircleOutline`. Since icon type match the file name, alert makes sense here.

PR #5501, opened 2024-08-21
feat: display underlying raw data for aggregated pivot cell

### LINKED ISSUE #4328: Add support for data viewer in pivot table
When using the pivot table, customers may be interested in retrieving the composite rows / underlying raw data for a particular cell. This will be especially useful for finance use case so the underlying numbers can be reviewed / audited. 

Additional context: https://rilldata.slack.com/archives/C04UY5NTXA9/p1710254899876739

djbarnwal:
Closes #4328


ericpgreen2:
Sweet! Probably merits a tooltip & a visual indication for the active cell. Maybe @jkhwu has input.

djbarnwal:
Addressed the review comments

>     * if the pivot table is empty, rather than showing any data, I think I'd expect the row viewer to say something like "no active cell"

For this do we want to only show data when we have an active cell or even if we have a valid pivot table but no active cell.

nishantmonu51:
When we have a valid pivot table and no active cell, we can show the data for all cells by just applying global filters. It would be consistent with how it works in explore as well. In order to make it clear visually indicate the difference in the table header.

ericpgreen2:
Styling nits:
- There should still be an indicator for the active cell when the row as a whole is highlighted. In the following pic, after clicking on the "Above the fold > Total Records" cell, the active cell highlight didn't show up because the row highlight was overpowering it.
    image
- To indicate what the active cell will be, we should highlight the cell on hover (in addition to the row as a whole).

djbarnwal:
Made the changes

With active cell
image

Without active cell or empty pivot
image

For explore, the header is same as before

ericpgreen2:
Nice, the on:hover styling feels great now!

Some more QA:
1. I'm not seeing the same cell interactions in Rill Cloud as I am in Rill Developer. (For a project where I've enabled the `cloudDataViewer`.)
2. It seems we're too aggressively clearing the active cell state. For example, when I click on a cell, drag the Model Data section up/down, the active cell is lost.
3. When the Model Data section is open & dragged to a custom height, I wouldn't expect that clicking on another active cell would reset the height to its default.
4. It's not totally clear that "for selected global filters (and highlighted pivot cell)" refers to the numerator not the denominator. IMO we should remove that italicized text and instead reconsider the "Model Data" copy.

web-common/src/features/dashboards/rows-viewer/RowsViewerAccordion.svelte
```diff
@@ -1,54 +1,99 @@
...
+    end: $timeControlsStore.timeEnd,
+  };
+  $: if (showPivot && activeCellFilters) {
+    if (!isOpen && !manualClose) {
+      height = PIVOT_HEIGHT_EXPANDED;
+      isOpen = true;
+    }
+    filters = sanitiseExpression(activeCellFilters.filters, undefined);
+    timeRange = activeCellFilters.timeRange;
+    label = "for selected global filters and highlited pivot cell";
```
ericpgreen2, line 66: "highlited" -> "highlighted"

web-common/src/features/dashboards/rows-viewer/row-viewer-store.ts
```diff
@@ -0,0 +1,3 @@
+import { writable } from "svelte/store";
+
+export const rowViewerStore = writable<HTMLElement | null>(null);
```
ericpgreen2, line 3: Mind just adding a comment to explain why this is needed?

ericpgreen2 (approved):
Looks good! Approving.

Just the one thing I still think should be reconsidered is the "for selected global filters (and highlighted pivot cell)" text. @ericokuma @mindspank @jkhwu any feedback?

djbarnwal:
Synced up with @jkhwu and we decided to use `Model Data` and `Model data for selected cell` terminology when a cell is highlighted.

PR #5505, opened 2024-08-22
Add slots and provisioner as annotations

himadrisingh:


admin/deployments.go
```diff
@@ -452,6 +452,8 @@ func (s *Service) NewDeploymentAnnotations(org *database.Organization, proj *dat
 		orgName:         org.Name,
 		projID:          proj.ID,
 		projName:        proj.Name,
+		projProdSlots:   "slot-" + fmt.Sprint(proj.ProdSlots),
```
kaspersjo, line 455: Let's just make sure the provisioner templates add `""` to all the annotation labels, updated here https://github.com/rilldata/rill-helm-charts/commit/b3a605959c31614d456eeea975155a5cd793829e

kaspersjo (changes requested):

kaspersjo (approved):

PR #5511, opened 2024-08-23
feat : cluster and direct SQL support for clickhouse models

k-anshul:
closes https://github.com/rilldata/rill-private-issues/issues/592


runtime/drivers/clickhouse/model_manager.go
```diff
@@ -47,21 +54,41 @@ type ModelOutputProperties struct {
 	TableSettings string `mapstructure:"table_settings"`
 	// QuerySettings sets the settings clause used in insert/create table as select queries.
 	QuerySettings string `mapstructure:"query_settings"`
+	// DistributedConfig is config for distributed table.
+	// Note: the table name in config should be table_local. Optional.
+	// TODO :: How to handle staged changes ?
+	DistributedConfig string `mapstructure:"distributed_config"`
```
k-anshul, line 60: Hey @begelundmuller 

We need to define a way to allow user to set `distributed_config`.
The challenge is that one of the parameter in the distributed engine is name of local table ( `Distributed(cluster, database, **local_table**[, sharding_key[, policy_name]])` )which in case of staging tables is different from actual table name. I can see following options now. Also the staged changes is only enabled on cloud I believe so user can't choose `__rill_tmp_xxx`. This also exposes implementation details.

Few suggestions to handle but please let me know in case you find something more convenient for user.
1. User sets a custom key like `table_local` in all cases which we replace.
2. We dome some string processing to replace the third parameter but can get complicated.

begelundmuller: Let's discuss it sync and then follow up here.

begelundmuller (approved):

PR #5514, opened 2024-08-23
Web: Use window.location or rel=external to navigate to /auth and /github

begelundmuller:
This is necessary to prevent Svelte from doing SPA routing when navigating to the /api/auth and /api/github paths on orgs with a custom domain configured. (When a custom domain is configured, the admin service is hosted on /api of the same domain, not a different host.)

### DIFF
--- a/web-admin/src/client/redirect-utils.ts
+++ b/web-admin/src/client/redirect-utils.ts
@@ -16,6 +16,7 @@ export function redirectToGithubLogin(remote: string) {
 }
 
 function buildLoginUrl() {
+  // The backend requires that we always use the canonical admin URL for redirects to /auth/login.
   const u = new URL(CANONICAL_ADMIN_URL);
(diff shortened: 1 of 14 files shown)


web-admin/src/client/redirect-utils.ts
```diff
@@ -0,0 +1,41 @@
...
+export function redirectToLogout() {
+  window.location.href = buildLogoutUrl();
+}
+
+export function redirectToGithubLogin(remote: string) {
+  window.location.href = buildGithubLoginUrl(remote);
+}
+
+function buildLoginUrl() {
+  const u = new URL(CANONICAL_ADMIN_URL);
```
begelundmuller, line 19:
```suggestion
  // The backend requires that we always use the canonical admin URL for redirects to /auth/login.
  const u = new URL(CANONICAL_ADMIN_URL);
```

AdityaHegde (approved):

PR #5516, opened 2024-08-23
feat: add support for linux aarch64

Kavinjsir:
Resolve:
https://rilldata.slack.com/archives/C01190F1R1D/p1724443548781379

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -64,6 +64,25 @@ builds:
     ldflags:
       - -s -w -X main.Version={{.Version}} -X main.Commit={{.ShortCommit}} -X main.BuildDate={{.Date}}
 
+  - id: linux_aarch64
+    main: ./cli/main.go
+    binary: rill
+    goos:
+      - linux
+    goarch:
+      - arm64
+    env:
+      - CC=aarch64-linux-gnu-gcc
+      - CXX=aarch64-linux-gnu-g++
+      - CGO_CFLAGS=--sysroot=/sysroot/linux/arm64
+      - CGO_LDFLAGS=--sysroot=/sysroot/linux/arm64
+      - PKG_CONFIG_SYSROOT_DIR=/sysroot/linux/arm64
+      - PKG_CONFIG_PATH=/sysroot/linux/arm64/usr/local/lib/pkgconfig
+    flags:
+      - "-mod=readonly"
+    ldflags:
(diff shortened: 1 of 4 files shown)


.goreleaser.yaml
```diff
@@ -64,6 +64,25 @@ builds:
     ldflags:
       - -s -w -X main.Version={{.Version}} -X main.Commit={{.ShortCommit}} -X main.BuildDate={{.Date}}
 
+  - id: linux_aarch64
+    main: ./cli/main.go
+    binary: rill
+    goos:
+      - linux
+    goarch:
+      - arm64
```
Kavinjsir, line 73: Looks like go env didn't provide literal arch name as `aarch`, so I'd assume `arm64` might be fit in this case?

himadrisingh:
can we try running a nightly/snapshot from this branch and test it out?

.goreleaser.yaml
begelundmuller: Yeah, `aarch64` refers to the same architecture as `arm64`

begelundmuller:
Here's a PR on top of this PR that adds DuckDB extension embedding for the new build target: https://github.com/rilldata/rill/pull/5519.

scripts/install.sh
```diff
@@ -10,6 +10,8 @@ initPlatform() {
         PLATFORM="darwin_amd64"
     elif [ "$OS" = "linux" ] && [ "$ARCH" = "x86_64" ]; then
         PLATFORM="linux_amd64"
+    elif [ "$OS" = "linux" ] && [ "$ARCH" = "arm64" ]; then
```
kaspersjo, line 13: I think on a lot of modern arm64 linux system `ARCH` will resolve to `aarch64` not `arm64`, so we might need to add support for both cases here

kaspersjo (changes requested):

Kavinjsir:
> can we try running a nightly/snapshot from this branch and test it out?

Sounds good! Schedule an action for a nightly release:
https://github.com/rilldata/rill/actions/runs/10569133157

scripts/install.sh
Kavinjsir: Good catch! Just updated the code to:
```bash
    elif [ "$OS" = "linux" ] && { [ "$ARCH" = "arm64" ] || [ "$ARCH" = "aarch64" ]; }; then
        PLATFORM="linux_arm64"
```

himadrisingh (approved):

himadrisingh (approved):

kaspersjo (approved):

Issue #5410, opened 2024-08-07
Ability to send alerts to project admins on reconcile failures


nishantmonu51:
When a reconcile fails due to an ingestion error, we need the ability to send an alert to project admins so that they can take necessary action to fix the project.

begelundmuller:
An idea for solving this is by adding a resolver for the a project's reconcile status, which could be used to configure an alert. Something like this:
```yaml
type: alert
refresh:
  cron: 0 * * * *
data:
  project_errors: {}
notify:
  email:
    recipients: [benjamin@example.org]

PR #5525, opened 2024-08-26
Implement `resource_status` resolver for alerts on project errors

begelundmuller:
This PR implements a new resolver `resource_status`, which can be used to configure an alert that triggers when one or more project resources have an error.

Example usage:
```yaml
# alerts/project_errors.yaml
type: alert

# Check the alert every 10 minutes.
refresh:
  cron: "*/10 * * * *"

# Get all resources with a reconcile error.
# The alert will trigger if this returns any rows.
data:
  resource_status:
    where_error: true

# Send notifications by email
notify:
  email:
    recipients: [benjamin@example.com]
```

Closes https://github.com/rilldata/rill/issues/5410


runtime/resolvers/resource_status.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	if r.props.WhereError {
+		// In-place trim resources that are not in error state
+		i := 0
+		for i < len(rs) {
+			r := rs[i]
+			if r.Meta.ReconcileError == "" {
+				// Remove from the slice
+				rs[i] = rs[len(rs)-1]
```
k-anshul, line 96: We can consider `slices.Delete` ?

k-anshul (approved):
Code wise looks good to me. 
But did we also consider configuring this kind of alert via `rill.yaml` or UI ? By default we could enable this for everybody where alerts are sent to all project admins. If users want they can configure recipients via `rill.yaml` or some CLI command or via UI (Any option can be taken).
Internally that may still get resolved like how it is getting resolved now. 
Here are my reasons why I would prefer that.
1. As of now user can configure an alert via UI and there is no resource definition visible to the user. 
2. The `data` seems like an implementation detail. `refresh_schedule` can be kept constant and it should be okay to tell users that emails on reconcile errors are triggered with a delay of maximum 10 minutes.
3. This alert is project specific and the other alerts are dashboard specific. With this approach a user can create multiple similar alerts for same project. Though it is a user mistake and easy to fix/avoid but can be avoided systematically as well.

runtime/resolvers/resource_status.go
begelundmuller: Unfortunately `slices.Delete` maintains order and is therefore much slower than this, especially for repeated invocations. (Larger projects may end up with 1000s of alerts and reports, so performance is if not hugely important, at least somewhat a factor here.)

begelundmuller:
> But did we also consider configuring this kind of alert via rill.yaml or UI ? By default we could enable this for everybody where alerts are sent to all project admins. If users want they can configure recipients via rill.yaml or some CLI command or via UI (Any option can be taken). Internally that may still get resolved like how it is getting resolved now. Here are my reasons why I would prefer that.

Yes, I think likely we will want to add a UI path for this at some point. The current issue was just about adding the capability (the new resolver can be used in the `admin` service's `CreateAlert` RPC), and then UI support can be prioritized separately (will probably also involve some design). 

I'm not sure I would favor a `rill.yaml` approach because there's sufficient room for adaptation here (e.g. using Slack instead of email, supporting only checking errors for specific resource types, etc.) that I think the few extra lines to be explicit about the resolver is okay.

PR #5540, opened 2024-08-28
fix: User invite not working for a single pill

AdityaHegde:
1. Fix user invite when there is a single entry and it is a pill.
2. Remove domain whitelist message when there are no domains added.
3. In user invite page, `Continue` => `Skip`

### DIFF
--- a/web-admin/src/features/projects/user-invite/UserInviteAllowlist.svelte
+++ b/web-admin/src/features/projects/user-invite/UserInviteAllowlist.svelte
@@ -10,21 +10,19 @@
   );
 </script>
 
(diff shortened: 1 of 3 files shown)


ericokuma:
All three looks good!

Only feedback is there is a bit of delay from when I click on the "Skip" button in the user invitation flow. Would be great to add a loading spinner to the button

AdityaHegde:
The slow down was probably from local dev server. It compiles pages when you first visit them. But on prod everything is precompiled.

ericokuma (approved):
Looks good and ready to merge!

PR #5546, opened 2024-08-28
tweak: invalid end date ux

briangregoryholmes:
When in end date selection mode, selecting a date prior to the current start date will re-assign the start date instead of clearing the range (updating both dates). Days prior to the start date are now a lighter shade of gray and the "potential range" hover state has been updated.

Screenshot 2024-08-28 at 5 02 42 PM

When in start date selection, selecting a date after the end date has a new visual cue, though the behavior is unchanged.

Screenshot 2024-08-28 at 5 02 28 PM


ericokuma (approved):
Looks good!

jkhwu (approved):
Looks good!

PR #5549, opened 2024-08-29
Add jobs package to admin with a River implementation

kaspersjo:


pjain1:
Couple of high level questions/feedback -
1. It seems to be using the same database as admin, I though we are going to use different db for river, can be in the same postgres instance. 
2. We will need an insert only river client as part of admin service to be able to submit jobs from anywhere, for example through webhook handlers or when a subscription change is done using an admin API etc.
3. One issue I was facing is that we cannot use `InsertTx` method of river client which submits job only when transaction succeeds as pgx driver expects transactions of type pgx.Tx which is different from our internal Tx implementation. We might need to implement `ExecutorTx` interface here, not fully sure. This is not high priority though, there can be workarounds for particular situations.

admin/jobs/river/reset_all_deployments.go
```diff
@@ -0,0 +1,80 @@
...
+	"go.uber.org/zap"
+)
+
+type ResetAllDeploymentsArgs struct{}
+
+func (ResetAllDeploymentsArgs) Kind() string { return "reset_all_deployments" }
+
+type ResetAllDeploymentsWorker struct {
+	admin *admin.Service
+	river.WorkerDefaults[ResetAllDeploymentsArgs]
```
begelundmuller, line 19: nit: put embedded fields above named fields

kaspersjo:
> Couple of high level questions/feedback -
> 
> 1. It seems to be using the same database as admin, I though we are going to use different db for river, can be in the same postgres instance.

True - I've just added a separate config entry for `RiverDatabaseURL`, then we have the option to choose at runtime.

> 2. We will need an insert only river client as part of admin service to be able to submit jobs from anywhere, for example through webhook handlers or when a subscription change is done using an admin API etc.

This is already available, the admin service always have access to the `jobs.Client` interface to insert jobs, if `Work()` is never called on the interface it's an insert only client (since `Start()` is never called on the underlying River client), this will happen if the worker is not started embedded in the admin server.

> 3. One issue I was facing is that we cannot use `InsertTx` method of river client which submits job only when transaction succeeds as pgx driver expects transactions of type pgx.Tx which is different from our internal Tx implementation. We might need to implement `ExecutorTx` interface here, not fully sure. This is not high priority though, there can be workarounds for particular situations.

This would be very interesting to explore as a next step to get transactional enqueueing, but I agree on it not being a high priority right now.

admin/jobs/river/river.go
```diff
@@ -0,0 +1,145 @@
...
+		newPeriodicJob(&ValidateDeploymentsArgs{}, "* */6 * * *", true),
+	}
+
+	riverClient, err := river.NewClient(riverpgxv5.New(dbPool), &river.Config{
+		Queues: map[string]river.QueueConfig{
+			river.QueueDefault: {MaxWorkers: 100},
+		},
+		Workers:      workers,
+		PeriodicJobs: periodicJobs,
+	})
```
begelundmuller, line 79: Should we also set:
- `JobTimeout` – we have some batch jobs that likely will run for more than a minute (the default). Should we maybe disable the timeout entirely?
- `Logger` – looks like it will default to STDOUT at above warn level. Might be nice to wire this up to our Zap logger so we get structured logs in the same format as our other logs.

cli/cmd/admin/start.go
```diff
@@ -43,6 +44,7 @@ import (
 type Config struct {
 	DatabaseDriver         string                 `default:"postgres" split_words:"true"`
 	DatabaseURL            string                 `split_words:"true"`
+	RiverDatabaseURL       string                 `split_words:"true"`
```
begelundmuller, line 47: Also set it in `.env` and do `rill devtool dotenv upload`

admin/jobs/river/river.go
```diff
@@ -0,0 +1,145 @@
...
+	"go.uber.org/zap"
+)
+
+var (
+	tracer              = otel.Tracer("github.com/rilldata/rill/admin/jobs/river")
+	meter               = otel.Meter("github.com/rilldata/rill/admin/jobs/river")
+	jobLatencyHistogram = observability.Must(meter.Int64Histogram("job_latency", metric.WithUnit("ms")))
+)
+
+type RiverClient struct {
```
begelundmuller, line 30: nit: this becomes `river.RiverClient`, consider `river.Client` instead

admin/jobs/river/river.go
```diff
@@ -0,0 +1,145 @@
...
+	if err != nil {
+		logger.Error("job failed", zap.String("name", name), zap.Error(err), zap.Duration("duration", time.Since(start)), observability.ZapCtx(ctx))
+		return err
+	}
+	logger.Info("job completed", zap.String("name", name), zap.Duration("duration", time.Since(start)), observability.ZapCtx(ctx))
+	return nil
+}
+
+// NOTE: Add new job trigger functions here
+func (c *RiverClient) ResetAllDeployments(ctx context.Context) error {
```
begelundmuller, line 139: nit: Move above the `work` and `newPeriodicJob` funcs as per the style guide https://github.com/uber-go/guide/blob/master/style.md#function-grouping-and-ordering

admin/jobs/river/validate_deployments.go
```diff
@@ -0,0 +1,147 @@
...
+func (ValidateDeploymentsArgs) Kind() string { return "validate_deployments" }
+
+type ValidateDeploymentsWorker struct {
+	admin *admin.Service
+	river.WorkerDefaults[ValidateDeploymentsArgs]
+}
+
+func (w *ValidateDeploymentsWorker) Work(ctx context.Context, job *river.Job[ValidateDeploymentsArgs]) error {
+	return work(ctx, w.admin.Logger, job.Kind, w.validateDeployments)
+}
```
begelundmuller, line 26: It's a shame to need to wrap all work functions like this, but not sure I have a great alternative suggestion – seems like the client itself doesn't have any hooks for this. If the River logger is changed to info-level, does it handle this for us by any chance?

Or could we maybe have some kind of approach similar to otel spans? Something like:
```go
func (w *Worker) Work(ctx context.Context, ...) (err error) {
  ctx := w.client.start(ctx, job)
  defer func() { w.client.done(ctx, job, err) } // Can use e.g. trace.SpanFromContext(ctx) to get the span and close it
}
```

Not sure if that's any better though...

admin/jobs/river/river.go
pjain1: might be good to add an error handler as well like [here](https://github.com/rilldata/rill/pull/5440/files#diff-aacc074d2e424275f2c45d80c99ece23ee71b9c6b456f07d1e47b0a41709e057R702) to properly log job failures. I think we should add a timeout of 10 minute. For refs I used these configs - https://github.com/rilldata/rill/pull/5440/files#diff-1ff6c7660e144613f9efda83929a9eefca974d681429e2a8dfc46fb11aa7fdfeR43

admin/jobs/river/river.go
```diff
@@ -0,0 +1,145 @@
...
+			return jobArgs, nil
+		},
+		&river.PeriodicJobOpts{RunOnStart: runOnStart},
+	)
+
+	return periodicJob
+}
+
+// Observability work wrapper for the job workers
+func work(ctx context.Context, logger *zap.Logger, name string, fn func(context.Context) error) error {
```
pjain1, line 122: how do I access job args here ?

pjain1:
> This is already available, the admin service always have access to the jobs.Client interface to insert jobs, if Work() is never called on the interface it's an insert only client (since Start() is never called on the underlying River client), this will happen if the worker is not started embedded in the admin server.

Few things should be available via Client, can you comment on these -
1. Depending on the situation I may need to schedule the same job either immediately or in future, so looks like we will need to send this information through the method on the Client. 
2. Another ability if to cancel a scheduled job, may be that can also be solved by creating `CancelJob(id)` method on the Client. 
3. Should be able to get jobId back after scheduling a job so that it can be cancelled in future if needed, so this can be exposed as return variable. Would be nice to know if job submitted was not inserted because it was a duplicate, that is also available as part of job result.

admin/jobs/river/reset_all_deployments.go
kaspersjo: Fixed

cli/cmd/admin/start.go
kaspersjo: Fixed

admin/jobs/river/river.go
kaspersjo: Fixed

admin/jobs/river/river.go
kaspersjo: Fixed

kaspersjo:
> 1. Depending on the situation I may need to schedule the same job either immediately or in future, so looks like we will need to send this information through the method on the Client.

I suggest we just add a `scheduledAt time.Time` arg to the relevant job trigger functions in the interface

> 2. Another ability if to cancel a scheduled job, may be that can also be solved by creating `CancelJob(id)` method on the Client.

Good point - I've added a `CancelJob(jobID)` to the interface
 
> 3. Should be able to get jobId back after scheduling a job so that it can be cancelled in future if needed, so this can be exposed as return variable. Would be nice to know if job submitted was not inserted because it was a duplicate, that is also available as part of job result.

Added `InsertResult` with `ID` and a `duplicate` bool to the interface

admin/jobs/river/river.go
kaspersjo: - I've added a global `JobTimeout` of 1 hour, the timeout can also be overriden on the worker level for very long running jobs
- @pjain1 good idea with the error handler, I've copied over the `ErrorHandler` from your branch
- The logger is a bit tricky since it expect a `slog` logger and it seems like we would need something like https://github.com/samber/slog-zap to get it to work with zap. Should we give that library a try?

begelundmuller: @kaspersjo How about just using this adapter from the core Zap repo: https://github.com/uber-go/zap/blob/master/exp/zapslog/example_test.go?

kaspersjo: Added a logger using the handler from Zap core

admin/jobs/river/river.go
kaspersjo: Just skip the `work()` wrapper (until we have a better solution)

begelundmuller (approved):

PR #5550, opened 2024-08-29
postgres column encryption for sensitive data

pjain1:
Fixes https://github.com/rilldata/rill-private-issues/issues/590

Before merging this, we will need to set following admin property in helm with secret in vault, also dev .env file should be updated, otherwise admin service will fail to start  -
```
RILL_ADMIN_DATABASE_ENCRYPTION_KEYRING='[{"key_id":"<id(can use numerical order for id starting from 1)>","key":"<32-bytes-base64-encoded-secret>"}]'
```

For key rotation, new key should be prepended to the array as the first key is used for encryption. New secret can be generated using `openssl rand -base64 32`

Included a utility to generate keyring - `go run scripts/generate_enc_keyring/main.go`. It accepts prepending to existing keyring with `go run scripts/generate_enc_keyring/main.go '[{"key_id":"<>","key":"<>"},{"key_id":"key2","key":"<>"}]'`

### DIFF
--- a/admin/admin.go
+++ b/admin/admin.go
@@ -16,17 +16,18 @@ import (
 )
 
 type Options struct {
-	DatabaseDriver     string
-	DatabaseDSN        string
-	ExternalURL        string
-	FrontendURL        string
-	ProvisionerSetJSON string
-	DefaultProvisioner string
-	VersionNumber      string
-	VersionCommit      string
-	MetricsProjectOrg  string
-	MetricsProjectName string
-	AutoscalerCron     string
+	DatabaseDriver            string
+	DatabaseDSN               string
+	ExternalURL               string
+	FrontendURL               string
+	ProvisionerSetJSON        string
+	DefaultProvisioner        string
+	VersionNumber             string
+	VersionCommit             string
+	MetricsProjectOrg         string
+	MetricsProjectName        string
+	AutoscalerCron            string
+	DatabaseEncryptionKeyring string
 }
 
 type Service struct {
@@ -49,9 +50,9 @@ type Service struct {
 	PaymentProvider  payment.Provider
 }
 
-func New(ctx context.Context, opts *Options, logger *zap.Logger, issuer *auth.Issuer, emailClient *email.Client, github Github, aiClient ai.Client, assets *storage.BucketHandle, biller billing.Biller, p payment.Provider, databaseEncryptionKeyring []*database.EncryptionKey) (*Service, error) {
+func New(ctx context.Context, opts *Options, logger *zap.Logger, issuer *auth.Issuer, emailClient *email.Client, github Github, aiClient ai.Client, assets *storage.BucketHandle, biller billing.Biller, p payment.Provider) (*Service, error) {
 	// Init db
-	db, err := database.Open(opts.DatabaseDriver, opts.DatabaseDSN, databaseEncryptionKeyring)
+	db, err := database.Open(opts.DatabaseDriver, opts.DatabaseDSN, opts.DatabaseEncryptionKeyring)
 	if err != nil {
 		logger.Fatal("error connecting to database", zap.Error(err))
 	}
--- a/admin/admin.go
+++ b/admin/admin.go
@@ -49,9 +49,9 @@ type Service struct {
 	PaymentProvider  payment.Provider
 }
 
-func New(ctx context.Context, opts *Options, logger *zap.Logger, issuer *auth.Issuer, emailClient *email.Client, github Github, aiClient ai.Client, assets *storage.BucketHandle, biller billing.Biller, p payment.Provider) (*Service, error) {
+func New(ctx context.Context, opts *Options, logger *zap.Logger, issuer *auth.Issuer, emailClient *email.Client, github Github, aiClient ai.Client, assets *storage.BucketHandle, biller billing.Biller, p payment.Provider, databaseEncryptionKeyring []*database.EncryptionKey) (*Service, error) {
 	// Init db
-	db, err := database.Open(opts.DatabaseDriver, opts.DatabaseDSN)
+	db, err := database.Open(opts.DatabaseDriver, opts.DatabaseDSN, databaseEncryptionKeyring)
 	if err != nil {
 		logger.Fatal("error connecting to database", zap.Error(err))
 	}
--- a/admin/admin.go
+++ b/admin/admin.go
@@ -18,6 +18,7 @@ import (
 type Options struct {
 	DatabaseDriver            string
 	DatabaseDSN               string
+	DatabaseEncryptionKeyring string
 	ExternalURL               string
 	FrontendURL               string
 	ProvisionerSetJSON        string
@@ -27,7 +28,6 @@ type Options struct {
 	MetricsProjectOrg         string
 	MetricsProjectName        string
 	AutoscalerCron            string
-	DatabaseEncryptionKeyring string
 }
 
 type Service struct {
--- a/admin/database/database.go
+++ b/admin/database/database.go
@@ -34,10 +34,6 @@ func Open(driver, dsn, encKeyringConfig string) (DB, error) {
 		return nil, fmt.Errorf("error parsing encryption keyring: %w", err)
 	}
 
-	if len(encKeyring) == 0 {
-		return nil, errors.New("encryption keyring is empty")
-	}
-
 	db, err := d.Open(dsn, encKeyring)
 	if err != nil {
 		return nil, err
@@ -939,6 +935,10 @@ type EncryptionKey struct {
 }
 
 func ParseEncryptionKeyring(keyring string) ([]*EncryptionKey, error) {
+	if keyring == "" {
+		return nil, nil
+	}
+
 	var encKeyring []*EncryptionKey
 	err := json.Unmarshal([]byte(keyring), &encKeyring)
 	if err != nil {
--- a/admin/database/database.go
+++ b/admin/database/database.go
@@ -929,6 +929,9 @@ type Asset struct {
 	CreatedOn      time.Time `db:"created_on"`
 }
 
+// EncryptionKey represents an encryption key for column-level encryption/decryption.
+// Column-level encryption provides an extra layer of security for highly sensitive columns in the database.
+// It is implemented on the application side before writes to and after reads from the database.
 type EncryptionKey struct {
 	ID     string `json:"key_id"`
 	Secret []byte `json:"key"`
--- a/admin/database/database.go
+++ b/admin/database/database.go
@@ -2,11 +2,13 @@ package database
 
 import (
 	"context"
-	"encoding/base64"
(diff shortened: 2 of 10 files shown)


admin/provisioner/static_test.go
```diff
@@ -17,7 +18,15 @@ func Test_staticProvisioner_Provision(t *testing.T) {
...
 	defer pg.Terminate(t)
 
-	db, err := database.Open("postgres", pg.DatabaseURL)
+	secret := make([]byte, 32) // 32 bytes for AES-256
+	_, err := rand.Read(secret)
+	require.NoError(t, err)
+
+	encKeyRing := []*database.EncryptionKey{
+		{ID: "test", Secret: secret},
+	}
```
begelundmuller, line 27: Maybe add a `database.NewRandomKeyring` util func? Or use empty keyrings (so encryption is disabled) for tests?

cli/cmd/admin/start.go
```diff
@@ -377,3 +391,18 @@ func StartCmd(ch *cmdutil.Helper) *cobra.Command {
...
+
+	var keyRing []*database.EncryptionKey
+
+	err := json.Unmarshal([]byte(keyRingConfig), &keyRing)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse encryption keyring: %w", err)
+	}
+
+	return keyRing, nil
+}
```
begelundmuller, line 408: Maybe we should push this parsing into the `database` package? Could have `database.NewRandomKeyring() string` and `database.ParseKeyring(keyring string) []*database.EncryptionKey`? Maybe even pass `keystring string` to `database.Open` and let it do the parsing there.

begelundmuller: This could also be used to add a script in `scripts/generate_db_keyring` for generating a safe keyring to set in Vault (similar to `scripts/generate_jwks`).

begelundmuller (changes requested):

admin/database/database.go
```diff
@@ -918,3 +921,33 @@ type Asset struct {
...
 	CreatedOn      time.Time `db:"created_on"`
 }
+
+type EncryptionKey struct {
+	ID     string `json:"key_id"`
+	Secret []byte `json:"key"`
+}
+
+// UnmarshalJSON custom method for EncryptionKey
+func (e *EncryptionKey) UnmarshalJSON(data []byte) error {
```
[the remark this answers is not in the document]
pjain1: ok yeah it works, thanks

admin/database/postgres/postgres.go
```diff
@@ -352,24 +358,42 @@ func (c *connection) FindProjectByName(ctx context.Context, orgName, name string
...
 
 func (c *connection) InsertProject(ctx context.Context, opts *database.InsertProjectOptions) (*database.Project, error) {
 	if err := database.Validate(opts); err != nil {
 		return nil, err
 	}
 
+	// encrypt the prod variables and store the key id used for encryption
+	if len(c.encKeyring) == 0 {
+		return nil, errors.New("no encryption key found")
+	}
```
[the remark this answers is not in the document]
pjain1: this was just a defensive check, actually admin starts to fail if keyring does not have any keys. I did this so that we don't miss encryption if there is some misconfiguration or we forgot to include encryption keys. What behaviour do we want here ?

admin/database/postgres/postgres.go
```diff
@@ -2106,3 +2168,49 @@ func (e *wrappedError) Error() string {
 func (e *wrappedError) Unwrap() error {
 	return e.err
 }
+
+func encrypt(plaintext, key []byte) (string, error) {
```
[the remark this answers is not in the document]
pjain1: ok changed both to `string -> string` for convenience

admin/provisioner/static_test.go
pjain1: Lets finalize on whether moving forward we allow empty keyrings or not - https://github.com/rilldata/rill/pull/5550#discussion_r1738078868.

cli/cmd/admin/start.go
pjain1: good suggestions, added

admin/database/postgres/postgres.go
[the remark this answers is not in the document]
begelundmuller: I think we should allow the empty keyring for now

admin/provisioner/static_test.go
begelundmuller: Responded on the other comment

admin/database/postgres/postgres.go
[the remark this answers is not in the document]
pjain1: ok

admin/provisioner/static_test.go
pjain1: removed use of keyring here and `admin_rbac_test` but kept it in `postgres_test` as it test updating prod variables.

admin/admin.go
```diff
@@ -16,17 +16,18 @@ import (
...
+	ExternalURL               string
+	FrontendURL               string
+	ProvisionerSetJSON        string
+	DefaultProvisioner        string
+	VersionNumber             string
+	VersionCommit             string
+	MetricsProjectOrg         string
+	MetricsProjectName        string
+	AutoscalerCron            string
+	DatabaseEncryptionKeyring string
```
begelundmuller, line 30: nit: move up below the `DatabaseDSN` so we group database-related properties together

admin/database/database.go
```diff
@@ -918,3 +928,36 @@ type Asset struct {
 	OwnerID        string    `db:"owner_id"`
 	CreatedOn      time.Time `db:"created_on"`
 }
+
+type EncryptionKey struct {
```
begelundmuller, line 932:
```suggestion
// EncryptionKey represents an encryption key for column-level encryption/decryption.
// Column-level encryption provides an extra layer of security for highly sensitive columns in the database.
// It is implemented on the application side before writes to and after reads from the database.
type EncryptionKey struct {
```

admin/database/database.go
```diff
@@ -918,3 +928,36 @@ type Asset struct {
...
+	ID     string `json:"key_id"`
+	Secret []byte `json:"key"`
+}
+
+func ParseEncryptionKeyring(keyring string) ([]*EncryptionKey, error) {
```
begelundmuller, line 937:
```suggestion
// ParseEncryptionKeyring parses a JSON string containing an array of EncryptionKey objects.
// If the provided string is empty, an empty keyring is returned.
// When using an empty keyring, columns will be read and written without applying encryption/decryption.
func ParseEncryptionKeyring(keyring string) ([]*EncryptionKey, error) {
```

admin/database/postgres/postgres.go
```diff
@@ -1915,32 +1931,86 @@ type projectDTO struct {
...
+			return nil, "", err
+		}
+		vars[k] = encrypted
+	}
+	return vars, encKeyID, nil
+}
+
+// returns the map with decrypted values and the encryption key id used
+func (c *connection) decryptMap(m map[string]string, encKeyID string) (map[string]string, error) {
+	if encKeyID != "" {
```
begelundmuller, line 1988: nit: can reduce nesting by returning early `if encKeyID == ""`

admin/database/database.go
```diff
@@ -18,14 +22,19 @@ func Register(name string, driver Driver) {
 	Drivers[name] = driver
 }
 
-// Open opens a new database connection.
-func Open(driver, dsn string) (DB, error) {
+// Open opens a new database connection. encryptionKeyring is for client-side column-level encryption
+func Open(driver, dsn, encKeyringConfig string) (DB, error) {
```
begelundmuller, line 26:
```suggestion
// Open opens a new database connection.
// See ParseEncryptionKeyring for the expected format for encKeyringConfig.
func Open(driver, dsn, encKeyringConfig string) (DB, error) {
```

admin/database/postgres/postgres.go
```diff
@@ -2106,3 +2176,55 @@ func (e *wrappedError) Error() string {
...
+	if err != nil {
+		return "", err
+	}
+
+	nonceSize := aesGCM.NonceSize()
+	if len(ciphertext) < nonceSize {
+		return "", errors.New("ciphertext too short")
+	}
+
+	nonce, ciphertext := ciphertext[:nonceSize], ciphertext[nonceSize:]
```
begelundmuller, line 2224: Seems weird that `aesGCM.Seal` does the append automatically, but we need to manually parse it here. Is this expected?

Also, just to avoid errors, did you triple check that this is the correct way to store the nonce (e.g. that the nonce size or output size never changes)?

begelundmuller (approved):

admin/database/postgres/postgres.go
pjain1:
```
	// Seal encrypts and authenticates plaintext, authenticates the
	// additional data and appends the result to dst, returning the updated
	// slice. The nonce must be NonceSize() bytes long and unique for all
	// time, for a given key.
	//
	// To reuse plaintext's storage for the encrypted output, use plaintext[:0]
	// as dst. Otherwise, the remaining capacity of dst must not overlap plaintext.
	Seal(dst, nonce, plaintext, additionalData []byte) []byte
```
We are deliberately using `nonce` slice as dst so that encrypted text is appended to it and no need to separately store the nonce. 
`crypto/cipher/gcm.go` ->
```
const (
	gcmBlockSize         = 16
	gcmTagSize           = 16
	gcmMinimumTagSize    = 12 // NIST SP 800-38D recommends tags with 12 or more bytes.
	gcmStandardNonceSize = 12
)
...
func NewGCM(cipher Block) (AEAD, error) {
	return newGCMWithNonceAndTagSize(cipher, gcmStandardNonceSize, gcmTagSize)
}
```
So I think nonce size is also not a problem. About storing nonce along with encrypted text, I am not an expert but I found this to be a common practice - https://crypto.stackexchange.com/questions/57895/would-it-be-safe-to-store-gcm-nonce-in-the-encrypted-output. Let me know if you have any other opinion.

pjain1: Upon looking more, seems like its not a global standard and libraries are free to choose the size. 12 is the common size but go implementation can choose to change the value of const `gcmStandardNonceSize` in newer version, not sure.

pjain1: Should we store nonce size ?

pjain1: If in newer versions go lib changes the nonce size then as per current code its not going to work anyways because they have 
```
if len(nonce) != g.nonceSize {
		panic("crypto/cipher: incorrect nonce length given to GCM")
	}
```
or they will have to make it backwards compatible somehow, so I doubt they will change it. Only use case is we use another implementation in future but there should be someway there as well to make it work with this size.

begelundmuller: We're discussing this offline and will follow up here

PR #5563, opened 2024-08-30
feat: Visual Metrics Editing

briangregoryholmes:


web-common/src/features/visual-metrics-editing/MetricsTable.svelte
```diff
@@ -0,0 +1,262 @@
...
+  onMount(() => {
+    const cells = clientWidth.children;
+    const initialNameWidth = cells[1].getBoundingClientRect().width;
+    const initialLabelWidth = cells[2].getBoundingClientRect().width;
+    const initialFormatWidth =
+      type === "measures" ? cells[4].getBoundingClientRect().width : 0;
+
+    nameWidth.set(initialNameWidth);
+    labelWidth.set(initialLabelWidth);
+    formatWidth.set(initialFormatWidth);
```
ericpgreen2, line 54: I might be missing it but I don't see these stores set again, so I'm wondering if the `MaxStore` layer is needed?

ericpgreen2 (approved):
Overall looks good! Given it's behind a feature flag, I'd like to merge and we can work on the final mile in follow-on PRs. In the final mile, the feature as a whole probably merits a dedicated e2e test.

Here are some UXQA findings:
1. I'm unable to submit my edits for any of the measure input fields, like `name`: [Jam](https://jam.dev/c/fc4c2ff3-56c6-47a6-986a-366e1ceec1d1)
2. Leftover "true" text: 
    image
3. Given point 1, I wasn't able to fully test it, but we should make sure that `count(*)` is allowed
    image
4. I think "Summable metric" should default to `true`
5. @jkhwu can we add some space between the Measures & Dimensions tables? They feel uncomfortably close.
6. I used the dropdown to switch to a new model, and I hit the following error. When using the model switcher, can we remove the `table` key (if any) from the YAML?
    image
7. We should store the view state (code or viz) somewhere (local storage?) so that refreshing the code view doesn't automatically flip you to the viz view. Similarly, navigating from code->Preview->"Edit metrics" should take me back to code.

web-common/src/features/visual-metrics-editing/MetricsTable.svelte
briangregoryholmes: They are all set once per table, so twice for each store.

web-common/src/features/canvas/ViewSelector.svelte
```diff
@@ -11,45 +11,48 @@
     { view: "viz", icon: Viz },
   ];
 
-  export let selectedView = "split";
+  export let selectedView: "code" | "split" | "viz" = "split";
+  export let split = true;
```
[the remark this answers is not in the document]
briangregoryholmes: Adjusted

briangregoryholmes:
1. You are missing Format, which is a required field. This is one of the reasons why I think inline editing is necessary for this surface (mismatch between UI and YAML requirements), but I've added a tooltip to the save changes button that lists out missing required fields. 
2. Removed
3. It works.
4. Agreed. Changed.
5. My spacing was off from Figma. Updated.
6. Fixed.
7. Working on it.

jkhwu:
Note to self: 
`features:
  visualEditing: true`

PR #5566, opened 2024-08-30
fix: infinite query glitch on invalid spec

djbarnwal:
Fixes

https://github.com/rilldata/rill-private-issues/issues/630
https://github.com/rilldata/rill-private-issues/issues/631

Instead of infinite loading state, we show the error message for the failed request

image

### DIFF
--- a/web-common/src/features/charts/prompt/ChartDataDisplay.svelte
+++ b/web-common/src/features/charts/prompt/ChartDataDisplay.svelte
@@ -36,8 +36,6 @@
   $: chartData = $chartDataQuery?.data?.data;
   $: isFetching = $chartDataQuery?.isFetching ?? false;
   $: errorMessage = $chartDataQuery?.error?.response?.data?.message;
-
-  $: console.log($chartDataQuery, resolverProperties, errorMessage, chartData);
 </script>
 
 {#if resolverProperties}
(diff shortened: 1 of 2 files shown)


web-common/src/features/charts/prompt/ChartDataDisplay.svelte
ericpgreen2, line 1: Now I'm not following what the `prompt` directory is supposed to represent. I thought it was supposed to encapsulate all the code related to AI chart generation?

ericpgreen2 (commented):
UXQA: when I edit values in the `data:` section, the chart does not respond. [See JAM](https://jam.dev/c/8746c49b-36f4-4539-bd85-c91ae395d6c5).

web-common/src/features/charts/prompt/ChartDataDisplay.svelte
djbarnwal: Don't have much context on this. Probably @AdityaHegde @briangregoryholmes would know more.

djbarnwal:
> UXQA: when I edit values in the `data:` section, the chart does not respond. [See JAM](https://jam.dev/c/8746c49b-36f4-4539-bd85-c91ae395d6c5).

Did a bit of digging. The issue is also on main. The `resource` query needs to be invalidated when the data query is changing. Will create a separate PR for that.

web-common/src/features/charts/prompt/ChartDataDisplay.svelte
AdityaHegde: It is. I think just this file is misplaced.

ericpgreen2 (approved):
Approving, just it'd be nice to move the non-AI code outside of the `charts/prompt` directory before merging.

djbarnwal:
> Approving, just it'd be nice to move the non-AI code outside of the `charts/prompt` directory before merging.

The whole custom dashboard / charts terminology needs to be updated to canvas and components. Will do it in a separate PR.

PR #5573, opened 2024-09-01
feat: enable time comparison for time dimension in pivot

djbarnwal:
Enables time comparison for time dimensions in the pivot table.

### DIFF
--- a/web-common/src/features/dashboards/pivot/pivot-infinite-scroll.ts
+++ b/web-common/src/features/dashboards/pivot/pivot-infinite-scroll.ts
@@ -11,8 +11,18 @@ import {
   sortAcessors,
 } from "./pivot-utils";
 
+/**
+ * We need to make sure that when we update these limits, we also update the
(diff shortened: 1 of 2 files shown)


ericpgreen2:
UXQA –
1. In the row for July 22nd, what does the `-` represent? It looks like the deltas should read `0` and `0%`?
    image
2. It's hard to tell what date is being used for the comparison calculations. Maybe we should add a tooltip that says what date is being compared?

djbarnwal:
1. Delta of 0 and 0% would mean that the data is defined (non-null) and  same as before. The `-` represents that either the current or the comparison period data is `null`.
2. We can add a tooltip to show previous period value as well as the comparison range. Although, we get the comparison values directly from the API so either we have to estimate the compared range on the client or update the API to also get the associated comparison range. Both of which would be non trivial and out of scope for this PR. We can get a UX design for the tooltip from @jkhwu and address it in a follow up.

ericpgreen2 (approved):

PR #5574, opened 2024-09-02
feat: Rill cloud E2E setup

AdityaHegde:
- [x] Get a working docker compose for rill cloud
- [x] Figure out auth (adding a container that has barebones oauth)
- [ ] Figure out github. We can perhaps skip app installation steps and start a local git server.
- [x] Figure out a good way to distribute .env (Will only contain keys for the dockerised components)
- [ ] Add a github action to build UI and start cloud


AdityaHegde:
We will be using dev env for running e2e

PR #5597, opened 2024-09-03
feat: undo zoom

briangregoryholmes:
Changes the copy of the chart Zoom UI elements and adds the ability to revert to the previous range via keyboard shortcut or button click.


briangregoryholmes:
Moving to draft to resolve an outstanding question.

lovincyrus (changes requested):
Nice QoL changes!

- [ ] Add font-semibold to the hotkeys
- [ ] Consistent copy (1) (Z) Zoom, (2) (⌘ Z) Undo Zoom
- [ ] ⌘Z should have a whitespace in between. Try adding tracking-widest to the span. Ref: https://ui.shadcn.com/docs/components/command
- [ ] Add windows support, Ctrl+Z to undo. Refer to https://github.com/reaviz/reakeys/blob/master/src/utils.ts

lovincyrus (changes requested):
The undo changes work great! One thing that I noticed, we shouldn't show Zoom (Z) when the interval is invalid.

See screenshot and recording:

CleanShot 2024-09-18 at 10 36 56@2x


https://github.com/user-attachments/assets/2d0fe414-6d1b-43ef-9c7a-3b1b6874cd85

lovincyrus (approved):

PR #5534, opened 2024-08-27
fix: Simplify `rill deploy` command and break up upload and github flows

AdityaHegde:
closes [rilldata/rill-private-issues#5493](https://github.com/rilldata/rill-private-issues/issues/533)

PR #5600, opened 2024-09-04
Docs deploy change

royendo:
Part two of https://github.com/rilldata/rill/pull/5534 to fix up docs for rill project upload and rill project connect-github


royendo:
https://rilldata.slack.com/archives/C01UFNH4W3B/p1725361752241639

wait for confirmation before merging!

royendo:
Looks like we pushed 0.49 to deprecate rill deploy so will push forward with docs.

AndrewRTsao (changes requested):
So it seems like we made some additional last minute-changes that we cherry picked into the v0.49 release. See [this thread](https://rilldata.slack.com/archives/CTZ8XBQ85/p1726309201981019) and https://github.com/rilldata/rill/pull/5698.

We should probably incorporate the following changes into this PR:
- It seems like `rill project deploy` is now being used in favor of `rill project upload`
- `rill deploy` remains available but is deprecated. We should still cover what happens when end users use `rill deploy` as there are a few Y/N questions they have to answer but call out that it's deprecated in favor of the other workflow (using an admonition).

royendo:
@AndrewRTsao Thanks for the information! 

I upgraded to v49.1 and updated the output information on the docs, looks like we changed some of the wording too!

Added a #Deprecated Rill Deploy section at the very end of the doc that links back to what it is actually doing with the main commands IE, disable automatic = rill project upload and enable automatic = rill project connect-github

Personally, from a UX perspective, i still dont like the command `rill project connect-github` as this sounds like youre not deploying but rather just added a github repo to an **existing project**. EITHER WAY, we are pushing for our users to use the UI so its just small nit stuff :) :)

AndrewRTsao (approved):

Issue #5282, opened 2024-07-16
Feat: SQL Expressions as URI support on dimensions


mindspank:
https://www.notion.so/rilldata/Support-for-exposing-clickable-URIs-derived-from-dimension-values-bf0218bd636043208095608b7c5e0947

k-anshul:
Hey @mindspank 

The backend changes for this are now complete.

PR #5603, opened 2024-09-04
feat : SQL expressions for URI in dimensions

k-anshul:
https://github.com/rilldata/rill/issues/5282

Allows caller to get URI for a dimension using a measure compute:
```
 {
      "name": "app_site_domain_uri",
      "uri": {
          "dimension": "app_site_domain"
      }
}
```
check `runtime/queries/metricsview_aggregation_test.go -> testMetricsViewsAggregationURI` for detailed request.


mindspank:
@nishantmonu51 we'd need some tiny frontend changes to make use of the new backend property as well. Are we tracking that anywhere?

k-anshul:
Hey @mindspank 

Just an FYI that we will change this approach. This has certain limitations.

proto/rill/runtime/v1/queries.proto
```diff
@@ -404,6 +404,7 @@ message MetricsViewAggregationMeasure {
     MetricsViewAggregationMeasureComputeComparisonDelta comparison_delta = 8;
     MetricsViewAggregationMeasureComputeComparisonRatio comparison_ratio = 9;
     MetricsViewAggregationMeasureComputePercentOfTotal percent_of_total = 10;
+    MetricsViewAggregationMeasureURI uri = 11;
```
begelundmuller, line 407: The type name should be `MetricsViewAggregationMeasureComputeURI` for consistency with the other "compute" options.

runtime/metricsview/query.go
```diff
@@ -32,13 +32,18 @@ type Dimension struct {
...
 }
 
 type DimensionComputeTimeFloor struct {
 	Dimension string    `mapstructure:"dimension"`
 	Grain     TimeGrain `mapstructure:"grain"`
 }
 
+type DimensionComputeURI struct {
+	Dimension string `mapstructure:"dimension"`
+}
```
begelundmuller, line 45: Seems like this isn't used?

begelundmuller (changes requested):

begelundmuller (approved):

PR #5605, opened 2024-09-04
improved idle detection

briangregoryholmes:
This PR adds improved idle detection timeout that closes the resource watchers (and displays a small banner) after two minutes of inactivity (keydown or mouse click). It also adds a priority inactivity detection that closes the watchers after 30 seconds if the tab's visibility has changed.

There are a few different scenarios that we can detect and set timeouts for:

1. The tab is no longer visible
2. A blur event has happened on the page, which does not necessarily mean it's no longer visible
3. No mouse click events
4. No keydown events
5. No mouse move events
6. IdleDetection API, though this is supported only in Chrome and in HTTPS contexts

### DIFF
--- a/web-common/src/features/entity-management/ResourceWatcher.svelte
+++ b/web-common/src/features/entity-management/ResourceWatcher.svelte
@@ -35,8 +35,8 @@
 
   function handleVisibilityChange() {
     if (document.visibilityState === "visible") {
-      fileWatcher.reconnect().catch(console.error);
-      resourceWatcher.reconnect().catch(console.error);
+      fileWatcher.heartbeat();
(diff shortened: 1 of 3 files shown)


web-common/src/features/entity-management/ResourceWatcher.svelte
```diff
@@ -36,13 +38,31 @@
       fileWatcher.reconnect().catch(console.error);
       resourceWatcher.reconnect().catch(console.error);
```
ericpgreen2, line 39: Wondering – should these be `heartbeat()` too?

ericpgreen2 (approved):
Looks good! I like the subtle banner as indicator.

PR #5613, opened 2024-09-05
Fix: handle missing instance on runtime in health check job

k-anshul:


admin/worker/deployments_health_check.go
```diff
@@ -38,54 +43,97 @@ func (w *Worker) deploymentsHealthCheck(ctx context.Context) error {
...
 		// fetch again
 	}
+
+	// compare expected and actual instances
+	for host, expected := range expectedInstances {
+		actual, ok := actualInstances[host]
+		if !ok {
+			// runtime health check failed
+			continue
+		}
```
begelundmuller, line 77: Shouldn't it error here?

begelundmuller (changes requested):

admin/worker/deployments_health_check.go
k-anshul: No logging of errors is handled `deploymentHealthCheck` itself. 
Here it is just comparing expected and actual instances across all runtimes.

begelundmuller: Okay, I see

begelundmuller (commented):
What about if there's a leaked instance on a runtime that doesn't exist in a deployment?

begelundmuller (approved):

PR #5615, opened 2024-09-05
Admin: Make error codes for database errors more consistent

begelundmuller:
This PR:
- Adds utility functions to `admin/database` for checking validation errors
- Adds handling of the `admin/database` errors to the fallback error handler in the gRPC server
- Changes many error cases away from explicit `codes.Internal` errors to non-annotated errors that will be handled by the fallback handler

The PR is not exhaustive and there are still many error cases that retain explicit error handling, some of them for justifiable errors (like needing to return `codes.Internal` for an unexpected not-found error), others because the expected behavior is not entirely clear and I'm erroring on the side of caution.

This PR was motivated by [this](https://rilldata.slack.com/archives/C05TXA7QSTY/p1725530133665909) false positive alert.

### DIFF
--- a/admin/server/alerts.go
+++ b/admin/server/alerts.go
@@ -35,10 +35,7 @@ func (s *Server) GetAlertMeta(ctx context.Context, req *adminv1.GetAlertMetaRequ
 
 	proj, err := s.admin.DB.FindProject(ctx, req.ProjectId)
 	if err != nil {
-		if errors.Is(err, database.ErrNotFound) {
-			return nil, status.Error(codes.NotFound, "project not found")
-		}
-		return nil, status.Error(codes.InvalidArgument, err.Error())
+		return nil, err
 	}
 
 	permissions := auth.GetClaims(ctx).ProjectPermissions(ctx, proj.OrganizationID, proj.ID)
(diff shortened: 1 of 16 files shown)


admin/server/alerts.go
```diff
@@ -35,10 +35,7 @@ func (s *Server) GetAlertMeta(ctx context.Context, req *adminv1.GetAlertMetaRequ
 
 	proj, err := s.admin.DB.FindProject(ctx, req.ProjectId)
 	if err != nil {
-		if errors.Is(err, database.ErrNotFound) {
-			return nil, status.Error(codes.NotFound, "project not found")
-		}
-		return nil, status.Error(codes.InvalidArgument, err.Error())
```
pjain1, line 41: not sure what was the purpose of using InvalidArgument here but now Internal error would be returned which is more appropriate I guess. There are many instances like this below.

begelundmuller: Yeah `Internal` is more appropriate here. There were many such cases of inconsistent error handling, hopefully there will be fewer now.

pjain1 (approved):

PR #5617, opened 2024-09-05
Disable cron refreshes in dev (i.e. on localhost)

begelundmuller:
This PR changes the `refresh:` YAML clause to ignore the `refresh.cron:` and `refresh.every:` properties in development (i.e. when running under `rill start` on `localhost`) unless `refresh.run_in_dev: true` is explicitly set.

Closes https://github.com/rilldata/rill-private-issues/issues/654


k-anshul (approved):
We also need to inform to users having self hosted deployments.

PR #5619, opened 2024-09-05
Docs: Update docs for cron refreshes in Rill Developer

begelundmuller:


AndrewRTsao (approved):
Thanks for the contribution, made some minor edits but otherwise good to go 🙂

PR #5620, opened 2024-09-05
CLI: New commands `rill sudo org join` and `rill sudo org list-admins`

begelundmuller:
This PR adds two commands to make it easier to manage orgs as a superuser:
- `rill sudo org list-admins <org>`: lists all the admins of an organization. This can be used together with `rill sudo user assume` to _temporarily_ assume an admin role on an organization that allows access from superusers.
- `rill sudo org join <org>`: adds the current user as an admin of an organization. This action is not temporary and the user will show up in member listings.

Closes https://github.com/rilldata/rill-private-issues/issues/624


admin/server/organizations.go
```diff
@@ -568,33 +582,33 @@ func (s *Server) AddOrganizationMemberUser(ctx context.Context, req *adminv1.Add
...
-				return nil, status.Error(codes.Internal, err.Error())
-			}
-			// If the user is already in the all user group, we can ignore the error
+			return fmt.Errorf("failed to add user to all user group: %w", err)
+		}
+
+		return tx.Commit()
+	}()
+	if err != nil {
+		if !errors.Is(err, database.ErrNotUnique) {
```
Kavinjsir, line 606: Noticed a similar pattern here.

Kavinjsir (approved):
LGTM!
Just left a small question to understand the coding pattern we are using :)

admin/server/organizations.go
begelundmuller: See answer above

admin/server/organizations.go
```diff
@@ -545,9 +547,21 @@ func (s *Server) AddOrganizationMemberUser(ctx context.Context, req *adminv1.Add
 			OrgID:     org.ID,
 			RoleID:    role.ID,
 		})
-		// continue sending an email if the user already exists
-		if err != nil && !errors.Is(err, database.ErrNotUnique) {
-			return nil, status.Error(codes.Internal, err.Error())
+		if err != nil {
```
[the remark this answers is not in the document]
begelundmuller: Both approaches would have been fine. But in theory, this approach is slightly better because a) there is no race condition between the check and the insertion/update, b) for the optimistic case where an invite does not yet exist, it only does one DB call instead of two, c) the DB will do a constraint check in any case, so there's no performance issue in relying on it.

pjain1 (approved):
Should we just error in case of role change of already added user and indicate to use `set-role` cmd. Earlier behaviour was incorrect in case of role change that it would do nothing and still send out an email with wrong role name. I am fine with this approach as well.

PR #5621, opened 2024-09-06
encrypt magic tokens

pjain1:
Fixes https://github.com/rilldata/rill-private-issues/issues/642


admin/server/magic_tokens.go
```diff
@@ -242,6 +242,19 @@ func magicAuthTokenToPB(tkn *database.MagicAuthTokenWithUser) (*adminv1.MagicAut
...
+			return nil, status.Error(codes.NotFound, fmt.Sprintf("project with id %s not found", tkn.ProjectID))
+		}
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+
+	org, err := s.admin.DB.FindOrganization(ctx, proj.OrganizationID)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to find organization for project: %v", err.Error())
+	}
+
```
pjain1, line 257: instead of doing db queries here, we can change the find magic auth token db queries to lookup org name, org custom domain and project name by doing join 3 way join with org and project tables and also we will need to overload the `MagicAuthToken` struct with org name, customDomain and project name, so was not sure. Should we do that instead ?

admin/database/database.go
```diff
@@ -607,18 +607,20 @@ type InsertDeploymentAuthTokenOptions struct {
...
-	CreatedByUserID       *string        `db:"created_by_user_id"`
-	Attributes            map[string]any `db:"attributes"`
-	MetricsView           string         `db:"metrics_view"`
-	MetricsViewFilterJSON string         `db:"metrics_view_filter_json"`
-	MetricsViewFields     []string       `db:"metrics_view_fields"`
-	State                 string         `db:"state"`
+	ID                      string
+	SecretHash              []byte         `db:"secret_hash"`
+	TokenStr                string         `db:"token_str"`
+	TokenStrEncryptionKeyID string         `db:"token_str_encryption_key_id"`
```
begelundmuller, line 613: For consistency with `SecretHash`, I would suggest calling it `Secret` instead of `TokenStr`. Also, should it be a `[]byte` instead of a `string`?

The auth code generally uses "token" to mean a serialized access token in the form `rill_<type>_<id><secret>`, but in this case it's only the `secret` part that is being encrypted and persisted.

admin/database/postgres/postgres.go
```diff
@@ -1142,16 +1142,21 @@ func (c *connection) InsertMagicAuthToken(ctx context.Context, opts *database.In
...
+
 	res := &magicAuthTokenDTO{}
-	err := c.getDB(ctx).QueryRowxContext(ctx, `
-		INSERT INTO magic_auth_tokens (id, secret_hash, project_id, expires_on, created_by_user_id, attributes, metrics_view, metrics_view_filter_json, metrics_view_fields, state)
-		VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10) RETURNING *`,
-		opts.ID, opts.SecretHash, opts.ProjectID, opts.ExpiresOn, opts.CreatedByUserID, opts.Attributes, opts.MetricsView, opts.MetricsViewFilterJSON, opts.MetricsViewFields, opts.State,
+	err = c.getDB(ctx).QueryRowxContext(ctx, `
+		INSERT INTO magic_auth_tokens (id, secret_hash, project_id, expires_on, created_by_user_id, attributes, metrics_view, metrics_view_filter_json, metrics_view_fields, state, token_str, token_str_encryption_key_id)
+		VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12) RETURNING *`,
+		opts.ID, opts.SecretHash, opts.ProjectID, opts.ExpiresOn, opts.CreatedByUserID, opts.Attributes, opts.MetricsView, opts.MetricsViewFilterJSON, opts.MetricsViewFields, opts.State, encTknStr, encKeyID,
```
begelundmuller, line 1154: nit: add fields in the same relative order as they are defined in the `database.MagicAuthToken` type – makes auditing for correctness easier

admin/server/magic_tokens.go
begelundmuller: Since it's a relatively infrequent code path, I would just suggest doing the project/org lookup separately in the server handler, and passing them to `magicAuthTokenToPB`. For example, in `ListMagicAuthTokens` it's already doing a project lookup anyway.

proto/rill/admin/v1/api.proto
```diff
@@ -2198,6 +2198,7 @@ message MagicAuthToken {
   rill.runtime.v1.Expression metrics_view_filter = 10;
   repeated string metrics_view_fields = 11;
   string state = 12;
+  string url = 13;
```
begelundmuller, line 2201: 1. Can we move this field up above `created_on`? Since it's a key field that's always set.
2. Can we also return the token string separately here? I.e. have a separate `string token = 14;` field?

admin/database/postgres/postgres.go
```diff
@@ -1121,7 +1121,7 @@ func (c *connection) FindMagicAuthToken(ctx context.Context, id string) (*databa
 	if err != nil {
 		return nil, parseErr("magic auth token", err)
 	}
-	return res.AsModel()
+	return c.magicAuthTokenFromDTO(res)
```
begelundmuller, line 1124: Can we make fetching/decrypting the secret optional? E.g. with a bool. And not do it in `FindMagicAuthToken`, which is only used for checking a token during auth, and can just rely on the secret_hash?

admin/database/database.go
pjain1: Will change it to secret, however its a `string` because encrypt/decrypt method accepts string and return string. So thought one more conversion is unnecessary, however if we want to be consistent with `SecretHash` then can convert back to `[]byte` or change encrypt/decrypt to accept `[]byte` and return `[]byte` but then prod vars needs to be converted to []byte before encrypting, so probably should go with more frequent usage.

admin/database/postgres/postgres.go
pjain1: I can do that but wouldn't it be confusing if `FindMagicAuthToken` being used somewhere else and caller does not get secret even though its in the `MagicAuthToken` struct ? Should we create another struct ?

begelundmuller: Maybe to keep it simple, you could add `withSecret bool` as an arg to `FindMagicAuthToken`?

admin/database/database.go
begelundmuller: Since encryption outputs bytes, it feels like it makes more sense just to store those.

The prod vars just apply base64 on top of the encrypted bytes, right? So you could just have `encryptAsString` and `decryptFromString` or something like that, which wraps the `[]byte` implementation with base64. Or maybe there are simpler approaches, just an example.

begelundmuller (approved):

Issue #5590, opened 2024-09-03
Bug: Azure Blob Storage Authentication Fails During Data Import


marius-hab:
**Describe the bug**
The source connection to Azure Blob Storage does not authenticate successfully. I've set the `AZURE_STORAGE_CONNECTION_STRING` variable in the .env file and ensured I’m logged in using `az login`. However, the connection still fails during the import process.

**To Reproduce**
1. Configure the source in `chat_session.yaml` as follows:

```yaml
# Source YAML
# Reference documentation: https://docs.rilldata.com/reference/project-files/sources

type: source

connector: "duckdb"
account: dainbrainstoragedev
sql: "select * from read_parquet('azure://db-export-container/chat_session.parquet')"
```

2. Attempt to import the source.

**Expected behavior**
The parquet file should be imported without any authentication errors.

**Screenshots**
Below are the logs obtained when running `rill start rill-dashboard`:
```
2024-09-03T13:49:02.679 INFO    Hydrating project '/Users/marius/Documents/projects/dain-brain/rill-dashboard'
2024-09-03T13:49:02.685 INFO    Reconciling resource    {"name": "chat_session", "type": "Source"}
2024-09-03T13:49:02.836 INFO    Ingesting source data   {"name": "chat_session", "connector": "duckdb"}
2024-09-03T13:49:04.490 WARN    Azure Blob Storage account does not have permission to list blobs. Falling back to anonymous access.
2024-09-03T13:49:04.554 WARN    Reconcile failed        {"name": "chat_session", "type": "Source", "elapsed": "1.869s", "error": "failed to ingest source: Object(\"chat_session.parquet\").NewReader: blob (key \"chat_session.parquet\") (code=Unknown): GET https://dainbrainstoragedev.blob.core.windows.net/db-export-container/chat_session.parquet\n--------------------------------------------------------------------------------\nRESPONSE 401: 401 Server failed to authenticate the request. Please refer to the information in the www-authenticate header.\nERROR CODE: NoAuthenticationInformation\n--------------------------------------------------------------------------------\n<?xml version=\"1.0\" encoding=\"utf-8\"?><Error><Code>NoAuthenticationInformation</Code><Message>Server failed to authenticate the request. Please refer to the information in the www-authenticate header.\nRequestId:1ad260e3-101e-0012-7ff7-fdd4f9000000\nTime:2024-09-03T11:49:04.5624505Z</Message></Error>\n--------------------------------------------------------------------------------\n"}

```

**Desktop**
[...]

PR #5626, opened 2024-09-06
fix : allow case insensitive variables

k-anshul:
closes https://github.com/rilldata/rill/issues/5590

**NOTE** : template resolution is still case sensitive


begelundmuller (approved):
This is a nice clean solution!

PR #5630, opened 2024-09-06
Support full templating in YAML connector properties

begelundmuller:
This PR adds supports for arbitrary templating in connector properties defined in a YAML file with `type: connector`. Originally we avoided this due to the performance implications, but it has now become necessary. If we find performance issues, we will need to investigate caching for resolved connector properties.

Among other things, this is needed to support user-specified DuckDB native secrets with `boot_queries` that reference other variables (such as S3 credentials).


runtime/connections.go
```diff
@@ -198,26 +200,20 @@ func (r *Runtime) ConnectorConfig(ctx context.Context, instanceID, name string)
 		}
 	}
 
-	// Search for connector definition in rill.yaml
+	// Search for connector definitions from YAML files
 	vars := inst.ResolveVariables(true)
```
k-anshul, line 204: This will make the template resolution case insensitive. This was okay previously because variables could only refer to `connector....` but now any variable can be referred. 
Though this should still be okay because mostly credentials will be referred.

k-anshul (approved):

runtime/connections.go
begelundmuller: Good point, don't want that! Changed it :)

PR #5631, opened 2024-09-06
DuckDB: Support configuring native secrets in DuckDB

begelundmuller:
This PR adds support for configuring native secrets in DuckDB using `init_sql` in a `duckdb.yaml` connector file.

This enables ingesting data from S3/GCS/Azure from DuckDB *models* (note that this is not needed for DuckDB *sources* due to some legacy hacks in the sources implementation that predate DuckDB's support for native secrets).

**Example:** DuckDB connector configuration that configures DuckDB to use S3 credentials from `~/.aws` on local and from Rill env variables on cloud:

```yaml
# connectors/duckdb.yaml
type: connector
driver: duckdb

init_sql: >
  CREATE OR REPLACE SECRET s3 (
    TYPE S3,
    KEY_ID '{{ .vars.connector.s3.aws_access_key_id }}',
    SECRET '{{ .vars.connector.s3.aws_secret_access_key }}',
    REGION 'eu-north-1'
  )

dev:
  init_sql: >
    CREATE OR REPLACE SECRET s3 (TYPE S3, PROVIDER CREDENTIAL_CHAIN)
```

### DIFF
--- a/scripts/embed_duckdb_ext/main.go
+++ b/scripts/embed_duckdb_ext/main.go
@@ -13,7 +13,7 @@ import (
 )
 
 // DuckDB extensions Rill depends on
-var extensions = []string{"json", "icu", "parquet", "httpfs", "sqlite_scanner", "motherduck"}
+var extensions = []string{"json", "icu", "parquet", "httpfs", "aws", "sqlite_scanner", "motherduck"}
 
 // DuckDB platforms to download extensions for
 var platforms = []string{"linux_amd64", "linux_arm64", "osx_amd64", "osx_arm64"}
--- a/scripts/embed_duckdb_ext/main.go
+++ b/scripts/embed_duckdb_ext/main.go
@@ -13,7 +13,7 @@ import (
 )
 
 // DuckDB extensions Rill depends on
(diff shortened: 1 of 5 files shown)


scripts/embed_duckdb_ext/main.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 // DuckDB extensions Rill depends on
-var extensions = []string{"json", "icu", "parquet", "httpfs", "sqlite_scanner", "motherduck"}
+var extensions = []string{"json", "icu", "parquet", "httpfs", "aws", "sqlite_scanner", "motherduck"}
```
k-anshul, line 16: I know the size is just ~7.5MB but given the use case is not going to be common. I am wondering if we should skip `aws`. Worried about progressively increasing binary size.

k-anshul (approved):

scripts/embed_duckdb_ext/main.go
begelundmuller: Makes sense, we can add it back if/when we move sources to use native credentials

Issue #5217, opened 2024-07-09
Code quality: Remove package `runtime/compilers/rillv1beta`


begelundmuller:

egor-ryashin:
https://github.com/rilldata/rill/pull/5637

PR #5637, opened 2024-09-09
`rillv1beta` removal

egor-ryashin:
Closes https://github.com/rilldata/rill/issues/5217


begelundmuller (changes requested):
See the CI failures

egor-ryashin:
Fixed.

runtime/compilers/rillv1/util.go
```diff
@@ -0,0 +1,29 @@
...
+	// Project variables
+	Variables   map[string]string `yaml:"env,omitempty"`
+	Title       string            `yaml:"title,omitempty"`
+	Description string            `yaml:"description,omitempty"`
+}
+
+func HasRillProject(dir string) bool {
+	_, err := os.Open(filepath.Join(dir, "rill.yaml"))
+	return err == nil
+}
```
begelundmuller, line 20: The goal is that this package doesn't do direct file system reads and instead using the `drivers.RepoStore` interface. 

So the callers should use something like `cmdutil.RepoForProjectPath` and then this function should be changed to accept a repo (i.e. have a type signature similar to `rillv1.ParseRillYAML(...)`

runtime/compilers/rillv1/util.go
```diff
@@ -0,0 +1,29 @@
...
+
+func ParseProjectConfig(content []byte) (*ProjectConfig, error) {
+	c := &ProjectConfig{Variables: make(map[string]string)}
+	if err := yaml.Unmarshal(content, c); err != nil {
+		return nil, err
+	}
+
+	return c, nil
+}
```
begelundmuller, line 29: We should avoid having this `ProjectConfig` type in this package, it will be confusing vs. the existing functionality here. The caller should either be refactored to use `ParseRillYAML` or since it's only one place and only seems to use the `Title` and `Description` fields, it *might* make sense to just inline that parsing functionality there.

begelundmuller (changes requested):

runtime/compilers/rillv1/util.go
egor-ryashin: like this? 
```
func HasRillProject(dir string) bool {
	repo, _, err := RepoForProjectPath(dir)
	_, err = repo.Get(context.Background(), "rill.yaml")
	return err == nil
}
```

egor-ryashin:
All done.

begelundmuller (approved):
LGTM

PR #5639, opened 2024-09-09
Prevent deployments.used_on being bumped by the deployment itself

begelundmuller:


pjain1 (approved):

pjain1:
hehe, is deployment not an entity :)

PR #5653, opened 2024-09-10
feat: Add basic url filter parser

AdityaHegde:
For human readable url params we need a parser for filter param.
Possible expressions,
```
country IN ('US','IN') and state = 'ABC'
country NIN ('US','IN') and (state = 'ABC' or lat >= 12.56)
country NIN ('US','IN') and state having (lat >= 12.56)
```

- [x] Write a grammer and parser for the expression.
- [x] Convert the parsed object to `V1Expressioin`


web-common/src/features/dashboards/url-state/filters/convertors.ts
```diff
@@ -0,0 +1,69 @@
...
+  createOrExpression,
+  createSubQueryExpression,
+  getAllIdentifiers,
+} from "@rilldata/web-common/features/dashboards/stores/filter-utils";
+import { V1Expression, V1Operation } from "@rilldata/web-common/runtime-client";
+import grammar from "./expression.js";
+import nearley from "nearley";
+
+export function convertFilterParamToExpression(filter: string) {
+  const parser = new nearley.Parser(nearley.Grammar.fromCompiled(grammar));
```
ericpgreen2, line 14: Should we re-use a single parser instance, rather than creating a new parser instance every time we call this `convertFilterParamToExpression` function?

web-common/src/features/dashboards/url-state/filters/convertors.ts
```diff
@@ -0,0 +1,69 @@
...
+
+const BinaryOperationMap = {
+  "=": V1Operation.OPERATION_EQ,
+  "!=": V1Operation.OPERATION_NEQ,
+  ">": V1Operation.OPERATION_GT,
+  ">=": V1Operation.OPERATION_GTE,
+  "<": V1Operation.OPERATION_LT,
+  "<=": V1Operation.OPERATION_LTE,
+};
+function convertToExpression(parsed: any) {
```
ericpgreen2, line 27: Can we name this function `convert[Something]ToExpression`?

web-common/src/features/dashboards/url-state/filters/expression.spec.ts
ericpgreen2, line 1: What happens when the string cannot be properly parsed? Can we add some test cases for the failure cases to show what happens? (Given user-provided strings will certainly fail occasionally)

web-common/src/features/dashboards/url-state/filters/convertors.ts
```diff
@@ -0,0 +1,69 @@
...
+  let subExpr: V1Expression | undefined;
+  switch (typeof parsed) {
+    case "object":
```
ericpgreen2, line 30: Given there's only 1 case here, maybe a guard clause would be better?

```ts
if (typeof parsed !== "object") return parsed
```

web-common/src/features/dashboards/url-state/filters/convertors.ts
```diff
@@ -0,0 +1,69 @@
...
+            subExpr,
+          );
+      }
+      break;
+
+    default:
+      return parsed;
+  }
+
+  return undefined;
```
ericpgreen2, line 68: Because neither `parsed` or `undefined` are Expressions, I'm wondering what's it mean when these cases are hit?

web-common/src/features/dashboards/url-state/filters/convertors.ts
```diff
@@ -0,0 +1,69 @@
...
+        case "!=":
+        case ">":
+        case ">=":
+        case "<":
+        case "<=":
+          return createBinaryExpression(
+            parsed[1],
+            BinaryOperationMap[parsed[0]],
+            parsed[2],
+          );
```
ericpgreen2, line 47: Nit: it could be cleaner for all the English operators (IN, NIN, AND, OR, HAVING) to be co-located, and these binary operators could either precede or follow the English operators.

web-common/src/features/dashboards/url-state/filters/convertors.ts
ericpgreen2, line 1: Typo: `convertors` -> `converters`

ericpgreen2 (changes requested):
Overall this is looking good, but left various comments/questions throughout. Additionally, for posterity, a README.md file in the `url-state/filters` directory would be helpful to explain our filter language, nearly, and how to generate the parser.

web-common/src/features/dashboards/url-state/filters/convertors.ts
AdityaHegde: It is meant to be a single use. We can reuse the compiled grammer.

web-common/src/features/dashboards/url-state/filters/convertors.ts
AdityaHegde: Removed this. We can directly create V1Expression instances.

web-common/src/features/dashboards/url-state/filters/convertors.ts
AdityaHegde: Removed

web-common/src/features/dashboards/url-state/filters/convertors.ts
AdityaHegde: Removed

web-common/src/features/dashboards/url-state/filters/convertors.ts
AdityaHegde: Removed

web-common/src/features/dashboards/url-state/filters/expression.spec.ts
AdityaHegde: Added a couple to demonstrate the error. We should get a UX pass eventually.

web-common/src/features/dashboards/url-state/filters/expression.ne
```diff
@@ -0,0 +1,51 @@
+@preprocessor esmodule
+# match an expression used in url filters
+# adapted from https://dev.mysql.com/doc/refman/5.7/en/expressions.html
+@builtin "whitespace.ne"
+@builtin "number.ne"
+@builtin "string.ne"
```
ericpgreen2, line 6: Where are these referenced? I don't see any explicit references, but I'm guessing maybe the primitive types at the bottom (e.g. `sqstring`, `int`, `decimal`) are coming from here?

web-common/src/features/dashboards/url-state/filters/expression.ne
```diff
@@ -0,0 +1,51 @@
...
+
+expr => boolean_expr                             {% id %}
+      | (boolean_expr _ "AND"i _):+ non_and_expr {% andOrPostprocessor %}
+      | (boolean_expr _ "OR"i _):+ non_or_expr   {% andOrPostprocessor %}
+
+# these are used to disambiguate matches
+non_and_expr => boolean_expr                            {% id %}
+              | (boolean_expr __ "OR"i __):+ non_and_expr {% andOrPostprocessor %}
+non_or_expr  => boolean_expr                            {% id %}
+              | (boolean_expr __ "AND"i __):+ non_or_expr {% andOrPostprocessor %}
```
ericpgreen2, line 25: I'm not totally following these. At first glance, it looks like adding these conditions would prevent multiple "ANDs" and multiple "ORs" in the same statement. But I see in the test cases that "X and Y and Z" is possible. So, mind just clarifying in the comment what these do?

ericpgreen2 (approved):

web-common/src/features/dashboards/url-state/filters/expression.ne
```diff
@@ -0,0 +1,51 @@
...
+    binaryPostprocessor,
+    inPostprocessor,
+    havingPostprocessor,
+    andOrPostprocessor,
+  } from "./post-processors.ts";
+%}
+
+expr => boolean_expr                             {% id %}
+      | (boolean_expr _ "AND"i _):+ non_and_expr {% andOrPostprocessor %}
+      | (boolean_expr _ "OR"i _):+ non_or_expr   {% andOrPostprocessor %}
```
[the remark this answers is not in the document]
AdityaHegde: They have the exact same implementation. This is the suggested usage of post post processors, being able to reuse them.

web-common/src/features/dashboards/url-state/filters/expression.ne
AdityaHegde: Yes, the docs arent here for these right now. They point to the source code of these files: https://github.com/kach/nearley/tree/master/builtin

web-common/src/features/dashboards/url-state/filters/expression.ne
AdityaHegde: These are used in `expr` `(boolean_expr _ "AND"i _):+ non_and_expr` for AND. It matches continuous AND ended by a non AND expr.

Added some comments around this.

Issue #5479, opened 2024-08-19
Runtime: Re-ingest local files if they are updated


begelundmuller:
We have gotten several reports about file sources not being updated when the local file is changed. This is especially problematic for `mapping.csv` files used for access management in security policies.

We need to be careful only to do this for small local files, to avoid accidentally scanning large/remote files (even for computing a hash). Maybe it should be an optional parameter for the `file:` connector?

begelundmuller:
I don't have a clear idea about how we might solve this, but here are some thoughts (probably too hacky):
- We could support a `filestore_to_self` executor for DuckDB and ClickHouse for ingesting from the `file`/`local_file` connector
- If `invalidate_on_change` is set to `true` in the `InputProps`, we could compute a hash of the ingested files and store it in the `ModelResult`
- In the call to `ModelManager.Exists`, we could re-compute the hash and return `false` if it changed

begelundmuller:
> We can also consider following alternate solutions if the use case is only for `mapping` files kind of situations:
1. Unfortunately they're often referenced in security policies and dimension/measure expressions, which get hit a ton (e.g. when listing dashboards, we check the security policy of each dashboard in the project), so I'd be too worried about the number of disk reads and CSV parses here. 

2. This might be okay, though it should be used cautiously so it doesn't invalidate large downstream models too frequently. However, there's an issue that applies here, which also applies to my suggestion above, which is that if a model file isn't edited, there is no call to `Reconcile` at all. `Reconcile` is only called on controller restart or when the resource spec changes.

> If we want to solve this for all cases then I think the solution listed above is the only way. Two considerations though :

About 1. here – for ClickHouse, the `file`/`local_file` connector should still load the *local path on the runtime* and upload/write it into a ClickHouse table. Because the goal for the local file connector is using small data files that are checked into Git.

k-anshul:
> About 1. here – for ClickHouse, the file/local_file connector should still load the local path on the runtime and upload/write it into a ClickHouse table. Because the goal for the local file connector is using small data files that are checked into Git.

Sure I think this can be handled separately. I will raise an issue.

PR #5656, opened 2024-09-10
feat: Re-ingest local files on updates

k-anshul:
closes https://github.com/rilldata/rill/issues/5479

Sample model file:
```
connector: local_file
path: "data/a.csv"

output:
  connector: duckdb
  materialize: true
```

**Some considerations**
1. Only works if local file is stored in the repo directory.
2. An input property `invalidate_on_change: false` can also be set to disable automatic updates. By default set to true.


runtime/compilers/rillv1/parse_model.go
```diff
@@ -144,6 +147,31 @@ func (p *Parser) parseModel(node *Node) error {
...
+		path, ok := inputProps["path"].(string)
+		if ok {
+			root, err := p.Repo.Root(ctx)
+			if err != nil {
+				return err
+			}
+			localPaths, err := doublestar.FilepathGlob(filepath.Join(root, path))
+			if err == nil {
+				for _, localPath := range localPaths {
+					localPath = strings.TrimPrefix(localPath, root)
```
begelundmuller, line 161: Could this use `p.Repo.ListRecursive` instead of resorting to a file system dependency? Currently I don't believe the parser has any direct file system dependency (which may come in handy for cloud editing eventually)

runtime/compilers/rillv1/parse_model.go
```diff
@@ -144,6 +147,31 @@ func (p *Parser) parseModel(node *Node) error {
...
+					resources := p.LocalDataToResourcepath[localPath]
+					if _, ok := p.LocalDataToResourcepath[localPath]; !ok {
+						resources = make(map[string]any)
+						p.LocalDataToResourcepath[localPath] = resources
+					}
```
begelundmuller, line 166: More repetition than necessary:
```suggestion
					resources, ok := p.LocalDataToResourcepath[localPath]
					if !ok {
						resources = make(map[string]any)
						p.LocalDataToResourcepath[localPath] = resources
					}
```

runtime/reconcilers/project_parser.go
```diff
@@ -723,3 +731,11 @@ func equalAPISpec(a, b *runtimev1.APISpec) bool {
 func equalConnectorSpec(a, b *runtimev1.ConnectorSpec) bool {
 	return proto.Equal(a, b)
 }
+
+func keys(m map[string]any) []string {
```
begelundmuller, line 735: Use `maps.Keys`?

runtime/reconcilers/project_parser.go
```diff
@@ -196,7 +196,11 @@ func (r *ProjectParserReconciler) Reconcile(ctx context.Context, n *runtimev1.Re
 					}
 					continue
 				}
-
+				paths, ok := parser.LocalDataToResourcepath[e.Path]
+				if ok {
+					changedPaths = append(changedPaths, keys(paths)...)
+					hasDuplicates = true
+				}
```
begelundmuller, line 203: Instead of leaking parser details here, I think this could be implemented inside the parser by adapting `parser.IsSkippable`, and adding the extra changed paths inside the parser's `parsePaths` function

runtime/compilers/rillv1/parser.go
```diff
@@ -178,6 +178,9 @@ type Parser struct {
 	Resources map[ResourceName]*Resource
 	Errors    []*runtimev1.ParseError
 
+	// Mapping of local data files to resource paths that depend on them
+	LocalDataToResourcepath map[string]map[string]any
```
begelundmuller, line 182: See other comment, it would be very nice if this could be kept as internal state

runtime/drivers/duckdb/model_executor_localfile_self.go
```diff
@@ -0,0 +1,156 @@
...
+		return "", err
+	}
+	defer file.Close()
+
+	hasher := md5.New()
+	if _, err := io.Copy(hasher, file); err != nil {
+		return "", err
+	}
+	return hex.EncodeToString(hasher.Sum(nil)), nil
+}
```
begelundmuller, line 156: Wouldn't it be simpler to produce one hash for all the files? You can keep copying file contents into the `hasher`. It anyways needs to reingest all files if one file is changed.

runtime/drivers/duckdb/model_manager.go
```diff
@@ -68,6 +68,9 @@ type ModelResultProperties struct {
 	Table         string `mapstructure:"table"`
 	View          bool   `mapstructure:"view"`
 	UsedModelName bool   `mapstructure:"used_model_name"`
+
+	// LocalFilePathsToHashes is only set for `local_file` input connector.
+	LocalFilePathsToHashes map[string]any `mapstructure:"paths_to_hashes"`
```
begelundmuller, line 73: See other comment – maybe one `LocalFilesHash` is enough

runtime/reconcilers/project_parser.go
```diff
@@ -685,6 +689,10 @@ func equalSourceSpec(a, b *runtimev1.SourceSpec) bool {
 }
 
 func equalModelSpec(a, b *runtimev1.ModelSpec) bool {
+	if a.InputConnector == "local_file" {
+		// hack to make sure that models with `local_file` as input connector are always reconciled
+		return false
+	}
```
begelundmuller, line 695: Could this hack be avoided by moving the computation of the hash into the parser instead of in the executor? So the parser would compute the hash and set `inputProps["filesHash"]`.

begelundmuller (changes requested):

runtime/reconcilers/project_parser.go
k-anshul: We are not on go 1.23 yet.

runtime/compilers/rillv1/parser.go
k-anshul: Yes that makes sense.

runtime/drivers/duckdb/model_executor_localfile_self.go
k-anshul: Not currently. 
If I change one file then only the model that depends on that file will be re-ingested.

runtime/reconcilers/project_parser.go
k-anshul: Yes. Good idea.

runtime/reconcilers/project_parser.go
begelundmuller: Just use `"golang.org/x/exp/maps"` instead then – already used various places in the code

runtime/drivers/duckdb/model_executor_localfile_self.go
begelundmuller: I only meant all the files used by the one model being executed here

k-anshul: Ah okay I misunderstood it is in the executor. 
Yes it can be a single hash.

runtime/compilers/rillv1/parse_model.go
```diff
@@ -106,6 +112,44 @@ func (p *Parser) parseModel(node *Node) error {
 		inputProps["sql"] = sql
 	}
 
+	// special handling to mark model as updated when local file changes
+	if inputConnector == "local_file" {
+		c, ok := inputProps["invalidate_on_change"].(bool)
```
begelundmuller, line 117: If not explicitly set to `false`, should we let it default to `true`? I guess that's the behavior most folks would expect by default.

runtime/compilers/rillv1/parse_model.go
```diff
@@ -106,6 +112,44 @@ func (p *Parser) parseModel(node *Node) error {
...
+			if err == nil && len(entries) > 0 {
+				var localPaths []string
+				root, err := p.Repo.Root(ctx)
+				if err != nil {
+					return err
+				}
+
+				// Update parser's localDataToResourcepath map to track which resources depend on the local file
+				for _, entry := range entries {
+					localPaths = append(localPaths, filepath.Join(root, entry.Path))
```
begelundmuller, line 131: The same parser is only used for one repo, so could it just store the project-relative path? That would reduce dependency on the local filesystem.

I suppose there's a case on local where you can reference a path outside of the project. I would actually prefer we do *not* hash or auto-refresh for files outside the project because a) it's a workflow we should discourage, and b) the watcher doesn't (and shouldn't) watch them.

runtime/compilers/rillv1/parse_model.go
```diff
@@ -106,6 +112,44 @@ func (p *Parser) parseModel(node *Node) error {
...
+					resources, ok := p.localDataToResourcepath[entry.Path]
+					if !ok {
+						resources = make(map[string]any)
```
begelundmuller, line 134: The nesting is getting very deep here, consider pulling the `inputConnector == "local_file"` handling into a separate function to support early return: https://github.com/uber-go/guide/blob/master/style.md#reduce-nesting

runtime/compilers/rillv1/parse_model.go
```diff
@@ -248,3 +292,20 @@ func findLineNumber(text string, pos int) int {
...
 	return lineNumber
 }
+
+func fileHash(paths []string) (string, error) {
+	hasher := md5.New()
+	for _, path := range paths {
+		file, err := os.Open(path)
+		if err != nil {
+			return "", err
+		}
```
begelundmuller, line 302: This again becomes the only place with a hard dependency on a real local file system in the `rillv1` package – any chance we could put this under an interface? I realize using `RepoStore.Get` won't do, but perhaps this could be solved by adding a buffered `GetReader(ctx, path) (io.Reader, error)` function to `RepoStore`?

runtime/compilers/rillv1/parser.go
```diff
@@ -184,6 +184,9 @@ type Parser struct {
 	Resources map[ResourceName]*Resource
 	Errors    []*runtimev1.ParseError
 
+	// Mapping of local data files to resource paths that depend on them
+	localDataToResourcepath map[string]map[string]any
+
 	// Internal state
 	resourcesForPath           map[string][]*Resource // Reverse index of Resource.Paths
 	resourcesForUnspecifiedRef map[string][]*Resource // Reverse index of Resource.rawRefs where kind=ResourceKindUnspecified
```
begelundmuller, line 192: Since the parser already keeps state of the paths for a particular `ResourceName` (through `Parser.Resources[name]`), could it just store the resource names that use a local data file?

Also, can it put it under the `// Internal state` section? Ordering of fields indicate their importance and relationship, and this is for example a less central piece of state than `resourcesForPath`.

 For example:

```suggestion
	// Internal state
	resourcesForPath           map[string][]*Resource // Reverse index of Resource.Paths
	resourcesForUnspecifiedRef map[string][]*Resource // Reverse index of Resource.rawRefs where kind=ResourceKindUnspecified
	resourceNamesForDataPaths map[string][]ResourceName // Index of local data files to resources that depend on them
```

begelundmuller (changes requested):

runtime/compilers/rillv1/parse_model.go
k-anshul: I was just worried about somebody adding a big file in local modelling. But yeah `true` by default can also be set.

runtime/compilers/rillv1/parse_model.go
k-anshul: Yeah I think it will be better to avoid dependency on local file system, I am assuming it is for near term use cases when `repo` may be stored in some database.
Looking from an implementation POV, in the admin driver I see that all operations happen within a lock. Keeping the lock open till the client closes the reader can be risky so I think we can expose a `FileHash(paths []string)` on the interface ?

begelundmuller: Yeah the lock is needed to avoid reading the repo while it is being cloned/pulled.

Having a `FileHash` also seems good to me :)

runtime/compilers/rillv1/parser.go
```diff
@@ -185,8 +185,9 @@ type Parser struct {
 	Errors    []*runtimev1.ParseError
 
 	// Internal state
-	resourcesForPath           map[string][]*Resource // Reverse index of Resource.Paths
-	resourcesForUnspecifiedRef map[string][]*Resource // Reverse index of Resource.rawRefs where kind=ResourceKindUnspecified
+	resourcesForPath           map[string][]*Resource    // Reverse index of Resource.Paths
+	resourcesForUnspecifiedRef map[string][]*Resource    // Reverse index of Resource.rawRefs where kind=ResourceKindUnspecified
+	resourceNamesForDataPaths  map[string][]ResourceName // Index of local data files to resources that depend on them
```
begelundmuller, line 190: How will this get garbage collected? The other index maps here are cleared in `deleteResource` ([link](https://github.com/rilldata/rill/blob/24626af89fb7642801a7d612d4a7c9cf5ee8c0cd/runtime/compilers/rillv1/parser.go#L860))

begelundmuller (changes requested):

runtime/compilers/rillv1/parser.go
k-anshul: Though I think we may still need to clean the `resourceNamesForDataPaths`. I think iterating the full map should be okay. It is not expected to be a very large map (only populated for local_files).

runtime/compilers/rillv1/parse_model.go
```diff
@@ -229,6 +239,43 @@ func (p *Parser) inferSQLRefs(node *Node) ([]ResourceName, error) {
...
+	}
+	path, ok := inputProps["path"].(string)
+	if !ok {
+		return nil
+	}
+	entries, err := p.Repo.ListRecursive(ctx, path, true)
+	if err != nil || len(entries) == 0 {
+		// The actual error will be returned by the model reconciler
+		return nil
+	}
```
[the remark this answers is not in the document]
k-anshul: Yeah can be done.

runtime/compilers/rillv1/parser.go
begelundmuller: Yeah I didn't mean memory garbage collection, just making sure the values are dropped from the map when they stop being correct :)

begelundmuller (approved):
LGTM

Issue #5575, opened 2024-09-02
Feat: Dashboard Width


mindspank:
We should not scale our custom dashboard widths to 100% but rather cap it at reasonable size such as for example 1280 or whatever the current best practice width is.

AC:
* Custom dashboards should have a max-width that is aligning with industry best practices for content heavy websites. Such as NYTimes etc.

PR #5657, opened 2024-09-10
canvas dashboards: set max width

briangregoryholmes:
#5575 

Open questions:

- [ ] What should the max width be? This PR sets it to 1200px.
- [ ] How should the dead space be presented? See below.

Screenshot 2024-09-10 at 9 57 00 AM

### DIFF
--- a/web-common/src/features/canvas-dashboards/CanvasDashboardEmbed.svelte
+++ b/web-common/src/features/canvas-dashboards/CanvasDashboardEmbed.svelte
@@ -45,7 +45,7 @@
 <DashboardWrapper
   bind:contentRect
   {scale}
-  height={maxBottom * gridCell}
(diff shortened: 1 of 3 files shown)


mindspank (approved):

mindspank:
Approved. `1200` looks fine as a width to me and easy to tweak if we'd want different widths.

mindspank:
@briangregoryholmes @ericpgreen2 


Some notes, I don't consider them blockers and we can tweak now that we have the initial stub in place.
- Should we go with 1440 instead?
- Should we scroll the container or the page?
- It looks a bit "dull" but at the same time it wont detract from the analysis. What do you think?

briangregoryholmes:
Bumped to 1440px and changed the scrolling element to the parent. Colors I'll leave for now, but I agree that it looks quite bland.

PR #5660, opened 2024-09-10
lazy load time series data

briangregoryholmes:
Moves responsibility for data fetching to the `ChartWithTotal` component rather than an external store and removes some contextual dependencies from the component so that it can be used outside of a dashboard context.


djbarnwal:
Did some QA, the table and charts are flashing on each selection. The table data apart from the totals row does not change so similar to main, we should be ideally preventing this flickering.

https://github.com/user-attachments/assets/99b297a2-e4e3-49d1-b4e6-9b98ae968ef7

ericpgreen2 (commented):
UXQA: 
1. Go to TDD -> compare by dimension -> attempt to highlight a time range -> app crashes w/ a bug ([Jam](https://jam.dev/c/8e03148e-d7af-4aad-a1dd-1823b6fc2b63))
2. When I load a dashboard with 2 measures, with comparison disabled, I'd expect 2 calls to `/timeseries`, but I see 5:
    image
3. Similarly, when I switch time grains, I see a number of `/timeseries` requests
    image

ericpgreen2 (commented):
When I add/remove filters, a bunch of `timeseries` requests are fired & cancelled ([Jam](https://jam.dev/c/4c6fdf64-7ade-4576-99c7-1843c3d50f02))

image

nishantmonu51:
@briangregoryholmes : Whats the status here ?

briangregoryholmes:
This is being worked on elsewhere

Issue #5488, opened 2024-08-20
Rill Developer: remove all "Importing source" blocking overlays


ericpgreen2:
These blocking overlays pre-date Async Reconcile. There's no need to brick the application with an overlay, as Sources are ingested in the background.

PR #5669, opened 2024-09-11
Rill Developer: Remove blocking overlays from Source ingestion

ericpgreen2:
This PR removes the legacy blocking overlays during Source ingest. Sources are ingested asynchronous, so there's no need to lock up the UI.

This PR does not remove the overlay during local file upload, which currently doesn't have another progress indicator. However, this PR debounces the local file upload overlay so that it does not flash quickly in cases of fast uploads.

Closes https://github.com/rilldata/rill/issues/5488


web-common/src/layout/overlay-store.ts
```diff
@@ -11,4 +9,32 @@ interface Overlay {
   };
 }
 
-export const overlay = writable<Overlay | null>(null);
+const { subscribe, set } = writable<Overlay | null>(null);
+let timeout: NodeJS.Timeout;
+let isCleared: boolean = false;
```
AdityaHegde, line 14: A class works best here since we have variables local to this. Perhaps good to make it a util so that it could be used elsewhere as well?

AdityaHegde (approved):

PR #5673, opened 2024-09-12
fix: alerts without an associated dashboard

AdityaHegde:
Alerts can be created without a dashboard through a YAML file artifact. While we do not have mocks for this, we should atleast not crash the app.

- Redirects users to alert page for emails for such alerts.
- The section in alerts page stating dashboard is removed for such alerts.

### DIFF
--- a/web-admin/src/features/alerts/metadata/AlertMetadata.svelte
+++ b/web-admin/src/features/alerts/metadata/AlertMetadata.svelte
@@ -41,9 +41,7 @@
   $: dashboardName = useAlertDashboardName($runtime.instanceId, alert);
   $: dashboard = useDashboard($runtime.instanceId, $dashboardName.data);
   $: dashboardTitle =
-    $dashboard.data?.metricsView.spec.title ||
-    $dashboardName.data ||
-    $alertQuery.data?.resource?.meta?.name?.name;
+    $dashboard.data?.metricsView.spec.title || $dashboardName.data;
   $: dashboardDoesNotExist = $dashboard.error?.response?.status === 404;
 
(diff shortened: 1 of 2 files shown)


web-admin/src/features/alerts/metadata/AlertMetadata.svelte
```diff
@@ -41,9 +41,7 @@
   $: dashboardName = useAlertDashboardName($runtime.instanceId, alert);
   $: dashboard = useDashboard($runtime.instanceId, $dashboardName.data);
   $: dashboardTitle =
-    $dashboard.data?.metricsView.spec.title ||
-    $dashboardName.data ||
-    $alertQuery.data?.resource?.meta?.name?.name;
+    $dashboard.data?.metricsView.spec.title || $dashboardName.data;
```
ericpgreen2, line 44: A general note – given [this convo](https://rilldata.slack.com/archives/CTCJ58H3M/p1725629184591609?thread_ts=1725577591.865249&cid=CTCJ58H3M), ultimately we should migrate from `metricsView.spec.[attribute]` to `metricsView.state.validSpec.[attribute]`

web-admin/src/routes/[organization]/[project]/-/alerts/[alert]/open/+page.svelte
Comment (ericpgreen2) [inline web-admin/src/routes/[organization]/[project]/-/alerts/[alert]/open/+page.svelte:1]: Another general note – these `open/+page.svelte` files would be good candidates to move to loader functions (i.e. `open/+page.ts`) because they're really just quick redirects.

ericpgreen2 (approved):

PR #5676, opened 2024-09-12
feat : Execute dashboard query as part of health check

k-anshul:
closes https://github.com/rilldata/rill-private-issues/issues/640


runtime/health.go
```diff
@@ -57,6 +52,163 @@ func (r *Runtime) Health(ctx context.Context) (*Health, error) {
...
+	if olap.Dialect() != drivers.DialectClickHouse {
+		return false
+	}
+	m := olap.(drivers.Handle).Config()
+	enabled, ok := m["enable_health_check_cache"].(bool)
+	if !ok {
+		return true
+	}
+	return enabled
+}
```
begelundmuller, line 184: Accessing a config key like this feels a bit hacky – how do you feel about adding something like a `MaybeScaledToZero() bool` on the OLAP driver that internally checks a config key in ClickHouse (which in the future could be powered by actual API calls to CH cloud)?

runtime/health.go
```diff
@@ -57,6 +52,163 @@ func (r *Runtime) Health(ctx context.Context) (*Health, error) {
...
+		return nil, false
+	}
+
+	c := &InstanceHealth{}
+	err = json.Unmarshal(cached.Health, c)
+	if err != nil {
+		return nil, false
+	}
+
+	hash, err := healthResultHash(ctrlVersion, resources)
```
begelundmuller, line 167: This computes the hash using the current controller version, not the one used for the cached health check. 

If the health check itself is refactored to check the cache per-metrics view (instead of for the whole instance) as discussed above, then I think this could be simplified so it doesn't need a hash. Instead, it should store the controller version as part of the cached health check. For example, the `instances` table might have an additional `health_check_controller_version` field or something like that.

Thinking about the high-level goal here, maybe these can be treated as separate steps:
1. Caching the most recent health check
2. Using the cached health check instead of running a new query when a) the controller version hasn't changed, b) the OLAP may be scaled to zero, c) the metrics view is present in the cache.

begelundmuller (changes requested):

runtime/drivers/catalog.go
```diff
@@ -78,3 +81,10 @@ type ModelSplit struct {
 	// Elapsed is the duration of the last execution of the split.
 	Elapsed time.Duration
 }
+
+// InstanceHealth represents the health of an instance.
+type InstanceHealth struct {
+	InstanceID string    `db:"instance_id"`
+	Health     []byte    `db:"health"`
+	CreatedOn  time.Time `db:"created_on"`
+}
```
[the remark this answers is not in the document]
k-anshul: Though the relationship is `1:1` but `health_data` about an instance and instance are separate logical entities.

runtime/health.go
```diff
@@ -2,50 +2,45 @@ package runtime
...
+	"encoding/binary"
+	"encoding/hex"
+	"encoding/json"
 	"errors"
 
 	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
 	"github.com/rilldata/rill/runtime/drivers"
 )
 
+type DashboardHealthQuery func(ctx context.Context, instanceID, name string) (Query, error)
```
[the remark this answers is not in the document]
k-anshul: Yes it was primarily to avoid circular dependencies. But it can also be used to configure multiple queries. 
`Resolver` also sounds good.

runtime/health.go
k-anshul: Actually `MaybeScaledToZero` can improve this logic but we would still need to check this config. A use case could be to wake up the service even if it is scaled to zero for health checks (may be for debugging).
If it feels hacky to access configs outside of clickhouse or olap then a method `HealthCheckCacheEnabled` can be added on the OLAP interface.

Overall I think we should take a call about the `Config` method. May be it should be removed and connections can expose methods to access their configs directly like `Config()` on s3 connection which returns `S3Config` struct.

runtime/health.go
```diff
@@ -57,6 +52,163 @@ func (r *Runtime) Health(ctx context.Context) (*Health, error) {
...
+
+	cacheEnabled := healthCheckCacheEnabled(olap)
+	if cacheEnabled {
+		instanceHealth, ok := r.cachedInstanceHealth(ctx, ctrl.InstanceID, ctrl.catalog.version, resources)
+		if ok {
+			res.OLAP = instanceHealth.OLAP
+			res.Dashboards = instanceHealth.Dashboards
+			return res, nil
+		}
+	}
```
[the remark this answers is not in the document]
k-anshul: Good catch.
I actually thought about multiple connectors for metrics views but skipped it thinking its not going to be common but missed the `olap_connector` part in `rill.yaml`.

runtime/health.go
```diff
@@ -57,6 +52,163 @@ func (r *Runtime) Health(ctx context.Context) (*Health, error) {
...
+
+	err = olap.(drivers.Handle).Ping(ctx)
+	if err != nil {
+		res.OLAP = err.Error()
+	} else if query != nil {
+		// run queries against dashboards
+		res.Dashboards = make(map[string]string, len(resources))
+		for _, mv := range resources {
+			q, err := query(ctx, ctrl.InstanceID, mv.Meta.Name.Name)
```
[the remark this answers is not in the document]
k-anshul: I was also split on this. 
A few times I have seen requests about current health check returning healthy for cases when dashboards aren't healthy so that's why proceeded to check on invalid metrics views.
May be let's throw errors for all to begin with and if we get complaints about too many alerts on user errors we can change this ?

runtime/health.go
k-anshul: > This computes the hash using the current controller version, not the one used for the cached health check.

I did not understand this. The idea was to discard cache if controller is restarted so that's why controller version is part of the hash along with the state version of all metrics views. Below this line we will compare the new hash with the hash stored in the health check.
Though a controller restart will reconcile all metrics views which will update their state version but it should still be okay to keep a controller version if the underlying metrics view reconciler logic is updated.

For caching at a metrics_view level also the current approach works with following changes:
1. `Hash` is just the controller version.
2. If OLAP is not scaled to zero always trigger health check.
3. Along with the cached `metrics_view` health also store the state.version to check whether cache can be used.

runtime/drivers/catalog.go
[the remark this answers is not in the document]
begelundmuller: Subjective, but I don't consider them separate. The latest health check is state about an instance, same way that for example the `deployments` table has `status` and `status_message`.

Databases are able to handle wide tables performantly, and they're usually easier to deal with both in code (no need for upserts or for handling 0 vs. 1 row cases) and in downstream systems.

runtime/health.go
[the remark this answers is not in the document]
begelundmuller: But if the metrics view is invalid (doesn't have `ValidSpec`), the query is guaranteed to fail.

I think parse/reconcile errors are a different case from queries to a valid metrics view not responding to queries, but I agree parse/reconcile errors are also useful to track.

What about if we skip invalid metrics views from the queries here, and instead add `uint parse_errors` and `uint reconcile_errors` to the `InstanceHealth` for reporting if and how many of such errors the instance has? Then we could optionally also monitor for that.

runtime/health.go
begelundmuller: Agree removing the `Config` method makes sense, until then let's avoid using it

runtime/health.go
begelundmuller: > I did not understand this. 

Sorry my bad! I seem to have got the logic wrong here.

> Hash is just the controller version.

Yes – then let's also rename it to `controller_version` for clarity :) 

> Along with the cached metrics_view health also store the state.version to check whether cache can be used.

Makes sense

runtime/drivers/catalog.go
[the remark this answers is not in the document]
k-anshul: I am not sure if a deployment's `status` is equivalent to the instance health which is composed of various parameters like status of various `metrics_view`. The consideration to separate 1:1 mappings are also a function of these nitty-gritty details.

Specific to this current implementation even storing instance health in instances table has few nuances and is not a simple `SELECT`/`INSERT` query.
1. The instance is always updated via a cache which also restarts the controller. We obviously don't want to restart the controller on health check updates.
2. So either we can update instance health separately which means the cache is not in sync with underlying DB which can be error prone.
3. Or I add some extra logic in cache to not restart controller on update of some fields. I am not sure if that solution is less simpler than the `UPSERT` logic which is contained in a single SQL query.

proto/rill/runtime/v1/api.proto
```diff
@@ -263,6 +263,10 @@ message InstanceHealth {
   string controller_error = 1;
   string olap_error = 2;
   string repo_error = 3;
+  map<string, string> metrics_view_errors = 4;
+
+  int32 parse_error_count = 5;
+  int32 reconcile_error_count = 6;
```
begelundmuller, line 269:
```suggestion
  map<string, string> metrics_view_errors = 4;
  int32 parse_error_count = 5;
  int32 reconcile_error_count = 6;
```

runtime/drivers/catalog.go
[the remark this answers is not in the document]
begelundmuller: Okay I see, yeah let's just have a separate table then

runtime/drivers/catalog.go
```diff
@@ -87,3 +90,10 @@ type FindModelSplitsOptions struct {
 	AfterIndex   int
 	AfterKey     string
 }
+
+// InstanceHealth represents the health of an instance.
+type InstanceHealth struct {
+	InstanceID string    `db:"instance_id"`
+	Health     []byte    `db:"health"`
+	CreatedOn  time.Time `db:"created_on"`
```
begelundmuller, line 98: I would propose this schema instead:
```suggestion
	InstanceID string    `db:"instance_id"`
	HealthJSON []byte    `db:"health_json"`
	UpdatedOn  time.Time `db:"updated_on"`
```

- Having `_json` clarifies the field content type when it's not clear from the SQL type
- The semantics here is `updated_on`, not `created_on` (it gets changed after first insertion)

runtime/health.go
```diff
@@ -58,5 +50,183 @@ func (r *Runtime) Health(ctx context.Context) (*Health, error) {
...
+	// run queries against metrics views
+	resources, err := ctrl.List(ctx, ResourceKindMetricsView, "", false)
+	if err != nil {
+		return nil, err
+	}
+	res.MetricsViews = make(map[string]metricsViewHealth, len(resources))
+	for _, mv := range resources {
+		if mv.GetMetricsView().State.ValidSpec == nil {
+			if mv.Meta.ReconcileError != "" {
+				res.ReconcileErrCount++
```
begelundmuller, line 103: This only captures reconcile error counts for metrics views, not other resources?

runtime/health.go
```diff
@@ -58,5 +50,183 @@ func (r *Runtime) Health(ctx context.Context) (*Health, error) {
...
+	if !canScaleToZero {
+		return res, nil
+	}
+
+	// save to cache
+	res.ControllerVersion = ctrl.catalog.version
+	bytes, err := json.Marshal(res)
+	if err != nil {
+		return nil, err
+	}
```
begelundmuller, line 149: It's missing `json:"..."` struct tags on the fields in the struct. Since this is persisted in a DB, it's important to be clear about what the field names will be.

runtime/health.go
```diff
@@ -58,5 +50,183 @@ func (r *Runtime) Health(ctx context.Context) (*Health, error) {
...
+			r.MetricsViewErrors[k] = v.err
+		}
+	}
+	return r
+}
+
+type metricsViewHealth struct {
+	err     string
+	Version int64
 }
```
begelundmuller, line 232: 1. `InstanceHealth` is an exported type, so any types it depends upon should also be.
2. Move this definition up under the `InstanceHealth` definition.
3. The `runtime` package is a large scope with a lot of functionality, so consider calling the type something like `InstanceHealthMetricsViewError`.

runtime/resolvers/metricsview_time_range.go
```diff
@@ -0,0 +1,129 @@
...
+	"io"
+
+	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
+	"github.com/rilldata/rill/runtime"
+	"github.com/rilldata/rill/runtime/pkg/mapstructureutil"
+	"github.com/rilldata/rill/runtime/queries"
+)
+
+func init() {
+	runtime.RegisterResolverInitializer("metricsview_time_range", newMetricsViewTimeRangeResolver)
```
begelundmuller, line 16: Call it `metrics_time_range` for consistency with the other metrics-related resolvers:

```suggestion
	runtime.RegisterResolverInitializer("metrics_time_range", newMetricsViewTimeRangeResolver)
```

runtime/resolvers/metricsview_time_range.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+type metricsViewTimeRangeResolver struct {
+	runtime    *runtime.Runtime
+	instanceID string
+	query      *queries.MetricsViewTimeRange
```
begelundmuller, line 22: It should not use the `queries` package, the goal of the `resolvers` package is to deprecate the `queries` package altogether. It would be better to copy the logic over here, or ideally implement it as a function on `metricsview.Executor` which is called here.

I know it's tricky to navigate partially refactored code, but since it's not possible to justify doing these refactors in one big push, we need to gradually make them as we require new functionality. The place we want to get to is:
- the `resolvers` package handles all data querying/exporting use cases, with consistent support for APIs, reports, alerts, etc.
- the `metricsview` package implements all metrics-related query generation functionality

begelundmuller (changes requested):

admin/worker/deployments_health_check.go
```diff
@@ -179,15 +179,32 @@ func (w *Worker) deploymentHealthCheck(ctx context.Context, d *database.Deployme
...
+			logger.Error("deployment health check: project has parse/reconcile errors", zap.Int32("parse_errors", health.ParseErrorCount), zap.Int32("reconcile_errors", health.ReconcileErrorCount))
+		}
+		for d, err := range health.MetricsViewErrors {
+			logger.Error("deployment health check: dashboard error", zap.String("dashboard", d), zap.String("error", err))
+		}
+
+		onlyUnhealthyDash := true
 		if health.OlapError != "" {
+			onlyUnhealthyDash = false
 			f = append(f, zap.String("olap_error", health.OlapError))
```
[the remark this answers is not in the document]
k-anshul: 1. This is just collecting `zap` fields, the actual log is incorporated into `runtime instance is unhealthy` log only. 
2. Done.

runtime/health.go
k-anshul: I thought we were only interested in the metrics views and not other resources. We can add all resources.

runtime/health.go
```diff
@@ -58,5 +50,183 @@ func (r *Runtime) Health(ctx context.Context) (*Health, error) {
...
+		mvHealth := metricsViewHealth{Version: mv.Meta.StateVersion}
+		if err != nil {
+			mvHealth.err = err.Error()
+		}
+		res.MetricsViews[mv.Meta.Name.Name] = mvHealth
+	}
+
+	if !canScaleToZero {
+		return res, nil
+	}
```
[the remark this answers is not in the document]
k-anshul: The reason I was not storing this for all OLAP engines was because it doesn't handle external tables very well. Till the metrics view is not updated it will keep reporting healthy even if the underlying OLAP engine goes down.

We can improve this logic in subsequent PRs and then we can cache it for all.
1. For other OLAP engines we can invalidate the cache after a TTL.
2. For clickhouse we can invalidate the cache if we are able to retrieve the actual status of the service.
Please let me me know what do you think.

runtime/health.go
k-anshul: Since this is for internal usage only I did not add json tags. The struct field name to json field name is expected to be backward compatible but I can add JSON tags as well.

runtime/health.go
```diff
@@ -15,26 +16,17 @@ type Health struct {
...
-	if h.OLAP != nil {
-		r.OlapError = h.OLAP.Error()
-	}
-	return r
+	// cached
+	OLAP         string
+	MetricsViews map[string]metricsViewHealth
+
+	ControllerVersion int64
 }
```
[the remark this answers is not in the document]
k-anshul: Currently it keeps it keeps cached fields separated from non cached ones which seems beneficial to quickly check what is cached.
```
// InstanceHealth contains health information for a single instance.
// The information about OLAP and metrics views is cached in the catalog.
// We want to avoid hitting the underlying OLAP engine when OLAP engine can scale to zero when no queries are generated within TTL.
// We do not want to keep it running just to check health. In such cases, we use the cached health information.
type InstanceHealth struct {
	// always recomputed on every health check
	Controller        string `json:"-"`
	Repo              string `json:"-"`
	ParseErrCount     int    `json:"-"`
	ReconcileErrCount int    `json:"-"`

	// cached health check information can be used if controller version is same and metrics view spec is same
	OLAP              string                                    `json:"olap"`
	MetricsViews      map[string]InstanceHealthMetricsViewError `json:"metrics_views"`
	ControllerVersion int64                                     `json:"controller_version"`
}
```

runtime/health.go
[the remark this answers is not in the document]
k-anshul: Though I think we can store but not use it here which means it will be refreshed after every health check.

runtime/health.go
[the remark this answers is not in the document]
begelundmuller: 1. I agree it's a noteworthy implementation detail, but I'm not sure it's the main thing guiding the API/interface design here. The approach of the proto message where fields are ordered roughly be severity of the error makes more sense to me, and I think would make sense to mirror internally here. The caching behavior could alternatively be noted in per-field comments instead of a section-level comment.

2. Something else as well – I'll propose not hiding any of the fields from the JSON output, even those that are recomputed. It will be useful to have the DB have an up-to-date snapshot of the latest health information across all these properties.

admin/worker/deployments_health_check.go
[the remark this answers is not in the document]
k-anshul: Okay. Parse and reconcile errors can be clubbed but metrics view errors need to be logged at warn level separately for each metrics view otherwise it losses context that this metrics view has this error. 
I can add a count in the one log line for a quick overview.

runtime/drivers/sqlite/migrations/0026.sql
```diff
@@ -0,0 +1,8 @@
+PRAGMA foreign_keys = ON;
+
+CREATE TABLE IF NOT EXISTS instance_health (
+    instance_id TEXT PRIMARY KEY,
+    health BLOB NOT NULL,
+    created_on TIMESTAMP NOT NULL,
```
begelundmuller, line 6: Should this be named `updated_on`? Since it changes when the health is updated. (It probably doesn't make sense to have a separate `created_on` field as it would just be roughly the same as that of the instance in the `instances` table.)

begelundmuller (approved):

runtime/drivers/sqlite/migrations/0026.sql
k-anshul: Yes. Forgot to update the db column names here. Thanks.

PR #5688, opened 2024-09-13
fix: remove vega validation

mindspank:
- removes vega validation from the runtime
- chart components now use the parsed spec
- remove the ChartTemplate component as it was no longer used.


begelundmuller (approved):
Backend changes look good

PR #5693, opened 2024-09-13
refac: Add canvas dashboard options to breadcrumbs

djbarnwal:
Add canvas option to the breadcrumbs menu.

Canvas options are added at the end of the list.


ericpgreen2:
1. The breadcrumb is empty when I don't provide a Canvas `title` ([see Jam](https://jam.dev/c/37a0fe9c-3689-4eab-9101-c9becf5d32a0)). In this case, we should fallback to the Canvas `name`.
2. Maybe you considered it, but another option is to use/adapt the `useValidVisualizations` function.

ericpgreen2 (approved):
Approving so you can merge when ready.

djbarnwal:
>  Maybe you considered it, but another option is to use/adapt the `useValidVisualizations` function.

We have to treat the resources differently downstream so I suppose it's alright to have their own call methods.

PR #5695, opened 2024-09-13
Ability to unset column sorting

### LINKED ISSUE #5428: Ability to unset column sorting
It's common to have tri-states in table design: unset, ascending, descending.

**Is your feature request related to a problem? Please describe.**
The current pivot table allows users to toggle between ascending and descending sort orders by clicking on a column header. However, clicking a different column header resets the sort order of the previously sorted column. We should allow users to unset a column on the third successive click.

https://github.com/user-attachments/assets/d6fcebe6-9f6b-42cc-bde4-5363ca2c82cb

**Describe the solution you'd like**
Allows users to easily toggle between ascending, descending, and unset states by simply clicking on the column header.

**Additional context**
Check the sorting example from Tanstack: https://tanstack.com/table/latest/docs/framework/svelte/examples/sorting

**Code reference**
Search for `enableSortingRemoval` in the codebase.

lovincyrus:
closes #5428

This PR allows sorting removal in the leaderboard header and pivot table.

https://github.com/user-attachments/assets/3efa1f8f-17f9-41ed-9760-74d7fcde3d10


ericpgreen2:
I should've commented on the original issue, but these unset states feel a little funky to me. The data is always sorted _somehow_, so IMO we should always indicate how the data is sorted. 

In the leaderboard case, it looks like the "unset" case is for the dimensional values to be sorted alphabetically Z->A. We should indicate that & we should make it possible to sort the data A->Z. (Well, this is possible today by entering into the Dimension Detail Table.)

In the pivot case, it looks like the "unset" case is for the data to be sorted by the measure in descending order. I would rather we indicate that than not.

But that's just my opinion! I'll defer to Product on the call. If this is the way we want to go, it looks well implemented :).

jkhwu:
I greenlit the original suggestion to unset a sort in Pivot (https://github.com/rilldata/rill/issues/5428), where the initial state has no arrow indicating sort (perhaps it should have an arrow on the totals column). 

But on leaderboards, the UX is different, because the default sort _should_ be the value column, and reverting to a default of alphabetical sort doesn't serve much use. You could argue that one shouldn't really ever need to see a leaderboard sorted by anything other than value.

Seeking input from @mindspank and @ericokuma . We could either get rid of the unsetting function altogether, or keep it but communicate the default sort state better.

ericokuma:
Is this a valid use case: "I would like to see my customers in ASC order on the leaderboard regardless of value"?

Feels like one but just wanted to double check with folks here

ericpgreen2:
> Is this a valid use case: "I would like to see my customers in ASC order on the leaderboard regardless of value"?

I think it's a valid use case to sort alphabetically by dimension values. It's very closely related to wanting to search for a dimension value.

I'm just not confident that this functionality belongs on a "leaderboard" (in which case the word "leaderboard" kind of loses its meaning), and maybe we should just keep it on the expanded Dimension Detail Table.

lovincyrus:
Reverted and addressed the changes based on https://rilldata.slack.com/archives/C01A9DYP013/p1726521277917949?thread_ts=1726269952.735399&cid=C01A9DYP013 @ericpgreen2

lovincyrus:
Latest demo:


https://github.com/user-attachments/assets/cd008d02-9751-4687-9839-ffe5cbf58e20

PR #5696, opened 2024-09-13
Fix (Alerts): Remove constraint on percent change

ericpgreen2:
Closes https://github.com/rilldata/rill-private-issues/issues/682


nishantmonu51 (approved):
👍

PR #5706, opened 2024-09-16
docs/ welcome to Rill

royendo:
Reorganization of the doc's splash page.

Simpler Welcome to Rill page
- link to tutorial
- link to quick start
- link to release notes

Quickstart & Example
- 60 seconds with Rill video
- Link to projects

FAQ
- Getting Started
- Install Rill
- Rill Develeper
- Rill Cloud
- Rill Cloud Trial


AndrewRTsao (approved):
Impressive update @royendo! 👏 

Given the new FAQ page and the fact that we hit on the Rill Cloud trial, we may want to consider timing this docs update with the self-serve rollout or temporarily redact it for now. Other options could be to reuse the "coming soon" icon here or include some admonition clarifying the Rill Cloud trial. Otherwise, it may lead to some confusion. 

cc @ericokuma on whether we're jumping the gun pushing the docs here and how we may want to time this in conjunction with the introduction of the Rill Cloud trial.

AndrewRTsao (approved):

Issue #5579, opened 2024-09-02
Rill Developer: show a Model's incremental state in the Inspector


ericpgreen2:
[Context in Slack](https://rilldata.slack.com/archives/CTZ8XBQ85/p1719997232795169?thread_ts=1719962827.464359&cid=CTZ8XBQ85)

PR #5720, opened 2024-09-17
Rill Developer: observability for incremental & split Models

ericpgreen2:
This PR adds observability for incremental and split models in the Rill Developer Model workspace. Closes https://github.com/rilldata/rill/issues/5579.

## Model Refresh Button

Previously in Rill Developer, models were only executed after an edit & save. Now, a refresh button can be used to rerun a model at any time.

For incremental models, the button provides two options:

- **Incremental refresh**: Rerun the model with the current state.
- **Full refresh**: Reset any incremental state and rerun the model.

image

---

## Inspector: Incremental Processing

Incremental models now include an "Incremental processing" section in the inspector, showing any data held in the `state`.

image

If the model is incremental but does not leverage any state, an informational message will be displayed.

image

---

## Inspector: Model Splits

Split models now have a "Model splits" section in the inspector. Clicking "View splits" will open a dialog with:

- A table displaying splits, execution timestamps, watermarks, and errors.
- Filtering for pending and errored splits.
- The ability to refresh specific splits.

image


lovincyrus (changes requested):
SplitsTable works as expected from the screenshots. 

On a side note, we should make the header sticky.
CleanShot 2024-09-23 at 00 20 17@2x

lovincyrus (changes requested):
I noticed that the scroller continues to scroll to the bottom of the table even when there are no more rows available. Check `splits_days_incremental.yaml` and `splits_glob.yaml`

https://github.com/user-attachments/assets/de6984f1-0777-4889-a1cc-27dbd2864fee

lovincyrus (approved):
+1

PR #5725, opened 2024-09-18
fix: right align measure headers

mindspank:


mindspank:
Right aligns measure column labels to stay consistent with number alignment.
Dimension column labels stay left aligned.

image

web-common/src/features/dashboards/pivot/PivotTable.svelte
```diff
@@ -459,6 +459,7 @@
                 class="header-cell"
                 class:cursor-pointer={header.column.getCanSort()}
                 class:select-none={header.column.getCanSort()}
+                class:flex-row-reverse={header.column.columnDef.isMeasure}
```
djbarnwal, line 462: As discussed, would be better to use
`class:flex-row-reverse={!hasDimension || i > 0}`
so if there's a dimension column, we want to skip the first column, other wise we apply the class to all columns.

djbarnwal (changes requested):

web-common/src/features/dashboards/pivot/PivotTable.svelte
djbarnwal: In case we only want to apply to the the last header row, we would have to add one more check for that

mindspank: Ye, I would image it's always the last row as measures are the only ones that should right align (technically number dimension columns should probably also do it but type checking the data seems iffy at this point).

djbarnwal (approved):

Issue #5500, opened 2024-08-21
Support access policies on explore and canvas dashboards


royendo:
**Describe the bug**
My dashboard has the following security policy: 

```
security:
  access: "{{ .user.admin }} "
```

So non-admin users cannot see, that works good. However, if I have a custom dashboard that is built using the metrics-view of the dashboard it is still visible in the UI. Any access to the custom dashboard leads to a 404, unexpected error.

**To Reproduce**
1. create dashboard.yaml with security policy
2. Create a rill KPI custom chart based on above dashboard,
3. Add chart to custom dashboard,
4. Deploy to UI

**Expected behavior**
1. Dont show the custom dashboard at all.
2. Show the custom dashboard, but block out the restricted KPI
3. Show the custom dashboard, as normal

It really depends, and maybe this can be an additional setting? It could make sense to have the underlying dashboard blocked but only the KPI showing.

**Screenshots**

mindspank:
@begelundmuller as we discussed when breaking out explore and metrics view we should implement access policies on individual resources such as explore and custom dashboards.

begelundmuller:
@mindspank Agreed. Let's pick this up as part of separating metrics views and dashboards.

PR #5728, opened 2024-09-18
Support `access:` security policies on `Explore` and `Canvas` resources

begelundmuller:
This PR adds support for `access:` security policies on canvas and explore resources. Among other use cases, this enables hiding dashboards that will anyway fail to render for people who don't have access to the underlying metrics views.

A policy on an explore/canvas resource does not prevent direct queries to the underlying metrics views, so it is still recommended to define security policies directly on metrics views. 

Changes:
- Adds support for `access:` security rules on resources of `type: explore` and `type: canvas`.
- Retains the current behavior of *additionally* inheriting the parent metrics view's security rules for `explore` resources.

Examples:
```yaml
# explores/my_explore.yaml
type: explore
title: My Explore
metrics_view: my_metrics
security:
  access: "{{ .user.admin }}"

# canvases/my_explore.yaml
type: canvas
title: My canvas
security:
  access: "{{ .user.admin }}"
```

Closes https://github.com/rilldata/rill/issues/5500.


proto/rill/runtime/v1/resources.proto
```diff
@@ -641,6 +641,7 @@ message CanvasSpec {
   uint32 gap = 3;
   repeated ComponentVariable variables = 5;
   repeated CanvasItem items = 4;
+  repeated SecurityRule security_rules = 6;
```
pjain1, line 644: should this be supported on component level or canvas level ? Seems like component will have mapping with either model or mv so that makes security mapping more appropriate ?

pjain1 (commented):
Shuold this

proto/rill/runtime/v1/resources.proto
begelundmuller: It may also be needed on `Component`s at some point, but the main goal here is to support blocking access to the end dashboards for folks they are not relevant for or that you know may not have access to the data inside them.

Queries made from components are still subject to the underlying API or metrics view's security policies, so not having security policies on the component itself does not leak data. See the PR comment above:

> Among other use cases, this enables hiding dashboards that will anyway fail to render for people who don't have access to the underlying metrics views.
> 
> A policy on an explore/canvas resource does not prevent direct queries to the underlying metrics views, so it is still recommended to define security policies directly on metrics views.

pjain1: I was more thinking on lines if canvas has multiple components accessing different mv and user has security policy on few and not all.

pjain1 (approved):

proto/rill/runtime/v1/resources.proto
begelundmuller: Okay that makes sense but would also involve some work on the canvas implementation side, so will discuss and implement that idea separately

PR #5735, opened 2024-09-19
fix: narrow generated API types

briangregoryholmes:
The lack of required properties in the Swagger yaml causes the generated types to have unnecessary optionals that require avoidable conditional checks on the applications side. This PR fixes that issue for a few of the most impactful properties.

### DIFF
--- a/proto/gen/rill/runtime/v1/runtime.swagger.yaml
+++ b/proto/gen/rill/runtime/v1/runtime.swagger.yaml
@@ -2862,6 +2862,9 @@ definitions:
     title: Selector for a dimension
   MetricsViewSpecDimensionV2:
     type: object
+    required:
+      - name
+      - unnest
     properties:
       name:
         type: string
@@ -2889,6 +2892,14 @@ definitions:
     title: Type of measure query to generate
   MetricsViewSpecMeasureV2:
     type: object
+    required:
+      - name
+      - expression
+      - type
+      - referencedMeasures
+      - validPercentOfTotal
+      - requiredDimensions
+      - perDimensions
     properties:
       name:
         type: string
@@ -4715,8 +4726,12 @@ definitions:
     title: Response message for QueryService.MetricsViewComparison
   v1MetricsViewComparisonRow:
     type: object
+    required:
+      - dimensionValue
+      - measureValues
     properties:
       dimensionValue:
+        type: string
         title: Not optional, not null
       measureValues:
         type: array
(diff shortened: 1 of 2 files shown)


begelundmuller (commented):
Unfortunately, I think these annotations will just be overridden next time we run `make proto.generate`. For the change to persist, I think it would require a config change in `proto/buf.gen.openapi-runtime.yaml` or one of its adjacent files, but I'm not sure if it's possible.

briangregoryholmes:
Looks like there are annotations we can add to the proto files as seen here: (https://grpc-ecosystem.github.io/grpc-gateway/docs/mapping/customizing_openapi_output/). Updated the PR.

briangregoryholmes:
One small issue I'm running into that is somewhat unrelated:

Is it possible to change the type of the `dimension_value` field to a string instead of the broader google.protobuf.Value? This resulted in some type errors via the linting, but I'm not sure how or why this would be returned as anything other than a string.

Relatedly, this is being used to populated the list of possible dimension values when adding a filter, though it doesn't seem designed for that. Is there a better way to get a list of these values (max 100) filtered by a search parameter?

begelundmuller:
> Looks like there are annotations we can add to the proto files as seen here: (https://grpc-ecosystem.github.io/grpc-gateway/docs/mapping/customizing_openapi_output/). Updated the PR.

Yeah this looks promising. Would it be satisfactory for you if we initially limited it's use to the `api.proto` files, especially the response types? 

> One small issue I'm running into:
> 
> Is it possible to change the type of the `dimension_value` field in `MetricsViewComparisonRow` to a `string` instead of the broader `google.protobuf.Value`? Changing it resulted in some type errors via the checks, but I'm not sure how or why this would be returned as anything other than a string. Want to hold off on fixing the errors if there's a reason I'm missing for using the other type.

Dimension values can also be integers or bools IIRC (and a number of other types from the SQL side, though those would become strings in the API response type).

> Relatedly, this `MetricsViewComparisonRow` type (and related query) is being used to populate the list of possible dimension values when adding a filter, though it doesn't seem to be designed for that considering it's also pulling in measure data. Is there a better way to just get a plain list of these values (max 100) filtered by a search parameter?

Yeah, you can use the `MetricsViewAggregation` API for that – it doesn't require requesting a measure.

In fact, the `MetricsViewAggregation` API can do everything that the `MetricsViewComparison`, `MetricsViewTotal` and  `MetricsViewToplist` APIs can do, and should be preferred over those (which will eventually be deprecated). We're also working on making the `MetricsViewAggregation` API support the use cases that `MetricsViewTimeSeries` supports (namely supporting null filling of missing time bins).

nishantmonu51:
@briangregoryholmes : Closing as stale.

PR #5732, opened 2024-09-19
refactor: bookmarks and filter menus

briangregoryholmes:

PR #5741, opened 2024-09-19
refactor: measure menu

briangregoryholmes:
This is a stopgap cleanup in preparation for https://github.com/rilldata/rill/pull/5732.


ericpgreen2 (approved):
UXQA: I'm not seeing any hover state when I mouse over the options in the leaderboard measure dropdown menu. I'd expect a hover state that matches the one in the dimension visibility dropdown menu to its left.

Approving so you can merge when ready.

briangregoryholmes:
Strange. Setting the role to `menuitem` breaks the hover state. Left it as is and changed the selectors in the tests.

PR #5769, opened 2024-09-24
convert add source modal to shadcn

briangregoryholmes:


ericpgreen2 (approved):
UXQA to fix before merging:
1. Z-index issue with the preview table:
    image
2. When on the second page of the dialog, clicking "X" to close the dialog seems to act like a submit button. It triggers yup validation and doesn't close the dialog. See [Jam](https://jam.dev/c/59c5220d-1266-4250-bce4-b69843808ec1)

ericpgreen2:
Looks good, except I'm seeing double validation errors:

image

briangregoryholmes:
This is existing behavior, though I agree it's a bit strange. Will follow up once I look into whether we should display the first or last error.

PR #5772, opened 2024-09-24
fix trial end check

pjain1:


admin/jobs/river/trial_checks.go
```diff
@@ -89,7 +89,7 @@ func (w *TrialEndCheckWorker) Work(ctx context.Context, job *river.Job[TrialEndC
 }
 
 func (w *TrialEndCheckWorker) trialEndCheck(ctx context.Context) error {
-	onTrialOrgs, err := w.admin.DB.FindBillingIssueByTypeNotOverdueProcessed(ctx, database.BillingIssueTypeOnTrial)
+	onTrialOrgs, err := w.admin.DB.FindBillingIssueByType(ctx, database.BillingIssueTypeOnTrial)
```
AdityaHegde, line 92: This could be pulling too many rows. One way to avoid is to use something like `next_check_time`. Index on it and always fetch `next_check_time < now and type=...`.

1. Set it to `now + trial_period - ending_soon_period` during creation initially.
2. In `TrialEndingSoonWorker` after success set it to `next_check_time += ending_soon_period`.
3. Here in `TrialEndCheckWorker` after success we could delete it as we are doing already.

Note that we would also need a sub type on top of this to not select fresh subscriptions in `TrialEndCheckWorker`.

One advantage is that we can use this for other types apart from trial based issues.

pjain1: Don't think its worth the complexity, want to keep it as simple as possible. We anyways typically have page sizes of around 100 or 1000 elsewhere in the code so don't think there will be orgs more than this on trial and we skip orgs for which trial not ended in the for loop. 

Its a background job so may be not a big deal and not all checks are time based so `next_check_time` field will not be generic.

pjain1: Added one optimization where trial end check will only fetch processed on trial issues thus only orgs which are within a week of trial end date.

AdityaHegde (approved):

PR #5777, opened 2024-09-25
update calendar component api

briangregoryholmes:
Updates the `Calendar` component API to accept a single `selection` prop and provides more flexibility around valid/invalid inputs to support a wider number of use cases.


lovincyrus (approved):
ack

Issue #5273, opened 2024-07-15
Feat: Sending reports to non-users


mindspank:
https://www.notion.so/rilldata/Sending-Reports-to-External-Users-3e929be4f9554910b36cc4eae24f30f4?pvs=4

PR #5778, opened 2024-09-25
allow sending reports to external users with magic tokens

pjain1:
Fixes https://github.com/rilldata/rill/issues/5273

- UI changes needs to be done to handle the new external urls for opening and exporting report containing magic token as path param. Urls for existing users will work as it is.

Clean up of unused magic tokens for reports directly created in code (not through UI) needs to be done. Its a bit tricky, may be some background job is required that queries `report_tokens` table and for each `report_name` queries the runtime and then does clean up.


admin/server/reports.go
```diff
@@ -502,6 +561,186 @@ func (s *Server) generateReportName(ctx context.Context, depl *database.Deployme
...
+				"domain": email[strings.LastIndex(email, "@")+1:],
+				"groups": []string{},
+				"admin":  false,
+			}
+		}
+
+		tkn, err := s.admin.IssueMagicAuthToken(cctx, mgcOpts)
+		if err != nil {
+			return nil, fmt.Errorf("failed to issue magic auth token for email %s: %w", email, err)
+		}
```
begelundmuller, line 698: If we just issue a plain magic auth token, then it will also show up in listings of magic auth tokens in the CLI and the UI. This might be confusing for people who wouldn't know it was created automatically for a report.

Not sure, but here are some ideas for fixing this:
- Having an `internal BOOL` on the magic auth tokens which if set hides it in the magic auth token API responses.
- Turning report tokens into full-blown auth tokens by themselves, separate from magic auth tokens (we already have four token types anyway... 😄)

runtime/drivers/admin.go
```diff
@@ -3,14 +3,21 @@ package drivers
 import (
 	"context"
 	"time"
+
+	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
 )
 
 type AdminService interface {
-	GetReportMetadata(ctx context.Context, reportName string, annotations map[string]string, executionTime time.Time) (*ReportMetadata, error)
+	GetReportMetadata(ctx context.Context, reportName string, reportSpec *runtimev1.ReportSpec, executionTime time.Time) (*ReportMetadata, error)
```
begelundmuller, line 11: Could it extract the list of recipients on the caller side, and just pass that here? Just to avoid having report spec handling logic spread too many places.

runtime/drivers/admin.go
```diff
@@ -3,14 +3,21 @@ package drivers
...
 	GetAlertMetadata(ctx context.Context, alertName string, annotations map[string]string, queryForUserID, queryForUserEmail string) (*AlertMetadata, error)
 }
 
 type ReportMetadata struct {
+	InternalUsersURL ReportURLs
+	ExternalUsersURL map[string]ReportURLs
```
begelundmuller, line 17: nit: what about naming the fields like this:

```suggestion
	BaseURLs ReportURLs
	RecipientURLs map[string]ReportURLs
```

Since the base/fallback URLs are also used for things like Slack destinations, which are not really a user recipient

begelundmuller (changes requested):

admin/database/postgres/migrations/0049.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE report_tokens (
+    id UUID DEFAULT uuid_generate_v4() PRIMARY KEY,
+    report_name TEXT NOT NULL,
```
begelundmuller, line 3: Does it need an index on this?

admin/server/projects.go
```diff
@@ -165,7 +166,7 @@ func (s *Server) GetProject(ctx context.Context, req *adminv1.GetProjectRequest)
 		// The magic token's resource
 		condition.WriteString(fmt.Sprintf(" OR '{{.self.kind}}'=%s AND '{{lower .self.name}}'=%s", duckdbsql.EscapeStringValue(mdl.ResourceType), duckdbsql.EscapeStringValue(strings.ToLower(mdl.ResourceName))))
 		// If the magic token's resource is an Explore, we also need to include its underlying metrics view
-		if mdl.ResourceType == runtime.ResourceKindExplore {
+		if mdl.ResourceType == runtime.ResourceKindExplore || mdl.ResourceType == runtime.ResourceKindReport {
```
begelundmuller, line 169: For report tokens, could we avoid needing to grant access to the explore or underlying metrics view if we make the report download API only require access to the report itself?

admin/server/reports.go
```diff
@@ -502,6 +561,154 @@ func (s *Server) generateReportName(ctx context.Context, depl *database.Deployme
...
+
+	// check if magic token already exists for the recipient, recipient list might have changed
+	tkns, err := s.admin.DB.FindReportTokensWithSecret(ctx, reportName)
+	if err != nil {
+		return nil, fmt.Errorf("failed to find report tokens: %w", err)
+	}
+
+	var unusedTknIDs []string
+	for _, tkn := range tkns {
+		if _, ok := emailSet[tkn.RecipientEmail]; ok {
```
begelundmuller, line 580: Should it maybe use `strings.ToLower` when setting/getting values from the email set?

admin/server/reports.go
```diff
@@ -53,10 +55,58 @@ func (s *Server) GetReportMeta(ctx context.Context, req *adminv1.GetReportMetaRe
...
+				ExportUrl: s.admin.URLs.WithCustomDomain(org.CustomDomain).ReportExportExternal(org.Name, proj.Name, req.Report, token),
+			}
+		}
+	} else {
+		// cleanup all report tokens if no external emails if found
+		err = s.cleanUpAllReportTokens(ctx, req.Report)
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "failed to clean up report tokens: %s", err.Error())
+		}
+	}
```
begelundmuller, line 101: Since `issueMagicTokensForExternalEmails` already revokes tokens for emails not present in the input, it basically works like `cleanUpAllReportTokens` when called with an empty list. So maybe the `else` case can be removed and the `if` statement removed?

proto/rill/admin/v1/api.proto
```diff
@@ -1839,14 +1840,19 @@ message GetReportMetaRequest {
   string project_id = 1;
   string branch = 2;
   string report = 3;
-  map<string, string> annotations = 4;
+  rill.runtime.v1.ReportSpec spec = 6;
+  map<string, string> annotations = 4 [deprecated = true];
```
begelundmuller, line 1844: Can this just be removed now? And in the handler just exit early without external user URLs if `spec` is not set. Then the backwards compatibility should be there (and anyway it only matters for the few seconds during release rollout where the runtime is on an older version).

proto/rill/runtime/v1/queries.proto
```diff
@@ -276,6 +284,25 @@ message ExportResponse {
   string download_url_path = 1;
 }
 
+message ExportReportRequest {
+  string instance_id = 1;
+  string name = 2;
```
begelundmuller, line 289: nit: maybe named it `string report` instead? We don't really have a convention around this so it's subjective

proto/rill/runtime/v1/queries.proto
```diff
@@ -276,6 +284,25 @@ message ExportResponse {
...
+message ExportReportResponse {
+  string download_url_path = 1;
+}
+
+message DownloadReportRequest {
+  string instance_id = 1;
+  int64 limit = 2;
+  ExportFormat format = 3;
+  Query query = 4;
+}
```
begelundmuller, line 304: This looks like a request type without a matching RPC. Could it be avoided?

runtime/queries/proto.go
```diff
@@ -62,8 +62,8 @@ func ProtoToQuery(q *runtimev1.Query, claims *runtime.SecurityClaims) (runtime.Q
 	}
 }
 
-// ProtoFromJSON builds a proto query from a query name, JSON args, and optional execution time.
-func ProtoFromJSON(qryName, qryArgsJSON string, executionTime *time.Time) (*runtimev1.Query, error) {
+// ProtoFromJSON builds a proto query from a query name, JSON args, and optional execution time. removeWhere is used for external reports where row filter is already added as additional security rule to the claims.
+func ProtoFromJSON(qryName, qryArgsJSON string, executionTime *time.Time, removeWhere bool) (*runtimev1.Query, error) {
```
begelundmuller, line 66: I think `removeWhere` isn't needed anymore, right?

runtime/server/downloads.go
```diff
@@ -88,6 +93,50 @@ func (s *Server) Export(ctx context.Context, req *runtimev1.ExportRequest) (*run
 	}, nil
 }
 
+func (s *Server) ExportReport(ctx context.Context, req *runtimev1.ExportReportRequest) (*runtimev1.ExportReportResponse, error) {
+	if !auth.GetClaims(ctx).CanInstance(req.InstanceId, auth.ReadMetrics) { // currently all reports are served by mv
+		return nil, ErrForbidden
+	}
```
begelundmuller, line 99: So the idea of the report export API is that it only requires access to the report itself. So instead, I think we can use logic like this on the report resource: https://github.com/rilldata/rill/blob/4d1f20c6a9a048d3bf16292bfce31f732f176b3f/runtime/server/controller.go#L177-L183.

That will trigger the report access logic to be evaluated (which may need some changing for this PR, not sure): https://github.com/rilldata/rill/blob/4d1f20c6a9a048d3bf16292bfce31f732f176b3f/runtime/security.go#L415

runtime/server/downloads.go
```diff
@@ -76,7 +76,12 @@ func (s *Server) Export(ctx context.Context, req *runtimev1.ExportRequest) (*run
 		req.BakedQuery = ""
 	}
 
-	tkn, err := s.generateDownloadToken(req, auth.GetClaims(ctx).SecurityClaims())
+	tkn, err := s.generateDownloadToken(&runtimev1.DownloadReportRequest{
+		InstanceId: req.InstanceId,
+		Limit:      req.Limit,
+		Format:     req.Format,
+		Query:      req.Query,
+	}, auth.GetClaims(ctx).SecurityClaims())
```
begelundmuller, line 84: Feels unintuitive to use a report-specific type for a non-report export.

Maybe the `generateDownloadToken` will need a way of distinguishing between token types? Ideally it could just receive a copy of `ExportReportRequest` then, so we don't need this separate `DownloadReportRequest` type (which from a naming perspective is quite confusing)

begelundmuller (changes requested):

admin/server/projects.go
pjain1: That works for the API however internally an export will be called on the underlying query which will require access to the mv

runtime/server/downloads.go
pjain1: Yeah may be the name of the proto message is not the best that may be changed. However to make this generic, we will have to support different token types ? We can also simply add the `ExportReportRequest` but then report spec will need to be fetched in the `downloadHandler` to resolve the actual query.
```
message ExportReportRequest {
  string instance_id = 1;
  string report = 2;
  int64 limit = 3;
  ExportFormat format = 4;
  google.protobuf.Timestamp execution_time = 5;
}

```

runtime/drivers/admin.go
pjain1: Its also used to get web open state and title, so we may need to pass those as well. In future if we want owner attr as part of this response then we may need to pass that as well. Let me know if you still want to change to just these 3 parameters instead of report spec.

runtime/server/downloads.go
```diff
@@ -23,67 +22,68 @@ import (
...
-		}
+	rep := r.GetReport()
+	t := req.ExecutionTime.AsTime()
 
-		req.Query = qry
-		req.BakedQuery = ""
+	qry, err := queries.ProtoFromJSON(rep.Spec.QueryName, rep.Spec.QueryArgsJson, &t)
+	if err != nil {
+		return nil, fmt.Errorf("failed to build export request: %w", err)
 	}
```
pjain1, line 71: Do we want to keep for one more version for backwards compatibility or small blip is okay during upgrade. I assume the upgrade order is admin, runtime and then UI ?

runtime/drivers/admin.go
begelundmuller: Hmm, maybe I'm misunderstanding, I don't see it accessing open state and title here? https://github.com/rilldata/rill/blob/61d69b09e0b739c53c1f6496a236ada771eb0cb1/admin/server/reports.go#L30

The web open state should just be retrieved by the frontend when looking up the report in the [report open route](https://github.com/rilldata/rill/blob/61d69b09e0b739c53c1f6496a236ada771eb0cb1/admin/urls.go#L308). The title I guess is needed for the email, but that's sent on the runtime side, which already has the report spec, right?

runtime/server/downloads.go
begelundmuller: I think a small blip is alright as long as it doesn't cause a crash (so no reuse of proto numbers for example)

admin/auth_token.go
```diff
@@ -177,6 +177,7 @@ type IssueMagicAuthTokenOptions struct {
 	Fields          []string
 	State           string
 	Title           string
+	Internal        bool
```
begelundmuller, line 180: Also just a though – do you think we should make this a `Type string` so we can also use these magic auth tokens for other use cases in the future (like alerts)? For example, we could have `type=public_url` for the previous usage, and `type=report_export` for the new use case.

runtime/drivers/admin.go
pjain1: I was setting up the state in the magic token for external recipients but yeah front end can directly fetch report resource with the jwt to get the state

admin/auth_token.go
pjain1: That can be done however wouldn't we always need one more table for storing mapping from resource name and recipient to magic tokens like we have for report tokens ? So storing type will be redundant ?

admin/database/database.go
```diff
@@ -660,6 +666,28 @@ type InsertMagicAuthTokenOptions struct {
...
+	RecipientEmail   string `db:"recipient_email"`
+	MagicAuthTokenID string `db:"magic_auth_token_id"`
+}
+
+type ReportTokenWithSecret struct {
+	ID                   string
+	ReportName           string `db:"report_name"`
+	RecipientEmail       string `db:"recipient_email"`
+	MagicAuthTokenID     string `db:"magic_auth_token_id"`
+	MagicAuthTokenSecret []byte `db:"secret"`
```
begelundmuller, line 684: nit: `db:"magic_auth_token_secret"` for consistency?

admin/server/projects.go
```diff
@@ -186,9 +186,11 @@ func (s *Server) GetProject(ctx context.Context, req *adminv1.GetProjectRequest)
 				return nil, status.Errorf(codes.Internal, "could not get resource for magic token: %s", err.Error())
 			}
 
-			spec := resp.Resource.GetExplore().State.ValidSpec
-			if spec != nil {
-				condition.WriteString(fmt.Sprintf(" OR '{{.self.kind}}'='%s' AND '{{lower .self.name}}'=%s", runtime.ResourceKindMetricsView, duckdbsql.EscapeStringValue(strings.ToLower(spec.MetricsView))))
+			if mdl.ResourceType == runtime.ResourceKindExplore {
```
begelundmuller, line 189: I don't think the extra nesting is needed here – this is already checked on line 168

admin/server/reports.go
```diff
@@ -502,6 +553,154 @@ func (s *Server) generateReportName(ctx context.Context, depl *database.Deployme
 	return uuid.New().String(), nil
 }
 
+func (s *Server) issueMagicTokensForExternalEmails(ctx context.Context, orgID, projectID, reportName, ownerID string, emails []string) (map[string]string, error) {
```
begelundmuller, line 556: Since `cleanUpAllReportTokens` does exactly the same as this function does when `len(emails) == 0`, what about removing `cleanUpAllReportTokens` and calling this function with an empty emails slice instead? Could be made clearer by changing the name of this function to `reconcileMagicTokensForExternalEmails` or something like that.

admin/urls.go
```diff
@@ -320,6 +320,18 @@ func (u *URLs) ReportEdit(org, project, report string) string {
...
+func (u *URLs) ReportOpenExternal(org, project, report, token string, executionTime time.Time) string {
+	reportURL := urlutil.MustJoinURL(u.Frontend(), org, project, "-", "reports", report, "open", token)
+	reportURL += fmt.Sprintf("?execution_time=%s", executionTime.UTC().Format(time.RFC3339))
+	return reportURL
+}
+
+// ReportExportExternal returns the URL for exporting a report in the frontend for an external user.
+func (u *URLs) ReportExportExternal(org, project, report, token string) string {
+	return urlutil.MustJoinURL(u.Frontend(), org, project, "-", "reports", report, "export", token)
+}
```
begelundmuller, line 333: 1. Can these be merged into `ReportOpen` and `ReportExport` with `token string` being an optional parameter? Many of the other URL functions here have optional parameters, but currently they try to have one function per route.
2. Did you check with the applications folks if they prefer `<path>/<token>` vs. `<path>/tokens/<token>` vs. `<path>?token=<token>`? I think the latter might be preferable since it's the same route, but not sure.

proto/rill/admin/v1/api.proto
```diff
@@ -1839,14 +1839,20 @@ message GetReportMetaRequest {
   string project_id = 1;
   string branch = 2;
   string report = 3;
-  map<string, string> annotations = 4;
-  google.protobuf.Timestamp execution_time = 5;
+  // 4 was previously used for annotations
+  string ownerID = 5;
```
begelundmuller, line 1843: Note lint error here

proto/rill/admin/v1/api.proto
```diff
@@ -1839,14 +1839,20 @@ message GetReportMetaRequest {
   string project_id = 1;
   string branch = 2;
   string report = 3;
-  map<string, string> annotations = 4;
-  google.protobuf.Timestamp execution_time = 5;
+  // 4 was previously used for annotations
```
begelundmuller, line 1842: Can remove this comment, just put a `reserved` statement on the last line of the message definition (if people need to know why, they can look in the Git history).

proto/rill/admin/v1/api.proto
```diff
@@ -1839,14 +1839,20 @@ message GetReportMetaRequest {
...
+  string ownerID = 5;
+  google.protobuf.Timestamp execution_time = 6;
+  repeated string email_recipients = 7;
 }
 
 message GetReportMetaResponse {
-  string open_url = 1;
-  string export_url = 2;
-  string edit_url = 3;
+  message Urls {
```
begelundmuller, line 1849: Can we call this type `URLs`? That's the naming convention we use elsewhere with manual naming (though auto-generated protos revert them to `Urls`, which is annoying).

begelundmuller (changes requested):

admin/urls.go
pjain1: @ericpgreen2 any thoughts on 2 above. I used `<path>/<token>` to be similar to public urls, but whatever is preferable I can make the change

pjain1: have changed to `<path>?token=<token>` for now

ericpgreen2: Yeah, `<path>?token=<token>` is preferable, thanks

begelundmuller (approved):
LGTM!

runtime/pkg/email/email.go
```diff
@@ -60,9 +61,14 @@ func (c *Client) SendScheduledReport(opts *ScheduledReport) error {
...
 
+	var err error
 	// Resolve template
 	buf := new(bytes.Buffer)
-	err := c.templates.Lookup("scheduled_report.html").Execute(buf, data)
+	if opts.External {
+		err = c.templates.Lookup("scheduled_report_external.html").Execute(buf, data)
+	} else {
+		err = c.templates.Lookup("scheduled_report.html").Execute(buf, data)
+	}
```
begelundmuller, line 71: Did you experiment with using something like `{{ if .OpenLink }}` in the `scheduled_report.mjml` instead of having a separate file? It's slightly tricky because we pre-compile the MJML, so the templating can't impact the HTML structure, but maybe something like this would work?
```html
<mjml>
  <mj-head>
    <mj-style> 
      .hidden {
          display: none !important;
          mso-hide: none;
      }   
    </mj-style>
  </mj-head>
  <mj-body>
    <mj-section>
      <mj-column>
        <mj-button css-class="{{ if empty .OpenLink }} hidden {{ end }}">
          Don't click me!
        </mj-button>
      </mj-column>
    </mj-section>
  </mj-body>
</mjml>
```

web-admin/src/routes/[organization]/[project]/-/reports/[report]/export/+page.svelte
```diff
@@ -5,15 +5,15 @@
   import CtaContentContainer from "@rilldata/web-common/components/calls-to-action/CTAContentContainer.svelte";
```
Comment (pjain1) [inline web-admin/src/routes/[organization]/[project]/-/reports/[report]/export/+page.svelte:5]: @AdityaHegde As per Alex's comment
1. Remove Rill logo for external report users
2. Remove Go To Report Page button for external report users

Comment (pjain1) [reply web-admin/src/routes/[organization]/[project]/-/reports/[report]/export/+page.svelte:5]: May be we still keep Rill logo, not sure.

runtime/pkg/email/email.go
pjain1: awesome it worked! made the change. also removed sending open url from get report metadata call.

runtime/pkg/email/templates/scheduled_report.mjml
```diff
@@ -40,7 +46,7 @@
 
         <mj-spacer height="20px" />
         
-        <mj-text align="center" font-size="12px">
+        <mj-text css-class="{{ if not .EditLink }}hidden{{ end }}" align="center" font-size="12px">
           To edit or unsubscribe from this report, <a href="{{ .EditLink }}">click here</a>.
         </mj-text>
```
begelundmuller, line 51: This is annoying, but we probably still need to find a way to let anonymous recipients unsubscribe from the report.

I guess this could work by setting up an unsubscribe page that will call the `UnsubscribeReport` RPC with the report token (and making the `UnsubscribeReport` RPC support calls with the report token)?

runtime/pkg/email/templates/scheduled_report.mjml
```diff
@@ -26,7 +32,7 @@
 
         <mj-spacer height="20px" />
 
-        <mj-button background-color="#ECF0FF" color="#3524C7" font-weight="bold" href="{{ .OpenLink }}">
+        <mj-button css-class="{{ if not .OpenLink }}hidden{{ end }}" background-color="#ECF0FF" color="#3524C7" font-weight="bold" href="{{ .OpenLink }}">
```
begelundmuller, line 35: Just double checking – did you send test emails both with and without the open link? Just to make sure it still shows for normal users.

runtime/pkg/email/templates/scheduled_report.mjml
pjain1: Modified `UnsubscribeReport` to accept magic token. Now a new page need to be added at `[organization]/[project]/-/reports/[report]/unsubscribe` which will call this rpc with magic token present as query param, the rpc also expects the email of the recipient which will also be present as query param. cc @AdityaHegde

runtime/pkg/email/templates/scheduled_report.mjml
```diff
@@ -41,9 +47,9 @@
         <mj-spacer height="20px" />
         
         <mj-text align="center" font-size="12px">
-          To edit or unsubscribe from this report, <a href="{{ .EditLink }}">click here</a>.
+          To{{ if not .External}} edit or{{ end }} unsubscribe from this report, <a href="{{ .EditLink }}">click here</a>.
```
pjain1, line 50: Added this explicit flag other way is to check if edit link contains `token` but for that will need to define functions to convert `template.URL` to string and also add then use string.contains function. Both these function will need to be added to template func map. So thought this is much cleaner and will also perform better.

runtime/pkg/email/templates/scheduled_report.mjml
pjain1: Found one issue, using magic token one can remove any other email from the list. Need to store info about the magic token recipient or encode email.

pjain1: Fixed - matching the magic token id for report token issued for this email

admin/server/reports.go
```diff
@@ -269,12 +270,21 @@ func (s *Server) UnsubscribeReport(ctx context.Context, req *adminv1.Unsubscribe
...
+	if claims.OwnerType() == auth.OwnerTypeUser {
+		user, err := s.admin.DB.FindUser(ctx, claims.OwnerID())
+		if err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+		userEmail = user.Email
+	} else if req.Email == nil {
+		return nil, status.Error(codes.InvalidArgument, "email is required for magic auth token")
+	} else {
+		userEmail = *req.Email
```
begelundmuller: This doesn't check that the email belongs to the current magic auth token. I think it would be better to do a DB lookup for the current magic auth token's email (can use `claims.OwnerID()` for the magic auth token ID).

admin/urls.go
```diff
@@ -319,8 +319,11 @@ func (u *URLs) ReportExport(org, project, report, token string) string {
 	return exportURL
 }
 
-// ReportEdit returns the URL for editing a report in the frontend.
-func (u *URLs) ReportEdit(org, project, report string) string {
+// ReportEdit returns the URL for editing a report in the frontend or unsubscribe for non-rill recipients.
+func (u *URLs) ReportEdit(org, project, report, recipient, token string) string {
```
begelundmuller, line 323: Can we avoid passing the recipient here and do a lookup instead? It's not so good to have PII in the URL string itself

proto/rill/admin/v1/api.proto
```diff
@@ -1896,6 +1896,7 @@ message UnsubscribeReportRequest {
   string organization = 1;
   string project = 2;
   string name = 3;
+  optional string email = 4 [(validate.rules).string.email = true]; // for non-rill recipients
```
begelundmuller, line 1899: See other comment, I think this is problematic (and can be inferred from the token)

begelundmuller (changes requested):

admin/server/reports.go
```diff
@@ -281,10 +281,22 @@ func (s *Server) UnsubscribeReport(ctx context.Context, req *adminv1.Unsubscribe
...
+
+		reportTkn, err := s.admin.DB.FindReportTokenForEmail(ctx, req.Name, userEmail)
+		if err != nil {
+			return nil, status.Errorf(codes.InvalidArgument, "could not find report token for email: %s", err.Error())
+		}
+
+		if reportTkn.MagicAuthTokenID != claims.OwnerID() {
+			return nil, status.Error(codes.PermissionDenied, "magic token not issued for this email")
+		}
 	}
```
begelundmuller, line 300: If you do a lookup by `claims.OwnerID()` to get the recipient email, then it wouldn't need to pass the user email as a request parameter.

pjain1: `claims.OwnerID()` will give magic auth token id, not sure if I understand this

begelundmuller: That ID can be used to look up the recipient's email address :)

admin/database/postgres/migrations/0049.sql
```diff
@@ -0,0 +1,10 @@
+ALTER TABLE magic_auth_tokens ADD COLUMN internal BOOLEAN NOT NULL DEFAULT FALSE;
+
+CREATE TABLE report_tokens (
+    id UUID DEFAULT uuid_generate_v4() PRIMARY KEY,
+    report_name TEXT NOT NULL,
+    recipient_email TEXT NOT NULL,
+    magic_auth_token_id UUID NOT NULL REFERENCES magic_auth_tokens (id) ON DELETE CASCADE
+);
+
+CREATE INDEX report_tokens_report_name_idx ON report_tokens (report_name);
```
begelundmuller, line 10: Now it also needs an index on `magic_auth_token_id`

begelundmuller (approved):
Note the merge conflict

admin/database/postgres/migrations/0049.sql
pjain1: added

pjain1:
@AdityaHegde does the UI changes need review or are we good to merge ?

ericpgreen2:
> @AdityaHegde @ericpgreen2 does the UI changes need review or are we good to merge ?

I took a quick look & looks pretty good, but I'll do a final review tomorrow.

ericpgreen2:
UXQA:
- For a specific report, my report is blank...
    (Expected) When I click on the email's "Open in browser", I see this data:
    image
    (Expected) When I export this data from the dashboard directly, I see:
    image
    (Unexpected) When I click on the email's "Download CSV file", I see:
    image





- For a project member, I clicked the "Open in browser" link, and I hit this type error. I haven't been able to consistently reproduce this, but it did happen once.
    image
- We're missing the bottom border underneath the top bar: 
    image
    image

pjain1:
@ericpgreen2 for the empty download whats the url ? is it missing the `execution_time` query param, I cannot reproduce that on my setup.

web-admin/src/routes/+layout.ts
ericpgreen2, line 1: Nit: it could be more clear to use the terms `routeToken` and `searchParamToken`, e.g.
```ts
const { organization, project, token: routeToken } = params;
...
let searchParamToken;
if (url.searchParams.has("token")) {
  searchParamToken = url.searchParams.get("token");
}
...
const token = searchParamToken ?? routeToken;
...
```

web-admin/src/routes/[organization]/[project]/-/reports/[report]/export/+page.svelte
```diff
@@ -66,11 +61,13 @@
         </CtaMessage>
       </div>
     {/if}
-    <CtaButton
-      variant="secondary"
-      href={`/${organization}/${project}/-/reports/${reportId}`}
-    >
-      Go to report page
-    </CtaButton>
+    {#if !token}
```
Comment (ericpgreen2) [inline web-admin/src/routes/[organization]/[project]/-/reports/[report]/export/+page.svelte:64]: A comment to explain this conditional would help

ericpgreen2:
> @ericpgreen2 for the empty download whats the url ? is it missing the `execution_time` query param, I cannot reproduce that on my setup.

http://localhost:3000/ericpgreen2/rill-test-covid/-/reports/test-external-recipient-asian-vaccinations-63894747/export?execution_time=2024-10-23T14%3A30%3A00Z&format=EXPORT_FORMAT_CSV

In case it helps to repro, here's my project: https://github.com/ericpgreen2/rill-test-covid

pjain1:
@ericpgreen2 I think its because the watermark used to run this report is trigger time and that is used in time filter so there won't be any data. This should have been the behaviour is past as well. I think if we configure `watermark_inherit` to true then it should use the actual data to figure out watermark but I don't think its configurable through UI.

ericpgreen2:
> @ericpgreen2 I think its because the watermark used to run this report is trigger time and that is used in time filter so there won't be any data. I confirmed this was the behaviour is past as well. I think if we configure `watermark_inherit` to true then it should use the actual data to figure out watermark but I don't think its configurable through UI.

Ah! Makes sense!

pjain1:
Once @ericpgreen2 is ok with the frontend changes, I will resolve the conflicts and merge the PR.

ericpgreen2 (approved):
LGTM

PR #5781, opened 2024-09-25
broaden conditional for readonly filter chips

briangregoryholmes:
Dimensions do not necessarily always have a `column` property and may sometimes, instead, have `expression`. I believe we could safely remove this conditional entirely, though it may have been included for a reason I'm not aware of.


lovincyrus (approved):
makes sense, also to account for this change https://github.com/rilldata/rill-examples/commit/75cadca5fc7daa14ee4801d1e9088b72c7cab004

Issue #5683, opened 2024-09-13
Alerts should not wake up idle clickhouse service


k-anshul:
Running alert queries every 10 minutes for tables created externally can prevent the clickhouse cloud service from becoming idle. 
1. Check the service status before checking for alerts. 
2. Do not trigger alert queries if service is idle.
3. Optionally have a instance var that checks whether to wake up service to check alerts or not.
4. Consider increasing this to 20 minutes since the default time for service idle check is 15 minutes.

Here is a exact vs best-effort implementation of checking service status. The `exact` implementation require more user inputs. The `best-effort` implementation is rudimentary and will only work when both connection max time and alert check time is less than service idle time.

```
// SacaledToZero implements OLAPStore.
// TODO :: clickhouse APIs have very strict rate limits, 10 requests per 10 second per API key
// Consider caching this result if we are going to add more use cases around this.
func (c *connection) ScaledToZero() bool {
	if c.config.APIKeyID == "" {
		// no api key provided resort to a rudimentary solution
		return c.scaleToZeroBasic()
	}
	req, err := http.NewRequest(http.MethodGet, fmt.Sprintf("https://api.clickhouse.cloud/v1/organizations/%s/services/%s", c.config.OrganizationID, c.config.ServiceID), http.NoBody)
	if err != nil {
		c.logger.Warn("failed to create clickhouse cloud API request", zap.Error(err))
		return c.scaleToZeroBasic()
	}
	req.SetBasicAuth(c.config.APIKeyID, c.config.APIKeySecret)

	resp, err := c.cloudAPI.Do(req)
	if err != nil {
		c.logger.Warn("failed to get clickhouse cloud API response", zap.Error(err))
		return c.scaleToZeroBasic()
	}
	defer resp.Body.Close()

	if resp.StatusCode != http.StatusOK {
		c.logger.Warn("failed to get clickhouse cloud API response", zap.Int("status_code", resp.StatusCode))
		return c.scaleToZeroBasic()
	}

	// parse response
	var response struct {
		Result struct {
			State string `json:"state"`
		} `json:"result"`
	}
	err = json.NewDecoder(resp.Body).Decode(&response)
	if err != nil {
		c.logger.Warn("failed to decode clickhouse cloud API response", zap.Error(err))
		return c.scaleToZeroBasic()
	}
	return strings.EqualFold(response.Result.State, "idle")
}

func (c *connection) scaleToZeroBasic() bool {
	stats := c.db.DB.Stats()
[...]

PR #5787, opened 2024-09-26
feat : Alerts should not wake up idle clickhouse service

k-anshul:
closes https://github.com/rilldata/rill/issues/5683


runtime/drivers/clickhouse/olap.go
```diff
@@ -503,7 +505,55 @@ func (c *connection) renameTable(ctx context.Context, oldName, newName, onCluste
...
+		// no api key provided resort to the config set
+		return c.config.CanScaleToZero
+	}
+
+	c.statusCheckMutex.Lock()
+	defer c.statusCheckMutex.Unlock()
+	// check if stauts is cached
+	if !c.statusCheckedAt.IsZero() && time.Since(c.statusCheckedAt) <= time.Minute*10 {
+		return c.scaledToZero
+	}
```
begelundmuller, line 518: I think caching for 10 mins is too long, makes it too likely to lead to a feedback loop that keeps it alive

runtime/drivers/clickhouse/clickhouse.go
```diff
@@ -105,9 +108,17 @@ type configProperties struct {
...
+	TempDir string `mapstructure:"temp_dir"`
+
+	// CanScaleToZero is used to determine if the OLAP instance can be scaled to zero.
+	// In absence of configs required to make requests to the cloud API or in case of API errors service status will be set to this flag.
+	CanScaleToZero bool `mapstructure:"can_scale_to_zero"`
+	// CloudAPIKeyID, CloudAPIKeySecret, OrganizationID, ServiceID are used to make requests to the cloud API
+	APIKeyID       string `mapstructure:"api_key_id"`
+	APIKeySecret   string `mapstructure:"api_key_secret"`
+	OrganizationID string `mapstructure:"organization_id"`
+	ServiceID      string `mapstructure:"service_id"`
```
begelundmuller, line 121: These fields should probably be prefixed with `cloud_` to clarify that they are CH Cloud specific.

begelundmuller (changes requested):
Reflecting on the four CH cloud related properties here, I wonder if this overcomplicates things too much.

Maybe we should just keep it simple and change the `AlertsDefaultStreamingRefreshCron` default to every 6 hours (or something like that) when the DB host matches CH cloud? It's easier to override a cron tab than set all these CH cloud environment variables.

And maybe additionally request a feature from CH that lets us check if the database is scaled down using the credentials in the CH DSN?

k-anshul:
I understand the issues in setting up four different properties that is why I initially rejected this solution to check for service status.

But I am somewhat sceptical about changing the alert check duration to 6 hours by default for all clickhouse cloud, given our positioning as operational BI. For some customers 6 hours can mean that users miss important alerts. 

May be we show this as a message on UI when users are configuring alerts and ask users to change alert duration check. The note can be somewhat `Caution : Rill alert checks can keep the olap service active and prevent it from scaling down to 0. Consider changing when alerts are checked by setting this property...`

We can also be smart about when this is shown. The decision can be taken on two factors:
1. Table should be streaming.
2. OLAP can scale down to zero. We can expose this property as part of the `ListConnectors`.

k-anshul:
For users who have their important data in clickhouse cloud, it is possible that their service never scales to zero. For hobbyist users, it will nudge them to configure alerts check correctly.
If users still want better status checks we can add the actual API based solution.

begelundmuller:
Ultimately the default interval is going to be a trade-off either way. The optimal interval for someone with an always-on cluster is going to be different than for people with one that's mostly scaled to zero.

However I think the current default of 10 mins is definitely optimized for always-on (like Druid), and will not be desirable for folks that scale to zero frequently. If you think 6 hours is too slow, then how about changing the default to maybe 1hr for ClickHouse Cloud?

I think the best we can do is pick a reasonable default and then ask people to override it for their use case – if we make sure to expose the right config keys for overriding, then we can defer the decision of how to prompt for it to the product folks (please follow up with that question in #team-product).

begelundmuller:
Closing this PR because it has been open a long time and probably requires a different approach. If the issue gets higher priority later, we can revisit this work.

PR #5797, opened 2024-09-27
update svelte kit and primary application packages to latest versions

briangregoryholmes:


web-common/.svelte-kit/ambient.d.ts
ericpgreen2, line 1: Looks like we should add `web-common/.svelte-kit` to `.gitignore` so we don't commit these files

ericpgreen2 (approved):
Awesome, Vite 5 is noticeably faster 🙌

PR #5802, opened 2024-09-30
v50 Docs and Tutorials

royendo:


royendo:
v50 docs ready to publish, first pass is complete.

Some pages to review
- concepts/metrics-layer
- manage/settings
- manage/migrations
- build/metrics-view
- build/dashboards (eventually will add canvas dashboards here, since we create a single dashboards folder in RD)

docs/docs/concepts/metrics-layer.md
```diff
@@ -0,0 +1,46 @@
+---
+title: "What is a Metrics View?"
+sidebar_label: "What is a Metrics View?"
+sidebar_position: 11
+hide_table_of_contents: true
+---
+
+What is a metrics layer, and why did we decide to split the dashboard into two individual components?
```
mindspank, line 8: Nit: If I have never used Rill before I don't care why we have done the split. The title and section aims to explain concepts in Rill that's current so it feels a bit odd that the main focus is on why we have done the split instead of explaning what a metrics view is.

royendo: Makes sense! I would say to merge for now and come back to revise this specific page? (I believe Janet is also going to make proper visuals for this page)

mindspank (approved):

PR #5816, opened 2024-10-02
Rill Developer: Add option to generate an Explore Dashboard from Sources, Models, and Tables

ericpgreen2:
"Generate dashboard" is a new option for Sources & Models. It does the following:
1. Creates a Metrics View file (with AI)
2. Creates a basic Explore file that references the new Metrics View
3. Navigates to the Explore Preview

This "Generate dashboard" action now is available in:
- The modal that pops up when a new Source has been ingested
- A Source's context menu in the File Explorer
- A Model's context menu in the File Explorer
- A Table's context menu in the Connector Explorer

Additionally, this PR removes "Generate chart" from the Source & Model menu items. "Generate chart" should be exclusively offered on top of Metrics Views.

### DIFF
--- a/web-common/src/features/metrics-views/ai-generation/generateMetricsView.ts
+++ b/web-common/src/features/metrics-views/ai-generation/generateMetricsView.ts
@@ -5,6 +5,8 @@ import { getScreenNameFromPage } from "@rilldata/web-common/features/file-explor
 import { eventBus } from "@rilldata/web-common/lib/event-bus/event-bus";
 import { get } from "svelte/store";
 import { overlay } from "../../../layout/overlay-store";
+import { queryClient } from "../../../lib/svelte-query/globalQueryClient";
+import { waitUntil } from "../../../lib/waitUtils";
 import { behaviourEvent } from "../../../metrics/initMetrics";
 import type { BehaviourEventMedium } from "../../../metrics/service/BehaviourEventTypes";
 import {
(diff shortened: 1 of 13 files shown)


web-common/src/features/metrics-views/ai-generation/generateMetricsView.ts
```diff
@@ -118,24 +123,64 @@ export function useCreateMetricsViewFromTableUIAction(
...
-      await goto(`/files${newFilePath}`);
-      void behaviourEvent.fireNavigationEvent(
-        newDashboardName,
-        behaviourEventMedium,
-        metricsEventSpace,
-        previousScreenName,
-        MetricsEventScreenName.Dashboard,
-      );
+
+      if (createExplore) {
```
AdityaHegde, line 133: Can we separate this part? This function is doing too much.

AdityaHegde (approved):

web-common/src/features/metrics-views/ai-generation/generateMetricsView.ts
ericpgreen2: Hmm, I agree it's a long function. A couple things:
- We shouldn't put `goto` in another function layer.
- I'm nervous to wrap another layer around `handleEntityCreate`, which itself wraps the actual file creation.

I'd propose that, in general, we need to standardize our various resource-to-resource creation functions. I think we should take this case as part of a broader refactor of `handleEntityCreate` that we can more succinctly use everywhere. This is good motivation.

PR #5817, opened 2024-10-02
Support org without subscription and limit trial orgs

pjain1:
Fixes https://github.com/rilldata/rill-private-issues/issues/668
- Organizations upon creation start with no subscription and never subscribed billing error
- Trial is started either on first project deployment or manually using update subscription API
- Upon sub cancellation, current sub will be valid until current billing cycle end and then no subscription. Quotas will be set to 0 and subscription cancelled billing error will be raised
- Subscription can be renewed again using renew subscription API even before or after cancelled subscription ends. Cannot renew to a trial plan thus payment method and billable address should be present before renewing.
- Simplifies subscription handling by assuming and enforcing single active subscription at all times


pjain1:
hmm...prettier check failing, not sure why

admin/jobs/river/river.go
```diff
@@ -111,12 +114,12 @@ func New(ctx context.Context, dsn string, adm *admin.Service) (jobs.Client, erro
...
 	riverClient, err := river.NewClient(riverpgxv5.New(dbPool), &river.Config{
 		Queues: map[string]river.QueueConfig{
-			river.QueueDefault: {MaxWorkers: 100},
+			river.QueueDefault: {MaxWorkers: 10},
 		},
 		Workers:      workers,
 		PeriodicJobs: periodicJobs,
 		Logger:       logger,
-		JobTimeout:   time.Hour,
+		JobTimeout:   30 * time.Minute,
```
begelundmuller, line 122: Is there a reason for these changes? I believe some of the non-billing jobs have cases where they might run very long

cli/cmd/billing/subscription/edit.go
```diff
@@ -29,14 +29,14 @@ func EditCmd(ch *cmdutil.Helper) *cobra.Command {
 			}
 
 			if subResp.Subscription == nil {
-				ch.PrintfWarn("No subscriptions found for organization %q\n", ch.Org)
-				return nil
+				ch.PrintfWarn("Organization %q has no subscription\n", ch.Org)
+			} else {
+				ch.PrintfBold("Organization %q has following subscription\n", ch.Org)
```
begelundmuller, line 34:
```suggestion
				ch.PrintfBold("Organization %q has the following subscription\n", ch.Org)
```

cli/cmd/billing/subscription/renew.go
```diff
@@ -0,0 +1,63 @@
...
+
+			subResp, err := client.GetBillingSubscription(ctx, &adminv1.GetBillingSubscriptionRequest{
+				Organization: ch.Org,
+			})
+			if err != nil {
+				return err
+			}
+
+			if subResp.Subscription != nil {
+				ch.PrintfBold("Organization %q has following subscription\n", ch.Org)
```
begelundmuller, line 32:
```suggestion
				ch.PrintfBold("Organization %q has the following subscription\n", ch.Org)
```

cli/cmd/billing/subscription/renew.go
```diff
@@ -0,0 +1,63 @@
...
+				ch.PrintSubscriptions([]*adminv1.Subscription{subResp.Subscription})
+
+				ch.PrintfWarn("\nSubscription renewal for %q will take place immediately.\n", ch.Org)
+				ch.PrintfWarn("\nTo edit plan for non cancelled subscription, please use `rill billing subscription edit` command.\n")
```
begelundmuller, line 36:
```suggestion
				ch.PrintfWarn("\nTo edit the plan of non-cancelled subscription, run `rill billing subscription edit`.\n")
```

cli/cmd/billing/subscription/renew.go
```diff
@@ -0,0 +1,63 @@
...
+				ok, err := cmdutil.ConfirmPrompt("Do you want to Continue ?", "", false)
```
begelundmuller, line 37:
```suggestion
				ok, err := cmdutil.ConfirmPrompt("Do you want to continue?", "", false)
```

cli/cmd/billing/subscription/edit.go
```diff
@@ -29,14 +29,14 @@ func EditCmd(ch *cmdutil.Helper) *cobra.Command {
...
+				ch.PrintSubscriptions([]*adminv1.Subscription{subResp.Subscription})
 			}
 
-			ch.PrintfBold("Organization %q has following subscription\n", ch.Org)
-			ch.PrintSubscriptions([]*adminv1.Subscription{subResp.Subscription})
-
 			ch.PrintfWarn("\nEditing plan for organization %q. Plan change will take place immediately.\n", ch.Org)
+			ch.PrintfWarn("\nTo renew a cancelled subscription, please use `rill billing subscription renew` command.\n")
 			ok, err := cmdutil.ConfirmPrompt("Do you want to Continue ?", "", false)
```
begelundmuller, line 40:
```suggestion
			ok, err := cmdutil.ConfirmPrompt("Do you want to continue?", "", false)
```

cli/cmd/billing/subscription/renew.go
```diff
@@ -0,0 +1,63 @@
...
+
+			resp, err := client.RenewBillingSubscription(cmd.Context(), &adminv1.RenewBillingSubscriptionRequest{
+				Organization: ch.Org,
+				PlanName:     plan,
+			})
+			if err != nil {
+				return err
+			}
+
+			ch.PrintfSuccess("Successfully renewed subscription with plan %q for org %q\n", plan, ch.Org)
```
begelundmuller, line 55:
```suggestion
			ch.PrintfSuccess("Successfully renewed subscription to plan %q for org %q\.n", plan, ch.Org)
```

cli/cmd/billing/subscription/renew.go
```diff
@@ -0,0 +1,63 @@
...
+			ch.PrintSubscriptions([]*adminv1.Subscription{resp.Subscription})
+			return nil
+		},
+	}
+	editCmd.Flags().StringVar(&plan, "plan", "", "Plan Name to renew subscription to")
```
begelundmuller, line 60:
```suggestion
	editCmd.Flags().StringVar(&plan, "plan", "", "Plan name to renew subscription to")
```

runtime/pkg/email/email.go
```diff
@@ -427,6 +427,23 @@ func (c *Client) SendSubscriptionEnded(opts *SubscriptionEnded) error {
...
+	TrialEndDate time.Time
+}
+
+func (c *Client) SendTrialStarted(opts *TrialStarted) error {
+	return c.SendInformational(&Informational{
+		ToEmail: opts.ToEmail,
+		ToName:  opts.ToName,
+		Subject: fmt.Sprintf("Your trial for %s has started", opts.OrgName),
+		Title:   fmt.Sprintf("Your trial for %s has started", opts.OrgName),
+		Body:    template.HTML(fmt.Sprintf("Welcome, Your trial for %q has started and will end on %s.", opts.OrgName, opts.TrialEndDate.Format("January 2, 2006"))),
```
begelundmuller, line 443:
```suggestion
		Body:    template.HTML(fmt.Sprintf("Welcome to Rill! Your trial for %q has started and will end on %s.", opts.OrgName, opts.TrialEndDate.Format("January 2, 2006"))),
```

admin/database/postgres/postgres.go
```diff
@@ -734,6 +734,23 @@ func (c *connection) CheckUserIsAProjectMember(ctx context.Context, userID, proj
...
+	return count, nil
+}
+
+func (c *connection) IncrementCurrentTrialOrgCount(ctx context.Context, userID string) error {
+	_, err := c.getDB(ctx).ExecContext(ctx, "UPDATE users SET current_trial_orgs_count = current_trial_orgs_count + 1 WHERE id=$1", userID)
+	if err != nil {
+		return parseErr("org count", err)
+	}
+	return nil
+}
```
begelundmuller, line 752: Could this be computed with a join query instead of tracked separately? It seems it won't be called often, so even if the query is inefficient it should be fine.

If not, shouldn't it be decremented when a trial expires or is upgraded?

begelundmuller (changes requested):

admin/jobs/river/river.go
pjain1: During my testing I found that if process is killed during a running river job then the job gets stuck in running mode even after the restart and only fail when on timeout so thought of decreasing it. I checked and it did not seem that it can take much time, but I can revert it.

admin/database/postgres/postgres.go
pjain1: The reason its being tracked separately because this quota is across the lifetime of user and once they have started two trial orgs then they cannot have another one even if previous ones were upgraded to paid plan. So no way to know at the moment how many orgs were on trial in past, user can directly upgrade to paid plan without going through trial and in that case this won't be incremented. Same reason why we don't need to decrement this.

admin/jobs/river/river.go
begelundmuller: Ah, that sounds like an issue with our graceful shutdowns. Maybe we don't wait for River to stop before stopping the process?

If that doesn't fix it, we may need to think about alternative solutions, either splitting long-running jobs into many small jobs, or seeing if there's a config to retry scheduled jobs earlier than their timeout in case of restarts.

@kaspersjo might have thoughts on this.

This can be handled in a follow up PR

admin/database/postgres/postgres.go
begelundmuller: Okay, didn't realize it's a lifetime thing.

Was it a request to make it a lifetime thing? Since we will probably forget to change it for each internal user, it sounds like this will lead to quota issues when folks are doing live demoes and POCs in the future.

pjain1: Yes it was a requirement from product cc @ericokuma

begelundmuller (approved):

PR #5818, opened 2024-10-02
Azure: Fix panic from nil iterator being returned without an error

begelundmuller:
Fixes the issue raised by [this alert](https://rilldata.slack.com/archives/C05TXA7QSTY/p1727868844887949).

### DIFF
--- a/runtime/drivers/gcs/object_store.go
+++ b/runtime/drivers/gcs/object_store.go
@@ -84,8 +84,8 @@ func (c *Connection) DownloadFiles(ctx context.Context, props map[string]any) (d
 
 	iter, err := rillblob.NewIterator(ctx, bucketObj, opts, c.logger)
 	if err != nil {
-		apiError := &googleapi.Error{}
 		// in cases when no creds are passed
+		apiError := &googleapi.Error{}
 		if errors.As(err, &apiError) && apiError.Code == http.StatusUnauthorized {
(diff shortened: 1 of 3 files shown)


runtime/drivers/gcs/object_store.go
```diff
@@ -97,12 +97,8 @@ func (c *Connection) DownloadFiles(ctx context.Context, props map[string]any) (d
 			return nil, drivers.NewPermissionDeniedError(fmt.Sprintf("can't access remote err: %v", retrieveError))
 		}
 
-		anonIt, anonErr := rillblob.NewIterator(ctx, bucketObj, opts, c.logger)
-		if anonErr == nil {
-			iter = anonIt
-			err = nil
-		}
```
pjain1, line 104: This was added explicitly by this [PR](https://github.com/rilldata/rill/pull/5598) to retrieve from public bucket so do we revert it, does creds work for gcs public buckets ?

begelundmuller: Yeah but it looks like it was added by mistake, if you look at the surrounding code, it doesn't change any of the parameters, it just repeats the previous authenticated request.

begelundmuller: I'm pretty sure GCS doesn't require separate auth for public buckets, I believe they can be accessed normally with your normal credentials.

pjain1: Yes it doesn't look like its creating some anon request

pjain1 (approved):

PR #5819, opened 2024-10-02
feat: header split

briangregoryholmes:
Creates a new global `ApplicationHeader` and updates the workspace application designs to use the new panelized layout and asset header.


web-common/src/components/forms/Input.svelte
```diff
@@ -218,27 +238,52 @@
...
 
   .multiline-input {
-    @apply overflow-auto break-words py-[5px] cursor-text;
+    @apply py-1;
+    line-height: 1.6;
   }
 
   input {
-    @apply h-[30px] py-0;
+    /* @apply pt-[7px] pb-[6px]; */
```
ericpgreen2, line 281: Leftover comment

web-common/src/features/entity-management/resource-icon-mapping.ts
```diff
@@ -21,4 +23,22 @@ export const resourceIconMapping = {
   [ResourceKind.Theme]: ThemeIcon,
   [ResourceKind.Report]: ReportIcon,
   [ResourceKind.Alert]: AlertCircleOutline,
+  [ResourceKind.ProjectConfig]: Settings,
+  file: FileIcon,
+};
+
+export const resourceColorMapping = {
```
ericpgreen2, line 30: Nice! We should use this in the sidebar too (but we can also do that in a different PR)

ericpgreen2 (commented):
Overall, this is great!

UXQA:
1. I noticed when a model is errored, the inspector disappears ([Jam](https://jam.dev/c/9fb3d749-01bd-4a93-a685-9b1349b8c63e)). Is that intentional? In the past, I think the inspector went blank, which wasn't a great state either, but this in its own way feels jumpy when I'm iterating on SQL. In the Metrics Editor, we show a little message "Fix the errors in the file to continue." Maybe we should standardize on that for now.
    image
2. @jkhwu, I feel all the "active" states on the panel controls (e.g. open/close inspector) compete for my attention and try to draw my focus away from the center of the screen. Thoughts on removing the buttons' "active" background color?
    image
3. @jkhwu, the "unsaved file" indicator to the right of the header looks mis-placed (it's too far to the right, but it's there because of the "rename" context menu)
    image

web-common/src/layout/ApplicationHeader.svelte
```diff
@@ -0,0 +1,189 @@
...
+      {/if}
+    {:else if mode === "Developer"}
+      <Button
+        square
+        type="secondary"
+        href={previewHref}
+        disabled={!previewHref}
+      >
+        <Play size="16px" />
+      </Button>
```
ericpgreen2, line 176: I noticed `features/explores/PreviewButton.svelte` can be deleted now

ericpgreen2 (approved):
Looks good! Some UXQA nits before merging:
1. When the Preview button is disabled, I'm not seeing the tooltip explaining why
    image
2. It'd also be helpful to have tooltips for these icon buttons:
    image
3. I'm seeing a `cursor-pointer` when I hover over the gap between the icon and the file name header, but there's no click interaction. Also, I think this gap might be a couple pixels too wide.
    image
4. The icons in the "Add [resource]" menu could use color too:
    image

PR #5827, opened 2024-10-03
Fix criteria form

lovincyrus:
Fix https://www.notion.so/rilldata/v0-50-0-QA-114ba33c8f5780bd897ecb99547ed623#114ba33c8f5780c29dbedd81e2cf7f8e

CleanShot 2024-10-03 at 16 02 37@2x

### DIFF
--- a/web-common/src/components/forms/Input.svelte
+++ b/web-common/src/components/forms/Input.svelte
@@ -43,7 +43,7 @@
   export let multiline = false;
   export let fontFamily = "inherit";
   export let sameWidth = false;
-  export let width: string | null = "100%";
+  export let width: string = "100%";
   export let textClass = "text-xs";
   export let enableSearch = false;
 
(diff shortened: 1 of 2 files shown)


briangregoryholmes:
The Notion doc is referring to a different element of the form. I'm also not able to reproduce the Notion screenshot to begin with.

As for the last element, it's not really that you want width to be "null", but rather "auto" or some fixed value like "120px". I'd rather not accept null here.

That said, I do see some issues with the Select component not always respecting the width property. I can look into that.

A larger issue is that this form does not appear to function properly. The measure name dropdown for each criteria only ever has one option and the ids are in conflict, so the triggers open up menus in the wrong places (though they do work). That latter one may have only surfaced when moving to shadcn a week or two back.

briangregoryholmes (approved):

Issue #5564, opened 2024-08-30
Feat: BigNumber and always show important digits


mindspank:
https://www.notion.so/rilldata/Expand-the-number-of-visible-digits-for-big-metrics-in-Explore-Overview-TTD-and-other-surfaces-28371608b0924fd99be243110b12f32c?pvs=4

PR #5832, opened 2024-10-06
Implement table formatting rules

djbarnwal:
Part of https://github.com/rilldata/rill/issues/5564

Remaining Item 
- [ ] Use updated tooltip values for all table context
- [ ] Update big number context rules
- [ ] Always format delta change

Leaderboard columns adjust based on data value size
image

Tooltips are more human readable
image

Implement Table formatting rules

image

image

https://www.notion.so/rilldata/Formatting-the-display-of-metrics-for-higher-precision-in-tables-and-other-contexts-28371608b0924fd99be243110b12f32c?d=10aba33c8f5780f98bbb001c0d0f6bf9#104ba33c8f578069b643cdb346dd2225


ericpgreen2:
> Display up to 12 digits with a comma-separated format, rounding to the nearest integer with no decimal precision (but still showing the decimal point)

I see [in the Notion doc](https://www.notion.so/rilldata/Formatting-the-display-of-metrics-for-higher-precision-in-tables-and-other-contexts-28371608b0924fd99be243110b12f32c?pvs=4#b113438bfa074f9eae0926dace567700) that the decimal point should _not_ be shown in this case (which is what I would expect).

web-common/src/features/dashboards/leaderboard/leaderboard-widths.ts
```diff
@@ -0,0 +1,91 @@
...
+      ),
+      delta: Math.max(currentWidths.delta, newWidths.delta),
+      deltaPercent: Math.max(
+        currentWidths.deltaPercent,
+        newWidths.deltaPercent,
+      ),
+    }));
+    processedDimensions.add(dimensionName);
+  }
+}
```
ericpgreen2, line 51: It could be simpler to make `updateMaxColumnWidths` a pure function that internalizes `processedDimensions` and returns `columnWidths`. It feels like `processedDimensions` is an implementation detail.

ericpgreen2 (changes requested):
UXQA:
1. I'm seeing "default" show up as an option in the Visual Metrics Editor. Per your code comment (and the runtime's validation), "default" isn't a valid option, it's only an internal helper, right?
    image
2. In an extreme case (using the `none` formatting option w/ many significant digits), I'm seeing overlapping numbers. While maybe an edge case, it'd be good to avoid overlapping digits.
    image
3. I think I'd expect that the icon in the column header would always be right-aligned with the rows below. Here, there are a number of rows where digits spill to the right of the icon (and could be the cause for the overlapping digits above)
    image

web-common/src/features/dashboards/leaderboard/leaderboard-widths.ts
ericpgreen2: Oh, now I realize I'm conflating widths within a single leaderboard & widths across all leaderboards. Maybe there's opportunity to more clearly distinguish between the two levels. And a pure function here could still help.

djbarnwal:
Have removed `default` as a preset option in favor of `none` which was being deprecated. This way we also preserve backward compatibility without introducing a new preset.

web-common/src/lib/number-formatting/strategies/per-range-default-options.ts
```diff
@@ -4,6 +4,45 @@ import {
   NumberKind,
 } from "../humanizer-types";
 
+export const defaultNoFormattingOptions: FormatterOptionsCommon &
```
ericpgreen2, line 7: This object doesn't strike me as "options", but rather a selected configuration.

web-common/src/lib/number-formatting/strategies/per-range-tooltip-options.ts
```diff
@@ -0,0 +1,58 @@
...
+};
+
+export const tooltipCurrencyOptions = (
+  numberKind: NumberKind,
+): FormatterOptionsCommon & FormatterRangeSpecsStrategy => ({
+  rangeSpecs: tooltipRangeSpec,
+  defaultMaxDigitsRight: 2,
+  upperCaseEForExponent: true,
+  numberKind,
+});
```
ericpgreen2, line 58: Likewise, these strike me as a selected configuration, not "options".

ericpgreen2 (approved):

web-local/tests/explores/number-formatting.spec.ts
```diff
@@ -82,13 +82,13 @@ dimensions:
...
-      ["No Format", "300.6k", "300576.83999999857"],
-      ["percentage", "30.1M%", "300576.83999999857"],
+      ["default format", "100.0k", "100,000"],
+      ["USD", "$300.6k", "$300,576.84"],
+      ["humanized chosen", "300.6k", "300,576.84"],
+      ["No Format", "300.6k", "300,576.84"],
+      ["percentage", "30.1M%", "30.1M%"],
       ["interval_ms", "5 m", "5m 576ms"],
-      ["d3 fixed", "300.6k", "300576.840"],
+      ["d3 fixed", "300.6k", "300,576.84"],
```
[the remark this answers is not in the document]
djbarnwal: `default` is referenced in terms of there being no preset defined. Will update this to `no preset format`.

web-common/src/lib/number-formatting/strategies/per-range-tooltip-options.ts
djbarnwal: Followed existing terminology here. Usage of `option` is across old types and objects as well. Will update this in a follow up.

PR #5835, opened 2024-10-07
refactor: dashboard menus

briangregoryholmes:
- Removes use of `WithTogglableFloatingElement` for certain components
- Removes the `showHideStore` and instead uses existing features of the global StateManagers
- Creates a new `DashboardVisibilityDropdown` component for controlling visibility of measures/dimensions
- Updates the dimension value query that populates the `DimensionFilter` dropdown to use the `MetricsViewAggregation` API instead of `MetricsViewComparisonRow`
- Does some consolidation of our "searchable" menus, though this is an ongoing project larger than this PR
- Makes some minor prop changes to Filter bar components to enable fixes to the bookmark UX (follow up PR)


ericpgreen2 (changes requested):
When I select a subset of dimensions, then refresh the page, the state isn't retained ([Jam](https://jam.dev/c/0c0df0d1-6a7e-4652-83f4-03f4b98ffc43)). There's a chance this was an older bug from `main` that has since been fixed, so I'd recommend merging `main` just in case.

ericpgreen2:
I'm still seeing this bug: 
- I select a subset of dimensions & measures
- I refresh the page
- I see the full set of dimensions & measures, not my subset

[See Jam](https://jam.dev/c/bd271ddb-5b5f-4686-9685-5087707fa0db)

briangregoryholmes:
Okay. Should be fixed now.

PR #5857, opened 2024-10-09
fix: explore state on rill developer not reset with explicit name on edit

AdityaHegde:
Editing an explore should reset the preview state to experience any change made. We were resetting using the explore name inferred from file path. If the explore contained an explicit name then it would fail to reset.

### DIFF
--- a/web-common/src/features/explores/ExploreEditor.svelte
+++ b/web-common/src/features/explores/ExploreEditor.svelte
@@ -5,6 +5,7 @@
   import { createPersistentDashboardStore } from "@rilldata/web-common/features/dashboards/stores/persistent-dashboard-state";
   import Editor from "@rilldata/web-common/features/editor/Editor.svelte";
   import { FileArtifact } from "@rilldata/web-common/features/entity-management/file-artifact";
+  import { findResourceNameInYAML } from "@rilldata/web-common/features/entity-management/infer-resource-kind";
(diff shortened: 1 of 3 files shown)


web-common/src/features/explores/ExploreEditor.svelte
```diff
@@ -35,10 +36,12 @@
     bind:autoSave
     bind:editor
     onSave={(content) => {
+      const name = findResourceNameInYAML(content) ?? exploreName;
+      console.log(name);
```
ericpgreen2, line 40: Leftover log

web-common/src/features/explores/ExploreEditor.svelte
```diff
@@ -35,10 +36,12 @@
     bind:autoSave
     bind:editor
     onSave={(content) => {
+      const name = findResourceNameInYAML(content) ?? exploreName;
```
ericpgreen2, line 39: We should fix `exploreName` at the top of the component tree, not just in this one place. It looks like `ExploreWorkspace.svelte` line 26 first defines `exploreName`. It should pull the name from the `meta` in the `GetResource` query, else maybe fallback to the filename.

ericpgreen2 (changes requested):

ericpgreen2 (approved):

PR #5858, opened 2024-10-09
fix: inferred resource kind for explore under dashboards folder

AdityaHegde:


web-common/src/features/entity-management/file-artifact.ts
```diff
@@ -111,12 +111,12 @@ export class FileArtifact {
   }
 
   updateRemoteContent = (newContent: string, alert = true) => {
-    this.remoteContent.set(newContent);
     if (alert && newContent !== get(this.remoteContent)) {
       for (const callback of this.remoteCallbacks) {
         callback(newContent);
       }
     }
+    this.remoteContent.set(newContent);
```
ericpgreen2, line 119: This makes me a little nervous – semantically, the callbacks should happen _after_ the remoteContent changes not _before_.

ericpgreen2 (approved):

web-common/src/features/entity-management/file-artifact.ts
AdityaHegde: Good point. `newContent !== get(this.remoteContent)` check should happen before updating content and firing callback should happen after.

web-common/src/features/entity-management/file-artifact.ts
```diff
@@ -313,6 +313,10 @@ export class FileArtifact {
       kind === ResourceKind.MetricsView &&
       resource.meta?.name?.kind === ResourceKind.Explore
     ) {
+      if (!get(this.remoteContent)) {
+        // inferred incorrectly since we didnt check contents
+        void this.fetchContent();
+      }
```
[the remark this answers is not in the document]
AdityaHegde: Fetching content is unnecessary in cases other. Added a comment to clarify things.

web-common/src/features/entity-management/file-artifact.ts
ericpgreen2: Nice, I really like the new approach

web-common/src/features/entity-management/file-artifact.ts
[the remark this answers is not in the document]
ericpgreen2: Gotcha, I didn't realize this would get called for all resources (and their files) on initialization. Agreed, best to not fetch all files on startup.

PR #5861, opened 2024-10-09
optimize druid query for complex measure

pjain1:
Fixes https://github.com/rilldata/rill-private-issues/issues/732

### DIFF
--- a/runtime/metricsview/executor_rewrite_druid_groups.go
+++ b/runtime/metricsview/executor_rewrite_druid_groups.go
@@ -17,8 +17,9 @@ func (e *Executor) rewriteDruidGroups(ast *AST) error {
 }
 
 func (e *Executor) rewriteDruidGroupsWalk(n *SelectNode) error {
-	// Skip if already grouped
-	if n.Group {
+	// Skip if already grouped or if there is no sub query with JOIN
+	// Druid requires GROUP BY and ANY_VALUE aggregate for measure if there is a subquery with JOIN
+	if n.Group || (n.SpineSelect == nil && len(n.LeftJoinSelects) == 0 && n.JoinComparisonSelect == nil) {
 		return nil
 	}
 
@@ -30,12 +31,6 @@ func (e *Executor) rewriteDruidGroupsWalk(n *SelectNode) error {
 	}
 
 	// Recurse
-	if n.FromSelect != nil {
-		err := e.rewriteDruidGroupsWalk(n.FromSelect)
-		if err != nil {
-			return err
-		}
-	}
 	if n.SpineSelect != nil {
 		err := e.rewriteDruidGroupsWalk(n.SpineSelect)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


begelundmuller (approved):
LGTM

runtime/metricsview/executor_rewrite_druid_groups.go
```diff
@@ -30,12 +31,6 @@ func (e *Executor) rewriteDruidGroupsWalk(n *SelectNode) error {
 	}
 
 	// Recurse
-	if n.FromSelect != nil {
-		err := e.rewriteDruidGroupsWalk(n.FromSelect)
-		if err != nil {
-			return err
-		}
-	}
```
begelundmuller, line 38: Or wait, shouldn't it recurse here? The `FromSelect` might have nested joins

runtime/metricsview/executor_rewrite_druid_groups.go
```diff
@@ -17,8 +17,9 @@ func (e *Executor) rewriteDruidGroups(ast *AST) error {
 }
 
 func (e *Executor) rewriteDruidGroupsWalk(n *SelectNode) error {
-	// Skip if already grouped
-	if n.Group {
+	// Skip if already grouped or if there is no sub query with JOIN
+	// Druid requires GROUP BY and ANY_VALUE aggregate for measure if there is a subquery with JOIN
+	if n.Group || (n.SpineSelect == nil && len(n.LeftJoinSelects) == 0 && n.JoinComparisonSelect == nil) {
```
begelundmuller, line 22: And what about `SpineSelect`? That is `RIGHT JOIN`ed to the query

begelundmuller (changes requested):

runtime/metricsview/executor_rewrite_druid_groups.go
pjain1: Just checked with measure filter, it seems like we may not add group by here however the query is getting converted to a new `windowOperator` query and I am already hitting some edge cases with this query. We may need to look into upgrading druid cluster to get latest version of this query, till then lets not make change for this.

runtime/metricsview/executor_rewrite_druid_groups.go
```diff
@@ -22,6 +22,20 @@ func (e *Executor) rewriteDruidGroupsWalk(n *SelectNode) error {
 		return nil
 	}
 
+	// Recurse over FromSelect as it may contain sub queries with joins
+	if n.FromSelect != nil {
+		err := e.rewriteDruidGroupsWalk(n.FromSelect)
+		if err != nil {
+			return err
+		}
+	}
```
begelundmuller, line 31: Now there are two "// Recurse ..." sections in this function – what about just moving the `// Recurse` section above the `// Rewrite` section, and having a `var hasJoins bool` that gets set in the join recurse cases?

begelundmuller (approved):

PR #5862, opened 2024-10-09
File Explorer: Remove client-side file rename constraint

ericpgreen2:
Previously, client-side, we were preventing files from being re-named to the same name as any other file in the project. Now, we merely prevent files from being re-named to the same name as another file in the same directory.

Addresses [this UXQA item](https://www.notion.so/rilldata/v0-50-0-QA-114ba33c8f5780bd897ecb99547ed623?pvs=4#119ba33c8f5780b082aafc64283468e3)

### DIFF
--- a/web-common/src/features/canvas-components/ComponentsHeader.svelte
+++ b/web-common/src/features/canvas-components/ComponentsHeader.svelte
@@ -6,7 +6,7 @@
   import { fileArtifacts } from "@rilldata/web-common/features/entity-management/file-artifacts";
   import {
     extractFileName,
-    splitFolderAndName,
+    splitFolderAndFileName,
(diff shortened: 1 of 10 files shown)


briangregoryholmes:
Looks like there may be an edge case where you have an old metrics view and a new Explore with the same name (but in different folders). No collision error is displayed, and only the dashboard for the new Explore is accessible.

briangregoryholmes (approved):

ericpgreen2:
> There appears to be an edge case where you have an old metrics view and a new Explore with the same name (but in different folders). No collision error is displayed, and only the dashboard for the new Explore is accessible.
 
Hmm, there should be a parser error stating the name collision. I'll track that in the UXQA doc for a follow-up.

PR #5877, opened 2024-10-10
fix: inferred resource behavior

briangregoryholmes:
- We cannot assume that files in the `dashboards` folder are of type MetricsView
- If a resource kind has been previously determined, do not overwrite that when it can no longer be inferred
- Adds a check to the `VisualMetrics` surface to ensure the YAML items in a sequence are of the map type

### DIFF
--- a/web-common/src/features/entity-management/entity-mappers.ts
+++ b/web-common/src/features/entity-management/entity-mappers.ts
@@ -124,7 +124,7 @@ export function addLeadingSlash(path: string): string {
 export const FolderNameToResourceKind: Record<string, ResourceKind> = {
   sources: ResourceKind.Source,
   models: ResourceKind.Model,
-  dashboards: ResourceKind.MetricsView,
+  // dashboards: ResourceKind.MetricsView,
   metrics: ResourceKind.MetricsView,
   // explores: ResourceKind.Explore, // This does not happen on backend
   charts: ResourceKind.Component,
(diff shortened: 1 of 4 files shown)


web-common/src/features/entity-management/entity-mappers.ts
```diff
@@ -124,7 +124,7 @@ export function addLeadingSlash(path: string): string {
 export const FolderNameToResourceKind: Record<string, ResourceKind> = {
   sources: ResourceKind.Source,
   models: ResourceKind.Model,
-  dashboards: ResourceKind.MetricsView,
+  // dashboards: ResourceKind.MetricsView,
```
ericpgreen2, line 127: Eh for backwards-compatibility, I think we need to retain this for now. Many existing projects have legacy metrics views in the `dashboards` directory. 

On this branch, if I edit one of these legacy Metrics Views, then I'll be exited from the Metrics View workspace ([see Jam](https://jam.dev/c/0699d5f8-ff20-4ac8-b075-5a1400d6d720)). On main, I'll be kept within the Metrics View workspace, which 1) provides context of the Inspector and 2) shows me the parse error (though in theory we could also show parse errors in the fallback editor) ([See Jam](https://jam.dev/c/9fab340c-9dcd-400e-9b7d-34b94a6e6bca)).

ericpgreen2: I suppose if we solved 1 by adding parse errors to the fallback file editor (though this has UX implications for non-Rill files), then losing the Metrics Inspector context wouldn't be the end-of-the-world.

ericpgreen2 (approved):

PR #5888, opened 2024-10-11
Use photo url in users table

lovincyrus:
CleanShot 2024-10-11 at 12 49 18@2x


ericpgreen2 (changes requested):
As discussed, pending https://github.com/rilldata/rill/issues/5889, we should also get the photoURL for the users who are not the one currently logged-in.

ericpgreen2 (approved):

Issue #5053, opened 2024-06-11
Runtime: Code quality burndown


begelundmuller:

nishantmonu51:
This is expected to be running for next few sprints.

PR #5901, opened 2024-10-15
Set `format_preset: humanize` when generating metrics views

begelundmuller:


k-anshul (approved):
Approving the PR but two unrelated comments:
1. Our docs indicate that the default format is `humanize` so is it possible that we have a gap in implementation either in docs or in code ?
2. Is it possible to create a enum for all presets?

begelundmuller:
> 1. Our docs indicate that the default format is `humanize` so is it possible that we have a gap in implementation either in docs or in code ?

@djbarnwal says they are changing this default to unformatted, so that's why this change was needed.

> 2. Is it possible to create a enum for all presets?

Good idea, added it to the code quality [backlog](https://github.com/rilldata/rill/issues/5053)

PR #5909, opened 2024-10-15
fix: alternate connector support for visual metrics

briangregoryholmes:


medriscoll:
Just adding to this per a Slack thread with @ericpgreen2 -- will this allow the drop-down menu to be populated correctly in the Visual Metrics Editor?  (For projects that reference a CH table, the drop-down is currently null). 

image

briangregoryholmes:
Yes

web-common/src/features/entity-management/resource-selectors.ts
```diff
@@ -44,7 +44,7 @@ export function prettyResourceKind(kind: string) {
 export function useResource<T = V1Resource>(
   instanceId: string,
   name: string,
-  kind: ResourceKind,
+  kind: ResourceKind | undefined,
```
ericpgreen2, line 47: Can we avoid setting this to `undefined`? It feels like it breaks the semantics

ericpgreen2 (approved):
LGTM! Just one nice-to-have would be a memory of the last Model/Table selected, so that when you toggle back-and-forth between Model/Table, you go back to your previous state.

PR #5912, opened 2024-10-15
Update embedding.md

royendo:
Adding a final tab on parameter to let customer know they can pass other attributes.

Based on discussion with Swyft on embed dashboards.


mindspank (changes requested):
You would use the `attributes` property which is already in the list. Such as

```
attributes: {
  email: Foo, 
  tenantid: bar
}

royendo:
modified the webpage for the above comments!

mindspank (approved):

PR #5920, opened 2024-10-16
fix quota trial orgs

pjain1:


begelundmuller (approved):
Will this quota get set to `-1` when using the noop biller?

pjain1:
This quota is not depended on biller and currently is getting set to -1 which is default in DB. After this patch it will be set to 2 which is default in code.

Issue #5804, opened 2024-09-30
APIs for Cloud Environment Variable management


ericpgreen2:
https://www.notion.so/rilldata/Environment-Variables-Cloud-Management-Table-API-Design-10cba33c8f578098acb3c0aebce18bcd

begelundmuller:
Updated the document with a detailed proposal here: https://www.notion.so/rilldata/Platform-support-for-environment-variables-UI-10cba33c8f578098acb3c0aebce18bcd.

PR #5929, opened 2024-10-17
APIs for Cloud Environment Variable management

k-anshul:
closes https://github.com/rilldata/rill/issues/5804


admin/database/postgres/migrations/0049.sql
```diff
@@ -0,0 +1,46 @@
...
+  project_data.project_id,
+  key AS name,
+  convert_to(value, 'UTF8')::bytea AS value,
+  project_data.prod_variables_encryption_key_id AS value_encryption_key_id
+FROM
+  project_data,
+  jsonb_each_text(project_data.prod_variables);
+
+-- Drop the old column
+ALTER TABLE projects DROP COLUMN prod_variables;
```
k-anshul, line 45: May be drop this in next release ?

cli/cmd/env/configure.go
```diff
@@ -95,6 +78,7 @@ func ConfigureCmd(ch *cmdutil.Helper) *cobra.Command {
 	configureCommand.Flags().SortFlags = false
 	configureCommand.Flags().StringVar(&projectPath, "path", ".", "Project directory")
 	configureCommand.Flags().StringVar(&projectName, "project", "", "")
+	configureCommand.Flags().StringVar(&environment, "environment", "production", "Environment for which variables apply (options: 'production', 'development', '')")
```
k-anshul, line 81: NOTE : The older CLIs will set environment as empty.

cli/cmd/env/rm.go
```diff
@@ -8,7 +8,7 @@ import (
 
 // RmCmd is sub command for env. Removes the variable for a project
 func RmCmd(ch *cmdutil.Helper) *cobra.Command {
```
k-anshul, line 10: The `rm` command from older CLIs will no longer work. As per the older behaviour it will delete removed variable and send remaining variable.
The only option I see it to create a `v2` API.

begelundmuller: It's not a commonly used command, so it's okay to just require folks to upgrade. You can reject requests to the RPC from old CLIs using this: https://github.com/rilldata/rill/blob/2396d5c36384a911502db944d3f7a4142738c5d1/admin/server/server.go#L41

k-anshul: Sounds good. 
But the same method is also used in `rill env configure` which is somewhat essential for most deploys. This means most users will need to upgrade their CLIs but that should be okay since we handle it gracefully now.

admin/database/postgres/migrations/0049.sql
begelundmuller: Sounds like a good idea. You will need to also keep it on the struct to avoid missing field errors (since some queries use `SELECT *` to query project rows).

admin/projects.go
```diff
@@ -207,6 +205,88 @@ func (s *Service) UpdateProject(ctx context.Context, proj *database.Project, opt
...
 
+// UpdateProjectVariables updates a project's variables and runs reconcile on the deployments.
+func (s *Service) UpdateProjectVariables(ctx context.Context, project *database.Project, environment string, vars map[string]string, unsetVars []string, userID string) ([]*database.ProjectVariable, error) {
+	if len(vars) == 0 && len(unsetVars) == 0 {
+		return nil, nil
+	}
+	txCtx, tx, err := s.DB.NewTx(ctx)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
```
begelundmuller, line 216: Rollback should be safe to call after commit, so it can be defer called here for better safety: https://go.dev/doc/database/execute-transactions#example

admin/projects.go
```diff
@@ -207,6 +205,88 @@ func (s *Service) UpdateProject(ctx context.Context, proj *database.Project, opt
...
+	if err != nil {
+		return nil, err
+	}
+
+	if len(vars) == 0 {
+		// edge case : no prod variables to set (variable was deleted)
+		// but the runtime does not update variables if the new map is empty
+		// so we need to set a dummy variable to trigger the update
+		vars = map[string]string{"nonce": time.Now().Format(time.RFC3339Nano)}
+	}
```
begelundmuller, line 269: `nonce` is too likely to collide with a user-provided variable in the long run. Maybe `rill.internal.nonce`? 

Also, shouldn't this be set in `ResolveProdVariables` for consistent handling?

admin/server/projects.go
```diff
@@ -680,42 +681,62 @@ func (s *Server) GetProjectVariables(ctx context.Context, req *adminv1.GetProjec
 		return nil, status.Error(codes.PermissionDenied, "does not have permission to read project variables")
 	}
 
-	return &adminv1.GetProjectVariablesResponse{Variables: proj.ProdVariables}, nil
+	var environment *string
+	if !req.ForAllEnvironments {
+		environment = &req.Environment
+	}
+	vars, err := s.admin.DB.FindProjectVariablesByEnvironment(ctx, proj.ID, environment)
```
begelundmuller, line 688: I think the expected behavior here is that it resolves the variables for the provided environment (i.e. also incorporating variables that are not environment-specific), not that it returns only the variables configured for that environment.

Resolving enables use of this endpoint to fetch variables for a specific environment, e.g. for getting all variables for the "dev" environment on local.

admin/projects.go
```diff
@@ -434,3 +517,23 @@ func (s *Service) TriggerParserAndAwaitResource(ctx context.Context, depl *datab
...
+	for _, v := range vars {
+		switch v.Environment {
+		case "":
+			// only add default variable if it doesn't already exist to not override production variables
+			if _, ok := res[v.Name]; !ok {
+				res[v.Name] = string(v.Value)
+			}
+		case "production":
+			res[v.Name] = string(v.Value)
+		}
```
begelundmuller, line 536: 1. Should this do case insensitive comparisons? 
2. The environment is called `"prod"`, not `"production"`: https://github.com/rilldata/rill/blob/2396d5c36384a911502db944d3f7a4142738c5d1/admin/deployments.go#L173

admin/database/postgres/migrations/0049.sql
```diff
@@ -0,0 +1,47 @@
+CREATE TABLE project_variables (
+    id UUID NOT NULL DEFAULT uuid_generate_v4() PRIMARY KEY,
+    project_id UUID NOT NULL REFERENCES projects(id) ON DELETE CASCADE,
+    name TEXT NOT NULL,
+    value BYTEA NOT NULL,
+    -- Encryption key ID if the value is encrypted
+    value_encryption_key_id TEXT NOT NULL DEFAULT '',
+    -- Environment it belongs to ("production" or "development").
```
begelundmuller, line 8: The built-in environments are currently called `"prod" or "dev"` (not fully spelled out)

admin/server/projects.go
```diff
@@ -680,42 +681,62 @@ func (s *Server) GetProjectVariables(ctx context.Context, req *adminv1.GetProjec
...
-		Subpath:              proj.Subpath,
-		ProdVariables:        req.Variables,
-		ProdDeploymentID:     proj.ProdDeploymentID,
-		ProdSlots:            proj.ProdSlots,
-		ProdTTLSeconds:       proj.ProdTTLSeconds,
-		Provisioner:          proj.Provisioner,
-		Annotations:          proj.Annotations,
-	})
+	if len(req.Variables) == 0 && len(req.UnsetVariables) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "no variables or unset_variables provided")
```
begelundmuller, line 727: I don't think this is necessarily an error, it can just return early?

admin/server/projects.go
```diff
@@ -680,42 +681,62 @@ func (s *Server) GetProjectVariables(ctx context.Context, req *adminv1.GetProjec
...
 	if err != nil {
-		return nil, status.Errorf(codes.Internal, "variables updated failed with error %s", err.Error())
+		return nil, status.Error(codes.Internal, err.Error())
 	}
 
-	return &adminv1.UpdateProjectVariablesResponse{Variables: proj.ProdVariables}, nil
+	resp := &adminv1.UpdateProjectVariablesResponse{}
+	for _, v := range vars {
+		resp.Variables = append(resp.Variables, projectVariableToDTO(v))
+	}
```
begelundmuller, line 738: Wouldn't the expected behavior here be that it returns all variables, not just the changed ones?

Since the endpoint semantics treats the variables as a single collection (even if the underlying database table doesn't).

cli/cmd/env/pull.go
```diff
@@ -105,6 +111,7 @@ func PullCmd(ch *cmdutil.Helper) *cobra.Command {
 
 	pullCmd.Flags().StringVar(&projectPath, "path", ".", "Project directory")
 	pullCmd.Flags().StringVar(&projectName, "project", "", "Cloud project name (will attempt to infer from Git remote if not provided)")
+	pullCmd.Flags().StringVar(&environment, "environment", "", "Environment for which variables apply (options: 'production', 'development', '')")
```
begelundmuller, line 114: `prod` or `dev`

cli/cmd/env/push.go
```diff
@@ -111,8 +113,9 @@ func PushCmd(ch *cmdutil.Helper) *cobra.Command {
 		},
 	}
 
-	pullCmd.Flags().StringVar(&projectPath, "path", ".", "Project directory")
-	pullCmd.Flags().StringVar(&projectName, "project", "", "Cloud project name (will attempt to infer from Git remote if not provided)")
+	pushCmd.Flags().StringVar(&projectPath, "path", ".", "Project directory")
+	pushCmd.Flags().StringVar(&projectName, "project", "", "Cloud project name (will attempt to infer from Git remote if not provided)")
+	pushCmd.Flags().StringVar(&environment, "environment", "", "Environment for which variables apply (options: 'production', 'development', '')")
```
begelundmuller, line 118: `prod` or `dev`

cli/cmd/env/show.go
```diff
@@ -28,14 +28,20 @@ func ShowCmd(ch *cmdutil.Helper) *cobra.Command {
...
 				return err
 			}
 
-			res, err := godotenv.Marshal(resp.Variables)
+			vars := make(map[string]string)
+			for _, v := range resp.Variables {
+				vars[v.Name] = v.Value
+			}
+
+			res, err := godotenv.Marshal(vars)
```
begelundmuller, line 44: I think if `--environment` is not passed, it should list all variable values for all environments (since that's what the UI does). So this probably needs to be replaced with a table that shows name, value and environment.

cli/cmd/env/show.go
```diff
@@ -48,6 +54,7 @@ func ShowCmd(ch *cmdutil.Helper) *cobra.Command {
 
 	showCmd.Flags().StringVar(&projectName, "project", "", "Cloud project name (will attempt to infer from Git remote if not provided)")
 	showCmd.Flags().StringVar(&projectPath, "path", ".", "Project directory")
+	showCmd.Flags().StringVar(&environment, "environment", "", "Environment for which variables apply (options: 'production', 'development', '')")
```
begelundmuller, line 57: `prod` or `dev`

cli/cmd/env/configure.go
```diff
@@ -95,6 +78,7 @@ func ConfigureCmd(ch *cmdutil.Helper) *cobra.Command {
 	configureCommand.Flags().SortFlags = false
 	configureCommand.Flags().StringVar(&projectPath, "path", ".", "Project directory")
 	configureCommand.Flags().StringVar(&projectName, "project", "", "")
+	configureCommand.Flags().StringVar(&environment, "environment", "", "Environment for which variables apply (options: 'production', 'development', '')")
```
begelundmuller, line 81: The options are `dev` and `prod`:

```suggestion
	configureCommand.Flags().StringVar(&environment, "environment", "", "Optional environment to resolve for (options: dev, prod)")
```

admin/database/database.go
```diff
@@ -276,6 +276,10 @@ type DB interface {
 	UpdateBillingIssueOverdueAsProcessed(ctx context.Context, id string) error
 	DeleteBillingIssue(ctx context.Context, id string) error
 	DeleteBillingIssueByTypeForOrg(ctx context.Context, orgID string, errorType BillingIssueType) error
+
+	FindProjectVariablesByEnvironment(ctx context.Context, projectID string, environment *string) ([]*ProjectVariable, error)
+	UpsertProjectVariable(ctx context.Context, projectID, environment string, vars map[string]string, userID string) ([]*ProjectVariable, error)
```
begelundmuller, line 281: As a reflection on this PR, I would consider breaking this into separate `InsertProjectVariable` and `UpdateProjectVariable` functions and making multiple calls from the API resolver.

The external API interface may pretend they are one set, but with multiple rows in the DB it makes the queries simpler (and performance/optimization is not really a concern on such small values anyway).

begelundmuller (changes requested):

admin/database/postgres/postgres.go
```diff
@@ -2060,25 +2049,108 @@ func (c *connection) DeleteBillingIssueByTypeForOrg(ctx context.Context, orgID s
...
+			Environment:          environment,
+			Name:                 key,
+			Value:                encryptedValue,
+			ValueEncryptionKeyID: valueEncryptionKeyID,
+			UpdatedByUserID:      &userID,
+		})
 	}
-	m, err := c.decryptMap(dto.Project.ProdVariables, dto.Project.ProdVariablesEncryptionKeyID)
+
+	rows, err := sqlx.NamedQueryContext(ctx, c.getDB(ctx), query, args)
```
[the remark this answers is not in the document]
k-anshul: This is at the bottom of the `ReadMe`. The arg can be a slice of structs or slice of maps. Need to scan all rows individually though.

admin/database/postgres/postgres.go
```diff
@@ -2060,25 +2049,108 @@ func (c *connection) DeleteBillingIssueByTypeForOrg(ctx context.Context, orgID s
...
+func (c *connection) UpsertProjectVariable(ctx context.Context, projectID, environment string, vars map[string]string, userID string) ([]*database.ProjectVariable, error) {
+	query := `INSERT INTO project_variables (project_id, environment, name, value, value_encryption_key_id, updated_by_user_id, updated_on)
+	VALUES (:project_id, :environment, :name, :value, :value_encryption_key_id, :updated_by_user_id, now())
+	ON CONFLICT (project_id, environment, lower(name)) DO UPDATE SET
+		value = EXCLUDED.value,
+		value_encryption_key_id = EXCLUDED.value_encryption_key_id,
+		updated_by_user_id = EXCLUDED.updated_by_user_id,
+		updated_on = now() RETURNING *`
+
+	args := make([]database.ProjectVariable, 0, len(vars))
```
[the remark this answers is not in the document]
k-anshul: The column specified in the `INSERT` statement should ensure that only the required fields are set. We can also create a map since the args can also be a slice of maps. Just wanted to re-use the struct for simple use case.

admin/database/postgres/tx.go
```diff
@@ -47,9 +47,16 @@ type dbHandle interface {
 	sqlx.QueryerContext
 	sqlx.PreparerContext
 	sqlx.ExecerContext
+	binder
 	SelectContext(ctx context.Context, dest interface{}, query string, args ...interface{}) error
 }
 
+type binder interface {
```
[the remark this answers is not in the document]
k-anshul: This is required so that we can call the `NamedQueryContext` from both tx and database. I can add some context here.

admin/projects.go
k-anshul: No this is only set when user is trying to remove the last set `variable`(This is just a side-effect on how the runtime API works). 
In other cases it will not be set and that's why it is not required in `ResolveProdVariables`. 
We can change it to `rill.internal.nonce` but it is better to only set this when it is required.

admin/projects.go
k-anshul: This switches on the environment set in the variable. I had thought it is `prod` and `dev`. I can change it.
Instead of doing case insensitive comparisons we can set environment in lower case only while inserting.

admin/server/projects.go
k-anshul: Our API resembles a patch API. I am not sure what is the best practice, whether to return all values or just what was updated. In the proposal it was specified to return the values that were created or updated by the request.

Also in order to return everything we will need to make a separate DB call which seems somewhat unnecessary.
Based on the feedback [here](https://github.com/rilldata/rill/pull/5929#discussion_r1810950851) our `UpdateProjectVariables` will become `GetProjectVariables` if user does not provide any variables to set :)

admin/database/postgres/migrations/0049.sql
```diff
@@ -0,0 +1,47 @@
...
+INSERT INTO project_variables (
+  project_id,
+  name,
+  value,
+  value_encryption_key_id
+)
+SELECT
+  project_data.project_id,
+  key AS name,
+  convert_to(value, 'UTF8')::bytea AS value,
```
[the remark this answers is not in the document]
k-anshul: As of now there are no encoded values set in the database so I avoided the additional decode complexity in the migration query.

admin/database/database.go
k-anshul: I updated the `UpsertProjectVariable` implementation to avoid usage of `NamedQuery`. Do you think we still need separate APIs?
The upsert query itself is fairly simple and similar upserts are in-use at other places as well. If we are not worried about performance we can consider running one query per variable which will make it even simpler.
The reason I want to favour `upsert` over `insert`/`update` is that we will need to add logic for what database already provides.

admin/database/postgres/migrations/0049.sql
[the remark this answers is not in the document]
begelundmuller: Okay, if there are no encoded values that sounds like a separate problem.

Anyway, it sounds to me like we should keep this field of type `TEXT` instead of `BYTEA` since it's storing raw strings when there's no encryption key. And then continue using base64 to serialize encrypted values as strings.

Using `BYTEA` to store raw strings means we additionally have to handle correct (de)serialization of text values (like using utf8 encoding as you do here), which does not seem worth it.

admin/projects.go
begelundmuller: Okay, I'm just thinking about future cases where variables may be resolved and saved from other code. For example a ClickHouse provisioner might remove variables when being disabled. 

Maybe have a `forWriting bool` arg on `ResolveVariables` that adds the nonce field if there aren't any variables? Just so the code for resolving variables is actually centralized in `ResolveVariables`.

admin/projects.go
begelundmuller: Sorry I meant case insensitive comparisons for `v.Name` (since it uses `lower` in the database schema for uniqueness)

admin/server/projects.go
begelundmuller: > our UpdateProjectVariables will become GetProjectVariables if user does not provide any variables to set

Which I think aligns with the other APIs, like `UpdateOrganization` where if you send an update request that doesn't make any changes, it returns the full value (not just the changed fields), i.e. acts similar to `GetOrganization`?

admin/database/database.go
begelundmuller: Okay it's fine to do upserts if you think it's simpler. It was mostly just a reflection on keeping things simple

admin/projects.go
k-anshul: I don't know how clickhouse provisioner will update variables but I think whenever variables are updated ideally it should call `UpdateProjectVariables`.
But yeah I am good with keeping it in `ResolveVariables` as well.

admin/database/postgres/migrations/0049.sql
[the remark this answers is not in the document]
k-anshul: Sure. I think this handling is also required in `InsertMagicAuthToken` but that will require migration and I think can be handled separately ?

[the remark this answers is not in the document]
begelundmuller: `InsertMagicAuthToken` is different because the `secret` is actually a `[]byte`, not a string. So it's storing bytes in both the encrypted and non-encrypted cases. Unlike the `secret` here, unencrypted variable values are strings.

admin/projects.go
```diff
@@ -207,6 +205,80 @@ func (s *Service) UpdateProject(ctx context.Context, proj *database.Project, opt
...
+	}
+	defer func() {
+		_ = tx.Rollback()
+	}()
+
+	// Upsert variables
+	if len(vars) > 0 {
+		_, err = s.DB.UpsertProjectVariable(txCtx, project.ID, environment, vars, userID)
+		if err != nil {
+			return status.Error(codes.Internal, err.Error())
```
begelundmuller, line 225: It should return the error directly. The `status` package is gRPC-specific so should only be used in the `admin/server` package or related server-specific packages.

Even in the `admin/server` package, the `err` should be returned directly because it may also be a context cancelled error, which are not considered internal errors. Our gRPC server has error handling middleware that checks un-annotated errors against common errors (such as context cancellation), falling back to an internal status code for unknown errors.

Note that there are multiple occurrences of this issue in this file.

begelundmuller (changes requested):

admin/projects.go
k-anshul: Sure. Modified the errors at both places.

begelundmuller (approved):

PR #5931, opened 2024-10-17
chore: monorepo updates

briangregoryholmes:
- Adds [syncpack](https://jamiemason.github.io/syncpack/guide/getting-started/) for version management across applications
- Runs `npx syncpack fix-mismatches` to fix version mismatches
- Removes explicit Vite dependency
- Updates `@sveltejs/kit` to latest version (features some necessary updates)
- Removes some TypeScript casts/workarounds that are no longer necessary
- Changes the `--allowed-origins` flag in the `dev` script to a wildcard
- Moves the `web-local` port assignment out of `vite.config` and into the root `package.json`
- Adds the "--" separator to the `dev-web` script so that a port can be passed

These last three changes enable us to run different branches/versions of the application against the same runtime very easily. In one branch, use `npm run dev` to start the application on port 3001. On another branch, run `npm run dev-web -- --port 4001` to run that version of the UI code on a different port.


ericpgreen2 (approved):
👍. Should we run `syncpack` in CI to ensure we don't get into a state of mismatched versions going forward?

briangregoryholmes:
I think that makes sense. Will follow up once I look into some configuration options.

PR #5932, opened 2024-10-17
Default quotas, sudo cmd to update internal orgs to superuser plan and support for extending trial

pjain1:
- Change the default quotas to be equivalent to trial plan as we take bigger of the value in the init org billing job 

Sudo cmd to update internal orgs to superuser plan `rill sudo org set-internal-plan`
- These will have all quotas as -1
- These orgs will still have no payment method and no billable address issue as actually corresponding Stripe customer does not have this info. This is fine as it just block plan upgrades which is not needed in this case anyways.

Sudo cmd to extend trial by num days provided as argument `rill sudo billing extend-trial --org --days`
- Supports extending plan on trial
- Supports extending plan with trial ended but still within grace period
- Supports extending plan with trial ended but grace period ended
- Supports extending plan when subscription was cancelled and valid till billing cycle end only when current sub is trial or subscription billing cycle ended.


begelundmuller (commented):
Since most orgs are created during deploys, which are now mostly triggered from the UI, I think it might be too big of an ask to have internal users create orgs manually using `rill sudo`. 

To stay compatible with the UI deploy flow that creates an org, would it be possible instead to have:
1. The quota on number of trial orgs be unlimited for superusers
2. Have a `rill sudo` command that moves an existing org (e.g. created with the normal flow) to the superuser plan?

pjain1:
@begelundmuller as per your comment removed `sudo org create` cmd and wrapped subscription update cmd to superuser plan in cmd `rill sudo org set-internal-plan` which retires until billing gets initialized. Now `rill org create NAME && rill sudo org set-internal-plan NAME` will do the same job as rill sudo org create.

Also added support extending trial using `rill sudo billing extend-trial --days` cmd here - https://github.com/rilldata/rill/pull/5932/commits/33cdf41271cee963be1e05985d55cc7686da8f25 which fixes https://github.com/rilldata/rill-private-issues/issues/759

begelundmuller (approved):

PR #5940, opened 2024-10-18
Unified auth experience

lovincyrus:
Closes https://github.com/rilldata/rill-private-issues/issues/743


ericpgreen2:
On staging, when I enter an email address and click "Continue with email", nothing happens.

lovincyrus:
> On staging, when I enter an email address and click "Continue with email", nothing happens.

Try again with Enter to submit password. I'll add the click event to the Continue.

web-auth/src/components/DiscordCTA.svelte
ericpgreen2, line 1: Maybe re-use the `CTANeedHelp.svelte` component in `web-common`?

web-auth/src/components/utils.ts
ericpgreen2, line 1: Why remove the types?

web-auth/src/components/Auth.svelte
```diff
@@ -6,69 +6,69 @@
...
+  const connectionMapObj = JSON.parse(connectionMap);
   const disableForgotPassDomainsArr = disableForgotPassDomains.split(",");
 
-  // By default show the LogIn page
-  $: isLoginPage = true;
   $: errorText = "";
-  $: isRillCloud = false;
 
   let isSSODisabled = false;
   let isEmailDisabled = false;
```
ericpgreen2, line 30: I don't see this `isEmailDisabled` variable ever get modified

ericpgreen2 (commented):
Can we only designate a sign-in option as "last used" once the user has successfully used it to log-in? [See this Jam](https://jam.dev/c/603a829d-1672-4211-8e0b-0f386e784ffe) for two problems: 
1. The "last used" parenthetical switches before the next page is loaded 
2. I do not complete the Microsoft SSO flow, so I'd consider it "never used" not "last used"

web-auth/src/components/Auth.svelte
lovincyrus: Good call, this was a remnant from the previous centralized implementation in Auth.svelte. Will remove it.

web-auth/src/components/utils.ts
lovincyrus: Reverting, these were causing issues in the build before brian's work

web-auth/src/components/Auth.svelte
lovincyrus: also removed unused `isSSODisabled`

web-auth/src/components/DiscordCTA.svelte
lovincyrus: Will do, didn't know this existed!

lovincyrus:
> Can we only designate a sign-in option as "last used" once the user has successfully used it to log-in? [See this Jam](https://jam.dev/c/603a829d-1672-4211-8e0b-0f386e784ffe) for two problems:
> 
> 1. The "last used" parenthetical switches before the next page is loaded
> 2. I do not complete the Microsoft SSO flow, so I'd consider it "never used" not "last used"

Will only set last used in a success callback

lovincyrus:
I can remove the last usage implementation in this ref and revisit this

ericpgreen2 (approved):
LGTM. @djbarnwal, since you're closer to this code, can you please also give it a review?

djbarnwal (commented):
https://jam.dev/c/647b8271-4469-4694-9f53-c0ec127c6b3d
Submitting email doesn't work for me. Tried firefox and chrome.

ericokuma:
UXQA items:

1. First time logging in or signing up always started with Log in with Email page
Screenshot 2024-11-07 at 5 01 52 PM

We don't want that. We want to show this page instead:
Screenshot 2024-11-08 at 6 22 38 AM

2. auth error of type "invalid_request": the connection was not found

Steps to repro:
- Enter eric.okuma@rilldata.com into email field
- Hit Continue
- See that it says email is SSO-supported
- Hit Continue, get error 

@lovincyrus

lovincyrus:
> 1. First time logging in or signing up always started with Log in with Email page

This is a bug caused from the latest commit. Will fix

lovincyrus:
> 2\. auth error of type "invalid_request": the connection was not found

I believe this is expected in staging. I was able to repro this error in the previous staging build.

https://github.com/user-attachments/assets/02b7c72b-6697-42aa-8679-6b7938e29032

lovincyrus:
Are legacy users allowed to use google or microsoft auth? @djbarnwal

PR #5941, opened 2024-10-19
refactor: welcome redirect flow

briangregoryholmes:
- Fixes broken links in the example project redirect flow (links are no longer hardcoded, new projects redirect to the first dashboard file)
- Removes dedicated `/welcome` route
- Conditionally renders out example project cards in the same place as the `Onboarding Workspace`
- Removes `EmptyProject` component
- Hides `ApplicationHeader` and `Navigation` when project is uninitialized
- Rewrites `Card` component to act as a link with disabled and loading states
- Clicking an example project disables the other cards and displays a loading spinner
- Fixes issue with svelte-check process


.github/workflows/web-test-code-quality.yml
```diff
@@ -52,6 +52,7 @@ jobs:
       - name: lint and type checks for web local
         if: steps.filter.outputs.local == 'true'
         run: |-
+          npm run build -w web-local
```
ericpgreen2, line 55: Double-checking – is this intentional? To catch build-time errors?

ericpgreen2 (changes requested):
I'm not seeing the background images for each of the example project cards: 
image

Somewhat out-of-scope for this PR, but I see that the "Preview dashboard" button should be disabled while the resource is reconciling, else I can click it and hit this 404 page:
image

briangregoryholmes:
Okay, should be fixed now. I've removed the `$img` alias and simplified and bypassed Tailwind for rendering out the images. I'm not seeing any issues in dev or production with these assets, so perhaps the newer versions of SvelteKit/Vite fixed whatever issue necessitated creating that alias.

ericpgreen2 (approved):

PR #5942, opened 2024-10-19
refactor Leaderboard/DimensionTable queries

### LINKED ISSUE #5753: Leaderboards: Missing data after selecting 9th+ dimension member in Dimension Table
To reproduce:
1. Find a Dimension Table that has 9 or more dimension members
2. Select/filter for the 9th (or later) dimension member
3. Go back to the Leaderboard
4. See the dimension member is presented without its measure value

For example, here's a Leaderboard:
image

Here's its Dimension Table:
image

After selecting for "Chess Royale", the leaderboard lacks its measure value (2.5M):
image

briangregoryholmes:
This PR refactors the way queries are handled within the `Leaderboard` and `DimensionTable` components. Namely, the reliance on externally derived readables that are accessible only within the StateManger context has been removed. Instead, they accept state-agnostic props and it is the responsibility of the parent component to pass them down accordingly. The logic for building out the queries/options objects is now collocated with the relevant component. In the future, `Leaderboard` and `DimensionTable` will be consolidated into a single component (along with `LeaderboardDisplay` and `DimensionTableDisplay`).

As part of this work, the `Leaderboard` now properly displays data for items "below the fold", as seen below.

Screenshot 2024-10-18 at 9 16 10 PM

Closes #5753


web-common/src/features/dashboards/leaderboard/Leaderboard.svelte
```diff
@@ -70,19 +74,58 @@
...
     $leaderboardSortedQueryBody(dimensionName),
     $leaderboardSortedQueryOptions(dimensionName, visible),
   );
 
+  $: belowTheFoldDataQuery = createQueryServiceMetricsViewAggregation(
+    instanceId,
+    $metricsViewName,
+    {
+      dimensions: [{ name: dimensionName }],
+      whereSql: selectedBelowTheFold
```
AdityaHegde, line 97: Would be better to use Expression while also merging with the general filters. Adding syntax mapping for things like measure filters would be redundant since backend does it from Expression object already.

Note that right now filters from other dimensions do not apply to this. You need to call `mergeMeasureFilters` and add this filter to the final filters.

web-common/src/features/dashboards/leaderboard/Leaderboard.svelte
```diff
@@ -70,19 +74,58 @@
...
+        end: timeControls.timeEnd,
+      },
+      comparisonTimeRange: timeControls.showTimeComparison
+        ? {
+            start: timeControls.comparisonTimeStart,
+            end: timeControls.comparisonTimeEnd,
+          }
+        : undefined,
+
+      measures: [
```
AdityaHegde, line 111: Can you replicate measure selection based on `leaderboardSortedQueryBody`.

AdityaHegde (changes requested):

ericpgreen2:
Is this one ready for review again? IMO this is an important one to get in the next release

AdityaHegde (commented):
Seeing some weird behaviour still,
1. Selecting below fold values, refreshing, expanding the dimension and coming back removes the values.
https://www.loom.com/share/41ba9ec4e6b84956859aabcc4d4b5ca1?sid=267e6432-8c69-49eb-bc76-2b94aa3a343d
2. I feel like there is some reactivity missing somewhere. Values do not update when I select filters for other dimensions. Also when values go above the fold they are duplicated.
https://www.loom.com/share/d1d8311b050145089c3a9592598b339d?sid=089fc062-7494-4f5d-b421-c1bd15c983e8

web-common/src/features/dashboards/leaderboard/Leaderboard.svelte
```diff
@@ -68,21 +86,126 @@
...
+            : []),
+        ),
+    },
+    $leaderboardSortedQueryOptions(
+      dimensionName,
+      visible && selectedBelowTheFold.length > 0,
+    ),
+  );
+
+  function getFiltersForOtherDimensions(
```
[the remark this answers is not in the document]
AdityaHegde: Thanks for updating this now.

briangregoryholmes:
I've decided to just make this PR the full refactor. These changes have resolved your two notes, but I recommend taking another look now that the scope has changed.

web-common/src/features/dashboards/state-managers/selectors/dashboard-queries.ts
```diff
@@ -1,181 +1,17 @@
-import { mergeMeasureFilters } from "@rilldata/web-common/features/dashboards/filters/measure-filters/measure-filter-utils";
```
AdityaHegde, line 1: Nice seeing this massive removal! This was needed when we had different APIs used. Now with a single AggregationRequest I was thinking of writing up a wrapper around it using builder pattern. Might decrease the code a lot more.

AdityaHegde (approved):
Looks good! One thing that is different from production is that below the fold values that get filtered out are not shown anymore. Not sure what the UX would be for those.

briangregoryholmes:
Added a catch for that situation. Those values are now displayed below the fold with no data.

web-common/src/features/dashboards/leaderboard/Leaderboard.svelte
```diff
@@ -30,95 +85,164 @@
...
+
+  $: leaderboardTotal = totalsData?.data?.[0]?.[activeMeasureName] as
+    | number
+    | null;
 
-  let aboveTheFold: LeaderboardItemData[] = [];
-  let selectedBelowTheFold: LeaderboardItemData[] = [];
-  let showExpandTable = false;
-  let noAvailableValues = true;
+  $: console.log(sortedData?.data);
```
AdityaHegde, line 177: Leftover console

AdityaHegde (approved):

PR #5944, opened 2024-10-19
monorepo package structure

briangregoryholmes:


lovincyrus:
> I think there are number of benefits to this structure (and Turbo), though I would be lying if I said I wasn't mostly motivated by shortening our exceedingly long import paths. 🤷

Oooh. Thanks for taking the first stab at this! We can start with something like this.

```
/rill
├── admin
├── docs
├── proto
├── runtime
├── scripts
├── packages
│   ├── ui
│   ├── states
│   ├── constants
│   ├── utils
│   │   └── number-formatting
│   └── icons
├── apps
│   ├── cli
│   ├── web
│   │   ├── admin
│   │   ├── auth
│   │   ├── local
│   │   └── common
```

References
https://github.com/vercel/turborepo/tree/main/examples/with-svelte
https://github.com/orval-labs/orval/tree/master/samples/svelte-query/basic
https://github.com/shadcn-ui/ui
https://github.com/calcom/cal.com

ericpgreen2:
I like the idea of using Turbo & I can see the benefit of creating smaller packages to improve its ability to cache.

One reservation I have is that if we nest `admin`, `auth`, `local` directories under `apps`, then I feel like the Application and Platform directory structures would be asymmetrical – since `admin` and `runtime` are at the top-level, not nested under a directory like `platform` or `services`. Or would you propose a re-shuffling of those directories too?

begelundmuller:
Just for inspiration, two other projects with a Go backend and SPA frontend are:
- https://github.com/grafana/grafana
- https://github.com/sourcegraph/sourcegraph-public-snapshot (they went private recently, this is a snapshot)

PR #5946, opened 2024-10-20
fix: type errors and prune whitelist

briangregoryholmes:
The slow process of eliminating this whitelist continues.


ericpgreen2 (approved):
🙌

PR #5949, opened 2024-10-22
child plan metadata, tax exemption for non US/Canada regions

pjain1:
Fixes https://github.com/rilldata/rill-private-issues/issues/804

Also add support to Inherit plan metadata from parent plan. Orb automatically creates a child plan in case we override any pricing info, only one level nesting is done.

### DIFF
--- a/admin/billing/orb.go
+++ b/admin/billing/orb.go
@@ -214,7 +214,7 @@ func (o *Orb) UnscheduleCancellation(ctx context.Context, subscriptionID string)
 	if err != nil {
 		return nil, err
 	}
-	return getBillingSubscriptionFromOrbSubscription(sub)
+	return o.getBillingSubscriptionFromOrbSubscription(ctx, sub)
 }
 
 func (o *Orb) CancelSubscriptionsForCustomer(ctx context.Context, customerID string, cancelOption SubscriptionCancellationOption) (time.Time, error) {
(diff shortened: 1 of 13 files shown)


admin/billing/orb.go
```diff
@@ -279,6 +279,35 @@ func (o *Orb) IsInvoicePaid(ctx context.Context, invoice *Invoice) bool {
 	return strings.EqualFold(invoice.Status, "paid")
 }
 
+func (o *Orb) MarkCustomerTaxExempt(ctx context.Context, customerID string) error {
+	// note - hard coded to use Avalara tax provider as of now, Orb supports Avalara and TaxJar
```
pjain1, line 283: do we need to make this configurable, if in future we change tax provider, not really urgent now ?

begelundmuller: I don't know, what do you think?

begelundmuller (approved):

admin/billing/orb.go
pjain1: Made it configurable in case we want to disable it (manually managing it) or change at some point in future. 

Configurable through `RILL_ADMIN_ORB_INTEGRATED_TAX_PROVIDER`, default is still `avalara` if not set for convenience as Kristen mentioned that this is going to be our tax provider going forward, can be set to `none` to disable.

PR #5967, opened 2024-10-23
billing devtool and docs

pjain1:
Fixes https://github.com/rilldata/rill-private-issues/issues/714

This does not include instructions on how to use utility cmd that @AdityaHegde created to manipulate end dates in billing issues metadata to simulate various scenarios. May be we can move the cmds into the project and add instructions in readme or point to that repo but that would be private repo.


cli/cmd/devtool/data/cloud-deps.docker-compose.yml
```diff
@@ -50,3 +50,8 @@ services:
       - "55679:55679" # zpages extension
     depends_on:
       - zipkin
+  stripe-cli:
+    image: stripe/stripe-cli:latest
+    command: listen --forward-to http://host.docker.internal:8080/payment/webhook
+    environment:
+      - STRIPE_API_KEY=
```
pjain1, line 57: need to discuss this

cli/cmd/devtool/data/template.env
pjain1, line 1: This file seems unused, not sure where it is used.

cli/cmd/billing/subscription/subscription.go
```diff
@@ -13,7 +13,6 @@ func SubscriptionCmd(ch *cmdutil.Helper) *cobra.Command {
 	}
 
 	subsCmd.PersistentFlags().StringVar(&ch.Org, "org", ch.Org, "Organization Name")
-	subsCmd.AddCommand(CreateCmd(ch))
```
pjain1, line 16: Edit cmd does the same thing

cli/cmd/devtool/start.go
```diff
@@ -710,6 +717,39 @@ func (s local) awaitUI(ctx context.Context) error {
...
+func prepareStripeConfig() error {
+	templateFile := "cli/cmd/devtool/data/stripe-config.template"
+	outputDir := "dev-cloud-state"
+	outputFile := filepath.Join(outputDir, "stripe-config.toml")
+
+	apiKey := lookupDotenv("RILL_DEVTOOL_STRIPE_CLI_API_KEY")
+
+	if apiKey == "" {
+		logWarn.Printf("No Stripe API key found in .env, Stripe webhook events will not be processed\n")
+	}
```
begelundmuller, line 729: Should there be a `return nil` here?

begelundmuller (approved):

cli/cmd/devtool/start.go
pjain1: What do you think, just didn't wanted to block devtool in case of missing api key

pjain1: oh misread the comment, wanted to create the file anyways so that docker compose volume thing does not fail

begelundmuller: Got it :)

PR #5972, opened 2024-10-24
tweak redirect flow

briangregoryholmes:
To address some concerns related to example project unpacking in certain situations, this PR makes the following changes:

- Reduce watch-request-client priority disconnect (which applies when the tab is put in the background) to 20 seconds
- Prevent multiple clicks on the same example project
- If `?redirect=true` is still in the URL after 5 seconds, reload the page


ericpgreen2 (approved):
These tweaks look reasonable. I'll approve and merge, and we can wait-and-see to see if anyone still reports problems.

Issue #5592, opened 2024-09-03
Release: Update version number in `package.json` files


ericpgreen2:
We should update our `rill-ui.yml` GitHub Action so that it writes the release number to our various `package.json` files:
- Root `package.json`
- `web-admin/package.json`
- `web-common/package.json`
- `web-local/package.json`

himadrisingh:
GHA will be confusing, we can try using bumpversion to update these files and create a tag for the release

Kavinjsir:
Hey @ericpgreen2, just wanted to let you know that after checking things out on the infra side, we've decided to stick with manual updates for now.

For releases, automating the version alignment across files between branches is a bit tricky at this stage.

So I'm gonna handle the updates manually from the main branch and push changes to the release branch, just like we do in our usual dev flow.

We’ll pick up some insights along the way and keep refining our approach.

If you have any suggestions or there’s anything specific you’d like to see, definitely give us a shout 😃

PR #5973, opened 2024-10-24
ops: align versions in monorepo

Kavinjsir:
https://github.com/rilldata/rill/issues/5592

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
   "name": "@rilldata/dev",
-  "version": "0.10.0",
+  "version": "0.0.0",
+  "private": true,
   "type": "module",
   "workspaces": [
     "docs",
--- a/web-auth/package.json
+++ b/web-auth/package.json
@@ -1,6 +1,6 @@
 {
 	"name": "web-auth",
-	"version": "0.0.1",
+	"version": "0.0.0",
 	"private": true,
 	"scripts": {
 		"dev": "vite dev",
--- a/docs/package.json
+++ b/docs/package.json
@@ -1,6 +1,7 @@
 {
   "name": "docs",
-  "version": "0.0.1",
+  "private": true,
+  "version": "0.0.0",
   "scripts": {
     "dev": "docusaurus start --port 4002",
     "build": "docusaurus build --out-dir dist",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@rilldata/dev",
-  "version": "0.10.0",
+  "version": "0.0.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@rilldata/dev",
-      "version": "0.10.0",
+      "version": "0.0.0",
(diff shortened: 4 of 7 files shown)


himadrisingh:
lets make them `0.51.0` as the next release would be cut from this branch.

ericpgreen2 (commented):
Given 1) we're not publishing any of our packages to npm, 2) we have our own way of managing code versions via git tags, and 3) we're not consuming any `version` anywhere.... what if we removed the `version` fields altogether?

Rather:
- We could add `"private": true` to our package.json files to be more explicit that they're not published to `npm`.
- We could cite `web-common` in our `devDependencies` like: `"web-common": "workspace:*"`, which seems to be a standard practice in monorepos.

Thoughts?

Kavinjsir:
> Given 1) we're not publishing any of our packages to npm, 2) we have our own way of managing code versions via git tags, and 3) we're not consuming any `version` anywhere.... what if we removed the `version` fields altogether?
> 
> Rather:
> 
>     * We could add `"private": true` to our package.json files to be more explicit that they're not published to `npm`.
> 
>     * We could cite `web-common` in our `devDependencies` like: `"web-common": "workspace:*"`, which seems to be a standard practice in monorepos.
> 
> 
> Thoughts?

@ericpgreen2 That sounds great to me! It helps reduce the complexity of version maintenance across different workspaces and during the release process. I also tried update `version: "0.0.0"` for each workspace in the project, in an attempt to explicitly "lock" the version value in package.json.
One issue here is, when specifying `"web-common": "workspace:*"`, it would crash by the following error:
workspace-issue

Not quite an expert over here, so I tried replace `"workspace:*"` with `"*"` to resolve the issue.
Here is the reference I got: https://stackoverflow.com/a/79135939

How would you like this?

cc: @himadrisingh I'm following @ericpgreen2 's suggestion which looks a good approach to reduce complexity. I also updated the version in the entry package.json to `0.51.0` regarding your suggestion :)

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@rilldata/dev",
-  "version": "0.10.0",
+  "version": "0.51.0",
```
ericpgreen2, line 3: My comments on not relying on any `version` also apply to this root package.json. What if we set this one to `"0.0.0"` too?

web-auth/package.json
```diff
@@ -1,6 +1,7 @@
 {
 	"name": "web-auth",
-	"version": "0.0.1",
+  "private": true,
```
ericpgreen2, line 3: Redundant `private: true` (see line 5)

package.json
Kavinjsir: Sounds good!

web-auth/package.json
Kavinjsir: Good Catch!

ericpgreen2 (approved):
LGTM! This seems like a good set-it-and-forget it approach for now. If we ever change how we version our releases, or if we ever publish some packages to npm, then we can revisit.

PR #5974, opened 2024-10-24
Support `format_d3_locale` in the metrics YAML

begelundmuller:
Enables passing locale config to D3. For details about the available properties, see: https://d3js.org/d3-format#formatLocale.

Example measure:

```yaml
measures:
  - name: b
    expression: count(*)
    format_d3: "$.2f"
    format_d3_locale:
        currency: ["£", ""]
```

The D3 locale properties are returned in the `MetricsViewSpec.MeasureV2` type for frontend consumption (cc @djbarnwal).


web-common/src/runtime-client/gen/index.schemas.ts
```diff
@@ -2505,6 +2505,8 @@ export interface NumericHistogramBinsBin {
   count?: number;
 }
 
+export type MetricsViewSpecMeasureV2FormatD3Locale = { [key: string]: any };
```
djbarnwal, line 2508: Can we type it more strongly?

djbarnwal: Probably something along the lines of 

```
interface FormatLocaleDefinition {
  decimal?: string; // Character for the decimal point (e.g., '.')
  thousands?: string; // Character for thousands separator (e.g., ',')
  grouping?: number[]; // Array defining group sizes (e.g., [3] for groups of 3 digits)
  currency?: [string, string]; // Array for prefix and suffix currency symbols (e.g., ['$', ''])
  numerals?: string[]; // Array of numeral representations, if different from 0-9
  percent?: string; // Symbol used for percentage (e.g., '%')
  minus?: string; // Symbol for minus (negative) values (e.g., '-')
  nan?: string; // Representation of "Not a Number" (e.g., 'NaN')
}
```

djbarnwal (commented):
Can we have more stricter type definitions?

web-common/src/runtime-client/gen/index.schemas.ts
begelundmuller: I can copy D3's spec and validation on the runtime side if you really think it's worth it, but I would prefer to just leave the validation and reporting of any errors in the locale to D3. The error will show up in the UI instead of the reconcile logs since it happens at runtime, but since this is an advanced feature and unlikely to act differently in production, I think that should be okay. Apart from the duplicated logic, it also saves us from needing to stay in sync with any additional options D3 adds.

djbarnwal (approved):

PR #5981, opened 2024-10-25
Refactor `.vars`/`--var` to `.env`/`--env`

begelundmuller:
Changes:

- Support `{{ .env.name}}` instead of `{{ .vars.name }}` to reference a variable in templating.
- Support `--env name=value` instead of `--var name=value` for specifying a variable in `rill start`.
- Support specifying environment variables in `rill.yaml` using `env:` (changed from `vars:`).
- For overriding the environment in `rill start`, now use `--environment prod` (changed from `--env prod`).
- For environment-specific overrides, use `dev:` and `prod:` directly (instead of nesting under `env:dev:` since `env:` is now used for setting environment variables in YAML).

The changes above are almost fully backwards compatible (there are some uncovered cases for custom environments, but that capability has never been used to my knowledge).

For the reviewer, some of the refactors in this PR are slightly peculiar due to the fact that we originally actually used `env:` instead of `vars:` to specify environment variables, so in some places, this PR actually just inverts old backwards compatibility logic.

Closes https://github.com/rilldata/rill/issues/5896


cli/cmd/start/start.go
```diff
@@ -182,23 +173,27 @@ func StartCmd(ch *cmdutil.Helper) *cobra.Command {
...
 	startCmd.Flags().StringVar(&logFormat, "log-format", "console", "Log format (options: \"console\", \"json\")")
 	startCmd.Flags().StringVar(&tlsCertPath, "tls-cert", "", "Path to TLS certificate")
 	startCmd.Flags().StringVar(&tlsKeyPath, "tls-key", "", "Path to TLS key file")
-
-	// --env was previously used for variables, but is now used to set the environment name. We maintain backwards compatibility by keeping --env as a slice var, and setting any value containing an equals sign as a variable.
-	startCmd.Flags().StringSliceVarP(&env, "env", "e", []string{}, `Environment name (default "dev")`)
-	startCmd.Flags().StringSliceVarP(&vars, "var", "v", []string{}, "Set project variables")
 	startCmd.Flags().StringSliceVarP(&allowedOrigins, "allowed-origins", "", []string{}, "Override allowed origins for CORS")
 
+	// Deprecated support for "--env": replaced by "--env".
```
k-anshul, line 191: // Deprecated support for "--var": replaced by "--env" ?

begelundmuller: Thanks!

runtime/reconcilers/model.go
```diff
@@ -1399,7 +1399,8 @@ func (r *ModelReconciler) analyzeTemplatedVariables(ctx context.Context, props m
 	vars := inst.ResolveVariables(false)
 
 	for k := range res {
-		k2 := strings.TrimPrefix(k, "vars.")
+		// Project variables are referenced with .env.name (current) or .vars.name (deprecated).
+		k2 := strings.TrimPrefix(strings.TrimPrefix(k, "vars."), "env.")
```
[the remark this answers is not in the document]
begelundmuller: Yeah I was lazy, but you're right, it's best to do this properly. Fixed it!

PR #5983, opened 2024-10-25
check org billing init in start trial job

pjain1:
retry start trial until org billing init is done

### DIFF
--- a/admin/jobs/river/river.go
+++ b/admin/jobs/river/river.go
@@ -342,6 +342,7 @@ func (c *Client) StartOrgTrial(ctx context.Context, orgID string) (*jobs.InsertR
 		UniqueOpts: river.UniqueOpts{
 			ByArgs: true,
 		},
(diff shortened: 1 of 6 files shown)


admin/jobs/river/river.go
```diff
@@ -342,6 +342,7 @@ func (c *Client) StartOrgTrial(ctx context.Context, orgID string) (*jobs.InsertR
 		UniqueOpts: river.UniqueOpts{
 			ByArgs: true,
 		},
+		MaxAttempts: 5, // override default retries as init org billing job should complete before this if org creation and project deployment were done in single flow
```
AdityaHegde, line 345: What is the default? Is it the 3 we set in the client? If so will going from 3  => 5 give enough time?

pjain1: yes default is 3, going to 5 should give around extra 2 mins which should be sufficient IMO. Its just that init job should complete before this. We should add alerts on failed jobs.

begelundmuller (approved):

PR #5984, opened 2024-10-28
review

pjain1:


cli/cmd/devtool/subscription.go
```diff
@@ -105,9 +105,10 @@ func AdvanceSubscriptionTimeCmd(ch *cmdutil.Helper) *cobra.Command {
 						OrgID: orgResp.Organization.Id,
 						Type:  database.BillingIssueTypeOnTrial,
 						Metadata: &database.BillingIssueMetadataOnTrial{
-							SubID:   subResp.Subscription.Id,
-							PlanID:  subResp.Subscription.Plan.Id,
-							EndDate: endDate,
+							SubID:              subResp.Subscription.Id,
+							PlanID:             subResp.Subscription.Plan.Id,
+							EndDate:            endDate,
+							GracePeriodEndDate: issue.Metadata.GetOnTrial().GracePeriodEndDate.AsTime(),
```
pjain1, line 111: after extend trial cmd PR, I have added bth end and grace end to on-trial issue as both needs to be change in case of trial extension and both should be same as there is grace period on extending trial

AdityaHegde (approved):

Issue #4852, opened 2024-05-07
File Exporer: Add menu option to directly duplicate a model in the file view


smartexpert:
**Is your feature request related to a problem? Please describe.**

At times, you need to create a lot of duplicate models of an existing model to start of as the base layer. In release 0.44.1, this is not very user friendly as you have to navigate to the tables in the left bar, right-click the model you want to duplicate and click 'create new model'.

In addition, adding a new model directly in the file system doesn't work. The file shows up but it's not recognized as a rill model.

**Describe the solution you'd like**
In the new left side bar interface in release 0.44.1, add option to duplicate model in the right-click context menu.

![image](https://github.com/rilldata/rill/assets/1627433/8ba1e058-e372-43a1-805a-05c67cec75f8)

**Describe alternatives you've considered**

**Additional context**
None

jaroet:
And not only a model but also a source and maybe even a dashboard?

ericpgreen2:
Closed by https://github.com/rilldata/rill/pull/5993

PR #5993, opened 2024-10-28
File Explorer: Add "Duplicate" menu item

ericpgreen2:
Closes https://github.com/rilldata/rill-private-issues/issues/832, https://github.com/rilldata/rill/issues/4852

image

### DIFF
--- a/web-common/src/features/entity-management/actions.ts
+++ b/web-common/src/features/entity-management/actions.ts
@@ -19,30 +19,6 @@ import {
   removeLeadingSlash,
 } from "./entity-mappers";
 
-export async function copyFileArtifact(
-  instanceId: string,
-  filePath: string,
-): Promise<string> {
-  // Get new file path
-  const [folder, fileName, extension] =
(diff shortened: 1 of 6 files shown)


AdityaHegde (commented):
1. `Copy` doesnt feel like the correct label since it is creating a copy and not just copying to clipboard. Perhaps we can call it `Create Copy`
2. For things like explore should we consider adding `(Copy)` to `display_name`.

AdityaHegde (approved):

ericpgreen2:
> 1. `Copy` doesnt feel like the correct label since it is creating a copy and not just copying to clipboard. Perhaps we can call it `Create Copy`

Totally right. I switched to "Duplicate".

> 2. For things like explore should we consider adding `(Copy)` to `display_name`.

Because this is an option that applies to all files, I'd like the behavior to be consistent across all file types. So a simple copy-and-paste would be consistent & conventional, IMO.

PR #6001, opened 2024-10-29
feat: Parser for rill time syntax

AdityaHegde:
Backend checklist
- [x] Add module for parsing rill time syntax in backend.
- [x] Add `ResolveTime` to return start and end times based on minTime, maxTime and now for a metrics view.
- [x] Replace usage of old duration parser.
- [ ] Map old format to new format
- [x] Add API to resolve times for a list of rill-times.
- [ ] Add a lot more tests.

UI checklist
- [x] Add a parser in UI.
- [ ] Check if we really need to replicate the parser in UI. We could make the resolve time API to return parsed object as well.
- [x] Integrate the new API to get start and end in UI.
- [ ] Categorise the time ranges into `latest` `periodToDate` and `previous`
- [ ] Add conversion to human readable label.
- [ ] Add E2E tests around time.
- [ ] Docs for the new rill time syntax @medriscoll


AdityaHegde:
Closing this since frontend code is pretty stale with human readable url changes. Will open separate PRs for backend and frontend.

PR #6003, opened 2024-10-29
feat: Add slack notification support for reports

AdityaHegde:
- [x] Add slack user and channel inputs to create.
- [x] Fix slack notification. (https://github.com/rilldata/rill/pull/6005)
- [x] Add slack user and channel to metadata.
- [x] Add slack user and channel inputs to edit.
- [x] Get feedback on design, the form is too long. Perhaps we can use the new multi entry input used in user invite? (Switched to MultiInput along with using superforms)


web-common/src/features/scheduled-reports/utils.ts
```diff
@@ -17,3 +67,93 @@ export function getDashboardNameFromReport(
...
+  return mappedEntries as {
+    [KEY in K]: string;
+  }[];
+}
+
+function mapAndAddEmptyEntryV2(entries: string[] | undefined) {
+  const finalEntries = entries ? [...entries] : [];
+  finalEntries.push("");
+  return finalEntries;
+}
```
ericpgreen2, line 159: Do we need to add an empty entry? It'd be simpler if we didn't.

web-common/src/features/scheduled-reports/utils.ts
```diff
@@ -17,3 +67,93 @@ export function getDashboardNameFromReport(
     null
   );
 }
+
+export function extractNotification(
```
ericpgreen2, line 71: "extract" seems like the wrong verb here

ericpgreen2, line 71: I only see this function called once, it doesn't use the `userEmail` parameter, and it only uses `isEdit=true`, so we could remove those two parameters? Or we could consolidate and use `extractNotificationV2`.

web-common/src/features/scheduled-reports/utils.ts
```diff
@@ -1,5 +1,55 @@
...
+    timeZone: reportSpec?.refreshSchedule?.timeZone ?? getLocalIANA(),
+    exportFormat: reportSpec
+      ? (reportSpec?.exportFormat ?? V1ExportFormat.EXPORT_FORMAT_UNSPECIFIED)
+      : V1ExportFormat.EXPORT_FORMAT_CSV,
+    exportLimit: reportSpec
+      ? reportSpec.exportLimit === "0"
+        ? ""
+        : reportSpec.exportLimit
+      : "",
+    ...extractNotificationV2(reportSpec?.notifiers, userEmail, !!reportSpec),
```
ericpgreen2, line 48: `getNotificationValues` could be a better name

web-common/src/features/scheduled-reports/ScheduledReportDialog.svelte
```diff
@@ -52,124 +47,101 @@
...
-            getRuntimeServiceGetResourceQueryKey($runtime.instanceId, {
-              "name.name": reportName,
-              "name.kind": ResourceKind.Report,
-            }),
-          );
-        }
+        },
+      });
 
+      if (reportSpec) {
```
ericpgreen2, line 102: Nit: it looks like an explanatory variable `const isEditForm = !!reportSpec` could help both here and in line 118

ericpgreen2 (changes requested):
Design (cc @jkhwu) :
1. For the recipients input, can we make the label "Email recipients"? Else "recipients" and "users" look too similar.
    image
2. Can we have separate tags for "Email sent" and "Slack message sent"? In this case, there was no email sent, only a Slack message. In other cases, it's possible for both an email and a Slack message to be sent.
    image

web-common/src/features/scheduled-reports/utils.ts
AdityaHegde: Moving this back to alerts. It is till using the old forms lib. We can merge them once we move it to superforms.

web-common/src/features/scheduled-reports/utils.ts
AdityaHegde: The empty entry is what is edited. It is mainly to support submit without needing to convert the currently being edited entry to a pill explicitly.

web-common/src/features/scheduled-reports/utils.ts
AdityaHegde: We extract the slack and email related fields from then notifiers array.

web-common/src/features/scheduled-reports/utils.ts
```diff
@@ -17,3 +67,93 @@ export function getDashboardNameFromReport(
     null
   );
 }
+
+export function extractNotification(
+  notifiers: V1Notifier[] | undefined,
```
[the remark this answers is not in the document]
AdityaHegde: Moved back to alerts with no changes.

web-common/src/features/scheduled-reports/utils.ts
```diff
@@ -17,3 +67,93 @@ export function getDashboardNameFromReport(
...
+    slackChannels,
+    slackUsers,
+
+    enableEmailNotification: isEdit ? !!emailNotifier : true,
+    emailRecipients,
+  };
+}
+
+export function extractNotificationV2(
+  notifiers: V1Notifier[] | undefined,
```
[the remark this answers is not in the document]
AdityaHegde: This is a single function that handles both new and edit use case.

AdityaHegde:
1. Done.
2. A generic `Report sent` is better IMO. If one of them fails to send we should get an error and show that instead.

web-common/src/features/scheduled-reports/ScheduledReportDialog.svelte
```diff
@@ -52,124 +47,102 @@
...
-              "name.name": reportName,
-              "name.kind": ResourceKind.Report,
-            }),
-          );
-        }
+        },
+      });
 
+      const isEdit = !!reportSpec;
+      if (isEdit) {
```
ericpgreen2, line 103: For clarity, this `isEdit` variable could be used in more places throughout this file – wherever there's a ternary based on `reportSpec`

ericpgreen2 (approved):

PR #6004, opened 2024-10-29
Fix ClickHouse timestamp format for comparison offsets

begelundmuller:
Addresses the problem reported [here](https://rilldata.slack.com/archives/C02T907FEUB/p1730209589645439?thread_ts=1730209125.774299&cid=C02T907FEUB)


nishantmonu51 (approved):

nishantmonu51:
@begelundmuller : Is it possible to add a unit test as well for the failure case to catch regressions in future?

begelundmuller:
> @begelundmuller : Is it possible to add a unit test as well for the failure case to catch regressions in future?

Yes, will do that in a follow up PR (slightly bigger change since I need to shuffle some ClickHouse init logic around to be able to test this).

PR #6006, opened 2024-10-30
DuckDB with backup on GCS

k-anshul:
closes https://github.com/rilldata/rill-private-issues/issues/692

Some of the previous logic removed from existing driver logic.
1. Clearing `.tmp` and `.wal` files if they are bigger than a certain size. Even on main it no longer works because the `tmp` and `wal` actually gets built in the attached DBs and not `main`.
2. String to `enum` conversion feature. Not being used anywhere and not documented as well. Can be added in the `rduckdb` if use case comes again.
3. `tx` support in duckdb driver. We no longer need to block read queries while attaching detaching.


begelundmuller (changes requested):

runtime/pkg/rduckdb/db.go
```diff
@@ -274,6 +274,15 @@ func NewDB(ctx context.Context, opts *DBOptions) (DB, error) {
...
 	}
+
+	// We want to prevent multiple rill process accessing same db files.
+	// All the files are accessed in read-only mode so it is possible for multiple rill process to access same db files.
+	// To prevent this we attach a dummy db file to the main in-memory db in write mode.
+	// This is required for local rill only but since there is no way to determine it in this package so we do it for all.
+	_, err = db.dbHandle.ExecContext(ctx, fmt.Sprintf("ATTACH %s AS __ymmud__", safeSQLString(filepath.Join(db.localPath, "main.db"))))
+	if err != nil {
+		return nil, err
+	}
```
begelundmuller, line 285: Instead of attaching, could it just pass `main.db` as the DSN to `openDBAndAttach`?

begelundmuller (changes requested):

runtime/drivers/duckdb/config.go
```diff
@@ -17,141 +14,67 @@ const (
...
 type config struct {
-	// DSN is the connection string. Also allows a special `:memory:` path to initialize an in-memory database.
-	DSN string `mapstructure:"dsn"`
-	// Path is a path to the database file. If set, it will take precedence over the path contained in DSN.
-	// This is a convenience option for setting the path in a more human-readable way.
-	Path string `mapstructure:"path"`
-	// DataDir is the path to directory where duckdb file named `main.db` will be created. In case of external table storage all the files will also be present in DataDir's subdirectories.
-	// If path is set then DataDir is ignored.
+	// DataDir is the path to directory where duckdb files will be created.
 	DataDir string `mapstructure:"data_dir"`
```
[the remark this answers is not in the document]
k-anshul: It still access data dir using storage. I kept it in case we want to configure this separately using envs.

runtime/drivers/duckdb/duckdb.go
```diff
@@ -203,43 +173,28 @@ func (d Driver) Open(instanceID string, cfgMap map[string]any, st *storage.Clien
...
-		walPath := cfg.DBFilePath + ".wal"
-		_ = os.Remove(walPath)
-		_ = os.Remove(cfg.DBFilePath)
-
-		// reopen connection again
-		if err := c.reopenDB(); err != nil {
-			return nil, err
+		// Check for another process currently accessing the DB
+		if strings.Contains(err.Error(), "Could not set lock on file") {
+			panic(fmt.Errorf("failed to open database (is Rill already running?): %w", err))
```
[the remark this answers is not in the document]
k-anshul: I didn't return error because ideally we should exit the program.
Changed to error now.

runtime/drivers/duckdb/duckdb.go
```diff
@@ -481,10 +485,23 @@ func (c *connection) reopenDB(ctx context.Context) error {
...
 	if !c.config.AllowHostAccess {
-		bootQueries = append(bootQueries, "SET preserve_insertion_order TO false")
+		tempDir, err := c.storage.TempDir()
+		if err != nil {
+			return err
+		}
+		bootQueries = append(bootQueries,
+			"SET preserve_insertion_order TO false",
+			fmt.Sprintf("SET temp_directory = %s", safeSQLString(tempDir)),
+			fmt.Sprintf("SET secret_directory = %s", safeSQLString(filepath.Join(dataDir, ".duckdb", "secrets"))),
```
begelundmuller, line 503: Will this interfere with external table storage or will it just be ignored?

begelundmuller (approved):

runtime/drivers/duckdb/duckdb.go
k-anshul: Good point. These queries are executed for every db file so setting `secret_directory` is what is expected. 
But there could be issues if all database share same tmp directory. May be for new removing `temp_directory` seems to be a safe option. 
More important was to set unique `secret_directory` for each instance's database.

PR #6018, opened 2024-10-31
Refactor resolver tests

begelundmuller:
- Improves the syntax and documentation for the new resolver YAML tests
- Makes initialization of test connectors more generic, paving the way for testing against more external connectors in the future


runtime/resolvers/resolvers_test.go
```diff
@@ -6,185 +6,248 @@ import (
...
-	Project    map[string]yaml.Node
-	Connectors map[string]*testruntime.InstanceOptionsForResolvers
-	Tests      map[string]*Test
+// TestFileYAML is the structure of a test file.
+// One runtime instance will created for each test file, with the provided file contents and connectors initialized.
+// Each test in the file will be run in sequence against that runtime instance.
+// The available connectors are defined in runtime/testruntime/connectors.go.
+type TestFileYAML struct {
+	Connectors []string             `yaml:"connectors,omitempty"`
+	Files      map[string]yaml.Node `yaml:"files"`
```
k-anshul, line 29: Can this be called `Resources` instead of `Files`? I got somewhat confused between test file vs this internal file while reading `ReadMe` and calling this resource would prevent confusion.

k-anshul (approved):

PR #6031, opened 2024-11-04
Tests for time dimension comparisons for ClickHouse and Druid

begelundmuller:
Adds tests for comparison metrics queries that have the time dimension as a dimension. Previously this was only tested for DuckDB, this PR adds resolver tests for it for all of DuckDB, Druid and ClickHouse.

This PR uses the new resolver test files support, which you can find more details about here: https://github.com/rilldata/rill/blob/main/runtime/resolvers/testdata/README.md

The tests here are not comprehensive for all comparison features, we'll be migrating and adding tests gradually. This PR addresses the test request here: https://github.com/rilldata/rill/pull/6004.

Currently the Druid test is failing, more details here: https://rilldata.slack.com/archives/CTCJ58H3M/p1730730758436969

### DIFF
--- a/runtime/resolvers/testdata/metrics_comparisons.yaml
+++ b/runtime/resolvers/testdata/metrics_comparisons.yaml
@@ -1,0 +1,149 @@
+connectors:
+  - clickhouse
+  - druid
+project_files:
+  clickhouse_data.yaml:
+    type: model
+    connector: clickhouse
+    sql: |
+      select parseDateTimeBestEffort('2024-01-01T00:00:00Z') as time, 'DK' as country, 1 as val union all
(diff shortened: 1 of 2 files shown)


runtime/resolvers/testdata/metrics_comparisons.yaml
```diff
@@ -0,0 +1,149 @@
+connectors:
+  - clickhouse
```
pjain1, line 2: where's the clickhouse cluster running ?

pjain1: ok I see `testruntime.NewInstanceForResolvers`

pjain1 (approved):
somewhat related: was wondering if there an easy way to exclude project files that test creates or clean them up after test is done. currently after running test, I see `runtime/testruntime/testdata/ad_bids/metrics/` as untracked files.

PR #6035, opened 2024-11-05
Billing UI follow ups

AdityaHegde:
- [ ] ~~Fix CSP issue with orb's portal.~~ In a follow up PR
- [x] Use organization quotas to show project and size limits. There could be overrides stored in db, so we cannot rely on plan's quotas.
- [x] Integrate the new usage metrics API.
- [x] Handle quota limits in deploy page.
- [x] Remove rogue domain in general settings page
- [x] Do not show billing banner on invite page.
- [x] Update the deploy modal in rill dev to mention free trial.
- [x] ~~Round durations instead of flooring. Also show UTC for free trial to be clear.~~ We are opting to use 31 days for trial so it appears 30 days instead.


web-admin/src/routes/+layout.ts
```diff
@@ -103,6 +108,7 @@ export const load = async ({ params, url, route }) => {
 
     return {
       organizationPermissions,
+      organizationQuotas,
```
ericpgreen2, line 111: It looks like `organizationQuotas` is only used in the `PlanQuotas.svelte` component. Rather than prop drill `organizationQuotas` through the app, I think it'd be cleaner to just pull the info from the TanStack Query cache in `PlanQuotas.svelte`.

web-admin/src/features/billing/plans/utils.ts
```diff
@@ -4,14 +4,19 @@ import { DateTime } from "luxon";
 import { writable } from "svelte/store";
 
 export function formatDataSizeQuota(
+  total: number,
```
ericpgreen2, line 7: Mind being more specific than "total"? It's hard for me to tell what the unit is

web-admin/src/features/billing/plans/selectors.ts
```diff
@@ -53,3 +59,59 @@ export function getNextBillingCycleDate(curEndDateRaw: string): string {
...
+        credsResp.data.instanceId,
+        credsResp.data.accessToken,
+      ).subscribe(set);
+    },
+  );
+}
+
+export type UsageMetricsResponse = {
+  project_name: string;
+  size: number;
```
ericpgreen2, line 85: Can we be more specific than "size"? What's the unit? Size of what exactly?

web-local/src/routes/(misc)/deploy/+page.svelte
```diff
@@ -31,6 +32,21 @@
...
+
+  let upgradeHref = "";
+  $: if ($org && $metadata.data) {
+    let cloudUrl = $metadata.data.adminUrl.replace(
+      "admin.rilldata",
+      "ui.rilldata",
+    );
+    if (cloudUrl === "http://localhost:9090") {
+      cloudUrl = "http://localhost:3000";
+    }
```
ericpgreen2, line 44: What's going on here? This seems like a hack. Can we solve the root problem?

web-common/src/features/project/ProjectDeployer.ts
```diff
@@ -160,9 +162,12 @@ export class ProjectDeployer {
     // eslint-disable-next-line no-constant-condition
     while (true) {
       try {
+        const tryOrgName = `${org}${i === 0 ? "" : "-" + i}`;
+        this.org.set(tryOrgName);
+        return;
```
ericpgreen2, line 167: Is this early return intentional? The code below is unreachable

web-admin/src/features/billing/plans/selectors.ts
```diff
@@ -53,3 +59,59 @@ export function getNextBillingCycleDate(curEndDateRaw: string): string {
...
+    ([credsResp], set) => {
+      if (!credsResp.data) return;
+      return getUsageMetrics(
+        credsResp.data.runtimeHost,
+        credsResp.data.instanceId,
+        credsResp.data.accessToken,
+      ).subscribe(set);
+    },
+  );
+}
```
ericpgreen2, line 81: Is it necessary to wrap this in a `derived` store? What about using a TanStack Query's `select` functionality?

ericpgreen2 (changes requested):

web-admin/src/features/billing/plans/selectors.ts
AdityaHegde: This is in the API response. Lets just add a comment here.

web-common/src/features/project/ProjectDeployer.ts
AdityaHegde: It was for some debugging.

web-admin/src/features/billing/plans/selectors.ts
AdityaHegde: `getUsageMetrics` is actually a query with a custom query function. I can rename it to clarify. Lets keep it as is to make sure it gets cached.

web-local/src/routes/(misc)/deploy/+page.svelte
AdityaHegde: It s a similar hack to runtime on local. I can add a comment.

web-local/src/routes/(misc)/deploy/+page.svelte
```diff
@@ -31,6 +32,23 @@
...
+    );
+    // hack for dev env
+    // TODO: find a better solution and get a url from backend
+    if (cloudUrl === "http://localhost:9090") {
+      cloudUrl = "http://localhost:3000";
+    }
+    const url = new URL(cloudUrl);
+    url.pathname = `/${$org}/-/settings/billing`;
+    url.searchParams.set("upgrade", "true");
+    upgradeHref = url.toString();
```
ericpgreen2, line 50: Given there's some complexity here, it'd be nice to quarantine this in a utility function `makeUpgradeHref` with comments that make it more readable.

ericpgreen2 (approved):

PR #6045, opened 2024-11-06
Refactor `splits` to `partitions`

begelundmuller:
Refactor our terminology to use "partitions" instead of "splits" in models. The main changes are:
- The `splits:` property in the model YAML has been changed to `partitions:`
- The `rill project splits` CLI command has been changed to `rill project partitions`

Contributes to https://github.com/rilldata/rill-private-issues/issues/788.


runtime/compilers/rillv1/parse_model.go
```diff
@@ -16,16 +16,17 @@ import (
...
-	InputProperties   map[string]any                          `yaml:",inline" mapstructure:",remain"`
-	Stage             struct {
+	commonYAML            `yaml:",inline" mapstructure:",squash"` // Only to avoid loading common fields into InputProperties
+	Refresh               *ScheduleYAML                           `yaml:"refresh"`
+	Timeout               string                                  `yaml:"timeout"`
+	Incremental           bool                                    `yaml:"incremental"`
+	State                 *DataYAML                               `yaml:"state"`
+	Partitions            *DataYAML                               `yaml:"partitions"`
+	Splits                *DataYAML                               `yaml:"splits"` // Deprecated: use "partitions" instead
+	PartitionsWatermark   string                                  `yaml:"partitions_watermark"`
```
pjain1, line 26: missing `splits_watermark` ?

proto/rill/runtime/v1/resources.proto
```diff
@@ -111,10 +111,10 @@ message ModelSpec {
   bool incremental = 13;
   string incremental_state_resolver = 14;
   google.protobuf.Struct incremental_state_resolver_properties = 15;
-  string splits_resolver = 18;
```
pjain1, line 114: Do we need to keep this for backwards compatibility ?

begelundmuller: This change is actually backwards compatible because the renamed field uses the same number (`18`). Protobuf stores fields with these numbers, not their names. So renaming only changes the names in JSON serialization in the APIs, but we don't need backwards compatibility there.

runtime/compilers/rillv1/parse_model.go
begelundmuller: No this is intentional – `splits` is the only field anyone uses right now (also it's a beta feature, so breaking changes are okay)

pjain1 (approved):

PR #6046, opened 2024-11-06
fix: pass original theme values

briangregoryholmes:
This PR updates the ThemeSpec to include the original raw color values in addition to the RGB object.


proto/rill/runtime/v1/resources.proto
```diff
@@ -614,6 +614,8 @@ message Theme {
 message ThemeSpec {
   optional Color primary_color = 1;
   optional Color secondary_color = 2;
+  string primary = 3;
+  string secondary = 4;
```
begelundmuller, line 618: Could we remove the old `Color` fields entirely or do you still use them?

If they can't be removed, could we call the new properties `primary_color_raw` and `secondary_color_raw` just to explicitly mirror the field names for the parsed colors?

briangregoryholmes: Removed

begelundmuller (approved):
LGTM

PR #6047, opened 2024-11-06
connector and name to API yaml

royendo:
Doing some testing with ClickHouse and API, noticed we dont have connector in the docs 
:)

### DIFF
--- a/docs/docs/reference/project-files/apis.md
+++ b/docs/docs/reference/project-files/apis.md
@@ -11,8 +11,13 @@ See comprehensive documentation on how to define and use [custom APIs](/integrat
 
 _**`type`**_ — Refers to the resource type and must be `api` _(required)_.
 
+_**`name`**_ — Refers to the name of the API. _(optional)_.
+
+
 Either one of the following:
 
 - _**`sql`**_ — General SQL query referring a [model](/build/models/models.md) _(required)_.
 
 - _**`metrics_sql`**_ — SQL query referring metrics definition and dimensions defined in the [metrics view](/build/dashboards/dashboards.md) _(required)_.
+
+_**`connector`**_ — Refers to the OLAP engine, if you are not using DuckDB, IE: [ClickHouse OLAP engine](../olap-engines/multiple-olap.md). _(optional)_.


docs/docs/reference/project-files/apis.md
```diff
@@ -11,8 +11,13 @@ See comprehensive documentation on how to define and use [custom APIs](/integrat
 
 _**`type`**_ — Refers to the resource type and must be `api` _(required)_.
 
+_**`name`**_ — Refers to the name of the API. _(optional)_.
```
begelundmuller, line 14: The `name:` property is not specific to API YAML, it works for all YAML resources. It defaults to the filename without file extension, which is sufficient for basically all use cases – so should we maybe not document it?

docs/docs/reference/project-files/apis.md
```diff
@@ -11,8 +11,13 @@ See comprehensive documentation on how to define and use [custom APIs](/integrat
...
+
 Either one of the following:
 
 - _**`sql`**_ — General SQL query referring a [model](/build/models/models.md) _(required)_.
 
-- _**`metrics_sql`**_ — SQL query referring metrics definition and dimensions defined in the [metrics view](/build/dashboards/dashboards.md) _(required)_.
\ No newline at end of file
+- _**`metrics_sql`**_ — SQL query referring metrics definition and dimensions defined in the [metrics view](/build/dashboards/dashboards.md) _(required)_.
+
+_**`connector`**_ — Refers to the OLAP engine, if you are not using DuckDB, IE: [ClickHouse OLAP engine](../olap-engines/multiple-olap.md). _(optional)_.
```
begelundmuller, line 23: Bit more complicated than that – it defaults to the current OLAP connector if not set, it's not hard-coded to DuckDB. Also it only matters for `sql:`, if you are using `metrics_sql:` it always uses the connector associated with the queried metrics view's underlying model.

docs/docs/reference/project-files/apis.md
royendo: true, will remove this one.

docs/docs/reference/project-files/apis.md
royendo: hmm, so maybe something like

- Refers to the OLAP engine if not already set in rill.yaml or in if using multiple OLAP connectors in a single project. Only applies when using `sql:`.

begelundmuller (approved):

PR #6053, opened 2024-11-07
nudge user to pass  flag when unable to infer project name

k-anshul:


begelundmuller (approved):

cli/cmd/env/configure.go
```diff
@@ -51,7 +51,7 @@ func ConfigureCmd(ch *cmdutil.Helper) *cobra.Command {
 				var err error
 				projectName, err = ch.InferProjectName(ctx, ch.Org, projectPath)
 				if err != nil {
-					return err
+					return fmt.Errorf("unable to infer project name, use `--project` to specify the project name")
```
begelundmuller, line 54: Are you sure it's safe to ignore the `err`? What if they are in the right Git repo, but there's a network error or permission error?

begelundmuller (changes requested):

cli/cmd/env/configure.go
k-anshul: Yes. IMO inference of project name is best effort and if it fails(mostly will succeed with `.rillcloud` being checked in now) best is to ask users to pass project name. 
For permissions the source of truth is permissions that are stored in rill so a user with no access to github repo but access to rill project should be able to make changes ?

begelundmuller: > mostly will succeed with .rillcloud being checked in now

Actually I think this is the problem – it will fail quite often now because someone else clones the repo, and for example tries to run Rill commands, but maybe hasn't run `rill login` yet. Then they will get an auth required error.

I think it might be better to still return the original error here along with the hint, like:
```go
fmt.Errorf("unable to infer project name (use `--project` to explicitly specify the name): %w", err)
```

k-anshul: Sure. Original error with `--project` message sounds good.

begelundmuller (approved):

PR #6065, opened 2024-11-08
fix(billing): Handle errored out metrics project request

AdityaHegde:
- [x] Handle the metrics project credentials request erroring out.
- [x] Use displayName in plan to keep the mapping in backend.
- [x] Show loading spinner instead of org selector during deploy.
- [x] Hide payment module for trial plans.

### DIFF
--- a/web-admin/src/features/billing/plans/PlanQuotas.svelte
+++ b/web-admin/src/features/billing/plans/PlanQuotas.svelte
@@ -23,7 +23,7 @@
       : "Unlimited";
 
   $: usageMetrics = getOrganizationUsageMetrics(organization);
-  $: total = $usageMetrics?.data?.reduce((s, m) => s + m.size, 0) ?? 0;
+  $: totalOrgSize = $usageMetrics?.data?.reduce((s, m) => s + m.size, 0) ?? 0;
 
   $: singleProjectLimit = $organizationQuotas.data?.projects === 1;
   $: storageLimitBytesPerDeployment =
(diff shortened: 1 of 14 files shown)


web-admin/src/features/billing/plans/PlanQuotas.svelte
```diff
@@ -23,7 +23,7 @@
       : "Unlimited";
 
   $: usageMetrics = getOrganizationUsageMetrics(organization);
-  $: total = $usageMetrics?.data?.reduce((s, m) => s + m.size, 0) ?? 0;
+  $: totalOrgSize = $usageMetrics?.data?.reduce((s, m) => s + m.size, 0) ?? 0;
```
ericpgreen2, line 26: Nit: I read `totalOrgSize` as "number of people in the organization". `totalOrgUsage`?

web-admin/src/features/billing/plans/utils.ts
```diff
@@ -4,18 +4,18 @@ import { DateTime } from "luxon";
...
 export function formatDataSizeQuota(
-  totalSizeInBytes: number,
+  sizeInBytes: number,
   storageLimitBytesPerDeployment: string,
 ): string {
   const maxSize = Number(storageLimitBytesPerDeployment);
   if (Number.isNaN(maxSize) || storageLimitBytesPerDeployment === "-1")
     return "";
-  const formattedTotal = formatMemorySize(totalSizeInBytes);
+  const formattedTotal = formatMemorySize(sizeInBytes);
```
ericpgreen2, line 13: Nit: "total" is hard to understand in this function. What if we used the terms "usage" (the numerator) and "quota" (the denominator)? The function name could be `formatUsageOfQuota`

web-admin/src/features/errors/error-utils.ts
```diff
@@ -139,6 +142,12 @@ export function createGlobalErrorCallback(queryClient: QueryClient) {
       return;
     }
 
+    // Handle case when usage metrics project is not available
+    // This is always the case on non-prod
+    if (isMetricsProjectQuery(query)) {
```
ericpgreen2, line 147: It's not clear to me why this function would return false outside of production?

ericpgreen2 (approved):

web-admin/src/features/errors/error-utils.ts
AdityaHegde: I have removed that comment. It could be down in prod as well.

ericpgreen2: Oh, this is a custom API, so if the Rill API resource has not successfully reconciled, the API path will be undefined or something?

AdityaHegde: Ya that could be the case on prod. Right now, on non-prod it is not preset at all.

PR #6069, opened 2024-11-10
Implement Incremental Merge Strategy for Clickhouse

rohithreddykota:
It involves the following steps:

1. Create a Temporary Table: A temporary table is created with the same schema as the target table.
2. Insert Data into Temporary Table: The new data is inserted into the temporary table.
3. List Temporary Table Partitions: The partitions in the temporary table are listed based on the specified key.
4. Replace Partitions: The partitions in the target table are replaced with the corresponding partitions from the temporary table.
5. Drop Temporary Table: The temporary table is dropped after the partitions have been replaced.

### Issue
https://github.com/rilldata/rill-private-issues/issues/892


runtime/drivers/clickhouse/olap.go
```diff
@@ -250,7 +250,57 @@ func (c *connection) InsertTableAsSelect(ctx context.Context, name, sql string,
...
+		// insert into temp table
+		err = c.Exec(ctx, &drivers.Statement{
+			Query:    fmt.Sprintf("INSERT INTO %s %s", safeSQLName(tempName), sql),
+			Priority: 1,
+		})
+		if err != nil {
+			return err
+		}
+		// list partitions from the temp table
+		res, err := c.Execute(ctx, &drivers.Statement{
```
k-anshul, line 279: I think it should use `partition` columns from `system.parts` table to list partition table. 
https://clickhouse.com/docs/en/sql-reference/statements/alter/partition#how-to-set-partition-expression

runtime/drivers/clickhouse/olap.go
```diff
@@ -250,7 +250,57 @@ func (c *connection) InsertTableAsSelect(ctx context.Context, name, sql string,
...
+			Priority: 1,
+		})
+		if err != nil {
+			return err
+		}
+		defer res.Close()
+		// iterate the partitions of the temp table
+		for res.Next() {
+			var part string
+			if err := res.Scan(&part); err != nil {
```
k-anshul, line 290: The result of the below query `fmt.Sprintf("SELECT DISTINCT %s FROM %s", strings.Join(key, ", "), safeSQLName(tempName))` can't be always scanned as string.

runtime/drivers/clickhouse/olap.go
```diff
@@ -250,7 +250,57 @@ func (c *connection) InsertTableAsSelect(ctx context.Context, name, sql string,
...
+		// create temp table with the same schema
+		tempName := tempName(name)
+		// drop the temp table
+		defer func() {
+			err := c.DropTable(ctx, tempName, false)
+			if err != nil {
+				c.logger.Error("clickhouse: failed to drop temp table", zap.String("name", tempName), zap.Error(err))
+			}
+		}()
+		err := c.Exec(ctx, &drivers.Statement{
```
k-anshul, line 263: Use `CREATE OR REPLACE`.
Is it guaranteed to work for distributed table in all case ? it is missing `on cluster clause` ?

runtime/drivers/duckdb/model_executor_warehouse_self.go
```diff
@@ -132,7 +132,16 @@ func (e *warehouseToSelfExecutor) queryAndInsert(ctx context.Context, opts *driv
...
 		if !create && opts.IncrementalRun {
-			err := olap.InsertTableAsSelect(ctx, outputTable, qry, false, true, outputProps.IncrementalStrategy, outputProps.UniqueKey)
+			var key []string
+			switch outputProps.IncrementalStrategy {
+			case drivers.IncrementalStrategyMerge:
+				key = outputProps.UniqueKey
+			case drivers.IncrementalStrategyReplace:
+				key = outputProps.ReplaceKey
+			default:
+				return fmt.Errorf("unsupported incremental strategy: %s", outputProps.IncrementalStrategy)
```
k-anshul, line 142: Throws error for `Append` strategy.

runtime/drivers/duckdb/model_manager.go
```diff
@@ -25,6 +25,7 @@ type ModelOutputProperties struct {
 	Table               string                      `mapstructure:"table"`
 	Materialize         *bool                       `mapstructure:"materialize"`
 	UniqueKey           []string                    `mapstructure:"unique_key"`
+	ReplaceKey          []string                    `mapstructure:"replace_key"`
```
k-anshul, line 28: Don't think we need replace strategy if both merge and replace works the same way. Also key should be same (`unique_key`).

k-anshul (changes requested):
Two more general comments:
1. May be I am missing something but replace looks exactly the same as merge so why call it a different strategy.
2. As I understand the partition replacement works only for the partition key of the table and not any other combination so we can't take any key input for `Merge`/`Replace` strategy and the partition key specified during table creation should work for unique_key as well. Also if the user specifies strategy as `Merge` we should validate that table is partitioned.

rohithreddykota:
> 1. May be I am missing something but replace looks exactly the same as merge so why call it a different strategy.
>

The Merge strategy in DuckDB is designed to replace data records based on any specified key, offering flexibility as the key can be any column. This approach is somewhat similar to using ReplacingMergeTree in ClickHouse, where data can be replaced based on primary keys.

The idea of the Replace strategy is to replace strictly at the partition level, using only the partition key to replace data. This means Replace does not support granular replacement on arbitrary fields, making it a more partition-specific approach. 

Also, it may not make sense for the Merge strategy to function differently in DuckDB compared to ClickHouse from an end-user perspective. By maintaining Merge and Replace as separate strategies, we can provide users clarity, ensuring they choose the appropriate method based on their specific data update requirements.

runtime/drivers/clickhouse/olap.go
rohithreddykota: Agree.

runtime/drivers/clickhouse/olap.go
rohithreddykota: Yes. that works on distributed table. This just creates an empty table with same schema. 

See https://clickhouse.com/docs/en/sql-reference/statements/create/table#with-a-schema-similar-to-other-table

>  it is missing on cluster clause

We can't pass cluster info here. It will be created on same cluster as existing table.

k-anshul:
> @k-anshul Here, we may be contradicting our own rules for merge. If user just uses merge, he must specify the key. Instead, we can allow user to just use `incremental_strategy: replace` and not allow to pass any key field. This can be another reason to have replace different from merge (fyi @begelundmuller )

I am still not fully convinced for adding another strategy because the end result is same for both. There can always be limitations around the functionalities depending on the underlying OLAP engine capabilities. 
Here is how I would put it.
`Merge` for duckdb works for arbitrary keys. `Merge` for clickhouse is limited in scope and only works for partition tables.

But yeah lets also wait for Benjamin/Eugene for their inputs.

k-anshul:
> Here, we may be contradicting our own rules for merge. If user just uses merge, he must specify the key. Instead, we can allow user to just use incremental_strategy: replace and not allow to pass any key field. This can be another reason to have replace different from merge 

The disparity between strategies for different OLAP engines will still exist. `Replace` is an alias for `Merge` for duckDB meaning user can(and must) provide keys. `Replace` for clickhouse does not take key input.

esevastyanov:
> Merge for duckdb works for arbitrary keys. Merge for clickhouse is limited in scope and only works for partition tables.

this sounds like a simpler approach since doesn't require introducing new syntax 

> If user just uses merge, he must specify the key

can the logic be updated so that no key is required in case of clickhouse?

rohithreddykota:
Thanks, I will make changes accordingly @k-anshul @esevastyanov

runtime/drivers/clickhouse/olap.go
esevastyanov: is `system.parts` strong or eventually consistent?

rohithreddykota: I think it does not matter. The data is being considered as ingested into temp table only after the partition key is added to the parts table. [See here](https://arc.net/l/quote/rhgwrgoe)

runtime/drivers/clickhouse/olap.go
rohithreddykota: Distributed is an abstract table => there is a base table => it is possible to manipulate on the base table. Thanks @esevastyanov

runtime/drivers/clickhouse/olap.go
```diff
@@ -251,7 +252,86 @@ func (c *connection) InsertTableAsSelect(ctx context.Context, name, sql string,
...
+			return err
+		}
+		defer res.Close()
+		if res.Next() {
+			if err := res.Scan(&engine, &engineFull); err != nil {
+				return err
+			}
+		}
+		// Distributed table cannot be altered directly, so we need to alter the underlying table
+		if engine == "Distributed" {
```
k-anshul, line 281: As of now the local table name is fixed : https://github.com/rilldata/rill/blob/b9e61923dd5b6fd074a0b370ae0c153b7dccbc07/runtime/drivers/clickhouse/olap.go#L1017-L1018

runtime/drivers/clickhouse/olap.go
```diff
@@ -251,7 +252,86 @@ func (c *connection) InsertTableAsSelect(ctx context.Context, name, sql string,
...
+			Priority: 1,
+		})
+		if err != nil {
+			return err
+		}
+		defer res.Close()
+		// iterate over partitions and replace them in the main table
+		for res.Next() {
+			var part string
+			if err := res.Scan(&part); err != nil {
```
k-anshul, line 320: I think we should collect all parts first and then loop over collected parts. Though I assume the alter queries are just meta queries but it will be good to free the connection that executed select.

runtime/drivers/clickhouse/olap.go
```diff
@@ -251,7 +252,86 @@ func (c *connection) InsertTableAsSelect(ctx context.Context, name, sql string,
...
+			// Example engine_full: Distributed('cluster', 'db', 'underlying_table', sharding_key)
+			re := regexp.MustCompile(`Distributed\('[^,]+',\s*'[^']+',\s*'([^']+)'`)
+			matches := re.FindStringSubmatch(engineFull)
+			if len(matches) < 1 {
+				return fmt.Errorf("unable to parse underlying table info from engine expression: %q", engineFull)
+			}
+			name = matches[1]
+		}
+		// create temp table with the same schema
+		tempName := tempName(name)
```
k-anshul, line 292: Also need to cleanup this table.

runtime/drivers/clickhouse/olap.go
```diff
@@ -251,7 +252,86 @@ func (c *connection) InsertTableAsSelect(ctx context.Context, name, sql string,
...
+		res, err = c.Execute(ctx, &drivers.Statement{
+			Query:    fmt.Sprintf("SELECT DISTINCT partition FROM system.parts WHERE table = '%s'", tempName),
+			Priority: 1,
+		})
+		if err != nil {
+			return err
+		}
+		defer res.Close()
+		// iterate over partitions and replace them in the main table
+		for res.Next() {
```
k-anshul, line 318: I don't think I have a solution for this but just surfacing the issue. 
Should we be worried about case if some partitions are replaced and then next query fails.

k-anshul (changes requested):

k-anshul:
I think it will also be good to add some validation that `incremental strategy = replace` is only added for models which create partitioned tables.

runtime/drivers/clickhouse/olap.go
esevastyanov: The operation might be safely repeated

runtime/drivers/clickhouse/olap.go
```diff
@@ -251,7 +251,94 @@ func (c *connection) InsertTableAsSelect(ctx context.Context, name, sql string,
...
+		var engine string
+		res, err := c.Execute(ctx, &drivers.Statement{
+			Query:    "SELECT engine FROM system.tables WHERE database = currentDatabase() AND name = ?",
+			Args:     []any{name},
+			Priority: 1,
+		})
+		if err != nil {
+			return err
+		}
+		defer res.Close()
```
k-anshul, line 273: May want to release this connection early. Two ideas :
1. Close this after scanning and if scan fails.
2. Create a simple utility function to get `engine` where the `defer` construct can be used.

runtime/drivers/clickhouse/olap.go
```diff
@@ -251,7 +251,94 @@ func (c *connection) InsertTableAsSelect(ctx context.Context, name, sql string,
...
+		}
+		// list partitions from the temp table using system.parts
+		res, err = c.Execute(ctx, &drivers.Statement{
+			Query:    fmt.Sprintf("SELECT DISTINCT partition FROM system.parts WHERE table = '%s'", tempName),
+			Priority: 1,
+		})
+		if err != nil {
+			return err
+		}
+		defer res.Close()
```
k-anshul, line 318: This still keeps the connection blocked till this function returns. You can use the same approach as noted above.
May be I am overthinking and it is fine to keep this connection blocked since the subsequent queries are going to be fast.

runtime/drivers/clickhouse/olap.go
k-anshul: But intermittently table will result partial data from older insert and partial data from newer insert.

k-anshul (changes requested):

runtime/drivers/clickhouse/olap.go
k-anshul: But intermittently table will report data from both older and newer(partial successful) inserts.

runtime/drivers/clickhouse/olap.go
```diff
@@ -659,6 +728,46 @@ func (c *connection) acquireConn(ctx context.Context) (*sqlx.Conn, func() error,
...
+	if res.Next() {
+		if err := res.Scan(&engine); err != nil {
+			return "", err
+		}
+	}
+	return engine, nil
+}
+
+func (c *connection) getTablePartitions(ctx context.Context, name string) ([]string, error) {
+	res, err := c.Execute(ctx, &drivers.Statement{
```
k-anshul, line 751: Pass it as `arg` similar to function above(preferable)  or wrap name in `safeSQLName`.

runtime/drivers/clickhouse/olap.go
```diff
@@ -477,6 +540,12 @@ func (c *connection) createTable(ctx context.Context, name, sql string, outputPr
 	} else {
 		fmt.Fprintf(&create, " %s ", outputProps.Columns)
 	}
+
+	// validate incremental strategy
+	if outputProps.IncrementalStrategy == drivers.IncrementalStrategyMerge && outputProps.PartitionBy == "" {
```
k-anshul, line 545: It is also possible to specify full engine directly without individual keys so I think we can just do string search for `PARTITION BY` in string returned from `tblConfig()`(https://github.com/rilldata/rill/blob/96c119ed46051c79649916826cdf43e8a1a773cd/runtime/drivers/clickhouse/model_manager.go#L104)

k-anshul (approved):

runtime/drivers/clickhouse/olap.go
```diff
@@ -251,7 +251,70 @@ func (c *connection) InsertTableAsSelect(ctx context.Context, name, sql string,
...
+			Query:    fmt.Sprintf("CREATE OR REPLACE TABLE %s %s AS %s", safeSQLName(tempName), onClusterClause, name),
+			Priority: 1,
+		})
+		if err != nil {
+			return err
+		}
+		// clean up the temp table
+		defer func() {
+			err = c.Exec(ctx, &drivers.Statement{
+				Query:    fmt.Sprintf("DROP TABLE %s %s", safeSQLName(tempName), onClusterClause),
```
begelundmuller, line 284: There is a risk that the `ctx` is cancelled (e.g. if the pod is restarted or it hits the reconciliation timeout). Since `tempName` function returns a random table name, this means the table will never be cleaned up.

Imagine if this operation is too slow for the current resources and keeps getting cancelled/retried. Then a lot of temp tables can accumulate.

Have you considered one or both of these options:
1. If `ctx.Err() != nil`, then creating a new short-lived background context to provide a brief opportunity for cleanup (e.g. `context.Background().WithTimeout(15*time.Second)`).
2. Changing `tempName` to return a deterministic temporary table name (not sure, but could be e.g. `__rill_temp_{table name}_{sql hash}`). This would ensure there's at most one orphaned temp table per table/split.

runtime/drivers/clickhouse/olap.go
```diff
@@ -251,7 +251,70 @@ func (c *connection) InsertTableAsSelect(ctx context.Context, name, sql string,
...
+		// list partitions from the temp table
+		partitions, err := c.getTablePartitions(ctx, tempName)
+		if err != nil {
+			return err
+		}
+		// iterate over partitions and replace them in the main table
+		for _, part := range partitions {
+			// alter the main table to replace the partition
+			err = c.Exec(ctx, &drivers.Statement{
+				Query:    fmt.Sprintf("ALTER TABLE %s %s REPLACE PARTITION %s FROM %s", safeSQLName(name), onClusterClause, part, safeSQLName(tempName)),
```
begelundmuller, line 308: So this is more like a "partition overwrite" strategy than a merge? (I.e. it replaces partitions instead of upserting/merging rows). If yes, would it make sense to add a new `drivers.IncrementalStrategyPartitionOverwrite` instead of using `drivers.IncrementalStrategyMerge`?

In ClickHouse, am I correct in saying that a normal "merge" strategy (that does upserts, not partition replaces like this does) can be achieved by using the [`ReplacingMergeTree`](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/replacingmergetree) table engine, and doesn't require special insert implementation?

esevastyanov: `ReplacingMergeTree` is eventually consistent so duplicate rows may exist temporarily until the background merges occur.
Current implementation of the merge strategy is pretty much similar to the DuckDB's but relies on a partition key instead of a custom one.

begelundmuller: > is eventually consistent so duplicate rows may exist temporarily until the background merges occur.
Got it, makes sense (though it sounds like an annoying property)

> but relies on a partition key instead of a custom one

So my question is if this means that it will replace/overwrite full partitions? I.e. can this lead to rows being removed (which is not possible in the DuckDB implementation). If it can lead to rows being removed, then I think we should call the strategy `partition_overwrite` instead (or something like that).

For reference, take a look at how DBT distinguishes between their `merge` and `insert_overwrite` incremental strategies: https://docs.getdbt.com/reference/resource-configs/bigquery-configs#merge-behavior-incremental-models

esevastyanov: > Got it, makes sense (though it sounds like an annoying property)

The background merge might be [triggered manually](https://clickhouse.com/docs/en/engines/table-engines/mergetree-family/replacingmergetree)
> Although you can run an unscheduled merge using the `OPTIMIZE` query, do not count on using it, because the `OPTIMIZE` query will read and write a large amount of data

We can implement such approach for small datasets in ClickHouse `merge` strategy.

> I think we should call the strategy `partition_overwrite` instead

Agree, lets call it so or `insert_overwrite` if we want to follow some common syntax

begelundmuller (approved):

runtime/drivers/clickhouse/olap.go
begelundmuller: > We can implement such approach for small datasets in ClickHouse merge strategy.

It sounds like there might still be a small period of inconsistency even if we trigger it manually? Might be best to avoid the manual triggers.

I wonder if CH has a way to not serve new rows for a `ReplacingMergeTree` until the background replace has completed. Or some kind of watermark indicating up to what row ID that the table is consistent, which we could pass in metrics queries.

> Agree, lets call it so or `insert_overwrite` if we want to follow some common syntax

IMO `partition_overwrite` sounds more correct than `insert_overwrite` (which sounds like a convoluted way of saying "upsert"). But if you prefer to align with DBT terminology, that makes sense and also sounds good to me – you decide :)

esevastyanov: > small period of inconsistency even if we trigger it manually? 

the period depends on data volume, it might be significant if the volume is large.

> which we could pass in metrics queries

yes, `FINAL` keyword might be [passed in queries](https://clickhouse.com/docs/en/guides/replacing-merge-tree#querying-replacingmergetree) as a guarantee of correctness.

> To obtain correct answers, users will need to complement background merges with query time deduplication and deletion removal. This can be achieved using the FINAL operator.

But `FINAL` has a [performance overhead](https://clickhouse.com/docs/en/guides/replacing-merge-tree#final-performance) and needs to be used wisely

PR #6071, opened 2024-11-11
Remove implicit edit instance invocation from validate deployments job

kaspersjo:
Fixes: https://github.com/rilldata/rill-private-issues/issues/833
Fixes: https://github.com/rilldata/rill-private-issues/issues/893


begelundmuller (approved):
LGTM

PR #6082, opened 2024-11-12
Rill E2E setup

himadrisingh:
`rill devtool start e2e`

### DIFF
--- a/cli/cmd/devtool/data/cloud-deps.docker-compose.yml
+++ b/cli/cmd/devtool/data/cloud-deps.docker-compose.yml
@@ -1,4 +1,3 @@
-version: "3.9"
 services:
   postgres:
     image: postgres
--- a/cli/cmd/devtool/data/cloud-deps.docker-compose.yml
+++ b/cli/cmd/devtool/data/cloud-deps.docker-compose.yml
@@ -1,6 +1,6 @@
 services:
   postgres:
-    image: postgres
+    image: postgres:17
     restart: always
     environment:
       - POSTGRES_USER=postgres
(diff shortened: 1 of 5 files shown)


begelundmuller (approved):
LGTM

ericpgreen2 (changes requested):
1. This looks incorrect. I'd expect it to pull `cloud-e2e.env`.
    image
2. I ran `rill devtool start e2e` and I see that it's connected to the same Postgres database as `rill devtool start cloud`. I'd like `rill devtool start e2e` to always start with a fresh Postgres database (and for my `rill devtool start cloud` Postgres database to remain untouched). Can we do that? (This comment may be resolved if this behavior is due to point 1.)
3. Also, it'd be nice to have the option to run `rill devtool seed e2e` to seed the e2e database in the same way as `rill devtool seed cloud`.

himadrisingh:
> 1. This looks incorrect. I'd expect it to pull `cloud-e2e.env`.
>    image
Fixed
> 2. I ran `rill devtool start e2e` and I see that it's connected to the same Postgres database as `rill devtool start cloud`. I'd like `rill devtool start e2e` to always start with a fresh Postgres database (and for my `rill devtool start cloud` Postgres database to remain untouched). Can we do that? (This comment may be resolved if this behavior is due to point 1.)
Created an ephemeral postgres for e2e only, will get cleaned up everytime we restart.
> 3. Also, it'd be nice to have the option to run `rill devtool seed e2e` to seed the e2e database in the same way as `rill devtool seed cloud`.
not able to make it work for cloud too :(

ericpgreen2:
> Created an ephemeral postgres for e2e only, will get cleaned up everytime we restart.

Wondering – what specifically do you mean by this? Locally, on a subsequent run, I only was able to reset the db via `rill devtool start e2e --reset`. When I didn't pass `--reset`, my state was retained (which is fine; the semantics are clear).

himadrisingh:
> > Created an ephemeral postgres for e2e only, will get cleaned up every time we restart.
> 
> Wondering – what specifically do you mean by this? Locally, on a subsequent run, I only was able to reset the db via `rill devtool start e2e --reset`. When I didn't pass `--reset`, my state was retained (which is fine; the semantics are clear).

So essentially if u restart docker then the volume is gone :) 
dev is persisted to disk under `dev-cloud-state` but not of e2e. So yeah we need to reset or restart the docker. 
we can change the behavior i believe.

ericpgreen2 (approved):
Great, this works.

PR #6083, opened 2024-11-12
model executor for postgres and mysql to duckdb

k-anshul:

- Adds a model executor that ingests data from postgres/mysql using duckdb extension.
- Also supports incremental ingestion.

Sample model yaml that ingests from postgres.
```
connector: postgres
sql: SELECT * FROM students 
dsn: "postgresql://postgres:postgres@localhost:5432/postgres"

output:
  connector: duckdb
```

### DIFF
--- a/runtime/drivers/duckdb/model_executor_sqlstore_self.go
+++ b/runtime/drivers/duckdb/model_executor_sqlstore_self.go
@@ -11,6 +11,18 @@ import (
 	"github.com/rilldata/rill/runtime/drivers/postgres"
 )
 
+type sqlStoreToSelfInputProps struct {
+	SQL string `mapstructure:"sql"`
+	DSN string `mapstructure:"dsn"`
+}
+
+func (p *sqlStoreToSelfInputProps) Validate() error {
+	if p.SQL == "" {
+		return fmt.Errorf("missing property 'sql'")
+	}
+	return nil
+}
+
 type sqlStoreToSelfExecutor struct {
 	c *connection
 }
@@ -25,7 +37,7 @@ func (e *sqlStoreToSelfExecutor) Concurrency(desired int) (int, bool) {
 }
 
 func (e *sqlStoreToSelfExecutor) Execute(ctx context.Context, opts *drivers.ModelExecuteOptions) (*drivers.ModelResult, error) {
-	inputProps := &ModelInputProperties{}
+	inputProps := &sqlStoreToSelfInputProps{}
 	if err := mapstructure.WeakDecode(opts.InputProperties, inputProps); err != nil {
 		return nil, fmt.Errorf("failed to parse input properties: %w", err)
 	}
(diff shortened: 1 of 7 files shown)


runtime/drivers/duckdb/model_executor_sqlstore_self.go
```diff
@@ -0,0 +1,101 @@
...
+	DSN string `mapstructure:"dsn"`
+}
+
+func (p *sqlStoreToSelfInputProps) Validate() error {
+	if p.SQL == "" {
+		return fmt.Errorf("missing property 'sql'")
+	}
+	if p.DSN == "" {
+		return fmt.Errorf("missing property `dsn`")
+	}
```
begelundmuller, line 23: The DSN or other connection properties (like `host`, etc.) may have been configured in a `postgres.yaml` file instead of inline. You should be able to access that info through `ModelExecutorOptions.InputHandle`.

runtime/drivers/duckdb/model_executor_sqlstore_self.go
```diff
@@ -0,0 +1,101 @@
...
+		return nil, err
+	}
+	clone.InputProperties = newInputProps
+
+	newOutputProps := make(map[string]any)
+	err = mapstructure.WeakDecode(outputProps, &newOutputProps)
+	if err != nil {
+		return nil, err
+	}
+	clone.OutputProperties = newOutputProps
```
begelundmuller, line 71: Is this needed? Could we just let `selfToSelfExecutor` handle the output props (since we aren't rewriting it here)?

runtime/drivers/duckdb/model_executor_sqlstore_self.go
```diff
@@ -0,0 +1,101 @@
...
+	newOpts := &clone
+
+	// execute
+	executor := &selfToSelfExecutor{c: e.c}
+	return executor.Execute(ctx, newOpts)
+}
+
+func (e *sqlStoreToSelfExecutor) modelInputProperties(modelName, inputDriver string, inputProps *sqlStoreToSelfInputProps) (map[string]any, error) {
+	m := &ModelInputProperties{}
+	dbName := modelName + "_external_db_"
```
begelundmuller, line 81: What about using `<model name>__<opts.InputConnector>` instead?

runtime/drivers/duckdb/transporter_mysql_to_duckDB_test.go
```diff
@@ -123,3 +132,55 @@ func allMySQLDataTypesTest(t *testing.T, db *sql.DB, dsn string) {
 	require.NoError(t, res.Close())
 	require.NoError(t, to.Close())
 }
+
+func mysqlToDuckDB(t *testing.T, dsn string) {
```
begelundmuller, line 136: Can we move this test to a test file that starts with `model_executor_` instead? For when we drop the `transporter` files?

runtime/drivers/duckdb/transporter_postgres_to_duckDB_test.go
```diff
@@ -86,3 +93,80 @@ func allDataTypesTest(t *testing.T, db *sql.DB, dbURL string) {
 	require.NoError(t, res.Close())
 	require.NoError(t, to.Close())
 }
+
+func pgxToDuckDB(t *testing.T, pgdb *sql.DB, dbURL string) {
```
begelundmuller, line 97: Similar comment about moving to a model executor related test file

begelundmuller (changes requested):

runtime/drivers/duckdb/model_executor_sqlstore_self.go
k-anshul: Sure. This is temporary so we just need unique name.

runtime/drivers/duckdb/transporter_mysql_to_duckDB_test.go
k-anshul: Actually I planned to rename this file and drop the transporter test when re drop the transporter files. Till then kept it in the same file to avoid creating more test containers and keep test time short.
I am okay to also create new test if you prefer that way.

runtime/drivers/postgres/postgres.go
```diff
@@ -56,6 +56,10 @@ var spec = drivers.Spec{
 
 type driver struct{}
 
+type ConfigProperties struct {
+	DatabaseURL string `mapstructure:"database_url"`
```
k-anshul, line 60: We are in this weird situation where everywhere else it is called a DSN but called `database_url` in postgres because it the more popular word in postgres world. May be we keep both and let user specify any one ?

runtime/drivers/duckdb/model_executor_sqlstore_self.go
k-anshul: We will need to create a model and connector both from a source definition.

k-anshul: But it will be easy to rewrite sources to models if allow inline properties. I have split opinions on this.
What do you suggest ?

runtime/drivers/postgres/postgres.go
begelundmuller: Yeah supporting both and aliasing one as the other seems good to me

runtime/drivers/duckdb/model_executor_sqlstore_self.go
begelundmuller: I'm also torn on this, I will generally favor the overall simplest solution, keeping in mind that we must keep backwards compatibility for existing sources.

One idea I can think of is to use nested values in `connector:` in models for inline-defined connectors. And then for backwards compatibility with sources, moving the few known properties like `dsn:` and similar to it. Specifically, the YAML would be:
```type
type: model
connector:
  driver: postgres
  dsn: psql://....
...
```

This should then emit a connector named `<model_name>__<connector>` and set it as the model's `InputConnector`. 

Would this work? Again, it's just a high-level idea and I'm overall biased towards not having any solutions here that get too complex/weird.

runtime/drivers/duckdb/model_executor_sqlstore_self.go
```diff
@@ -0,0 +1,118 @@
...
+		if dsn == "" {
+			// may be configured via a connector
+			var config *postgres.ConfigProperties
+			if err := mapstructure.Decode(inputHandle.Config(), &config); err != nil {
+				return nil, err
+			}
+			dsn = config.ResolveDSN()
+		}
+		if dsn == "" {
+			return nil, fmt.Errorf("must set `dsn` for models that transfer data from `postgres` to `duckdb`")
```
begelundmuller, line 105:
```suggestion
			return nil, fmt.Errorf("must set `database_url` or `dsn` for models that transfer data from `postgres` to `duckdb`")
```

begelundmuller (approved):

PR #6088, opened 2024-11-13
null filling and coalescing for mv agg query

pjain1:
These PR adds supports for two features-
1. `fill_missing` - This is a metrics view aggregation query level config and when a computed time dimension is used, it fills in the missing time buckets in the response.
2. `treat_nulls_as` - This is a measure level config used to configure what value to fill in for missing time buckets. This also works generally as COALESCING over non empty time buckets.

Closes https://github.com/rilldata/rill-private-issues/issues/786


runtime/metricsview/executor.go
```diff
@@ -214,6 +215,10 @@ func (e *Executor) Query(ctx context.Context, qry *Query, executionTime *time.Ti
 			return nil, err
 		}
 
+		for k, v := range ast.olapContext {
+			args = append(args, sql2.NamedArg{Name: drivers.DialectDruid.ContextKeyArgPrefix() + k, Value: v})
```
begelundmuller, line 219: Could we avoid a `Druid` dialect hard-coding here by having something like `additionalArgs` instead of `olapContext`? 

Then when processing the spine in `ast.go`, the dialect might either return extra args or a range SELECT query.

runtime/metricsview/query.go
```diff
@@ -124,6 +124,7 @@ type TimeSpine struct {
 	Start time.Time `mapstructure:"start"`
 	End   time.Time `mapstructure:"end"`
 	Grain TimeGrain `mapstructure:"grain"`
+	Alias string    `mapstructure:"alias"`
```
begelundmuller, line 127: What does this do? Ideally it should apply only to the time dimension if it's requested, and in that case, the time dimension already has an alias provided in `Dimension`

runtime/queries/metricsview_aggregation.go
```diff
@@ -305,6 +310,40 @@ func (q *MetricsViewAggregation) rewriteToMetricsViewQuery(export bool) (*metric
...
+		if qry.Spine != nil {
+			// should we silently ignore instead of error ?
+			return nil, fmt.Errorf("cannot have both where and time spine")
+		}
+		if len(computedTimeDims) > 1 {
+			return nil, fmt.Errorf("cannot have multiple time spines")
+		}
+		if q.TimeRange == nil || q.TimeRange.Start == nil || q.TimeRange.End == nil {
+			return nil, fmt.Errorf("time range is required for null fill")
+		}
```
begelundmuller, line 324: I think it might still work if there's a relative time range (like `q.TimeRange.IsoDuration`), and this might be needed e.g. for alerts/reports that use a time spine.

It would be resolved to a fixed time range before the query is turned into an AST. It happens here: https://github.com/rilldata/rill/blob/1d51e8ca44aa5a762cdf9bccffc841e858d514c4/runtime/metricsview/executor_rewrite_time.go#L15

runtime/queries/metricsview_aggregation.go
```diff
@@ -305,6 +310,40 @@ func (q *MetricsViewAggregation) rewriteToMetricsViewQuery(export bool) (*metric
...
+
+		qry.Spine = &metricsview.Spine{}
+		s := timeutil.TruncateTime(q.TimeRange.Start.AsTime(), timeDim.Compute.TimeFloor.Grain.ToTimeutil(), tz, 1, 1)
+		e := q.TimeRange.End.AsTime()
+		qry.Spine.TimeRange = &metricsview.TimeSpine{
+			Start: s,
+			End:   e,
+			Grain: timeDim.Compute.TimeFloor.Grain,
+			Alias: timeDim.Name,
```
begelundmuller, line 343: I think it should be handled inside the `metricsview` package by applying it to any dimension that has a `TimeFloor` applied (and erroring if zero or multiple dimensions have a `TimeFloor`)

begelundmuller (changes requested):

begelundmuller (commented):
As a separate comment, can we add some tests for this feature for all of DuckDB, Druid and ClickHouse in `runtime/resolvers/testdata`?

runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go
```diff
@@ -415,43 +423,45 @@ func (chc *coordinatorHTTPCheck) IsHardFailure(ctx context.Context) (bool, error
...
+	queryCtx := make(map[string]any)
+	queryCtx[sqlQueryIDContextKey] = uuid.New().String()
+	queryCtx[enableTimeBoundaryPlanningContextKey] = true
+
+	var parameters []DruidParameter
+	for _, arg := range args {
+		if strings.HasPrefix(arg.Name, drivers.DialectDruid.ContextKeyArgPrefix()) {
+			queryCtx[strings.TrimPrefix(arg.Name, drivers.DialectDruid.ContextKeyArgPrefix())] = arg.Value
+			continue
+		}
```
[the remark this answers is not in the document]
pjain1: removing this, going to use https://druid.apache.org/docs/latest/querying/datasource/#inline

runtime/metricsview/ast.go
```diff
@@ -689,7 +699,87 @@ func (a *AST) buildSpineSelect(alias string, spine *Spine, tr *TimeRange) (*Sele
 	}
 
 	if spine.TimeRange != nil {
-		return nil, errors.New("time_range not yet supported in spine")
+		if a.dialect == drivers.DialectDruid {
```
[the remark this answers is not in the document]
pjain1: moving this to olap package causing cyclic dependency because of `metricsview.TimeGrain`, metricsview package need drivers package

runtime/metricsview/ast.go
```diff
@@ -689,7 +699,87 @@ func (a *AST) buildSpineSelect(alias string, spine *Spine, tr *TimeRange) (*Sele
...
+				FromTable: &from,
+			}
+		} else if a.dialect == drivers.DialectClickHouse {
+			rangeSelect = &SelectNode{
+				Alias: alias,
+			}
+
+			// ClickHouse does not support range() function, so we need to generate a series of timestamps
+			for t := spine.TimeRange.Start; t.Before(spine.TimeRange.End); t = timeutil.AddTime(t, spine.TimeRange.Grain.ToTimeutil(), 1) {
+				rangeSelect.UnionAllSelects = append(rangeSelect.UnionAllSelects, &SelectNode{
```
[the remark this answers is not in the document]
pjain1: moving this to `drivers/olap.go` causing cyclic dependency because of metricsview.TimeGrain, metricsview package need drivers package

runtime/metricsview/executor.go
pjain1: removed

runtime/metricsview/query.go
pjain1: This is used while constructing the inline select statement for time ranges and for also removing time dim from dim select node which will be cross joined with range select. Yes its present in the query dimension however to use it here it will need to passed down separately as in this context only ast dimensions are present.

runtime/queries/metricsview_aggregation.go
pjain1: `WhereSpine` is created in the query package so did similar. You mean to create this inside `NewAST` ?

runtime/metricsview/ast.go
```diff
@@ -648,6 +655,9 @@ func (a *AST) buildBaseSelect(alias string, comparison bool) (*SelectNode, error
 		if err != nil {
 			return nil, err
 		}
+		if sn == nil {
+			return n, nil
+		}
```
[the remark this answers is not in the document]
pjain1: removed this

runtime/metricsview/ast.go
```diff
@@ -730,44 +820,49 @@ func (a *AST) addMeasureField(n *SelectNode, m *runtimev1.MetricsViewSpec_Measur
...
 	// We recurse on the sub-select and add a pass-through field to the current node.
 
 	if !hasMeasure(n.FromSelect, m.Name) { // Don't recurse if already in scope in sub-query
-		err := a.addSimpleMeasure(n.FromSelect, m)
+		f, err := a.addSimpleMeasure(n.FromSelect, m)
 		if err != nil {
-			return err
+			return nil, err
 		}
+		f.TreatNullAs = "" // only have to set this at the top level
```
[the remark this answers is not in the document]
pjain1: > If it only sets it at the top level, does that mean it won't be applied to the underlying when there's a derived window measures? I think that might be unexpected.

Doesn't we only need to set this for simple measure and any derived window measure will just get the correct values to work on after being coalesced and joined with the time spine ? I tried it with [bids_1day_rolling_avg](https://github.com/rilldata/rill/blob/fef2386bb980101ebbdd9f7c5ab2b41c3e0d60b4/runtime/testruntime/testdata/ad_bids/dashboards/ad_bids_metrics_advanced.yaml#L51) and added `treat_nulls_as` on the base measure and it works fine.

proto/rill/runtime/v1/resources.proto
```diff
@@ -205,6 +205,7 @@ message MetricsViewSpec {
     string format_d3 = 7;
     google.protobuf.Struct format_d3_locale = 13;
     bool valid_percent_of_total = 6;
+    string treat_nulls_as = 14; // TODO what should the type, using string values will not work when coalescing numeric cols
```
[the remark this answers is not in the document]
pjain1: yes thats what it does currently

runtime/metricsview/ast.go
```diff
@@ -644,7 +648,7 @@ func (a *AST) buildBaseSelect(alias string, comparison bool) (*SelectNode, error
 	// If there is a spine, we wrap the base SELECT in a new SELECT that we add the spine to.
 	// We do not join the spine directly to the FromTable because the join would be evaluated before the GROUP BY,
 	// which would impact the measure aggregations (e.g. counts per group would be wrong).
-	if a.query.Spine != nil {
+	if a.query.Spine != nil && !(a.query.Spine.TimeRange != nil && comparison) {
```
begelundmuller, line 651: This was quite hard to understand, can this be made more explicit? E.g. at the very least a comment:
```suggestion
	if a.query.Spine != nil && !(a.query.Spine.TimeRange != nil && comparison) { // Skip time range spines in the comparison select
```

begelundmuller (changes requested):

runtime/metricsview/ast.go
pjain1: Agree it was hard to write as well

begelundmuller (approved):

PR #6091, opened 2024-11-15
fix: autogenerate metrics indentation

AdityaHegde:
We are generating metrics with additional spaces. Also `name` is missing in dimensioins.
```
# Metrics view YAML

version: 1
type: metrics_view

display_name: Ad Bids
model: ad_bids
timeseries: timestamp

dimensions:
  - 
    display_name: Publisher
    column: publisher
  - 
    display_name: Domain
    column: domain

measures:
  - name: total_records
    
    display_name: Total records
    expression: COUNT(*)
    description: ""
    format_preset: humanize
  - name: bid_price_sum
    
    display_name: Sum of Bid Price
    expression: SUM(bid_price)
    description: ""
    format_preset: humanize
```

Removing the new line before display_name and adding after type instead.

After,
```
# Metrics view YAML

version: 1
type: metrics_view 

display_name: Ad Bids
model: ad_bids
timeseries: timestamp

dimensions:
  - name: publisher
    display_name: Publisher
    column: publisher
  - name: domain
    display_name: Domain
    column: domain

measures:
  - name: total_records
    display_name: Total records
    expression: COUNT(*)
    description: ""
    format_preset: humanize
  - name: bid_price_sum
    display_name: Sum of Bid Price
    expression: SUM(bid_price)
    description: ""
    format_preset: humanize
```

### DIFF
--- a/runtime/server/generate_metrics_view.go
+++ b/runtime/server/generate_metrics_view.go
@@ -379,6 +379,7 @@ func generateMetricsViewYAMLSimpleDimensions(schema *runtimev1.StructType) []*me
 		switch f.Type.Code {
 		case runtimev1.Type_CODE_BOOL, runtimev1.Type_CODE_STRING, runtimev1.Type_CODE_BYTES, runtimev1.Type_CODE_UUID:
 			dims = append(dims, &metricsViewDimensionYAML{
+				Name:        f.Name,
 				DisplayName: identifierToDisplayName(f.Name),
 				Column:      f.Name,
 			})
@@ -447,6 +448,7 @@ type metricsViewYAML struct {
 }
 
 type metricsViewDimensionYAML struct {
+	Name        string `yaml:"name"`
 	DisplayName string `yaml:"display_name"`
 	Column      string `yaml:"column"`
 }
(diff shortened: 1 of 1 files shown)


runtime/server/generate_metrics_view.go
```diff
@@ -447,6 +448,7 @@ type metricsViewYAML struct {
 }
 
 type metricsViewDimensionYAML struct {
+	Name        string `yaml:"name"`
```
begelundmuller, line 451: Are you sure we want to include `name` here? The default behavior is that it's the same as `column`, which is simpler and sufficient for most dimensions

AdityaHegde: Ya for consistency it makes sense. While it is not needed it will be good for users to get this while autogenerated.

AdityaHegde: Tagging @ericpgreen2 who asked for it for more clarity.

ericpgreen2: Yeah I think we should populate it – because the Visual Metrics Editor considers `name` a required field in the Add/Edit Dimension form.

begelundmuller: Sounds like that needs to be fixed regardless in the visual editor then? Since there are many metrics views with only a `column:` field for dimensions?

Also, what's the workflow here for people who are adding/changing a dimension that maps directly to an underlying column? Do they have to enter/update the column in two places to get the expected behavior where the dimension name matches the underlying column name?

AdityaHegde: We will be making `name` mandatory during parsing as well. Adding it during autogeneration is the 1st step

PR #6093, opened 2024-11-15
fix: bookmark query by doing a case insensitive query

AdityaHegde:
Our top navigation bar uses lower case resource name so if a bookmark was created with original case and navigated to using top bar the older bookmarks were not found anymore.

Making the bookmark query case insensitive since resource name is already case insensitive in our system.

### DIFF
--- a/admin/database/postgres/migrations/0053.sql
+++ b/admin/database/postgres/migrations/0053.sql
@@ -1,2 +1,2 @@
 DROP INDEX bookmarks_search_idx;
-CREATE INDEX bookmarks_search_idx ON bookmarks (project_id, lower(resource_name), resource_kind);
+CREATE INDEX bookmarks_search_idx ON bookmarks (project_id, resource_kind, lower(resource_name));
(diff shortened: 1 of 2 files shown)


begelundmuller (changes requested):
Thanks! Can you also update the `bookmarks_search_idx` to index on `lower(resource_name)` and consider if other indexes would make sense to add?

AdityaHegde:
Updated `bookmarks_search_idx`. Do you mean other indices or other columns in this index?

For other columns it doesnt make sense, the only one we could possibly do it is `resource_kind` but since we use constants for that it would be overkill.

For other indices, for sure. Lets take as a follow up of this PR since it would need some thorough check.

admin/database/postgres/migrations/0053.sql
```diff
@@ -0,0 +1,2 @@
+DROP INDEX bookmarks_search_idx;
+CREATE INDEX bookmarks_search_idx ON bookmarks (project_id, lower(resource_name), resource_kind);
```
begelundmuller, line 2:
```suggestion
CREATE INDEX bookmarks_search_idx ON bookmarks (project_id, resource_kind, lower(resource_name));
```

begelundmuller (approved):

PR #6094, opened 2024-11-18
fix: templated uri dimension by making a request

### LINKED ISSUE #5282: Feat: SQL Expressions as URI support on dimensions

AdityaHegde:
closes #5282

For URI dimensions we need to add a measure to get the final value after templating is done. This PR replaces the current implementation of replacing the template in frontend to backend.

### DIFF
--- a/web-common/src/features/dashboards/dashboard-utils.ts
+++ b/web-common/src/features/dashboards/dashboard-utils.ts
@@ -4,6 +4,7 @@ import {
   ComparisonDeltaPreviousSuffix,
   ComparisonDeltaRelativeSuffix,
 } from "@rilldata/web-common/features/dashboards/filters/measure-filters/measure-filter-entry";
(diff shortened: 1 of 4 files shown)


briangregoryholmes (approved):
It looks like the previous approach is being entirely deprecated, which means the makeHref function needs to change a bit. Perhaps this:

```
function makeHref(uriStringOrBoolean: string | boolean | null) {
    if (!uriStringOrBoolean) {
      return undefined;
    }

    const uri =
    uriStringOrBoolean === true
        ? label
        : uriStringOrBoolean
            .replace(/\s/g, "")

    const hasProtocol = /^[a-zA-Z][a-zA-Z\d+\-.]*:/.test(uri);

    if (!hasProtocol) {
      return "https://" + uri;
    } else {
      return uri;
    }
  }
  ```

PR #6098, opened 2024-11-19
fix: billing UI minor issues

AdityaHegde:
- [x] Hide quotas when plan has ended/expired. Quotas will be set to 0 so no point in showing the quotas module.
- [x] During deploy, back button in upgrade wall should return to org selection page if org count > 1.
- [x] Fix issues with time remaining < 1 hour showing in milliseconds.
- [x] Avoid the quick pop-in of welcome dialog before navigating after upgrading during deploy.
- [x] `canceled` => `cancelled`
- [x] Update input description to match mocks.

### DIFF
--- a/web-admin/src/features/organizations/settings/OrgNameSettings.svelte
+++ b/web-admin/src/features/organizations/settings/OrgNameSettings.svelte
@@ -119,6 +119,7 @@
       description={`Your org URL will be https://ui.rilldata.com/${sanitizeOrgName($form.name)}, to comply with our naming rules.`}
       textClass="text-sm"
       alwaysShowError
+      width="520px"
     />
     <Input
       bind:value={$form.description}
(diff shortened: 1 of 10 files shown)


web-admin/src/features/organizations/settings/OrgNameSettings.svelte
```diff
@@ -119,6 +119,7 @@
       description={`Your org URL will be https://ui.rilldata.com/${sanitizeOrgName($form.name)}, to comply with our naming rules.`}
       textClass="text-sm"
       alwaysShowError
+      width="520px"
```
ericpgreen2, line 122: Looks like this breaks responsiveness. Switch it to `maxWidth`?
image
image

ericpgreen2 (approved):

PR #6102, opened 2024-11-19
fix : remove table directory while holding tx lock

k-anshul:
A concurrent goroutine can start creating table which can make `os.RemoveAll` fail.


begelundmuller (approved):
Good catch!

PR #6103, opened 2024-11-19
Create public url and offer copy to clipboard cta

lovincyrus:
Fix part of https://github.com/rilldata/rill-private-issues/issues/920.

Safari doesn't allow async clipboard copy unless it's directly triggered by a user interaction. Split the CTA into two separate actions: one for creation and another for copying to the clipboard.

### DIFF
--- a/web-admin/src/features/public-urls/CreatePublicURLForm.svelte
+++ b/web-admin/src/features/public-urls/CreatePublicURLForm.svelte
@@ -119,14 +119,12 @@
   );
 
   function onCopy() {
-    const success = copyToClipboard(url, "URL copied to clipboard");
-    if (success) {
-      copied = true;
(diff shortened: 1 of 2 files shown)


ericpgreen2 (approved):
Code looks good. I just find it distracting that we're notified in 3 different places of the copy-to-clipboard action:
1. the button text changes
2. the text above the button changes
3. and there's a toast notification, which disappears at a different speed than 1 & 2.

I'd propose we be consistent with the behavior in the first "Copy URL" tab:
- The button text could be "Copy Public URL" (to specify what's different between this URL and the URL from the first tab)
- The button text should temporarily change to "√ Copied URL" (1)
- We do not change the text above the button (2) nor emit a toast notification (3)

PR #6106, opened 2024-11-20
feat: big num new formatting rules

djbarnwal:
Format big numbers based on https://www.notion.so/rilldata/Formatting-the-display-of-metrics-for-higher-precision-in-tables-and-other-contexts-28371608b0924fd99be243110b12f32c?pvs=4#12fba33c8f5780c18ac5e89fbbf2853e

Ex - 
image


ericpgreen2 (approved):
For a small number with `none` formatting, I see <.0001...
image

... yet for an even smaller number, I see scientific notation:
image

Double-checking – is that per design? My first instinct is that it feels like two different strategies for small numbers, when I'd expect one consistent strategy.

djbarnwal:
I am a bit conflicted on this because something like 10^(-20) is very different that let's say 10^(-5) but both would be then represented as "<0.0001".  The context behind these big numbers is important but I would say it's a safer choice to show exponential formatting for extremely large and extremely small numbers. I will update this Order of magnitude to 10^(-10) for showing exponents for getting this PR out for the release. In the meanwhile we can discuss offline what limit/rules make more sense.

PR #6116, opened 2024-11-20
Organizations: remove old CTA to add users via CLI

ericpgreen2:
This CTA isn't needed now that we have a dedicated Users tab.

Before:
image

After:
image


lovincyrus (changes requested):
nice! it feels like we should retain the spacing above the "check out your projects below." text.

looping in @jkhwu

jkhwu:
I resolved directly with @ericpgreen2 in a sync

lovincyrus (approved):

PR #6119, opened 2024-11-21
feat: support locale formatting using d3 locale

djbarnwal:
Integrate `format_d3_locale` in the UI. This enables support for all currencies, grouping, thousands string, decimal string etc.

Usage details here 

https://d3js.org/d3-format#formatLocale

### DIFF
--- a/web-common/src/features/metrics-views/editor/metrics-schema.json
+++ b/web-common/src/features/metrics-views/editor/metrics-schema.json
@@ -123,6 +123,35 @@
             "type": "string",
             "description": "Controls the formatting of this measure using a d3-format string."
           },
+          "format_d3_locale": {
+            "type": "object",
+            "description": "Defines a custom locale for d3-format.",
+            "properties": {
+              "decimal": { "type": "string", "description": "The decimal point (e.g., \".\")." },
+              "thousands": { "type": "string", "description": "The group separator (e.g., \",\")." },
+              "grouping": {
+                "type": "array",
+                "items": { "type": "integer" },
+                "description": "The array of group sizes (e.g., [3]), cycled as needed."
+              },
+              "currency": {
+                "type": "array",
+                "items": { "type": "string" },
+                "minItems": 2,
+                "maxItems": 2,
+                "description": "The currency prefix and suffix (e.g., [\"$\", \"\"])."
+              },
+              "numerals": {
+                "type": "array",
+                "items": { "type": "string" },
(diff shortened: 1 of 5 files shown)


web-common/src/features/metrics-views/editor/metrics-schema.json
```diff
@@ -123,6 +123,35 @@
...
+              "numerals": {
+                "type": "array",
+                "items": { "type": "string" },
+                "description": "Optional array of ten strings to replace the numerals 0-9."
+              },
+              "percent": { "type": "string", "description": "Optional symbol to replace the percent suffix." },
+              "minus": { "type": "string", "description": "Optional minus sign (defaults to \"−\")." },
+              "nan": { "type": "string", "description": "Optional value for not-a-number (defaults to \"NaN\")." }
+            },
+            "required": ["decimal", "thousands", "grouping", "currency"]
```
ericpgreen2, line 153: I don't think these should all be required. Shouldn't we be able to provide a subset?
image

web-common/src/lib/number-formatting/format-measure-value.ts
```diff
@@ -1,5 +1,14 @@
+import {
+  currencyHumanizer,
+  includesCurrencySymbol,
+  isValidD3Locale,
+} from "@rilldata/web-common/lib/number-formatting/utils/d3-format-utils";
 import type { MetricsViewSpecMeasureV2 } from "@rilldata/web-common/runtime-client";
-import { format as d3format } from "d3-format";
+import {
+  format as d3format,
+  formatLocale,
```
ericpgreen2, line 9: Nit: `formatLocale as d3FormatLocale` for symmetry with the line above?

ericpgreen2 (approved):
Sweet! I'll approve, so you can merge when ready.

UXQA: with the following spec...
image
... I see the Big Number uses my provided currency, but the leaderboards do not.
image

web-common/src/features/metrics-views/editor/metrics-schema.json
djbarnwal: Yeah, you are getting this mismatch between values because of the incomplete spec. Following https://d3js.org/d3-format#formatLocale `decimal`, `thousands`, and `grouping` are required fields. I can add a subset but then our definition and the d3 docs would diverge and we would have to maintain our own copy of the docs for it. I am fine either way though.

djbarnwal: Made the change!

PR #6121, opened 2024-11-21
feat : Add validation for env names

k-anshul:
closes https://github.com/rilldata/rill/issues/6112

### DIFF
--- a/runtime/pkg/env/env.go
+++ b/runtime/pkg/env/env.go
@@ -1,0 +1,15 @@
+package env
+
+import (
+	"fmt"
+	"regexp"
+)
+
+var re = regexp.MustCompile(`^[a-zA-Z_][a-zA-Z0-9_.]*$`)
+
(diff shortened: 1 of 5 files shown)


runtime/pkg/env/env.go
```diff
@@ -0,0 +1,15 @@
...
+import (
+	"fmt"
+	"regexp"
+)
+
+var re = regexp.MustCompile(`^[a-zA-Z_][a-zA-Z0-9_.]*$`)
+
+func ValidateName(name string) error {
+	if !re.MatchString(name) {
+		return fmt.Errorf("invalid variable name: %q. Must start with a letter or underscore, can contain letters, digits, underscores and dots", name)
```
begelundmuller, line 12:
```suggestion
		return fmt.Errorf("invalid variable name %q: must start with a letter or underscore and contain only letters, digits, underscores and dots", name)
```

PR #6132, opened 2024-11-22
fix: billing email  and deploy paywall copy

AdityaHegde:
- [x] Update various email related copy https://www.notion.so/rilldata/Email-adjustments-146ba33c8f57808e9eddc1a3828d6bdc
- [x] Update the paywall during deploy to be aware of trial plan.
- [x] Update the percent calculation on trial orgs usage.

### DIFF
--- a/runtime/pkg/email/templates/welcome_to_trial.mjml
+++ b/runtime/pkg/email/templates/welcome_to_trial.mjml
@@ -15,7 +15,7 @@
   <mj-body>
     <mj-section padding-top="24px">
       <mj-column>
-        <mj-image container-background-color="#FFF" width="48px" src="https://cdn.rilldata.com/email-transactional/rill-logo-purple-square" />
+        <mj-image container-background-color="#FFF" width="48px" src="https://cdn.rilldata.com/email-transactional/rill-logo-purple-square.png" />
 
         <mj-text font-size="20px">
           Welcome to Rill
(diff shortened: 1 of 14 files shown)


runtime/pkg/email/templates/welcome_to_trial.mjml
```diff
@@ -15,7 +15,7 @@
   <mj-body>
     <mj-section padding-top="24px">
       <mj-column>
-        <mj-image container-background-color="#FFF" width="48px" src="https://cdn.rilldata.com/email-transactional/rill-logo.png" />
+        <mj-image container-background-color="#FFF" width="48px" src="https://cdn.rilldata.com/email-transactional/rill-logo-purple-square" />
```
ericpgreen2, line 18: Shouldn't this asset have a file extension?

web-common/src/features/project/deploy-errors.ts
```diff
@@ -41,7 +41,10 @@ export type DeployError = {
   message: string;
 };
 
-export function extractDeployError(error: ConnectError): DeployError {
+export function extractDeployError(
+  error: ConnectError,
+  orgOnTrial: boolean,
```
ericpgreen2, line 46: This feels like a random parameter to add to this function – because it doesn't relate to "extracting" information from the error message. Maybe it'd make sense with a small name change: `makeDeployError`, `makePrettyDeployError`?

ericpgreen2 (approved):
Code looks good. Note that I did not QA the product flow – assuming you have.

PR #6134, opened 2024-11-22
Managed connectors

begelundmuller:
Closes https://github.com/rilldata/rill-private-issues/issues/853


admin/provision.go
```diff
@@ -0,0 +1,143 @@
...
+type ProvisionOptions struct {
+	DeploymentID string
+	Type         provisioner.ResourceType
+	Name         string
+	Provisioner  string
+	Args         map[string]any
+	Annotations  map[string]string
+}
+
+func (s *Service) Provision(ctx context.Context, opts *ProvisionOptions) (*database.ProvisionerResource, error) {
```
pjain1, line 23: In general if provisioning fails then deployment should be deleted ?

runtime/queries/column_numeric_histogram.go
```diff
@@ -149,6 +149,11 @@ func (q *ColumnNumericHistogram) calculateFDMethod(ctx context.Context, rt *runt
 		return fmt.Errorf("not available for dialect %q", olap.Dialect())
 	}
 
+	if olap.Dialect() == drivers.DialectClickHouse {
+		// Returning early with empty results because this query tends to hang on ClickHouse.
```
pjain1, line 153: just confirming - this is intentional ?

admin/deployments.go
```diff
@@ -478,107 +477,42 @@ type provisionRuntimeOptions struct {
...
-	if err != nil {
-		return nil, err
-	}
-
 	return pr, nil
 }
 
 func (s *Service) findProvisionedRuntimeResource(ctx context.Context, deploymentID string) (*database.ProvisionerResource, bool, error) {
-	prs, err := s.DB.FindProvisionerResourcesForDeployment(ctx, deploymentID)
+	pr, err := s.DB.FindProvisionerResourceByName(ctx, deploymentID, string(provisioner.ResourceTypeRuntime), "")
```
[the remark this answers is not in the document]
begelundmuller: Okay, I changed the query name to `...ByTypeAndName`. 

The runtime resource is not currently named because there is only supposed to be one resource of that type for each deployment. This is different than for ClickHouse, where it matches to the connector name in the runtime and you could in theory have multiple ClickHouses for one deployment (e.g. if you define `clickhouse-a.yaml` and `clickhouse-b.yaml`).

admin/provision.go
begelundmuller: Not anymore – if provisioning fails, the deployment will still be in Postgres with an error status. Provisioning can then be retried by doing a reset (the reset will delete the old deployment record in Postgres, so no leaks).

This enables moving provisioning to a retryable background job, where the UI/CLI can poll the deployment until its status becomes "ok". Kasper is planning to do this move when he gets some time.

runtime/queries/column_numeric_histogram.go
begelundmuller: Yeah, should probably have been a separate PR, but we want to keep that.

pjain1 (approved):

PR #6138, opened 2024-11-22
feat: workspace breadcrumbs

briangregoryholmes:
Screenshot 2024-11-22 at 1 34 46 PM


ericpgreen2 (approved):
This is great! It feels like these breadcrumbs might become the primary mode of navigation. Some UXQA:
1. It's unclear to me when the downstream breadcrumbs are present. In [this Jam](https://jam.dev/c/75445944-b659-4770-b222-95fa44d87ed7), it looks like only the one-step-away downstream breadcrumb is shown. It'd be nice to show all downstream breadcrumbs.
2. Could you add an e2e test for navigating through these breadcrumbs?
3. I had a source & a metrics view, then I created a model from the source. The model's breadcrumb was named by the metrics view "Adbids_metrics", not by the model  "Adbids_model":
    image

Approving so you can merge at your discretion.

PR #6149, opened 2024-11-25
docs / metrics view locale_d3 example

royendo:
Not sure if necessary but based on Dhiraj showcase, might need some examples, we link the website but would be nice to able to reference one example via rill docs 

```
    format_d3: "$,"
    format_d3_locale: 
      grouping: [3, 2]
      currency: ["₹", ""]
```


nishantmonu51 (approved):
👍

Issue #6150, opened 2024-11-25
Allow variable creation with different target environments


lovincyrus:
- There can only be 1 unique `key_name` + `environment` for a project’s environment variables
    - Users can create `key_names` with same name but different `environments`
        - e.g. key_name = AWS_SECRET_KEY; environment = development & key_name = AWS_SECRET_KEY; environment = production

PR #6168, opened 2024-11-25
Allow variable creation with different target environments 

lovincyrus:
closes https://github.com/rilldata/rill/issues/6150

https://www.notion.so/rilldata/Environment-Variables-management-UI-42304f145b07410d86c2a523496aa6bc

- [x] allow 1 unique variable key + environment
- [x] disallow updating existing variable in create dialog perhaps
- [x] https://rilldata.slack.com/archives/C07NUPQ02EA/p1733167795245509
- [x] A variable with the name `test` already exists for the target development

CleanShot 2024-11-27 at 09 47 49@2x

https://github.com/user-attachments/assets/3769feac-b5d4-43b4-9046-95b8b6a9c049

### DIFF
--- a/web-admin/src/features/projects/environment-variables/utils.ts
+++ b/web-admin/src/features/projects/environment-variables/utils.ts
@@ -1,0 +1,39 @@
+import { EnvironmentType, type VariableNames } from "./types";
+
+// FIXME: we should rename UNDEFINED to ALL for more clarity
+// FOR NOW, we keep UNDEFINED based on the API
+export function getEnvironmentLabel(environment: string) {
+  return environment === EnvironmentType.UNDEFINED ? "all" : environment;
+}
+
+export function isDuplicateKey(
+  environment: string,
+  variableName: string,
+  variableNames: VariableNames,
+) {
+  return variableNames.some((existingVariable) => {
+    return (
+      existingVariable.environment === environment &&
+      existingVariable.name === variableName
+    );
+  });
+}
+
+export function getCurrentEnvironment(
+  isDevelopment: boolean,
+  isProduction: boolean,
+) {
+  if (isDevelopment && isProduction) {
+    return EnvironmentType.UNDEFINED;
+  }
+
+  if (isDevelopment) {
+    return EnvironmentType.DEVELOPMENT;
+  }
+
+  if (isProduction) {
+    return EnvironmentType.PRODUCTION;
+  }
+
+  return EnvironmentType.UNDEFINED;
+}
--- a/web-admin/src/features/projects/environment-variables/utils.ts
+++ b/web-admin/src/features/projects/environment-variables/utils.ts
@@ -5,7 +5,7 @@ import { EnvironmentType, type VariableNames } from "./types";
  * - Environment it belongs to ("prod" or "dev").
  * - If empty, then the value is used as the fallback for all environments.
  */
-export function getEnvironmentLabel(environment: string) {
+export function getEnvironmentType(environment: string) {
   return environment === EnvironmentType.UNDEFINED ? "" : environment;
 }
 
@@ -13,7 +13,7 @@ export function getEnvironmentLabel(environment: string) {
  * Checks if a key would create a duplicate environment variable.
  *
  * Rules for environment variables:
- * 1. A key can only be tied to one environment configuration
+ * 1. A key can be tied to multiple environment configurations, e.g. my_key[dev] & my_key[prod]
  * 2. If a key exists in all environments (UNDEFINED), it cannot be created in dev or prod
  * 3. If a key exists in dev or prod, it cannot be created in all environments
  * 4. If a key exists in dev, it cannot be created in dev but CAN be created in prod
--- a/web-admin/src/features/projects/environment-variables/utils.ts
+++ b/web-admin/src/features/projects/environment-variables/utils.ts
@@ -1,21 +1,59 @@
 import { EnvironmentType, type VariableNames } from "./types";
 
-// FIXME: we should rename UNDEFINED to ALL for more clarity
-// FOR NOW, we keep UNDEFINED based on the API
+/**
+ *
+ * - Environment it belongs to ("prod" or "dev").
+ * - If empty, then the value is used as the fallback for all environments.
+ */
 export function getEnvironmentLabel(environment: string) {
-  return environment === EnvironmentType.UNDEFINED ? "all" : environment;
+  return environment === EnvironmentType.UNDEFINED ? "" : environment;
 }
 
+/**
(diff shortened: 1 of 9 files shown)


ericpgreen2:
It should probably be a server-side check, but it shouldn't be possible to have multiple `test` keys for a given environment. If `test[dev,prod]=abc` and `test[prod]=def`, what is the value of `test` in production? I see that Vercel prevents such a submission:
image

lovincyrus:
> It should probably be a server-side check, but it shouldn't be possible to have multiple `test` keys for a given environment. If `test[dev,prod]=abc` and `test[prod]=def`, what is the value of `test` in production? I see that Vercel prevents such a submission: image

Agreed - This should ideally be a server-side check.

begelundmuller:
> It should probably be a server-side check, but it shouldn't be possible to have multiple `test` keys for a given environment

This already is not possible at the DB level, so you may have a UI state issue if that seems to be the case. If you see this, can you try refreshing the page?

Also note that the `UpdateProjectVariables` API is an *update* API, so it's not going to give you errors like the one Vercel returns – if `test[dev,prod]=abc` and you update `test[prod]=def`, then the new state is `test[dev]=abc` and `test[prod]=def`.

-------

It sounds like it may be helpful with some context about how variables are stored in the database. This is the schema:
```sql
CREATE TABLE project_variables (
  id UUID NOT NULL DEFAULT uuid_generate_v4() PRIMARY KEY,
  project_id UUID NOT NULL REFERENCES projects(id) ON DELETE CASCADE,
  name TEXT NOT NULL,
  value TEXT NOT NULL,
  -- Encryption key ID if the value is encrypted
  value_encryption_key_id TEXT NOT NULL DEFAULT '',
  -- Environment it belongs to ("prod" or "dev").
  -- If empty, then it should be used as the fallback for all environments.
  environment TEXT NOT NULL DEFAULT '',
  -- The user who most recently edited the variable
  updated_by_user_id UUID REFERENCES users(id) ON DELETE SET NULL,
  created_on TIMESTAMPTZ DEFAULT now() NOT NULL,
  updated_on TIMESTAMPTZ DEFAULT now() NOT NULL
);

CREATE UNIQUE INDEX project_variables_project_id_environment_name_idx ON project_variables (project_id, environment, lower(name));
```

Note that there's a unique index on `project_id, environment, lower(name)`. This unique index is used to determine if a variable should be added or updated in calls to `UpdateProjectVariables` in the following upsert statement:
```sql
INSERT INTO project_variables (project_id, environment, name, value, value_encryption_key_id, updated_by_user_id, updated_on)
VALUES ...
ON CONFLICT (project_id, environment, lower(name)) DO UPDATE SET
  value = EXCLUDED.value,
  value_encryption_key_id = EXCLUDED.value_encryption_key_id,
  updated_by_user_id = EXCLUDED.updated_by_user_id,
  updated_on = now()
RETURNING *
```

lovincyrus:
Latest demo:

https://github.com/user-attachments/assets/ebf9c668-cf16-4289-9bf9-e60858e5cd25

lovincyrus:
> > It should probably be a server-side check, but it shouldn't be possible to have multiple `test` keys for a given environment
> 
> This already is not possible at the DB level, so you may have a UI state issue if that seems to be the case. If you see this, can you try refreshing the page?
> 
> Also note that the `UpdateProjectVariables` API is an _update_ API, so it's not going to give you errors like the one Vercel returns – if `test[dev,prod]=abc` and you update `test[prod]=def`, then the new state is `test[dev]=abc` and `test[prod]=def`.
> 
> It sounds like it may be helpful with some context about how variables are stored in the database. This is the schema:
> 
> ```sql
> CREATE TABLE project_variables (
>   id UUID NOT NULL DEFAULT uuid_generate_v4() PRIMARY KEY,
>   project_id UUID NOT NULL REFERENCES projects(id) ON DELETE CASCADE,
>   name TEXT NOT NULL,
>   value TEXT NOT NULL,
>   -- Encryption key ID if the value is encrypted
>   value_encryption_key_id TEXT NOT NULL DEFAULT '',
>   -- Environment it belongs to ("prod" or "dev").
>   -- If empty, then the value is used as the fallback for all environments.
>   environment TEXT NOT NULL DEFAULT '',
>   -- The user who most recently edited the variable
>   updated_by_user_id UUID REFERENCES users(id) ON DELETE SET NULL,
>   created_on TIMESTAMPTZ DEFAULT now() NOT NULL,
>   updated_on TIMESTAMPTZ DEFAULT now() NOT NULL
> );
> 
> CREATE UNIQUE INDEX project_variables_project_id_environment_name_idx ON project_variables (project_id, environment, lower(name));
> ```
> 
> Note that there's a unique index on `project_id, environment, lower(name)`. This unique index is used to determine if a variable should be added or updated in calls to `UpdateProjectVariables` in the following upsert statement:
> 
> ```sql
> INSERT INTO project_variables (project_id, environment, name, value, value_encryption_key_id, updated_by_user_id, updated_on)
> VALUES ...
> ON CONFLICT (project_id, environment, lower(name)) DO UPDATE SET
>   value = EXCLUDED.value,
>   value_encryption_key_id = EXCLUDED.value_encryption_key_id,
>   updated_by_user_id = EXCLUDED.updated_by_user_id,
>   updated_on = now()
> RETURNING *
> ```
> 
> When resolving variables for the runtime, it first builds a list of all variables with `environment = ''` (the fallback values for all environments), and then overwrites it with all variables with `environment = '<target env>'` (the override values for the specific environment).

Added more checks on the client side to check for duplicates before hitting the actual _Update_ API operation call.

web-admin/src/features/projects/environment-variables/utils.ts
```diff
@@ -0,0 +1,77 @@
...
+      // If trying to create in the same environment as existing variable
+      if (variable.environment === environment) {
+        return true;
+      }
+    }
+    return false;
+  });
+}
+
+export function getCurrentEnvironment(
```
ericpgreen2, line 60: Nit: `getEnvironmentType`?

web-admin/src/features/projects/environment-variables/utils.ts
```diff
@@ -0,0 +1,77 @@
...
+ */
+export function getEnvironmentLabel(environment: string) {
+  return environment === EnvironmentType.UNDEFINED ? "" : environment;
+}
+
+/**
+ * Checks if a key would create a duplicate environment variable.
+ *
+ * Rules for environment variables:
+ * 1. A key can only be tied to one environment configuration
```
ericpgreen2, line 16: Nit: a key *can* be tied to multiple environment configurations, e.g. `my_key[dev]` & `my_key[prod]`

ericpgreen2 (approved):
LGTM!

PR #6170, opened 2024-11-26
feat : data bucket for persisting data to gcs

k-anshul:
PrefixedBucket closes underlying bucket so no instance will be able to read other instance's data.


runtime/drivers/drivers.go
```diff
@@ -32,16 +33,20 @@ func Register(name string, driver Driver) {
 	Drivers[name] = driver
 }
 
+type OpenDataBucketFn func(ctx context.Context, instanceID string, elem ...string) (*blob.Bucket, error)
+
+var OpenNilDataBucket OpenDataBucketFn = func(ctx context.Context, instanceID string, elem ...string) (*blob.Bucket, error) { return nil, nil }
```
begelundmuller, line 38: It would be nice to avoid returning a `nil` value and `nil` error. The usual expectation is that after checking `if err != nil`, you are free to access a returned value. There are exceptions, but they often require extra caution.

Here, it can easily be fixed by either:
1. Returning `bucket, ok, err := ...`
2. Returning a constant `drivers.ErrBucketNotAvailable` that the caller can check for

begelundmuller (commented):
It feels weird that the data directory and data bucket are configured differently. What do you think about grouping storage/persistence related functionality in a single place and dependency-injecting it similar to the logger and activity client?

For example, we could have a `runtime/storage` package:
```go
// runtime/storage/storage.go
type Client struct {...}

func New(dataDir, tmpDir string, bucketCfg map[string]any) *Client {...}

func NewTmp() *Client {...}

func (c *Client) WithPrefix(instanceID string) *Client {...}

func (c *Client) DataDir(elem ..string) string {...}

func (c *Client) TempDir(elem ..string) string {...}

func (c *Client) OpenBucket() (*blob.Bucket, bool, error) {...}
```

Which we create in `start.go` and pass to the runtime, like `runtime.New(ctx, opts, logger, ..., storage)`. In tests, we can have `storage.NewTmp()` similar to `zap.NewNop()`.

Then when opening a driver, it can pass `r.storage.WithPrefix(instanceID)` to it. 

The advantages I see to this are:
1. All disk persistence is managed in one place, making it simpler to understand and easier to track storage for billing and periodic cleanup
2. In the future, enables exposing buckets in other ways than `*blob.Bucket`. This might be useful for e.g. caching exports where writing directly from the OLAP might be faster.

k-anshul:
> It would be great if we could configure the data directory and data bucket together (instead of passing one as config vars and another as a constructor function). What do you think about grouping storage/persistence related functionality in a single place and dependency-injecting it similar to the logger and activity client?

Yeah sounds good. 
I did the changes as per suggested design. Two changes/considerations though:
1. Even for tests we will need `data_dir` so better to let callers use same constructor and pass a temporary directory which can later be cleaned.
2. As of now it is not possible to set connector `name` so all connectors of same driver will use same name but this can be easily added by making some changes in `AcquireConnection`. I haven't done it yet, will make the changes as part of a different PR. As of now there is no impact because of this.

runtime/connection_cache.go
```diff
@@ -110,7 +110,7 @@ func (r *Runtime) openAndMigrate(ctx context.Context, cfg cachedConnectionConfig
 		}
 	}
 
-	handle, err := drivers.Open(cfg.driver, cfg.instanceID, cfg.config, activityClient, logger)
+	handle, err := drivers.Open(cfg.driver, cfg.instanceID, cfg.config, r.storage.WithPrefix(cfg.instanceID), activityClient, logger)
```
begelundmuller, line 113: I think we should include the connector name here (e.g. `WithPrefix(cfg.instanceID, cfg.name)`) to correctly namespace the connectors. This would mirror the old behavior which also includes the `name` in the prefix.
```
	res.setPreset("data_dir", r.DataDir(instanceID, name), true)
```

It would need to track the `name` in `cachedConnectionConfig`. This change is already present in this open PR, you can lift it from there and then I'll fix the merge conflicts: https://github.com/rilldata/rill/pull/6134

runtime/drivers/duckdb/duckdb.go
```diff
@@ -145,7 +146,7 @@ func (d Driver) Open(instanceID string, cfgMap map[string]any, ac *activity.Clie
 		logger.Warn("failed to install embedded DuckDB extensions, let DuckDB download them", zap.Error(err))
 	}
 
-	cfg, err := newConfig(cfgMap)
+	cfg, err := newConfig(cfgMap, st.DataDir("duckdb"))
```
begelundmuller, line 149: See above comment – then you can avoids, which has the problem of not supporting multiple connectors that use the same driver.

runtime/storage/storage.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+	paths = append(paths, "tmp")
+	paths = append(paths, elem...)
+	return filepath.Join(paths...)
+}
+
+func (c *Client) OpenBucket(ctx context.Context, elem ...string) (*blob.Bucket, bool, error) {
+	if c.bucketConfig == nil {
+		return nil, false, nil
+	}
```
begelundmuller, line 80: To be on the safe side:

```suggestion
	if len(c.bucketConfig) == 0 {
		return nil, false, nil
	}
```

runtime/storage/storage.go
```diff
@@ -0,0 +1,151 @@
...
+	if err != nil {
+		return nil, err
+	}
+	// the token source returned from credentials works for all kind of credentials like serviceAccountKey, credentialsKey etc.
+	return gcp.NewHTTPClient(gcp.DefaultTransport(), gcp.CredentialsTokenSource(creds))
+}
+
+type gcsBucketConfig struct {
+	Bucket     string `mapstructure:"bucket"`
+	SecretJSON string `mapstructure:"google_application_credentials"`
```
begelundmuller, line 150: 1. Use the same struct field name and map key name to keep naming consistent
2. I would suggest suffixing the credentials field with `_json`. The problem is that the behavior in most of Google's native tools is to parse `google_application_credentials` as a file path, so this clarifies that a file path is not acceptable here.

```suggestion
	GoogleApplicationCredentialsJSON string `mapstructure:"google_application_credentials_json"`
```

begelundmuller (changes requested):

runtime/connection_cache.go
k-anshul: Yeah I planned to do this as a follow up. I wasn't too sure if `name` should be part of cache key or not.
Can port changes from your PR :)

runtime/storage/storage.go
k-anshul: Actually `bucketConfig` is a parsed struct not the input map.

begelundmuller: I see!

runtime/storage/storage.go
```diff
@@ -0,0 +1,151 @@
...
+		return fmt.Errorf("could not remove instance tmp directory: %w", err)
+	}
+	if err := os.Mkdir(tmpDir, os.ModePerm); err != nil && !errors.Is(err, fs.ErrExist) {
+		return err
+	}
+
+	return nil
+}
+
+func RemoveInstance(c *Client, instanceID string) error {
```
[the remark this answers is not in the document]
begelundmuller: > Actually for 1. The temp dir is part of every instance so we can't delete directory in New

I think iterating and deleting is okay (not great). An alternative is generating temp paths in `{root}/tmp/{prefix}` so we can easily clear temp data by deleting `tmp`? 

Or another idea is to get a temp root by calling `os.TempDir` and using `{temp root}/{prefix}` for temp paths? Then the OS will automatically clean it for us on restarts. What do you think?

> Not a strong opinion but it felt simpler to just return paths and let callers take care of creating/deleting directories if any sub directory is required similar to the current API.

Yeah that also sounds good to me, but then I don't think we need `AddInstance` right? I mean if the caller anyways is required to call `MkDirAll` before using a path, then `AddInstance` is not needed. 

But I also think it will be fine just to create the directory (but yes then it needs to return the error), callers can check contents if they really need to know if they are the first or not.

> Yes cleanup of instance data in bucket needs to be added. But do we want in a sync path or via bg job similar to DeleteAssets ?

Ideally a background job, at some point we will move all of `DeleteInstance` to be invoked idempotently by a background job. But until then, it's fine to do it sync :)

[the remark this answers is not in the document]
k-anshul: > An alternative is generating temp paths in {root}/tmp/{prefix} so we can easily clear temp data by deleting tmp?

This will mean that `tmp` data is stored outside of instance directory in`data_dir`. At some point we wanted to store all data including any tmp files in a single directory. 

> Or another idea is to get a temp root by calling os.TempDir and using {temp root}/{prefix} for temp paths? Then the OS will automatically clean it for us on restarts. What do you think?

This also has the problem specified above. We may also want to use mounted disks for storing tmp files like file generated by bigquery connector, exported files etc.

> I mean if the caller anyways is required to call MkDirAll before using a path, then AddInstance is not needed.

I think the idea was that base directory for each instance is always created but if any subdirectory is still required then caller need to create that. I think this is similar to how go's API like `TempDir` and `UserCacheDir` operates ?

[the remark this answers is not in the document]
k-anshul: I think the main problem in current `storage` abstraction is that the concept of `instance` is neither coupled nor decoupled and that is what `AddInstance` and `RemoveInstance` tries to compensate but I can go ahead with suggested changes as well.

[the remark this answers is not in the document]
k-anshul: > But I also think it will be fine just to create the directory (but yes then it needs to return the error), callers can check contents if they really need to know if they are the first or not.

Given the elem is a variable list, another consideration is whether we return error if last element is a clash or any element clashes.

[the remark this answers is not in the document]
begelundmuller: > This will mean that tmp data is stored outside of instance directory indata_dir. At some point we wanted to store all data including any tmp files in a single directory.

The main thing is having some way to monitor how much temp data is being written by an instance, but it doesn't strictly speaking need to be in the same directory as the persistent data (but yes that's how we did it initially). As long as all temp data goes in one directory with sub-directories for each instance, then we will still be able to add stats/monitoring of per-instance temp data in the future.

> This also has the problem specified above. We may also want to use mounted disks for storing tmp files like file generated by bigquery connector, exported files etc.

I think this has been more of a weakness of our previous approach. It would actually be better to use an ephemeral local SSD for temp data than a networked persistent disk. I think this would be a good change to make, but we should give infra a heads up so they can ensure that `os.TempDir` is an ephemeral SSD (it probably  already is) and has decent capacity.

> I think the idea was that base directory for each instance is always created but if any subdirectory is still required then caller need to create that.

We almost never access the `data` directory directly, we are anyway appending the connector name to it, so the directory doesn't exist even now I think?

> I think this is similar to how go's API like TempDir and UserCacheDir operates ?

The docs for `TempDir` say "The directory is neither guaranteed to exist nor have accessible permissions." Looking at the implementation, the same seems to be the case for `UserCacheDir`.

> Given the elem is a variable list, another consideration is whether we return error if last element is a clash or any element clashes.

If you decide to create the directory in these calls, then I think the normal behavior/errors of `MkdirAll` would be expected.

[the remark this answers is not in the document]
k-anshul: Sure. Using `os.TempDir` sounds good in that case.

runtime/drivers/snowflake/sql_store.go
```diff
@@ -86,7 +86,7 @@ func (c *connection) QueryAsFiles(ctx context.Context, props map[string]any) (dr
 		return nil, drivers.ErrNoRows
 	}
 
-	tempDir, err := os.MkdirTemp(c.configProperties.TempDir, "snowflake")
+	tempDir, err := c.storage.RandomTempDir("snowflake")
```
begelundmuller, line 89: Is `"snowflake"` needed here? Thinking the `c.storage` will already have `"snowflake"` as a prefix?

runtime/storage/storage.go
```diff
@@ -0,0 +1,174 @@
...
+	path := c.path(c.dataDirPath, elem...)
+	err := os.MkdirAll(path, os.ModePerm)
+	if err != nil {
+		return "", err
+	}
+	return path, nil
+}
+
+func (c *Client) TempDir(elem ...string) (string, error) {
+	path := c.path(os.TempDir(), elem...)
```
begelundmuller, line 106: I don't believe `os.TempDir` is guaranteed to be unique for this process. So it probably needs to call something like `os.MkdirTemp("", "rill")` in `New` and store that directory in the `Client`. 

This is to avoid collision with other temp files, and to ensure that listing `TempDir` on the root `storage.Client` can be used to capture metrics about temp data usage for all instances.

begelundmuller (changes requested):

runtime/drivers/snowflake/sql_store.go
k-anshul: This is a random directory within parent `instance_id/snowflake` where temporarily download parquet files will be written.
Different sources can be ingesting data in parallel so need a unique directory for each.
Though as of now only one file is generated but the rest of the code is structured to support multiple files and hence ingesting them in a unique directory and later removing the directory seems like a safer option.

k-anshul: We can also name it after the sink table but will need to pass it here.

begelundmuller: Okay makes sense, no need to change

runtime/storage/storage.go
```diff
@@ -0,0 +1,180 @@
...
+		panic(err)
+	}
+	return c
+}
+
+func (c *Client) WithPrefix(prefix ...string) *Client {
+	newClient := &Client{
+		dataDirPath:  c.dataDirPath,
+		bucketConfig: c.bucketConfig,
+	}
```
begelundmuller, line 58: Forgot to handle `tempDirPath` here

k-anshul: OOPS. Thanks :)

begelundmuller (approved):

PR #6171, opened 2024-11-26
use creators id in security rule for reports

pjain1:
We add creator id in the jwt attributes for magic token. Now the security claims can use this to check against the owner of the report, if match then allow. Idea is that magic token should have same privileges as its creator for viewing report.

### DIFF
--- a/admin/server/projects.go
+++ b/admin/server/projects.go
@@ -165,7 +165,7 @@ func (s *Server) GetProject(ctx context.Context, req *adminv1.GetProjectRequest)
 		// All themes
 		condition.WriteString(fmt.Sprintf("'{{.self.kind}}'='%s'", runtime.ResourceKindTheme))
 		// The magic token's resource
-		condition.WriteString(fmt.Sprintf(" OR '{{.self.kind}}'=%s AND '{{lower .self.name}}'=%s", duckdbsql.EscapeStringValue(mdl.ResourceType), duckdbsql.EscapeStringValue(strings.ToLower(mdl.ResourceName))))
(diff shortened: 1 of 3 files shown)


admin/server/projects.go
```diff
@@ -190,11 +190,15 @@ func (s *Server) GetProject(ctx context.Context, req *adminv1.GetProjectRequest)
...
-				condition.WriteString(fmt.Sprintf(" OR '{{.self.kind}}'='%s' AND '{{lower .self.name}}'=%s", runtime.ResourceKindMetricsView, duckdbsql.EscapeStringValue(strings.ToLower(spec.MetricsView))))
+				condition.WriteString(fmt.Sprintf(" OR ('{{.self.kind}}'='%s' AND '{{lower .self.name}}'=%s)", runtime.ResourceKindMetricsView, duckdbsql.EscapeStringValue(strings.ToLower(spec.MetricsView))))
 			}
 		}
 
 		attr = mdl.Attributes
+		// add the magic token's creator id to the attributes useful for checks in security rules
+		if mdl.CreatedByUserID != nil && *mdl.CreatedByUserID != "" {
+			attr["creator_id"] = *mdl.CreatedByUserID
+		}
```
begelundmuller, line 201: Just commenting here to reflect the discussion in Slack: the problem with this approach is that it breaks the report if the creator of it leaves (`CreatedByUserID` is nullable), even though other admins can still manage the report.

Is there a way we can explicitly grant access to the magic token's resource? If you look at the rule added below, it is in fact a deny rule for anything that doesn't match the conditions here – which is slightly different.

PR #6174, opened 2024-11-26
Allow calls to `GetOrganization` if the org has public projects

begelundmuller:


admin/server/organizations.go
```diff
@@ -933,12 +943,17 @@ func organizationToDTO(o *database.Organization) *adminv1.Organization {
 			OutstandingInvites:             int32(o.QuotaOutstandingInvites),
```
AdityaHegde, line 943: Should quotas be privileged as well? It does expose what the plan the user could be on.

AdityaHegde (approved):

admin/server/organizations.go
begelundmuller: We can do that – is the frontend able to handle quotas being empty for users who are not admins?

begelundmuller: Going to merge this as is, can do this in a follow up if it's worth it

Issue #6154, opened 2024-11-25
Project Status: Ability to refresh individual sources/models


ericpgreen2:
[Requested in Slack](https://rilldata.slack.com/archives/C02T907FEUB/p1732549697046109)

Follows https://github.com/rilldata/rill/pull/6080

ericpgreen2:
I think we have a couple options. Either:
1. Rows be toggleable/selectable, so the refresh request can be batched. If any are selected, then the "Refresh all sources & models" button transforms into a primary-colored "Refresh all selected resources".
2. Every individual row has its own refresh button

@jkhwu @ericokuma @mindspank, any thoughts?

lovincyrus:
Fixed in https://github.com/rilldata/rill/pull/6175

PR #6175, opened 2024-11-26
Refresh button for source rows in Status page

lovincyrus:
address https://github.com/rilldata/rill/issues/6154

- [x] invalidate resources data when refresh source is clicked
- [x] unable to invalidate individual source resource

CleanShot 2024-11-26 at 13 44 08@2x


ericpgreen2:
Note: we should also include the refresh button for Models because 1) a Model can be a root node of the DAG, 2) an intermediate model could still have logic that needs a refresh independent from an upstream node, and 3) we're soon deprecating the Source concept anyways.

lovincyrus:
> Note: we should also include the refresh button for Models because 1) a Model can be a root node of the DAG, 2) an intermediate model could still have logic that needs a refresh independent from an upstream node, and 3) we're soon deprecating the Source concept anyways.
> 
> And actually, if it's not too much a scope expansion, it'd be nice to be able to refresh Alerts & Reports too. See: image

Added row click refresh button to `Source` and `Model` resource kind. I also added a tooltip to indicate that a resource trigger will update all dependent resources.

---

Latest demo:

https://github.com/user-attachments/assets/393afa07-1076-4792-9d3b-ab99274609ec

ericpgreen2 (changes requested):
There's a lot of action when clicking the refresh button: flickering, jumpiness, table resizing. Some things I noticed:
1. The Resources table unmounts temporarily. [See Jam](https://jam.dev/c/941ace36-aa0c-4e77-afc4-27c89cf3ecd1) and screenshot:
    image
2. The page gets auto-scrolled to the top ([Jam](https://jam.dev/c/941ace36-aa0c-4e77-afc4-27c89cf3ecd1)), probably due to point 1.
3. A `RefreshTrigger` resource is shown temporarily. Can we hide it from the list? Here's a screenshot from your video:
    image

lovincyrus:
Latest demo:

https://github.com/user-attachments/assets/209550f4-9cbd-47f1-b27c-6c361779a766

scripts/tsc-with-whitelist.sh
```diff
@@ -63,6 +63,7 @@ web-common/src/runtime-client/http-request-queue/HttpRequestQueue.ts: error TS25
 web-common/src/runtime-client/http-request-queue/HttpRequestQueueTypes.ts: error TS18048
 web-common/src/runtime-client/http-request-queue/HttpRequestQueueTypes.ts: error TS2322
 web-common/src/lib/time/config.ts: error TS2322
+web-common/src/features/file-explorer/new-files.ts: error TS2741
```
ericpgreen2, line 66: Is this a challenging type error? If easy, it'd be good to fix rather than expand this list.

ericpgreen2 (commented):
Mostly looks good, but I ran into a weird state: ~50% of the time when refreshing a Source with an error, the source is marked "Idle", yet for about 30s the "Refresh all sources and models" button is disabled and I see polling `ListResources` network requests. [Here's a Jam](https://jam.dev/c/6ca87666-2f41-412e-adb2-5996017189d9).

I also hit this 500 error, but I think it's been fixed on `main`, right? So it'd help if you could merge `main` into this branch.
image

scripts/tsc-with-whitelist.sh
lovincyrus: I'm not sure if I made any changes that caused this in this PR

lovincyrus:
> Mostly looks good, but I ran into a weird state: ~50% of the time when refreshing a Source with an error, the source is marked "Idle", yet for about 30s the "Refresh all sources and models" button is disabled and I see polling `ListResources` network requests. [Here's a Jam](https://jam.dev/c/6ca87666-2f41-412e-adb2-5996017189d9).
> 
> I also hit this 500 error, but I think it's been fixed on `main`, right? So it'd help if you could merge `main` into this branch. image

I can see from the Jam that the specific resource didn't get refreshed properly due to the "failed to ingest source" error. Here's a video of a happy path:

https://github.com/user-attachments/assets/5a4b5abd-1b97-4635-b1ef-297a2eb383ab

Is it expected to disable the refresh icon when a resource errors out? I can see in the Jam, the resource continues to error out even after a refresh..

scripts/tsc-with-whitelist.sh
ericpgreen2: I think it's from this addition to the `ResourceKind` enum:

```
RefreshTrigger = "rill.runtime.v1.RefreshTrigger",
```

lovincyrus:
> I also hit this 500 error, but I think it's been fixed on `main`, right? So it'd help if you could merge `main` into this branch.
> image

After a rebase, I still run into the 500 error occasionally when I fire a `$createTrigger` call from the individual rows. This doesn't need to block this PR. Would you have more context? @begelundmuller 

Error message — "grpc: error while marshaling: marshaling rill.runtime.v1.ListResourcesResponse: size mismatch (see https://github.com/golang/protobuf/issues/1609): calculated=229, measured=177"

CleanShot 2024-12-10 at 14 01 21@2x

begelundmuller:
> After a rebase, I still run into the 500 error occasionally when I fire a `$createTrigger` call from the individual rows. This doesn't need to block this PR. Would you have more context? @begelundmuller

Yeah that should be fixed by this PR: https://github.com/rilldata/rill/pull/6233

ericpgreen2 (changes requested):
UXQA: I've triggered a Source refresh, yet I don't see any polling of the `ListResource` API.

https://github.com/user-attachments/assets/28609b30-fe85-4934-80f3-de487f4432ae

web-admin/src/features/projects/status/ProjectResources.svelte
```diff
@@ -10,41 +10,104 @@
...
+              resource.meta.name.kind !== ResourceKind.ProjectParser &&
+              resource.meta.name.kind !== ResourceKind.RefreshTrigger,
           );
         },
         refetchOnMount: true,
-        refetchInterval: isReconciling ? 500 : false,
+        keepPreviousData: true,
+        onError: () => {
+          stopPolling();
+        },
```
ericpgreen2, line 43: 1. `onError` will be deprecated in TanStack Query v5 ([see here](https://tkdodo.eu/blog/breaking-react-querys-api-on-purpose))
2. Rather than the imperative `startPolling()`, `stopPolling()`, have you considered using the built-in declarative `refetchInterval: () => number | false`?

lovincyrus:
> UXQA: I've triggered a Source refresh, yet I don't see any polling of the `ListResource` API.
> 
>  Kapture.2024-12-16.at.14.40.05.mp4

I cannot consistently reproduce this..

https://github.com/user-attachments/assets/d104c6c4-c214-468f-b605-ad44f81a2d55

lovincyrus:
> UXQA: I've triggered a Source refresh, yet I don't see any polling of the `ListResource` API.
> 
>  Kapture.2024-12-16.at.14.40.05.mp4

In this PR, I've used the `ListResources` API to refresh a resource.

Taking another look at this,
image

Specifically, the Refresh a model command.
CleanShot 2024-12-18 at 23 53 52@2x

I see `runtimev1.RefreshModelTrigger` and `runtimev1.CreateTriggerRequest` from the `refresh.go`. Any reason why we don't have `RefreshModelTrigger` runtime service? @ericpgreen2

begelundmuller:
@lovincyrus @ericpgreen2 FYI `RefreshModelTrigger` is used in `CreateTriggerRequest.models` ([link](https://github.com/rilldata/rill/blob/221b323aff68072a164c86a1bdb9cfda5c665293/proto/rill/runtime/v1/api.proto#L697)) to enable fine-grained refreshes of models (incremental/full refreshes or refreshes of a specific partition). You can still use specify a model in `CreateTriggerRequest.resources` ([link](https://github.com/rilldata/rill/blob/221b323aff68072a164c86a1bdb9cfda5c665293/proto/rill/runtime/v1/api.proto#L695)), but then it will always be a "normal" model refresh (which is a full refresh for non-incremental models and an incremental refresh for incremental models).

ericokuma:
Screenshot 2025-01-17 at 2 59 46 PM
I'm getting a 400 error when trying to access this PR

ericokuma:
from @jkhwu: After they click refresh we can display a temp toast “🔘 Refreshing <resource name>” Followed by another toast “✔️ Successfully refreshed <resource name>”

ericokuma:
Thanks Cyrus! Can we do the following changes:

1. When a user clicks on refresh, surface a confirmation prompt that reads: "Refresh {resource_name}? Two buttons: `Cancel` and `Yes, refresh`. 

2. When user clicks on `Yes, refresh`, it'll trigger a toast notification that reads "Refreshing {resource_name}

3. When refresh successful, it'll trigger a toast notification that reads "Successfully refreshed {resource_name}"

4. When refresh fail, it'll trigger a toast notification that reads "Refreshing {resource_name} failed" Action: `Try again` which reattempts the refresh

cc @jkhwu, what icons would want in the toast notifications?

lovincyrus:
> cc @jkhwu, what icons would want in the toast notifications?

I'll reuse the existing icons in the codebase.

web-admin/src/features/projects/status/ProjectResources.svelte
lovincyrus: Just realized that my response was stuck in a pending state, sorry!

Update: Okay, I tried out refresh interval but the imperative polling seems to be a more straightforward approach in updating the resource status. With the refresh interval approach, the loading spinner would remain stuck even after the resource had been reconciled. It's easier to stop polling when different conditions have been met.

ericokuma:
@lovincyrus, great work here! Only feedback is for the initial toast notification, can we not show the "x"? In this context, it feels like you can click on the "X" to cancel the refresh

Screenshot 2025-01-27 at 9 38 05 AM

lovincyrus:
1, 2. The reason is that there is no promise-based notification implementation. I used the persisted option as a workaround to keep the notification visible while waiting for the next success or error status. The "x" in the notification indicates that the notification is being dismissed, not the inflight request being canceled.

3. I believe this is expected when we attempt to refresh a resource. We get a new refresh timestamp regardless of the status.

ericokuma:
> 1, 2. The reason is that there is no promise-based notification implementation. I used the persisted option as a workaround to keep the notification visible while waiting for the next success or error status. The "x" in the notification indicates that the notification is being dismissed, not the inflight request being canceled.
> 
> 3. I believe this is expected when we attempt to refresh a resource. We get a new refresh timestamp regardless of the status.

#1,2: Gotcha. The persisted option for when a success toast notification (without the "x") is displayed might not be the right choice. We should remove the notification after a certain amount of time (8 seconds) if it's a toast notification that doesn't have the "x", otherwise, there's no way to dismiss it. for the "refresing resource_name" notification, perhaps we remove the "x" in that toast then and when it either fails or succeeds, we auto-dismiss it

#3: might be misleading for it to say it has been refreshed when the refresh attempt failed

lovincyrus:
> #3: might be misleading for it to say it has been refreshed when the refresh attempt failed

This requires platform support to not update the timestamp when a trigger attempt fails.

ericokuma:
> > #3: might be misleading for it to say it has been refreshed when the refresh attempt failed
> 
> This requires platform support to not update the timestamp when a trigger attempt fails.

Gotcha! fyi @begelundmuller

it's probably not a blocker for release

ericokuma:
https://www.loom.com/share/968406d6fc8a410d983d83ae106bc725?sid=f0ece7e1-5560-4fad-8620-26ee55bc65c4

For the "Refreshing {resource_name}" notification, it should only dismiss once we've received either a success or failure response

ericokuma:
Screenshot 2025-01-27 at 4 09 04 PM

im still seeing that the success toast notification persists.

ericokuma:
but other than that, the PR is good!

ericokuma:
Looks good to me!

ericokuma (approved):

ericpgreen2 (changes requested):
After 30s, I see this "Failed to refresh" message, yet the resources still have a spinner refresh indicator.

Can we not stop polling after 30s? To avoid hitting the runtime with requests twice-a-second for perpetuity, could we use a backoff function that slows polling to once per 10s and continues forever?

Note: we will also want to use this polling logic on the each of the project resources tables (Dashboards/Alerts/Reports).

image

web-admin/src/features/projects/status/ProjectResources.svelte
ericpgreen2: Are you sure about this? This is precisely what a `refetchInterval` function is supposed to be used for. The imperative logic – split over a few functions and a handful of reactive statements – is looking pretty complicated. 

I'm able to break the current approach by: refreshing a source, navigating away from the status page, and navigating back to the status page. I'm observing that the polling has stopped, yet I'd expect that the polling continues.

Ideally (as mentioned below), we can make a `refetchInterval` function that we can re-use for the `ListDashboards`, `ListReports`, and `ListAlerts` requests on those resources' respective pages.

lovincyrus: Referred to https://github.com/rilldata/rill/pull/6421 and it worked

web-admin/src/features/projects/status/ProjectResources.svelte
```diff
@@ -2,54 +2,154 @@
...
+      // Filter out project parser and refresh triggers
+      resources: data?.resources?.filter(
+        (resource: V1Resource) =>
+          resource.meta.name.kind !== ResourceKind.ProjectParser &&
+          resource.meta.name.kind !== ResourceKind.RefreshTrigger,
+      ),
+    }),
+    refetchInterval: (data) => {
+      if (
+        !isPollingEnabled ||
```
ericpgreen2, line 62: We shouldn't need this `isPollingEnabled` variable. The `refetchInterval` function should dictate whether or not there is polling happening. If the conditions do not merit polling (all resources are idle), the function should return `false`.

Though, I could see you wanting to set an `isCurrentlyPolling` variable inside the `refetchInterval` function, so that you could know its polling status outside the hook.

web-admin/src/features/projects/status/ProjectResources.svelte
```diff
@@ -2,54 +2,154 @@
...
+        $resources?.isError ||
```
ericpgreen2, line 63: This looks circular. The `refetchInterval` function should be a pure function that only cares about the `data`.

web-admin/src/features/projects/status/ProjectResources.svelte
```diff
@@ -2,54 +2,154 @@
...
+
+      // Calculate time elapsed since polling started
+      const elapsedTime = Date.now() - pollStartTime;
+
+      // After threshold, gradually increase interval to MAX_POLL_INTERVAL
+      if (elapsedTime > BACKOFF_THRESHOLD_MS) {
+        return MAX_POLL_INTERVAL;
+      }
+
+      return INITIAL_POLL_INTERVAL;
```
ericpgreen2, line 83: Nit: can we do more graceful scaling? Firing 30 requests in 30 seconds seems a little heavy.

web-common/src/lib/event-bus/events.ts
```diff
@@ -4,15 +4,16 @@ interface Link {
 }
 
 export interface NotificationMessage {
-  type?: "default" | "success" | "error";
-  message: string;
+  type?: "default" | "success" | "loading" | "error" | "clear-all";
```
ericpgreen2, line 7: All these types refer to different kinds of toasts. Rather than adding a `clear-all` type, I think it'd be cleaner to add a new `clear-all-notifications` event altogether.

web-common/src/features/file-explorer/new-files.ts
```diff
@@ -116,6 +116,11 @@ export const ResourceKindMap: Record<
     baseName: "alert",
     extension: ".yaml",
   },
+  [ResourceKind.RefreshTrigger]: {
+    folderName: "triggers",
+    baseName: "refresh_trigger",
+    extension: ".yaml",
+  },
```
ericpgreen2, line 123: This list is "user-facing kinds", and I wouldn't consider the refresh trigger user-facing. Instead, you could add `RefreshTrigger` to the exclusion list here:
```
export type UserFacingResourceKinds = Exclude<
  ResourceKind,
  ResourceKind.ProjectParser,
  ResourceKind.RefreshTrigger // Add this
>;
```

ericpgreen2 (changes requested):

web-common/src/features/file-explorer/new-files.ts
lovincyrus: thanks!

web-common/src/lib/event-bus/events.ts
lovincyrus: for sure, it could go one or another, that's why we have a review process here :)

web-admin/src/features/projects/status/selectors.ts
```diff
@@ -18,3 +25,18 @@ export function useProjectDeployment(orgName: string, projName: string) {
...
+    ErrorType<RpcStatus>,
+    V1ListResourcesResponse
+  >,
+) {
+  return createRuntimeServiceListResources(instanceId, undefined, {
+    query: {
+      ...queryOptions,
+    },
+  });
+}
```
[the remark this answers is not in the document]
lovincyrus: this seems like a pattern from the other part of the codebase, https://github.com/rilldata/rill/blob/main/web-common/src/features/explores/selectors.ts#L28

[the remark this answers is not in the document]
lovincyrus: i'll revert this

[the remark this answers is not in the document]
ericpgreen2: Yeah, I think we should avoid it there too :)

web-admin/src/features/projects/status/ProjectResources.svelte
```diff
@@ -2,73 +2,166 @@
...
+        // Filter out project parser and refresh triggers
+        resources: data?.resources?.filter(
+          (resource: V1Resource) =>
+            resource.meta.name.kind !== ResourceKind.ProjectParser &&
+            resource.meta.name.kind !== ResourceKind.RefreshTrigger,
+        ),
+      }),
+      refetchInterval: (data) => {
+        if (
+          $resources?.isError ||
```
ericpgreen2, line 59: This still isn't a pure function, given `$resources.isError` is still here. I'm worried about a circular reference.

web-admin/src/features/projects/status/ActionsCell.svelte
```diff
@@ -0,0 +1,69 @@
...
+            kind: resourceKind,
+            name: resourceName,
+          },
+        ],
+      },
+    });
+
+    // Then update UI state and invalidate query
+    triggerRefresh(resourceName);
+  }
```
ericpgreen2, line 36: It's a bit hard-to-follow that the refresh action has logic in 3 different places: this function, the inline function that starts on line 65, and the `triggerRefresh` callback function defined in `ProjectResources.svelte`. I'd propose unifying the functions in one place.

ericpgreen2 (changes requested):
When the user navigates away from the status page, we'll have to get rid of this toast message, else it'll spin forever.

image

Tangentially, @ericokuma / @jkhwu, I'd propose we remove this toast message altogether, since it's redundant to the spinner in the table. And, in its current form, it's not accurately describing what's happening when: the initial resource has been refreshed successfully, and a downstream resource (or multiple) are refreshing.

image

web-admin/src/features/projects/status/ActionsCell.svelte
lovincyrus: This was needed to pass the resourceName upstream to display all the notifications in ProjectResources. I'll unify the functions.

web-common/src/lib/event-bus/event-bus.ts
```diff
@@ -52,6 +52,7 @@ export const eventBus = new EventBus();
 
 export interface Events {
   notification: NotificationMessage;
+  "clear-all-notifications": void;
```
lovincyrus, line 55: I can remove these notification changes before landing @ericpgreen2

web-admin/src/features/projects/status/ProjectResources.svelte
```diff
@@ -2,54 +2,92 @@
...
+        // Filter out project parser and refresh triggers
+        resources: data?.resources?.filter(
+          (resource: V1Resource) =>
+            resource.meta.name.kind !== ResourceKind.ProjectParser &&
+            resource.meta.name.kind !== ResourceKind.RefreshTrigger,
+        ),
+      }),
+      refetchInterval: (data) =>
+        calculateRefetchInterval(
+          $resources?.data ? currentRefetchInterval : INITIAL_REFETCH_INTERVAL,
```
ericpgreen2, line 75: The `currentRefetchInterval` isn't getting updated, so this is currently polling every 1s w/o a backoff.

PR #6176, opened 2024-11-26
Integrate PostHog session replay

ericpgreen2:
Closes https://github.com/rilldata/rill-private-issues/issues/851


web-common/src/features/project/ProjectDeployer.ts
```diff
@@ -165,7 +172,13 @@ export class ProjectDeployer {
       projectResp.localProjectName,
       checkNextOrg,
     );
-    if (frontendUrl) window.open(frontendUrl + "/-/invite", "_self");
+    const projectUrl = frontendUrl; // https://ui.rilldata.com/<org>/<project>
```
AdityaHegde, line 175: Why not just rename the above to `projectUrl`?

web-common/src/lib/analytics/posthog.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  if (!POSTHOG_API_KEY) {
+    console.warn("PostHog API Key not found");
+    return;
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-member-access
+  posthog.init(POSTHOG_API_KEY, {
+    api_host: "https://us.i.posthog.com", // TODO: use a reverse proxy https://posthog.com/docs/advanced/proxy
+    session_recording: {
```
AdityaHegde, line 17: Is it possible to configure to not upload when on slow network? I wouldnt want an app I use to upload videos when I am using mobile network for example.

ericpgreen2: We could consider disabling it for slow connections, but I'm not sure we should. It's important for us to understand user behavior on all connection types (e.g. users on slow networks might quit in frustration if we don't have proper loading states). The PostHog SDK is designed to be efficient, e.g. it batches & compresses events to minimize bandwidth. That said, if we see significant issues, we could add a network condition check in the future.

web-common/src/features/project/ProjectDeployer.ts
ericpgreen2: Agreed, just changed

AdityaHegde (approved):

PR #6177, opened 2024-11-27
fix: logged out users getting a login page for public projects

AdityaHegde:
After the refactors in Billings UI PR logged out users are getting a login page for public page. There is a 403 request for org which should be ignored. Another 403 for subscription that should not be called at all.


nishantmonu51 (approved):
Lgtm

Issue #6058, opened 2024-11-07
Improve support for `INTERVAL` in metrics APIs


begelundmuller:
- [ ] Cast intervals to milliseconds in our metrics APIs (treating one month as 30 days)
- [ ] Add an INTERVAL type to `StructSchema` and make sure it's set in the metrics response meta for measures of type `INTERVAL`

See [this thread](https://rilldata.slack.com/archives/C01A9DYP013/p1730895589500589?thread_ts=1730755312.798259&cid=C01A9DYP013) for details.

PR #6187, opened 2024-12-02
Support `INTERVAL` type in APIs

begelundmuller:
This PR improves the support for `INTERVAL` types in our APIs by:
1. Adding a type code for intervals
2. Casting intervals to milliseconds with the limitation that one month is counted as 30 days

Interval types are now handled consistently for both DuckDB and ClickHouse. Druid does not currently have an interval type.

Closes https://github.com/rilldata/rill/issues/6058.


runtime/queries/metricsview.go
```diff
@@ -707,6 +707,15 @@ func WriteParquet(meta []*runtimev1.MetricsViewColumn, data []*structpb.Struct,
...
 
 				recordBuilder.Field(idx).(*array.StringBuilder).Append(string(bts))
+			case runtimev1.Type_CODE_INTERVAL:
+				switch v := v.GetKind().(type) {
+				case *structpb.Value_NumberValue:
+					s := fmt.Sprintf("%f", v.NumberValue)
+					recordBuilder.Field(idx).(*array.StringBuilder).Append(s)
+				case *structpb.Value_StringValue:
+					recordBuilder.Field(idx).(*array.StringBuilder).Append(v.StringValue)
+				default:
```
pjain1, line 717: wondering if we should error here ? may its fine to give some data than no data at all

pjain1 (approved):

runtime/queries/metricsview.go
begelundmuller: Yeah I thought it would be better to fail silently here so people get some data

PR #6193, opened 2024-12-03
fix: refetch explore resource

briangregoryholmes:
This shouldn't strictly be necessary, but this is a patch fix for the following bug:

1. Go to the Visual Explore Editing surface
2. Navigate to the upstream metrics view
3. Change a property (like the name of a dimension)
4. Navigate back to the Visual Explore Editing surface

The new name will not be reflected because the query used to create the underlying state management object (`useExploreValidSpec`) does not refetch. I believe this may be a bug related to our specific version of Svelte Query. It looks like we ran into this previously here: https://github.com/rilldata/rill/pull/2027#discussion_r1161672656

### DIFF
--- a/web-common/src/features/entity-management/WatchResourcesClient.ts
+++ b/web-common/src/features/entity-management/WatchResourcesClient.ts
@@ -218,7 +218,7 @@ export class WatchResourcesClient {
             }
 
             queryClient
-              .invalidateQueries(
+              .refetchQueries(
                 getRuntimeServiceGetExploreQueryKey(this.instanceId, {
                   name: res.name.name,
                 }),


ericpgreen2:
This sounds like it's the same "bug" we keep running into due to our Query Client's global `refetchOnMount: false` setting. See [this tech debt issue](https://github.com/rilldata/rill-private-issues/issues/719). 

If that's right, in the short-term, I think a better solution is to set `refetchOnMount: true` in the `useExploreValidSpec()` Query Observer instance. Then, at a later point, we can set `refetchOnMount: true` in the `globalQueryClient`.

ericpgreen2 (approved):

briangregoryholmes:
That would cause the same issue, though. This is one of the specific queries that causes the global spinner to appear.

ericpgreen2:
> This is one of the specific queries that causes the global spinner to appear.

Ah yeah, I see that now.

PR #6199, opened 2024-12-03
Validate measures have a numeric type

begelundmuller:
This PR:
- Moves metrics view validation logic to the `runtime/metricsview` package
- Adds a new validation step that checks that measures return a numeric data type

Closes https://github.com/rilldata/rill/issues/6111


runtime/metricsview/executor_validate.go
```diff
@@ -302,10 +309,41 @@ func validateDimension(ctx context.Context, olap drivers.OLAPStore, t *drivers.T
...
+
+	// Check that the measures are not strings
+	for i, m := range e.metricsView.Measures {
+		typ, ok := types[m.Name]
+		if !ok {
+			return fmt.Errorf("internal: measure %q not found in schema", m.Name)
+		}
+
+		switch typ.Code {
+		case runtimev1.Type_CODE_TIMESTAMP, runtimev1.Type_CODE_DATE, runtimev1.Type_CODE_TIME, runtimev1.Type_CODE_STRING, runtimev1.Type_CODE_BYTES, runtimev1.Type_CODE_ARRAY, runtimev1.Type_CODE_STRUCT, runtimev1.Type_CODE_MAP, runtimev1.Type_CODE_JSON, runtimev1.Type_CODE_UUID:
```
pjain1, line 340: Not sure if could happen but should we also include `CODE_UNSPECIFIED` as well ? Also I think min and max like functions can be used on string columns on duckdb and return string types but I don't see obvious usage for them.

pjain1 (approved):

PR #6216, opened 2024-12-05
token based report

pjain1:
**Changes:**

- This PR changes all the report links to always have a magic token. 
- Magic tokens are created for each run of the report with expiry of 60 days for both internal and external users. 
- This enables the links to never expire and everytime its clicked it just works (if user is removed or added etc. in all scenarios) unless the token expires.

**UI changes needed:**

- Report Creation API - Since even external users can now open the report in explore, the API now expects either the (metrics view and explore name) or canvas name on which report is created. UI also needs to specify a open mode which indicates if open link should be created for the user or not. It has following options
```
open_mode: legacy|none|creator|filter
open_filter: <filter to use if open_mode=filter>

- `legacy` (default): if the email is not a Rill user, don't provide an open link; otherwise, provide an open link without a magic token (so it uses their current credentials)
- `none`: don't provide an open link
- `creator`: provide an open link with a magic token that has the same attributes as the report creator
- `filter` (later): provide an open link with magic token that has the same attributes as the report creator, but filtered by `open_filter`
```
- Report Edit API - should include open_mode in the request now.
- Open link - open link will have magic token and can be used to show the explore (similar to a public URL). If a user is logged in, it should display a banner to say "this is a preview link" or something like that (may need product input here). The report's token may have different permissions/attributes than the current user (e.g. if the report was created by an admin, but the recipient is not an admin), so it cannot simply show the dashboard as the current user.
- Download link - use token to enable download, it already happens for external user now so probably only removing internal download page (if exists).
- Unsubscribe link - It will need to grab the `email` or `slack_user` query param from the url and call the unsubscribe API with this param.

**Important Note:** This does not support locking time ranges as of now (locking dimension filter works), as they would be evaluated during each report run at runtime. The queries that are sent by explore have a separate time range apart from the filter, magic token only supports row filter. If mgc token has time range then it will need to be reconciled with the actual time range sent.

Contributes to https://github.com/orgs/rilldata/projects/38/views/8?pane=issue&itemId=85181742&issue=rilldata%7Crill-private-issues%7C855


admin/server/reports.go
```diff
@@ -57,48 +61,38 @@ func (s *Server) GetReportMeta(ctx context.Context, req *adminv1.GetReportMetaRe
...
+			ExportUrl:      s.admin.URLs.WithCustomDomain(org.CustomDomain).ReportExport(org.Name, proj.Name, req.Report, tokens[recipient]),
+			UnsubscribeUrl: s.admin.URLs.WithCustomDomain(org.CustomDomain).ReportUnsubscribe(org.Name, proj.Name, req.Report, tokens[recipient], recipient),
+		}
+	}
+
+	if req.AnonRecipients {
+		urls[""] = &adminv1.GetReportMetaResponse_URLs{
+			OpenUrl:        s.admin.URLs.WithCustomDomain(org.CustomDomain).ReportOpen(org.Name, proj.Name, req.Report, tokens[""], req.ExecutionTime.AsTime()),
+			ExportUrl:      s.admin.URLs.WithCustomDomain(org.CustomDomain).ReportExport(org.Name, proj.Name, req.Report, tokens[""]),
+			UnsubscribeUrl: s.admin.URLs.WithCustomDomain(org.CustomDomain).ReportUnsubscribe(org.Name, proj.Name, req.Report, tokens[""], ""),
```
begelundmuller, line 90: How is this different from the previous implementation that looped over `tokens`? Since `""` is in `tokens`, could it automatically be added in the same loop?

admin/server/reports.go
```diff
@@ -283,7 +277,23 @@ func (s *Server) UnsubscribeReport(ctx context.Context, req *adminv1.Unsubscribe
...
+			}
+			userEmail = *req.SlackUser
+		} else {
+			userEmail = reportTkn.RecipientEmail
+			if req.Email == nil {
+				return nil, status.Error(codes.InvalidArgument, "no email provided for unsubscribing")
+			}
+			if strings.EqualFold(userEmail, *req.Email) {
+				return nil, status.Error(codes.InvalidArgument, "email does not match token")
+			}
```
begelundmuller, line 295: Should this be a NOT?

```suggestion
			if !strings.EqualFold(userEmail, *req.Email) {
				return nil, status.Error(codes.InvalidArgument, "email does not match token")
			}
```

admin/server/reports.go
```diff
@@ -555,63 +565,46 @@ func (s *Server) generateReportName(ctx context.Context, depl *database.Deployme
...
-	if len(emailSet) == 0 {
-		return emailTokens, nil
-	}
-
+func (s *Server) createMagicTokens(ctx context.Context, orgID, projectID, reportName, ownerID, metricsViewName, exploreName, canvasName string, emails []string) (map[string]string, error) {
 	var createdByUserID *string
 	if ownerID != "" {
 		createdByUserID = &ownerID
 	}
+	ttl := 24 * 30 * 6 * time.Hour // approx 6 months
```
begelundmuller, line 573: Let's start with just 3 months

```suggestion
	ttl := 2 * 30 * 24 * time.Hour // approx 3 months
```

admin/server/reports.go
```diff
@@ -36,6 +35,11 @@ func (s *Server) GetReportMeta(ctx context.Context, req *adminv1.GetReportMetaRe
 		attribute.String("args.report", req.Report),
 		attribute.StringSlice("args.email_recipients", req.EmailRecipients),
 		attribute.String("args.execution_time", req.ExecutionTime.String()),
+		attribute.Bool("args.anon_recipients", req.AnonRecipients),
+		attribute.String("args.owner_id", req.OwnerId),
+		attribute.String("args.metrics_view", req.MetricsView),
+		attribute.String("args.explore", req.Explore),
+		attribute.String("args.canvas", req.Canvas),
```
begelundmuller, line 42: Can we pass an array of `refs` with `runtimev1.ResourceName` instead?

admin/server/reports.go
```diff
@@ -488,6 +489,15 @@ func (s *Server) yamlForManagedReport(opts *adminv1.ReportOptions, ownerUserID s
...
 	res.Annotations.WebOpenState = opts.WebOpenState
+	if opts.MetricsViewName != nil || opts.ExploreName != nil {
+		if opts.MetricsViewName == nil || opts.ExploreName == nil {
+			return nil, errors.New("both metricsViewName and exploreName must be provided")
+		}
+		res.Annotations.MetricsViewName = *opts.MetricsViewName
+		res.Annotations.ExploreName = *opts.ExploreName
+	} else if opts.CanvasName != nil {
+		res.Annotations.CanvasName = *opts.CanvasName
+	}
```
begelundmuller, line 500: 1. Can it just set all of them and leverage `omitempty` in the YAML serialization?
2. Can we just call these `Explore` and `Canvas` (i.e. omit the `Name` suffix). I think it is clear from context that they are names.
3. Can we avoid including the metrics view here?
    - In the UI they just need the explore name
    - In the report reconciler, it can lookup the explore and find its underlying metrics view

proto/rill/admin/v1/api.proto
```diff
@@ -1918,17 +1918,21 @@ message GetReportMetaRequest {
...
+  string canvas = 11;
   reserved 4;
 }
 
 message GetReportMetaResponse {
   message URLs {
     string open_url = 1;
     string export_url = 2;
-    string edit_url = 3;
+    string unsubscribe_url = 3;
```
begelundmuller, line 1932: How will the report owner/admin open the report if they want to edit it? (E.g. change the frequency or the title)

begelundmuller (changes requested):
Also, do we have a job that clears expired magic auth tokens periodically? If not, can you add one?

admin/database/database.go
```diff
@@ -642,22 +642,28 @@ type InsertDeploymentAuthTokenOptions struct {
...
+	State                 string                      `db:"state"`
+	DisplayName           string                      `db:"display_name"`
+	Internal              bool                        `db:"internal"`
+	DependentResources    []MgcTokenDependentResource `db:"dependent_resources"`
+}
+
+type MgcTokenDependentResource struct {
+	Name string
+	Kind string
 }
```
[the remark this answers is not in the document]
pjain1: Ok my thinking was that we can have base `Resource` for which magic token was issued and dependent resources separately wondering if we will ever need to know about this distinction but anyways lets have this for now.

admin/server/reports.go
pjain1: sure

proto/rill/admin/v1/api.proto
pjain1: They will need to visit the reports page on UI. 

We would need to detect if user is a project or org admin or creator of the report and add edit link. Currently we just add edit link for all Rill users but it has an issue if the user is not part of project (or not admin) otherwise it will be broken experience for them. For this reasons I just removed the edit link and thought admin can visit the reports page, let me know if we want to add these checks.

pjain1:
> Also, do we have a job that clears expired magic auth tokens periodically? If not, can you add one?

There is already a `deleteExpiredAuthTokens` job that will take care of it since we set an expiry.

proto/rill/admin/v1/api.proto
begelundmuller: Though not great, this is alright. But would it be possible to keep the edit link for at least the owner user (I'm wondering if that would just require a simple check against the owner user ID)?

admin/server/reports.go
```diff
@@ -283,7 +265,23 @@ func (s *Server) UnsubscribeReport(ctx context.Context, req *adminv1.Unsubscribe
 			return nil, status.Errorf(codes.InvalidArgument, "failed to find report token: %s", err.Error())
 		}
 
-		userEmail = reportTkn.RecipientEmail
+		if reportTkn.RecipientEmail == "" {
```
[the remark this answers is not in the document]
begelundmuller: Is it easy to break it out by notifier type? If not, I think it's fine to unsubscribe both places because this will probably never happen in practice.

begelundmuller:
@pjain1 I responded to the two open questions on this PR. Can you let me know when you address them (also okay if you decide they are not worth it) and fix the merge conflict?

admin/server/reports.go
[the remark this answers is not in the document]
pjain1: done

proto/rill/admin/v1/api.proto
pjain1: Added edit link for owner, also added unsubscribe as a separate link since owner can be removed from project or org but should have option to unsubscribe.

admin/server/reports.go
```diff
@@ -57,48 +57,39 @@ func (s *Server) GetReportMeta(ctx context.Context, req *adminv1.GetReportMetaRe
...
+	var recipients []string
+	recipients = append(recipients, req.EmailRecipients...)
+	if req.AnonRecipients {
+		// add empty email for slack and other notifiers token
+		recipients = append(recipients, "")
 	}
 
-	externalUrls := make(map[string]*adminv1.GetReportMetaResponse_URLs, len(externalEmails))
-	emailTokens, err := s.reconcileMagicTokensForExternalEmails(ctx, proj.OrganizationID, proj.ID, req.Report, req.OwnerId, externalEmails)
+	tokens, ownerEmail, err := s.createMagicTokens(ctx, proj.OrganizationID, proj.ID, req.Report, req.OwnerId, recipients, req.Resources)
```
begelundmuller, line 69: If a filter was applied to the report then that doesn't get propagated to the magic auth token here – so in theory, people can explore more data than that which the report filters by. Is that right? This is probably desirable in many cases, but not necessarily for external partners.

I wonder if we should perhaps add a flag to the report YAML and `CreateReport` API for whether or not to generate an open URL or only an export URL. E.g. it could be named `open_mode: none|full` (and then we could add support for a `filtered` mode later)

admin/server/reports.go
```diff
@@ -488,6 +487,11 @@ func (s *Server) yamlForManagedReport(opts *adminv1.ReportOptions, ownerUserID s
 	res.Annotations.AdminNonce = time.Now().Format(time.RFC3339Nano)
 	res.Annotations.WebOpenPath = opts.WebOpenPath
 	res.Annotations.WebOpenState = opts.WebOpenState
+	if opts.Explore != "" && opts.Canvas != "" {
+		return nil, fmt.Errorf("cannot set both explore and canvas")
+	}
+	res.Annotations.Explore = opts.Explore
+	res.Annotations.Canvas = opts.Canvas
```
begelundmuller, line 494: Should we check somewhere if the report has an explore or canvas? And if not, then skip generating the open link (since it would not actually work).

Some legacy reports and YAML-managed reports may not have an explore or canvas.

admin/server/magic_tokens.go
```diff
@@ -49,13 +47,19 @@ func (s *Server) IssueMagicAuthToken(ctx context.Context, req *adminv1.IssueMagi
 		return nil, status.Error(codes.PermissionDenied, "not allowed to create a magic auth token")
 	}
 
+	resources := make([]database.ResourceName, len(req.Resources))
```
begelundmuller, line 50: It seems like it's missing backwards compatibiltiy for `req.ResourceType` and `req.ResourceName`? (Can be fixed by adding them to `req.Resources` if they are not empty)

begelundmuller (changes requested):

admin/database/postgres/migrations/0065.sql
```diff
@@ -0,0 +1,11 @@
+ALTER TABLE magic_auth_tokens ADD COLUMN resources JSONB NOT NULL DEFAULT '[]'::jsonb;
+
+UPDATE magic_auth_tokens
+SET resources = jsonb_build_array(
+        jsonb_build_object(
+                'type', resource_type,
+                'name', resource_name
+        )
+                );
```
begelundmuller, line 9: nit: Indentation on the last line here

admin/server/reports.go
```diff
@@ -769,13 +792,25 @@ type reportYAML struct {
...
 }
 
+type WebOpenMode string
+
+const (
+	WebOpenModeRecipient WebOpenMode = "recipient"
+	WebOpenModeCreator   WebOpenMode = "creator"
+	WebOpenModeNone      WebOpenMode = "none"
+	WebOpenModeFiltered  WebOpenMode = "filtered"
+)
```
begelundmuller, line 812: This enum is both defined here in Go and defined in the protos. Can we use just one or the other type? IMO using just the Go enum seems simpler to me since we anyway need to store/parse it as a string. You can add a util function here for validating it when the frontend passes it as input:
```go
func (m WebOpenMode) Valid() bool {
  switch m {
  case WebOpenModeRecipient, WebOpenModeCreator, WebOpenModeNone, WebOpenModeFiltered:
    return true
  }
  return false
}
```

admin/server/reports.go
```diff
@@ -787,6 +822,20 @@ func parseReportAnnotations(annotations map[string]string) reportAnnotations {
...
+	case "recipient":
+		res.WebOpenMode = WebOpenModeRecipient
+	case "creator":
+		res.WebOpenMode = WebOpenModeCreator
+	case "none":
+		res.WebOpenMode = WebOpenModeNone
+	case "filtered":
+		res.WebOpenMode = WebOpenModeFiltered
+	case "": // backwards compatibility
+		res.WebOpenMode = WebOpenModeRecipient
```
begelundmuller, line 837: See comment above – since you have the `WebOpenMode` enum type, ideally you can just cast to that, e.g. something like:
```go
res.WebOpenMode := WebOpenMode(annotations["web_open_mode"])
if res.WebOpenMode == "" {
  res.WebOpenMode = WebOpenModeRecipient // Backwards compatibility
}
if !res.WebOpenMode.Valid() {
  return fmt.Errorf("...")
}
```

proto/rill/admin/v1/api.proto
```diff
@@ -1977,6 +1984,9 @@ message GetReportMetaRequest {
   string owner_id = 5;
   google.protobuf.Timestamp execution_time = 6;
   repeated string email_recipients = 7;
+  bool anon_recipients = 8;
+  repeated ResourceName resources = 9;
+  ReportOptions.OpenMode web_open_mode = 10;
```
begelundmuller, line 1989: Try to avoid nested types especially when they are used outside the nested scope itself (we haven't always been consistent with that, but try to going forward). 

So based on the previous, if you decide to go with the Go-defined enum, this can just be a `string` type. Otherwise, I suggest moving this to the outer scope as a `ReportOpenMode` enum or something like that.

proto/rill/admin/v1/api.proto
```diff
@@ -2031,6 +2042,8 @@ message UnsubscribeReportRequest {
   string organization = 1;
   string project = 2;
   string name = 3;
+  optional string email = 4 [(validate.rules).string.email = true];
+  optional string slack_user = 5 [(validate.rules).string.email = true];
```
begelundmuller, line 2046: The practice is to avoid `optional` unless unset and empty value have different meanings, which I don't think is the case here given that the empty string is not a valid email.

If `optional` is just for validation, I believe there's a validation rule you can set that will skip the other validation rules if the value is empty.

proto/rill/admin/v1/api.proto
```diff
@@ -2478,6 +2492,17 @@ message ReportOptions {
...
+  string explore = 16; // either canvas or explore should be set
+  string canvas = 17;
+  enum OpenMode {
+    OPEN_MODE_UNSPECIFIED = 0;
+    OPEN_MODE_RECIPIENT = 1; // send this for older reports
+    OPEN_MODE_CREATOR = 2; // send this for reports that should be opened with creators permissions
+    OPEN_MODE_NONE = 3; // send this for reports that should not be opened or older reports which do not have any web_open_path
+    OPEN_MODE_FILTERED = 4; // send this for reports that should be opened with creators permissions but with locked filters
+  }
+  OpenMode web_open_mode = 18;
```
begelundmuller, line 2504: See other comments – I suggest picking either the Go or proto enum, not having both. If you decide to proceed with the proto enum, then I suggest moving this out to the outer scope.

proto/rill/admin/v1/api.proto
```diff
@@ -2478,6 +2492,17 @@ message ReportOptions {
...
+  optional rill.runtime.v1.Expression filter = 19; // use with OPEN_MODE_FILTERED
```
begelundmuller, line 2505: Again, I don't think `optional` is needed – the empty value will already be `nil`, which I don't think we need to disambiguate further.

runtime/reconcilers/report.go
```diff
@@ -394,28 +397,77 @@ func (r *ReportReconciler) sendReport(ctx context.Context, self *runtimev1.Resou
...
+			}
+		}
+		if explore != "" {
+			webOpenMode = "recipient"
+		}
+	}
 
+	var reqWebOpenMode adminv1.ReportOptions_OpenMode
+	switch webOpenMode {
+	case "recipient":
```
begelundmuller, line 444: 1. We really want to avoid a dependency on the `adminv1` package in the runtime if possible, except perhaps in the `runtime/drivers/admin` package. Trying to keep circular dependencies between the two services at a minimum. 
2. Can it just pass this through as an opaque `string` and let the admin service interpret it? That would also avoid repeating the string literals again in the cases here.

admin/server/reports.go
```diff
@@ -488,6 +503,23 @@ func (s *Server) yamlForManagedReport(opts *adminv1.ReportOptions, ownerUserID s
 	res.Annotations.AdminNonce = time.Now().Format(time.RFC3339Nano)
 	res.Annotations.WebOpenPath = opts.WebOpenPath
 	res.Annotations.WebOpenState = opts.WebOpenState
+	switch opts.WebOpenMode {
+	case adminv1.ReportOptions_OPEN_MODE_RECIPIENT, adminv1.ReportOptions_OPEN_MODE_UNSPECIFIED: // backwards compatibility
+		res.Annotations.WebOpenMode = WebOpenModeRecipient
```
begelundmuller, line 508: 1. Is `web_open_mode` maybe becoming a bit of a misnomer? It doesn't sound like it dictates the access tokens being generated. Can you think of a clearer name? Maybe something like `security_mode`, `token_mode`, or just `open_mode`?
2. Should these modes perhaps apply both to the download and the open link? E.g. if the mode is "recipient", I wonder if both the download and open links should use recipient attributes for consistency?

begelundmuller (approved):

admin/server/reports.go
pjain1: 1. `web_open_mode` is just dictating how and if open link is generated [here](https://github.com/rilldata/rill/pull/6216/files#diff-c2bcafed81fcbf8eb2837df22cb63d055009c3c64d5e9d354fbc41659e7b36aeR103-R106) so seems like appropriate name?
2. In the last change we made sure that all recipients use magic token to download link (using creators' attribute if there is an owner) then if we change open mode behaviour then that will change this behaviour for `recipient` mode and thus they won't be able to download if they don't have access.

proto/rill/admin/v1/api.proto
pjain1: yeah it was validation, let me see the rule you mentioned

admin/server/reports.go
pjain1: we decided to go with current approach of using `recipient` mode for now as default and once `filtered` mode is implemented we will just use that.

PR #6220, opened 2024-12-05
tweak: update tab bar implementation

briangregoryholmes:
Moves use of `getStateManagers` out of the TabBar component.


web-common/src/features/explores/selectors.ts
```diff
@@ -237,7 +237,13 @@ export function shouldRedirectToViewWithParams(
     defaultExplorePreset,
   );
   // copy over any partial params. this will include the view and measure param
-  url.searchParams.forEach((value, key) => newUrl.searchParams.set(key, value));
+  url.searchParams.forEach((value, key) => {
+    // Ignore default view
+    if (key === "view" && value === "explore") return;
```
AdityaHegde, line 242: This is not a good way of doing this I think. But lets merge this I can do a follow up.

AdityaHegde (approved):

PR #6222, opened 2024-12-05
feat: move code/no-code toggle

briangregoryholmes:
Screenshot 2024-12-05 at 2 38 48 PM

### DIFF
--- a/web-common/src/components/forms/InputWithConfirm.svelte
+++ b/web-common/src/components/forms/InputWithConfirm.svelte
@@ -113,7 +113,7 @@
 
   .input-wrapper {
     @apply overflow-hidden;
-    @apply flex justify-center items-center pl-2;
+    @apply flex justify-center items-center pl-1;
     @apply justify-center;
     @apply border border-transparent rounded-[2px];
     @apply h-fit;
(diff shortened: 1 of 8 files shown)


jkhwu:
# Review feedback

1. Tweening animation for toggle switch
2. Increase click target to anywhere within the switch (not just the gray side)
3. Add hover tooltips that read  `Code view` and `Leave code view` or something similar.
https://www.loom.com/share/f756025f6ad84b3aba7e536403ce798c?sid=2c82b81c-d642-4ab4-bd99-81a8b90cfd0e

ericpgreen2 (approved):
LGTM. Nit (cc @jkhwu): rather than the tooltip saying "Leave code view", I'd expect it to say the name of the view that you're navigating to. Like, "no-code view", "viz view", or something like that.

jkhwu:
@briangregoryholmes I just looked at the changes and they look good to me.
@ericpgreen2 Part of the motivation for this change was that the previous name "viz" couldn't get team-buy-in, and the team didn't offer an alternative name that we could all agree on. I personally am fine with "no-code view" but not sure if others will agree.

jkhwu:
@briangregoryholmes not sure if this is related to this PR or another issue but when I was testing this branch and started a fresh project and clicked on an example project, all the dashboards stalled on the spinner until I navigated away and back (either to and from the code view or to another asset).

PR #6225, opened 2024-12-06
fix: URL params UXQA items

AdityaHegde:
- [x] Compare dimension is not shared between explore and TTD views on a fresh tab. Once sessionStore is populated this is not the case.
- [x] Dashboards without timezone shouldn't add the `tz` param. Note that the default of UTC / localIANA is used still.
- [x] Add `view` in the correct order when switching using the tabs.
- [x] `grain` should not appear in pivot.
- [x] Timezone selected should appear in the url instead of silently getting applied. This is a different behaviour that previous system.

### DIFF
--- a/web-common/src/features/dashboards/stores/dashboard-store-defaults.ts
+++ b/web-common/src/features/dashboards/stores/dashboard-store-defaults.ts
@@ -6,7 +6,6 @@ import {
 import { getPersistentDashboardState } from "@rilldata/web-common/features/dashboards/stores/persistent-dashboard-state";
 import { convertPresetToExploreState } from "@rilldata/web-common/features/dashboards/url-state/convertPresetToExploreState";
 import { getDefaultExplorePreset } from "@rilldata/web-common/features/dashboards/url-state/getDefaultExplorePreset";
-import { getLocalUserPreferencesState } from "@rilldata/web-common/features/dashboards/user-preferences";
 import type {
   V1ExploreSpec,
   V1MetricsViewSpec,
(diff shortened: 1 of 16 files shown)


web-common/src/features/dashboards/stores/dashboard-store-defaults.ts
```diff
@@ -77,5 +72,8 @@ export function restorePersistedDashboardState(
   if (persistedState.sortDirection) {
     metricsExplorer.sortDirection = persistedState.sortDirection;
   }
+
+  // const
+
```
ericpgreen2, line 77: Is this intentional?

web-common/src/features/explores/selectors.ts
```diff
@@ -239,7 +237,12 @@ export function shouldRedirectToViewWithParams(
     defaultExplorePreset,
   );
   // copy over any partial params. this will include the view and measure param
-  url.searchParams.forEach((value, key) => newUrl.searchParams.set(key, value));
+  url.searchParams.forEach((value, key) => {
+    // ignore `view` param since it is already part of partialExploreState
+    // in case `view` was same as default adding it here would mess up the order.
+    if (key === ExploreStateURLParams.WebView) return;
+    newUrl.searchParams.set(key, value);
+  });
```
ericpgreen2, line 245: It looks like this logic should go in the `convertExploreStateToURLSearchParams()` function. Whenever the `view` is the default view, it shouldn't be stringified into query params (which it looks like the `shouldSetParam()` function is for).

ericpgreen2 (approved):

web-common/src/features/dashboards/stores/dashboard-store-defaults.ts
AdityaHegde: Nope. cleaning up.

web-common/src/features/explores/selectors.ts
AdityaHegde: Good point. But `convertExploreStateToURLSearchParams` is not the one adding the params here. It is `getUrlForWebView` that always set the view, it should not set it if it is default.

ericpgreen2: I think it's fine for it to set explicit parameters, then the `load` function can remove (via redirect) any parameters that are defaults

PR #6232, opened 2024-12-06
Trino Connector 

NamanMahor:
Follow the same concept as mysql connector and create this connector for trino.

Test with trino mysql and memory catallog.


begelundmuller:
Hi @NamanMahor, thank you for contributing this PR. For a Trino connector, instead of deriving from the MySQL connector I would recommend looking at how the Snowflake, BigQuery or Athena drivers are implemented.

Since the connector doesn't serve dashboard data, but instead is for ETL into the project's OLAP database, they transfer data as exported Parquet or Arrow buffers instead of reading rows sequentially as the MySQL connector does.

This provides much faster ingestion for the scale of data that Trino often operates on. We previously had a POC connector for BigQuery that transferred data row-by-row, but we had to disable it because the performance was so poor that it threatened stability (since all the transferred rows are created in Go memory, which adds a lot of garbage collection overhead).

PR #6235, opened 2024-12-09
Fix "Source imported successfully" modal

ericpgreen2:
The modal was showing as expected for local file Sources, but it was not showing for other Source types (at least Google Cloud Storage).

Partially addresses https://github.com/rilldata/rill-private-issues/issues/991


web-common/src/features/entity-management/WatchResourcesClient.ts
```diff
@@ -181,6 +182,15 @@ export class WatchResourcesClient {
             // The following invalidations are only needed if the Source/Model has an active table
             if (!connectorName || !tableName) return;
 
+            // If it's a new source, show the "Source imported successfully" modal
+            const isNewSource =
+              res.name.kind === ResourceKind.Source &&
+              res.resource.meta.specVersion === "1";
```
AdityaHegde, line 188: Is this check enough? I remember we would sometimes get `2` as well. That is why the earlier implementation checked whether we are seeing the resource for the 1st time.

AdityaHegde (approved):

web-common/src/features/entity-management/WatchResourcesClient.ts
ericpgreen2: I believe so. We have this condition above it...
```
// The following invalidations are only needed if the Source/Model has an active table
if (!connectorName || !tableName) return;
```
... which ensures the source reconciler successfully yielded a table. It's at this point that `stateVersion === "2"`, but I don't think a check for that is important.

PR #6239, opened 2024-12-09
feat: Allow users to specify `pre_exec` and `post_exec` SQL in models

k-anshul:
This enables users to directly configure DuckDB connectors, secrets, etc. on a per-model basis.
 
Example model:
```yaml
type: model

pre_exec: ATTACH 'dbname=postgres host=localhost port=5432 user=postgres password=postgres' AS postgres_db (TYPE POSTGRES);
sql: SELECT * FROM postgres_query('postgres_db', 'SELECT * FROM USERS')
post_exec: DETACH postgres_db # Note : this is not mandatory but nice to have 

output:
  materialize: true
```

**Follow ups**:
1. Also implement the support for clickhouse models.


runtime/drivers/clickhouse/olap.go
```diff
@@ -209,7 +209,7 @@ func (c *connection) AlterTableColumn(ctx context.Context, tableName, columnName
 }
 
 // CreateTableAsSelect implements drivers.OLAPStore.
-func (c *connection) CreateTableAsSelect(ctx context.Context, name string, view bool, sql string, tableOpts map[string]any) error {
+func (c *connection) CreateTableAsSelect(ctx context.Context, name string, view bool, sql, beforeCreateSQL, afterCreateSQL string, tableOpts map[string]any) error {
```
begelundmuller, line 212: It seems like it's not executing the before/after SQL here?

runtime/drivers/clickhouse/olap.go
```diff
@@ -240,7 +240,7 @@ func (c *connection) CreateTableAsSelect(ctx context.Context, name string, view
 }
 
 // InsertTableAsSelect implements drivers.OLAPStore.
-func (c *connection) InsertTableAsSelect(ctx context.Context, name, sql string, byName, inPlace bool, strategy drivers.IncrementalStrategy, uniqueKey []string) error {
+func (c *connection) InsertTableAsSelect(ctx context.Context, name, sql, beforeInsert, afterInsert string, byName, inPlace bool, strategy drivers.IncrementalStrategy, uniqueKey []string) error {
```
begelundmuller, line 243: Seems like it's not being executed?

runtime/drivers/duckdb/model_manager.go
```diff
@@ -10,8 +10,10 @@ import (
 )
 
 type ModelInputProperties struct {
-	SQL  string `mapstructure:"sql"`
-	Args []any  `mapstructure:"args"`
+	SQL         string `mapstructure:"sql"`
+	Args        []any  `mapstructure:"args"`
+	PreInitSQL  string `mapstructure:"pre_init_sql"`
+	PostInitSQL string `mapstructure:"post_init_sql"`
```
begelundmuller, line 16: "Pre init" sounds like something that runs *before* init, and "post init" sounds like something that runs *after* init (but still *before* the actual model).

I'll suggest one of these combinations instead:
- `pre_exec` and `post_exec` (advantage: "exec" conveys that it's a statement, not a SELECT like the normal `sql:`)
- `pre_sql` and `post_sql` (advantage: aligns closer with the existing `sql:` terminology, although it might be confusing since they're not SELECT statements)
- `init_sql` and `post_sql` (advantage: aligns with the `init_sql` terminology in the connector YAML)

runtime/drivers/duckdb/olap.go
```diff
@@ -221,20 +221,33 @@ func (c *connection) AlterTableColumn(ctx context.Context, tableName, columnName
 
 // CreateTableAsSelect implements drivers.OLAPStore.
 // We add a \n at the end of the any user query to ensure any comment at the end of model doesn't make the query incomplete.
-func (c *connection) CreateTableAsSelect(ctx context.Context, name string, view bool, sql string, tableOpts map[string]any) error {
+func (c *connection) CreateTableAsSelect(ctx context.Context, name string, view bool, sql, beforeCreateSQL, afterCreateSQL string, tableOpts map[string]any) error {
```
begelundmuller, line 224: Could we pass the input props in the `CreateTableAsSelect` and `InsertTableAsSelect` functions instead of passing the different options separately?

runtime/drivers/duckdb/olap.go
```diff
@@ -257,11 +270,23 @@ func (c *connection) InsertTableAsSelect(ctx context.Context, name, sql string,
...
+					return err
+				}
+				return nil
+			}
 			// Create a temporary table with the new data
 			tmp := uuid.New().String()
 			_, err := conn.ExecContext(ctx, fmt.Sprintf("CREATE TEMPORARY TABLE %s AS (%s\n)", safeSQLName(tmp), sql))
 			if err != nil {
-				return err
+				return errors.Join(err, aferExec())
```
begelundmuller, line 289: Since this is duplicated four times (easy to mess up for the next maintainer), it would be better to use `defer` and overwrite the function return values if it errors.

runtime/drivers/duckdb/olap.go
```diff
@@ -257,11 +270,23 @@ func (c *connection) InsertTableAsSelect(ctx context.Context, name, sql string,
 
 	if strategy == drivers.IncrementalStrategyMerge {
 		err = db.MutateTable(ctx, name, func(ctx context.Context, conn *sqlx.Conn) error {
+			// Execute the pre-init SQL first
+			if beforeInsert != "" {
+				_, err := conn.ExecContext(ctx, beforeInsert)
+				return err
+			}
+			aferExec := func() error {
```
begelundmuller, line 278: Spelling: `afterExec`

runtime/pkg/rduckdb/db.go
```diff
@@ -725,15 +734,22 @@ func (d *db) openDBAndAttach(ctx context.Context, uri, ignoreTable string, read
...
-				_, err = execer.ExecContext(context.Background(), qry+" FROM 'http://nightly-extensions.duckdb.org'", nil)
+				_, err = execer.ExecContext(ctx, qry+" FROM 'http://nightly-extensions.duckdb.org'", nil)
 			}
 			if err != nil {
 				return err
 			}
 		}
+		if !read {
+			// disable any more configuration changes on the write handle via init queries
+			_, err = execer.ExecContext(ctx, "SET lock_configuration TO true", nil)
```
begelundmuller, line 748: Does this prevent the `afterCreateFn` from cleaning up e.g. a secret?

begelundmuller (changes requested):

runtime/drivers/clickhouse/olap.go
k-anshul: Yeah clickhouse is a planned follow up. Need to evaluate if these queries should this be run by the same connection. If yes, then some refactors are required. Currently there are multiple queries being executed and those can be executed on any connection. 
Also added a draft issue in my bucket to track this.

runtime/drivers/clickhouse/olap.go
k-anshul: Replied above.

runtime/drivers/duckdb/model_manager.go
k-anshul: `pre_exec` and `post_exec` sounds good to me.

runtime/drivers/duckdb/olap.go
k-anshul: Done

runtime/pkg/rduckdb/db.go
k-anshul: No it does not. The following chain of queries succeed:
```
CREATE SECRET secret...
SET lock_configuration TO true;
DROP SECRET secret;
```

begelundmuller (approved):

PR #6240, opened 2024-12-09
fix: decode known yaml fields, display errors in generic file workspace

briangregoryholmes:
Closes: https://github.com/rilldata/rill-private-issues/issues/923


runtime/compilers/rillv1/parse_rillyaml.go
```diff
@@ -92,10 +95,20 @@ func (p *Parser) parseRillYAML(ctx context.Context, path string) error {
...
 	if err := yaml.Unmarshal([]byte(data), tmp); err != nil {
 		return newYAMLError(err)
 	}
 
+	dec := yaml.NewDecoder(strings.NewReader(data))
+	dec.KnownFields(true)
+
+	err = dec.Decode(tmp)
+
+	if err != nil && !errors.Is(err, io.EOF) {
```
begelundmuller, line 108: nit: Go convention is to handle an `err` in the same code block as the one where it's set
```suggestion
	dec := yaml.NewDecoder(strings.NewReader(data))
	dec.KnownFields(true)

	err = dec.Decode(tmp)
	if err != nil && !errors.Is(err, io.EOF) {
```

runtime/compilers/rillv1/parse_rillyaml.go
```diff
@@ -40,6 +42,7 @@ type VariableDef struct {
 
 // rillYAML is the raw YAML structure of rill.yaml
 type rillYAML struct {
```
begelundmuller, line 44: I think you will also need to add handling of `mock_users:` here, which I believe the frontend currently adds to enable testing security policies locally. (Eventually it would probably be good to expose those over the APIs so the UI can avoid parsing them separately, but that's for another PR.)

begelundmuller (changes requested):

begelundmuller (approved):
LGTM!

PR #6247, opened 2024-12-10
feat: measure filter ux

briangregoryholmes:
Closes: https://github.com/rilldata/rill-private-issues/issues/980


web-common/src/features/dashboards/filters/measure-filters/MeasureFilter.svelte
```diff
@@ -1,34 +1,148 @@
...
+          value1: filter?.value1 ?? "",
+          value2: filter?.value2 ?? "",
+        },
+      });
     }
   }
+
+  function expressionIsBetween(op: MeasureFilterOperation | string) {
+    return (
+      isMFO(op) &&
```
AdityaHegde, line 124: Why is this extra check needed?

web-common/src/features/dashboards/filters/measure-filters/MeasureFilter.svelte
```diff
@@ -1,34 +1,148 @@
...
+  }
 </script>
 
-<DropdownMenu.Root
+<svelte:window
+  on:keydown={(e) => {
+    if (e.key === "Escape") {
+      handleDismiss();
+    } else if (e.key === "Enter") {
+      submit();
```
AdityaHegde, line 140: Why do we need this on top of `onEnter` for inputs? Would this interfere with other inputs outside of this form?

AdityaHegde (approved):

web-common/src/features/dashboards/filters/measure-filters/MeasureFilter.svelte
briangregoryholmes: We have @typescript-eslint/no-unsafe-enum-comparison set to warn. I haven't done the work to make our Input/Select components accept generic props (mostly because there isn't much benefit), so the change function where we need to use `expressionIsBetween` only knows that the new value is a string. Can't type cast in the Svelte template section yet and didn't want to pull the function into the script tag. So, I'm type guarding the string before making the enum comparison.

web-common/src/features/dashboards/filters/measure-filters/MeasureFilter.svelte
briangregoryholmes: We have @typescript-eslint/no-unsafe-enum-comparison set to warn. I haven't done the work to make our Input/Select components accept generic props (mostly because there isn't much benefit), so the change function where we need to use expressionIsBetween only knows that the new value is a string. Can't type cast in the Svelte template section yet and didn't want to pull the function into the script tag. So, I'm type guarding the string before making the enum comparison.

PR #6255, opened 2024-12-11
fix: dimension values with null measure values in dimension table are not shown when filter by that measures

AdityaHegde:


djbarnwal:
As discussed, the sorted values show NULL values at top when sorting in descending order. 
https://www.notion.so/rilldata/Bug-Bash-Dec-2024-159ba33c8f578027970beb993567142f?pvs=4#159ba33c8f5780549f6dc4f34a4636dc

AdityaHegde:
@djbarnwal lets take that separately. The fix will be quite independant

djbarnwal (approved):

PR #6260, opened 2024-12-11
Don't show the "Source imported successfully" modal when unpacking example projects

ericpgreen2:
Reported in the [bug bash](https://www.notion.so/rilldata/Bug-Bash-Dec-2024-159ba33c8f578027970beb993567142f?pvs=4#159ba33c8f5780dab177dd5ba8af2ed5)

### DIFF
--- a/web-common/src/features/feature-flags.ts
+++ b/web-common/src/features/feature-flags.ts
@@ -30,7 +30,6 @@ type FeatureFlagKey = keyof Omit<FeatureFlags, "set">;
 class FeatureFlags {
   adminServer = new FeatureFlag("rill", false);
   readOnly = new FeatureFlag("rill", false);
-  sourceImportedModal = new FeatureFlag("rill", true);
 
   ai = new FeatureFlag("user", !import.meta.env.VITE_PLAYWRIGHT_TEST);
   exports = new FeatureFlag("user", true);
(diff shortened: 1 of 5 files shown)


web-common/src/features/feature-flags.ts
```diff
@@ -30,6 +30,7 @@ type FeatureFlagKey = keyof Omit<FeatureFlags, "set">;
 class FeatureFlags {
   adminServer = new FeatureFlag("rill", false);
   readOnly = new FeatureFlag("rill", false);
+  sourceImportedModal = new FeatureFlag("rill", true);
```
AdityaHegde, line 33: We should not use `FeatureFlags` for this. Lets have a separate set of values for such use cases.

AdityaHegde (approved):

briangregoryholmes:
I agree with Aditya, but also think we should just remove this modal entirely for the time being. Forcing the user to choose one of these options isn't a good user experience even when it works properly. This should be part of a more robust toast notification system or to a way of pointing out the CTA button when certain conditions are met.

If you do go forward with this PR, the reason it shouldn't be shown is because there is already a downstream dashboard, so I would look into basing the hide/show rules around that, which may not require the use of a store.

ericpgreen2:
> I agree with Aditya, but also think we should just remove this modal entirely for the time being. Forcing the user to choose one of these options isn't a good user experience even when it works properly. This should be part of a more robust toast notification system or we should develop a way of pointing out the CTA button via a small animation when certain conditions are met.

I agree this modal feels quite ham-fisted and that we could give the user a much more elegant nudge. For the time being, for the sake of keeping _some_ CTA (which I believe Mike feels strongly about), I'll patch this as is.

> If you do go forward with this, the reason it shouldn't be shown is because there is already a downstream dashboard, so I would look into basing the hide/show rules around that, which may not require the use of a store.

Done!

PR #6265, opened 2024-12-12
feat: enable cache for external tables using a cache control on metrics views

k-anshul:
…rics views


runtime/reconcilers/metrics_view.go
```diff
@@ -119,11 +122,69 @@ func (r *MetricsViewReconciler) Reconcile(ctx context.Context, n *runtimev1.Reso
...
+		cache.KeyTtlSeconds = 60
+	}
+
+	if spec.Cache != nil && spec.Cache.KeySql != "" {
+		cache.KeySql = spec.Cache.KeySql
+	} else {
+		if streaming {
+			cache.KeySql = fmt.Sprintf("SELECT %s FROM %s", watermarkForMetricsView(spec), dialect.EscapeTable(spec.Database, spec.DatabaseSchema, spec.Table))
+		} else {
+			cache.KeySql = "SELECT " + fmt.Sprintf("'%d:%d'", updatedOn.GetSeconds(), updatedOn.GetNanos()/int32(time.Millisecond))
```
k-anshul, line 173: May be the version is unnecessary for rill managed models because caches will anyways expire. We can also return an empty cache key ?

runtime/reconcilers/metrics_view.go
```diff
@@ -119,11 +122,69 @@ func (r *MetricsViewReconciler) Reconcile(ctx context.Context, n *runtimev1.Reso
...
 	}
+	// Override the cache control with our calculated cache control
+	if validSpec != nil {
+		validSpec.Cache = metricsViewCacheControl(validSpec, mv.State.Streaming, self.Meta.StateUpdatedOn, olap.Dialect())
+	}
 
 	return runtime.ReconcileResult{Err: validateErr}
 }
+
+func metricsViewCacheControl(spec *runtimev1.MetricsViewSpec, streaming bool, updatedOn *timestamppb.Timestamp, dialect drivers.Dialect) *runtimev1.MetricsViewSpec_Cache {
```
begelundmuller, line 151: I think it would be better to move the logic (which basically just applies defaults) in this function to the `metricsview_cache_key` resolver – to keep the cache key logic in a single place. (It also makes refactoring the defaults easier because they will only be applied at runtime, not persisted in the DB as here.)

runtime/reconcilers/metrics_view.go
begelundmuller: See note above – if you move this logic to the resolver, you won't need to worry about creating a SQL statement that returns a constant

runtime/server/batch_query_test.go
```diff
@@ -7,6 +7,7 @@ import (
 	"testing"
 
 	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
+	_ "github.com/rilldata/rill/runtime/resolvers"
```
begelundmuller, line 10: Underscore imports should be in a separate group (linter enforces this for non-test files)

begelundmuller (changes requested):

runtime/metricsview/executor.go
```diff
@@ -67,8 +67,7 @@ func (e *Executor) Close() {
 
 // Cacheable returns whether the result of running the given query is cacheable.
 func (e *Executor) Cacheable(qry *Query) bool {
-	// TODO: Get from OLAP instead of hardcoding
-	return e.olap.Dialect() == drivers.DialectDuckDB
+	return *e.metricsView.Cache.Enabled
```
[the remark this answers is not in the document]
k-anshul: I am making sure that `Enabled` is not set as nil in reconciler but I can check the nil checks as well.

runtime/reconcilers/metrics_view.go
k-anshul: 1. We still need to validate for `key_sql` i.e. whether we are able to resolve a valid key_sql for streaming metrics views. It will also be ideal to do that validation in reconciler and not during query time. Since we are anyways doing some validations it made sense to also apply defaults.
2. In reconciler I always mark cache enabled or disabled which the resolvers can use to check if caching should be enabled and get `metricsViewCacheKey`. Now resolvers will need to query every time and handle an error thrown by `metrics_cache_key` resolver.

>It also makes refactoring the defaults easier because they will only be applied at runtime, not persisted in the DB as here

Since defaults can only be changed with a code change so it should make sure the defaults are also updated on metrics view reconcile on restart.

Having said that I also thought about keeping this in resolver and was indifferent to both. 
I can move this logic to resolver as well.

runtime/query.go
```diff
@@ -101,18 +98,6 @@ func (r *Runtime) Query(ctx context.Context, instanceID string, query Query, pri
...
-	// NOTE: This hack is removed in the new resolvers by letting the resolver itself decide whether to cache or not.
-	olap, release, err := r.OLAP(ctx, instanceID, olapConnector)
-	if err != nil {
-		return err
-	}
-	dialect := olap.Dialect()
-	release()
-	if dialect != drivers.DialectDuckDB {
-		return query.Resolve(ctx, r, instanceID, priority)
-	}
```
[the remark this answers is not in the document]
k-anshul: This will make sure profiling queries are cached for clickhouse as well which is what we want right ? And the results will get invalidated if source/model change similar to how it happens in duckdb.

runtime/reconcilers/metrics_view.go
k-anshul: This was more of a convenient way of generating `runtime.ResolverResult` using results of a SQL query. May be we can still keep the same behaviour but return a constant (not driven by `updatedOn`). Or do you have something else in mind ?

runtime/reconcilers/metrics_view.go
k-anshul: By the way if we don't apply defaults in reconciler than we can't resolve this todo since executor can't call `metrics_cache_key` resolver. 
https://github.com/rilldata/rill/blob/ba6f2ba9d62b591db91f8bc1af202f64e68fa64e/runtime/metricsview/executor.go#L68-L72

runtime/metricsview/executor.go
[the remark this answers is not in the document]
begelundmuller: Yeah I think it's good to check

runtime/query.go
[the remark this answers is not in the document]
begelundmuller: I don't really know – what if someone refreshes their underlying data and then don't understand why Rill Developer doesn't reflect the changes?

Maybe caching profiling for a few mins would be good, but waiting until the next restart sounds like it might be too confusing.

runtime/reconcilers/metrics_view.go
k-anshul: I can think of following ideas:
1. We discussed on resolving this via a refresh schedule on metrics_view. This could be evaluated. In that case we can compute the key and store it directly in spec.
2. Another option is to let runtime expose APIs to set/get data in query_cache which the executor can utilise to set/get `metrics_cache_key` in the cache. This could be useful for other use cases as well. Like as of now we are always querying watermark so it could be cached as well(May be it is not queried always, just thinking out aloud).

k-anshul: Option 2 sounds most convenient to me. Good chance we may come up with similar use cases in the future as well.

runtime/query.go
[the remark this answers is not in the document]
k-anshul: Okay my understanding was that for rill managed tables (source and models) users should not add data externally.
I can revert to original implementation as well.

runtime/reconcilers/metrics_view.go
begelundmuller: Since the `key_sql` will require another DB roundtrip, it would be great only to do it for dynamic results (like `MAX(__time)`). When it's a constant, returning it immediately in Go (without the SQL roundtrip) would be nice.

runtime/query.go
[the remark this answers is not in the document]
begelundmuller: Maybe I misunderstood the context here – I was thinking about profiling on user-managed tables, not Rill-managed tables.

I think the UI may still run profiling on user-managed tables in some cases, but I might be wrong. Anyway they might start doing it, so it would be nice not to have a tricky-to-debug cache issue surface if they do.

runtime/reconcilers/metrics_view.go
begelundmuller: Right now we basically only have caching happening in `runtime.Runtime.Resolve` (and the legacy `runtime.Runtime.Query`), which is nice because it's much simpler than when caching happens at multiple levels. I think it would be very nice to avoid caching happening inside the `metricsview` package – so the guarantee basically becomes that it always returns fresh data, which upstream callers can cache (we may need multi-level caching at some point, but I think we may be able to avoid it for now).

Looking at the resolvers, the `Cacheable` and `Key` functions are only called from here: https://github.com/rilldata/rill/blob/8e7bca1dc80ebeb7fb1f32df2312a02f0d4228e1/runtime/resolver.go#L126-L136.

What if we replaced these with a `CacheKey(ctx context.Context) (key []byte, ok bool)` function (where caching should be skipped if `!ok`)? Then the `metrics`, `metrics_sql` and `metrics_time_range` resolvers could call the `metrics_cache_key` resolver from their `CacheKey` implementation – if they pass that call through `opts.Runtime.Resolve`, then it in turn will also be cached.

Do you think that will work? The advantage of this approach is that there continues only to be caching happening in `runtime.Runtime.Resolve`, and resolvers are only being called from other resolvers (not from the `metricsview` package). This doesn't quite account for the legacy `runtime.Runtime.Query` implementation, but hopefully the above idea can also work for that in a similar way?

When it comes to the actual execution (but not caching) of the `metrics_cache_key` resolver, that in turn could actually be implemented inside the `metricsview` package (similar to how the logic for the `metrics` resolver is implemented there).

Lastly, to comment on this:
> Like as of now we are always querying watermark so it could be cached as well
You are thinking about this, right? https://github.com/rilldata/rill/blob/8e7bca1dc80ebeb7fb1f32df2312a02f0d4228e1/runtime/metricsview/executor_rewrite_time.go#L45

I agree that it would be nice to have this cached (it hasn't been urgent since the UI currently always passes fixed time ranges). In this case, we could also avoid multi-level caching by instead moving this to the resolver – e.g. you could have something like this in the `metrics` resolver (pseudocode):
```go
// Create an executor
executor := ...

// Call watermark resolver for the metrics view (can be cached in the resolver cache)
watermark := opts.Runtime.Resolve(...)

// Bind the query using the provided watermark (to avoid nested, uncached queries for the watermark)
qry := executor.BindWatermark(qry, watermark)

// Now execute the qry
res := executor.Execute(qry)
```

k-anshul: Overall the `CacheKey` API seems good to me.
But I am somewhat unconvinced with how it is computed in resolvers.
1. Feels like too much indirection for `CacheKey`. It is hard to understand and trace the calls mentally. `metrics` and other resolvers need to call another resolver via the `runtime.Resolve` method to resolve cache key so that data can be cached. 
2. Some APIs on executor like `Query` can be called directly but other APIs like `CacheKey` need to go via the resolver path. So the decision to cache something is split across multiple places.
3. Overall the resolver construct is a generic construct but the results are stored in a `query_cache`. Though no resolver apart from the ones that operate on a metrics_view cache results as of now but the construct allows it. It may be easily fixable by calling the cache as `resolver_cache`.


May be for internal resolvers there should be a uncomplicated way to get and set things in query_cache to avoid too many indirections.

runtime/query.go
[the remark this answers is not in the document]
k-anshul: Okay so for external tables the following check(already present) will make sure caching is disabled :
```
// If there were no known dependencies, skip caching
if len(depKeys) == 0 {
	return query.Resolve(ctx, r, instanceID, priority)
}
```
For models and sources it will be good to cache results similar to duckdb.
What do you think ?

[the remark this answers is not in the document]
begelundmuller: Makes sense, yeah that check seems sufficient!

runtime/metricsview/executor.go
```diff
@@ -65,10 +67,46 @@ func (e *Executor) Close() {
...
+	if err != nil {
+		return nil, false, err
+	}
+	defer res.Close()
+	var key string
+	for res.Next() {
+		if err := res.Scan(&key); err != nil {
+			return nil, false, err
+		}
+	}
```
begelundmuller, line 105: If the cache key SQL doesn't have an aggregation or `LIMIT`, this may accidentally process a ton of data. Maybe using `if` instead would provide at least a little protection against that?

```suggestion
	if res.Next() {
		if err := res.Scan(&key); err != nil {
			return nil, false, err
		}
	}
```

runtime/metricsview/executor.go
```diff
@@ -65,10 +67,46 @@ func (e *Executor) Close() {
...
+	if res.Err() != nil {
+		return nil, false, err
+	}
+	return []byte(key), true, nil
```
begelundmuller, line 109: Not sure if we want to protect agains this, but often the key will be a timestamp and people may forget to cast it to a string. 

An easy way to protect here is to use `var key any` and then run it through `jsonval.ToValue` and then `json.Marshal` – that will almost always give you a safely serialized string representation.

runtime/query.go
```diff
@@ -78,21 +76,18 @@ func (r *Runtime) Query(ctx context.Context, instanceID string, query Query, pri
...
 		// Add to cache key.
 		// Using StateUpdatedOn instead of StateVersion because the state version is reset when the resource is deleted and recreated.
 		key := fmt.Sprintf("%s:%s:%d:%d", res.Meta.Name.Kind, res.Meta.Name.Name, res.Meta.StateUpdatedOn.Seconds, res.Meta.StateUpdatedOn.Nanos/int32(time.Millisecond))
+		if mv := res.GetMetricsView(); mv != nil {
+			cacheKey, err := r.metricsViewCacheKey(ctx, instanceID, res.Meta.Name.Name, priority)
+			if err != nil {
+				// skip caching
+				// the cache_key_resolver should ideally only return an error if caching is disabled or context is cancelled
+				return query.Resolve(ctx, r, instanceID, priority)
+			}
```
begelundmuller, line 88: Can we have a dedicated `skip bool` for skipping caching. Because if the `err` is a context cancellation, that's ideally an error we return immediately. It will also make it easier to understand if an unexpected error is actually returned.

runtime/resolvers/metrics.go
```diff
@@ -87,16 +88,34 @@ func (r *metricsResolver) Close() error {
...
+	// get the underlying executor's cache key
+	key, ok, err := cacheKeyForMetricsView(ctx, r.runtime, r.instanceID, r.query.MetricsView, r.args.Priority)
+	if err != nil {
+		return nil, false, err
+	}
+	if !ok {
+		return nil, false, nil
+	}
+
+	hasher := md5.New()
```
begelundmuller, line 101: This is already being hashed in `runtime/resolver.go` so could we return the full pointer here?

runtime/resolvers/metricsview_cache_key.go
```diff
@@ -0,0 +1,159 @@
...
+	streaming  bool
+	exectuor   *metricsview.Executor
+	args       *metricsViewCacheKeyResolverArgs
+}
+
+type metricsViewCacheKeyResolverArgs struct {
+	Priority int `mapstructure:"priority"`
+}
+
+type metricsViewCacheKey struct {
```
begelundmuller, line 39: Should be `metricsViewCacheKeyProps`

runtime/query.go
```diff
@@ -158,6 +141,24 @@ func (r *Runtime) Query(ctx context.Context, instanceID string, query Query, pri
...
+		InstanceID:         instanceID,
+		Resolver:           "metrics_cache_key",
+		ResolverProperties: map[string]any{"metrics_view": name},
+		Args:               map[string]any{"priority": priority},
+		Claims:             &SecurityClaims{SkipChecks: true},
+	})
+	if err != nil {
+		return nil, err
+	}
+	cacheKey, err := cacheKeyResolver.MarshalJSON()
```
begelundmuller, line 155: Maybe use `.Next()` and access `res["key"]` since it's already being serialized to a string/bytes upstream? 

Or alternatively, you could consider returning `*drivers.Result` from `metricsview.Executor.CacheKey`? Then we would get support for the cache key SQL returning multiple columns.

runtime/resolvers/metricsview_cache_key.go
```diff
@@ -0,0 +1,159 @@
...
+func (r *metricsViewCacheKeyResolver) Cacheable() bool {
+	return true
+}
+
+func (r *metricsViewCacheKeyResolver) CacheKey(ctx context.Context) ([]byte, bool, error) {
+	var sb strings.Builder
+	sb.WriteString(runtime.ResourceKindMetricsView)
+	sb.WriteString(":")
+	sb.WriteString(r.mvName)
+	sb.WriteString(":")
```
begelundmuller, line 107: If you add the metrics view in `Refs()`, then the caller will take care of adding this to the cache key.

runtime/resolvers/metricsview_cache_key.go
```diff
@@ -0,0 +1,159 @@
...
+	sb.WriteString("cahe_key")
```
begelundmuller, line 108: Spelling: `cache_key`

runtime/resolvers/metricsview_cache_key.go
```diff
@@ -0,0 +1,159 @@
...
+	hash, err := hashstructure.Hash(sb.String(), hashstructure.FormatV2, nil)
+	if err != nil {
+		return nil, false, err
+	}
+	return []byte(strconv.FormatUint(hash, 16)), true, nil
+}
+
+func (r *metricsViewCacheKeyResolver) Refs() []*runtimev1.ResourceName {
+	return []*runtimev1.ResourceName{}
+}
```
begelundmuller, line 128: I think it should have a ref to the metrics view to ensure invalidation if the underlying model is updated when `streaming` is false?

runtime/resolvers/metricsview_cache_key.go
```diff
@@ -0,0 +1,159 @@
...
+	sb.WriteString("cahe_key")
+	ttlSeconds := r.mv.CacheKeyTtlSeconds
+	if ttlSeconds == 0 && r.streaming {
+		// If streaming, we need to cache the key for 60 seconds
+		// For non streaming metrics view we don't need to expire the key as data itself will be invalidated basis the ref's state version
+		ttlSeconds = 60
+	}
+	if ttlSeconds != 0 {
+		sb.WriteString(":")
+		sb.WriteString(truncateTime(time.Now(), r.mv.CacheKeyTtlSeconds).Format(time.RFC3339))
```
begelundmuller, line 117: It's not using `ttlSeconds`:

```suggestion
		sb.WriteString(truncateTime(time.Now(), ttlSeconds).Format(time.RFC3339))
```

runtime/resolvers/metricsview_cache_key.go
```diff
@@ -0,0 +1,159 @@
...
+	sb.WriteString(":")
+	sb.WriteString(r.mvName)
+	sb.WriteString(":")
+	sb.WriteString("cahe_key")
+	ttlSeconds := r.mv.CacheKeyTtlSeconds
+	if ttlSeconds == 0 && r.streaming {
+		// If streaming, we need to cache the key for 60 seconds
+		// For non streaming metrics view we don't need to expire the key as data itself will be invalidated basis the ref's state version
+		ttlSeconds = 60
+	}
```
begelundmuller, line 114: This might be simpler:
```go
ttl := r.mv.CacheKeyTtlSeconds * time.Second
if ttl == 0 && r.streaming {
  ttl = time.Minute
}
```
and then you don't need `truncateTime` because you can just call `time.Now().Truncate(ttl)`

runtime/resolvers/sql.go
```diff
@@ -94,8 +94,11 @@ func (r *sqlResolver) Cacheable() bool {
 	return false
```
begelundmuller, line 94: The `Cacheable` func here can be removed now, right?

begelundmuller (changes requested):

runtime/reconcilers/metrics_view.go
begelundmuller: > Feels like too much indirection for CacheKey

Agree, but I think it's intrinsic to the idea of a "cached cache key", not this implementation.

> Some APIs on executor like Query can be called directly but other APIs like CacheKey need to go via the resolver path. So the decision to cache something is split across multiple places.

You must have misunderstood `Executor.Query` – it never does any caching. So on the contrary, this ensures consistency because caching is *never* implemented in the `metricsview` package, and only implemented one place (in the `runtime/resolver.go` function (and the legacy `runtime/query.go`, but that's being deprecated)).

> It will be ideal to not let resolvers call other resolvers to prevent some cyclic calls.

I think resolvers calling other resolvers can actually be very powerful. Think of resolvers as memoized functions – it's completely fine for functions to call other functions, but of course, you should be careful if you call the same function recursively. 

> there should be a uncomplicated way to get and set things in query_cache to avoid too many indirections.

I'm conversely more worried about the potential for bugs if stuff is getting cached at many different levels in the call chain. Having caching only happen in `Resolve` IMO makes it easier to reason about when things are or are not cached.

runtime/metricsview/executor.go
k-anshul: Weirdly I did not find this issue in my testing and on digging further I realise that the `sql` package already does conversion when source is `time.Time` and dest is `*string` : https://github.com/golang/go/blob/669d87a935536eb14cb2db311a83345359189924/src/database/sql/convert.go#L286
It does this conversions for most common types.

But I think it is good to run it through `jsonval.ToValue` since it handles more types.

runtime/query.go
k-anshul: Okay. In that case I think it will be best to expose `errCachingDisabled` and check for it directly for now. Once query package is removed that can be made internal as well.

runtime/resolvers/metrics.go
k-anshul: The values were already being hashed in the older implementation : https://github.com/rilldata/rill/blob/9061a24267f0b562c5304f0d897813fed88055c7/runtime/resolvers/metrics.go#L94

Since the key is now composed of two things (`cache_key` and `query`) I thought of just using a md5 hasher. 
Do you mean we create a new struct and just return any from `CacheKey` instead of bytes ?

runtime/query.go
k-anshul: > Maybe use .Next() and access res["key"] since it's already being serialized to a string/bytes upstream?

Sounds good. 

> You could consider returning *drivers.Result from metricsview.Executor.CacheKey

I don't understand how this would work since the callers need to access this via `ResolverResult` interface.

> Then we would get support for the cache key SQL returning multiple columns.

Given `Executor.CacheKey` returns `[]bytes`, technically it can still support multiple columns.

runtime/resolvers/metricsview_cache_key.go
k-anshul: Yes. Previously when not having the cacheable interface I was calling this for every metrics view and adding a ref on metrics_view would lead to circular dependency but now given it is called by resolvers it can add a ref on metrics view.
Though invalidation due to model update still happens because the queries also have the model as ref.

runtime/reconcilers/metrics_view.go
k-anshul: > Agree, but I think it's intrinsic to the idea of a "cached cache key", not this implementation.

Same indirection will be present if resolvers need to load watermark as well.

> You must have misunderstood Executor.Query – it never does any caching

No I didn't meant that. I meant `metrics` resolver and other such resolvers need to call `Executor.Query` directly but call `Executor.CacheKey` via `runtime.Resolve`.

> I think resolvers calling other resolvers can actually be very powerful. 

I mean there is no denying that it is powerful. My reservations are more around ease of usage and ease of understanding. Creating resolvers for internal usage feels like adding too much bloat. I think there has to be a tradeoff between "perfect" abstractions and ease of understanding/usage. May be we are now in _it is subjective_ territory.

> It's subjective, but I'm conversely more worried about the potential for bugs if stuff is getting cached at many different levels in the call chain.

I don't know if the current approach is less bug prone. It is easy to miss the fact calling the `Executor. CacheKey` in resolver needs to go via `runtime.Resolve` and not to be called directly which is not the case with other APIs on executor.

runtime/resolvers/metrics.go
begelundmuller: I just meant if one way or another it could use the same hasher – e.g. it might return a combined `[]byte` or a `*drivers.Result`, or accept a `io.Writer`, or something like that.

It's not an important comment, certainly you shouldn't do anything weird to achieve this (like returning a struct would probably be weird). Just wondered if there was a clean way to have just one hasher.

runtime/reconcilers/metrics_view.go
begelundmuller: I think this is all subjective, and we may need multiple levels of caching in the future, but as long as possible, I think we should aim to keep caching and invalidation in a single place (`Resolve`). If someone forgets to use it, the worst case is that something doesn't get cached until we notice.

runtime/resolvers/metrics.go
k-anshul: Okay. Added another way to compute key without hasher.

runtime/metricsview/executor.go
```diff
@@ -65,10 +68,56 @@ func (e *Executor) Close() {
...
+	spec := e.metricsView
+	// Cache is disabled for metrics views based on external table
+	if (spec.CacheEnabled != nil && !*spec.CacheEnabled) || (spec.CacheEnabled == nil && e.streaming) {
+		return nil, false, nil
+	}
+
+	if spec.CacheKeySql == "" {
+		if !e.streaming {
+			// for metrics views on rill managed tables no cache key specific to the metrics view is required
+			return []byte(""), true, nil
```
begelundmuller, line 83:
```suggestion
			// for metrics views on rill managed tables, we can cache forever
			// (until the metrics view is refreshed/edited, which always leads to cache invalidations)
			return []byte(""), true, nil
```

runtime/resolvers/metricsview_cache_key.go
```diff
@@ -0,0 +1,133 @@
...
+	runtime.RegisterResolverInitializer("metrics_cache_key", newMetricsViewCacheKeyResolver)
+}
+
+type metricsViewCacheKeyResolver struct {
+	runtime    *runtime.Runtime
+	instanceID string
+	mvName     string
+	mv         *runtimev1.MetricsViewSpec
+	streaming  bool
+	exectuor   *metricsview.Executor
```
begelundmuller, line 26: nit: typo, should be `executor`

begelundmuller (approved):

PR #6267, opened 2024-12-12
Auth: check if user exists

lovincyrus:
https://www.notion.so/rilldata/Auth0-Actions-159ba33c8f5780c79969dc5036beeacf


web-auth/src/components/Auth.svelte
```diff
@@ -20,111 +19,264 @@
...
-    if (cloudClientIDsArr.includes(config?.clientID)) {
-      isLegacy = true;
+    webAuth.authorize({
+      connection: connectionName,
+      login_hint: email,
+      prompt: "login",
+    });
+  }
+
+  // TODO: Serverless function
```
lovincyrus, line 123: Where can we deploy this token generation function on the server side in our current techstack? @ericpgreen2

https://www.notion.so/rilldata/Post-Canvas-Implementation-Notes-190ba33c8f5780c99e8bd66bda682125#192ba33c8f5780c4b4ebf73797b96b49

lovincyrus: Where can I deploy this piece of server-side code? Primarily the usage is for generating bearer token and checking user existence @begelundmuller

ericpgreen2: What about in an Auth0 Action? IIRC, that's what our Auth0 contact advised.

lovincyrus: Explained in https://www.notion.so/rilldata/Post-Canvas-Implementation-Notes-190ba33c8f5780c99e8bd66bda682125

ericpgreen2: So, one of Auth0's recommendations was to move HRD from client-side to the Auth0-side. If we did that, and let's say HRD doesn't recognize an email – does that help us?

lovincyrus: If HRD doesn't recognize an email, the user will have to switch to the sign-up form to create a new account. HRD is only useful for Enterprise and Social logins based on the provided user email.

If we want to ensure a unified auth flow, we need to add a user existence check hence https://rilldata.slack.com/archives/C080SHAMZU2/p1738953939372419

PR #6270, opened 2024-12-12
fix: comparison time ranges resetting when switching time ranges

AdityaHegde:
- When switching time ranges, comparison time range would reset.
- Just after switching on compare dimension toggling comparison time range would not apply.
- If comparison is disabled comparison based criteria was still allowed in alerts.

### DIFF
--- a/web-common/src/features/alerts/criteria-tab/getTypeOptions.ts
+++ b/web-common/src/features/alerts/criteria-tab/getTypeOptions.ts
@@ -17,6 +17,7 @@ export function getTypeOptions(
     disabled?: boolean;
     tooltip?: string;
   }[] = [...MeasureFilterBaseTypeOptions];
+  console.log(formValues.comparisonTimeRange);
 
   if (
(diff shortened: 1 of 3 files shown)


web-common/src/features/alerts/criteria-tab/getTypeOptions.ts
```diff
@@ -17,6 +17,7 @@ export function getTypeOptions(
     disabled?: boolean;
     tooltip?: string;
   }[] = [...MeasureFilterBaseTypeOptions];
+  console.log(formValues.comparisonTimeRange);
```
djbarnwal, line 20: Stale log

djbarnwal (commented):
Is the change to the `Select` component intentional in this PR? I see the change in the component but no change in it's usage

AdityaHegde:
Looks like a refactor to Select component removed the capability needed in alerts. We still need the readonly option with tooltip so adding it back.

djbarnwal (approved):

PR #6271, opened 2024-12-12
Fix the generated password not being idempotent in the ClickHouse provisioner

begelundmuller:
When editing the `clickhouse.yaml` file with keystroke-by-keystroke editing, provisioning calls can be rapidly cancelled and retried. This surfaced an idempotency issue where calling `CREATE USER IF NOT EXISTS` did not use the generated random password, causing the generated DSN to be invalid. This PR fixes the problem by leveraging ClickHouse's support for adding multiple passwords to one user.

### DIFF
--- a/admin/provisioner/clickhousestatic/provisioner.go
+++ b/admin/provisioner/clickhousestatic/provisioner.go
@@ -102,14 +102,15 @@ func (p *Provisioner) Provision(ctx context.Context, r *provisioner.Resource, op
 		return nil, fmt.Errorf("failed to create clickhouse database: %w", err)
 	}
 
-	// Idempotently create the user
-	_, err = p.ch.ExecContext(ctx, fmt.Sprintf("CREATE USER IF NOT EXISTS %s DEFAULT DATABASE %s GRANTEES NONE", user, dbName))
+	// Idempotently create the user.
+	_, err = p.ch.ExecContext(ctx, fmt.Sprintf("CREATE USER IF NOT EXISTS %s IDENTIFIED WITH sha256_password BY ? DEFAULT DATABASE %s GRANTEES NONE", user, dbName), password)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create clickhouse user: %w", err)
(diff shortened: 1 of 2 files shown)


kaspersjo (approved):

admin/provisioner/clickhousestatic/provisioner.go
```diff
@@ -102,12 +102,20 @@ func (p *Provisioner) Provision(ctx context.Context, r *provisioner.Resource, op
...
+	// Idempotently create the user.
 	_, err = p.ch.ExecContext(ctx, fmt.Sprintf("CREATE USER IF NOT EXISTS %s IDENTIFIED WITH sha256_password BY ? DEFAULT DATABASE %s GRANTEES NONE", user, dbName), password)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create clickhouse user: %w", err)
 	}
 
+	// When creating the user, the password assignment is not idempotent (if there are two concurrent invocations, we don't know which password was used).
+	// By adding the password separately, we ensure all passwords will work.
+	// NOTE: Requires ClickHouse 24.9 or later.
+	_, err = p.ch.ExecContext(ctx, fmt.Sprintf("ALTER USER %s ADD IDENTIFIED WITH sha256_password BY ?", user), password)
```
nishantmonu51, line 114: Seems risky, in a rare case of id collision, would it end up giving an existing project access to previously created DB ? However, in case of id collision we would have even bigger troubles..
Usually, locking on table name would be less prone to errors...

begelundmuller: Thanks for taking a look – this should not be a problem because:
1. The risk of an ID collision is the same as the risk of guessing the password (both have 16 bytes of randomness), so extremely unlikely.
2. Even if there's an ID collision, the resource is inserted in our Postgres before we provision the ClickHouse account. This will cause a unique constraint error on the ID in Postgres and ClickHouse will not be provisioned until the next retry (which will generate a new ID).

PR #6280, opened 2024-12-16
fix: avoid using global runtime in loader funciton

AdityaHegde:
We were setting the global runtime in project's loader function and using it in explore's loader function to fetch explore spec and other data. This would cause issue with preloading routes where loader function gets called. Hovering over other projects in breadcrumbs would update the runtime causing the dashboard list to be incorrect.

Here we are making explicit calls to `httpRuntime` with a custom `Runtime` object. This should not be the final solution since changes to API is not auto generated. We should look at autogenerating these `fetch` wrappers for every API.


AdityaHegde:
We are opting to disable preload on links to other projects in the breadcrumbs and any other places.

PR #6293, opened 2024-12-17
fix : Clickhouse driver should always use native protocol unless a DSN is set directly

k-anshul:
In my testing on local, running a query 100 times that returns 30K rows, native worked ~30% faster.

Also fixes https://github.com/rilldata/rill-private-issues/issues/1013 
The HTTP protocol does not return column types if empty results whereas native protocol does (Issue is already open in clickhouse).

### DIFF
--- a/runtime/drivers/clickhouse/clickhouse.go
+++ b/runtime/drivers/clickhouse/clickhouse.go
@@ -149,14 +149,11 @@ func (d driver) Open(instanceID string, config map[string]any, st *storage.Clien
 			host = fmt.Sprintf("%s:%d", conf.Host, conf.Port)
 		}
 		opts.Addr = []string{host}
+		opts.Protocol = clickhouse.Native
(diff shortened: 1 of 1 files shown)


nishantmonu51:
We should aim to keep this backwards compatible and not make this a breaking change. Maybe add a config connection-type, default to http.

k-anshul:
> We should aim to keep this backwards compatible and not make this a breaking change. Maybe add a config connection-type, default to http.

k-anshul:
> Before merging this, can you investigate how many projects this would impact? If it impacts many, are there any heuristics or migrations we could write to maintain compatibility?

@nishantmonu51 suggested to fallback to http port when connection with native port failed and log a warning message. This will allow us to deploy without breaking change and we can migrate existing projects to use native port as per convenience.

begelundmuller (approved):

PR #6298, opened 2024-12-17
fix: allows for . in dimension names

mindspank:
Naive fix for allowing . in dimension names in the parser. We should look at which allowed characters we support and extend the parser syntax for it.


AdityaHegde:
Hmmm we escape the name in backend and not really restrict the names. We could do the same in UI filters as well. We should update our docs to reflect this.

AdityaHegde:
Updating this in a separate PR: https://github.com/rilldata/rill/pull/6301/files#diff-5a4f1e3856243633d83a9ed61a7d1a57f28d7791134fddecd29cdb527a75fa02R81
The above also handles `'` and `\n` in values.

mindspank:
Gotcha, closing in favor of your PR

PR #6303, opened 2024-12-18
e2e github actions

himadrisingh:


.github/workflows/web-test-e2e.yml
```diff
@@ -23,25 +26,45 @@ jobs:
...
+        id: matrix
+        run: |
+          echo "matrix=$(jq -n '[
+              "web-admin" | select(env.FILTER_ADMIN == "true" or env.FILTER_COMMON == "true"),
+              "web-local" | select(env.FILTER_LOCAL == "true" or env.FILTER_COMMON == "true")
+            ]' | jq -c)" >> "$GITHUB_OUTPUT"
+        env:
+          FILTER_ADMIN: ${{ steps.filter.outputs.admin }}
+          FILTER_COMMON: ${{ steps.filter.outputs.common }}
+          FILTER_LOCAL: ${{ steps.filter.outputs.local }}
```
ericpgreen2, line 58: It feels like these environment variables are adding an extra layer of indirection. Could we use `filter.outputs` directly in the `jq` command like: `select(${{ steps.filter.outputs.admin }} == "true" or ${{ steps.filter.outputs.common }} == "true")`?

ericpgreen2: Further, can we treat the filter output like a typical boolean, not a string, like: 
`select(${{ steps.filter.outputs.admin }} or ${{ steps.filter.outputs.common }})`
or even without the `select` function:
`if ${{ steps.filter.outputs.admin }} or ${{ steps.filter.outputs.common }} then "web-admin" else empty end`

ericpgreen2 (changes requested):
Nice! Functionality looks good. Just looks like there's one spot where the code could be simpler.

.github/workflows/web-test-e2e.yml
himadrisingh, line 58:
```suggestion
        run: |
          echo "matrix=$(jq -n '[
              "web-admin" | select(${{ steps.filter.outputs.admin }} == "true" or ${{ steps.filter.outputs.common }} == "true"),
              "web-local" | select(${{ steps.filter.outputs.local }} == "true" or ${{ steps.filter.outputs.common }} == "true")
            ]' | jq -c)" >> "$GITHUB_OUTPUT"
```

himadrisingh: yeah, it doesn't have access to `steps.filter.outputs`.

ericpgreen2: Are you sure? [This example](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#example-usage-of-the-steps-context) seems to show it's possible.

himadrisingh: i tried https://github.com/rilldata/rill/pull/6303/commits/f63fb070500ef086c8d8ecf654e8a4ba6f846202 but it didnt work out. probably its jq or single quote.

ericpgreen2: Gotcha

PR #6305, opened 2024-12-18
Intelligent default connectors for state and partition resolvers in models

begelundmuller:
Changes:
- If the `partitions:` resolver in a model is a `sql` resolver, it will default to the model's **input** connector
- If the `state:` resolver in a model is a `sql` resolver, it will default to the model's **output** connector

### DIFF
--- a/runtime/compilers/rillv1/parse_model.go
+++ b/runtime/compilers/rillv1/parse_model.go
@@ -61,43 +61,6 @@ func (p *Parser) parseModel(ctx context.Context, node *Node) error {
 		}
 	}
 
-	// Parse incremental state resolver
-	var incrementalStateResolver string
-	var incrementalStateResolverProps *structpb.Struct
-	if tmp.State != nil {
-		var refs []ResourceName
-		incrementalStateResolver, incrementalStateResolverProps, refs, err = p.parseDataYAML(tmp.State)
-		if err != nil {
-			return fmt.Errorf(`failed to parse "state": %w`, err)
-		}
-		node.Refs = append(node.Refs, refs...)
-	}
-
-	// Parse partitions resolver
-	var partitionsResolver string
-	var partitionsResolverProps *structpb.Struct
-	if tmp.Splits != nil { // Backwards compatibility: "splits" is deprecated and has been renamed to "partitions"
-		if tmp.Partitions != nil {
-			return fmt.Errorf(`"partitions" and "splits" are mutually exclusive`)
-		}
-		tmp.Partitions = tmp.Splits
-	}
-	if tmp.Partitions != nil {
-		var refs []ResourceName
-		partitionsResolver, partitionsResolverProps, refs, err = p.parseDataYAML(tmp.Partitions)
-		if err != nil {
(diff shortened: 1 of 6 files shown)


runtime/compilers/rillv1/parse_model.go
```diff
@@ -162,6 +125,43 @@ func (p *Parser) parseModel(ctx context.Context, node *Node) error {
...
+	var partitionsResolverProps *structpb.Struct
+	if tmp.Splits != nil { // Backwards compatibility: "splits" is deprecated and has been renamed to "partitions"
+		if tmp.Partitions != nil {
+			return fmt.Errorf(`"partitions" and "splits" are mutually exclusive`)
+		}
+		tmp.Partitions = tmp.Splits
+	}
+	if tmp.Partitions != nil {
+		var refs []ResourceName
+		partitionsResolver, partitionsResolverProps, refs, err = p.parseDataYAML(tmp.Partitions, inputConnector)
```
rohithreddykota, line 151: @begelundmuller I am trying to understand the rationale bind considering inputConnector as default. From my point of view, users may always want to pull the partition or state based on the current state of the output table data. It makes more sense to connect to the output table and get the latest available data (like timestamp) to derive the required partition or state and pull the source data based on derived partitions. Pls let me know if I am missing anything here.

begelundmuller: Hey Rohith, it's actually using the `outputConnector` as the default for the `state:` resolver (since that will often be something like `max(updated_on)` on the output). But I picked `inputConnector` for the `partitions:` resolver because I figured people would usually discover partitions by inspecting the input data (like `SELECT DISTINCT partition_key FROM source_table`).

Does that make sense? Or is there something I'm not thinking of?

rohithreddykota: Regarding `state`, it is clear that the default must be the outputConnector. I see your point with `partitions` and I agree with your argument too. 

I imagine the partitions generation as a continuous incremental process in Rill Cloud, which runs at the scheduled time. Let's take a scenario where the start of source data partitions aren't the end of output data. There can be a data gap. In this case, I would think to pull the partitions from the source based on the max(timestamp) of the output data. This approach will always pulling relevant data and avoiding potential gaps between the source and output data. Also by defaulting both connectors to outputConnector, user wouldn't need to learn which connector will be used be default in each situation.

begelundmuller: > I would think to pull the partitions from the source based on the max(timestamp) of the output data

Could a solution here be to allow you to reference `{{ .state.max_time }}` from inside the `partitions:` SQL? Or are you thinking about something else here?

rohithreddykota: Okay, now I understand what you are thinking. It totally makes sense as we can use the state template in the partitions SQL. This approach would allow us to reference` {{ .state.max_time }}` directly within the partitions SQL, ensuring that we are always pulling partitions based on the latest state of the output data. Thank you for the clarification!

rohithreddykota (approved):

PR #6308, opened 2024-12-19
fix: add accessible labels for workspace toggles

briangregoryholmes:
Adds the ability to select these buttons using `page.getByLabel("Toggle table visibility")` and `page.getByLabel("Toggle inspector visibility")` when constructing Playwright tests


royendo (approved):
Ty! LGTM

PR #6317, opened 2024-12-20
fix: redirect loop when home bookmark has missing compare dimension

AdityaHegde:


ericpgreen2 (approved):
I'll approve, though this makes the component even more complex. I'd encourage you to ask AI for suggestions on how to simplify all this code.

AdityaHegde:
Sounds good. Lets do a simplification follow up

PR #6326, opened 2024-12-23
feat: Add new rill time parser backend support

AdityaHegde:
- [x] Add module for parsing rill time syntax in backend
- [x] Support syntax like, `2024-12-15+5d,2024-12-15-5d`
- [x] Add `ResolveTime` to return start and end times based on minTime, maxTime and now for a metrics view
- [x] Replace usage of old duration parser
- [x] Map old format to new format
- [x] Add API to resolve times for a list of rill-times
- [ ] Add a lot more tests
- [ ] ~~Update `MetricsViewTimeSeries` to use `TimeRange` and make use of `ResolveTime`~~ We will just use the MetricsViewAggregation in a future PR. If it is not ready then this can be added.

Note that this doesnt add support for rill-time in alerts and report intervals (The underlying query is support in this PR). That will be added in a future PR.


runtime/pkg/metricssql/time_range_parser.go
```diff
@@ -102,7 +116,38 @@ func (q *query) getWatermark(ctx context.Context, colName string) (watermark tim
 	return watermark, nil
 }
 
-func parseTimeRangeArgs(args []ast.ExprNode) (duration.Duration, int, string, error) {
+func (q *query) getMinTime(ctx context.Context, colName string) (time.Time, error) {
```
k-anshul, line 119: Nite : Can directly use `metricsview.Executor.GetMinTime` ? The existing code does not use `watermark` from `Executor` because the implementations are somewhat different.

runtime/pkg/metricssql/time_range_parser.go
```diff
@@ -102,7 +116,38 @@ func (q *query) getWatermark(ctx context.Context, colName string) (watermark tim
...
+
+	sql := fmt.Sprintf("SELECT MIN(%s) FROM %s", olap.Dialect().EscapeIdentifier(colName), olap.Dialect().EscapeTable(q.metricsViewSpec.Database, q.metricsViewSpec.DatabaseSchema, q.metricsViewSpec.Table))
+	result, err := olap.Execute(ctx, &drivers.Statement{Query: sql, Priority: q.priority})
+	if err != nil {
+		return time.Time{}, err
+	}
+	defer result.Close()
+
+	var t time.Time
+	for result.Next() {
```
k-anshul, line 142: Also need to check for `result.Err()`

runtime/pkg/metricssql/time_range_parser.go
```diff
@@ -10,11 +10,11 @@ import (
 	"github.com/pingcap/tidb/pkg/parser/ast"
 	"github.com/rilldata/rill/runtime/drivers"
 	"github.com/rilldata/rill/runtime/metricsview"
-	"github.com/rilldata/rill/runtime/pkg/duration"
+	"github.com/rilldata/rill/runtime/pkg/rilltime"
 )
 
 func (q *query) parseTimeRangeStart(ctx context.Context, node *ast.FuncCallExpr) (*metricsview.Expression, error) {
-	d, unit, colName, err := parseTimeRangeArgs(node.Args)
+	rt, unit, colName, err := parseTimeRangeArgs(node.Args)
```
k-anshul, line 17: More of a product call but given the unit can now be expressed with the new time range syntax, may be we should just omit the unit from `time_range_start` and `time_range_end` definition.
I don't think they are used externally so should be okay to make a backward incompatible change.

k-anshul (changes requested):

runtime/compilers/rillv1/parse_explore.go
```diff
@@ -165,18 +166,26 @@ func (p *Parser) parseExplore(node *Node) error {
...
+			isNewFormat := false
 			if ctr.Range != "" {
-				if err := validateISO8601(ctr.Range, false, false); err != nil {
+				rt, err := rilltime.Parse(ctr.Range)
+				if err != nil {
 					return fmt.Errorf("invalid comparison range %q: %w", ctr.Range, err)
 				}
+				isNewFormat = rt.IsNewFormat
+			}
+			if ctr.Offset != "" {
```
begelundmuller, line 182: Consider having a `rilltime.ParseCompatibility(ctr.Range, ctr.Offset)` function – this would avoid duplicating the error handling if `isoOffset` is combined with the new format, and also avoid exposing the `IsNewFormat` outside the package

runtime/metricsview/executor_rewrite_time.go
```diff
@@ -8,9 +8,41 @@ import (
 
 	"github.com/rilldata/rill/runtime/drivers"
 	"github.com/rilldata/rill/runtime/pkg/duration"
+	"github.com/rilldata/rill/runtime/pkg/rilltime"
 	"github.com/rilldata/rill/runtime/pkg/timeutil"
 )
 
+func (e *Executor) GetMinTime(ctx context.Context, colName string) (time.Time, error) {
```
begelundmuller, line 15: 1. Can drop the `Get`, so just `MinTime`, similar to `Watermark`.
2. All the executors exported functions should be in `runtime/metricsview/executor.go` file – this probably belongs next to the `Watermark` function.
3. Why does this accept a column name? We currently only support one time dimension for metrics views.

runtime/metricsview/executor_rewrite_time.go
```diff
@@ -41,6 +73,47 @@ func (e *Executor) resolveTimeRange(ctx context.Context, tr *TimeRange, tz *time
 		return nil
 	}
 
+	if tr.RillTime == "" {
+		return e.resolveTimeRangeLegacy(ctx, tr, tz, executionTime)
+	}
+
+	rt, err := rilltime.Parse(tr.RillTime)
```
begelundmuller, line 80: Can we avoid calling this `rt` – we often use that abbreviation for the `Runtime` object, e.g. `e.rt` in this context is the `Runtime` object. So it's a bit too confusing.

runtime/metricsview/executor_rewrite_time.go
```diff
@@ -41,6 +73,47 @@ func (e *Executor) resolveTimeRange(ctx context.Context, tr *TimeRange, tz *time
 		return nil
 	}
 
+	if tr.RillTime == "" {
+		return e.resolveTimeRangeLegacy(ctx, tr, tz, executionTime)
+	}
```
begelundmuller, line 78: This gets called if `tr.Start` or `tr.End` are set – I wouldn't really consider that legacy behavior (which entails we will be deprecating it), it's something we will probably always support

runtime/metricsview/executor_rewrite_time.go
```diff
@@ -41,6 +73,47 @@ func (e *Executor) resolveTimeRange(ctx context.Context, tr *TimeRange, tz *time
...
+
+	rt, err := rilltime.Parse(tr.RillTime)
+	if err != nil {
+		return err
+	}
+
+	t, err := e.loadWatermark(ctx, executionTime)
+	if err != nil {
+		return err
+	}
```
begelundmuller, line 88: `t` is very ambiguous now, probably call it `watermark`

runtime/metricsview/executor_rewrite_time.go
```diff
@@ -41,6 +73,47 @@ func (e *Executor) resolveTimeRange(ctx context.Context, tr *TimeRange, tz *time
...
+
+	minTime, err := e.GetMinTime(ctx, e.metricsView.TimeDimension)
+	if err != nil {
+		return err
+	}
+
+	tr.Start, tr.End, err = rt.Resolve(rilltime.ResolverContext{
+		Now:        time.Now(),
+		MinTime:    minTime,
+		MaxTime:    t,
```
begelundmuller, line 98: The watermark is not necessarily `MaxTime` – it may be, but users can override it to something else (e.g. to avoid partial data). 

The design doc says to differentiate between the `earliest`, `latest`, `now` and `watermark` constants: https://www.notion.so/rilldata/Time-range-and-time-range-series-syntax-0be92cca367a425682bb5ea09032b260?pvs=4#131ba33c8f578010b792cac30273e818

runtime/metricsview/executor_rewrite_time.go
```diff
@@ -41,6 +73,47 @@ func (e *Executor) resolveTimeRange(ctx context.Context, tr *TimeRange, tz *time
...
+
+	t, err := e.loadWatermark(ctx, executionTime)
+	if err != nil {
+		return err
+	}
+
+	minTime, err := e.GetMinTime(ctx, e.metricsView.TimeDimension)
+	if err != nil {
+		return err
+	}
```
begelundmuller, line 93: It feels a bit heavy to run two queries to obtain these values, and if the watermark is not equal to the max time, it will need to run three queries. Since I believe most expressions can be resolved using only one constant, can these be lazily loaded instead of eagerly?

runtime/metricsview/executor_rewrite_time.go
begelundmuller: Actually it also doesn't handle the error case where e.g. both `Start` and `RillTime` is specified

runtime/metricsview/query.go
```diff
@@ -134,13 +134,14 @@ type Sort struct {
 type TimeRange struct {
 	Start        time.Time `mapstructure:"start"`
 	End          time.Time `mapstructure:"end"`
+	RillTime     string    `mapstructure:"rill_time"`
```
begelundmuller, line 137: As per the other comment, I would suggest calling this `Expression`

runtime/pkg/metricssql/parser.go
```diff
@@ -160,6 +163,17 @@ func (q *query) parseFrom(ctx context.Context, node *ast.TableRefsClause) error
...
+	security, err := q.controller.Runtime.ResolveSecurity(q.instanceID, q.claims, mv)
+	if err != nil {
+		return fmt.Errorf("metrics sql: failed to resolve security: %w", err)
+	}
+
+	ex, err := metricsview.NewExecutor(ctx, q.controller.Runtime, q.instanceID, q.metricsViewSpec, false, security, q.priority)
+	if err != nil {
+		return fmt.Errorf("metrics sql: failed to create executor: %w", err)
+	}
+	q.executor = ex
```
begelundmuller, line 175: Can we avoid needing an `Executor` in this package? It is ideally just supposed to parse the SQL and rewrite it into a `metricsview.Query`, which can accept the raw `rilltime` expression

proto/rill/runtime/v1/queries.proto
```diff
@@ -153,6 +153,13 @@ service QueryService {
     };
   }
 
+  rpc MetricsViewResolveTimeRanges(MetricsViewResolveTimeRangesRequest) returns (MetricsViewResolveTimeRangesResponse) {
```
begelundmuller, line 156: Can we call this `MetricsViewTimeRanges` – to align with format of the other `MetricsViewXXX` API names

runtime/server/queries_metrics.go
```diff
@@ -401,6 +401,47 @@ func (s *Server) MetricsViewSearch(ctx context.Context, req *runtimev1.MetricsVi
...
+		ResolvedMVSecurity: security,
+	}
+	err = s.runtime.Query(ctx, req.InstanceId, timeRangeQuery, int(req.Priority))
+	if err != nil {
+		return nil, err
+	}
+
+	q := &queries.MetricsViewResolveTimeRanges{
+		MetricsViewName: req.MetricsViewName,
+		MinTime:         timeRangeQuery.Result.TimeRangeSummary.Min.AsTime(),
```
begelundmuller, line 433: It's weird that it resolves `MinTime` separately here, but it resolves the watermark/max time inside the `MetricsViewResolveTimeRanges` implementation.

I would expect a) that it generally resolves these timestamps the same way in most places, b) that it resolves these timestamps at the same place / subject to the same caching.

runtime/pkg/rilltime/rilltime_test.go
```diff
@@ -0,0 +1,73 @@
...
+func Test_Resolve(t *testing.T) {
+	now := parseTestTime(t, "2024-08-09T10:32:36Z")
+	maxTime := parseTestTime(t, "2024-08-06T06:32:36Z")
+	testCases := []struct {
+		timeRange string
+		start     string
+		end       string
+	}{
+		// Earliest = 2023-08-09T10:32:36Z, Latest = 2024-08-06T06:32:36Z, = Now = 2024-08-09T10:32:36Z
+		{`m : |s|`, "2024-08-09T10:32:00Z", "2024-08-09T10:32:36Z"},
```
begelundmuller, line 19: In the (resolved) comments on the doc, we discussed open intervals that end on a time boundary, but I don't think there's a test case for that here. Can you add it? I think an example could be:
```
		{`m : s`, "2024-08-09T10:32:00Z", "2024-08-09T10:32:37Z"},
```

runtime/pkg/rilltime/rilltime.go
```diff
@@ -0,0 +1,372 @@
...
+		"d": timeutil.TimeGrainDay,
+		"D": timeutil.TimeGrainDay,
+		"W": timeutil.TimeGrainWeek,
+		"Q": timeutil.TimeGrainQuarter,
+		"M": timeutil.TimeGrainMonth,
+		"Y": timeutil.TimeGrainYear,
+	}
+)
+
+type RillTime struct {
```
begelundmuller, line 71: Instead of `rilltime.RillTime`, consider calling this `rilltime.Expression`

runtime/pkg/rilltime/rilltime.go
```diff
@@ -0,0 +1,372 @@
...
+			rt.Start.Grain.Grain = "s"
+		default:
+			return nil, fmt.Errorf("unexpected field %q in duration", name)
+		}
+	}
+
+	return rt, nil
+}
+
+func (r *RillTime) Resolve(resolverCtx ResolverContext) (time.Time, time.Time, error) {
```
begelundmuller, line 227: This is a matter of convention, but we generally use "resolve" for something that makes database queries and "context" for `context.Context` objects. Could we call this `Eval(opts EvalOptions)` or something like that instead?

begelundmuller (changes requested):

proto/rill/runtime/v1/queries.proto
```diff
@@ -551,6 +558,9 @@ message TimeRange {
   TimeGrain round_to_grain = 5;
   // Optional. IANA format, ie Europe/Copenhagen. Defaults to UTC
   string time_zone = 6;
+  // Optional. Rill format time range. Should only be used for alerts and reports.
+  // Resolve the times and use `start` and `end` while querying in UI to keep time consistent across queries.
+  string rill_time = 7;
```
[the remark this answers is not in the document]
AdityaHegde: Will we ever have other expression similar to watermark? If so would it be better to have a separate param for this: `rill_expression` or something?

runtime/metricsview/executor_rewrite_time.go
AdityaHegde: Huh. I thought according Mike we only used `earliest` and `latest`. Lemme add watermark as a separate thing.

runtime/pkg/metricssql/parser.go
AdityaHegde: This was as per @k-anshul request in his review. @k-anshul any comments?

runtime/pkg/rilltime/rilltime_test.go
AdityaHegde: Wouldnt it be `2024-08-09T10:32:00Z` to `2024-08-09T10:32:36Z`?

runtime/pkg/rilltime/rilltime.go
AdityaHegde: Changed. So `resolveTimeRange` was called `resolveXX` just because it loaded the watermark from database?

runtime/pkg/metricssql/parser.go
k-anshul: The parser is already doing some OLAP queries to query watermark so it is not a "typical" parser pkg.
Given we are already making OLAP queries it should be okay to use `Executor` as well since the same function is already part of executor.
Might be wrong but IIRC there was a review comment to use executor's `loadWatermark` instead of defining another `getWatermark` function here : https://github.com/rilldata/rill/blob/fef2386bb980101ebbdd9f7c5ab2b41c3e0d60b4/runtime/pkg/metricssql/time_range_parser.go#L70
I could not use it previously because the implementation is slightly different.

proto/rill/runtime/v1/queries.proto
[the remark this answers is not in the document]
begelundmuller: I think we are planning on leaning into this time syntax across the product, so I think it's to be expected that `time_range.expression` would use our time range syntax (and of course, the docstring should mention that it uses our own syntax).

If we need to keep separate fields for ISO or SQL syntax (which I hope not), those in turn could be called `iso_expression` and `sql_expression`. But putting a `rill_` prefix IMO does not disambiguate much because the product and package name here are both called Rill already.

runtime/pkg/rilltime/rilltime_test.go
begelundmuller: I believe it should be `...:37Z` – there was a long comment thread about it in the design doc, which seems to have been resolved without updating the main document. Here's one of the relevant comments (to find it, change the comment filter to include resolved comments in Notion):

Screenshot 2025-01-10 at 11 16 44

runtime/pkg/rilltime/rilltime.go
begelundmuller: Yeah

runtime/pkg/metricssql/parser.go
begelundmuller: I see, in that case I agree that the executor should be used to get the watermark so there's not too much duplication.

I'm not sure if it's possible, but it would still be even better if the SQL time expressions could just be translated to the new time range syntax and then passed into the `metricsview.Query` instead of being resolved inside the parser. However, that's not urgent or a blocker for this PR.

runtime/compilers/rillv1/parse_explore.go
AdityaHegde: Update. Had to move `ValidateISO8601` to achieve this.

runtime/metricsview/executor_rewrite_time.go
AdityaHegde: Update

runtime/metricsview/executor_rewrite_time.go
AdityaHegde: Update for 1 and 2. For 3, we are currently reusing the method for `time_range_start` and `time_range_end` which accepts a column name.

runtime/server/queries_metrics.go
AdityaHegde: Will raise a follow up to use the metrics_time_range resolver.

runtime/metricsview/executor_rewrite_time.go
AdityaHegde: I think lazy load will make it a little complex. Will do a follow up that will use metrics_time_range resolver so that these things are cached.

begelundmuller:
Awaiting integration of this PR before the next review: https://github.com/rilldata/rill/pull/6413#pullrequestreview-2550332129

runtime/queries/metricsview_resolve_time_ranges.go
```diff
@@ -0,0 +1,108 @@
...
+	"io"
+	"time"
+
+	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
+	"github.com/rilldata/rill/runtime"
+	"github.com/rilldata/rill/runtime/pkg/rilltime"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+type MetricsViewTimeRanges struct {
```
begelundmuller, line 16: This basically just a) calls the timestamps resolver (which may be cached in the resolver cache), b) calls `rilltime`, which doesn't involve heavy operations and should be fast. So can the logic here just be moved directly to the API handler for `MetricsViewTimeRanges`? Since there is no need for separate caching of the result here.

It's also nice not to add new logic to the `runtime/queries` package since it's being deprecated.

runtime/pkg/rilltime/rilltime.go
```diff
@@ -0,0 +1,394 @@
...
+		var err error
+		rt.timeZone, err = time.LoadLocation(strings.Trim(*rt.AtModifiers.TimeZone, "{}"))
+		if err != nil {
+			return nil, fmt.Errorf("invalid time zone %q: %w", *rt.AtModifiers.TimeZone, err)
+		}
+	}
+
+	if rt.End == nil {
+		rt.End = &TimeAnchor{
+			Now: true,
```
begelundmuller, line 161: Shouldn't the watermark be the default end if not specified?

runtime/pkg/rilltime/rilltime.go
```diff
@@ -0,0 +1,394 @@
...
+
+	if rt == nil {
+		rt, err = rillTimeParser.ParseString("", from)
+		if err != nil {
+			return nil, err
+		}
+		rt.isNewFormat = true
+	}
+
+	rt.timeZone = time.UTC
```
begelundmuller, line 138: Can we add support for passing in the default time zone? E.g. we have had some requests to use the same time zone as the `refresh:` clause when evaluating and formatting data for alerts and reports.

runtime/pkg/rilltime/rilltime_test.go
```diff
@@ -0,0 +1,82 @@
...
+	maxTime := parseTestTime(t, "2024-08-06T06:32:36Z")
+	watermark := parseTestTime(t, "2024-08-05T06:32:36Z")
+	testCases := []struct {
+		timeRange string
+		start     string
+		end       string
+	}{
+		// Earliest = 2023-08-09T10:32:36Z, Latest = 2024-08-06T06:32:36Z, = Now = 2024-08-09T10:32:36Z
+		{`m : |s|`, "2024-08-09T10:32:00Z", "2024-08-09T10:32:37Z"},
+		{`m : s`, "2024-08-09T10:32:00Z", "2024-08-09T10:32:37Z"},
```
begelundmuller, line 21: Shouldn't it be:
```
		{`m : |s|`, "2024-08-09T10:32:00Z", "2024-08-09T10:32:36Z"},
```
like it was before? I.e. `|s|` leads to `:36Z` and `s` leads to `:37Z`.

The idea being that `|s|` indicates that only completed bins should be included – but if `maxTime=...:36Z`, more events may yet appear with `time=...:36Z`. And callers who are good with bins that have potentially partial data should use `m : s`.

begelundmuller (changes requested):

runtime/pkg/rilltime/rilltime_test.go
AdityaHegde: Hmmm. So the idea is that our bins are `ts >= 0 and ts < next 0`? So a data point at the boundary is actually part of next grain. That is why `|s|` will not include it since it is incomplete. We should make sure we explain all these things in our docs, will note it down.

runtime/pkg/rilltime/rilltime.go
AdityaHegde: Lemme check with mike today. The docs mention that it should be now. But it probably make sense to do latest/watermark.

runtime/pkg/rilltime/rilltime_test.go
begelundmuller: Yeah that's the idea!

runtime/pkg/rilltime/rilltime.go
AdityaHegde: Got a confirmation from @nishantmonu51 to use watermark.

begelundmuller (approved):

k-anshul (approved):

PR #6327, opened 2024-12-23
Backend support for custom logos

begelundmuller:
- You can upload a custom logo using `rill org upload-logo`.
- A URL for the custom logo is returned in `organization.logo_url` in the `GetOrganization` RPC (empty if not configured).
- The custom logo's URL will always have the same host as the `admin` service (so we should already be good for CSP).
- For implementing UI uploads, see `cli/cmd/org/upload_logo.go` for an example of the network calls to make.

Contributes to https://github.com/rilldata/rill-private-issues/issues/962.


cli/cmd/org/upload_logo.go
```diff
@@ -0,0 +1,150 @@
...
+			}
+			defer f.Close()
+
+			// Confirmation prompt
+			if ok, err := cmdutil.ConfirmPrompt(fmt.Sprintf("You are changing the custom logo for %q. Continue?", ch.Org), "", false); err != nil || !ok {
+				return err
+			}
+
+			// Generate the asset upload URL
+			asset, err := client.CreateAsset(cmd.Context(), &adminv1.CreateAssetRequest{
```
pjain1, line 101: can you not follow the network calls to upload a logo of size less than 100 MB (signed url header limit)

begelundmuller: I just intended it as a convenience to avoid uploading an image that would load very slowly, but I agree it's probably safer and more future proof to enforce the size on the backend. I pushed a change that enforces different upload limits on the backend depending on whether it's a deploy archive or an image.

proto/rill/admin/v1/api.proto
```diff
@@ -1184,6 +1184,7 @@ message CreateAssetRequest {
   string name = 3 [(validate.rules).string.min_len = 1]; // Eg. "rilldata__project_name"
   string extension = 4; // Eg. ".tar.gz"
   bool cacheable = 5;
+  int64 estimated_size_bytes = 6;
```
pjain1, line 1187: Now this needs to be set while uploading project assets as well I don't see that change

begelundmuller: It is optional – if not set, it will just be enforced during the upload itself (which was the previous behavior). Just added it here to error early with a nicer error message.

It is not quite a trivial refactor for the project upload use cases, so adding a separate backlog issue to make the refactor there :)

pjain1: OK I think its fine then

pjain1 (approved):

proto/rill/admin/v1/api.proto
pjain1: don't think its really necessary to add fail early check

PR #6324, opened 2024-12-21
upgrade snowflake, bigquery and arrow version

k-anshul:
Hoping to fix the memory leak observed in projects using snowflake connector. 
The new versions fixes a memory leak : https://github.com/snowflakedb/gosnowflake/pull/1270

PR #6330, opened 2024-12-24
fix : patch snowflake driver for memory fix

k-anshul:
This relates to https://github.com/rilldata/rill/pull/6324
Forgot to patch the snowflake driver after upgrading :(


pjain1 (approved):

k-anshul:
Thanks for the review. 
Well as it turns out the above issue does not have anything to do with the leaks that we are seeing. We do not hit that code path. We do not decode the arrow records and directly pass it to parquet converter.
But it seems likely that leak is somewhere in the snowflake code path since projects using snowflake connector seems to be getting OOMs. I will take a deeper look at the library.

Issue #6331, opened 2024-12-25
with Rill 0.52.2 on Linux AMD64 getting a 'conflicting Schema URL' error


jaywgraves:
**Describe the bug**
installing Rill 0.52.2 from a downloaded/unzipped release

```shell
rill start example_project 
Error: 1 errors occurred detecting resource:
	* conflicting Schema URL: https://opentelemetry.io/schemas/1.25.0 and https://opentelemetry.io/schemas/1.26.0
```

**To Reproduce**
1. Download 0.52.2 for Linux AMD64
2. unzip to get `rill` binary
3. type `rill start example_project`
```shell
Error: 1 errors occurred detecting resource:
	* conflicting Schema URL: https://opentelemetry.io/schemas/1.25.0 and https://opentelemetry.io/schemas/1.26.0
```

**Expected behavior**
`rill start example_project`
does not error

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Ubuntu 22.04.1 LTS
 - Browser: N/A  (never launches browser
 - Version:  N/A
 
**Additional context**
Downgrading to 0.52.1 works.
#6324 bumps a bunch of deps, maybe something got missed?

k-anshul:
Hey @jaywgraves 

Thanks for reporting. We merged a fix for the issue. The fix should be available in the next release. 
In case you want to try it out early you can use upgrade to nightly with `rill upgrade --nightly` after the next run (runs at midnight UTC).

PR #6332, opened 2024-12-26
fix : resolve conflict schema URL error

k-anshul:
closes https://github.com/rilldata/rill/issues/6331

On a longer term we need to investigate on how this can be caught earlier. Didn't see either in prod servers or on my laptotp.


pjain1:
Were not able to reproduce locally, how can we test that this resolves the issue, merging is ok but just asking.

pjain1 (approved):

k-anshul:
I tested on my linux laptop.

PR #6333, opened 2024-12-27
Emit ClickHouse database size metrics

begelundmuller:
- Adds a metric event named `clickhouse_estimated_size_bytes` giving the estimated disk size used by ClickHouse (limited to the tables the current user has access to).
- Adds an attribute `managed: true` to all events emitted from a Rill-managed ClickHouse connector. This enables billing logic to bill differently for externally managed ClickHouse connections.
- Adds last-used-on logic to the ClickHouse driver, which is used to only estimate the size when the database was recently used. This ensures we play nice with scale-to-zero ClickHouse.
- Adds a new grant to Rill-managed ClickHouse users to enable reporting disk usage.
- Adds support for configuring a `console` activity logger on the runtime for debugging activity event emission.

Contributes to https://github.com/rilldata/rill-private-issues/issues/1002

### DIFF
--- a/runtime/drivers/clickhouse/clickhouse.go
+++ b/runtime/drivers/clickhouse/clickhouse.go
@@ -97,7 +97,10 @@ var maxOpenConnections = 20
 type driver struct{}
 
 type configProperties struct {
-	// Provision is set on local if `managed: true` is set for the connector.
+	// Managed is set internally if the connector has `managed: true`.
+	Managed bool `mapstructure:"managed"`
+	// Provision is set when Managed is true and provisioning should be handled by this driver.
+	// (In practice, this gets set on local and means we should start an embedded Clickhouse server).
 	Provision bool `mapstructure:"provision"`
 	// DSN is the connection string. Either DSN can be passed or the individual properties below can be set.
 	DSN      string `mapstructure:"dsn"`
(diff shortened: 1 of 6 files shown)


runtime/drivers/clickhouse/clickhouse.go
```diff
@@ -402,3 +425,69 @@ func (c *connection) AsNotifier(properties map[string]any) (drivers.Notifier, er
...
+			}
+		case <-c.ctx.Done():
+			return
+		}
+	}
+}
+
+// estimateSize returns the estimated combined disk size of all resources in the database in bytes.
+func (c *connection) estimateSize(ctx context.Context) (int64, error) {
+	rows, err := c.db.QueryxContext(ctx, `SELECT sum(bytes_on_disk) AS size FROM system.parts WHERE (active = 1) AND lower(database) NOT IN ('information_schema', 'system')`)
```
k-anshul, line 476: Nit: Can use `QueryRowxContext` to keep the scanning and error handling simpler ?

runtime/drivers/clickhouse/clickhouse.go
```diff
@@ -402,3 +425,69 @@ func (c *connection) AsNotifier(properties map[string]any) (drivers.Notifier, er
...
+			// Skip if it hasn't been used recently and may be scaled to zero.
+			if c.config.CanScaleToZero && time.Since(c.lastUsedOn()) > 2*d {
+				continue
+			}
+
+			// Emit the estimated size of the database.
+			size, err := c.estimateSize(c.ctx)
+			if err == nil {
+				c.activity.RecordMetric(c.ctx, "clickhouse_estimated_size_bytes", float64(size))
+			} else if !errors.Is(err, c.ctx.Err()) {
```
k-anshul, line 465: What if we are querying external tables and the user does not have access to query `system.parts` ?

k-anshul (commented):
Added a question and a nit.
I am assuming this will be mostly used for rill managed databases since for external databases this can report a big number across tables even if no rill dashboard exists on top of it.

runtime/drivers/clickhouse/clickhouse.go
begelundmuller: Done!

runtime/drivers/clickhouse/clickhouse.go
begelundmuller: Good catch! I updated it to only log at `error` level (which triggers alerts) for Clickhouse connectors that are managed by Rill.

k-anshul (approved):

PR #6337, opened 2024-12-30
2 phase comparison query

pjain1:
Split comparison time range query into two phase -
1. Run a groupby query on base time range
2. Run another query which joins the inline results from first query with groupby comparison query having added filter from base query
    - This optimization is only possible when sorting by base column. (can extend to sort by comparison column but UI does not seem to use that anywhere)
    - CTE rewrite is disabled by default now as it cause double scan of CTE in Druid, Duckdb and CH. It can be enabled but does not take effect if 2 phase optimization is done. It can be enabled to run queries on very high cardinality (~1M or greater) when sorting by other than base column which may fail in engines like Druid with ResourceLimitException.

Side note
- May be I can reuse the inline select concept for the null filling PR, also figured that druid support inline datasource with alternative syntax so that can be used and partial support with context keys can be removed

### DIFF
--- a/runtime/drivers/olap.go
+++ b/runtime/drivers/olap.go
@@ -500,8 +500,8 @@ func (d Dialect) DateDiff(grain runtimev1.TimeGrain, t1, t2 time.Time) (string, 
 	}
 }
 
-// SelectInlineResults returns a SQL query which inline results from the result set supplied.
-func (d Dialect) SelectInlineResults(result *Result) (string, []any, error) {
+// SelectInlineResults returns a SQL query which inline results from the result set supplied along with the positional arguments and dimension values.
+func (d Dialect) SelectInlineResults(result *Result) (string, []any, []any, error) {
 	values := make([]any, len(result.Schema.Fields))
 	valuePtrs := make([]any, len(result.Schema.Fields))
 	for i := range values {
@@ -509,6 +509,7 @@ func (d Dialect) SelectInlineResults(result *Result) (string, []any, error) {
 	}
 
 	var dimVals []any
+	var args []any
 
 	rows := 0
 	prefix := ""
@@ -516,7 +517,7 @@ func (d Dialect) SelectInlineResults(result *Result) (string, []any, error) {
 	// creating inline query for all dialects in one loop, accumulating field exprs first and then creating the query can be more cleaner
 	for result.Next() {
 		if err := result.Scan(valuePtrs...); err != nil {
-			return "", nil, fmt.Errorf("select inline: failed to scan value: %w", err)
+			return "", nil, nil, fmt.Errorf("select inline: failed to scan value: %w", err)
 		}
 		if d == DialectDruid {
 			// format - select * from (values (1, 2), (3, 4)) t(a, b)
@@ -551,17 +552,19 @@ func (d Dialect) SelectInlineResults(result *Result) (string, []any, error) {
 			if i > 0 {
 				prefix += ", "
 			}
-			ok, expr, err := d.GetValExpr(v, result.Schema.Fields[i].Type.Code)
-			if err != nil {
-				return "", nil, fmt.Errorf("select inline: failed to get value expression: %w", err)
-			}
-			if !ok {
-				return "", nil, fmt.Errorf("select inline: unsupported value type %q: %w", result.Schema.Fields[i].Type.Code, ErrOptimizationFailure)
-			}
+
 			if d == DialectDruid {
+				ok, expr, err := d.GetValExpr(v, result.Schema.Fields[i].Type.Code)
+				if err != nil {
+					return "", nil, nil, fmt.Errorf("select inline: failed to get value expression: %w", err)
+				}
+				if !ok {
+					return "", nil, nil, fmt.Errorf("select inline: unsupported value type %q: %w", result.Schema.Fields[i].Type.Code, ErrOptimizationFailure)
+				}
 				prefix += expr
 			} else {
-				prefix += fmt.Sprintf("%s AS %s", expr, d.EscapeIdentifier(result.Schema.Fields[i].Name))
+				prefix += fmt.Sprintf("%s AS %s", "?", d.EscapeIdentifier(result.Schema.Fields[i].Name))
+				args = append(args, v)
 			}
 		}
 
@@ -579,7 +582,7 @@ func (d Dialect) SelectInlineResults(result *Result) (string, []any, error) {
 		prefix += ") "
 	}
 
-	return prefix + suffix, dimVals, nil
+	return prefix + suffix, args, dimVals, nil
 }
 
 func (d Dialect) GetValExpr(val any, typ runtimev1.Type_Code) (bool, string, error) {
--- a/runtime/drivers/olap.go
+++ b/runtime/drivers/olap.go
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"math"
 	"strings"
 	"time"
 
@@ -16,8 +17,12 @@ import (
 	_ "time/tzdata"
 )
 
-// ErrUnsupportedConnector is returned from Ingest for unsupported connectors.
-var ErrUnsupportedConnector = errors.New("drivers: connector not supported")
+var (
+	// ErrUnsupportedConnector is returned from Ingest for unsupported connectors.
+	ErrUnsupportedConnector = errors.New("drivers: connector not supported")
+	// ErrOptimizationFailure is returned when an optimization fails.
+	ErrOptimizationFailure = errors.New("drivers: optimization failure")
+)
 
 // WithConnectionFunc is a callback function that provides a context to be used in further OLAP store calls to enforce affinity to a single connection.
 // It also provides pointers to the actual database/sql and database/sql/driver connections.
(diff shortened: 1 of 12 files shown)


pjain1:
Another approach is to do two separate topn queries for base and comparison time range and then a 3rd query with results from both queries to calculate measures and return the results, have a working code for this approach as well in a branch. However, this seems more cleaner and does only 2 queries.

runtime/drivers/registry.go
```diff
@@ -136,6 +140,8 @@ func (i *Instance) Config() (InstanceConfig, error) {
 		ModelDefaultMaterialize:           false,
 		ModelMaterializeDelaySeconds:      0,
 		MetricsApproximateComparisons:     true,
+		MetricsApproximateComparisonsCTE:  false,
+		MetricsTwoPhaseComparisons:        true,
```
begelundmuller, line 144: Would it make sense to differentiate between queries with a small limit and (unlimited) queries like exports? Since I guess the two-phase approach will not be suitable for high cardinality results?

If yes, do you think it would be best to keep falling back to the CTE approach for unlimited queries, or should it just fall back to the non-CTE approach?

runtime/metricsview/ast.go
```diff
@@ -43,6 +43,9 @@ type SelectNode struct {
 	IsCTE                bool             // Whether this node is a Common Table Expression
 	DimFields            []FieldNode      // Dimensions fields to select
 	MeasureFields        []FieldNode      // Measure fields to select
+	IsInline             bool             // Whether this node is an inline view
+	InlineDimFields      [][]FieldNode    // Dimension fields to select in the inline view, row based representation of the dimensions. Each index in the outer array represents a row in the inline view.
+	InlineMeasureFields  [][]FieldNode    // Measure fields to select in the inline view, row based representation of the measures. Each index in the outer array represents a row in the inline view.
```
begelundmuller, line 48: Based on our sync discussion, would it make sense instead to:
1. Support a field like: `RawSelect *ExprNode // Raw SELECT statement with args. If set, other fields like FromSelect, Where, etc. are ignored.`
2. Populating `DimFields` and `MeasureFields` with the columns returned by the `RawSelect` subquery (instead of having separate `InlineXXX` variants)

(This is just a suggestion, not sure if I have enough context yet to determine if it's the best approach.)

runtime/metricsview/ast.go
```diff
@@ -623,6 +628,70 @@ func (a *AST) buildUnderlyingWhere() (*ExprNode, error) {
...
+					if a.dialect == drivers.DialectClickHouse {
+						f.Expr = fmt.Sprintf("parseDateTimeBestEffort('%s')", v.Format(time.RFC3339))
+					} else {
+						f.Expr = fmt.Sprintf("CAST('%s' AS TIMESTAMP)", v.Format(time.RFC3339))
+					}
+				case bool:
+					f.Expr = fmt.Sprintf("%v", v)
+				default:
+					f.Expr = fmt.Sprintf("'%v'", v)
+				}
```
begelundmuller, line 659: 1. As we discussed, this should ideally be pushed into the dialects. 
2. I'm not sure we can really rely just on the Go types here. It would be best to take into account the DB-level types, to prevent comparison/union issues in joins. This can perhaps be accomplished by leveraging the `drivers.Result.Schema` which is returned when querying the OLAP.
    - There are also people who use non-basic types for dimensions, e.g. a `CODE_STRUCT`. Ideally we handle all these types, but that is probably not feasible initially. Instead, whenever we see a dimension/measure type used that we can't handle, we should probably fail gracefully by skipping the two-phase optimization.
4. Also, did you look into if we can leverage positional args instead of putting literals into the strings (similar to how `ExprNode` works)? I'm not sure if that's a good idea or not, just asking.

runtime/metricsview/ast.go
```diff
@@ -623,6 +628,70 @@ func (a *AST) buildUnderlyingWhere() (*ExprNode, error) {
...
+			inDims = append(inDims, fields)
+		}
+
+		// Add the measure fields to the inline view.
+		inMeasures := make([][]FieldNode, 0, len(a.query.inlineMeasures))
+		for measure, values := range a.query.inlineMeasures {
+			fields := make([]FieldNode, 0, len(values))
+			for _, v := range values {
+				if v == nil {
+					v = nilExpr
```
begelundmuller, line 674: Why can't we use `null` in SQL to represent nils?

runtime/metricsview/ast.go
```diff
@@ -640,6 +709,72 @@ func (a *AST) buildBaseSelect(alias string, comparison bool) (*SelectNode, error
...
+						foundNil = true
+					} else {
+						vals = append(vals, k)
+					}
+				}
+				inExpr = &Expression{
+					Condition: &Condition{
+						Operator: OperatorIn,
+						Expressions: []*Expression{
+							{Name: dim.(string)},
```
begelundmuller, line 735: If `dim` comes from an OLAP query, I don't think you can be sure it's a string. As mentioned above, we support e.g. struct/map dims.

runtime/metricsview/query.go
```diff
@@ -23,6 +23,11 @@ type Query struct {
 	Offset              *int64      `mapstructure:"offset"`
 	TimeZone            string      `mapstructure:"time_zone"`
 	UseDisplayNames     bool        `mapstructure:"use_display_names"`
+
+	// internal state
+	inlineBaseSelect bool          `mapstructure:"-"`
+	inlineDims       map[any][]any `mapstructure:"-"` // dimension -> column values
+	inlineMeasures   map[any][]any `mapstructure:"-"` // measure -> column values
```
begelundmuller, line 30: It somehow feels dirty to add this internal state to the `Query` type (it's widely used many places, but the internal state here applies to a specific internal optimization). Contrast this with the `rewriteQueryDruidExactify`, which despite rewriting the query does not need to add any internal fields. To the extent possible, it would be nice if the two-phase approach could live mainly in the `executor_rewrite_two_phase` file, and only need minimal changes to the generic types in this package. 

Did you consider implementing the optimization by traversing the `AST` type instead of `Query`, similar to how `rewriteApproxComparisons` works?

begelundmuller (changes requested):
> > CTE rewrite is disabled by default now as it cause double scan of CTE in Druid, Duckdb and CH. It can be enabled but does not take effect if 2 phase optimization is done. It can be enabled to run queries on very high cardinality (~1M or greater) when sorting by other than base column which may fail in engines like Druid with ResourceLimitException.

Would it make sense to keep it enabled for exports, since we probably shouldn't use the two-phase approach for exports, which may have high cardinality results?

runtime/drivers/registry.go
pjain1: I think it should fall back to non-CTE approach which works faster than CTE in general. However it causes issues with druid dimensions having high cardinality column so for those customer we can enable CTE approach. Best would to be have non-CTE enabled, monitor for ResourceLimitExceeced exception and automatically fall back to CTE approach

runtime/metricsview/ast.go
pjain1: outdated and removed, however we are using NULL while writing sql when we encounter this exception

runtime/metricsview/query.go
pjain1: removed

runtime/metricsview/ast.go
pjain1: I have not tried `ExprNode` approach, just was sql as of now. Will try it out

runtime/metricsview/ast.go
pjain1: using query dim name now directly from the query

pjain1:
> Would it make sense to keep it enabled for exports, since we probably shouldn't use the two-phase approach for exports, which may have high cardinality results?

For exports of high cardinality columns even with CTE approach we don't add limit to the CTE query and that fails with Resource Limit Exceeded exception. I think we need another config to limit base query as well ? Thoughts ?

begelundmuller:
FYI the tests are still failing

runtime/drivers/olap.go
```diff
@@ -495,6 +500,128 @@ func (d Dialect) DateDiff(grain runtimev1.TimeGrain, t1, t2 time.Time) (string,
...
+		return false, "", fmt.Errorf("unsupported type %q", typ)
+	}
+}
+
+func (d Dialect) GetTimeExpr(t time.Time) (bool, string) {
+	switch d {
+	case DialectClickHouse:
+		return true, fmt.Sprintf("parseDateTimeBestEffort('%s')", t.Format(time.RFC3339))
+	case DialectDuckDB, DialectDruid, DialectPinot:
+		return true, fmt.Sprintf("CAST('%s' AS TIMESTAMP)", t.Format(time.RFC3339))
```
begelundmuller, line 619: Should it use `time.RFC3339Nano`? The databases usually have at least millisecond precision, but `RFC3339` only has second precision IIRC

runtime/drivers/olap.go
```diff
@@ -495,6 +500,128 @@ func (d Dialect) DateDiff(grain runtimev1.TimeGrain, t1, t2 time.Time) (string,
...
+	return prefix + suffix, dimVals, nil
+}
+
+func (d Dialect) GetValExpr(val any, typ runtimev1.Type_Code) (bool, string, error) {
+	if val == nil {
+		return true, "NULL", nil
+	}
+	switch typ {
+	case runtimev1.Type_CODE_STRING:
+		return true, fmt.Sprintf("'%v'", val), nil
```
begelundmuller, line 591: Is there a reason you didn't go with putting `?` and returning an array of positional args to bind to the query?

begelundmuller (changes requested):

runtime/metricsview/executor_rewrite_two_phase_comparisons.go
```diff
@@ -0,0 +1,157 @@
...
+
+	var bm []Measure
+	for _, qm := range qry.Measures {
+		if qm.Compute == nil || (qm.Compute.ComparisonValue == nil && qm.Compute.ComparisonDelta == nil && qm.Compute.ComparisonRatio == nil && qm.Compute.PercentOfTotal == nil) {
+			bm = append(bm, qm)
+			continue
+		}
+
+		if qm.Name == sortField.Name {
+			// only supported sorting on base value and dims TODO extend for comparison value
```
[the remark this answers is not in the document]
pjain1: not urgent as UI does not seem to be sorting by comparison value anywhere

runtime/drivers/olap.go
pjain1: changed to use positional args for CH and duckdb, druid does not support positional args for inline queries

runtime/metricsview/executor_rewrite_two_phase_comparisons.go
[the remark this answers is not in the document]
begelundmuller: What about this? But we can handle it in a follow up if it's not easy

Screenshot 2025-01-16 at 12 00 47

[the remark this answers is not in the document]
pjain1: this is sorting by delta not comparison

begelundmuller (approved):

PR #6338, opened 2025-01-02
cache plan name

pjain1:
Fixes https://github.com/rilldata/rill-private-issues/issues/973
- Cache and update plan name on performing any subscription related action 
- Cached plan name is returned as part of GetOrganization rpc
- GetBillingSubscription does not use cache always fetches afresh so should not be used on each page load but only on billing page
- For handling external changes (directly done in Orb, although it should not be done), webhook are used to update the cache. They can be removed if we are ok with some staleness when external modifications are done.


admin/billing/orb.go
```diff
@@ -30,6 +30,13 @@ const (
 
 var ErrCustomerIDRequired = errors.New("customer id is required")
 
+var planCache = make(map[string]planCacheEntry)
+
+type planCacheEntry struct {
+	planDisplayName string
+	lastUpdated     time.Time
+}
```
begelundmuller, line 38: Since this is updated with webhooks, it should be consistent, right? So can it be stored in Postgres instead? 

It would be useful to use Postgres because a) we have multiple replicas of the `admin` service, b) it makes introspection easier and provides easy access to the plan name in downstream pipelines

admin/database/database.go
```diff
@@ -333,6 +333,7 @@ type Organization struct {
 	PaymentCustomerID                   string    `db:"payment_customer_id"`
 	BillingEmail                        string    `db:"billing_email"`
 	CreatedByUserID                     *string   `db:"created_by_user_id"`
+	CachedPlanDisplayName               *string   `db:"cached_plan_display_name"`
```
begelundmuller, line 336: 1. Keep billing-related fields together, i.e. put the plan name above `CreatedByUserID`
2. See other comment, I don't think a `*string` is warranted here – an empty string should be a sufficient indicator that there's no plan.

admin/server/organizations.go
```diff
@@ -74,6 +75,36 @@ func (s *Server) GetOrganization(ctx context.Context, req *adminv1.GetOrganizati
 		perms.ReadProjects = true
 	}
 
+	if org.CachedPlanDisplayName == nil {
+		sub, err := s.admin.Biller.GetActiveSubscription(ctx, org.BillingCustomerID)
```
begelundmuller, line 79: Maybe add a comment that it's for backwards compatibility? Hopefully we can eventually remove it.

admin/jobs/river/biller_event_handlers.go
```diff
@@ -308,3 +309,58 @@ func (w *PaymentFailedGracePeriodCheckWorker) checkFailedInvoicesForOrg(ctx cont
...
+		QuotaDeployments:                    org.QuotaDeployments,
+		QuotaSlotsTotal:                     org.QuotaSlotsTotal,
+		QuotaSlotsPerDeployment:             org.QuotaSlotsPerDeployment,
+		QuotaOutstandingInvites:             org.QuotaOutstandingInvites,
+		QuotaStorageLimitBytesPerDeployment: org.QuotaStorageLimitBytesPerDeployment,
+		BillingCustomerID:                   org.BillingCustomerID,
+		PaymentCustomerID:                   org.PaymentCustomerID,
+		BillingEmail:                        org.BillingEmail,
+		CreatedByUserID:                     org.CreatedByUserID,
+		CachedPlanDisplayName:               &sub.Plan.DisplayName,
```
begelundmuller, line 359: Is it guaranteed to have a display name? If it may be empty (e.g. in case we forgot to set one in Orb), it should perhaps fall back to the unique name or ID instead (so it doesn't look like no plan is assigned).

admin/database/postgres/migrations/0059.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE orgs ADD COLUMN cached_plan_display_name TEXT;
```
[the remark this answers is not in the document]
pjain1: 1. ok 
2. ok
> null plan name and empty plan name is effectively the same, no need for additional states 

I want to be able to distinguish between two as there are states when there will be no plan - for example when org is just created or when org subscription ended, in this case if plan name is just empty then will know that there is no plan and would not call orb to fetch plan name again on each call to `GetOrganization` API.

admin/server/organizations.go
pjain1: There can be some delay in receiving or handling webhook request so its not just for backwards compatibility

admin/jobs/river/biller_event_handlers.go
pjain1: Plan display name is not set in orb at all, its generated logically as per product requirement here -https://github.com/rilldata/rill/blob/main/admin/billing/orb.go#L636

admin/database/postgres/migrations/0059.sql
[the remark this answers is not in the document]
begelundmuller: Okay, so it's to build a bridge for old rows, which won't initially have the plan stored? It's ugly but makes sense. It definitely needs a docstring about that difference though, it's not easy to guess that `null` means not loaded and `""` means no plan.

admin/server/organizations.go
begelundmuller: I think in practice it's better to return the plan name as unpopulated or `"Unknown plan"` than try to call the Orb API in that case. The `GetOrganization` API call is important for every page load and many CLI commands, which would create a dangerous dependency if Orb is unavailable.

pjain1: Makes sense but for now we keep it and remove in later versions ? Any trick to not use it for new orgs but only for older orgs. May be we run a one time job to populate display name or triggerable cmd that syncs plan names with org and we manually run it once and remove this code ?

admin/database/database.go
pjain1: 1. ok
2. based on discussion [above](https://github.com/rilldata/rill/pull/6338#discussion_r1952690650) we should keep this `*string`

admin/jobs/jobs.go
```diff
@@ -27,6 +27,8 @@ type Client interface {
 	RepairOrgBilling(ctx context.Context, orgID string) (*InsertResult, error) // biller is just used for deduplication
 	StartOrgTrial(ctx context.Context, orgID string) (*InsertResult, error)
 	PurgeOrg(ctx context.Context, orgID string) (*InsertResult, error)
+
+	PlanCacheUpdate(ctx context.Context, billingCustomerID string) (*InsertResult, error)
```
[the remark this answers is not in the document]
pjain1: sure I can make that change

pjain1:
lint job failing with
```
Failed executing command with error: the configuration contains invalid elements
```

begelundmuller (approved):
LGTM

PR #6340, opened 2025-01-02
nit : close results in exports in sql resolver

k-anshul:


begelundmuller (approved):
Also commented on this separately on this PR: https://github.com/rilldata/rill/pull/6341

PR #6341, opened 2025-01-02
nit : Remove checked allocator in arrow code usages

k-anshul:
The `CheckedAllocator` is expected to be used in test code only. Mostly adds extra overheads as well.

### DIFF
--- a/runtime/resolvers/sql.go
+++ b/runtime/resolvers/sql.go
@@ -160,6 +160,7 @@ func (r *sqlResolver) generalExport(ctx context.Context, w io.Writer, filename s
 	if err != nil {
 		return err
 	}
+	defer res.Close()
 
 	meta := make([]*runtimev1.MetricsViewColumn, len(res.Schema.Fields))
 	for i, f := range res.Schema.Fields {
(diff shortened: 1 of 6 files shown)


runtime/resolvers/sql.go
```diff
@@ -160,6 +160,7 @@ func (r *sqlResolver) generalExport(ctx context.Context, w io.Writer, filename s
 	if err != nil {
 		return err
 	}
+	defer res.Close()
```
begelundmuller, line 163: Good catch! I don't think the `sqlResolver` is currently used for exporting anywhere, so it's probably not the cause of memory leaks. But there may be other places – maybe we should generally search for calls to `OLAPStore.Execute` and audit that results are closed?

begelundmuller (approved):

runtime/resolvers/sql.go
k-anshul: > so it's probably not the cause of memory leaks

Yeah.

> maybe we should generally search for calls to OLAPStore.Execute and audit that results are closed

Yeah that's how I found this one as well.

PR #6345, opened 2025-01-03
fix: custom compare time range not applied

AdityaHegde:


web-common/src/features/dashboards/time-controls/comparison-pill/ComparisonPill.svelte
```diff
@@ -54,12 +54,6 @@
     start: Date,
     end: Date,
   ) {
-    if (!showTimeComparison) {
```
AdityaHegde, line 57: Moved to `setSelectedComparisonRange` to make sure this action leads to a single store update.

djbarnwal (approved):

PR #6352, opened 2025-01-06
fix : Use database_url for postgres to duckdb transporter

k-anshul:


runtime/drivers/duckdb/utils.go
```diff
@@ -23,19 +23,23 @@ func parseSinkProperties(props map[string]any) (*sinkProperties, error) {
...
 	if err := mapstructure.Decode(props, cfg); err != nil {
 		return nil, fmt.Errorf("failed to parse source properties: %w", err)
 	}
-	if cfg.DSN != "" { // For mysql, postgres the property is called as dsn and not db
+	if cfg.DSN != "" { // For mysql the property is called as dsn
 		cfg.Database = cfg.DSN
 	}
+	if cfg.DatabaseURL != "" { // For postgres the property is called as database_url
+		cfg.Database = cfg.DatabaseURL
+	}
```
begelundmuller, line 42: Usually `database` would be only the database name, but `database_url` would be a connection string (all of scheme, host, port, user, password, database name, ...). What about just having `database_url` be an alias for `dsn`?

begelundmuller (changes requested):

runtime/drivers/duckdb/utils.go
k-anshul: I am setting everything in `database` to simplify the check here in `duckb-to-duckdb` transporter: https://github.com/rilldata/rill/blob/c801193c5ef814e4d354450014201604c103635c/runtime/drivers/duckdb/transporter_duckDB_to_duckDB.go#L49-L50

This is a throw away code and is expected to be removed soon so we can keep the same check for now ?

begelundmuller (approved):

PR #6353, opened 2025-01-06
model executor for s3/gcs/azure to duckdb

k-anshul:

Adds a model executor that ingests data from s3/gcs/azure using duckdb extension (Note :  GCS works via SQL compatibility mode and native both).
Also supports incremental ingestion.
Sample model yaml that ingests from s3.

```
connector: s3
path: s3://rill-developer.rilldata.io/AdBids.csv.gz

output:
  connector: duckdb
```

**Note** : Tested only for GCS and S3.


runtime/drivers/gcs/gcs.go
```diff
@@ -71,7 +71,10 @@ var spec = drivers.Spec{
 type driver struct{}
 
 type configProperties struct {
-	SecretJSON      string `mapstructure:"google_application_credentials"`
+	SecretJSON string `mapstructure:"google_application_credentials"`
+	// When working in s3 compatible mode
+	AccessKeyID     string `mapstructure:"aws_access_key_id"`
+	SecretAccessKey string `mapstructure:"aws_secret_access_key"`
```
begelundmuller, line 77: It seems like Google uses the terms `google_access_key_id` and `google_access_key_secret` (let me know if you can find another source): https://cloud.google.com/storage/docs/aws-simple-migration#storage-list-buckets-s3-python.

Maybe we should use those in the context of the `gcs` connector? If people pass `aws_xxx` keys, then they should probably use `type: s3` and leverage `endpoint` to customize the endpoint.

runtime/drivers/duckdb/model_executor_objectstore_self.go
```diff
@@ -0,0 +1,166 @@
...
+		}
+		if s3Config.Region != "" {
+			sb.WriteString(", REGION ")
+			sb.WriteString(safeSQLString(s3Config.Region))
+		}
+		sb.WriteRune(')')
+		m.PreExec = sb.String()
+	case "gcs":
+		// GCS works via S3 compatibility mode
+		s3Config := &s3.ConfigProperties{}
```
begelundmuller, line 112: See other comment, maybe we should use `gcs.ConfigProperties` here and use Google-related property names?

runtime/drivers/duckdb/model_executor_objectstore_self.go
```diff
@@ -0,0 +1,166 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse s3 config properties: %w", err)
+		}
+		var sb strings.Builder
+		sb.WriteString("CREATE OR REPLACE TEMPORARY SECRET ")
+		sb.WriteString(safeSecretName)
+		sb.WriteString(" (TYPE GCS")
+		if s3Config.AllowHostAccess {
+			sb.WriteString(", PROVIDER CREDENTIAL_CHAIN")
+		}
```
begelundmuller, line 123: For GCS, what does `CREDENTIAL_CHAIN` do? If it doesn't support `GOOGLE_APPLICATION_CREDENTIALS` (which I believe it doesn't?), then if we start using this implementation for legacy sources, it will be a regression, right? I'm wondering if we need to keep a fallback to the old implementation where we download files in that case.

begelundmuller (changes requested):

runtime/drivers/duckdb/model_executor_objectstore_self.go
k-anshul: 1. In case of duckdb, GCS is just a convenient wrapper over s3 which takes care of endpoint implicitly. So if I set `aws_access_key_id` and `aws_secret_access_key` as environment variables and set the following secret in duckdb then I can query files stored on GCS.
```
CREATE SECRET gcs_secret (
     TYPE GCS,
     PROVIDER CREDENTIAL_CHAIN
)
SELECT * FROM read_json(['gcs://.../*.json.gz'])
```

2. It does not support `GOOGLE_APPLICATION_CREDENTIALS`.
3. My understanding was that given so far we are the only ones using GCS so we will migrate our projects to use s3 compatibility. Given both duckdb and clickhouse does not have native GCS support, I get an impression that either people are comfortable using s3 compatibility mode or nobody is using GCS :)
In my opinion it will be ideal to deprecate our home grown connectors for object stores as well and just rely on duckdb's connectors but I will let you make the call.
If we do end up supporting our home grown connectors for GCS then I think it will be best to not support GCS via duckdb at all and if somebody wants to use duckdb's connectors they can use s3 with endpoint set.

runtime/drivers/gcs/gcs.go
k-anshul: As noted in this [comment](https://github.com/rilldata/rill/pull/6353#discussion_r1906790808), given GCS works by setting `aws_access_key_id` and `aws_secret_access_key` environment variables I thought of keeping keys as `aws_access_key_id` and `aws_secret_access_key` itself. 
DuckDB just calls them `KEY_ID` and `SECRET` so calling them `google_xxx` for GCS and `aws_xxx` for S3 may also be fine but that naming convention is not used frequently in their docs so somewhat skeptical with `google_access_key_id` and `google_access_key_secret`.

begelundmuller: Got it, then how about having:
- `key_id` and `secret` (like DuckDB calls them)
- or, `access_key_id` and `access_key_secret`

I understand it's just a layer over S3, but it just seems too weird to have `aws_access_key_id` as a property on a GCS connector.

runtime/drivers/duckdb/model_executor_objectstore_self.go
begelundmuller: 1. Thanks, I responded in the other comment :) 
3. I'm definitely on board with removing the homegrown connectors if possbile. Can you think of a log search in Datadog or similar we could do to find out if anyone other than us is using GCS in cloud?

runtime/drivers/gcs/gcs.go
k-anshul: Sure. `key_id` and `secret` sounds good.

runtime/drivers/duckdb/model_executor_objectstore_self.go
k-anshul: I can't find any easy way to find what all users use GCS connector so I added a log here that we can merge and release in upcoming release : https://github.com/rilldata/rill/pull/6380

begelundmuller: You could also consider doing a little script that combines `rill sudo project search %%` with `rill sudo clone` to `grep` for sources that contain `gs://` or `: gcs`...

k-anshul: Yeah valid idea but I think we can wait for the log to be added. Would be somewhat easier ?

runtime/drivers/duckdb/model_executor_objectstore_self.go
```diff
@@ -0,0 +1,167 @@
...
+		var sb strings.Builder
+		sb.WriteString("CREATE OR REPLACE TEMPORARY SECRET ")
+		sb.WriteString(safeSecretName)
+		sb.WriteString(" (TYPE GCS")
+		if gcsConfig.AllowHostAccess {
+			sb.WriteString(", PROVIDER CREDENTIAL_CHAIN")
+		}
+		if gcsConfig.KeyID != "" {
+			fmt.Fprintf(&sb, ", KEY_ID %s, SECRET %s", safeSQLString(gcsConfig.KeyID), safeSQLString(gcsConfig.Secret))
+		}
```
begelundmuller, line 127: 1. If `gcsConfig.SecretJSON` is set, but not `gcsConfig.KeyID`, should we perhaps return an error here?
2. Can you also update `rill env configure` to request `key_id` and `secret` instead of `google_application_credentials` for GCS? I guess that would be more appropriate now, right?

begelundmuller (changes requested):

runtime/drivers/duckdb/model_executor_objectstore_self.go
begelundmuller: Up to you :) Has the log been rolled out to prod yet? Looking forward to the findings!

runtime/drivers/duckdb/model_executor_objectstore_self.go
k-anshul: 1. Sure.
2. I planned to do this after we migrate sources to be run as models because as of now sources backed by GCS are taking usual code path.

begelundmuller: 2. Okay, do you have an issue where you are tracking these follow ups? If not, can you add them to the unification issue?

k-anshul: I think we will stick with `google_application_credentials` for now and add that later as per Mike and Nishant's inputs.

runtime/drivers/duckdb/model_executor_objectstore_self.go
```diff
@@ -0,0 +1,254 @@
...
+
+	"github.com/mitchellh/mapstructure"
+	"github.com/rilldata/rill/runtime/drivers"
+	"github.com/rilldata/rill/runtime/drivers/azure"
+	"github.com/rilldata/rill/runtime/drivers/gcs"
+	"github.com/rilldata/rill/runtime/drivers/s3"
+	"github.com/rilldata/rill/runtime/pkg/fileutil"
+)
+
+var errObjectStoreUsesNativeCreds = errors.New("Uses native credentials")
```
begelundmuller, line 19: What does "native" credentials mean? If you mean credentials from the environment, probably better to use a word like "host" or "environment" (since we usually "use" native to refer to connectors that are native to the database).

begelundmuller (approved):

runtime/drivers/duckdb/model_executor_objectstore_self.go
k-anshul: Okay so this was specifically for GCS in cases when it uses native credentials and not the s3 compatibility ones.
I can change it to `errGCSUsesNativeCreds` as well.

PR #6359, opened 2025-01-06
Update .gitignore to ignore local PostgreSQL mount directory

grahamplata:
## Changes
Small change to the `.gitignore` to ignore local PostgreSQL mount directory.

## Context
I encountered this when setting up services for the first time via `Make`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,6 +41,7 @@ tsc-tmp/
 /models
 /dashboards
 /sources
+/postgres
 
 # check in generated data files
 !/web-local/tests/data/*.parquet


ericpgreen2:
@himadrisingh I wonder if we have a recent regression here. Shouldn't this `postgres` directory live in the `dev-cloud-state` directory?

himadrisingh:
> @himadrisingh I wonder if we have a recent regression here. Shouldn't this `postgres` directory live in the `dev-cloud-state` directory?

this should be within some cloud-state directory. do we have steps to reproduce this?

begelundmuller:
@grahamplata @ericpgreen2 @himadrisingh This is due to Himadri's recent change to `rill devtool` – you need to use a nightly version of Rill when using the devtool for the configuration to be parsed and applied correctly. Using the latest release with the new config files leads to files being created in the repo root instead.

PR #6377, opened 2025-01-08
fix: public projects with logged out user fetching billing and bookmark data

AdityaHegde:


AdityaHegde:
Ya using this in all places would be clean.

Fixed the issue with bookmark.

PR #6380, opened 2025-01-09
Add a log for gcs usage

k-anshul:
refers to https://github.com/rilldata/rill/pull/6353#discussion_r1907452585

This is to identify if some external user is using GCS.

### DIFF
--- a/runtime/connection_cache.go
+++ b/runtime/connection_cache.go
@@ -146,7 +146,7 @@ func (r *Runtime) openAndMigrate(ctx context.Context, cfg cachedConnectionConfig
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


begelundmuller (approved):

runtime/connection_cache.go
```diff
@@ -145,10 +146,18 @@ func (r *Runtime) openAndMigrate(ctx context.Context, cfg cachedConnectionConfig
 		}
 	}
 
+	r.Logger.Info("opening connection", zap.String("instance_id", cfg.instanceID), zap.String("driver", cfg.driver), zap.String("name", cfg.name), zap.Bool("provision", cfg.provision))
```
begelundmuller, line 149: This should not show up on local if you run `rill start`. Have you checked that it doesn't?

Otherwise it might make sense to use `debug` level instead (which is hidden in the local console unless `--verbose` is specified).

k-anshul: Good point. Changed to debug.

PR #6384, opened 2025-01-09
Feat: Canvas cosmetic changes

djbarnwal:
1. Add Vega lite config input field and merge with default config
2. Update styles for markdown
3. Update styles for the default vega lite config
4. Add menu for positional fields in chart to show/hide axis title
5. Hide axis title by default
6. Remove title and description for Markdown
7. Split field selector component into single and multi select
8. Update styles for KPI component, add default title


web-common/src/features/canvas/components/charts/util.ts
```diff
@@ -29,3 +31,34 @@ export const chartMetadata: ChartMetadata[] = [
...
+  let parsedConfig: Config;
+
+  try {
+    parsedConfig = JSON.parse(config) as Config;
+  } catch {
+    console.warn("Invalid JSON config");
+    return defaultConfig;
+  }
+
+  const reverseArrayMerge = (
```
AdityaHegde, line 46: Could you add a comment on why we need this reverse array merge?

Also since this is not really dependant on any function arguments, could you move this outside?

AdityaHegde (approved):
Didnt fully review the input framework yet. Will leave my comments in a day or so, you can address them in follow ups.

web-common/src/features/canvas/components/charts/util.ts
djbarnwal: This is required to put the values added by the user in the config before our default values. For example for category colors, we have our own set of 14 values. When users add their color, those would be put first.

I will add a comment and move this out to a separate file in a separate PR to avoid conflicts with PR forked against this.

PR #6387, opened 2025-01-09
Add positional argument to `env` commands for project

grahamplata:
**Description**
Normalizing `env` command usage to how `project` is implemented. 

`rill env` supports flag and positional arg
- `rill env show --org=demo --project=rill-openrtb-prog-ads`
- `rill env show rill-openrtb-prog-ads --org=demo`

---

`--env` flag allows for the export of `.env` output
```bash
# Example: rill env show rill-openrtb-prog-ads --org=demo --env > .env
rill env show rill-openrtb-prog-ads --org=demo --env
TEST_KEY=TEST_VALUE
```

Additionally the character row count was increased https://github.com/lensesio/tableprinter/blob/master/tableprinter.go#L83-L84

```
// if RowCharLimit > 0 && RowTextWrap == true then wrap the line otherwise replace the trailing with "..."
```


cli/cmd/env/set.go
```diff
@@ -38,6 +36,11 @@ func SetCmd(ch *cmdutil.Helper) *cobra.Command {
 				}
 			}
 
+			key, value, err := envValidator.ParseAndValidate(keyValPair)
+			if err != nil {
+				return err
+			}
```
begelundmuller, line 42: Could `Flags().StringToStringVar` ([docs](https://pkg.go.dev/github.com/spf13/pflag#StringToStringVar)) be used for this instead?

cli/cmd/env/set.go
```diff
@@ -11,23 +11,21 @@ import (
 
 // SetCmd is sub command for env. Sets the variable for a project
 func SetCmd(ch *cmdutil.Helper) *cobra.Command {
-	var projectPath, projectName, environment string
+	var projectPath, projectName, environment, keyValPair string
 
 	setCmd := &cobra.Command{
-		Use:   "set <key> <value>",
-		Args:  cobra.ExactArgs(2),
+		Use:   "set [<project-name>] [--env=key=value]",
```
begelundmuller, line 17: 1. Did you consider `set [<project>] <key> <value>`, similar to e.g. `rill project describe` ([link](https://github.com/rilldata/rill/blob/cf0bcf4dbd4806268f90a0c9fa071ba186bbb472/cli/cmd/project/describe.go#L20-L32))? Not necessarily saying that's better, just curious which approach you think is more idiomatic? (IMO neither mixed required/optional positional args nor required flags feel intuitive, but can't think of other alternatives.)
2. If you go ahead with this approach, note there are some non-generated docs that need to be updated (search `docs/` for `rill env set`)

cli/cmd/env/show.go
```diff
@@ -38,18 +42,21 @@ func ShowCmd(ch *cmdutil.Helper) *cobra.Command {
...
 				})
 			}
 
-			ch.PrintfSuccess("\nVariables\n\n")
-			ch.PrintData(table)
+			if cmd.Flags().Changed("env") {
+				printEnv(envVars)
+			} else {
+				ch.PrintData(envVars)
+			}
```
begelundmuller, line 59: A few thoughts:
1. Using `--env` for this maybe feels confusing because we use `--env` to specify variable values in a couple other commands
3. I wonder if people generally would prefer the shell export format instead of a table? I know I would. We could just have that as the only format if you agree.
3. The `printEnv` function does not include `v.Environment` in the output (it allows for environment-specific overrides of variable values). One idea is to group using comments in the output, like this:
```
foo=bar
hello=world

# development
foo=bazz

# production
foo=foo
```

cli/cmd/env/set.go
grahamplata: Yeah I see what you mean, I think I will stub it out and try

grahamplata: Normalized on 1.

cli/cmd/env/show.go
grahamplata: 1. Agreed I'll adjust that
2. I had the exact same thought I figured the use case would be to pipe to a dotenv or vault
3. I think that is reasonable ~ is it worth providing a filter on that?

begelundmuller: > I think that is reasonable ~ is it worth providing a filter on that? ~ specify env

It kind of already does with the `rill env show --environment` flag. I say "kind of" because if an environment is provided, it actually evaluates the variables for that environment, i.e. it will return non-environment specific variables overwritten by the variables for the requested environment.

Sounds like this isn't completely clear – if you have any ideas for help menu or docs improvements here, feel free to send a PR :)

cli/cmd/env/set.go
```diff
@@ -56,6 +61,5 @@ func SetCmd(ch *cmdutil.Helper) *cobra.Command {
 	setCmd.Flags().StringVar(&projectName, "project", "", "Cloud project name (will attempt to infer from Git remote if not provided)")
 	setCmd.Flags().StringVar(&projectPath, "path", ".", "Project directory")
 	setCmd.Flags().StringVar(&environment, "environment", "", "Optional environment to resolve for (options: dev, prod)")
```
begelundmuller, line 63: Unrelated, but I guess this should be:

```suggestion
	setCmd.Flags().StringVar(&environment, "environment", "", "Optional environment to set for (options: dev, prod)")
```

cli/cmd/env/show.go
```diff
@@ -62,6 +69,33 @@ func ShowCmd(ch *cmdutil.Helper) *cobra.Command {
 	return showCmd
 }
 
+func formatEnvVar(name, value string) string {
+	return fmt.Sprintf("%s=%s", name, value)
+}
```
begelundmuller, line 74: If `value` contains a newline, this might throw off the formatting. Newlines could happen e.g. for GCS credentials where we store the JSON payload from the service account `.json` file that the user provides.

Maybe it's as simple as:
```suggestion
func formatEnvVar(name, value string) string {
	return fmt.Sprintf("%s=%q", name, value)
}
```

(unless you want to be fancy and only add quotes conditionally when the value has non-standard characters.)

cli/cmd/env/show.go
```diff
@@ -62,6 +69,33 @@ func ShowCmd(ch *cmdutil.Helper) *cobra.Command {
...
+		if _, ok := envMap[v.Environment]; !ok {
+			envMap[v.Environment] = []*variable{}
+		}
+		envMap[v.Environment] = append(envMap[v.Environment], v)
+	}
+
+	for env, vars := range envMap {
+		if env != tmpKey {
+			fmt.Printf("# %s\n", env)
+		}
```
begelundmuller, line 92: It would be nice to guarantee that the variables without an environment are emitted at the top of the file (before the environment-specific overrides)

begelundmuller (changes requested):

cli/cmd/env/show.go
grahamplata: Good call, adjusting

grahamplata:
```shell
➜  rill env show rill-openrtb-prog-ads
key="value"
TEST_KEY="TEST_VALUE"

# Environment: dev
foo="bar"

# Environment: prod
cat="dog"

# Environment: test
meow="mix"
```

begelundmuller (approved):

PR #6393, opened 2025-01-10
Estimate tar sizes for rill deploy command

grahamplata:
Set `estimated_size_bytes` field when calling `AdminService.CreateAsset` during a project deploy. I split apart the compound function so size could be measured prior to upload.


begelundmuller (approved):
Looks good!

PR #6396, opened 2025-01-12
feat: dimension and measure filter for canvas

djbarnwal:
Adds dimension and measure filters for canvas.


web-common/src/features/dashboards/filters/Filters.svelte
```diff
@@ -78,6 +86,8 @@
     dimensionIdMap,
   );
 
+  $: console.log($dashboardStore.whereFilter);
```
ericpgreen2, line 89: Leftover console log

web-common/src/features/dashboards/stores/filter-utils.ts
```diff
@@ -386,3 +390,34 @@ export function isExpressionUnsupported(expression: V1Expression) {
 
   return false;
 }
+
+export function getValidFilterForMetricView(
```
ericpgreen2, line 394: A few things:
1. Nit: for consistency, "MetricView" -> "MetricsView"
2. Given this function's name, I'd expect it to take a `metricsView` as a parameter
3. `compose` might be a more descriptive verb? As is, the function could use a comment to explain what it actually does.

web-common/src/features/dashboards/filters/Filters.svelte
ericpgreen2, line 1: @briangregoryholmes is your proposal that we lift the StateManagers even further up the component tree, or keep them here?

web-common/src/features/canvas/components/table/selector.ts
```diff
@@ -46,6 +46,31 @@ export function useComparisonStartEndTime(
   });
 }
 
+export function useStartEndTime(
+  instanceId: string,
+  metricsViewName: string,
+  timeRange: string,
```
ericpgreen2, line 52: I'm not following how this `timeRange` is represented. Is this the time range name? Or is it an ISO 8601 time range?

web-common/src/features/canvas/components/time-filter-store.ts
```diff
@@ -0,0 +1,62 @@
...
+import { derived, type Writable } from "svelte/store";
+
+// This helper returns a derived store that yields the final timeRange and where clause
+export function createTimeAndFilterStore(
+  ctx: StateManagers,
+  instanceId: string,
+  metricsViewName: string,
+  {
+    timeRangeStore,
+    overrideTimeRange,
```
ericpgreen2, line 18: Could you add a comment to explain when the `overrideTimeRange` comes into play?

web-common/src/features/canvas/components/time-filter-store.ts
ericpgreen2, line 1: This strikes me as a derivation of state in the CanvasEntity, so it should be a selector contained within that class. 

Or is this meant to also account for when specific component state diverges from the overall canvas state?

ericpgreen2 (approved):
Left some comments, but generally LGTM!

web-common/src/features/canvas/components/table/selector.ts
djbarnwal: This is just a placeholder for now using 8601. Will be replaced by the new time syntax

web-common/src/features/canvas/stores/canvas-spec.ts
```diff
@@ -1,24 +1,199 @@
...
+  type MetricsViewSpecDimensionV2,
+  type MetricsViewSpecMeasureV2,
+  type V1CanvasSpec,
 } from "@rilldata/web-common/runtime-client";
-import { derived, type Readable } from "svelte/store";
+import { derived, get, type Readable } from "svelte/store";
 
 export class CanvasResolvedSpec {
   canvasSpec: Readable<V1CanvasSpec | undefined>;
   isLoading: Readable<boolean>;
```
ericpgreen2, line 12: Do we need `isError` and `error` for error handling?

ericpgreen2: If so, might it be easier to expose the `$validSpecStore` itself, rather than duplicating the fields that map 1:1 to the TanStack Query response? If so, maybe the exposure of the `useCanvas` response belongs in the top-level `CanvasEntity` class?

web-common/src/features/entity-management/WatchResourcesClient.ts
```diff
@@ -220,6 +221,8 @@ export class WatchResourcesClient {
           }
 
           case ResourceKind.Explore: {
+            console.log("logggExplore");
```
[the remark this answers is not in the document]
djbarnwal: https://github.com/rilldata/rill/pull/6483

PR #6397, opened 2025-01-12
feat: two tiered nav in embeds

mindspank:
For internal use only.


ericpgreen2:
I know this is still in draft, but I'll just note that we'll need tests for this. We should:
 1. Add an example project to https://github.com/rilldata/rill-qa
 2. Add an integration test for the `BreadcrumbItemEmbed` component

PR #6411, opened 2025-01-14
refac: integrate canvas resolver API

djbarnwal:
Integrate canvas resolver API and newly added properties to the YAML.

- Integrate the new resolved canvas API
- Create new class for spec readables
- Add keys and style canvas page panel as per mocks
- Add all time range store

There are more changes and refactoring needed for time filters. Those will be addressed in a different PR.


web-common/src/features/canvas/stores/canvas-spec.ts
ericpgreen2, line 1: It seems we have two layers of selector functions:
1. In the `select` function provided to the `ResolveCanvas` API's TanStack Query request
2. The functions provided to the derived stores in this file

Maybe consider if this should be a one-tier system, not two-tier? I wonder if the TanStack Query call and all  selectors could all be co-located in this class.

ericpgreen2 (approved):

web-common/src/features/canvas/stores/canvas-spec.ts
djbarnwal: I wonder if there's any performance cost of creating a derivable out of query response. The reason I chose to use derivables instead of `select` function was the ease of integration in the components. All these selectors derive from a single query so we wouldn't have to check for fetch status or error at every place the selectors are used. 

I have added more methods to `canvas-spec.ts` in https://github.com/rilldata/rill/pull/6396 I have fixed the nits in this PR. We can continue this discussion around selection there.

PR #6412, opened 2025-01-14
Not to be merged : Source spec defined at OLAP level

k-anshul:
**Motivation**
The source spec for each driver is not independent but is coupled with OLAP engine.
As an example, for postgres, user can define a SQL that duckdb can run against postgres db but for clickhouse user needs to specify database and table as well.
Since the source spec is coupled with OLAP engine so it seems logical to keep the source/ingestion spec in OLAP driver itself. 
We are also moving to using OLAP engine's internal connectors so it also fits well with that design. 

**Solution**
1. Add another method to drivers.OLAPStore : `SourcesSpec` which returns a map of drivers and their properties.
2. Define this independently for each OLAP driver for all drivers that it supports ingestion from.
3. Add an API that returns the spec for all driver that a particular OLAP engine supports.
4. The config properties will continue to be part of individual drivers.

### DIFF
--- a/proto/rill/runtime/v1/api.proto
+++ b/proto/rill/runtime/v1/api.proto
@@ -219,7 +219,7 @@ service RuntimeService {
   }
 
   rpc ListSourceConnectors(ListSourceConnectorsRequest) returns (ListSourceConnectorsResponse) {
-    option (google.api.http) = {get: "/v1/connectors/{olap}/meta"};
+    option (google.api.http) = {get: "/v1/source-connectors/{olap}/meta"};
   }
   
   // Access management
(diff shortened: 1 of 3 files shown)


begelundmuller (commented):
I agree the combinatorics of input/output/executor options would be nice to capture. 

Regarding putting the spec on the OLAP, some questions on my mind are:
- What about models with a staging connector, such as `bigquery -> s3 -> clickhouse`?
- What about the upcoming export functionality (Rohith working on that now), where the OLAP _outputs_ to another connector such as S3?

I don't have a concrete proposal here, but did you consider:
- Supporting annotations on `Spec` that give extra detail specific to a given input or output connector
- Maybe having a separate global schema (maybe using JSON schema) that contains all possible keys/combinations for all connectors (instead of trying to put specs inside individual connectors as we have done so far)

k-anshul:
Thanks for the inputs. We will discuss and take this separately.

PR #6413, opened 2025-01-14
feat: improve watermark calculation by adding the option to cache it

AdityaHegde:
This PR centralises the calculation of min, max and watermark to executors.
- [x] Adds `Executor.Timestamps` that does the calculation of min, max (previously in metrics_time_range resolver) and watermark.
- [x] Updates `metrics_time_range` resolver to call `Executor.Timestamps`.
- [x] Adds `Executor.BindQuery` that allows for setting the timestamps from outside. This allows for getting a cached timestamps by getting it from `metrics_time_range` resolver.
- [x] Update `MetricsViewTimeRange` to use `metrics_time_range` resolver.
- [x] Update usages of `Executor.Query` to set cached timestamps.

A future PR will refactor metrics_sql to move time range resolution to `Executor`. So it is not updated to use the new resolver.


proto/rill/runtime/v1/queries.proto
```diff
@@ -1024,6 +1024,8 @@ message TimeRangeSummary {
   google.protobuf.Timestamp min = 1;
   // Not optional, not null
   google.protobuf.Timestamp max = 2;
+  // Not optional, not null
+  google.protobuf.Timestamp watermark = 4;
   message Interval {
```
begelundmuller, line 1029: Do you use the `Interval` on the frontend? Or can we just remove it?

runtime/metricsview/executor.go
```diff
@@ -127,10 +127,40 @@ func (e *Executor) ValidateQuery(qry *Query) error {
 	panic("not implemented")
 }
 
-// Watermark returns the current watermark of the metrics view.
-// If the watermark resolves to null, it defaults to the current time.
-func (e *Executor) Watermark(ctx context.Context) (time.Time, error) {
-	return e.loadWatermark(ctx, nil)
+// Timestamps queries min, max and watermark for the metrics view
+func (e *Executor) Timestamps(ctx context.Context, executionTime *time.Time) (time.Time, time.Time, time.Time, error) {
```
begelundmuller, line 131: How about creating a result type here? Like:
```
type TimestampsResult struct {
  Min time.Time
  Max time.Time
  Watermark time.Time
}
```

Then it can also be used instead of the `min, max, watermark time.Time` in `type Executor`

runtime/metricsview/executor.go
```diff
@@ -127,10 +127,40 @@ func (e *Executor) ValidateQuery(qry *Query) error {
...
+		e.min, e.max, e.watermark, err = e.resolveClickHouseAndPinot(ctx)
+	default:
+		return time.Time{}, time.Time{}, time.Time{}, fmt.Errorf("not available for dialect '%s'", e.olap.Dialect())
+	}
+	if err != nil {
+		return time.Time{}, time.Time{}, time.Time{}, err
+	}
+	if executionTime != nil {
+		e.watermark = *executionTime
+	}
```
begelundmuller, line 154: A tricky problem here is that in many cases (like most alerts/reports), only the watermark is used, so this override here actually means the timestamps query was redundant.

Any thoughts on a simple way to optimize that? Otherwise we can go with this for now and think about optimizing this later.

runtime/metricsview/executor_timestamps.go
```diff
@@ -0,0 +1,243 @@
...
+	} else {
+		watermarkExpr = fmt.Sprintf("MAX(%s)", timeDim)
+	}
+
+	rangeSQL := fmt.Sprintf(
+		"SELECT min(%[1]s) as \"min\", max(%[1]s) as \"max\", %[2]s as \"watermark\", max(%[1]s) - min(%[1]s) as \"interval\" FROM %[3]s %[4]s",
+		timeDim,
+		watermarkExpr,
+		escapedTableName,
+		e.security.RowFilter(),
```
begelundmuller, line 33: This is not prefixed with a `WHERE` clause, so would break if there's a row filter.

We have a new test framework for resolvers that supports easily testing different OLAPs and different user attributes / security policies. See here:
- For more details: runtime/resolvers/testdata/README.md
- For an example: runtime/resolvers/testdata/metrics_comparisons.yaml

runtime/metricsview/executor_timestamps.go
```diff
@@ -0,0 +1,243 @@
...
+		if err != nil {
+			return time.Time{}, time.Time{}, time.Time{}, err
+		}
+		if v := rowMap["min"]; v != nil {
+			minTime, ok := v.(time.Time)
+			if !ok {
+				return time.Time{}, time.Time{}, time.Time{}, fmt.Errorf("not a timestamp column")
+			}
+			return minTime, rowMap["max"].(time.Time), rowMap["watermark"].(time.Time), nil
+		}
```
begelundmuller, line 58: I think it would be simpler to do:
```go
var minTime, maxTime, watermark time.Time
...
err = rows.Scan(&minTime, &maxTime, &watermark)
```

`Scan` will then handle nulls, casting it to a `time.Time`, etc.

runtime/metricsview/executor_timestamps.go
```diff
@@ -0,0 +1,243 @@
...
+
+	var watermarkExpr string
+	if e.metricsView.WatermarkExpression != "" {
+		watermarkExpr = e.metricsView.WatermarkExpression
+	} else {
+		watermarkExpr = fmt.Sprintf("MAX(%s)", timeDim)
+	}
+
+	rangeSQL := fmt.Sprintf(
+		"SELECT min(%[1]s) as \"min\", max(%[1]s) as \"max\", %[2]s as \"watermark\", max(%[1]s) - min(%[1]s) as \"interval\" FROM %[3]s %[4]s",
```
begelundmuller, line 29: `interval` is not used in the result here. If you remove it, I think the implementation here can be combined with `resolveClickHouseAndPinot`

runtime/queries/metricsview_aggregation.go
```diff
@@ -130,6 +132,48 @@ func (q *MetricsViewAggregation) Export(ctx context.Context, rt *runtime.Runtime
...
+	}
+	watermark, err := anyToTime(row["watermark"].(string))
+	if err != nil {
+		return err
+	}
+
+	err = e.BindQuery(ctx, qry, minTime, maxTime, watermark)
+	if err != nil {
+		return err
+	}
```
begelundmuller, line 175: The export can take a long time, so it would be best for `res.Close()` to run earlier than the end of the function. Perhaps pull this into a smaller scope or util function

runtime/queries/metricsview_aggregation.go
```diff
@@ -130,6 +132,48 @@ func (q *MetricsViewAggregation) Export(ctx context.Context, rt *runtime.Runtime
...
+	maxTime, err := anyToTime(row["max"].(string))
+	if err != nil {
+		return err
+	}
+	watermark, err := anyToTime(row["watermark"].(string))
+	if err != nil {
+		return err
+	}
```
begelundmuller, line 170: Casting to `string` here is redundant, `anyToTime` does it

runtime/queries/metricsview_aggregation.go
```diff
@@ -355,3 +399,11 @@ func metricViewExpression(expr *runtimev1.Expression, sql string) (*metricsview.
 	}
 	return nil, nil
 }
+
+func anyToTime(tm any) (time.Time, error) {
+	tmStr, ok := tm.(string)
+	if !ok {
+		return time.Time{}, errors.New("invalid type")
+	}
```
begelundmuller, line 407: Perhaps also try casting to `time.Time` before failing

runtime/queries/metricsview_aggregation.go
```diff
@@ -355,3 +399,11 @@ func metricViewExpression(expr *runtimev1.Expression, sql string) (*metricsview.
...
+	return time.Parse(time.RFC3339, tmStr)
```
begelundmuller, line 408: Consider `time.RFC3339Nano` to be safe

runtime/queries/metricsview_aggregation_test.go
```diff
@@ -4807,6 +4813,59 @@ func testMetricsViewsAggregation_comparison_with_offset(t *testing.T, rt *runtim
 	require.Equal(t, "news.google.com,3.59,3.69", fieldsToString2digits(rows[i], "dom", "m1", "m1__p"))
 }
 
+func testMetricsViewsAggregation_comparison_iso_timerange(t *testing.T, rt *runtime.Runtime, instanceID string) {
```
begelundmuller, line 4816: See the other note about testing using a new file in `runtime/resolvers/testdata` – it will make it easier to do many tests for different values/OLAPs/security policies/etc.

begelundmuller (changes requested):

proto/rill/runtime/v1/queries.proto
AdityaHegde: No we do not. Safe to remove it.

runtime/queries/metricsview_aggregation.go
AdityaHegde: Moved to a util funciton

proto/rill/runtime/v1/queries.proto
AdityaHegde: We are doing a `max-min` in UI. There was one place in queries where it was used. But it is better to move the math to that query.

runtime/queries/metricsview_aggregation_test.go
AdityaHegde: Moved this to `runtime/resolvers/testdata`

runtime/metricsview/executor_timestamps.go
```diff
@@ -0,0 +1,199 @@
...
+		ExecutionTimeout: defaultExecutionTimeout,
+	})
+	if err != nil {
+		return TimestampsResult{}, err
+	}
+	defer rows.Close()
+
+	if rows.Next() {
+		var minTime, maxTime, watermark *time.Time
+		err = rows.Scan(&minTime, &maxTime, &watermark)
```
begelundmuller, line 52: This ends up sending `**time.Time` to `rows.Scan`, might not be well handled by all implementations. Also makes the pointers unsafe to access since they may still be nil. I think this would be safer:
```suggestion
		var minTime, maxTime, watermark time.Time
		err = rows.Scan(&minTime, &maxTime, &watermark)
```

runtime/metricsview/executor_timestamps.go
```diff
@@ -0,0 +1,199 @@
...
+			return TimestampsResult{}, err
+		}
+		// Table is empty for the filters. duckdb returns null in this case
+		if minTime == nil {
+			return TimestampsResult{}, nil
+		}
+		return TimestampsResult{
+			Min:       *minTime,
+			Max:       *maxTime,
+			Watermark: *watermark,
```
begelundmuller, line 63: Dereferencing pointers without `nil` checks is not safe (maxTime and watermark are not checked). See comment above – I think the pointers can be avoided. Then the downstream logic can just check with `time.Time.IsZero()`

runtime/queries/metricsview_aggregation.go
```diff
@@ -331,6 +343,49 @@ func (q *MetricsViewAggregation) rewriteToMetricsViewQuery(export bool) (*metric
 	return qry, nil
 }
 
+func (q *MetricsViewAggregation) resolveTimestampResult(ctx context.Context, rt *runtime.Runtime, instanceID string) (metricsview.TimestampsResult, error) {
```
begelundmuller, line 346: Can you also add this logic in `runtime/resolvers/metrics.ResolveInteractive`? It's okay to copy-paste since we're deprecating the `runtime/queries` package.

runtime/resolvers/metricsview_time_range.go
```diff
@@ -110,54 +102,22 @@ func (r *metricsViewTimeRangeResolver) Validate(ctx context.Context) error {
...
-		row["max"] = tr.Max.AsTime()
-		row["interval"] = map[string]any{
-			"days":   tr.Interval.Days,
-			"months": tr.Interval.Months,
-			"micros": tr.Interval.Micros,
-		}
+	if !ts.Min.IsZero() {
+		row["min"] = ts.Min
+		row["max"] = ts.Max
+		row["watermark"] = ts.Watermark
```
begelundmuller, line 114: Actually it would be best if this always has the same map keys – otherwise the `schema` will not match the `row` values. So if a timestamp is null, it should be added as `row["key"] = nil`.

runtime/server/queries_metrics.go
```diff
@@ -456,3 +476,36 @@ func lookupMetricsView(ctx context.Context, rt *runtime.Runtime, instanceID, nam
...
 	return res, mv.State, nil
 }
+
+type decodedTimeRangeSummary struct {
+	Min, Max, Watermark string
+}
+
+func decodeTimeRangeSummary(row map[string]any) (*runtimev1.TimeRangeSummary, error) {
+	timeRangeSummary := decodedTimeRangeSummary{} // TODO: move this to a good place
+	err := mapstructure.Decode(row, &timeRangeSummary)
```
begelundmuller, line 486: Have you tried using `WeakDecode`? Then I think it would be able to directly decode ISO strings into `time.Time` types (but I may be wrong)

begelundmuller (changes requested):

runtime/metricsview/executor_timestamps.go
AdityaHegde: This is needed for duckdb. It returns `null` when there are no rows. Perhaps that is why we had different implementation for duckdb. We should probably go back to that. What do you think?

runtime/server/queries_metrics.go
AdityaHegde: It wont work by itself. We need to create a decoder with `DecoderConfig` and `StringToTimeHookFunc`. I just reused the code from `metricsview_aggregation` since we will also need for rilltime resolver.

runtime/metricsview/executor_timestamps.go
begelundmuller: And you checked it can't scan `null` into a `time.Time`? (Where the output would have `t.IsZero() == true`)

If yes, then no need for separate implementations, having an extra pointer layer is fine, as long as we check for `nil` before dereferencing.

begelundmuller (approved):

runtime/metricsview/executor_timestamps.go
AdityaHegde: Ya it throws a type error. Saying it cant store nil into `time.Time`

PR #6419, opened 2025-01-14
Update googlesheets.md

royendo:
Removed unnecessary screenshot and added note

### DIFF
--- a/docs/docs/reference/connectors/googlesheets.md
+++ b/docs/docs/reference/connectors/googlesheets.md
@@ -24,5 +24,6 @@ Make sure to replace `SPREADSHEET_ID` and `SHEET_ID` with the ID of your spreads
 
 ![Connecting to Google Sheets](/img/reference/connectors/googlesheets/googlesheets.png)
 
-<img src = '/img/reference/connectors/bigquery/bigquery.png' class='rounded-png' />
-<br />
+:::note gsheets DuckDB Community Extension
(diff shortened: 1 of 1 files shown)


docs/docs/reference/connectors/googlesheets.md
```diff
@@ -24,5 +24,6 @@ Make sure to replace `SPREADSHEET_ID` and `SHEET_ID` with the ID of your spreads
 
 ![Connecting to Google Sheets](/img/reference/connectors/googlesheets/googlesheets.png)
 
-<img src = '/img/reference/connectors/bigquery/bigquery.png' class='rounded-png' />
-<br />
\ No newline at end of file
+:::note gsheets DuckDB Community Extension
+We have not yet implemented the DuckDB community extension for gsheets in our deployment of DuckDB. Please contact us if you are interested in using this feature.
```
mindspank, line 28: Maybe add a note on why someone would want to use that extension, Auth etc

royendo: based on our slack convo, do you think its worthwhile to put an example here using init_sql, or as you mentioned since its not ready yet, should we wait on that?

mindspank: We can wait for now, we should be good to support from a security pov but it probably deserve it's own page for loading extensions

mindspank (approved):

PR #6425, opened 2025-01-15
fix: Disallow sort by hidden measure in leaderboard

### LINKED ISSUE #6342: Explore: Expanding a Leaderboard and visible measures
1. Show only a single measure in the timeseries.
2. Set the leaderboards to sort/display deltas by a measure that is not visible
3. Expand the leaderboard

Observe that only the visible measure and not the sorted by measure is visible as a column and thus no time comparison can be shown. 

I expected that: we harmonize visible columns in a expanded leaderboard and/or only allow sort by measures for measures that are currently visible.

AdityaHegde:
closes #6342

We allow leaderboard sort by measure to be hidden. This looks weird when a dimension is expanded. We have decided to disallow sorting by a measure that is hidden to keep things consistent.

### DIFF
--- a/web-common/src/features/dashboards/state-managers/selectors/measures.ts
+++ b/web-common/src/features/dashboards/state-managers/selectors/measures.ts
@@ -175,6 +175,18 @@ export const getIndependentMeasures = (
   return [...measures];
 };
 
+export const getFilteredSimpleMeasures = (
+  measures: MetricsViewSpecMeasureV2[],
+) => {
+  return (
+    measures?.filter(
+      (m) =>
+        !m.window &&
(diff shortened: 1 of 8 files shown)


web-common/src/features/dashboards/state-managers/selectors/measures.ts
```diff
@@ -175,6 +175,18 @@ export const getIndependentMeasures = (
   return [...measures];
 };
 
+export const getFilteredSimpleMeasures = (
```
ericpgreen2, line 178: Nit: rename to `getSimpleMeasures()`. `getFilteredSimpleMeasures()` almost sounds like the simple measures are somehow further filtered.

ericpgreen2 (approved):

PR #6431, opened 2025-01-15
Support security policies on API resources

grahamplata:
Add support for security policies on api resources. https://github.com/rilldata/rill-private-issues/issues/1065

> Example api.yaml config file

This example adds a security policy directly on the type: `api` resource via api yaml file

```yaml
# Endpoint requires user to hold the role of admin
type: api
sql: SELECT 1
security:
  access: '{{ .user.admin }}'
```

Additionally add a property on the type: `api` resource that determines whether the underlying resolver's security policy should be applied or not. For example, the property could be `skip_resolver_security: <bool>`

```
metrics:
  metrics_view: sales
  dimensions: [country]
  measures: [profits]
  filter: country='DK'

skip_resolver_security: true

security:
  access: '{{ .user.admin }}'
```


runtime/compilers/rillv1/parse_api.go
```diff
@@ -77,6 +77,16 @@ func (p *Parser) parseAPI(node *Node) error {
 	}
 	node.Refs = append(node.Refs, resolverRefs...)
 
+	securityRules, err := tmp.Security.Proto()
+	if err != nil {
+		return fmt.Errorf("failed to parse security rules: %w", err)
+	}
+	for _, rule := range securityRules {
+		if rule.GetAccess() == nil {
+			return fmt.Errorf("the 'custom api' resource type only supports 'access' security rules")
```
begelundmuller, line 86: nit: for normal users they just call them `api`, e.g. `type: api` (it's only internally we words like "custom" or "user-defined" to disambiguate from internal/built-in APIs)

```suggestion
			return fmt.Errorf("the 'api' resource type only supports 'access' security rules")
```

runtime/compilers/rillv1/parse_api.go
```diff
@@ -88,20 +98,23 @@ func (p *Parser) parseAPI(node *Node) error {
...
-	ResourceStatus map[string]any `yaml:"resource_status"`
+	Connector            string             `yaml:"connector"`
+	SQL                  string             `yaml:"sql"`
+	MetricsSQL           string             `yaml:"metrics_sql"`
+	API                  string             `yaml:"api"`
+	Args                 map[string]any     `yaml:"args"`
+	Glob                 yaml.Node          `yaml:"glob"` // Path (string) or properties (map[string]any)
+	ResourceStatus       map[string]any     `yaml:"resource_status"`
+	Security             SecurityPolicyYAML `yaml:"security"`
+	SkipResolverSecurity bool               `yaml:"skip_resolver_security"`
```
begelundmuller, line 117: I think these should be defined on the `APIYAML` type since they are specific to it – the `DataYAML` type is a partial which is used by multiple resources (and should probably be moved to its own file, similar to `parse_partial_field_selector.go`, though you don't need to do that now)

runtime/compilers/rillv1/parse_api.go
```diff
@@ -88,20 +98,23 @@ func (p *Parser) parseAPI(node *Node) error {
...
-	Glob           yaml.Node      `yaml:"glob"` // Path (string) or properties (map[string]any)
-	ResourceStatus map[string]any `yaml:"resource_status"`
+	Connector            string             `yaml:"connector"`
+	SQL                  string             `yaml:"sql"`
+	MetricsSQL           string             `yaml:"metrics_sql"`
+	API                  string             `yaml:"api"`
+	Args                 map[string]any     `yaml:"args"`
+	Glob                 yaml.Node          `yaml:"glob"` // Path (string) or properties (map[string]any)
+	ResourceStatus       map[string]any     `yaml:"resource_status"`
+	Security             SecurityPolicyYAML `yaml:"security"`
```
begelundmuller, line 116: I also think this should be a pointer, i.e. `*SecurityPolicyYAML` (not sure if it matters, but we use that the other places and the YAML library has subtle differences in its treatment of pointer vs. literal types, so probably best to stay consistent with the other places its used)

runtime/compilers/rillv1/parse_api.go
```diff
@@ -114,6 +127,10 @@ func (p *Parser) parseDataYAML(raw *DataYAML, contextualConnector string) (strin
 	var refs []ResourceName
 	resolverProps := make(map[string]any)
 
+	if raw.SkipResolverSecurity {
+		resolverProps["skip_resolver_security"] = raw.SkipResolverSecurity
+	}
```
begelundmuller, line 132: Can you make this a typed field on the `APISpec` instead? Since it's a property that should be handled upstream, not inside the individual resolver implementations.

runtime/server/api.go
```diff
@@ -65,14 +65,22 @@ func (s *Server) apiHandler(w http.ResponseWriter, req *http.Request) error {
 	}
 
 	// TODO: Should it resolve security and check access here?
```
begelundmuller, line 67: This comment can be removed now :)

begelundmuller (changes requested):
I'm wondering if we can come up with a better property name than `skip_resolver_security` – perhaps `skip_nested_security` or `skip_transitive_security`. What do you think is clearest? Let's also ask @mindspank.

runtime/compilers/rillv1/parse_api.go
grahamplata: Ahh, agreed. Moving

runtime/compilers/rillv1/parse_api.go
```diff
@@ -37,6 +40,7 @@ func (p *Parser) parseAPI(node *Node) error {
 	var openapiSummary string
 	var openapiParams []*structpb.Struct
 	var openapiSchema *structpb.Struct
+	var securityRules []*runtimev1.SecurityRule
```
begelundmuller, line 43: nit: I don't think you need this if you use `:=` on line 84

runtime/compilers/rillv1/parse_api.go
```diff
@@ -77,6 +81,26 @@ func (p *Parser) parseAPI(node *Node) error {
...
+
+	if tmp.SkipResolverSecurity {
+		// Skip security rules for nested resolvers
+		propsMap := resolverProps.AsMap()
+		propsMap["skip_nested_security"] = tmp.SkipResolverSecurity
+		resolverProps, err = structpb.NewStruct(propsMap)
+		if err != nil {
+			return fmt.Errorf("failed to update resolver properties: %w", err)
+		}
+	}
```
begelundmuller, line 102: In the earlier comment:
> Can you make this a typed field on the APISpec instead? Since it's a property that should be handled upstream, not inside the individual resolver implementations.

I meant that instead of adding it in the resolver props, better to put it directly on the `APISpec` in `proto/rill/runtime/v1/resources.proto` and then set it like this around line 112:
```
r.APISpec.SkipNestedSecurity = tmp.SkipNestedSecurity
```

runtime/server/api.go
```diff
@@ -64,15 +64,22 @@ func (s *Server) apiHandler(w http.ResponseWriter, req *http.Request) error {
...
 
-	// TODO: Should it resolve security and check access here?
+	resolverProps := api.Spec.ResolverProperties.AsMap()
+	skipResolverSecurity := resolverProps["skip_nested_security"]
+	claims := auth.GetClaims(ctx).SecurityClaims()
+	// for each claim skip checks
+	if skipResolverSecurity != nil && skipResolverSecurity.(bool) {
+		// SkipChecks
+		claims.SkipChecks = true
+	}
```
begelundmuller, line 74: Referencing my comment above – then you'd be able to just have something like this:
```suggestion
	// Rewrite the claims before passing them to the resolver
	claims := auth.GetClaims(ctx).SecurityClaims()
	if api.Spec.SkipNestedSecurity {
		claims.SkipChecks = true
	}
```

begelundmuller (changes requested):

begelundmuller (approved):

PR #6439, opened 2025-01-16
fix : certain db level settings in duckdb can't be set again

k-anshul:
This is just a fix for release blocker. The permanent solution will be as follows:
1. Most of our settings are global configuration except `max_expression_depth`.
2. It seems extensions are now loaded at db level and not connection level. The following code now works:
```
	_, err = conn1.ExecContext(context.Background(), "INSTALL spatial; LOAD spatial;")
	if err != nil {
		log.Fatal(err)
	}

	row := conn2.QueryRowContext(context.Background(), "SELECT ST_Area('POLYGON((0 0, 0 1, 1 1, 1 0, 0 0))'::GEOMETRY);")
	var area float64
	err = row.Scan(&area)
	if err != nil {
		log.Fatal(err)
	}
	log.Println(area)
```
3. Introduce two new options in `rduckdb` : 
       a. `DBInitQueries` : Runs for db handle once. Extensions, global configurations and `init_sql` in connector config will be DBInitQueries.
       b. `ConnInitQueries`: Runs for each connection. Local configurations will be `ConnInitQueries`.
4. After opening DB run `DBInitQueries`.
5. Run `ConnInitQueries` before returning a read or write connection.

### DIFF
--- a/runtime/drivers/duckdb/duckdb.go
+++ b/runtime/drivers/duckdb/duckdb.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"log/slog"
 	"net/url"
-	"path/filepath"
 	"strings"
 	"sync"
 	"time"
@@ -500,10 +499,7 @@ func (c *connection) reopenDB(ctx context.Context) error {
 	// We want to set preserve_insertion_order=false in hosted environments only (where source data is never viewed directly). Setting it reduces batch data ingestion time by ~40%.
 	// Hack: Using AllowHostAccess as a proxy indicator for a hosted environment.
 	if !c.config.AllowHostAccess {
-		bootQueries = append(bootQueries,
-			"SET preserve_insertion_order TO false",
-			fmt.Sprintf("SET secret_directory = %s", safeSQLString(filepath.Join(dataDir, ".duckdb", "secrets"))),
-		)
+		bootQueries = append(bootQueries, "SET preserve_insertion_order TO false")
 	}
 
 	// Add init SQL if provided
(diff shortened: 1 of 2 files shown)


runtime/drivers/duckdb/duckdb.go
```diff
@@ -500,10 +499,7 @@ func (c *connection) reopenDB(ctx context.Context) error {
 	// We want to set preserve_insertion_order=false in hosted environments only (where source data is never viewed directly). Setting it reduces batch data ingestion time by ~40%.
 	// Hack: Using AllowHostAccess as a proxy indicator for a hosted environment.
 	if !c.config.AllowHostAccess {
-		bootQueries = append(bootQueries,
-			"SET preserve_insertion_order TO false",
-			fmt.Sprintf("SET secret_directory = %s", safeSQLString(filepath.Join(dataDir, ".duckdb", "secrets"))),
```
begelundmuller, line 505: Why does this need to be removed when `lock_configuration` is no longer set to true?

begelundmuller: It seems like an important setting to reduce chance of persistent secrets getting shared across instances

k-anshul: The proper fix as I mentioned in the PR desc is to allow two kinds of init_queries one that works at db level and other that works at connection level.

Given all instances are isolated right now so I went ahead by removing this setting completely but if you prefer we can add the proper fix as well.

begelundmuller: Okay, can you add an issue to the next sprint with reference to these settings so we remember to add them back?

begelundmuller (approved):

PR #6441, opened 2025-01-16
quote inline string values

pjain1:


runtime/drivers/olap.go
```diff
@@ -596,7 +596,7 @@ func (d Dialect) GetValExpr(val any, typ runtimev1.Type_Code) (bool, string, err
 	}
 	switch typ {
 	case runtimev1.Type_CODE_STRING:
-		return true, fmt.Sprintf("'%v'", val), nil
+		return true, fmt.Sprintf("%s", d.EscapeStringValue(val.(string))), nil
```
begelundmuller, line 599: Can you make the cast safe? Usually `Type_CODE_STRING` should be a Go `string`, but there may be unpredictable edge cases (maybe around `null`s)

begelundmuller (approved):

runtime/drivers/olap.go
pjain1: null is already checked before switch but will add check anyways

pjain1: added

PR #6446, opened 2025-01-17
use literals in distributed engine definition in ch

pjain1:
otherwise it was quoting the local table name like `ENGINE = Distributed('abc', 'default', '"table_local"', rand())` and queries were failing on it saying local table not found

### DIFF
--- a/runtime/drivers/clickhouse/olap.go
+++ b/runtime/drivers/clickhouse/olap.go
@@ -609,7 +609,7 @@ func (c *connection) createTable(ctx context.Context, name, sql string, outputPr
 		database = "currentDatabase()"
 	}
 	fmt.Fprintf(&distributed, "CREATE OR REPLACE TABLE %s %s AS %s", safeSQLName(name), onClusterClause, safelocalTableName(name))
-	fmt.Fprintf(&distributed, " ENGINE = Distributed('%s', %s, '%s'", c.config.Cluster, database, localTableName(name))
+	fmt.Fprintf(&distributed, " ENGINE = Distributed(%s, %s, %s", safeSQLString(c.config.Cluster), database, safeSQLString(localTableName(name)))
 	if outputProps.DistributedShardingKey != "" {
 		fmt.Fprintf(&distributed, ", %s", outputProps.DistributedShardingKey)
 	} else {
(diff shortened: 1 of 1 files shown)


runtime/drivers/clickhouse/olap.go
```diff
@@ -609,7 +609,7 @@ func (c *connection) createTable(ctx context.Context, name, sql string, outputPr
 		database = "currentDatabase()"
 	}
 	fmt.Fprintf(&distributed, "CREATE OR REPLACE TABLE %s %s AS %s", safeSQLName(name), onClusterClause, safelocalTableName(name))
-	fmt.Fprintf(&distributed, " ENGINE = Distributed(%s, %s, %s", safeSQLName(c.config.Cluster), database, safelocalTableName(name))
+	fmt.Fprintf(&distributed, " ENGINE = Distributed('%s', %s, '%s'", c.config.Cluster, database, localTableName(name))
```
k-anshul, line 612: We should use `safeSQLString` instead of `'%s'`. `safeSQLString` takes care of escaping any single quotes within string.

k-anshul (changes requested):

runtime/drivers/clickhouse/olap.go
```diff
@@ -609,7 +609,7 @@ func (c *connection) createTable(ctx context.Context, name, sql string, outputPr
 		database = "currentDatabase()"
 	}
 	fmt.Fprintf(&distributed, "CREATE OR REPLACE TABLE %s %s AS %s", safeSQLName(name), onClusterClause, safelocalTableName(name))
-	fmt.Fprintf(&distributed, " ENGINE = Distributed(%s, %s, %s", safeSQLName(c.config.Cluster), database, safelocalTableName(name))
+	fmt.Fprintf(&distributed, " ENGINE = Distributed(%s, %s, %s", safeSQLString(c.config.Cluster), database, safeSQLString(localTableName(name)))
```
k-anshul, line 612: I know this is an existing code but shouldn't database also be quoted ?
If you do quote this then need to take care of fact that database can also be `currentDatabase()`.

k-anshul (approved):
Adding a comment

runtime/drivers/clickhouse/olap.go
pjain1: Yes I did not quote because of existing code but added it now

k-anshul (approved):

PR #6453, opened 2025-01-18
No auto-push for docs, only manual push from main allowed

himadrisingh:


royendo (commented):
I'm not too familiar with this but , here removes CLI. Is this expected?

( it's not on the other file)

royendo:
Ah nvm looked at old file LGTM

royendo (approved):

PR #6458, opened 2025-01-21
fix : only allow a single instance of embedded clickhouse server

k-anshul:
closes https://github.com/rilldata/rill-private-issues/issues/943
Also blocks change of port.

### DIFF
--- a/runtime/drivers/clickhouse/clickhouse.go
+++ b/runtime/drivers/clickhouse/clickhouse.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/ClickHouse/clickhouse-go/v2"
@@ -25,9 +24,8 @@ import (
 )
 
 func init() {
-	d := &driver{}
-	drivers.Register("clickhouse", d)
-	drivers.RegisterAsConnector("clickhouse", d)
+	drivers.Register("clickhouse", driver{})
+	drivers.RegisterAsConnector("clickhouse", driver{})
 }
 
 var spec = drivers.Spec{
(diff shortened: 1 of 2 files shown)


runtime/drivers/clickhouse/clickhouse.go
```diff
@@ -94,7 +96,57 @@ var spec = drivers.Spec{
...
+
+func (d *driver) startEmbeddedClickhouse(port int, dataDir, tempDir string, logger *zap.Logger) (*clickhouse.Options, func() error, error) {
+	d.embedMu.Lock()
+	defer d.embedMu.Unlock()
+	if d.embed != nil {
+		if d.embedPort == port {
+			d.refs++
+			return d.embedOpts, d.stopEmbeddedClickhouse, nil
+		}
+		return nil, nil, fmt.Errorf("change of `embed_port` is not allowed while application is running, please restart rill")
```
begelundmuller, line 120: nit:
```suggestion
		return nil, nil, fmt.Errorf("change of `embed_port` is not allowed while the application is running, please restart Rill")
```

runtime/drivers/clickhouse/clickhouse.go
```diff
@@ -300,6 +353,8 @@ type connection struct {
 	opts *clickhouse.Options
 	// embed is embedded clickhouse server for local run
 	embed *embedClickHouse
+	// stopEmbed should be called stop the embedded clickhouse server
+	stopEmbed func() error
```
begelundmuller, line 357: Instead of having `startEmbeddedClickhouse` and `stopEmbeddedClickhouse` on the `driver` and having a separate `stopEmbed` state variable here, could all this logic be stored/handled in the `embedClickhouse`? I.e. implement the reference counting inside of it, as well as return errors from it if the config is changed after it is started. And then just create and store one `embedClickhouse` globally on the `driver`.

That would prevent embed logic from spilling into this file, so we can keep it contained to the `embedClickhouse` implementation.

begelundmuller (changes requested):

runtime/drivers/clickhouse/clickhouse.go
k-anshul: Done

begelundmuller (approved):

PR #6461, opened 2025-01-21
use values stmt for inline select duckdb

pjain1:


begelundmuller (commented):
Have you checked if very long `UNION ALL` queries are efficient in ClickHouse? Wondering if it also supports `VALUES` or something more efficient.

pjain1:
> Have you checked if very long UNION ALL queries are efficient in ClickHouse? Wondering if it also supports VALUES or something more efficient.

Very long queries on CH with `UNION ALL` works, tested on a dataset with high cardanility column. However, inline query like `SELECT c1 AS a, c2 AS b FROM VALUES((1, 2), (3, 4))` works but only issue with using `c1`, `c2` conventional column names. Asked about it in the CH support as well, some suggested to use JSON like
```
SELECT *
FROM FORMAT(JSONEachRow, '{"name":"blablubb","type":1,"data":["127.0.0.1","127.0.0.2"]}{"a":123}')
```
but this has same problem of not being able to use positional args. Anyways have pushed `SELECT c1 AS a, c2 AS b FROM VALUES((1, 2), (3, 4))` change

begelundmuller (approved):

PR #6462, opened 2025-01-21
fix: compare time and dimension conflicts in url param

AdityaHegde:
Since we use comparison mode in explore preset there could be issues when both time and dimension comparison are enabled. Since we actually support for both to be present at the same time this fixes some edge cases with it. The main one is setting both and switching between pivot and explore.


web-common/src/features/dashboards/url-state/convertExploreStateToPreset.ts
```diff
@@ -95,6 +95,7 @@ function getTimeRangeFields(
   ) {
     preset.selectTimeRange = toTimeRangeParam(exploreState.selectedScrubRange);
   }
+  console.log(preset);
```
djbarnwal, line 98: Stale log

djbarnwal (approved):

PR #6466, opened 2025-01-21
Bug: Use Custom River Error Handler

grahamplata:
Allow the custom handler to manage attempts instead of worker job.

Resolves: https://github.com/rilldata/rill-private-issues/issues/1101

### DIFF
--- a/admin/jobs/river/org_jobs.go
+++ b/admin/jobs/river/org_jobs.go
@@ -35,10 +35,13 @@ func (w *InitOrgBillingWorker) Work(ctx context.Context, job *river.Job[InitOrgB
 		return fmt.Errorf("failed to find organization %s: %w", job.Args.OrgID, err)
 	}
 
-	// rare case but if its retried, we should repair the billing as it might be in some inconsistent state
-	_, _, err = w.admin.RepairOrganizationBilling(ctx, org, false)
-	if err != nil {
-		return fmt.Errorf("failed to repair billing for organization %s: %w", org.Name, err)
+	if job.Attempt > 1 {
+		// rare case but if its retried, we should repair the billing as it might be in some inconsistent state
+		_, _, err = w.admin.RepairOrganizationBilling(ctx, org, false)
+		if err != nil {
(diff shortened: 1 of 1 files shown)


begelundmuller:
@pjain1 Deferring to you to review this (for context, error logs lead to alerts, warning logs are usually overlooked unless we are specifically searching for them).

begelundmuller (commented):
Just an idea – have you considered configuring the log level to error only if `job.Attempt == job.MaxAttempts`? (You can use `w.logger.Log(level, ...)` to pass a level dynamically.)

Alternatively I wonder if we should just remove these logs and format the returned errors nicely instead (since River will log the error from the last job attempt at error level, right?)

admin/jobs/river/org_jobs.go
```diff
@@ -125,7 +115,7 @@ func (w *StartTrialWorker) Work(ctx context.Context, job *river.Job[StartTrialAr
 		TrialEndDate: sub.TrialEndDate,
 	})
 	if err != nil {
-		w.logger.Error("failed to send trial started email", zap.String("org_name", trialOrg.Name), zap.String("org_id", trialOrg.ID), zap.String("billing_email", trialOrg.BillingEmail), zap.Error(err))
+		return fmt.Errorf("failed to send trial started email for organization %s: %w", trialOrg.Name, err)
```
pjain1, line 118: I wouldn't return an error here as job will try to start the trial again, although currently `w.admin.StartTrial` is idempotent but I would not bet on it on future. I think we should just get an alert when this happens.

pjain1 (changes requested):
In general we wanted to prevent error logging in case of retires so this PR does removes explicit logging from our end however river inherently logs an error in case of retires like
```
{"level":"error","ts":1738056846.8063529,"caller":"river@v0.11.4/job_executor.go:305","msg":"jobExecutor: Job errored","error":"organization already has billing initialized","job_id":5457,"job_kind":"init_org_billing"}
``` 
see the caller above, I have no idea how we can do that, probably we just use some logic on datadog to avoid that.

pjain1 (approved):

PR #6469, opened 2025-01-22
fix: reconnect to a different github failing with `remote already exists`

### LINKED ISSUE #6468: [Regression] Unable to reconnect to different Repository from UI
**Describe the bug**
After disconnecting from remote A in the UI, and trying to connect remote B. The following error occurs 
`failed to create remote: remote already exists`

**To Reproduce**
1. Connect to remote A via UI.
2. Disconnect via the UI.
3. Try to reconnect a different remote to write RC projects to
4. Error occurs

**Expected behavior**
Disconnecting from a repository should remove any git parameters from the existing deployment to allow RC project to connect to new repo.

**Screenshots**
Image
Image

**Additional context**
Prod and release-53

AdityaHegde:
closes #6468

`.git` folder was getting added to the archive after disconnecting from github. `drivers.IsIgnored` expects a leading `/`.

Also there could already be projects with the `.git` folder. So handle this case when reconnecting to a new repo.

### DIFF
--- a/admin/server/github.go
+++ b/admin/server/github.go
@@ -1053,7 +1053,20 @@ func (s *Server) pushToGit(ctx context.Context, copyData func(projPath string) e
 	}
 
 	if empty {
-		// we need to add a remote if the repo was completely empty
+		remote, err := ghRepo.Remote("origin")
+		if err != nil && !errors.Is(err, git.ErrRemoteNotFound) {
+			return err
+		}
+		// remove the remote if it already exists
+		// this could happen when previously connected to another repo and .git got added to assets
+		if remote != nil {
+			err = ghRepo.DeleteRemote("origin")
+			if err != nil {
+				return err
+			}
+		}
+
+		// we need to add a remote as the new repo if the repo was completely empty
 		_, err = ghRepo.CreateRemote(&config.RemoteConfig{Name: "origin", URLs: []string{repo}})
 		if err != nil {
 			return fmt.Errorf("failed to create remote: %w", err)
--- a/admin/server/github.go
+++ b/admin/server/github.go
@@ -346,7 +346,7 @@ func (s *Server) ConnectProjectToGithub(ctx context.Context, req *adminv1.Connec
 			defer os.RemoveAll(downloadDir)
 			downloadDst := filepath.Join(downloadDir, "zipped_repo.tar.gz")
 			// extract the archive once the folder is prepped with git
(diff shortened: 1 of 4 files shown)


admin/server/github.go
```diff
@@ -1053,7 +1053,20 @@ func (s *Server) pushToGit(ctx context.Context, copyData func(projPath string) e
 	}
 
 	if empty {
-		// we need to add a remote if the repo was completely empty
+		remote, err := ghRepo.Remote("origin")
```
k-anshul, line 1056: Why not simply delete the `.git` folder ?

AdityaHegde: Since we are making a commit just before it would remove all of that.

AdityaHegde: Ignoring .git during untar. So this is not needed.

runtime/pkg/archive/archive.go
```diff
@@ -172,11 +172,18 @@ func untar(src, dest string) error {
 			return err
 		}
 
+		// nolint:gosec // adding the '..' check here is still triggering gosec for GSC-G305
+		if strings.Contains(header.Name, "..") ||
```
begelundmuller, line 176: How does `..` end up in the paths here?

AdityaHegde: Right now it cant. But if we allow users to upload their own archives then it can lead to this security issue. So better to have this safeguard now.

https://app.deepsource.com/directory/analyzers/go/issues/GSC-G305
https://security.snyk.io/research/zip-slip-vulnerability

k-anshul: Why don't we just delete `.git` folder after untar in `Download` since this is just to circumnavigate an older issue?
That way we don't need to worry about the security issue as well ?

AdityaHegde: Since we are untaring to a location it will overwrite exiting `.git`. Then removing the `.git` will remove anything we did before the untar. This could be an issue if we are pushing to a previously populated repo.

So it is better to ignore it here.

k-anshul (approved):

runtime/pkg/archive/archive.go
AdityaHegde: Removing the `..` check since `sanitizeArchivePath` already does it. But the linter is not smart enough to know this so we still need the `nolint`

PR #6470, opened 2025-01-22
Canvas: Improve charts 

djbarnwal:
- Simplify specs
- Support time dimension for charts
- Style tooltips
- Update label for color selector
- Format tooltip values and axis labels for measure
- Add rule for line charts


ericpgreen2 (approved):
Code LGTM 👍

PR #6479, opened 2025-01-22
.env auto deploy doc updates

royendo:
Build/ configure local credentials

setting up creds for RD, 4 options,
mention to use .env when deploying to RC

Deploy/configure deployment credentials
More details on differnece between local creds and RC creds,
env is deployed with, else youll need to deploy then define creds using rill env configure

Manage/project-management/ configure environmental variables and credentials

fix references to these docs ++ nit


royendo:
@mindspank can we push these docs now with the .env deploys update?

PR #6484, opened 2025-01-23
feat : Add a `read_write_ratio` for duckdb to split resources between serving and ETL

k-anshul:
The default split of resources is 50/50 for reads and writes. The ratio can be used to change the split.

### DIFF
--- a/runtime/drivers/duckdb/config.go
+++ b/runtime/drivers/duckdb/config.go
@@ -21,8 +21,8 @@ type config struct {
 	CPU int `mapstructure:"cpu"`
 	// MemoryLimitGB is the amount of memory available for the DB. If no ratio is set then this is split evenly between read and write.
 	MemoryLimitGB int `mapstructure:"memory_limit_gb"`
-	// ReadWriteResourceSplitRatio is the ratio of resources to allocate to the read DB. If set, CPU and MemoryLimitGB are distributed based on this ratio.
-	ReadWriteResourceSplitRatio float64 `mapstructure:"read_write_resource_split_ratio"`
+	// ReadWriteRatio is the ratio of resources to allocate to the read DB. If set, CPU and MemoryLimitGB are distributed based on this ratio.
+	ReadWriteRatio float64 `mapstructure:"read_write_ratio"`
 	// BootQueries is SQL to execute when initializing a new connection. It runs before any extensions are loaded or default settings are set.
 	BootQueries string `mapstructure:"boot_queries"`
 	// InitSQL is SQL to execute when initializing a new connection. It runs after extensions are loaded and and default settings are set.
@@ -32,7 +32,9 @@ type config struct {
 }
 
 func newConfig(cfgMap map[string]any) (*config, error) {
-	cfg := &config{}
+	cfg := &config{
+		ReadWriteRatio: 0.5,
+	}
 	err := mapstructure.WeakDecode(cfgMap, cfg)
 	if err != nil {
 		return nil, fmt.Errorf("could not decode config: %w", err)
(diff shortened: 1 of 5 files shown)


runtime/drivers/duckdb/config.go
```diff
@@ -18,14 +17,12 @@ type config struct {
...
 	CPU int `mapstructure:"cpu"`
-	// MemoryLimitGB is the amount of memory available for the read DB. If no MemoryLimitGBWrite is set then this is split evenly between read and write.
+	// MemoryLimitGB is the amount of memory available for the DB. If no ratio is set then this is split evenly between read and write.
 	MemoryLimitGB int `mapstructure:"memory_limit_gb"`
-	// CPUWrite is CPU available for the DB when writing data.
-	CPUWrite int `mapstructure:"cpu_write"`
-	// MemoryLimitGBWrite is the amount of memory available for the DB when writing data.
-	MemoryLimitGBWrite int `mapstructure:"memory_limit_gb_write"`
+	// ReadWriteResourceSplitRatio is the ratio of resources to allocate to the read DB. If set, CPU and MemoryLimitGB are distributed based on this ratio.
+	ReadWriteResourceSplitRatio float64 `mapstructure:"read_write_resource_split_ratio"`
```
begelundmuller, line 25: Perhaps this name feels unnecessarily long without making the purpose much clearer. People will probably anyway need to read the docs to understand it. Maybe just call it something like `read_write_split` or `read_write_ratio`?

runtime/pkg/rduckdb/db.go
```diff
@@ -76,27 +81,22 @@ type DBOptions struct {
 }
 
 func (d *DBOptions) ValidateSettings() error {
-	read := &settings{}
-	err := mapstructure.Decode(d.ReadSettings, read)
-	if err != nil {
-		return fmt.Errorf("read settings: %w", err)
+	if d.ReadWriteResourceSplitRatio < 0 || d.ReadWriteResourceSplitRatio >= 1 {
```
begelundmuller, line 84: If we start having separate ETL and read nodes, then they may set it to `0` or `1` to have all resources go to writes or reads respectively.

runtime/pkg/rduckdb/db.go
```diff
@@ -111,87 +111,49 @@ func (d *DBOptions) ValidateSettings() error {
...
+		d.ReadSettings["memory_limit"] = fmt.Sprintf("%d bytes", memoryLimitBytes/2)
+		d.WriteSettings["memory_limit"] = fmt.Sprintf("%d bytes", memoryLimitBytes/2)
+		if threads < 2 {
+			d.ReadSettings["threads"] = "1"
 		} else {
-			threads = writeThread
+			d.ReadSettings["threads"] = strconv.Itoa(threads / 2)
 		}
+		d.WriteSettings["threads"] = strconv.Itoa((threads + 1) / 2)
+	} else {
```
begelundmuller, line 141: Isn't this the same as having the ration be `0.5` by default? Given the comment above, perhaps it's easiest to just have the `drivers/duckdb` package set the ratio to 0.5 in `DBOptions` by default.

begelundmuller (changes requested):

runtime/drivers/duckdb/config.go
k-anshul: Sure. `read_write_ratio` sounds good.

runtime/pkg/rduckdb/db.go
k-anshul: I imagined that we will need a separate config `mode` or similar to denote that this is a read only or write only db and then we will need to revisit this logic. Setting `threads` to 0 means unlimited CPU. 
But I can remove this check for now and set 0.5 as default as suggested in other comment.

k-anshul: We will need to make some minor tweaks in the code here to support ETL and read nodes. As of now the read path is always initialised.

begelundmuller (approved):

PR #6487, opened 2025-01-23
feat: file upload input

AdityaHegde:
- [x] Add a FileInput with single or multiple file upload support.
- [x] Drag-drop support.
- [x] Add drop-over state.
- [x] Add loading and error with file upload state.
- [ ] See if it is possible to merge this with Input. Probably not since the caller needs to handle the `File` from the input. We cannot generalise it since we use different endpoint for uploading on cloud vs runtime.

Check https://github.com/rilldata/rill/pull/6521 for usage example.


ericpgreen2:
I'll let @briangregoryholmes review this as main author of `Input.svelte`

briangregoryholmes (approved):

PR #6489, opened 2025-01-23
Enable admins to view all resources on the status page

grahamplata:
This PR enables admins to view all resources on the status page. This will better help administrators diagnose and triage resources within their projects.

### DIFF
--- a/runtime/server/controller.go
+++ b/runtime/server/controller.go
@@ -61,6 +61,10 @@ func (s *Server) ListResources(ctx context.Context, req *runtimev1.ListResources
 		return strings.Compare(an.Name, bn.Name)
 	})
 
+	// Check if the requester is an admin
+	claims := auth.GetClaims(ctx)
+	isAdmin := claims.SecurityClaims().Admin()
+
 	i := 0
 	for i < len(rs) {
 		r := rs[i]
@@ -68,10 +72,10 @@ func (s *Server) ListResources(ctx context.Context, req *runtimev1.ListResources
 		if err != nil {
 			return nil, status.Error(codes.InvalidArgument, err.Error())
 		}
-		if !access {
+		// Admins can see all resources, even if they don't have access.
+		if !isAdmin && !access {
 			// Remove from the slice
 			rs[i] = rs[len(rs)-1]
-			rs[len(rs)-1] = nil
 			rs = rs[:len(rs)-1]
 			continue
 		}


runtime/server/controller.go
```diff
@@ -61,17 +61,21 @@ func (s *Server) ListResources(ctx context.Context, req *runtimev1.ListResources
 		return strings.Compare(an.Name, bn.Name)
 	})
 
+	// Check if the requester is an admin
+	claims := auth.GetClaims(ctx)
+	isAdmin := claims.SecurityClaims().Admin()
```
begelundmuller, line 66: The problem here is that `ListResources` is used many places, e.g. both the main dashboard listings and the status page listings. Outside the status page listings, people *do* expect the policies to be enforced, even for admins.

One option is to add a `skip_checks` bool flag to the `ListResources` and `GetResource` RPCs that the UI can pass in the contexts where it wants the checks to be skipped (and of course, setting it to true for a non-admin should error).

begelundmuller (changes requested):

runtime/server/controller.go
```diff
@@ -68,7 +69,8 @@ func (s *Server) ListResources(ctx context.Context, req *runtimev1.ListResources
 		if err != nil {
 			return nil, status.Error(codes.InvalidArgument, err.Error())
 		}
-		if !access {
+		// Check of the request is to skip security checks
+		if !req.SkipChecks && !access {
```
begelundmuller, line 73: Only admins should be allowed to pass `skip_checks=true` – otherwise anyone would be able to skip checks

begelundmuller (changes requested):

runtime/server/controller.go
```diff
@@ -61,13 +62,19 @@ func (s *Server) ListResources(ctx context.Context, req *runtimev1.ListResources
 		return strings.Compare(an.Name, bn.Name)
 	})
 
+	isAdmin := auth.GetClaims(ctx).SecurityClaims().Admin()
+	if isAdmin && req.SkipChecks {
+		return &runtimev1.ListResourcesResponse{Resources: rs}, nil
+	}
```
begelundmuller, line 68: If `skip_checks` is passed and the caller is not an admin, it should error instead of failing silently by applying checks anyway. For example:
```suggestion
	if req.SkipChecks {
		if !auth.GetClaims(ctx).SecurityClaims().Admin() {
			return nil, ErrForbidden
		}
		return &runtimev1.ListResourcesResponse{Resources: rs}, nil
	}
```

runtime/server/controller.go
```diff
@@ -174,11 +182,17 @@ func (s *Server) GetResource(ctx context.Context, req *runtimev1.GetResourceRequ
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 
+	isAdmin := auth.GetClaims(ctx).SecurityClaims().Admin()
+
+	if isAdmin && req.SkipChecks {
+		return &runtimev1.GetResourceResponse{Resource: r}, nil
+	}
```
begelundmuller, line 189: Same comment as above – should not fail silently

runtime/server/controller.go
```diff
@@ -174,11 +182,17 @@ func (s *Server) GetResource(ctx context.Context, req *runtimev1.GetResourceRequ
...
+
 	r, access, err := s.applySecurityPolicy(ctx, req.InstanceId, r)
 	if err != nil {
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
-	if !access {
+	if !req.SkipChecks && !access {
```
begelundmuller, line 195: I think this change is not needed given the earlier check?

begelundmuller (changes requested):

proto/rill/runtime/v1/api.proto
```diff
@@ -651,6 +651,8 @@ message ListResourcesRequest {
   string kind = 2;
   // Filter by resource path (optional).
   string path = 3;
+  // Skip security checks
+  bool skip_checks = 4;
```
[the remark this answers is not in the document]
grahamplata: I agree with that, let me make the swap

Issue #6474, opened 2025-01-22
Support user deletion


begelundmuller:
Proposed changes:

1. Add an RPC on the `AdminService` for deleting a user
    - Normal users should only be able to delete their own user
    - Superusers should be allowed to delete any user
2. Handle the case where the user being deleted is the only remaining admin of an org
    - Error if the org has other members
    - If the org has no other members, it should tear down the org and all its projects
3. Add a command `rill sudo user remove <email>` that superusers can use to invoke the RPC to delete any user

See [this Slack thread](https://rilldata.slack.com/archives/C01UFNH4W3B/p1737567808558919?thread_ts=1737503563.495859&cid=C01UFNH4W3B) for context.

grahamplata:
Proposed changes:

- [ ] Add an RPC on the AdminService for deleting a user
  - [ ] Normal users should only be able to delete their own user
  - [ ] Superusers should be allowed to delete any user
- [ ] Handle the case where the user being deleted is the only remaining admin of an org
  - [ ] Error if the org has other members
  - [ ] If the org has no other members, it should tear down the org and all its projects
- [ ] Add a command rill sudo user remove <email> that superusers can use to invoke the RPC to delete any user

grahamplata:
>  If the org has no other members, it should tear down the org and all its projects

For the above I think it could be better handled as an error and subsequent flow. ~

cc @begelundmuller

begelundmuller:
> For the above I think it could be better handled as an error and subsequent flow. ~

What do you mean by a subsequent flow? 

But overall I'm definitely open to doing it outside the RPC, e.g. by kicking off an async job or having a periodic background job that cleans up orphaned orgs/projects.

grahamplata:
> "slowness is okay since user deletion isn't going to happen very often"

That was where the thought was coming from. 

Knowing that I'll  do it within the RPC

lovincyrus:
Hey @grahamplata, I noticed that "Add a command rill sudo user remove that superusers can use to invoke the RPC to delete any user" is marked as completed. Could you link me to the PR? The change might be useful to https://github.com/rilldata/rill-private-issues/issues/1230

grahamplata:
[...]

PR #6495, opened 2025-01-23
Support user deletion and offboarding

grahamplata:
This Pull Request enables project admins the ability to query their data directly from the rill cli tool. Admins can target resolvers and issue sql queries.

Closes: https://github.com/rilldata/rill/issues/6474

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!


admin/server/users.go
```diff
@@ -284,6 +284,76 @@ func (s *Server) GetUser(ctx context.Context, req *adminv1.GetUserRequest) (*adm
...
+	}
+
+	claims := auth.GetClaims(ctx)
+
+	isCurrentUser := currentUser.User.Email == req.Email
+	isSuperuser := claims.Superuser(ctx)
+
+	if !isCurrentUser && !isSuperuser {
+		return nil, status.Error(codes.PermissionDenied, "only superusers can delete other users")
+	}
```
begelundmuller, line 316: 1. The `s.GetCurrentUser(...)` function call means one gRPC RPC handler calls another without using a real gPRC client. Even though it's probably safe here, we generally avoid that as it can mess with instrumentation and lead to edge cases where the correct middleware is not applied. Instead we prefer to pull shared logic between handlers into shared/utility functions or just duplicating the logic if it's just a few lines of code.
2. In this case, you can avoid the call by moving up the `s.admin.DB.FindUserByEmail(ctx, req.Email)` call later in the function and doing a check like `claims.OwnerType() == auth.OwnerTypeUser && claims.OwnerID() == user.ID`

admin/server/users.go
```diff
@@ -284,6 +284,76 @@ func (s *Server) GetUser(ctx context.Context, req *adminv1.GetUserRequest) (*adm
...
+
+	orgAdminUsers, err := s.admin.DB.FindOrganizationMemberUsersByRole(ctx, org.ID, role.ID)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(orgAdminUsers) == 1 && orgAdminUsers[0].Email == req.Email {
+		return nil, status.Error(codes.PermissionDenied, "cannot delete the last admin of an organization, please transfer ownership first")
```
begelundmuller, line 324:
```suggestion
		return nil, status.Error(codes.PermissionDenied, "cannot delete the last admin of an organization, please add another admin first")
```

admin/server/users.go
```diff
@@ -284,6 +284,76 @@ func (s *Server) GetUser(ctx context.Context, req *adminv1.GetUserRequest) (*adm
...
+	}
+
+	memberCount, err := s.admin.DB.CountMembersByOrganization(ctx, org.ID)
+	if err != nil {
+		return nil, err
+	}
+	if memberCount == 1 {
+		// TODO: Tear down the org and all its projects
```
begelundmuller, line 332: We need to implement this because many users will have created their own org and deployed test projects (especially the ones Himadri is looking to clean up). But if you prefer, we can merge this PR and do it in a follow up.

cli/cmd/sudo/user/remove.go
```diff
@@ -0,0 +1,59 @@
...
+			}
+
+			email := args[0]
+			if email == "" {
+				return errors.New("email is required")
+			}
+
+			if !valid(email) {
+				return errors.New("invalid email format")
+			}
```
begelundmuller, line 36: Subjective, but given the util function is only used once, consider inlining it and propagating the nested error message:

```suggestion
			if _, err := mail.ParseAddress(email); err != nil {
				return fmt.Errorf("invalid email: %w", err)
			}
```

cli/cmd/sudo/user/remove.go
```diff
@@ -0,0 +1,59 @@
...
+				err = auth.SelectOrgFlow(ctx, ch, true)
+				if err != nil {
+					return err
+				}
+			}
+
+			_, err = client.DeleteUser(ctx, &adminv1.DeleteUserRequest{
+				Email:        email,
+				Organization: ch.Org,
+			})
```
begelundmuller, line 48: A user can be a member of many orgs, so one org can't be passed here. The `DeleteUser` RPC needs to load all their memberships and check/handle each of them.

(It's okay for this to be sequential/slow, user deletion is not going to happen very often.)

begelundmuller (changes requested):

admin/server/users.go
grahamplata: Nah no need, just something I wanted clarity

cli/cmd/sudo/user/remove.go
```diff
@@ -0,0 +1,53 @@
...
+			if err != nil {
+				return fmt.Errorf("failed to list organizations for user %q: %w", email, err)
+			}
+
+			for _, org := range orgs.Organizations {
+				_, err = client.DeleteUser(ctx, &adminv1.DeleteUserRequest{Email: email, Organization: org.Id})
+				if err != nil {
+					return fmt.Errorf("failed to remove user %q from organization %q: %w", email, org.Id, err)
+				}
+			}
```
begelundmuller, line 45: Users exist separate of orgs, not nested under an org (check the database schemas) – so they can only be deleted once. Look at my previous comment, the user's org memberships probably need to be inspected and handled inside the `DeleteUser` RPC implementation.

To clarify, removing a user's _membership_ of an org is already possible using `rill user remove`. The purpose of this PR is to physically delete a user from our system (i.e. remove them from the `users` table and all references to them).

On reflection, to avoid ambiguity with `rill user remove`, let's call this command e.g. `rill user delete` (or can you think of a better name?)

begelundmuller (changes requested):

cli/cmd/sudo/user/remove.go
grahamplata: Oh! Ok that makes so much more sense. I'll leverage the schemas

cli/cmd/sudo/user/delete.go
```diff
@@ -0,0 +1,51 @@
...
+			}
+
+			_, err = client.GetUser(ctx, &adminv1.GetUserRequest{
+				Email: email,
+			})
+			if err != nil {
+				return fmt.Errorf("user %q not found: %w", email, err)
+			}
+
+			_, err = client.DeleteUser(ctx, &adminv1.DeleteUserRequest{Email: email})
```
begelundmuller, line 40: Perhaps add a `cmdutil.ConfirmPrompt` here to warn that the user and their relationships will be permanently deleted (and a `--force` CLI flag for circumventing it in scripts)

proto/rill/admin/v1/api.proto
```diff
@@ -1639,6 +1644,13 @@ message UpdateUserPreferencesResponse {
   UserPreferences preferences = 1;
 }
 
+message DeleteUserRequest {
+  string email = 1 [(validate.rules).string.email = true];
+  string organization = 2 [(validate.rules).string.min_len = 1];
```
begelundmuller, line 1649: This field doesn't make sense since the user may be a member of multiple orgs, and permanently deleting the user will/should remove all those relationships. It's also not populated by the CLI now.

admin/server/users.go
```diff
@@ -284,6 +284,59 @@ func (s *Server) GetUser(ctx context.Context, req *adminv1.GetUserRequest) (*adm
 	return &adminv1.GetUserResponse{User: userToPB(user)}, nil
 }
 
+func (s *Server) DeleteUser(ctx context.Context, req *adminv1.DeleteUserRequest) (*adminv1.DeleteUserResponse, error) {
```
begelundmuller, line 287: The RPC here still relies on `req.Organization`, which is no longer passed from the CLI, so it can't work.

begelundmuller (changes requested):

cli/cmd/sudo/user/delete.go
```diff
@@ -0,0 +1,65 @@
...
+				return err
+			}
+
+			if confirm {
+				_, err = client.DeleteUser(ctx, &adminv1.DeleteUserRequest{Email: email})
+				if err != nil {
+					return fmt.Errorf("failed to delete user %q: %w", email, err)
+				}
+				fmt.Printf("User %q deleted successfully\n", email)
+			}
```
begelundmuller, line 55: nit: Better to check `!confirm` and exit early, which enables an "aborted" message and reduces indentation:

```suggestion
			confirm, err := cmdutil.ConfirmPrompt(fmt.Sprintf("Are you sure you want to delete user %q?", email), "", false)
			if err != nil {
				return err
			}
			if !confirm {
				ch.PrintfWarn("Aborted\n")
				return nil
			}

			_, err = client.DeleteUser(ctx, &adminv1.DeleteUserRequest{Email: email})
			if err != nil {
				return fmt.Errorf("failed to delete user %q: %w", email, err)
			}
			ch.PrintfSuccess("User %q deleted successfully\n", email)
```

begelundmuller (changes requested):

cli/cmd/sudo/user/delete.go
```diff
@@ -0,0 +1,65 @@
...
+
+			return nil
+		},
+	}
+
+	deleteCmd.Flags().BoolVar(&force, "force", false, "Allows superusers to bypass certain checks")
+	_ = deleteCmd.Flags().MarkHidden("force")
+
```
begelundmuller, line 63: `force` is not checked. On reflection, let's just remove it and check against `ch.Interactive` instead (the goal here is just to be able to skip the "Confirm?" prompt when scripting)

begelundmuller (changes requested):

cli/cmd/sudo/user/delete.go
grahamplata: oh smh, can do

grahamplata:
```
go run ./cli sudo user delete graham.plata@gmail.com --interactive=false

Deleting a user is a permanent action and cannot be undone.
The user will be removed from all organizations and their data will be deleted.
User "graham.plata@gmail.com" deleted successfully```

File: cli/cmd/sudo/user/delete.go [hunk 12]
```diff
@@ -0,0 +1,71 @@
...
+			_, err = client.GetUser(ctx, &adminv1.GetUserRequest{
+				Email: email,
+			})
+			if err != nil {
+				return fmt.Errorf("user %q not found: %w", email, err)
+			}
+
+			ch.PrintfError("\nDeleting a user is a permanent action and cannot be undone.\n")
+			ch.PrintfError("The user will be removed from all organizations and their data will be deleted.\n")
```
begelundmuller, line 46: No error has happened here:

```suggestion
			ch.PrintfWarn("\nDeleting a user is a permanent action and cannot be undone.\n")
			ch.PrintfWarn("The user will be removed from all organizations and their data will be deleted.\n")
```

cli/cmd/sudo/user/delete.go
```diff
@@ -0,0 +1,71 @@
...
+
+			// If not interactive, confirm the deletion
+			if !ch.Interactive {
```
begelundmuller, line 49: Should be the reverse, right? I.e. you don't want to prompt if interactive is disabled (it's on by default).

```suggestion
			// If interactive, confirm the deletion
			if ch.Interactive {
```

begelundmuller (changes requested):

begelundmuller (approved):

PR #6496, opened 2025-01-23
fix: embedded dashboards without time series

briangregoryholmes:
Dashboards that lacked time series data were prevented from loading due to a conditional in `DashboardUrlStateSyncWrapper`

### DIFF
--- a/web-common/src/features/dashboards/url-state/DashboardURLStateSyncWrapper.svelte
+++ b/web-common/src/features/dashboards/url-state/DashboardURLStateSyncWrapper.svelte
@@ -59,9 +59,11 @@
 
   // only reactive to url and defaultExplorePreset
   $: parseUrl($page.url, defaultExplorePreset);
+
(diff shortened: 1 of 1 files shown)


AdityaHegde (approved):

web-common/src/features/dashboards/url-state/DashboardURLStateSyncWrapper.svelte
```diff
@@ -61,7 +61,7 @@
   $: parseUrl($page.url, defaultExplorePreset);
 </script>
 
-{#if !$validSpecStore.isLoading && !$metricsViewTimeRange.isLoading}
+{#if !$validSpecStore.isLoading}
```
AdityaHegde, line 64: We probably still need `metricsViewTimeRange.isLoading` when the dashboard has a timestamp. Maybe the reason E2E is failing.

PR #6493, opened 2025-01-23
Canvas: toggle filter bar visibility

djbarnwal:
Merge after https://github.com/rilldata/rill/pull/6486

Add toggle to filter bar in side panel to show/hide filter bar at top.

PR #6499, opened 2025-01-24
Canvas: Rewrite table state manangement

djbarnwal:
Merge after https://github.com/rilldata/rill/pull/6493

- Remove usage of explore state manager from canvas
- Use canvas entity spec methods for queries
- Use schema validator to validate table schema to avoid app crash when `metrics_view` is changed.
- Integrate time and filter store with table selector


web-common/src/features/canvas/stores/canvas-spec.ts
```diff
@@ -66,6 +70,13 @@ export class CanvasResolvedSpec {
       Object.keys($validSpecStore?.data?.metricsViews || {}),
     );
 
+    this.getMetricsViewFromName = (metricViewName: string) =>
```
ericpgreen2, line 73: Nit: I'd rename to `getMetricsViewValidSpecFromName`, since it returns a `validSpec`, not the whole `metricsView` resource

web-common/src/features/canvas/components/table/selector.ts
```diff
@@ -130,23 +132,25 @@ export function getTableConfig(
   );
 }
 
-export function hasValidTableSchema(
-  instanceId: string,
-  tableProperties: TableProperties,
-) {
+export function validateTableSchema(
+  ctx: StateManagers,
```
ericpgreen2, line 136: Ideally this weren't coupled to a Canvas `ctx` and rather accepted a `metricsView`

ericpgreen2 (changes requested):

djbarnwal:
Merged as part of https://github.com/rilldata/rill/pull/6438

PR #6502, opened 2025-01-24
fix: various time ranges issues in alerts

### LINKED ISSUE #6447: Alerts dialog issues
From [this Slack message](https://rilldata.slack.com/archives/CTZ8XBQ85/p1736347418250419):
> The Alerts dialog has some funky behaviour that can block the user from using it, especially if you split the data by a high cardinality field.
>  - For certain time ranges, the metric total doesn't match the real figure when certain time ranges are selected. I tried Today, Yesterday, Last 1 Week.
>  - The sample figures are sorted in ascending order, which means for high cardinality fields you only see a bit of the long tail instead of the largest dimensions. Descending order with significant values first would work better.
>     - Also, when trying to set criteria, it means that no data appear to meet even a low threshold
>  - The options to alert based on changes are greyed out, even when starting with a comparison period selected
>  - Not a bug but: it would be useful to be able to change the time period from within the alert dialog, especially when reacting to the tooltip if you started the dialog without a comparison period selected..
>  
> Tested on [this dashboard](https://ui.rilldata.com/demo/rill-openrtb-prog-ads/explore/bids_explore?view=pivot&tr=rill-PDC&compare_tr=rill-PP&grain=hour&compare_dim=campaign_name&rows=campaign_name&cols=overall_spend).

AdityaHegde:
closes #6447

- [x] Fix time zone not getting passed through for metrics aggregation query.
- [x] Sort preview by descending.
- [x] Fix time comparison not enabled in alert even when enabled while in pivot view.

### DIFF
--- a/runtime/queries/metricsview_aggregation.go
+++ b/runtime/queries/metricsview_aggregation.go
@@ -314,6 +314,9 @@ func (q *MetricsViewAggregation) rewriteToMetricsViewQuery(export bool) (*metric
 		res.IsoDuration = q.TimeRange.IsoDuration
 		res.IsoOffset = q.TimeRange.IsoOffset
 		res.RoundToGrain = metricsview.TimeGrainFromProto(q.TimeRange.RoundToGrain)
+		if q.TimeRange.TimeZone != "" {
+			qry.TimeZone = q.TimeRange.TimeZone
+		}
 		qry.TimeRange = res
 	}
 
@@ -329,6 +332,9 @@ func (q *MetricsViewAggregation) rewriteToMetricsViewQuery(export bool) (*metric
 		res.IsoDuration = q.ComparisonTimeRange.IsoDuration
 		res.IsoOffset = q.ComparisonTimeRange.IsoOffset
 		res.RoundToGrain = metricsview.TimeGrainFromProto(q.ComparisonTimeRange.RoundToGrain)
+		if q.ComparisonTimeRange.TimeZone != "" && qry.TimeZone == "qry.TimeZone" {
+			qry.TimeZone = q.ComparisonTimeRange.TimeZone
+		}
 		qry.ComparisonTimeRange = res
 	}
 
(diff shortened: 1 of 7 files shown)


web-common/src/features/dashboards/time-controls/time-range-mappers.ts
```diff
@@ -82,10 +85,11 @@ export function mapComparisonTimeRange(
   dashboardState: MetricsExplorerEntity,
   timeControlState: TimeControlState,
   timeRange: V1TimeRange | undefined,
+  forceTimeComparison: boolean,
```
ericpgreen2, line 88: I'm not following, why is this necessary? This feels like something that should be fixed upstream

AdityaHegde: This fixes the case where both time and dimension comparison is enabled. For alerts we should add time comparison in either cases.

runtime/queries/metricsview_aggregation.go
```diff
@@ -329,6 +332,9 @@ func (q *MetricsViewAggregation) rewriteToMetricsViewQuery(export bool) (*metric
 		res.IsoDuration = q.ComparisonTimeRange.IsoDuration
 		res.IsoOffset = q.ComparisonTimeRange.IsoOffset
 		res.RoundToGrain = metricsview.TimeGrainFromProto(q.ComparisonTimeRange.RoundToGrain)
+		if q.ComparisonTimeRange.TimeZone != "" && qry.TimeZone == "qry.TimeZone" {
```
begelundmuller, line 335: Why would it be `"qry.TimeZone"` as a string? It's expected to be an IANA time zone code.

Also since we don't support different time zones in `qry.TimeZone`, IMO it should return an error if `q.TimeRange.TimeZone` and `q.ComparisonTimeRange.TimeZone` specify different time zones.

begelundmuller (changes requested):

web-common/src/features/dashboards/time-controls/time-range-mappers.ts
ericpgreen2: Instead, can we delete line 92 altogether? I'm not immediately seeing why `selectedComparisonDimension` should remove the comparison time range.

begelundmuller (approved):
Backend looks good now

web-common/src/features/dashboards/time-controls/time-range-mappers.ts
AdityaHegde: Hmmm I dont see any issue with that. It was just some old code to remove time comparison when dimension comparison is enabled.

@mindspank Ok to remove this check? It affects scheduled reports for TDD and dimension table.

mindspank: Should be fine

ericpgreen2 (approved):

PR #6505, opened 2025-01-24
feat: add quick actions for pivot

mindspank:
Adds quick add actions for pivot pills

https://www.notion.so/rilldata/Quick-add-Dimension-and-Measures-in-Pivot-15aba33c8f5780c482b4ddfd8a35dd52?pvs=4


web-common/src/features/dashboards/pivot/DragList.svelte
```diff
@@ -147,6 +155,14 @@
     $ghostIndex = null;
     swap = false;
   }
+
+  function handleRowClick(item: PivotChipData) {
```
mindspank, line 159: Duplicated the methods here as I anticipate we will add more row/column related logic once we work on flat/nested tables

briangregoryholmes (approved):

PR #6508, opened 2025-01-27
feat: Add monthly report support

### LINKED ISSUE #6503: Scheduled Reports: Add option to issue report monthly
[Slack request here](https://rilldata.slack.com/archives/CTCJ58H3M/p1737680685355149)

AdityaHegde:
closes #6503


ericpgreen2 (changes requested):
UXQA: 
1. The "Create" button should be disabled until the all the form's required fields are input & when there's input validation errors.
2. The "Invalid" text should be left-aligned with the main body of the form (and it should be a more informative message).
    image
3. The "repeats" section should be written in more natural English, e.g. "on the 31st of each month" 
    image
4. I see that when the next month doesn't have the provided day, like February 30th, the report is skipped for that month. @mindspank, can you please do the Product review for this PR? Is this what you'd expect?
    image

mindspank:
> I see that when the next month doesn't have the provided day, like February 30th, the report is skipped for that month.

I would expect the report to run on the 28th in this case. Same for reports scheduled to run on 31st I would expect to run on the 30th if there is no 31 for the next month.

AdityaHegde:
Disabled the input for now.

web-common/src/features/scheduled-reports/time-utils.ts
ericpgreen2, line 1: `convertFormValuesToCronExpression()` and `getFrequencyFromCronExpression()` could use some unit tests

ericpgreen2 (approved):

web-common/src/features/scheduled-reports/time-utils.ts
AdityaHegde: True. None of this code has any tests.

PR #6509, opened 2025-01-27
fix: rill time issues around truncate and arithmatic

AdityaHegde:
1. Fix truncate when arithmetic is involved. For example a value like `0W, 0W+1d` was being incorrectly truncated by week for both start and end.
2. Support `@ now`, `@ latest` `@ 2024-02-09` modifiers to apply the full time range for now/latest/abs time. EG: Support `0W, 0W+1D @ now`, should return beginning of week to a day after beginning of week relative to now.
3. Support case insensitive grains for day, week, quarter and year.
4. Fix `timeutil.Truncate` when `firstMonth` is invalid.


begelundmuller (approved):
LGTM

PR #6513, opened 2025-01-28
Set max date for public url form

lovincyrus:
closes https://github.com/rilldata/rill-private-issues/issues/1149

CleanShot 2025-01-27 at 16 17 04@2x


ericpgreen2 (dismissed):
UXQA: After I set an expiration date, clicking outside of the popover does not close the popover, as I'd expect.

ericpgreen2 (approved):

PR #6517, opened 2025-01-28
Update Project parser to support commit timestamps

grahamplata:
This PR updates the project parser to support commit timestamps (exposed as `projectParser.state.currentCommitOn`). Enabling the UI to display the last timestamp the source project files were updated.

Git-connected projects, will utilize the current commit's timestamp, while non-Git deploys that will be the timestamp when the project files being served were uploaded.


runtime/reconcilers/project_parser.go
```diff
@@ -116,6 +117,23 @@ func (r *ProjectParserReconciler) Reconcile(ctx context.Context, n *runtimev1.Re
...
+	if err != nil {
+		r.C.Logger.Error("failed to get commit timestamp", zap.String("error", err.Error()))
+	}
+	// Convert to timestamppb.Timestamp
+	timeStamp := timestamppb.New(ts)
+
+	// Update the commit timestamp
+	if pp.State.CurrentCommitOn != timeStamp {
+		pp.State.CurrentCommitOn = timeStamp
+		err = r.C.UpdateState(ctx, n, self)
```
begelundmuller, line 131: IMO it's fine to skip the check for whether the commit time changed, and just update both fields if the hash changed. The commit hash and timestamp will almost always change in unison.

Also, merging merging this update logic with the `UpdateState` call for `CurrentCommitSha` below would be nice as there would be one less DB write roundtrip.

begelundmuller (changes requested):

runtime/reconcilers/project_parser.go
grahamplata: Sounds good to me, I will make that change

runtime/drivers/admin/repo.go
```diff
@@ -26,6 +27,38 @@ func (h *Handle) Root(ctx context.Context) (string, error) {
...
+		return time.Time{}, err
+	}
+
+	defer h.repoMu.RUnlock()
+
+	if h.downloadURL != "" {
+		// use downloadURL as a proxy for CommitTimestamp for one-time uploads
+		// It will change when new data is uploaded
+		return time.Now(), nil
+	}
```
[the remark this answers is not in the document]
grahamplata: @begelundmuller that makes sense to me. Made some adjustments

proto/rill/admin/v1/api.proto
```diff
@@ -1922,11 +1922,13 @@ message GetRepoMetaRequest {
...
 message GetRepoMetaResponse {
   string git_url = 1;
+  google.protobuf.Timestamp created_on = 5;
   google.protobuf.Timestamp git_url_expires_on = 2;
   string git_subpath = 3;
   // archive_download_url is set when repo is managed by Rill
   // either archive_download_url or git related details will be set
   string archive_download_url = 4;
+  string asset_id = 6;
 }
```
begelundmuller, line 1932: It's not clear that `created_on` and `asset_id` are coupled to the `archive_download_url`. What about something like this:

```suggestion
  // If the Git-related fields are set, the archive-related fields will not be set (and vice versa).
  string git_url = 1;
  google.protobuf.Timestamp git_url_expires_on = 2;
  string git_subpath = 3;
  string archive_download_url = 4;
  string archive_id = 5;
  google.protobuf.Timestamp archive_created_on = 6;
```

runtime/drivers/admin/admin.go
```diff
@@ -125,6 +125,8 @@ type Handle struct {
 
 	// downloadURL is set when using one-time uploads
 	downloadURL string
+	assetID     string
+	createdOn   time.Time
```
begelundmuller, line 129: Similar to the API signature, perhaps adding `archive` prefixes here might also be helpful to disambiguate

runtime/drivers/admin/repo.go
```diff
@@ -26,6 +27,38 @@ func (h *Handle) Root(ctx context.Context) (string, error) {
...
+	if err != nil {
+		return time.Time{}, err
+	}
+
+	defer h.repoMu.RUnlock()
+
+	if h.downloadURL != "" {
+		// use downloadURL as a proxy for CommitTimestamp for one-time uploads
+		// It will change when new data is uploaded
+		return h.createdOn, nil
```
begelundmuller, line 41: The comment doesn't quite fit here

runtime/reconcilers/project_parser.go
```diff
@@ -124,6 +133,7 @@ func (r *ProjectParserReconciler) Reconcile(ctx context.Context, n *runtimev1.Re
 	}
 	if pp.State.CurrentCommitSha != hash {
 		pp.State.CurrentCommitSha = hash
+		pp.State.CurrentCommitOn = timeStamp
```
begelundmuller, line 136: nit:
```suggestion
		pp.State.CurrentCommitOn = timestamppb.New(ts)
```

begelundmuller (changes requested):

runtime/drivers/admin/admin.go
```diff
@@ -124,7 +124,9 @@ type Handle struct {
 	gitURLExpiresOn time.Time
 
 	// downloadURL is set when using one-time uploads
-	downloadURL string
+	downloadURL      string
+	ArchiveID        string
+	ArchiveCreatedOn time.Time
```
begelundmuller, line 129: nit: first letter should be lowercase for private fields (that are not accessed outside of the package)

runtime/drivers/admin/admin.go
```diff
@@ -405,6 +407,14 @@ func (h *Handle) checkHandshake(ctx context.Context) error {
...
+		h.ArchiveID = meta.ArchiveId
+	}
+
+	if meta.ArchiveCreatedOn != nil {
+		h.ArchiveCreatedOn = meta.ArchiveCreatedOn.AsTime()
+	}
+
 	if meta.ArchiveDownloadUrl != "" {
 		h.downloadURL = meta.ArchiveDownloadUrl
 		return nil
```
begelundmuller, line 420: Possibility for some cleanup here:

```suggestion
	h.archiveID = meta.ArchiveId
	h.archiveCreatedOn = time.Time{}
	if meta.ArchiveCreatedOn != nil {
		h.archiveCreatedOn = meta.ArchiveCreatedOn.AsTime()
	}
	h.downloadURL = meta.ArchiveDownloadUrl
```

PR #6518, opened 2025-01-28
Support custom attributes in `mock_users` and `IssueDevJWT`

begelundmuller:


medriscoll:
@nishantmonu51 can you review?

ericpgreen2 (commented):
The `mock_users` and `IssueDevJWT` part looks good to me. But are we allowed to reference custom attributes like this?

image

begelundmuller:
> The `mock_users` and `IssueDevJWT` part looks good to me. But are we allowed to reference custom attributes like this?

Yes you are allowed to reference it. But if you want the security policy to work without the attribute set (as I assume is the case here, i.e. you're not using a mock user that has that attribute?), you need to write it more defensively, something like this:
```
security:
  access: '{{ if hasKey .user "custom_attribute" }} {{ .user.custom_attribute }} {{ else }} FALSE {{ end}}'
```

The problem is that the templating engine outputs the string `<no value>` when a field is not populated. So your access check expression here becomes the following DuckDB statement
```
SELECT <no value>
```
giving the syntax error you see here. This gotcha is difficult for us to handle as it is just templating+DuckDB. But if you have any ideas for improvement let me know.

ericpgreen2:
Ah, okay, that makes sense. Yes, I'm able to get a working example using that defensive logic. I'll follow this PR with one of my own to pass these custom attributes into the `IssueDevJWT` API call.

PR #6520, opened 2025-01-29
fix: public urls with computed measures and dependant measures are hidden

AdityaHegde:
If a public url is created by selecting a computed measure but not the measure it is computed on, the dashboard fails to load since all required measures are not added to the token.

~~So we need to add all `referencedMeasures` to fields as well. This has a side effect of showing measures that were not selected.~~

We do not need to send the `referencedMeasures` for non-window computed measures.

### DIFF
--- a/web-common/src/features/dashboards/state-managers/selectors/measures.ts
+++ b/web-common/src/features/dashboards/state-managers/selectors/measures.ts
@@ -82,10 +82,12 @@ export const filteredSimpleMeasures = ({
 };
 
 /**
- * Selects measure valid for current dashboard selections.
- * Also includes additional dimensions needed for any advanced measures.
+ * Selects measure valid for current dashboard selections. We filter out advanced measures that are,
+ * 1. Of type MEASURE_TYPE_TIME_COMPARISON.
+ * 2. Dependent on a time dimension with a defined grain and not equal to the current selected grain.
+ * 3. Window measures if includeWindowMeasures=false. Right now totals query does not support these.
  */
-export const getFilteredMeasures = (
+export const removeSomeAdvancedMeasures = (
   exploreState: MetricsExplorerEntity,
   metricsViewSpec: V1MetricsViewSpec,
   measureNames: string[],
--- a/web-common/src/features/dashboards/state-managers/selectors/measures.ts
+++ b/web-common/src/features/dashboards/state-managers/selectors/measures.ts
@@ -1,5 +1,5 @@
+import type { MetricsExplorerEntity } from "@rilldata/web-common/features/dashboards/stores/metrics-explorer-entity";
 import {
-  type MetricsViewSpecDimensionSelector,
   MetricsViewSpecMeasureType,
   type MetricsViewSpecMeasureV2,
   type V1MetricsViewSpec,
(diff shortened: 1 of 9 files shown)


begelundmuller:
FYI it's `referencedMeasures`, not `requiredMeasures` (maybe getting confused with `requiredDimensions`)

AdityaHegde:
I used `referencedMeasures` in code but mixed it up in the description, updated now.

web-common/src/features/dashboards/state-managers/selectors/measures.ts
```diff
@@ -145,10 +147,11 @@ export const getFilteredMeasuresAndDimensions = ({
       });
       if (skipMeasure) return;
       measures.add(measureName);
-      measure.referencedMeasures?.filter((refMes) => measures.add(refMes));
+      if (!measure.window) nonWindowMeasures.add(measureName);
```
ericpgreen2, line 150: I'm having trouble following this function. 
1. What is a "filtered" measure and dimension?
2. I'm nervous about this line. Adding measures to the list of user-selected measures feels like exactly what caused the original bug. Double-checking... are you sure this is necessary?

AdityaHegde: Adding these measures break the totals query, since we dont really get totals data anyway (`no data` can be seen) it is better to filter them out.

Would be great to add E2E once the setup PR is merged.

ericpgreen2: It seems one way this function is overloaded is that it's preparing two different query requests: `totals` and `timeseries`, which each have a different set of measures & dimensions.

Can we split this into two functions:
1. `getMeasuresAndDimensionsForTimeSeriesRequest()`
2. `getMeasuresForTotalsRequest()`

AdityaHegde: Updated the method. I dont think the method should differentiate between timeseries vs totals request. Added a param to include/exclude window measures.

ericpgreen2 (approved):

web-common/src/features/dashboards/state-managers/selectors/measures.ts
```diff
@@ -82,95 +82,49 @@ export const filteredSimpleMeasures = ({
...
+        // filter out measures with dependant dimensions not matching the selected grain
+        skipMeasure = true;
+        return;
+      }
+    });
+    if (skipMeasure) return;
+
     measures.add(measureName);
-    measure.referencedMeasures?.filter((refMes) => measures.add(refMes));
   });
```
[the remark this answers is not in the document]
AdityaHegde: We dont need this anymore. That was the big fix in the PR. The earlier code should have been a `forEach`.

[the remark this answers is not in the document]
ericpgreen2: Sorry, GitHub's UI made this comment confusIng... I meant the whole code block that's highlighted. At the top-level, this code is initializing an empty set `const measures = new Set()`, then building up an array. Rather, I'd expect the input `measureNames` to be filtered like `measureNames.filter(...)`.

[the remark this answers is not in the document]
AdityaHegde: Ah, will change in a follow up PR.

PR #6521, opened 2025-01-29
feat: Custom logo UI support

AdityaHegde:
- [x] Show logo in nav bar when it is set for the org. (Moved to #6530)
- [x] Add a logo editing setting component
- [x] Support uploading logo from UI. (Need to figure out the CORS issue while uploading)
- [x] Support removing the log from UI.


ericokuma:
Some UXQA feedback:

1. Instead of 
image
Can we have a preview of the image I uploaded? Here's how Notion does it:

Screenshot 2025-01-30 at 8 55 55 AM

The one on the left is light mode, right is dark mode. Because we don't have dark mode for now, we can just show the light mode preview

2. When I first click on the logo edit icon, the modal appears below
Screenshot 2025-01-30 at 8 57 10 AM

After I upload an image, it jumps above
Screenshot 2025-01-30 at 8 57 56 AM

3. Scaling. I'm not sure if the logo upload is scaling properly. I uploaded a .png file and the new logo is much smaller than the default Rill logo

Screenshot 2025-01-30 at 8 59 07 AM

vs

Screenshot 2025-01-30 at 8 59 20 AM

PR #6530, opened 2025-01-30
feat: Show custom logo in UI

AdityaHegde:
Separates out the "showing" part from #6521 

Add a logo using `rill org upload-logo <org> <path-to-img>`. The logo will show up in the top nav bar.

### DIFF
--- a/web-admin/src/routes/+layout.ts
+++ b/web-admin/src/routes/+layout.ts
@@ -17,6 +17,7 @@ import {
   type V1User,
 } from "@rilldata/web-admin/client";
 import { redirectToLoginOrRequestAccess } from "@rilldata/web-admin/features/authentication/checkUserAccess";
+import { getFetchOrganizationQueryOptions } from "@rilldata/web-admin/features/organizations/selectors";
 import { fetchProjectDeploymentDetails } from "@rilldata/web-admin/features/projects/selectors";
 import { initPosthog } from "@rilldata/web-common/lib/analytics/posthog";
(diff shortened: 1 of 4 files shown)


web-admin/src/routes/+layout.ts
```diff
@@ -75,10 +76,16 @@ export const load = async ({ params, url, route }) => {
...
+  let organizationLogoUrl: string | undefined = undefined;
   if (organization && !token) {
     try {
-      organizationPermissions =
-        await fetchOrganizationPermissions(organization);
+      const organizationResp = await queryClient.fetchQuery({
+        queryKey: getAdminServiceGetOrganizationQueryKey(organization),
+        queryFn: () => adminServiceGetOrganization(organization),
+        staleTime: Infinity,
+      });
```
ericpgreen2, line 86: Nit/idea: if interested in abstracting out the guts of a `fetchQuery` call like this one, we could factor out the `queryOptions` into a separate file, like [this blog post](https://tkdodo.eu/blog/the-query-options-api) advocates

ericpgreen2: Remind me why `staleTime: Infinity` is needed?

ericpgreen2 (approved):
Code looks good. @mindspank, want to do the Product review?

web-admin/src/routes/+layout.ts
AdityaHegde: Every switch of org fetches the org details otherwise. Seems redundant to me. There are a few more of those we should probably clamp on.

PR #6534, opened 2025-01-30
Emiting reconcile_elapsed_ms metrics

NamanMahor:
Emiting reconcile_elapsed_ms metrics

### DIFF
--- a/runtime/controller.go
+++ b/runtime/controller.go
@@ -1359,28 +1359,30 @@ func (c *Controller) processCompletedInvocation(inv *invocation) error {
 	}
 
 	commonDims := []attribute.KeyValue{
-		attribute.String("resource_id", inv.name.Name),
+		attribute.String("resource_name", inv.name.Name),
 		attribute.String("resource_type", PrettifyResourceKind(inv.name.Kind)),
 	}
+
 	if inv.isDelete {
-		commonDims = append(commonDims, attribute.Bool("is_deleted", true))
-	}
-	if inv.isRename {
-		commonDims = append(commonDims, attribute.Bool("is_renamed", true))
+		commonDims = append(commonDims, attribute.String("reconcile_operation", "delete"))
+	} else if inv.isRename {
+		commonDims = append(commonDims, attribute.String("reconcile_operation", "rename"))
+	} else {
+		commonDims = append(commonDims, attribute.String("reconcile_operation", "normal"))
 	}
 
 	if !inv.cancelledOn.IsZero() {
-		commonDims = append(commonDims, attribute.String("reconcile_status", "Cancelled"))
+		commonDims = append(commonDims, attribute.String("reconcile_result", "canceled"))
 	} else if inv.result.Err != nil {
 		if errors.Is(inv.result.Err, context.Canceled) {
-			commonDims = append(commonDims, attribute.String("reconcile_status", "Cancelled"))
+			commonDims = append(commonDims, attribute.String("reconcile_result", "canceled"))
 		} else if errors.Is(inv.result.Err, context.DeadlineExceeded) {
-			commonDims = append(commonDims, attribute.String("reconcile_status", "Timeout"))
+			commonDims = append(commonDims, attribute.String("reconcile_result", "timeout"))
 		} else {
-			commonDims = append(commonDims, attribute.String("reconcile_status", "Error"), attribute.String("reconcile_error", inv.result.Err.Error()))
+			commonDims = append(commonDims, attribute.String("reconcile_result", "error"), attribute.String("reconcile_error", inv.result.Err.Error()))
 		}
 	} else {
-		commonDims = append(commonDims, attribute.String("reconcile_status", "Succeeded"))
+		commonDims = append(commonDims, attribute.String("reconcile_result", "success"))
 	}
 
 	if !inv.result.Retrigger.IsZero() {
--- a/runtime/controller.go
+++ b/runtime/controller.go
@@ -1333,7 +1333,7 @@ func (c *Controller) processCompletedInvocation(inv *invocation) error {
 	delete(c.invocations, nameStr(inv.name))
 
 	// Emit metrics
-	reconcile_time_ms := time.Since(inv.startedOn).Milliseconds()
+	reconcileTimeMs := time.Since(inv.startedOn).Milliseconds()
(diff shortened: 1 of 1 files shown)


k-anshul (changes requested):
Hey

So we already have the metric being [logged](https://github.com/rilldata/rill/blob/baa6274cf3a53ab8c468aeb658819c74ebbba5ab/runtime/controller.go#L1379-L1380). But there are some changes that needs to be done.
1. It should be a `EventTypeMetric` and not log.
2. Needs to log for cancellations as well.
3. Need to separate out normal errors from timeouts.
4. Need to include `retrigger_time`.
5. Need to include rill_version. It is not present in runtime as of now but you can populate the same by making changes similar to admin service here : https://github.com/rilldata/rill/blob/baa6274cf3a53ab8c468aeb658819c74ebbba5ab/cli/cmd/admin/start.go#L292-L293

NamanMahor:
> Hey
> 
> So we already have the metric being [logged](https://github.com/rilldata/rill/blob/baa6274cf3a53ab8c468aeb658819c74ebbba5ab/runtime/controller.go#L1379-L1380). But there are some changes that needs to be done.
> 
> 1. It should be a `EventTypeMetric` and not log.
> 2. Needs to log for cancellations as well.
> 3. Need to separate out normal errors from timeouts.
> 4. Need to include `retrigger_time`.
> 5. Need to include rill_version. It is not present in runtime as of now but you can populate the same by making changes similar to admin service here : https://github.com/rilldata/rill/blob/baa6274cf3a53ab8c468aeb658819c74ebbba5ab/cli/cmd/admin/start.go#L292-L293

@k-anshul rill_version should already be there because it got added in [activety](https://github.com/rilldata/rill/blob/main/cli/cmd/runtime/start.go#L199) 
I added new EvenTypeMetric(which handled retrigger_time n cancellations ) but did't removed the log type which was there.
 Seems like we do not need log so removed it in latest commit. 
And also Handling timeout using DeadlineExceeded error.

runtime/controller.go
```diff
@@ -1360,24 +1358,35 @@ func (c *Controller) processCompletedInvocation(inv *invocation) error {
...
-		if inv.isDelete {
-			eventArgs = append(eventArgs, attribute.Bool("deleted", true))
-		}
-		if inv.isRename {
-			eventArgs = append(eventArgs, attribute.Bool("renamed", true))
-		}
-		if inv.result.Err != nil {
-			eventArgs = append(eventArgs, attribute.String("error", inv.result.Err.Error()))
+	commonDims := []attribute.KeyValue{
+		attribute.String("resource_id", inv.name.Name),
```
begelundmuller, line 1362:
```suggestion
		attribute.String("resource_name", inv.name.Name),
```

runtime/controller.go
```diff
@@ -1360,24 +1358,35 @@ func (c *Controller) processCompletedInvocation(inv *invocation) error {
...
+		attribute.String("resource_type", PrettifyResourceKind(inv.name.Kind)),
+	}
+	if inv.isDelete {
+		commonDims = append(commonDims, attribute.Bool("is_deleted", true))
+	}
+	if inv.isRename {
+		commonDims = append(commonDims, attribute.Bool("is_renamed", true))
+	}
```
begelundmuller, line 1370: Grammar:
```suggestion
	if inv.isDelete {
		commonDims = append(commonDims, attribute.Bool("is_delete", true))
	}
	if inv.isRename {
		commonDims = append(commonDims, attribute.Bool("is_rename", true))
	}
```

runtime/controller.go
```diff
@@ -1360,24 +1358,35 @@ func (c *Controller) processCompletedInvocation(inv *invocation) error {
...
+
+	if !inv.cancelledOn.IsZero() {
+		commonDims = append(commonDims, attribute.String("reconcile_status", "Cancelled"))
```
begelundmuller, line 1373: `reconcile_status` is not a good key here – we already have something called `ReconcileStatus` and it takes different values: https://github.com/rilldata/rill/blob/d4fa2a11a540d2745d056abf8ed10bfc1980072b/proto/rill/runtime/v1/resources.proto#L12-L17

Maybe consider `reconcile_result` instead?

runtime/controller.go
```diff
@@ -1360,24 +1358,35 @@ func (c *Controller) processCompletedInvocation(inv *invocation) error {
...
+		if errors.Is(inv.result.Err, context.Canceled) {
+			commonDims = append(commonDims, attribute.String("reconcile_status", "Cancelled"))
+		} else if errors.Is(inv.result.Err, context.DeadlineExceeded) {
+			commonDims = append(commonDims, attribute.String("reconcile_status", "Timeout"))
+		} else {
+			commonDims = append(commonDims, attribute.String("reconcile_status", "Error"), attribute.String("reconcile_error", inv.result.Err.Error()))
 		}
-		c.Activity.Record(context.Background(), activity.EventTypeLog, "reconciled_resource", eventArgs...)
+	} else {
+		commonDims = append(commonDims, attribute.String("reconcile_status", "Succeeded"))
```
begelundmuller, line 1383: nit: call it "success" instead of "Succeeded"

runtime/controller.go
```diff
@@ -1360,24 +1358,35 @@ func (c *Controller) processCompletedInvocation(inv *invocation) error {
...
+	}
+	if inv.isRename {
+		commonDims = append(commonDims, attribute.Bool("is_renamed", true))
+	}
+
+	if !inv.cancelledOn.IsZero() {
+		commonDims = append(commonDims, attribute.String("reconcile_status", "Cancelled"))
+	} else if inv.result.Err != nil {
+		if errors.Is(inv.result.Err, context.Canceled) {
+			commonDims = append(commonDims, attribute.String("reconcile_status", "Cancelled"))
```
begelundmuller, line 1376: Given the key is snake case, consider also using snake case for the values (this seems like pascal case or title case). E.g. `canceled` instead of `Cancelled`.

runtime/controller.go
NamanMahor: added reconcile_operation instead of two different dimensions

runtime/controller.go
NamanMahor: done

runtime/controller.go
NamanMahor: done

PR #6540, opened 2025-01-30
fix : do not log cancellation errors in worker jobs

k-anshul:


admin/worker/worker.go
```diff
@@ -107,7 +108,7 @@ func (w *Worker) RunJob(ctx context.Context, name string) error {
 func (w *Worker) schedule(ctx context.Context, name string, fn func(context.Context) error, every time.Duration) error {
 	for {
 		err := w.runJob(ctx, name, fn)
-		if err != nil {
+		if err != nil && !errors.Is(err, context.Canceled) {
```
begelundmuller, line 111: What about line 156?

begelundmuller (changes requested):

begelundmuller (approved):

PR #6572, opened 2025-02-04
Update invite page button logic to change when inviting or enabling domain allowlisting

ericokuma:
On a first time deploy flow, the user invitation page, currently, the skip button turns into a Continue only when a user has invited someone. 

This ticket is requesting that the button logic be updated so that it changes from `skip` to `continue` when a user either invites someone or domain allowlisting is enabled


ericpgreen2 (changes requested):
Check out the Files Changed :)

ericpgreen2 (approved):

PR #6580, opened 2025-02-04
fix: dimension uri coming in as int instead of bool breaking UI

AdityaHegde:
For certain cases dimension uri comes in as `0/1` instead of `false/true`. Until we figure out the query/data issue this fixes the UI crashing.

### DIFF
--- a/web-common/src/features/dashboards/leaderboard/leaderboard-utils.spec.ts
+++ b/web-common/src/features/dashboards/leaderboard/leaderboard-utils.spec.ts
@@ -1,0 +1,22 @@
+import { makeHref } from "@rilldata/web-common/features/dashboards/leaderboard/leaderboard-utils";
+import { describe, it, expect } from "vitest";
+
+describe("makeHref", () => {
+  const TestCases: [
+    string | number | boolean | null,
+    string,
+    string | undefined,
+  ][] = [
+    [null, "http://localhost/dim", undefined],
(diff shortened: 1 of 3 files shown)


web-common/src/features/dashboards/leaderboard/leaderboard-utils.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+  const TestCases: [
+    string | number | boolean | null,
+    string,
+    string | undefined,
+  ][] = [
+    [null, "http://localhost/dim", undefined],
+    ["http://localhost", "http://localhost/dim", "http://localhost"],
+    [true, "http://localhost/dim", "http://localhost/dim"],
+    [1, "http://localhost/dim", "http://localhost/dim"],
+  ];
```
ericpgreen2, line 14: The test cases would be more readable if we used a descriptive object rather than a raw tuple:
```
const testCases = [
    { uri: null, dimValue: "http://localhost/dim", expected: undefined },
    { uri: "http://localhost", dimValue: "http://localhost/dim", expected: "http://localhost" },
    { uri: true, dimValue: "http://localhost/dim", expected: "http://localhost/dim" },
    { uri: 1, dimValue: "http://localhost/dim", expected: "http://localhost/dim" },
  ];
```

royendo:
indeed uri isnt part of the docs, i can add that in

PR #6592, opened 2025-02-05
fix : minor tweaks to temporary view cleanups during clickhouse modeling

k-anshul:
closes https://github.com/rilldata/rill-private-issues/issues/1112

**Note** : This is a best effort cleanup and there are chances that views can still linger.

### DIFF
--- a/runtime/drivers/clickhouse/olap.go
+++ b/runtime/drivers/clickhouse/olap.go
@@ -8,11 +8,11 @@ import (
 	"strings"
 	"time"
 
-	"github.com/google/uuid"
 	"github.com/jmoiron/sqlx"
 	"github.com/mitchellh/mapstructure"
 	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
 	"github.com/rilldata/rill/runtime/drivers"
+	"github.com/rilldata/rill/runtime/pkg/graceful"
 	"github.com/rilldata/rill/runtime/pkg/observability"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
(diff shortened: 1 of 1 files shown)


runtime/drivers/clickhouse/olap.go
```diff
@@ -269,33 +269,29 @@ func (c *connection) InsertTableAsSelect(ctx context.Context, name, sql string,
...
-			if ctx.Err() != nil {
-				ctx, cancel = context.WithTimeout(context.Background(), 15*time.Second)
-			} else {
-				cancel = func() {}
-			}
+			// cleanup using a different ctx to prevent cleanups being impacted by the main ctx cancellation
+			// this is a best effort cleanup and query can still timeout and we don't want to wait forever due to blocked calls
+			// this is triggered before the table is even created to handle situations
+			// where before the client can trigger query cancel the query succeeds and the view is created but the driver stil reports query cancelled
+			ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
```
begelundmuller, line 278: Did you consider using `graceful.WithMinimumDuration`? https://github.com/rilldata/rill/blob/d2fa07e2cf1b02ab5191d90f91fe1d1112231e0c/runtime/pkg/graceful/context.go#L31

runtime/drivers/clickhouse/olap.go
```diff
@@ -569,15 +565,19 @@ func (c *connection) createTable(ctx context.Context, name, sql string, outputPr
 		}
 		// infer columns
 		v := tempName("view")
```
begelundmuller, line 567: I don't think it's possible, but can you see a way that we could use a consistent temporary name (instead of a random one)? Then if cleanup fails in one invocation, it would at least be cleaned up in the next successful invocation.

begelundmuller (approved):

runtime/drivers/clickhouse/olap.go
k-anshul: No I did not and that sounds like a much better idea 👍

runtime/drivers/clickhouse/olap.go
k-anshul: I think `InsertTableAsSelect` can be called in parallel as well. Clickhouse driver allows executing partitions in parallel so getting a consistent name with just name can be problematic.
We could generate one with name and sql combination. From the existing code I see this : https://github.com/rilldata/rill/blob/baa6274cf3a53ab8c468aeb658819c74ebbba5ab/runtime/drivers/clickhouse/olap.go#L271-L272
I believe something like this could work ?

runtime/drivers/clickhouse/olap.go
begelundmuller: Did you forget to push this change? :)

runtime/drivers/clickhouse/olap.go
begelundmuller: I'm up for that if you think it's a good idea?

runtime/drivers/clickhouse/olap.go
k-anshul: My bad. Yes. :)

PR #6599, opened 2025-02-05
Canvas: Time improvements and stability fixes

djbarnwal:
- Mimics TimeControlsStore for canvas using independent stores
- Rename class and variables
- Fix validation error in charts when Time is picked
- Allow empty measure for Table


web-common/src/features/canvas/stores/time-control.ts
```diff
@@ -0,0 +1,331 @@
...
+          this.selectedComparisonTimeRange.set(newComparisonRange);
+          this.showTimeComparison.set(true);
+        }
+
+        this.selectedTimeRange.set(newTimeRange);
+      },
+    );
+
+    // Subscribe to ensure the derived code runs
+    store.subscribe(() => {});
```
AdityaHegde, line 214: Dont we need to unsubscribe when the entity is removed/navigated from? Might lead to unintended callbacks.

AdityaHegde (approved):
Feel free to merge, once the issue of missing unsub.

PR #6610, opened 2025-02-06
feat: add dashboard banner

mindspank:
Adds a `banner` key to the `explore.yaml`.
Rill Cloud will display the banner whenever a dashboard is viewed in Rill Cloud.

Edge cases:

- When a Rill Admin on a trial plan directly navigates to a dashboard route the trial banner will overwrite the dashboard banner, race condition where the billing banner loads slower
- When a Rill Admin on a trial plan navigates away from a dashboard with a banner there is a slight delay before the trial banner is fetched again


begelundmuller (commented):
Backend LGTM

web-admin/src/routes/[organization]/[project]/explore/[dashboard]/+page.svelte
```diff
@@ -76,11 +76,25 @@
     });
   }
 
+  // Display a dashboard banner
+  $: if ($explore.data?.explore?.explore?.state?.validSpec?.banner) {
```
Comment (ericpgreen2) [inline web-admin/src/routes/[organization]/[project]/explore/[dashboard]/+page.svelte:80]: Nit: since you're repeating this twice, it'd be marginally neater to pull it out as an explanatory variable:
```
$: hasBanner = !!$explore.data?.explore?.explore?.state?.validSpec?.banner
```

Then also, the below `$explore...banner` does not need optional chaining.

ericpgreen2 (approved):

PR #6615, opened 2025-02-07
Canvas: Chart improvements

djbarnwal:
- Non zero origin for measure by default
- Add a option in dropdown to use 0 based origin
- More descriptive chart title
- Remove interactivity from lenged
- Show auto chart title if description is added


lovincyrus (approved):
ack

PR #6616, opened 2025-02-07
Tweak: Do not show "New source" prompt if source has previously been ingested

ericpgreen2:
Addresses [this Slack conversation](https://rilldata.slack.com/archives/C02T907FEUB/p1738871868841619)

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!

### DIFF
--- a/web-common/src/features/entity-management/WatchResourcesClient.ts
+++ b/web-common/src/features/entity-management/WatchResourcesClient.ts
@@ -187,7 +187,8 @@ export class WatchResourcesClient {
             // If it's a new source, show the "Source imported successfully" modal
             const isNewSource =
               res.name.kind === ResourceKind.Source &&
-              res.resource.meta.specVersion === "1" &&
+              res.resource.meta.specVersion === "1" && // First file version
+              res.resource.meta.stateVersion === "2" && // First ingest is complete
               (await isLeafResource(res.resource, this.instanceId)); // Protects against existing projects reconciling anew
             if (isNewSource) {
               const filePath = res.resource?.meta?.filePaths?.[0] as string;


medriscoll:
Can we also include a 'Cancel' button here or X in the upper right of the model?  Right now user is forced to View Source or Generate a Dashboard.

ericpgreen2:
> Can we also include a 'Cancel' button here or X in the upper right of the model? 

Yep, just added an "X" button.

image

web-common/src/features/entity-management/WatchResourcesClient.ts
```diff
@@ -187,7 +187,8 @@ export class WatchResourcesClient {
             // If it's a new source, show the "Source imported successfully" modal
             const isNewSource =
               res.name.kind === ResourceKind.Source &&
-              res.resource.meta.specVersion === "1" &&
+              res.resource.meta.specVersion === "1" && // First file version
+              res.resource.meta.stateVersion === "2" && // First ingest is complete
```
AdityaHegde, line 191: Is this `2` because `1` will be when it is queued for reconcile?

Also would it always be spec=1 and state=2 when the source gets successfully ingested for the 1st time? Like if the source had an error, the user fixed it in later iterations?

ericpgreen2: > Is this 2 because 1 will be when it is queued for reconcile?

Exactly

> Also would it always be spec=1 and state=2 when the source gets successfully ingested for the 1st time? Like if the source had an error, the user fixed it in later iterations?

If the user later fixed it in-editor, then spec != 1. In this case, if we wanted to show a "source ingested for the first time" pop-up, I'd argue it should not be a full-screen blocking pop-up, but rather a less obstructive toast notification. We also have the top-right "Create Dashboard" CTA that is useful here.

AdityaHegde (approved):

PR #6624, opened 2025-02-07
Enable admins to directly query the OLAP database (resolvers)

grahamplata:
INSERT DESCRIPTION HERE

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


runtime/resolvers/sql.go
```diff
@@ -62,6 +64,14 @@ func newSQL(ctx context.Context, opts *runtime.ResolverOptions) (runtime.Resolve
...
 
+	// If there is a limit, convert it to an int
+	interactiveRowLimit := cfg.InteractiveSQLRowLimit
+	if props.Limit != "" {
+		interactiveRowLimit, err = strconv.ParseInt(props.Limit, 10, 64)
+		if err != nil {
+			return nil, err
+		}
+	}
 	olap, release, err := opts.Runtime.OLAP(ctx, opts.InstanceID, props.Connector)
```
begelundmuller, line 75: nit: logic above is unrelated to OLAP initialization, so add a newline

```suggestion
	}

	olap, release, err := opts.Runtime.OLAP(ctx, opts.InstanceID, props.Connector)
```

runtime/resolvers/sql.go
```diff
@@ -32,6 +33,7 @@ type sqlResolver struct {
 type sqlProps struct {
 	Connector string `mapstructure:"connector"`
 	SQL       string `mapstructure:"sql"`
+	Limit     string `mapstructure:"limit"`
```
begelundmuller, line 36: If you change `mapstructure.Decode` to `mapstructure.WeakDecode`, then you can make this an `int` and `mapstructure` will handle the conversion (if needed):
```suggestion
	Limit     int `mapstructure:"limit"`
```

That's better because then we won't error if the caller actually does the right thing and passes an `int` in the first place.

begelundmuller (changes requested):

proto/rill/runtime/v1/api.proto
```diff
@@ -587,6 +596,18 @@ message GenerateRendererResponse {
   google.protobuf.Struct renderer_properties = 2;
 }
 
+message QueryResolverRequest {
+  string instance_id = 1;
+  string resolver = 2;
+  google.protobuf.Struct resolver_properties = 3;
+  google.protobuf.Struct resolver_args = 4;
+}
```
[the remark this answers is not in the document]
grahamplata: I Agree, I found the current implementation... awkward

runtime/server/query_resolver.go
```diff
@@ -0,0 +1,74 @@
...
+	})
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+	defer resolver.Close()
+
+	// Query the resolver
+	res, err := resolver.ResolveInteractive(ctx)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
```
begelundmuller, line 48: Usually it would be a cancellation/timeout/query error, so probably best not to mark it internal (which triggers alerts)

```suggestion
		return nil, status.Error(codes.InvalidArgument, err.Error())
```

runtime/server/query_resolver.go
```diff
@@ -0,0 +1,74 @@
...
+	resolver, err := initializer(ctx, &runtime.ResolverOptions{
+		Runtime:    s.runtime,
+		InstanceID: req.InstanceId,
+		Properties: req.ResolverProperties.AsMap(),
+		Args:       req.ResolverArgs.AsMap(),
+		Claims:     claims.SecurityClaims(),
+		ForExport:  false,
+	})
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
```
begelundmuller, line 41:
```suggestion
		return nil, status.Error(codes.InvalidArgument, err.Error())
```

runtime/server/query_resolver.go
```diff
@@ -0,0 +1,74 @@
...
+
+	data := make([]*structpb.Struct, 0)
+	for {
+		// Next returns the next row of data. It returns io.EOF when there are no more rows.
+		row, err := res.Next()
+		if err != nil {
+			if errors.Is(err, io.EOF) {
+				break
+			}
+			return nil, status.Error(codes.Internal, err.Error())
```
begelundmuller, line 60: No chance of user error here, but it may be a `ctx` cancel/timeout error (since the `ctx` was passed to `ResolveInteractive`). The middleware handles that (falling back to `codes.Internal` for unknown error types), so can just return the plain error:

```suggestion
			return nil, err
```

begelundmuller (changes requested):

grahamplata:
> Output samples

`rill query --sql "select ad_position, bid_floor, pub_name, app_or_site from auction_data_raw" --local --format json --limit 3`

```json
[
  {
    "ad_position": "Not Available",
    "app_or_site": "App",
    "bid_floor": 0.1218,
    "pub_name": "GG Software"
  },
  {
    "ad_position": "Not Available",
    "app_or_site": "App",
    "bid_floor": 0.425,
    "pub_name": "Singular One"
  },
  {
    "ad_position": "Not Available",
    "app_or_site": "Site",
    "bid_floor": 0.89,
    "pub_name": "Not Available"
  }
]
```

`rill query --sql "select ad_position, bid_floor, pub_name, app_or_site from auction_data_raw" --local --format human --limit 3`

> Note this would be the default

```bash
➜  rill query --sql "select ad_position, bid_floor, pub_name, app_or_site from auction_data_raw" --local --limit 3
  AD POSITION   BID FLOOR   PUB NAME   APP OR SITE
 ------------- ----------- ---------- -------------
  Not Available   0.1218      GG Software   App
  Not Available   0.425       Singular One   App
  Not Available   0.89        Not Available   Site
```

runtime/server/query_resolver.go
```diff
@@ -0,0 +1,84 @@
...
+				break
+			}
+			return nil, err
+		}
+		rowStruct, err := pbutil.ToStruct(row, res.Schema())
+		if err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+		data = append(data, rowStruct)
+	}
```
begelundmuller, line 77: Since the limit pushdown earlier is optimistic, you need to manually check it here also (e.g. with a counter; maybe treating `limit==0` as no limit?)

begelundmuller (changes requested):

begelundmuller (approved):

PR #6625, opened 2025-02-08
Canvas: add local filters icon and tooltip

djbarnwal:
Merge after https://github.com/rilldata/rill/pull/6615

For local component filters, add filter icon to component header


lovincyrus (approved):
ack

PR #6626, opened 2025-02-08
Canvas: KPI component improvements

djbarnwal:
- For negative comparison change, display change in red text
- Add comparison label for all KPI cards
- Add support for orienting sparkline on right
- Add interactivity to sparkline on hover
- Add comparison multi select toggle


lovincyrus (changes requested):
- [ ] Line clamp the measure label when orientation is horizontal

CleanShot 2025-02-10 at 13 13 17@2x

https://github.com/user-attachments/assets/83b62ec5-51d0-4af6-b935-c1c71b9d6b48

Nit (non-blocker): Nice to have title attribute or tooltip on the comparison options and sparkling orientation as it wasn't clear what they're supposed to do

https://github.com/user-attachments/assets/f58c92ed-60ca-4219-a7f6-1267e22541d0

lovincyrus (approved):
lgtm

PR #6629, opened 2025-02-10
Canvas: Gracefully handle error states

djbarnwal:
- Create an error state component and use it for all canvas components
- Relax constraints for table component


ericpgreen2 (approved):
Looks good!

PR #6635, opened 2025-02-10
feat: Dump resources command for debuggability

k-anshul:

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!

### DIFF
--- a/cli/cmd/sudo/project/dump_resources.go
+++ b/cli/cmd/sudo/project/dump_resources.go
@@ -194,40 +194,39 @@ func printResources(p *printer.Printer, resources map[string]map[string][]*runti
 	for org, projectRes := range resources {
 		for proj, res := range projectRes {
 			for _, r := range res {
-				// convert each resource to a flattened JSON
-				row := make(map[string]any)
+				row := make(map[string]any, 0)
+				rows = append(rows, row)
 				row["org"] = org
 				row["project"] = proj
 				row["resource_type"] = runtime.PrettifyResourceKind(r.Meta.Name.Kind)
 				row["resource_name"] = r.Meta.Name.Name
-				// marshal meta separately
-				meta, err := protojson.MarshalOptions{Indent: " ", Multiline: true}.Marshal(r.Meta)
-				if err != nil {
-					row["meta"] = fmt.Sprintf("failed to marshal meta: %v", err)
-				} else {
-					row["meta"] = json.RawMessage(meta)
-				}
 
-				// convert resource proto to json
+				// each resource has a meta field and a resource(source/model/metricsview etc) which has spec and state fields
+				// we want to flatten the resource to have the meta fields and spec and state fields at the top level
 				rowJSON, err := protojson.Marshal(r)
 				if err != nil {
-					row["error"] = fmt.Sprintf("failed to marshal resource: %v", err)
+					row["error"] = fmt.Sprintf("Failed to marshal resource: %v", err)
+					continue
 				}
-
-				// unmarshal the json to get the spec and state which is inside a model/source/metrics_view etc.
-				parsed := make(map[string]map[string]any)
-				err = json.Unmarshal(rowJSON, &parsed)
+				err = json.Unmarshal(rowJSON, &row)
 				if err != nil {
-					row["error"] = fmt.Sprintf("failed to unmarshal resource: %v", err)
+					row["error"] = fmt.Sprintf("Failed to unmarshal resource: %v", err)
+					continue
 				}
-				for k := range parsed {
-					if k != "meta" {
-						row["spec"] = parsed[k]["spec"].(map[string]any)
-						row["state"] = parsed[k]["state"].(map[string]any)
-						break
+				for k := range row {
+					if k == "meta" {
+						continue
 					}
+					resource, ok := row[k].(map[string]any)
+					if !ok {
+						delete(row, k)
+						continue
+					}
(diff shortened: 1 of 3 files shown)


cli/cmd/sudo/project/dump_resources.go
```diff
@@ -0,0 +1,161 @@
...
+				grp.Go(func() error {
+					row, err := dumpResourcesForProject(ctx, client, org, project, typ)
+					if err != nil {
+						m.Lock()
+						failedProjects[name] = err
+						m.Unlock()
+						return nil
+					}
+					m.Lock()
+					resources[name] = row
```
begelundmuller, line 80: What about flattening this just a bit? So we output a table like:
```
org,project,resource_type,resource_name,meta,spec,state
```

Would make it easier to load into and analyze in e.g. DuckDB or Google Sheets.

cli/cmd/sudo/project/dump_resources.go
```diff
@@ -0,0 +1,161 @@
...
+	if depl == nil {
+		return nil, nil
+	}
+
+	rt, err := runtimeclient.New(depl.RuntimeHost, proj.Jwt)
+	if err != nil {
+		return nil, fmt.Errorf("failed to connect to runtime: %w", err)
+	}
+
+	res, err := rt.ListResources(ctx, &runtimev1.ListResourcesRequest{InstanceId: depl.RuntimeInstanceId})
```
begelundmuller, line 138: If filtering by a type, you can pass it to the request here instead of checking locally

begelundmuller (changes requested):

cli/cmd/sudo/project/dump_resources.go
k-anshul: Okay so you meant a CSV ?

cli/pkg/printer/resources.go
```diff
@@ -540,3 +543,106 @@ type billingIssue struct {
 	Metadata     string `header:"metadata" json:"metadata"`
 	EventTime    string `header:"event_time,timestamp(ms|utc|human)" json:"event_time"`
 }
+
+func (p *Printer) PrintResource(resources map[string]map[string][]*runtimev1.Resource) {
```
begelundmuller, line 547: In any other case, we most likely do not want to include org and project names when listing resources (e.g. `rill project status`). It's confusing to keep it here as a generic function, someone else may try to reach for it when they need to list resources. It would be better to just keep the printing logic in `dump_resources.go`.

cli/pkg/printer/resources.go
```diff
@@ -540,3 +543,106 @@ type billingIssue struct {
...
+	var (
+		spec, state       json.RawMessage
+		specErr, stateErr error
+	)
+
+	switch r.Meta.Name.Kind {
+	case runtime.ResourceKindSource:
+		spec, specErr = protojson.MarshalOptions{Indent: " "}.Marshal(r.GetSource().Spec)
+		state, stateErr = protojson.MarshalOptions{Indent: " "}.Marshal(r.GetSource().State)
+	case runtime.ResourceKindModel:
```
begelundmuller, line 578: All these cases seem excessive for an internal dump tool. Did you consider just doing a JSON roundtrip to avoid handling each type specifically? E.g. (pseudocode):
```go
rowJSON := protojson.Marshal(r)
row := json.Unmarshal(rowJSON)
for k, v := range row {
  if k != "meta" {
    row["spec"] = row[k]["spec"]
    row["state"] = row[k]["state"]
    delete(row, k)
    break
  }
}

row["org"] = org
row["project"] = project
row["resource_type"] = runtime.PrettifyResourceKind(r.Meta.Name.Kind)
row["resource_name"] = r.Meta.Name.Name

return row
```

cli/pkg/printer/resources.go
```diff
@@ -540,3 +543,106 @@ type billingIssue struct {
...
+
+type resource struct {
+	Org          string          `header:"org" json:"org"`
+	Project      string          `header:"project" json:"project"`
+	ResourceType string          `header:"resource_type" json:"resource_type"`
+	ResourceName string          `header:"resource_name" json:"resource_name"`
+	Meta         json.RawMessage `header:"meta" json:"meta"`
+	Spec         json.RawMessage `header:"spec" json:"spec"`
+	State        json.RawMessage `header:"state" json:"state"`
+}
```
begelundmuller, line 648: CLI formatted tables don't really make sense here since meta/spec/state are so big. IMO can just remove and only have JSON output from the command itself (like in the first PR).

cli/pkg/printer/resources.go
k-anshul: No I did not consider handling via JSON but sounds like a better idea. Modified the code.

cli/cmd/sudo/project/dump_resources.go
```diff
@@ -0,0 +1,240 @@
...
+)
+
+func DumpResources(ch *cmdutil.Helper) *cobra.Command {
+	var pageSize uint32
+	var pageToken string
+	var annotations map[string]string
+	var typ string
+
+	searchCmd := &cobra.Command{
+		Use:   "dump-resources [<pattern>]",
```
begelundmuller, line 30: Someone might think it's a resource name pattern (not project name pattern):

```suggestion
		Use:   "dump-resources [<project-pattern>]",
```

begelundmuller (approved):

PR #6640, opened 2025-02-11
fix : change duckdb connection held check to log after 3 hours

k-anshul:
The default model and source timeout is now 3 hours so a connection can be held for 3 hours.
**Note** : A user can override with a longer timeout as well but that seems unlikely. 

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/runtime/reconcilers/model.go
+++ b/runtime/reconcilers/model.go
@@ -28,6 +28,7 @@ import (
 )
 
 const (
+	// If changing this value also update maxAcquiredConnDuration in runtime/drivers/duckdb/duckdb.go
 	_modelDefaultTimeout = 3 * time.Hour
 
 	_modelSyncPartitionsBatchSize    = 1000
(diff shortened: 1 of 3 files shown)


runtime/reconcilers/model.go
```diff
@@ -28,6 +28,7 @@ import (
 )
 
 const (
+	// If changing this value also update maxAcquiredConnDuration in runtime/drivers/duckdb/duckdb.go
 	_modelDefaultTimeout = 3 * time.Hour
```
begelundmuller, line 32: This is just a default timeout, but a higher one can be provided with `timeout: 6h` in the YAML. So this may be more a stopgap than a solution.

begelundmuller (approved):

runtime/reconcilers/model.go
k-anshul: Yes. I think we either need to modify the logic of hanging conns or remove it completely since we haven't seen this error in a long time now.

Issue #6313, opened 2024-12-19
Ability to toggle off RC features


royendo:
**Is your feature request related to a problem? Please describe.**
For some security purposes, having the ability to toggle off features in RC such as Public URL would be handy.

**Describe the solution you'd like**
Whether its a RC setting or a rill.yaml (like we do for pivot embeds or features) , have the ability to disable specific RC functionality

Something like... 
```
features:
   -  public_url: false
   - scheduled_report: false
```

or...
```
rill_cloud:
   - public_url: false
   - ...
```

nishantmonu51:
would be in favor of using feature flags

ericpgreen2:
I'll consider this closed per https://github.com/rilldata/rill/pull/6641. If there are more specific features that should be guarded by a feature flag, feel free to re-open this issue and list them out.

PR #6641, opened 2025-02-11
feat: add toggle for hiding public url

mindspank:
Toggle to hide publicUrl.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-admin/src/features/dashboards/share/ShareDashboardPopover.svelte
+++ b/web-admin/src/features/dashboards/share/ShareDashboardPopover.svelte
@@ -14,9 +14,11 @@
     TabsTrigger,
   } from "@rilldata/web-common/components/tabs";
   import Check from "@rilldata/web-common/components/icons/Check.svelte";
+  import { featureFlags } from "@rilldata/web-common/features/feature-flags";
 
   export let createMagicAuthTokens: boolean;
 
+  const { hidePublicUrl } = featureFlags;
   let isOpen = false;
   let copied = false;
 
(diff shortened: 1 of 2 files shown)


mindspank:
Closes #6313

royendo:
Does it make more sense to have hidepublicUrl default to false and have a customer require to set this true to remove publicurl or have publicurl and default this to true and have them set it to false when they want to remove it?

mindspank:
> Does it make more sense to have hidepublicUrl default to false and have a customer require to set this true to remove publicurl or have publicurl and default this to true and have them set it to false when they want to remove it?

Right now it defaults to false. So if you set the flag it would remove the option to create public links

```
#rill.yaml

features:
  - hidePublicUrl
```

mindspank:
@ericpgreen2 but maybe we should be descriptive in the toggle name for permanent toggles.

Both from a code perspective but also for the user.

```
features:
  - hideExport
```

Reads cleaner then
```
features:
  - export: false

**Comment by ericpgreen2** [comment]
I can see it either way. But the longer the list of toggled features, then I think this...
```
features:
  canvasDashboards: true
  cloudDataViewer: true
  exports: false
  publicUrls: false
```
... is more clear than...
```
features:
  - canvasDashboards
  - cloudDataViewer
  - hideExports
  - hidePublicUrls
```

It's confusing to see a mix of positives & negatives. And, semantically, I'd argue "exports" is the feature, not "hiding exports".

royendo:
I don't have any particular strong feelings on this one, whatever seems to make the most sense overall in the application and as long as it's consistent 💪

PR #6650, opened 2025-02-11
fix: adds banner to rd preview

mindspank:
Adds the banner to preview of explores in rill developer. Verified it works with the inactive banner which stacks on top.

**Checklist:**
- [x] Covered by tests
- [ ] I'm proud of this work!


web-local/src/routes/(viz)/explore/[name]/+page.svelte
```diff
@@ -53,12 +54,28 @@
     enabled: $selectedMockUserStore?.admin,
   });
 
+  $: hasBanner = !!$exploreQuery.data?.explore?.banner;
+
+  $: if (hasBanner) {
+    eventBus.emit("banner", {
+      type: "default",
+      message: $exploreQuery.data?.explore?.banner ?? "",
```
Comment (ericpgreen2) [inline web-local/src/routes/(viz)/explore/[name]/+page.svelte:62]: Nit: It looks like this doesn't need optional chaining given the `hasBanner` condition

ericpgreen2 (changes requested):
Would you mind adding an e2e test to cover the banner? It could be a short test placed at `tests/explores/banner.spec.ts`

web-local/src/routes/(viz)/explore/[name]/+page.svelte
Comment (mindspank) [reply web-local/src/routes/(viz)/explore/[name]/+page.svelte:62]: It shouldn't but eslint complained 🤷

ericpgreen2 (approved):

PR #6657, opened 2025-02-12
Fix: Org rename not reflected in the breadcrumb in Rill Cloud UI

grahamplata:
Issue: https://github.com/rilldata/FE-private-issues/issues/221
Relates to: https://github.com/rilldata/rill-private-issues/issues/1245

```bash
➜  rill org rename

```

**Checklist:**
- [x] Covered by tests
- [x] Checked if the docs need to be updated - Will run doc gen once tests pass
- [ ] I'm proud of this work!


grahamplata:
> 2\. Do you think we might be better off just skipping prompting altogether and just using flags? Then we could just:

For sure this

begelundmuller:
> > 2. Do you think we might be better off just skipping prompting altogether and just using flags? Then we could just:
> 
> For sure this

Feel free to make that change for this command in this PR now

grahamplata:
Behavior without flags on an edit command. 

```bash
rill org edit
? Select org to edit nullstring
? Do you want to update the display name No
? Do you want to update the description No
? Do you want to update the billing email No
Updated organization
  NAME         CREATED AT
 ------------ ---------------------
  nullstring   2025-02-14 14:36:43
```

```bash
rill org edit nullstring --display-name="NULLSTRING"
Updated organization
  NAME         CREATED AT
 ------------ ---------------------
  nullstring   2025-02-14 14:36:43
```

cli/cmd/org/rename.go
```diff
@@ -40,51 +39,95 @@ func RenameCmd(ch *cmdutil.Helper) *cobra.Command {
...
-			if !force {
-				msg := fmt.Sprintf("Do you want to rename org \"%s\" to \"%s\"?", color.YellowString(name), color.YellowString(newName)) // nolint:gocritic // Because it uses colors
-				ok, err := cmdutil.ConfirmPrompt(msg, "", false)
+			if cmd.Flags().Changed("new-name") {
+				req.NewName = &newName
+				if !cmd.Flags().Changed("display-name") && org.DisplayName != "" {
+					ch.PrintfWarn("Warn: Changing org name without updating display name. Consider using --display-name to keep names consistent.\n")
+				}
+			} else if ch.Interactive && !force {
+				ok, err := confirmFieldChange("Org name", force)
```
begelundmuller, line 77: The case handling here is quite difficult to follow, but it seems like running `rill org rename NAME --display-name DISPLAY_NAME` will lead to a prompt to change the slug name? That would be dangerous because a non-breaking change could easily lead to accidentally breaking URLs. It would also violate the idea that we avoid prompts for optional fields when flags are provided. 

Overall, do you think you can simplify the various cases here? If it makes it simpler, it's okay to prompt less and/or have more errors.

cli/cmd/project/edit.go
```diff
@@ -12,6 +12,7 @@ func EditCmd(ch *cmdutil.Helper) *cobra.Command {
 	var name, description, prodVersion, prodBranch, subpath, path, provisioner string
 	var public bool
 	var prodTTL int64
+	var slots int
```
begelundmuller, line 15: `slots` and the `prod-slots` flag was removed recently – perhaps this is a broken merge?

cli/cmd/project/edit.go
```diff
@@ -49,38 +50,47 @@ func EditCmd(ch *cmdutil.Helper) *cobra.Command {
...
 			}
 
 			if cmd.Flags().Changed("prod-ttl-seconds") {
-				promptFlagValues = false
 				req.ProdTtlSeconds = &prodTTL
 			}
 
-			if promptFlagValues {
+			// Only prompt interactively if no flags were explicitly set
+			if ch.Interactive && !anyFlagsChanged {
```
begelundmuller, line 93: If you prefer, I'm also good with removing the prompt cases here entirely. If you're invoking `rill project edit`, you probably know what you want to edit so relying on flags should be enough (and is idiomatic for a CLI).

cli/cmd/project/jwt.go
```diff
@@ -36,6 +36,9 @@ func JwtCmd(ch *cmdutil.Helper) *cobra.Command {
 				if err != nil {
 					return err
 				}
+			} else if cmd.Flags().Changed("project") {
+				// Use the flag value directly when provided
+				name = cmd.Flag("project").Value.String()
```
begelundmuller, line 41: Something's off here – this shouldn't be needed. When Cobra parses flags, it automatically assigns values into the pointer passed in `jwtCmd.Flags().StringVar(&name, "project", "", "Project Name")`.

cli/cmd/project/partitions.go
```diff
@@ -26,7 +26,8 @@ func PartitionsCmd(ch *cmdutil.Helper) *cobra.Command {
 				model = args[1]
 			}
 
-			if !local && !cmd.Flags().Changed("project") && len(args) <= 1 && ch.Interactive {
+			if !local && !cmd.Flags().Changed("project") && !cmd.Flags().Changed("model") &&
+				len(args) <= 1 && ch.Interactive {
```
begelundmuller, line 30: I don't think it makes sense to bring `!cmd.Flags().Changed("model")` in here. Because we need to know the project to query. So if a project is not specified, we _need_ to either prompt for it or error. 

Again, it's a bit difficult to formalize, but I think "don't prompt when flags are specified" is a principle for optional (and unrelated) fields, not for necessary/required fields (where we need to either prompt or error).

cli/cmd/project/refresh.go
```diff
@@ -26,7 +26,7 @@ func RefreshCmd(ch *cmdutil.Helper) *cobra.Command {
 			if len(args) > 0 {
 				project = args[0]
 			}
-			if !local && !cmd.Flags().Changed("project") && len(args) == 0 && ch.Interactive {
+			if !local && !cmd.Flags().Changed("project") && len(args) == 0 && ch.Interactive && cmd.Flags().NFlag() == 0 {
 				var err error
 				project, err = ch.InferProjectName(cmd.Context(), ch.Org, path)
```
begelundmuller, line 31: Same comment as above, I think. In this case, would the following perhaps be reasonable?
```suggestion
			if !local && project == "" {
				if !ch.Interactive {
					return fmt.Errorf("project not specified and could not be inferred from context")
				}

				var err error
				project, err = ch.InferProjectName(cmd.Context(), ch.Org, path)
				...
```

cli/cmd/project/show.go
```diff
@@ -23,7 +23,7 @@ func ShowCmd(ch *cmdutil.Helper) *cobra.Command {
 				name = args[0]
 			}
 
-			if !cmd.Flags().Changed("project") && len(args) == 0 && ch.Interactive {
+			if !cmd.Flags().Changed("project") && len(args) == 0 && ch.Interactive && !cmd.Flags().Changed("path") {
 				name, err = ch.InferProjectName(cmd.Context(), ch.Org, path)
```
begelundmuller, line 27: Doesn't this end up disabling inference of the project name from `path`? Since that happens in `InferProjectName`? We need to be careful with project names because there are some complicated flows:
- It is usually required (unless it's a command that supports `--local`)
- It can be provided explicitly either with a positional arg OR the `--project` flag
- Or you can specify a `--path` and then we use the path to infer the project
- Or we try to infer it from the path of the current working directory

cli/cmd/project/status.go
```diff
@@ -29,7 +29,7 @@ func StatusCmd(ch *cmdutil.Helper) *cobra.Command {
 				name = args[0]
 			}
 
-			if !cmd.Flags().Changed("project") && len(args) == 0 && ch.Interactive {
+			if !cmd.Flags().Changed("project") && !cmd.Flags().Changed("path") && len(args) == 0 && ch.Interactive {
```
begelundmuller, line 32: See comment above, I don't think it's safe to skip if `path` is set

cli/cmd/publicurl/create.go
```diff
@@ -33,7 +33,7 @@ func CreateCmd(ch *cmdutil.Helper) *cobra.Command {
 			if len(args) == 2 {
 				project = args[0]
 			}
-			if !cmd.Flags().Changed("project") && len(args) == 0 && ch.Interactive {
+			if !cmd.Flags().Changed("project") && len(args) == 0 && ch.Interactive && !cmd.Flags().Changed("path") {
```
begelundmuller, line 36: Same comment

cli/cmd/project/edit.go
grahamplata: Most likely, will adjust

cli/cmd/org/delete.go
```diff
@@ -11,27 +12,24 @@ import (
 
 func DeleteCmd(ch *cmdutil.Helper) *cobra.Command {
 	var force bool
-	var name string
 
 	deleteCmd := &cobra.Command{
 		Use:   "delete [<org-name>]",
```
begelundmuller, line 17: Since you removed the flag, the positional arg is no longer optional:

```suggestion
		Use:   "delete <org-name>",
```

cli/cmd/org/edit.go
```diff
@@ -57,9 +57,14 @@ func EditCmd(ch *cmdutil.Helper) *cobra.Command {
...
 
+			flagsSet := cmd.Flags().Changed("display-name") || cmd.Flags().Changed("description") || cmd.Flags().Changed("billing-email")
+			if !flagsSet {
+				return fmt.Errorf("at least one flag must be set")
+			}
+
 			if cmd.Flags().Changed("display-name") {
 				req.DisplayName = &displayName
-			} else if ch.Interactive {
+			} else if ch.Interactive && !flagsSet {
```
begelundmuller, line 67: Can't you remove all these prompt cases now that you return early when `!flagsSet` on line 61?

cli/cmd/org/rename.go
```diff
@@ -26,65 +27,92 @@ func RenameCmd(ch *cmdutil.Helper) *cobra.Command {
...
 				}
+				return err
 			}
+			org := resp.Organization
 
-			if newName == "" {
-				return fmt.Errorf("please provide valid org new-name, provided: %q", newName)
+			// Require at least one change
+			if !cmd.Flags().Changed("new-name") && !cmd.Flags().Changed("display-name") {
+				return fmt.Errorf("at least one of --new-name or --display-name must be provided")
```
begelundmuller, line 57: This is duplicated – same as on line 30

cli/cmd/org/rename.go
```diff
@@ -26,65 +27,92 @@ func RenameCmd(ch *cmdutil.Helper) *cobra.Command {
...
-					return err
-				}
-				if !ok {
-					return nil
+			// Build update request
+			req := &adminv1.UpdateOrganizationRequest{
+				Name: org.Name,
+			}
+
+			if cmd.Flags().Changed("new-name") {
```
begelundmuller, line 65: When you have `if cmd.Flags().Changed("...")` two places for the same name, the code can probably be simplified. For example, you have `cmd.Flags().Changed("new-name")` both here and on line 56. Instead, you could have (pseudocode):
```go
var flagSet bool
req := &adminv1.UpdateOrganizationRequest{ ... }

if cmd.Flags().Changed("new-name") {
	flagSet = true
	// ...
}

if cmd.Flags().Changed("display-name") {
	flagSet = true
	// ...
}

if !flagSet {
	return  fmt.Errorf("must specify at least one update flag")
}
```

begelundmuller: This comment applies in several files throughout this PR. Please go over the diff and look for all cases where it applies.

cli/cmd/project/hibernate.go
```diff
@@ -21,7 +21,8 @@ func HibernateCmd(ch *cmdutil.Helper) *cobra.Command {
 			if len(args) > 0 {
 				project = args[0]
 			}
-			if !cmd.Flags().Changed("project") && len(args) == 0 && ch.Interactive {
+			// Only prompt interactively if no flags or args are provided
+			if !cmd.Flags().Changed("project") && len(args) == 0 && ch.Interactive && !cmd.Flags().Changed("path") {
```
begelundmuller, line 25: Should be reverted – it should still proceed to inference even if `path` is changed (in fact, the whole point of `path` is to support inference)

cli/cmd/project/jwt.go
```diff
@@ -21,24 +21,20 @@ func JwtCmd(ch *cmdutil.Helper) *cobra.Command {
 				return err
 			}
 
-			if len(args) > 0 {
-				name = args[0]
-			}
```
begelundmuller, line 26: Isn't this important to keep? See the CLI command signature

cli/cmd/project/refresh.go
```diff
@@ -113,6 +116,13 @@ func RefreshCmd(ch *cmdutil.Helper) *cobra.Command {
 				return fmt.Errorf("the --full flag can only be used with --all or --model")
 			}
 
+			// Return an error for ineffective use of --partition
+			if len(modelPartitions) > 0 {
+				if !cmd.Flags().Changed("model") {
+					return fmt.Errorf("--partition requires --model to be explicitly set")
+				}
+			}
```
begelundmuller, line 124: How can this case ever happen? Isn't it already covered by the check on line 86?

cli/cmd/user/list.go
```diff
@@ -8,6 +8,12 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const (
+	UserTokenPrefix   = "usr" // User token prefix
+	InviteTokenPrefix = "inv" // Invite token prefix
```
begelundmuller, line 13: Should not be exported, they are and should only be accessed inside the `ListCmd` command. (These are not global prefixes, they are internal to the command to disambiguate user input.)

cli/cmd/user/list.go
```diff
@@ -78,7 +84,7 @@ func ListCmd(ch *cmdutil.Helper) *cobra.Command {
 	listCmd.Flags().StringVar(&ch.Org, "org", ch.Org, "Organization")
 	listCmd.Flags().StringVar(&projectName, "project", "", "Project")
 	listCmd.Flags().StringVar(&groupName, "group", "", "User group")
-	listCmd.Flags().Uint32Var(&pageSize, "page-size", 50, "Number of users to return per page")
+	listCmd.Flags().Uint32Var(&pageSize, "page-size", DefaultPageSize, "Number of users to return per page")
```
begelundmuller, line 87: Using a constant here is IMO not helpful and just adds indirection – it is natural to specify the default value for a flag on the line of code that defines the flag

cli/cmd/usergroup/create.go
```diff
@@ -23,9 +23,11 @@ func CreateCmd(ch *cmdutil.Helper) *cobra.Command {
...
 			}
 
-			err = cmdutil.StringPromptIfEmpty(&name, "Enter user group name")
-			if err != nil {
-				return err
+			if name == "" {
+				err = cmdutil.StringPromptIfEmpty(&name, "Enter user group name")
+				if err != nil {
+					return err
+				}
```
begelundmuller, line 30: Same comment as in the previous review – I think `StringPromptIfEmpty` already checks this

begelundmuller (changes requested):

cli/cmd/org/delete.go
grahamplata: Is this preferred across all arguments?

cli/cmd/org/edit.go
```diff
@@ -12,6 +12,7 @@ import (
 
 func EditCmd(ch *cmdutil.Helper) *cobra.Command {
 	var orgName, displayName, description, billingEmail string
+	var flagsSet bool
```
begelundmuller, line 15: This variable is only used inside `RunE` so should not be declared in the outer scope (which makes it look like a flag value). E.g. around line 48 seems like a more suitable place.

This applies in several files.

cli/cmd/org/rename.go
```diff
@@ -26,65 +25,66 @@ func RenameCmd(ch *cmdutil.Helper) *cobra.Command {
...
-			err = dotrill.SetDefaultOrg(newName)
-			if err != nil {
-				return err
+			// Print results
+			ch.PrintfSuccess("Updated organization\n")
+			if req.NewName != nil {
+				ch.Printf("Updated name: %s to %s\n", name, updatedOrg.Organization.Name)
+			}
+			if req.DisplayName != nil {
+				ch.Printf("Updated display name: %s to %s\n", org.DisplayName, updatedOrg.Organization.DisplayName)
```
begelundmuller, line 68: Previously you had a check where it printed a warning if it already has a display name, but the user only updated the unique name, i.e. something like `req.NewName != nil && req.DisplayName == nil && org.DisplayName != ""`.

Not a cause to prompt or error, but it's a nice warning because often someone would want to change both `Name` and `DisplayName` in that case.

cli/cmd/project/edit.go
```diff
@@ -10,7 +10,7 @@ import (
 
 func EditCmd(ch *cmdutil.Helper) *cobra.Command {
 	var name, description, prodVersion, prodBranch, subpath, path, provisioner string
-	var public bool
+	var flagSet, public bool
```
begelundmuller, line 13: Same comment as earlier, but notice how this makes it look like `flagSet` is populated from a flag (instead of being internal state)?

cli/cmd/project/edit.go
```diff
@@ -49,66 +36,47 @@ func EditCmd(ch *cmdutil.Helper) *cobra.Command {
 				OrganizationName: ch.Org,
 				Name:             name,
 			}
-			promptFlagValues := ch.Interactive
+
+			// Set values from flags if they were changed
+			if cmd.Flags().Changed("prod-ttl-seconds") && prodTTL < 0 {
+				return fmt.Errorf("prod-ttl-seconds must be non-negative")
+			}
```
begelundmuller, line 43: Since it's rare this field is changed, this kind of validation is nicer to do once in the backend than in the client.

Probably in the database model layer, around here (look for the `validate:` struct tags elsewhere in the file for examples of how to add validation): https://github.com/rilldata/rill/blob/a4ced7d46771430003c26ba3f3fd21812c875865/admin/database/database.go#L446

cli/cmd/project/refresh.go
```diff
@@ -113,6 +116,11 @@ func RefreshCmd(ch *cmdutil.Helper) *cobra.Command {
 				return fmt.Errorf("the --full flag can only be used with --all or --model")
 			}
 
+			// Return an error for ineffective use of --partition
+			if !cmd.Flags().Changed("model") {
+				return fmt.Errorf("--partition requires --model to be explicitly set")
+			}
```
begelundmuller, line 122: This breaks the command for all other resource types than a model (e.g. `rill project refresh --source NAME`)! Also I don't think this is even needed, see my earlier comment here: https://github.com/rilldata/rill/pull/6657#discussion_r1977997711

cli/cmd/org/rename.go
```diff
@@ -26,65 +24,71 @@ func RenameCmd(ch *cmdutil.Helper) *cobra.Command {
...
-				name, err = cmdutil.SelectPrompt("Select org to rename", orgNames, "")
-				if err != nil {
-					return err
-				}
+			// Build update request
+			req := &adminv1.UpdateOrganizationRequest{Name: org.Name}
+
+			if req.NewName != nil && req.DisplayName == nil && org.DisplayName != "" {
+				ch.PrintfWarn("\nWarn: Changing org name will invalidate dashboard URLs.\n")
 			}
```
begelundmuller, line 42: The warning here doesn't match the `if` check. And conversely, the warning printed here was previously accompanied by a confirmation prompt. I think you want this here:

```suggestion
			if req.NewName != nil {
				ch.PrintfWarn("Changing the name will invalidate dashboard URLs.\n")
				ok, err := cmdutil.ConfirmPrompt("Do you want to continue?", "", false)
				if err != nil {
					return err
				}
				if !ok {
					ch.PrintfWarn("Aborted\n")
					return nil
				}
			}
```

And then the note about having forgotten to update the display name should go after the update (since it can easily be done as a follow up). See my next comment.

cli/cmd/org/rename.go
```diff
@@ -26,65 +24,71 @@ func RenameCmd(ch *cmdutil.Helper) *cobra.Command {
...
 
-			err = dotrill.SetDefaultOrg(newName)
-			if err != nil {
-				return err
+			// Print results
+			ch.PrintfSuccess("Updated organization\n")
+			if req.NewName != nil {
+				ch.Printf("Updated name: %s to %s\n", name, updatedOrg.Organization.Name)
+			}
```
begelundmuller, line 70: For example something like this for letting the user know that they may also want to update the display name:

```suggestion
			if req.NewName != nil {
				ch.Printf("Updated name %q to %q\n", name, updatedOrg.Organization.Name)
				if org.DisplayName != "" && req.DisplayName == nil {
					ch.Warnf("You updated the org's unique name, but not its display name; use --display-name to update the display name\n")
				}
			}
```

begelundmuller (changes requested):

begelundmuller:
Can you also merge `main` into this branch to see if that resolves the failing CI checks?

cli/cmd/org/rename.go
```diff
@@ -26,65 +24,81 @@ func RenameCmd(ch *cmdutil.Helper) *cobra.Command {
...
 				}
+				return err
+			}
+			org := resp.Organization
 
-				name, err = cmdutil.SelectPrompt("Select org to rename", orgNames, "")
+			// Build update request
+			req := &adminv1.UpdateOrganizationRequest{Name: org.Name}
+
+			if req.NewName != nil {
```
begelundmuller, line 40: `req.NewName` has not been assigned yet (happens on line 55), so this check is ineffectual.

I think if the `if` clause here is moved to after the assignments to `req` (around line 67), then things look good

begelundmuller (approved):

cli/cmd/billing/subscription/subscription.go
```diff
@@ -5,6 +5,15 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const (
+	PlanStarter    = "starter"
+	PlanPro        = "pro"
+	PlanEnterprise = "enterprise"
```
pjain1, line 11: publicly exposed plan names are `free_trial` and `team` only

PR #6661, opened 2025-02-12
Source to model

k-anshul:
closes https://github.com/rilldata/rill-private-issues/issues/854

**Checklist:**
- [x] Covered by tests
- [ ] I'm proud of this work!


runtime/drivers/duckdb/model_manager.go
```diff
@@ -14,6 +14,8 @@ type ModelInputProperties struct {
 	Args     []any  `mapstructure:"args"`
 	PreExec  string `mapstructure:"pre_exec"`
 	PostExec string `mapstructure:"post_exec"`
+	// DB is set if sql is to be run against an external database
+	DB string `mapstructure:"db"`
```
begelundmuller, line 18: nit: subjective, but can we call this `Database` instead? Somehow `db` sounds like a connection handle

runtime/drivers/duckdb/model_executor_objectstore_self.go
```diff
@@ -14,20 +14,27 @@ import (
...
 
 func (p *objectStoreInputProps) Validate() error {
-	if p.Path == "" {
+	if p.Path == "" && p.URI == "" {
 		return fmt.Errorf("missing property `path`")
 	}
+	if p.Path != "" && p.URI != "" {
+		return fmt.Errorf("cannot specify both `path` and `uri`")
+	}
+	p.Path = p.URI
```
begelundmuller, line 37: If `Path` is set and `URI` is not set, won't this override it? Shouldn't it be:

```suggestion
	if p.URI != "" { // Backwards compatibility
		p.Path = p.URI
	}
```

runtime/drivers/duckdb/model_executor_self.go
```diff
@@ -69,21 +111,37 @@ func (e *selfToSelfExecutor) Execute(ctx context.Context, opts *drivers.ModelExe
...
-			BeforeCreate: inputProps.PreExec,
-			AfterCreate:  inputProps.PostExec,
-		}
-		res, err := olap.CreateTableAsSelect(ctx, stagingTableName, inputProps.SQL, createTableOpts)
-		if err != nil {
-			_ = olap.DropTable(ctx, stagingTableName)
-			return nil, fmt.Errorf("failed to create model: %w", err)
+		if inputProps.DB != "" {
+			// special handling for ingesting from an external database
+			// not handling incremental use cases since ingesting from an external database is mostly for small,experimental use cases
```
begelundmuller, line 116: This is fine, but then perhaps add a check that returns an error if `.DB` is set when `opts.IncrementalRun` is true

runtime/drivers/duckdb/model_executor_self.go
```diff
@@ -125,3 +183,136 @@ func (e *selfToSelfExecutor) Execute(ctx context.Context, opts *drivers.ModelExe
...
+		if _, err := conn.ExecContext(ctx, fmt.Sprintf("USE %s;", safeDBName)); err != nil {
+			return err
+		}
+
+		userQuery := strings.TrimSpace(inputProps.SQL)
+		userQuery, _ = strings.CutSuffix(userQuery, ";") // trim trailing semi colon
+		query := fmt.Sprintf("CREATE OR REPLACE TABLE %s.%s.%s AS (%s\n);", safeName(localDB), safeName(localSchema), safeTempTable, userQuery)
+		_, execErr := conn.ExecContext(ctx, query)
+		// revert to localdb and schema before returning
+		_, err := conn.ExecContext(ctx, fmt.Sprintf("USE %s.%s;", safeName(localDB), safeName(localSchema)))
```
begelundmuller, line 215: Is this critical or nice-to-have? Just wondering if it's okay if `ctx` was cancelled when this is called

begelundmuller (changes requested):
This is so nice!

begelundmuller (commented):
Also, can it now delete the source transporters or are they still used?

runtime/drivers/duckdb/model_executor_self.go
k-anshul: This is on the write handle which is created new for each call.
On the happy path it must revert to localdb and schema so that further query can be passed.
In case of errors like `ctx` cancelled this is not required since this handle will anyways not be used anymore.

runtime/drivers/duckdb/model_manager.go
k-anshul: It is externally called `db` so I kept the `mapstructure` field to `db` but changed the internal field to `database`.

k-anshul:
> Also, can it now delete the source transporters or are they still used?

Yes I planned to handle that in a follow up PR :)

runtime/compilers/rillv1/parse_source.go
```diff
@@ -85,25 +85,33 @@ func (p *Parser) parseSource(ctx context.Context, node *Node) error {
...
-	r, err := p.insertResource(ResourceKindSource, node.Name, node.Paths, node.Refs...)
+	outputProps, err := structpb.NewStruct(map[string]any{"materialize": true})
+	if err != nil {
+		return fmt.Errorf("internal error: can not create output properties for source: %w", err)
+	}
+
+	// Track as a model
+	// We allowed a special resource type (source) to ingest data from external connector
+	// After the unification of sources and models everything is a model
+	r, err := p.insertResource(ResourceKindModel, node.Name, node.Paths, node.Refs...)
```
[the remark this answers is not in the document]
k-anshul: Done

begelundmuller (approved):

PR #6665, opened 2025-02-12
olap cache ctrl, disable caching timeboundary results

pjain1:
External customers do want to cache results from timeboundary queries, its ok to not cache generally as it just hits segment metadata and saves cache space. If already cached then results will be used.


runtime/drivers/drivers.go
```diff
@@ -157,3 +157,23 @@ func NewPermissionDeniedError(msg string) error {
 func (e *PermissionDeniedError) Error() string {
 	return e.msg
 }
+
+type OlapCacheCtrl struct {
+	UseCache      bool
+	PopulateCache bool
+}
```
begelundmuller, line 164: 1. As a Druid-specific setting, it should belong entirely in the `druidsqldriver` package, not pollute the `runtime/drivers` package. The `druidsqldriver` package is also supposed to be a generic `database/sql` driver so having a dependency on `runtime/drivers` is odd.
2. Did you consider passing it as a query argument instead and implementing the `NamedValueChecker` interface? See the docs [here](https://pkg.go.dev/database/sql/driver#NamedValueChecker). This is the recommended way to pass query-level settings through the `database/sql` interface.
3. This code definitely needs a docstring – without context, it's very unclear what it does and what caches it relates to.

begelundmuller (changes requested):

runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go
```diff
@@ -419,6 +421,27 @@ func (chc *coordinatorHTTPCheck) IsHardFailure(ctx context.Context) (bool, error
...
+	UseCache      bool
+	PopulateCache bool
+}
+type ctxKey int
+
+var queryCfgKey ctxKey
+
+func WithQueryCfg(ctx context.Context, cfg *QueryCfg) context.Context {
+	return context.WithValue(ctx, queryCfgKey, cfg)
+}
```
begelundmuller, line 438: 1. Should be `QueryConfig` instead of `QueryCfg` – abbreviations are generally used for variable names, but not for exported type names.
2. The context key type should be `struct{}` not `int`. Search the codebase for `ctxKey` for examples.
3. Newline missing between lines 431 and 432
4. These definitions don't really fit here in the file – it's between `DruidQueryContext` and `DruidParameter` which are both Druid JSON types used in `DruidRequest`

runtime/drivers/olap.go
```diff
@@ -72,9 +72,19 @@ type Statement struct {
 	DryRun           bool
 	Priority         int
 	LongRunning      bool
+	OlapQueryCfg     *OlapQueryCfg
```
begelundmuller, line 75: A separate "OLAP query config" type doesn't really make sense here because all the fields in it are already query config keys (args, priority, timeouts, etc.).

So it should be something like this (but with the docstrings added):

```suggestion
	UseCache bool
	PopulateCache bool
```

begelundmuller (changes requested):

runtime/drivers/druid/druidsqldriver/druid_api_sql_driver.go
pjain1: 1. ok
2. ok, I looked the go [doc](https://github.com/golang/go/blob/master/src/context/context.go#L145) but anyways changed.
3. ok
4. moved to the end before unexported definitions.

runtime/drivers/olap.go
pjain1: I wanted to have it as struct here because I wanted to both of them to be set but now I think about it, its ok to omit. Changed

begelundmuller (approved):

Issue #6490, opened 2025-01-23
Support `.env` files in nested directories


begelundmuller:
If a `.env` file is observed in a nested directory, we should parse it and merge it with variables loaded from other `.env` files. It is okay to assume the same key is not present in multiple `.env` files.

PR #6671, opened 2025-02-12
Feat: Support .env files in nested project directories

grahamplata:
This PR updates the Project parser to handle multiple `.env` files. Nested `.env` files will be parsed and merged with variables loaded from other `.env` files.

Issue: https://github.com/rilldata/rill/issues/6490

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/runtime/compilers/rillv1/parse_dotenv.go
+++ b/runtime/compilers/rillv1/parse_dotenv.go
@@ -2,18 +2,14 @@ package rillv1
 
 import (
 	"context"
-	"os"
 
 	"github.com/joho/godotenv"
 )
 
-// parseDotEnv parses the env file present at repo root
+// parseDotEnv parses the .env file at the given path and merges it with the existing env vars
 func (p *Parser) parseDotEnv(ctx context.Context, path string) error {
 	data, err := p.Repo.Get(ctx, path)
 	if err != nil {
-		if os.IsNotExist(err) {
-			return nil
-		}
 		return err
 	}
 	envMap, err := godotenv.Unmarshal(data)
@@ -21,6 +17,35 @@ func (p *Parser) parseDotEnv(ctx context.Context, path string) error {
 		return err
 	}
 
-	p.DotEnv = envMap
+	if p.DotEnv == nil {
+		p.DotEnv = envMap
+		return nil
+	}
+
+	p.DotEnv = mergeDotEnvMaps(p.DotEnv, envMap)
+
 	return nil
+}
+
+// parseDotEnvs parses and merges .env files at the given paths
+func (p *Parser) parseDotEnvs(ctx context.Context, paths []string) error {
+	for _, path := range paths {
+		err := p.parseDotEnv(ctx, path)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// mergeDotEnvMaps merges two env maps
+func mergeDotEnvMaps(a, b map[string]string) map[string]string {
+	merged := make(map[string]string, len(a)+len(b))
+	for k, v := range a {
+		merged[k] = v
+	}
+	for k, v := range b {
+		merged[k] = v
+	}
+	return merged
 }
--- a/runtime/compilers/rillv1/parse_dotenv_test.go
+++ b/runtime/compilers/rillv1/parse_dotenv_test.go
@@ -1,0 +1,114 @@
+package rillv1_test
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	"github.com/rilldata/rill/runtime/compilers/rillv1"
+	"github.com/rilldata/rill/runtime/drivers"
+	"github.com/rilldata/rill/runtime/pkg/activity"
+	"github.com/rilldata/rill/runtime/storage"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
+)
+
+// TestEnvParsing tests repo without a .env file
+func TestEnvParsing(t *testing.T) {
+	t.Run("without env file", func(t *testing.T) {
+		ctx := context.Background()
+		repo := makeRepo(t, map[string]string{
+			`rill.yaml`: ``,
+		})
+
+		parser, err := rillv1.Parse(ctx, repo, "", "", "duckdb")
+		require.NoError(t, err)
+		require.Equal(t, len(parser.DotEnv), 0)
+	})
+
+	t.Run("with single env file", func(t *testing.T) {
+		ctx := context.Background()
+		repo := makeRepo(t, map[string]string{
+			"rill.yaml": ``,
+			".env": `
+TEST=test
+FOO=bar
+`,
+		})
+
+		parser, err := rillv1.Parse(ctx, repo, "", "", "duckdb")
+		require.NoError(t, err)
+
+		require.Equal(t, "test", parser.DotEnv["TEST"])
+		require.Equal(t, "bar", parser.DotEnv["FOO"])
+	})
+
+	t.Run("with multiple env files", func(t *testing.T) {
+		ctx := context.Background()
+		repo := makeRepo(t, map[string]string{
+			`rill.yaml`: ``,
+			".env": `
+ROOT_TEST_VAR=root
+`,
+			"models/.env": `
+MODELS_FOLDER_TEST_VAR=models
+`,
+		})
+
+		parser, err := rillv1.Parse(ctx, repo, "", "", "duckdb")
+		require.NoError(t, err)
+
+		require.Empty(t, parser.Errors)
+		require.Equal(t, "root", parser.DotEnv["ROOT_TEST_VAR"], "root-only variable should be preserved")
+		require.Equal(t, "models", parser.DotEnv["MODELS_FOLDER_TEST_VAR"], "models-only variable should be preserved")
+	})
+
+	t.Run("env value merge behavior", func(t *testing.T) {
+		ctx := context.Background()
+		repo := makeRepo(t, map[string]string{
+			`rill.yaml`: ``,
+			".env": `
+SHARED_VAR=root_value
+ROOT_ONLY=root_value
+`,
+			"models/.env": `
+SHARED_VAR=models_value
+MODELS_ONLY=models_value
+`,
+			"models/nested/.env": `
+SHARED_VAR=nested_value
+NESTED_ONLY=nested_value
+`,
+		})
+
+		parser, err := rillv1.Parse(ctx, repo, "", "", "duckdb")
+		require.NoError(t, err)
+		require.Empty(t, parser.Errors)
+
+		// Check that variables from all levels exist
+		require.Equal(t, "root_value", parser.DotEnv["ROOT_ONLY"], "root-only variable should be preserved")
+		require.Equal(t, "models_value", parser.DotEnv["MODELS_ONLY"], "models-only variable should be preserved")
+		require.Equal(t, "nested_value", parser.DotEnv["NESTED_ONLY"], "nested-only variable should be preserved")
+
+		// Check that shared variable takes the value from the deepest .env file
+		require.Equal(t, "nested_value", parser.DotEnv["SHARED_VAR"], "shared variable should take value from deepest .env file")
+	})
+}
+
+// Helper functions
+
+// makeRepo is a helper function that creates a new repo with the given files
+func makeRepo(t testing.TB, files map[string]string) drivers.RepoStore {
+	root := t.TempDir()
+	handle, err := drivers.Open("file", "default", map[string]any{"dsn": root}, storage.MustNew(root, nil), activity.NewNoopClient(), zap.NewNop())
+	require.NoError(t, err)
+
+	repo, ok := handle.AsRepoStore("")
+	require.True(t, ok)
+
+	for path, data := range files {
+		err := repo.Put(context.Background(), path, strings.NewReader(data))
+		require.NoError(t, err)
+	}
+	return repo
+}
--- a/runtime/compilers/rillv1/parser.go
+++ b/runtime/compilers/rillv1/parser.go
@@ -563,6 +563,24 @@ func (p *Parser) parsePaths(ctx context.Context, paths []string) error {
 		if pathIsRillYAML(b) {
 			return 1
 		}
+
+		// Sort .env files by path depth with always root first. This ensures that .env files are parsed in order from root to deepest nested
+		aIsDotEnv := pathIsDotEnv(a)
+		bIsDotEnv := pathIsDotEnv(b)
+		if aIsDotEnv && !bIsDotEnv {
+			return -1
+		}
+		if !aIsDotEnv && bIsDotEnv {
+			return 1
+		}
+		if aIsDotEnv && bIsDotEnv {
+			aDepth := strings.Count(a, "/")
+			bDepth := strings.Count(b, "/")
+			if aDepth != bDepth {
+				return aDepth - bDepth
+			}
+		}
+
 		return strings.Compare(a, b)
 	})
 
@@ -1045,9 +1063,9 @@ func pathIsRillYAML(path string) bool {
 	return path == "/rill.yaml" || path == "/rill.yml"
 }
 
-// pathIsDotEnv returns true if the path is .env
+// pathIsDotEnv returns true if the path is a .env file (in any directory)
 func pathIsDotEnv(path string) bool {
-	return path == "/.env"
+	return strings.HasSuffix(path, "/.env")
 }
 
 // pathIsIgnored returns true if the path should be ignored by the parser.


runtime/compilers/rillv1/parse_dotenv.go
```diff
@@ -2,25 +2,39 @@ package rillv1
...
 		return err
 	}
 
-	p.DotEnv = envMap
+	if p.DotEnv == nil {
+		p.DotEnv = envMap
+		return nil
+	}
+
+	p.DotEnv = mergeDotEnvMaps(p.DotEnv, envMap)
```
begelundmuller, line 25: This gets into some of the complex parts of this parser, but on local where the file watcher is on, the parser only re-parses changed files (see `p.Reparse`). In that case, I think deleting a nested `.env` file would not clear `p.DotEnv`, leading to it having inconsistent values.

Two quick thoughts on how to address that:
1. Add the path to `p.DotEnv`, i.e. so `p.DotEnv[path][key] = value`. And add appropriate handling for clearing if the path is deleted here: https://github.com/rilldata/rill/blob/f55c081b0b86b4e6e84eab4cac305b51813b44f4/runtime/compilers/rillv1/parser.go#L410-L426
2. Just do a full parse of all files if a `.env` file is edited. That logic would have to go somewhere around here: https://github.com/rilldata/rill/blob/f55c081b0b86b4e6e84eab4cac305b51813b44f4/runtime/compilers/rillv1/parser.go#L260-L266

begelundmuller (changes requested):

cli/cmd/env/pull.go
```diff
@@ -76,8 +76,10 @@ func PullCmd(ch *cmdutil.Helper) *cobra.Command {
 				resVars[v.Name] = v.Value
 			}
 
+			dotEnv := parser.GetDotEnv()
+
 			// If the variables match any existing .env file, do nothing
-			if maps.Equal(resVars, parser.DotEnv) {
+			if reflect.DeepEqual(resVars, dotEnv) {
```
begelundmuller, line 82: What is wrong with `maps.Equal`? The members are `string`s, so don't require deep equality checks

cli/cmd/env/push.go
```diff
@@ -71,7 +71,9 @@ func PushCmd(ch *cmdutil.Helper) *cobra.Command {
 			}
 			added := 0
 			changed := 0
-			for k, v := range parser.DotEnv {
+
+			dotEnv := parser.GetDotEnv()
+			for k, v := range dotEnv {
```
begelundmuller, line 76:
```suggestion
			for k, v := range parser.GetDotEnv() {
```

cli/pkg/local/project.go
```diff
@@ -20,5 +20,8 @@ func ParseDotenv(ctx context.Context, projectPath string) (map[string]string, er
 	if parser.RillYAML == nil {
 		return nil, fmt.Errorf("not a valid Rill project (missing a rill.yaml file)")
 	}
-	return parser.DotEnv, nil
+
+	env := parser.GetDotEnv()
+
+	return env, nil
```
begelundmuller, line 26:
```suggestion
	return parser.GetDotEnv(), nil
```

runtime/compilers/rillv1/parser.go
```diff
@@ -168,6 +167,10 @@ type Diff struct {
 	Deleted        []ResourceName
 }
 
+// DotEnvMap is a map of environment variables for a specific path.
+// The key is the path, and the value is a map of environment variable names to values.
+type DotEnvMap map[string]map[string]string
```
begelundmuller, line 172: It's subjective, but I don't think defining a type alias increases readability here (means needing to jump more to find out how it can be used).

Type aliases provide most value when the caller just uses member functions of the type alias, and therefore doesn't need to know the underlying type. But in this case, `DotEnv` is directly used as a map multiple places, so it would be helpful for its type signature to clearly show it's a map.

runtime/compilers/rillv1/parser.go
```diff
@@ -414,7 +428,7 @@ func (p *Parser) reparseExceptRillYAML(ctx context.Context, paths []string) (*Di
 			oldDotEnv := p.DotEnv
 			err := p.parseDotEnv(ctx, checkPaths[i])
 			if err == nil {
-				modifiedDotEnv = !maps.Equal(p.DotEnv, oldDotEnv)
+				modifiedDotEnv = !reflect.DeepEqual(p.DotEnv, oldDotEnv)
```
begelundmuller, line 431: I don't think this works:
1. Previously `parseDotEnv` overwrote `p.DotEnv`, so `oldDotEnv` would be a different pointer. But now `parseDotEnv` adds to `p.DotEnv`, so with `oldDotEnv` being a pointer, the equality check will always return `true`.
2. Now that there can be multiple `.env` files, I don't think it's safe to overwrite `modifiedDotEnv` each time a `.env` file is parsed.

runtime/compilers/rillv1/parser.go
```diff
@@ -563,6 +577,24 @@ func (p *Parser) parsePaths(ctx context.Context, paths []string) error {
...
+		if !aIsDotEnv && bIsDotEnv {
+			return 1
+		}
+		if aIsDotEnv && bIsDotEnv {
+			aDepth := strings.Count(a, "/")
+			bDepth := strings.Count(b, "/")
+			if aDepth != bDepth {
+				return aDepth - bDepth
+			}
+		}
```
begelundmuller, line 596: I don't think this matters when you store the values as separate sub-maps of `p.DotEnv`? Instead the precedence would need to be applied in the implementation of `GetDotEnv`.

runtime/compilers/rillv1/parse_dotenv.go
```diff
@@ -2,25 +2,27 @@ package rillv1
...
 )
 
-// parseDotEnv parses the env file present at repo root
+// parseDotEnv parses the .env file at the given path and merges it with the existing env vars
 func (p *Parser) parseDotEnv(ctx context.Context, path string) error {
 	data, err := p.Repo.Get(ctx, path)
 	if err != nil {
-		if os.IsNotExist(err) {
-			return nil
-		}
```
begelundmuller, line 16: Why remove this check?

runtime/compilers/rillv1/parse_dotenv.go
```diff
@@ -2,25 +2,27 @@ package rillv1
...
 		return err
 	}
 
-	p.DotEnv = envMap
+	if p.DotEnv == nil {
+		p.DotEnv = make(DotEnvMap)
+	}
+
+	// Store the env vars for this specific path
+	p.DotEnv[path] = envMap
```
begelundmuller, line 25: If a `.env` file is deleted, how is it subsequently removed from here?

Is that why you removed the check for file existence above? I see how that would work, but it feels fragile. It would be better to clear the entry explicitly in `reparseExceptRillYAML` (which is responsible for clearing state that will be reloaded) and let `parsePaths` re-populate it.

begelundmuller (changes requested):

runtime/compilers/rillv1/parser.go
grahamplata: Fair enough, I will make that change

runtime/reconcilers/project_parser.go
```diff
@@ -300,7 +300,7 @@ func (r *ProjectParserReconciler) reconcileParser(ctx context.Context, inst *dri
 	}
 
 	// not setting restartController=true when diff is actually nil prevents infinite restarts
-	updateConfig := diff == nil || diff.ModifiedDotEnv || diff.Reloaded
+	updateConfig := diff == nil || diff.Reloaded
```
begelundmuller, line 303: Isn't it important to have the `diff.ModifiedDotEnv` here? Otherwise how will it update the variables when `.env` is interactively edited?

runtime/compilers/rillv1/parse_dotenv.go
```diff
@@ -2,25 +2,50 @@ package rillv1
...
+	"maps"
 	"os"
 
 	"github.com/joho/godotenv"
 )
 
-// parseDotEnv parses the env file present at repo root
-func (p *Parser) parseDotEnv(ctx context.Context, path string) error {
+// parseAndMergeEnvironmentVariables parses the .env file at the given path and merges it with the existing env vars
+func (p *Parser) parseAndMergeEnvironmentVariables(ctx context.Context, path string) error {
```
begelundmuller, line 13: Let's keep the `parseDotEnv` name, all the file-type-specific parse functions in the package have a name like `parseXYZ` (and they all internally modify the parser's state).

runtime/compilers/rillv1/parse_dotenv.go
```diff
@@ -2,25 +2,50 @@ package rillv1
...
+	// Unlikely, but if the path is empty, return an error
+	if path == "" {
+		return fmt.Errorf("empty path provided for .env file")
+	}
```
begelundmuller, line 17: Isn't this already checked upstream? Did you encounter an empty name propagating to here?

runtime/compilers/rillv1/parse_dotenv.go
```diff
@@ -2,25 +2,50 @@ package rillv1
...
 			return nil
 		}
 		return err
 	}
+
+	// Unmarshal the .env file
 	envMap, err := godotenv.Unmarshal(data)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to unmarshal .env file at %s: %w", path, err)
```
begelundmuller, line 32: No need to include the `path` in the name here. We use a special error type to attach paths to errors, and it's already handled by the caller: https://github.com/rilldata/rill/blob/a4ced7d46771430003c26ba3f3fd21812c875865/runtime/compilers/rillv1/parser.go#L583-L586

runtime/compilers/rillv1/parse_dotenv.go
```diff
@@ -2,25 +2,50 @@ package rillv1
...
+		return nil
+	}
+
+	// Check if the map is the same as the current state
+	if maps.Equal(currentEnv, envMap) {
+		return nil
+	}
+
+	// Overwrite the current state with the new state
+	p.DotEnv[path] = envMap
```
begelundmuller, line 48: Isn't this the same as just doing `p.DotEnv[path] = envMap` outright?

runtime/compilers/rillv1/parser.go
```diff
@@ -302,11 +300,22 @@ func (p *Parser) TrackedPathsInDir(dir string) []string {
...
+// GetDotEnv returns all the project's environment variables
+func (p *Parser) GetDotEnv() map[string]string {
+	env := make(map[string]string)
+	for _, envMap := range p.DotEnv {
+		for k, v := range envMap {
+			env[k] = v
+		}
+	}
+	return env
+}
```
begelundmuller, line 312: Would be nice to make this deterministic in case the same env var is defined multiple places (bad thing to do, but non-determinism would make it trickier to debug) – e.g. sort the paths and add them in sorted order

runtime/compilers/rillv1/parse_dotenv_test.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/rilldata/rill/runtime/compilers/rillv1"
+	"github.com/rilldata/rill/runtime/drivers"
+	"github.com/rilldata/rill/runtime/pkg/activity"
+	"github.com/rilldata/rill/runtime/storage"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
+)
+
+// TestEnvParsing tests repo without a .env file
+func TestEnvParsing(t *testing.T) {
```
begelundmuller, line 17: This doesn't test `Reparse`, but that's where all the edge cases are

runtime/compilers/rillv1/parse_dotenv_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+	repo, ok := handle.AsRepoStore("")
+	require.True(t, ok)
+
+	for path, data := range files {
+		err := repo.Put(context.Background(), path, strings.NewReader(data))
+		require.NoError(t, err)
+	}
+	return repo
+}
```
begelundmuller, line 119: This util is already defined in `parser_test.go`, why duplicate it? You just need to change the package name in this test file from `package rillv1_test` to `package_rillv1` to get it into scope.

runtime/compilers/rillv1/parser.go
```diff
@@ -400,29 +408,20 @@ func (p *Parser) reparseExceptRillYAML(ctx context.Context, paths []string) (*Di
...
-				modifiedDotEnv = !maps.Equal(p.DotEnv, oldDotEnv)
-			} else {
-				// any error means .env is under change
-				modifiedDotEnv = true
-			}
-			if !modifiedDotEnv {
-				continue
+			err := p.parseAndMergeEnvironmentVariables(ctx, path)
+			if err != nil {
+				p.addParseError(path, err, false)
```
begelundmuller, line 423: 1. I don't see how the issue from this comment is handled? https://github.com/rilldata/rill/pull/6671#discussion_r1967632614. I.e. if you delete a `.env` file, I don't think it gets cleared during incremental parsing.
2. It shouldn't add a parse error here because it already does that in `parsePaths`. The duplicate call to `parseDotEnv` in the old implementation is pretty hacky so I understand the confusion, but it's important that the implementation in this function (`reparseExceptRillYAML`) only clears old state, but doesn't add new state. I know the parser (especially the incremental "reparsing") is complex, try to go through it and see if it makes sense how it works.

begelundmuller (changes requested):

runtime/compilers/rillv1/parse_dotenv.go
grahamplata: I see what you mean, is it worth just reassigning always?

begelundmuller: Updating in the map here is not expensive at this point (the assignment is very cheap, the garbage collection of one of the maps unavoidable at this point). Doing a `maps.Equal` will just be extra cycles.

The place where we don't want unnecessary updates is upstream, i.e. we don't want to the `Diff` to indicate the `.env` was changed if it wasn't, since that causes some very expensive upstream operations (reconciliation of all resources since we don't know which resources use a given env key or not).

runtime/compilers/rillv1/parse_dotenv.go
```diff
@@ -18,9 +18,10 @@ func (p *Parser) parseDotEnv(ctx context.Context, path string) error {
 	}
 	envMap, err := godotenv.Unmarshal(data)
 	if err != nil {
-		return err
+		p.addParseError(path, err, false)
 	}
```
begelundmuller, line 22: 1. `addParseError` should only be called in `parser.go`, not in the functions for individual file types
2. If you return the `err`, the upstream code will automatically call `addParserError`
3. It should return here! It should not use this function to do cleanup (e.g. setting the map entry to `nil`).

About point 3., if you look at how `Reparse` works, it's essentially a two-step process: first it removes/cleans up all state related to the files that were changed, such that the parser's state looks like those files never existed in the first place. Then secondly, it parses the changed files, calling the `parseXYZ` functions, which parse the files and add the results to the parser's state.

This means the `parseXYZ` functions are only responsible for adding to the parser's state, not cleaning it up during a reparse (since that happens in the first part of the process, which is implemented separately in the `reparseExceptRillYAML` function.

runtime/compilers/rillv1/parser.go
```diff
@@ -411,18 +431,17 @@ func (p *Parser) reparseExceptRillYAML(ctx context.Context, paths []string) (*Di
 		// Watcher sends multiple events for a single edit. We want to only restart the controller when .env actually changes.
 		// So we check the new .env contents against the contents stored in parser state.
 		if isDotEnv {
-			oldDotEnv := p.DotEnv
-			err := p.parseDotEnv(ctx, checkPaths[i])
+			oldDotEnv := p.DotEnv[path]
+			err := p.parseDotEnv(ctx, path)
 			if err == nil {
-				modifiedDotEnv = !maps.Equal(p.DotEnv, oldDotEnv)
+				modifiedDotEnv = !maps.Equal(p.DotEnv[path], oldDotEnv)
```
begelundmuller, line 437: This looks right now, I think. Given my comment above, I understand it's confusing to see a `parseDotEnv` call here (since that should only happen in the second phase, not the "cleanup" phase here). The reason it calls it here is a hacky way to easily work out if the `.env` file has been modified.

runtime/compilers/rillv1/parse_dotenv.go
begelundmuller: If you just revert this line to `return err` and make no other changes, do all the tests still pass? I think that's all that should be needed.

PR #6676, opened 2025-02-13
Discoverable copy nit for refresh individual resources

lovincyrus:
Updated copy change in refresh all dialog

CleanShot 2025-02-13 at 12 45 12@2x

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-admin/src/features/projects/status/RefreshAllSourcesAndModelsConfirmDialog.svelte
+++ b/web-admin/src/features/projects/status/RefreshAllSourcesAndModelsConfirmDialog.svelte
@@ -28,7 +28,8 @@
       <AlertDialogTitle>Refresh all sources and models?</AlertDialogTitle>
       <AlertDialogDescription>
         <div class="mt-1">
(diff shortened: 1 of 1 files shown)


ericokuma:
I believe the only resources that can be single refreshed are sources and models. In that case the copy should say:

"This will refresh all project sources and models.

**Note:** To refresh a single source or model, use the refresh option behind the '...' button."

ericokuma:
if it's easier to discover, we can replace the "..." dropdown button to a singular "refresh" secondary button

cc @jkhwu

lovincyrus:
Latest:

CleanShot 2025-02-13 at 16 13 38@2x

ericokuma (changes requested):
Sorry update to: "<span class="font-medium">Note:</span>To refresh a single resource, scroll to the source or model, click the '...' button, and select the refresh option."

Just to be safe and explicitly clear

ericokuma (approved):

PR #6680, opened 2025-02-14
feat: public api for embeds

mindspank:
Adds a RPC API for Rill embeds.

methods:
 - `setState`
 - `getState`
 
 notifications:
  - `stateChange`

**Checklist:**
- [x] Covered by tests
- [ ] I'm proud of this work!


web-admin/src/features/embeds/EmbedPublicAPI.svelte
```diff
@@ -0,0 +1,85 @@
...
+  });
+
+  registerMethod("setState", (state: string) => {
+    const currentUrl = new URL(get(page).url);
+    currentUrl.search = state;
+    goto(currentUrl, { replaceState: true });
+    return true;
+  });
+
+  $: stateString = decodeURIComponent(
```
mindspank, line 70: Something weird happening in Svelte that I don't understand. For notifications I had to move the reactivity to its own variable otherwise it would trigger twice.

Using the same reactive variable inside of the getState method would not work correctly.
If there is any Svelte gurus - teach me senpai

web-admin/src/routes/-/embed/+layout.svelte
```diff
@@ -1,12 +1,18 @@
...
   import RuntimeProvider from "@rilldata/web-common/runtime-client/RuntimeProvider.svelte";
+  import { initRPC } from "@rilldata/web-common/lib/rpc";
 
   const instanceId = $page.url.searchParams.get("instance_id");
   const runtimeHost = $page.url.searchParams
     .get("runtime_host")
     .replace("localhost:9091", "localhost:8081");
   const accessToken = $page.url.searchParams.get("access_token");
+
+  onMount(() => {
```
mindspank, line 13: kept the initializing here if we want to support state for canvas in the future

ericpgreen2 (commented):
Regarding the JSON-RPC 2.0 interface, I think it's too heavy to ask that host application developers implement their own JSON-RPC over `postMessage` client (since there's no widely adopted NPM library for this). I looked around a little bit and didn't find any other BI tools that take this approach (correct me if I'm wrong). 

Instead, I'd propose we either:
1. Use a simple custom JSON interface, much like what [Looker](https://cloud.google.com/looker/docs/embedded-javascript-events#posting_the_request_to_the_iframes_contentwindow_property), [Sigma](https://help.sigmacomputing.com/docs/inbound-and-outbound-events-in-embeds#example-inbound-event-update-controls-in-sigma), and [Metabase](https://www.metabase.com/docs/latest/embedding/interactive-embedding#supported-postmessage-messages-to-embedded-metabase) offer.
2. Build a Javascript SDK that hides the JSON-RPC exchange under-the-hood.

I'd suggest we do Option 1 in the short-run and explore Option 2 in the long-run.

web-common/src/lib/rpc.ts
```diff
@@ -0,0 +1,98 @@
...
+    id: string | number | null,
+    error: { code: number; message: string; data?: unknown },
+) {
+    if (window.parent !== window) {
+        window.parent.postMessage({ id, error } as JSONRPCResponse, "*");
+    }
+}
+
+export function initRPC() {
+    window.removeEventListener("message", (_event: MessageEvent) => { });
```
AdityaHegde, line 79: Why is this needed? If we want to cleanup then the reference to the listener has to be same.

A better way is to return an "unregister" handler that can be returned in `onmount`. Check `ErrorEventHandler.addJavascriptErrorListeners` for an example.

mindspank:
> Regarding the strict JSON-RPC 2.0 interface, I think it's too heavy to ask that host application developers implement their own JSON-RPC 2.0 over `postMessage` client (since there's no widely adopted NPM library for this). I looked around a little bit and didn't find any other BI tools that take this approach (correct me if I'm wrong).

For ref: removed references to JSON-RPC and made id optional if you do not want a response back.
It diverges slightly from the standard but still maintains the core rpc functionality.

web-common/src/lib/rpc.ts
```diff
@@ -0,0 +1,97 @@
...
+    const result = await methods[method](params);
+    if (id !== null) {
+      sendResponse(id, result);
+    }
+  } catch (error) {
+    sendError(id, {
+      code: JSONRPC_ERRORS.INTERNAL_ERROR.code,
+      message: (error as Error).message,
+    });
+  }
```
ericpgreen2, line 60: I'm a little nervous about this cast. Something like the below might be safer?
```
  const errorMessage = error instanceof Error ? error.message : String(error);
  sendError(id, {
    code: JSONRPC_ERRORS.INTERNAL_ERROR.code,
    message: errorMessage,
  });
```

web-common/src/lib/rpc.ts
```diff
@@ -0,0 +1,97 @@
...
+}
+
+export function initRPC() {
+  window.addEventListener("message", (event: MessageEvent) => {
+    if (event.source && event.data) {
+      void handleRPCMessage(event as MessageEvent<JSONRPCRequest>);
+    }
+  });
+}
```
ericpgreen2, line 84: We don't really need it immediately, but for extra caution, we should return a cleanup function so we can unregister the event listener on component dismount

web-admin/tests/embeds.spec.ts
ericpgreen2, line 1: Awesome to see these tests! Thank you!

web-admin/tests/utils/generateEmbed.ts
ericpgreen2, line 1: Nit: switch file name to kebab case, which we've slowly tried to standardize on for TS files: `generate-embed.ts`

web-admin/tests/utils/generateEmbed.ts
```diff
@@ -0,0 +1,59 @@
...
+import path from "path";
+import { fileURLToPath } from "url";
+
+export async function generateEmbed(
+  resourceId: string,
+  serviceToken: string,
+): Promise<void> {
+  try {
+    const response: AxiosResponse<{ iframeSrc: string }> = await axios.post(
+      "http://localhost:8080/v1/organizations/e2e/projects/openrtb/iframe",
```
ericpgreen2, line 13: Nit: pass the `organization` and `project` as function parameters

web-admin/src/features/embeds/EmbedPublicAPI.svelte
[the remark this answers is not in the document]
mindspank: Sure, feels like we are spreading around logic though as we would potentially extend to canvas etc.
I can also imagine we might want to introduce methods that are not dashboard specific such as displaying a pop-up and so on

web-admin/tests/utils/generateEmbed.ts
mindspank: Moved org and project names into consts.

web-common/src/features/dashboards/workspace/Dashboard.svelte
```diff
@@ -95,6 +96,31 @@
...
+
+  });
+
+  $: if (isReady && initEmbedPublicAPI) {
+    try {
+      initEmbedPublicAPI(instanceId);
+    } catch (error) {
+      console.error("Error running initEmbedPublicAPI:", error);
+    }
+  };
```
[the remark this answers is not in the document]
mindspank: He, ye o4 actually suggested the same but there is something strange going on inside of svelte. With the above approach the `ready` notification is swallowed. I'm guessing we might need two ticks to wait for everything to initialize which kinda what observing the reactive variable does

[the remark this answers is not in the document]
ericpgreen2: Hmm, if that's the case here's another suggestion. Mainly I think we could avoid needing the `isReady` variable.

```
onMount(async () => {
  if (isEmbedded) {
    const { default: initEmbedPublicAPI } = await import("@rilldata/web-admin/features/embeds/init-embed-publicapi.ts");
  }
});

$: if (initEmbedPublicAPI) {
  try {
    await tick(); // Hack: ensure the dashboard is fully loaded before initializing the embed API
    initEmbedPublicAPI(instanceId);
  } catch (error) {
    console.error("Error running initEmbedPublicAPI:", error);
  }
};
```

web-admin/tests/setup/base.ts
```diff
@@ -44,4 +54,34 @@ export const test = base.extend<MyFixtures>({
...
+      );
+      const rillServiceToken = fs.readFileSync(readPath, "utf-8");
+
+      await generateEmbed(
+        RILL_ORG_NAME,
+        RILL_PROJECT_NAME,
+        "bids_explore",
+        rillServiceToken,
+      );
+      const filePath = "file://" + path.resolve(__dirname, "..", "embed.html");
```
ericpgreen2, line 75: Suggestion, so we don't have `embed.html` hardcoded in two places:
```
const embedPath = await generateEmbed(
  RILL_ORG_NAME,
  RILL_PROJECT_NAME,
  "bids_explore",
  rillServiceToken
);
const filePath = "file://" + path.resolve(embedPath);
```

web-admin/tests/setup/constants.ts
```diff
@@ -4,3 +4,7 @@ export const VIEWER_STORAGE_STATE = "playwright/.auth/viewer.json";
 
 export const RILL_DEVTOOL_BACKGROUND_PROCESS_PID_FILE =
   "playwright/rill-devtool-background-process-pid.txt";
+export const RILL_EMBED_SERVICE_TOKEN = "playwright/rill-service-token.txt";
```
ericpgreen2, line 7: Nit: `RILL_EMBED_SERVICE_TOKEN_FILE`

web-admin/tests/setup/git/repos/rill-examples
ericpgreen2, line 1: This is a good point that we should be pinning the test repos to a specific commit, but I don't think this on its own does much because the `setup.ts` script is still overwriting the directory with a clone of the `main` branch of `rill-examples`. This is a good idea, but mind breaking it out into its own PR? Another approach could be that we just checkout a specific hash after the `git clone` command.

ericpgreen2 (approved):
Great feature. Big thanks – especially for the tests!

web-admin/tests/setup/base.ts
mindspank: I'll keep the two versions for now since `resolve` is relative to the callers directory. Could make sense to move into a constant though since the file location would always have to be in the root of tests

PR #6682, opened 2025-02-14
moving prod-slots to superuser

NamanMahor:
moving prod-slots to superuser

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


proto/rill/admin/v1/api.proto
```diff
@@ -1178,6 +1178,7 @@ message UpdateProjectRequest {
   optional string new_name = 9;
   optional int64 prod_ttl_seconds = 10;
   optional string prod_version = 11;
+  optional bool superuser_force_access = 14;
```
begelundmuller, line 1181: Doesn't have to be `optional`, the default value when not `optional` will be `false`, which is desired. (`optional` has a different meaning in protos than you would expect – even non-optional fields are not required).

admin/server/projects.go
```diff
@@ -591,7 +591,14 @@ func (s *Server) UpdateProject(ctx context.Context, req *adminv1.UpdateProjectRe
 
 	claims := auth.GetClaims(ctx)
 	if !claims.ProjectPermissions(ctx, proj.OrganizationID, proj.ID).ManageProject {
```
begelundmuller, line 593: Check forced access as part of the normal access check, not for prod slots specifically.

```suggestion
	claims := auth.GetClaims(ctx)
	forceAccess := claims.Superuser(ctx) && req.SuperuserForceAccess
	if !claims.ProjectPermissions(ctx, proj.OrganizationID, proj.ID).ManageProject && !forceAccess {
```

admin/server/projects.go
```diff
@@ -591,7 +591,14 @@ func (s *Server) UpdateProject(ctx context.Context, req *adminv1.UpdateProjectRe
...
-		return nil, status.Error(codes.PermissionDenied, "does not have permission to delete project")
+		return nil, status.Error(codes.PermissionDenied, "does not have permission to manage project")
+	}
+
+	if req.ProdSlots != nil {
+		forceAccess := req.SuperuserForceAccess != nil && *req.SuperuserForceAccess && claims.Superuser(ctx)
+		if !forceAccess {
+			return nil, status.Error(codes.PermissionDenied, "does not have permission to change prod-slots")
+		}
```
begelundmuller, line 601: See other comment – this doesn't work, the earlier access check would prevent a superuser's call from making it here.

There's also no need to check `superuser_force_access` on a per-field basis. If a superuser sets that flag, it's okay to allow editing any field from an API point-of-view.

cli/cmd/sudo/project/edit.go
```diff
@@ -0,0 +1,64 @@
...
+			return nil
+		},
+	}
+
+	editCmd.Flags().IntVar(&prodSlots, "prod-slots", 0, "Slots to allocate for production deployments")
+	if !ch.IsDev() {
+		if err := editCmd.Flags().MarkHidden("prod-slots"); err != nil {
+			panic(err)
+		}
+	}
```
begelundmuller, line 62: Now that it's a `rill sudo` command, it's not necessary to hide this flag

cli/cmd/sudo/project/edit.go
```diff
@@ -0,0 +1,64 @@
...
+			req := &adminv1.UpdateProjectRequest{
+				OrganizationName:     args[0],
+				Name:                 args[1],
+				SuperuserForceAccess: &[]bool{true}[0],
+			}
+
+			if cmd.Flags().Changed("prod-slots") {
+				if prodSlots <= 0 {
+					ch.PrintfError("--prod-slots must be greater than zero\n")
+					return nil
```
begelundmuller, line 28: nit: when exiting with an error, just return the error (it will be logged and a non-zero status code returned)

```suggestion
					return fmt.Errorf("--prod-slots must be greater than zero")
```

begelundmuller (changes requested):

cli/cmd/sudo/project/edit.go
```diff
@@ -0,0 +1,64 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx := cmd.Context()
+			isEditRequested := false
+			req := &adminv1.UpdateProjectRequest{
+				OrganizationName:     args[0],
+				Name:                 args[1],
+				SuperuserForceAccess: &[]bool{true}[0],
+			}
+
+			if cmd.Flags().Changed("prod-slots") {
```
k-anshul, line 25: Nit: I think the code can be simplified a bit by checking first that `prod-slots` is set:
```
if !cmd.Flags().Changed("prod-slots") {
    ch.PrintfWarn("No edit requested\n")
    return nil
}
```
And then checking value is +ve
```
if prodSlots <= 0 {
   ch.PrintfError("--prod-slots must be greater than zero\n")
   return nil
}
```

k-anshul (changes requested):

cli/cmd/sudo/project/edit.go
NamanMahor: done

cli/cmd/sudo/project/edit.go
```diff
@@ -0,0 +1,64 @@
...
+
+			ch.PrintfSuccess("Updated project\n")
+			ch.PrintProjects([]*adminv1.Project{updatedProj.Project})
+
+			return nil
+		},
+	}
+
+	editCmd.Flags().IntVar(&prodSlots, "prod-slots", 0, "Slots to allocate for production deployments")
+	if !ch.IsDev() {
```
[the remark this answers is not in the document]
NamanMahor: done

admin/server/projects.go
NamanMahor: is there case where user have manage permission but not a superuser? then this kind of user can able to change the slots with rill project edit --prod-slots  with old version rill

begelundmuller: Yes that happens for any normal project admin. But it's fine, from an API perspective they should be allowed to edit slots (we already have quotas so they can't assign very high number of slots). However from a UX point-of-view, our product folks currently prefer us to handle slots assignments internally so we don't expose or document the ability to external users.

begelundmuller (approved):

k-anshul (approved):

PR #6683, opened 2025-02-14
fix information schema returning all table types as view

k-anshul:
closes https://github.com/rilldata/rill-private-issues/issues/1000

**Checklist:**
- [x] Covered by tests
- [ ] I'm proud of this work!


runtime/drivers/duckdb/information_schema.go
```diff
@@ -62,33 +39,18 @@ func (i informationSchema) All(ctx context.Context, like string) ([]*drivers.Tab
...
-			array_agg(c.is_nullable = 'YES' order by c.ordinal_position) as "column_nullable"
-		from information_schema.tables t
-		join information_schema.columns c on t.table_schema = c.table_schema and t.table_name = c.table_name
-		where database = current_database() and t.table_schema = current_schema() and lower(t.table_name) = lower(?)
-		group by 1, 2, 3
-		order by 1, 2, 3
-	`
-
-	rows, err := conn.QueryxContext(ctx, q, name)
+	rows, err := db.Schema(ctx, name, true)
```
begelundmuller, line 50: What if `name` has a special character in it? E.g. `LIKE` interprets underscores as "any character", so that will give conflicts for any snake case name.

runtime/pkg/rduckdb/information_schema.go
```diff
@@ -0,0 +1,79 @@
...
+	"fmt"
+)
+
+type Table struct {
+	Database       string `db:"database"`
+	Name           string `db:"name"`
+	View           bool   `db:"view"`
+	ColumnNames    []any  `db:"column_names"`
+	ColumnTypes    []any  `db:"column_types"`
+	ColumnNullable []any  `db:"column_nullable"`
```
begelundmuller, line 14: Shouldn't these be `[]string`, `[]string` and `[]bool`?

begelundmuller (changes requested):

runtime/drivers/duckdb/information_schema.go
k-anshul: Good point. I did not think about it. Will modify to take name separately

runtime/pkg/rduckdb/information_schema.go
k-anshul: I just ported this from the existing code and thought may be I am missing some case where datatype can be different. Can make them []string, []string and []bool.

k-anshul: Okay I see following error on scan :
```
sql: Scan error on column index 3, name "column_names": unsupported Scan, storing driver.Value type []interface {} into type *[]string
```

runtime/drivers/duckdb/information_schema.go
```diff
@@ -18,41 +18,18 @@ func (c *connection) InformationSchema() drivers.InformationSchema {
 	return &informationSchema{c: c}
 }
 
-func (i informationSchema) All(ctx context.Context, like string) ([]*drivers.Table, error) {
-	conn, release, err := i.c.acquireMetaConn(ctx)
+func (i informationSchema) All(ctx context.Context, ilike string) ([]*drivers.Table, error) {
+	// TODO: this bypasses the acquireMetaConn call in the original implementation. Fix this.
+	db, release, err := i.c.acquireDB()
```
begelundmuller, line 23: Is there any important code that relies on fast information schema results?

begelundmuller (approved):

runtime/drivers/duckdb/information_schema.go
k-anshul: Nothing on rill cloud.
And the `acquireDB` should return early since it bypass single metadata call semaphore.

PR #6695, opened 2025-02-17
do not delete billing and payment customer on org deletion

pjain1:
fixes - https://rilldata.slack.com/archives/C05TXA7QSTY/p1739514901815009


begelundmuller (approved):
LGTM

PR #6696, opened 2025-02-17
fix (e2e): Prevent CLI from opening URLs in the OS default browser

ericpgreen2:
Prevents the Rill CLI from automatically opening certain URLs in the OS default browser when running e2e tests with Playwright’s Chromium.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/cli/cmd/auth/login.go
+++ b/cli/cmd/auth/login.go
@@ -75,7 +75,9 @@ func Login(ctx context.Context, ch *cmdutil.Helper, redirectURL string) error {
 
 	ch.PrintfBold("\nOpen this URL in your browser to confirm the login: %s\n\n", deviceVerification.VerificationCompleteURL)
 
-	_ = browser.Open(deviceVerification.VerificationCompleteURL)
+	if ch.Interactive {
+		_ = browser.Open(deviceVerification.VerificationCompleteURL)
+	}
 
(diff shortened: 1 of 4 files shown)


cli/cmd/auth/login.go
```diff
@@ -75,7 +75,9 @@ func Login(ctx context.Context, ch *cmdutil.Helper, redirectURL string) error {
 
 	ch.PrintfBold("\nOpen this URL in your browser to confirm the login: %s\n\n", deviceVerification.VerificationCompleteURL)
 
-	_ = browser.Open(deviceVerification.VerificationCompleteURL)
+	if ch.Interactive {
```
AdityaHegde, line 78: Perhaps would be good to wrap the browser.Open? Like `ch.Open`? Would be good to make it a standard in all usages. There are a lot more places where we call `browser.Open`

AdityaHegde (commented):
Perhaps would be good

cli/cmd/auth/login.go
ericpgreen2: I see the motivation, but at the moment, I'm hesitant to do that since commands like `rill docs` and `rill sudo user open` are explicitly meant to open the browser. For now, I think it's fine as is.

AdityaHegde (approved):

PR #6697, opened 2025-02-17
Implement `rows:` in the canvas YAML

begelundmuller:
**Changes:**
- Introduces a `rows:` property on `type: canvas` resources.
- Each row has a `height` (optional integer with required `px` unit) and list of nested `items:`.
- Each item has a `width` (optional integer without unit)
- Items use `component: name` to reference external components; other properties are considered an inline component definition
- Removes `items:` as a root property on `type: canvas` resources/

**Example syntax:**
```yaml
type: canvas
...
rows:
  - items:
    - component: c1 # External component
  - height: 100px
    items:
    - component: c2
    - width: 2
      markdown: # Inline component definition
        content: Foo
```

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!


jkhwu:
@begelundmuller transcribing answers for you from @briangregoryholmes and @mindspank in our canvas war room meeting.
1. Both width and height can be optional, is it a requirement of protobuff that optional values are returned as 0? Would be great to have default numeric value be other than 0.
2. We're good with your proposed reduced-nesting structure.

briangregoryholmes (approved):

PR #6706, opened 2025-02-18
Disable cloud API calls during rill-dev E2E

AdityaHegde:
We make a few API calls to cloud in rill-dev to get user related info. While running these tests locally if there is a login token stored then cloud is hit. This is unnecessary load on cloud that should be disabled.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


web-common/src/features/authentication/LocalAvatarButton.svelte
ericpgreen2, line 1: It looks like the feature flag should be pulled into `ApplicationHeader.svelte` and conditionally render the whole `LocalAvatarButton` component

web-common/src/features/authentication/RepresentingUserBanner.svelte
ericpgreen2, line 1: Similarly, I think the feature flag should exist in the parent component and conditionally render the whole `RepresentingUserBanner`

web-common/src/features/dashboards/workspace/DeployProjectCTA.svelte
ericpgreen2, line 1: Here too, the feature flag should be placed in the parent component `ApplicationHeader`

web-common/src/features/feature-flags.ts
```diff
@@ -32,6 +32,14 @@ class FeatureFlags {
   readOnly = new FeatureFlag("rill", false);
 
   ai = new FeatureFlag("user", !import.meta.env.VITE_PLAYWRIGHT_TEST);
+  /**
+   * We make some calls to cloud for fetching user related stats. But this is not needed in playwring tests.
+   * While there wont be token in CI there could be one when running on a dev's laptop.
+   */
+  disableCloud = new FeatureFlag(
```
ericpgreen2, line 39: We should be clear on the difference between "disableCloud" and "adminServer" (which also is a toggle for Cloud features). It seems the difference is whether or not we show Cloud features _in Rill Developer_, so I'd propose something like: `rillDevCloudFeatures`, `rillDevAdminServer`, or something along those lines. 

Also, I prefer the feature flags are positive not negative. It's easier to read `if (feature)` rather than `if (!noFeature)`.

web-common/src/features/feature-flags.ts
```diff
@@ -32,6 +32,14 @@ class FeatureFlags {
...
+    "user",
```
ericpgreen2, line 40: This should be moved up to a "rill" feature flag because we don't anticipate users toggling this on themselves.

ericpgreen2 (approved):
Approving with various suggested changes

PR #6723, opened 2025-02-19
Fix wrong field in trial started log

kaspersjo:


admin/billing.go
```diff
@@ -277,11 +277,16 @@ func (s *Service) StartTrial(ctx context.Context, org *database.Organization) (*
 		return org, sub, nil
 	}
 
+	user, err := s.DB.FindUser(ctx, *org.CreatedByUserID)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to get user info: %w", err)
+	}
```
begelundmuller, line 283: Seems like it was an issue before, but it needs to check if the pointer is `nil` to protect against nil pointer access panics.
```suggestion
	var email string
	if org.CreatedByUserID != nil {
		user, err := s.DB.FindUser(ctx, *org.CreatedByUserID)
		if err != nil {
			return nil, nil, fmt.Errorf("failed to get user info: %w", err)
		}
		email = user.Email
	}
```

begelundmuller (changes requested):

grahamplata (approved):

begelundmuller (approved):

PR #6725, opened 2025-02-19
feat : Save user query execution time in model state

k-anshul:

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


runtime/drivers/duckdb/model_executor_self.go
```diff
@@ -95,20 +99,28 @@ func (e *selfToSelfExecutor) Execute(ctx context.Context, opts *drivers.ModelExe
...
+		stats.Size += prevResult.Size
 	}
 
 	// Build result props
 	resultProps := &ModelResultProperties{
 		Table:         tableName,
 		View:          asView,
 		UsedModelName: usedModelName,
+		Size:          stats.Size,
+		ExecTime:      stats.ExecTime,
```
begelundmuller, line 120: Did you consider adding these directly to the `ModelState`? It might make it easier/clearer for the UI to integrate and CLI to print.

The execution duration could be returned in `*drivers.ModelResult` and the table size could be loaded with a separate call to the `OLAPStore` interface (would require a new function, but that might be nice for other use cases, including introspecting ClickHouse table sizes).

Not a strong opinion though, will let you decide how you want to proceed.

k-anshul: >Did you consider adding these directly to the ModelState? It might make it easier/clearer for the UI to integrate and CLI to print.

Yeah I thought about it. But kept it in `result_properties` because the godoc mentions that it is supposed to keep metadata about the result. The client will then need to look at the field of the map. We can also add more fields in future.
But I can keep it at root level as well.

> The table size could be loaded with a separate call to the OLAPStore interface

Sure we can keep it a separate call as well. For clickhouse I planned to query information_schema and populate results but I am now wondering if the size changes due to background compaction in clickhouse and thus it may better to keep it a separate function.

runtime/reconcilers/model.go
```diff
@@ -1009,6 +1021,7 @@ func (r *ModelReconciler) executeAll(ctx context.Context, self *runtimev1.Resour
 		// Start the worker goroutines
 		grp, ctx := errgroup.WithContext(ctx)
 		counter := &atomic.Int64{}
+		duration := &atomic.Int64{}
```
begelundmuller, line 1024: Consider just making `totalExecDuration` an `atomic.Int64` from the start, then you don't need two different but similarly named variables

runtime/reconcilers/model.go
```diff
@@ -1067,11 +1082,12 @@ func (r *ModelReconciler) executeAll(ctx context.Context, self *runtimev1.Resour
 
 	// Should not happen, could also have been a panic
 	if prevResult == nil {
-		return "", nil, fmt.Errorf("partition execution succeeded but did not produce a non-nil result")
+		return "", nil, false, fmt.Errorf("partition execution succeeded but did not produce a non-nil result")
 	}
 
+	prevResult.ExecDuration = totalExecDuration
 	// We have continuously updated prevResult with new partition results, so we return it here
```
begelundmuller, line 1089:
```suggestion
	// We have continuously updated prevResult with new partition results, so we complete and return it here
	prevResult.ExecDuration = totalExecDuration
```

runtime/reconcilers/model.go
```diff
@@ -1139,6 +1155,7 @@ func (r *ModelReconciler) executeSingle(ctx context.Context, executor *wrappedMo
 		return nil, err
 	}
 
+	var stageDuration time.Duration
 	// Execute the stage step if configured
 	if executor.stage != nil {
```
begelundmuller, line 1160:
```suggestion
	// Execute the stage step if configured
	var stageDuration time.Duration
	if executor.stage != nil {
```

begelundmuller (changes requested):

begelundmuller (approved):

runtime/reconcilers/model.go
```diff
@@ -885,64 +891,69 @@ func (r *ModelReconciler) executeAll(ctx context.Context, self *runtimev1.Resour
...
 	err = r.resolveAndSyncPartitions(ctx, self, model, incrementalState)
 	if err != nil {
-		return "", nil, fmt.Errorf("failed to sync partitions: %w", err)
+		return "", nil, false, fmt.Errorf("failed to sync partitions: %w", err)
 	}
 
+	var (
+		totalExecDuration     time.Duration
+		firstRunIsIncremental = incrementalRun
+	)
```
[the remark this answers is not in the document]
k-anshul: Thanks. Definitely makes sense.

Issue #6711, opened 2025-02-18
Pivot: Flat table view


ericpgreen2:
https://www.notion.so/rilldata/Pivot-Table-Flat-and-Nested-176ba33c8f5780ea9e81dc49fd81707e

PR #6727, opened 2025-02-19
Feat: flat pivot table

djbarnwal:
Addresses https://github.com/rilldata/rill/issues/6711

Remaining Items - 

- [ ] Ability to re-arrange columns to mix measures and dimensions (_deferred to next PR due to scale of change across codebase_) https://github.com/rilldata/rill/pull/6784

**Checklist:**
- [x] Covered by tests
- [x] Checked if the docs need to be updated (*Updates needed*)
- [ ] I'm proud of this work!


mindspank:
First pass: https://www.notion.so/rilldata/Pivot-Table-Flat-and-Nested-176ba33c8f5780ea9e81dc49fd81707e?pvs=4#176ba33c8f578178a682ea2f8df3fd7d

Small bugs, nothing major. Looks good!

mindspank:
Looking forward to tests!

djbarnwal:
@AdityaHegde tagging you for review of URL and relevant proto changes, @ericpgreen2 for other files.

web-common/src/features/dashboards/url-state/convertPresetToExploreState.ts
```diff
@@ -494,7 +510,7 @@ function fromPivotUrlParams(
         rowPage: 1,
         enableComparison: true,
         activeCell: null,
-        rowJoinType: "nest",
+        rowJoinType: rowJoinType,
```
AdityaHegde, line 513: nit: can be simplified with just `rowJoinType,`

AdityaHegde (changes requested):
You also need to update `dashboard.proto` and add proto conversions. Bookmarks and public urls still use the old proto.

AdityaHegde (approved):
Approving from URL params POV. `PivotRowJoinType` is already in dashboard.proto so it is not in the diffs.

web-common/src/features/dashboards/url-state/url-params.ts
```diff
@@ -25,4 +25,5 @@ export enum ExploreStateURLParams {
 
   PivotRows = "rows",
   PivotColumns = "cols",
+  PivotRowJoinType = "table_type",
```
ericpgreen2, line 28: We should use the same terminology internally as externally. I see that `PivotRowJoinType` was a term added 13 months ago, but AFAIK we're not using it yet, so IMO we should rename it to be more intuitive and to match what we communicate externally.

web-common/src/features/dashboards/pivot/pivot-utils.ts
```diff
@@ -516,25 +515,61 @@ export function getSortFilteredMeasureBody(
   return { sortFilteredMeasureBody, isMeasureSortAccessor, sortAccessor };
 }
 
+function getValuesForFlatTable(
+  tableData: PivotDataRow[],
+  rowDimensions: string[],
```
ericpgreen2, line 520: Why are these the `rowDimensions`, not the `colDimensions`? Given the pills are labeled as "Columns":

image

web-common/src/features/dashboards/stores/dashboard-stores.ts
```diff
@@ -300,12 +300,19 @@ const metricsViewReducers = {
...
+          if (!validAccessors.includes(accessor)) {
+            metricsExplorer.pivot.sorting = [];
+          }
+        } else {
+          const anchorDimension = metricsExplorer.pivot.rows.dimension?.[0]?.id;
+          if (accessor !== anchorDimension) {
+            metricsExplorer.pivot.sorting = [];
+          }
         }
       }
```
ericpgreen2, line 317: Note: there's a lot of nesting here

web-common/src/features/dashboards/pivot/FlatTable.svelte
```diff
@@ -0,0 +1,304 @@
...
+    COLUMN_WIDTH_CONSTANTS as WIDTHS,
+  } from "@rilldata/web-common/features/dashboards/pivot/pivot-column-width-utils";
+  import Resizer from "@rilldata/web-common/layout/Resizer.svelte";
+  import { modified } from "@rilldata/web-common/lib/actions/modified-click";
+  import type { Cell, Column, HeaderGroup, Row } from "@tanstack/svelte-table";
+  import { flexRender } from "@tanstack/svelte-table";
+  import type { PivotDataRow } from "./types";
+
+  // State props
+  export let assembled: boolean;
```
ericpgreen2, line 21: What's this mean?

ericpgreen2 (changes requested):
Looks pretty good!

In general, I'm wary of many `isFlat` conditionals that now permeate the pivot code. There's a lot of nested conditionals. As the pivot/table configuration grows, we should think about how we can avoid even more conditionals (though I don't have a concrete proposal at the moment).

web-common/src/features/dashboards/url-state/url-params.ts
mindspank: I would prefer if we avoid `row_type` or `row_join_type` for external communication. From a user pov it's a table that has various modes/types so I would stick with `table_type` or `table_mode`

proto/rill/runtime/v1/resources.proto
```diff
@@ -427,6 +427,7 @@ message ExplorePreset {
   repeated string pivot_cols = 25;
   optional string pivot_sort_by = 26;
   optional bool pivot_sort_asc = 27;
+  optional string pivot_row_join_type = 28;
```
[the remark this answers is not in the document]
mindspank: See my comment below, if we want to unify the naming of properties and url_params

web-common/src/features/dashboards/url-state/url-params.ts
ericpgreen2: I very much agree `row_type` / `row_join_type` isn't good. I'm okay with `table_type` or `table_mode`

web-common/src/features/dashboards/pivot/FlatTable.svelte
djbarnwal: There are multiple stages in a pivot construction. `assembled: true` signifies that all the stages have been completed.

djbarnwal:
> Looks pretty good!
> 
> In general, I'm wary of many `isFlat` conditionals that now permeate the pivot code. There's a lot of nested conditionals. As the pivot/table configuration grows, we should think about how we can avoid even more conditionals (though I don't have a concrete proposal at the moment).

Have made the proto changes to rename `rowJoinType` to `tableMode`. 

FWith regards to nesting, the other option is to create a separate data store for flat table. Though, the implementation paths for both are very similar with flat table needed few steps so a lot of the conditionals are used to skip those steps. We can create a different data store for flat but it would mean duplicating a lot of existing nested table code.

web-common/src/features/dashboards/pivot/pivot-utils.ts
djbarnwal: This is because the way implementation is set up for Pivot tables. Column dimension internally are treated as values on which measures would be split and hence have a very different way of working. For flat table we need just a basic query so it made sense to use rowDimensions as a proxy for implementing it. I have added a comment for this in `pivot-data-config`.

web-common/src/features/dashboards/pivot/DragList.svelte
```diff
@@ -208,7 +213,7 @@
       />
 
       <div class="icons">
-        {#if zone === "Time" || zone === "Dimensions"}
+        {#if (zone === "Time" || zone === "Dimensions") && rowJoinType === "nest"}
```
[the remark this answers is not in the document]
djbarnwal: Can't seem to do it. Probably have to create a separate enum for this.

`Type 'DashboardState_PivotTableMode.NEST' is not assignable to type 'PivotTableMode'.ts(2322)`

ericpgreen2:
> With regards to nesting, the other option is to create a separate data store for flat table. Though, the implementation paths for both are very similar with flat table needed few steps so a lot of the conditionals are used to skip those steps. We can create a different data store for flat but it would mean duplicating a lot of existing nested table code.

I discussed it with Claude-3.7-sonnet in Cursor, and got this high-level recommendation:
```ts
// Create specialized stores
function createFlatPivotDataStore(ctx, config) {
  // Flat-specific implementation
}

function createNestedPivotDataStore(ctx, config) {
  // Nested-specific implementation
}

// Facade that delegates to the appropriate store
export function createPivotDataStore(ctx, configStore) {
  return derived(configStore, (config, set) => {
    const storeFactory = config.isFlat 
      ? createFlatPivotDataStore 
      : createNestedPivotDataStore;
    
    return storeFactory(ctx, config).subscribe(set);
  });
}
```

For any of the common processing steps, we could of course use a common utility function. But I know this might be a largish refactor of what you have, so I will defer to you on what you think is best.

ericpgreen2 (approved):

web-common/src/features/dashboards/url-state/url-state-variations.spec.ts
```diff
@@ -41,6 +41,7 @@ import {
   AD_BIDS_TOGGLE_BID_DOMAIN_DIMENSION_VISIBILITY,
   AD_BIDS_TOGGLE_BID_PRICE_MEASURE_VISIBILITY,
   AD_BIDS_TOGGLE_PIVOT,
+  AD_BIDS_TOGGLE_PIVOT_ROW_JOIN_TYPE,
```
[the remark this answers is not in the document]
djbarnwal: Resolved in https://github.com/rilldata/rill/pull/6784/commits/8f31215c209baf98f6f5fb318814ecfc1a7b6687

PR #6730, opened 2025-02-20
tweak: rewrite kpi component, canvas cleanup

briangregoryholmes:
Screenshot 2025-02-19 at 8 56 41 PM
Screenshot 2025-02-19 at 8 56 59 PM

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!


djbarnwal:
Great work! Looks neat and exciting to see the new changes to KPI. Some initial thoughts - 

UXQA
1. Charts size is broken (blocker)
2. We need an empty state for the KPI
3. When local time filters are applied to KPI without a local comparison range, do not inherit global comparison (blocker)

https://jam.dev/c/417355de-240e-4002-902b-95822763247d


image

This feels a bit too dense. I wonder if we can add some padding on both sides of the `/` separator.

image

Hide blue dot when null

djbarnwal (changes requested):
Blockers

- Expanding chart container
- Unable to set local comparison range for KPI grid
- Do not inherit global range when local time range is set

briangregoryholmes:
- Hid points when null
- Added padding between total hovered value
- Resolved chart container issue, though needs some follow up tweaks
- Updated the conditional on when to show comparisons

`Unable to set local comparison range for KPI grid`

This only happens on first load and seems to be an issue in `canvas-entity.ts`. Can you take a look at this one and hold off on blocking this PR?

djbarnwal (approved):

djbarnwal (commented):
UXQA:

The main line dot should be above comparison dot - 
image

KPI layout is broken when sparkline is placed on the right

image

If primary point is null and comparison point has a value, `no change` is incorrectly displayed

image

Adding a title breaks layout and increase row height 
image

PR #6736, opened 2025-02-21
feat: small improvements to search

AdityaHegde:
- Add loading state to dimension filter search.
- Add error state to dimension filter search.
- Limit the leaderboard to max 15 values for above and below fold values.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-common/src/features/dashboards/filters/dimension-filters/DimensionFilter.svelte
+++ b/web-common/src/features/dashboards/filters/dimension-filters/DimensionFilter.svelte
@@ -3,7 +3,8 @@
   import { Chip } from "@rilldata/web-common/components/chip";
   import RemovableListBody from "@rilldata/web-common/components/chip/removable-list-chip/RemovableListBody.svelte";
   import * as DropdownMenu from "@rilldata/web-common/components/dropdown-menu";
-  import SearchableMenuContent from "@rilldata/web-common/components/searchable-filter-menu/SearchableMenuContent.svelte";
+  import LoadingSpinner from "@rilldata/web-common/components/icons/LoadingSpinner.svelte";
+  import { Search } from "@rilldata/web-common/components/search";
   import Tooltip from "@rilldata/web-common/components/tooltip/Tooltip.svelte";
   import TooltipContent from "@rilldata/web-common/components/tooltip/TooltipContent.svelte";
   import TooltipTitle from "@rilldata/web-common/components/tooltip/TooltipTitle.svelte";
@@ -40,13 +41,14 @@
     timeEnd,
     Boolean(timeControlsReady && open),
   );
+  $: ({ data, error, isFetching } = $searchValues);
 
   $: allSelected = Boolean(
-    selectedValues.length && $searchValues?.length === selectedValues.length,
+    selectedValues.length && data?.length === selectedValues.length,
   );
 
   function onToggleSelectAll() {
-    $searchValues?.forEach((dimensionValue) => {
+    data?.forEach((dimensionValue) => {
       if (!allSelected && selectedValues.includes(dimensionValue)) return;
 
       onSelect(dimensionValue);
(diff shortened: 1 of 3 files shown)


web-common/src/features/dashboards/filters/dimension-filters/dimensionFilterValues.ts
```diff
@@ -37,15 +38,37 @@ export function useDimensionSearch(
...
-      return true;
-    });
+    return {
+      data: values.filter((value) => {
+        if (seen.has(value)) {
+          return false;
+        }
+        seen.add(value);
+        return true;
+      }),
```
ericpgreen2, line 69: Suggestion to remove this `seen` complexity:
```
const values = items.map((item) => item[dimensionName] as string);
// Convert to Set and back to array to remove duplicates
const data = [...new Set(values)];
return ...
```

web-common/src/features/dashboards/leaderboard/Leaderboard.svelte
```diff
@@ -221,6 +222,7 @@
       timeRange,
       comparisonTimeRange,
       measures,
+      limit: (maxValuesToShow - aboveTheFold.length).toString(),
```
ericpgreen2, line 225: I'm wondering if we'd get a `400` if `limit` were 0 or negative. Maybe guard against `limit <= 0` in the `enabled` clause?

web-common/src/features/dashboards/filters/dimension-filters/dimensionFilterValues.ts
```diff
@@ -37,15 +38,37 @@ export function useDimensionSearch(
   );
 
   return derived(queries, ($queries) => {
+    const someQueryFetching = $queries.some((q) => q.isFetching);
```
ericpgreen2, line 41: Typically we should use `isLoading` not `isFetching`

web-common/src/features/dashboards/filters/dimension-filters/DimensionFilter.svelte
```diff
@@ -113,29 +115,87 @@
...
-    selectableGroups={[
-      {
-        name: "DIMENSIONS",
-        items: $searchValues.map((dimensionValue) => ({
-          name: dimensionValue,
-          label: dimensionValue,
-        })),
-      },
-    ]}
+  <!-- There will be some custom controls for this. Until we have the full design have a custom dropdown here. -->
```
ericpgreen2, line 118: I'm a little worried about divergence between our menu components. For example, all other `SearchableMenuContent` would also benefit from the loading & error states you've added here. I wonder if we could add the requisite functionality to `SearchableMenuComponent` (e.g. make it pluggable?). I'll leave it up to you.

ericpgreen2 (approved):

web-common/src/features/dashboards/filters/dimension-filters/DimensionFilter.svelte
AdityaHegde: Ya I wanted to wait for the design to stabilise before refactoring this too much. The end goal would be a common component with hooks for dimension filter.

web-common/src/features/dashboards/leaderboard/Leaderboard.svelte
AdityaHegde: It shouldnt happen, but doesnt hurt to add a safeguard.

web-common/src/features/dashboards/filters/dimension-filters/dimensionFilterValues.ts
AdityaHegde: Hmmm my understanding was `isLoading` is true only on the 1st fetch. But i guess it is for the 1st fetch of the query object. Since we create a new one on every searchText it will be true for showing the spinner.

AdityaHegde: But if for whatever reason there a refetch the data would go from blank to showing values without a spinner. So I think it is fine to use `isFetching` here.

PR #6748, opened 2025-02-24
feat: Add gzip compression for url params to support large filters

AdityaHegde:
We need to support bulk dimension filters. As one of the steps we need to be able to support larger urls. This PR does a 1st pass by compressing the url using gzip. The end goal would be using bookmarks.

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!


web-common/src/features/dashboards/url-state/convertURLToExplorePreset.ts
```diff
@@ -41,7 +42,7 @@ import {
   V1Operation,
 } from "@rilldata/web-common/runtime-client";
 
-export function convertURLToExplorePreset(
+export async function convertURLToExplorePreset(
```
ericpgreen2, line 45: (Apologies if this is outside the scope of this PR) Why do we need this function at all? Shouldn't we always be converting a URL directly to `ExploreState`?

web-common/src/features/dashboards/url-state/convertURLToExplorePreset.ts
```diff
@@ -64,6 +65,8 @@ export function convertURLToExplorePreset(
     (d) => d.name!,
   );
 
+  searchParams = await decompressUrlParams(searchParams);
```
ericpgreen2, line 68: Along the lines of my other comments, I'd propose we pull the conditional out of `decompressUrlParams` and into this function here, e.g. something like:
```ts
export async function convertURLToExploreState(url, ...) {
    let urlState;
    if (url.searchParams.has(ExploreStateURLParams.GzippedParams)) {
        urlState = decompressUrlSearchParams(searchParams)
    } else {
        urlState = url.searchParams;
    }
}
```

ericpgreen2 (changes requested):

web-common/src/features/dashboards/url-state/convertURLToExplorePreset.ts
AdityaHegde: I have this on the backlog for a refactor. Out of scope of this PR

web-admin/src/features/dashboards/query-mappers/utils.ts
```diff
@@ -205,7 +205,7 @@ export async function getExplorePageUrl(
     });
   }
 
-  url.search = convertExploreStateToURLSearchParams(
+  url.search = await convertExploreStateToURLSearchParams(
```
ericpgreen2, line 208: I'm generally concerned that so many of these functions have turned async

web-common/src/features/dashboards/url-state/DashboardURLStateSync.svelte
```diff
@@ -117,60 +117,7 @@
...
-    }
-
-    prevUrl = redirectUrl.toString();
-    // using `replaceState` directly messes up the navigation entries,
-    // `from` and `to` have the old url before being replaced in `afterNavigate` calls leading to incorrect handling.
-    void goto(redirectUrl, {
-      replaceState: true,
-      state: $page.state,
-    });
+    void handleAfterNavigate(to.url, backButtonUsed);
```
ericpgreen2, line 120: Given all the above code in the `afterNavigate` scope, it's not entirely clear to me what's supposed to be encapsulated in the `handleAfterNavigate` function.

web-common/src/features/dashboards/url-state/compression.ts
ericpgreen2, line 1: Given URL parameter processing is inherently a blocking operation that needs to happen immediately during page load, and to avoid the proliferation of `await` statements, could we make these compression functions synchronous? I did some quick research and found https://github.com/101arrowz/fflate.

ericpgreen2 (changes requested):

web-common/src/features/dashboards/url-state/compression.ts
AdityaHegde: I wanted to keep it async because the end goal will probably be to store data in the backend. But there will be other race condition issues when a network call is involved.

web-common/src/features/dashboards/url-state/DashboardURLStateSync.svelte
AdityaHegde: Reverted since we dont use async methods

web-common/src/features/dashboards/url-state/convertExploreStateToURLSearchParams.ts
```diff
@@ -75,14 +80,19 @@ export function convertExploreStateToURLSearchParams(
   // But it is only available in TimeControlState and not MetricsExplorerEntity
   timeControlsState: TimeControlState | undefined,
   preset: V1ExplorePreset,
+  // Used to decide whether to compress or not based on the full url length
+  url: URL,
```
ericpgreen2, line 84: Nit: it'd be nice if we didn't expand (i.e. make more complex) the function signature. What if we decided to compress based on the length of the search parameters, not the full URL? We'd trade-off ultra precision for more simplicity.

web-common/src/features/dashboards/url-state/convertExploreStateToURLSearchParams.ts
```diff
@@ -75,14 +80,19 @@ export function convertExploreStateToURLSearchParams(
...
+  disableCompression = false,
```
ericpgreen2, line 85: If this were provided as an object, it'd be more clear what it represents where it's called:
```
// init-embed-api.ts
convertExploreStateToURLSearchParams(
  dashboard,
  exploreSpec,
  timeControlsState,
  defaultExplorePreset,
  { disableCompression: true },
),
```

ericpgreen2 (approved):
I left two more suggestions, then LGTM!

web-common/src/features/dashboards/url-state/convertExploreStateToURLSearchParams.ts
AdityaHegde: Hmmm with the new token paths this will start becoming less accurate. Also it depends on project and dashboard names. So lets keep the check on URL

web-common/src/features/dashboards/url-state/convertExploreStateToURLSearchParams.ts
AdityaHegde: Huh, that sounds bad. I'd expect developers to look at the function signature. Adding an object for a single optional params seems unnessary.

ericpgreen2: Ok maybe it's overkill in this scenario, but I'm thinking of it similar to how TanStack's `queryOptions` are passed as an object, and all call sites are very readable.

AdityaHegde: Hmmm probably makes sense in that case. There are a lot of params for in `queryOptions`, not having an object will make it ugly.

web-common/src/features/dashboards/url-state/convertExploreStateToURLSearchParams.ts
ericpgreen2: To play this out...

Here's a reasonably small URL:
```
https://ui.rilldata.com/my-org/my-project/explore/my-explore-name
```

Here's a roughly maximum size URL (since we limit org names to 40 characters):
```
https://forty-character-custom-domain-xxxxxxxxxx.com/forty-character-organization-name-yyyyyy/forty-character-project-name-zzzzzzzzzzz/-/share/rill_mgc_8U2ofiQ9ehmcXz0JhbPEE2CMYOnSYg680Ed3wZEl3WbRg0anZerwNW/explore/forty-character-explore-name-aaaaaaaaaaa
```

65 vs 256 characters, so the difference is 191 characters.

I could see it either way. I support any push for code simplicity, but I also get that reducing the capacity of human-readable URL length by ~10% might be a little much.

AdityaHegde: Lets keep this, our compression ration is already pretty low.

PR #6763, opened 2025-02-25
feat: visual canvas editor yaml structure integration and refinement

briangregoryholmes:
UI integration of the Visual Canvas editor with the new YAML structure.

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!


web-common/src/features/canvas/stores/spec.ts
```diff
@@ -214,21 +214,27 @@ export class CanvasResolvedSpec {
...
-    const componentName = this.getComponentNameFromIndex(index);
+  getComponentFromIndex = (pos: {
+    row: number;
+    column: number;
+  }): Readable<V1ComponentSpec | undefined> => {
+    const componentName = this.getComponentNameFromPos(pos);
     return derived(
       [componentName, this.components],
       ([componentName, components]) => {
+        console.log(componentName, components);
```
djbarnwal, line 237: Stale log

web-common/src/features/canvas/stores/canvas-entity.ts
```diff
@@ -107,12 +109,13 @@ export class CanvasEntity {
     const componentSpecStore = spec.getComponentResourceFromName(componentName);
 
     return derived(componentSpecStore, (componentSpec, set) => {
+      console.log(componentSpec);
```
djbarnwal, line 112: Leftover log

web-common/src/features/canvas/layout-util.ts
```diff
@@ -0,0 +1,227 @@
...
+  bar_chart: 400,
+  area_chart: 400,
+  stacked_bar: 400,
+  stacked_bar_normalized: 400,
+  markdown: 160,
+  kpi: 200,
+  kpi_grid: 200,
+  image: 420,
+  table: 400,
+};
```
djbarnwal, line 20: I am wondering if these values should be inside the component class. We already have `minSize` and `defaultSize` which are not being used.

Having them inside the class would co-locate all component related config at one place. Secondly, in future if we want these values to be derived from the current properties of the component (for ex. min height for a table with just measures would be different to a table with row dimensions) we can expose a method from the component's class.

web-common/src/features/canvas/layout-util.ts
```diff
@@ -0,0 +1,227 @@
...
+
+type YAMLRow = {
+  items: (YAMLItem | null)[];
+  height?: string;
+};
+
+type V1CanvasRow = Omit<APIV1CanvasRow, "items"> & {
+  items: (V1CanvasItem | null)[];
+};
+
```
djbarnwal, line 40: Can you add some context on why we need to modify this type?

djbarnwal (approved):
Looks good! Have left some nit comments. Feel free to merge after addressing the CI issues

Found some UX issues. Not sure if these are related but adding it anyway. We can triage them later.

On selecting markdown block, I would expect to see a border. It's confusing without the border if the block is selected or not

image

Table component has multiple issues - 
https://jam.dev/c/8da7cdc5-90e8-47f0-861c-94ebc1e59d73

Sorting is not registered most of the time (Can be reproduced by editing markdown and then interacting with the table)

When sorting is possible, the arrow direction changes but it does not update the table contents (https://github.com/rilldata/rill/pull/6767)

Adding a row or column doesn't update the table until refreshed (https://github.com/rilldata/rill/pull/6767)

Do not show KPI comparison line when comparison toggle in the sidebar is off
image

PR #6773, opened 2025-02-26
feat: advanced dimension filter

### LINKED ISSUE #6710: Bulk and advanced filtering
[PRD in Notion](https://www.notion.so/rilldata/Bulk-and-advanced-filtering-196ba33c8f5780b4b25dce617ce55ed6)

AdityaHegde:
closes #6710

- [x] Add comma separated "In List" search mode that is essentially an `in` filter.
- [x] Add a "Contains" search mode that is essentially a `like` filter.
- [x] Improve the way we differentiate select and bulk list since both are `in` query in the end.
- [x] Compress the url params using gzip https://github.com/rilldata/rill/pull/6748
- [ ] ~~Add tests for canvas filters.~~ Will be in a separate PR
- [x] UXQA 

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!


mindspank (approved):

ericpgreen2:
Leaving our notes from our live discussion about a filter state refactor:
- This PR further motivates a refactor to our client-side filters state. It's now making less sense for our client-side mental model to be anchored on the `where: V1Expression` parameter that the backend expects. Rather, our mental model should mimic the user-facing UX, which now notably includes different filter modes. 
- Our dimension filter state should revolve around a type that makes `mode` explicit, e.g.:
    ```ts
   interface DimensionFilter {
      dimensionName: string;
      mode: "Select" | "Contains" | "In List";
      payload: string | string[]; // this is just for illustration because we should avoid this mixed type
      exclude: boolean;
    }
    ```
- Regardless of what the URL parameters are, we can deserialize them into the above structure (e.g. `const dimensionFilters = filters.fromURLParams(...)`). Regardless of what form our client-side mental model takes, we can serialize it into the `where` property that the metrics view queries expect (e.g. `const where = filters.toWhereExpression(...)`.

ericpgreen2 (changes requested):
Leaving various observations from my first pass

web-common/src/components/button/IconButton.svelte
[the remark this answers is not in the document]
AdityaHegde: Hmmm not sure why this change showed up. It is not being used at all.

web-common/src/features/dashboards/filters/dimension-filters/DimensionFilter.svelte
```diff
@@ -1,21 +1,37 @@
...
+    useDimensionSearch,
+    useAllSearchResultsCount,
+  } from "./dimensionFilterValues";
 
   export let name: string;
   export let metricsViewNames: string[];
   export let label: string;
   export let selectedValues: string[];
+  export let searchText: string | undefined;
+  export let isMatchList: boolean | undefined;
```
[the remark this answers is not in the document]
AdityaHegde: We can reuse `DimensionFilterMode`

web-common/src/features/dashboards/filters/Filters.svelte
```diff
@@ -405,6 +409,10 @@
                 onToggleFilterMode={() => toggleDimensionFilterMode(name)}
                 onSelect={(value) =>
                   toggleDimensionValueSelection(name, value, true)}
+                onBulkSelect={(values) =>
+                  applyDimensionBulkSearch(name, values)}
```
ericpgreen2, line 413: We need to use consistent terminology. It's hard to keep track of "bulk select" vs "bulk search" vs  "in list".

web-common/src/features/dashboards/filters/Filters.svelte
```diff
@@ -405,6 +409,10 @@
...
+                onSearch={(searchText) =>
+                  applyDimensionSearch(name, searchText)}
```
ericpgreen2, line 415: Similarly, this should more clearly map to the UI's "contains" terminology

ericpgreen2 (changes requested):
I haven't finished going through the whole PR yet, but here's some feedback for now.

UXQA:

1. The menu item hover state is mis-aligned with the above elements. 
    image
2. Currently, when I submit the "Contains" filter with an empty input, all values are selected. Instead, there should be form validation that prevents the form from being submitted with an empty string. There should be a small error message in red text.
    image
3. Similar to 2, the InList mode needs form validation to prevent a filter from having an empty string input.
4. Why does the "Contains" filter match a different number of values in the dashboard (700) and the alert context (1018)?
    image

web-admin/src/features/alerts/metadata/AlertFilters.svelte
```diff
@@ -47,8 +48,13 @@
     (measure) => measure.name,
   );
 
-  $: currentDimensionFilters = getDimensionFilters(dimensionIdMap, whereFilter);
+  $: currentDimensionFilters = getDimensionFilters(
+    dimensionIdMap,
+    whereFilter,
+    dimensionsWithInlistFilter,
+  );
   $: currentMeasureFilters = getMeasureFilters(measureIdMap, havingFilter);
```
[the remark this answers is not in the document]
AdityaHegde: You mean just for `dimensionsWithInlistFilter`. The other fields are taken from the query, those could be different.

web-admin/src/features/alerts/selectors.ts
```diff
@@ -97,3 +102,40 @@ export function useIsAlertCreatedByCode(instanceId: string, name: string) {
     },
   );
 }
+
+export function useAlertDashboardState(
+  instanceId: string,
+  alertSpec: V1AlertSpec | undefined,
+) {
```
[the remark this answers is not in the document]
AdityaHegde: Ya that is one option. It would lead to `.subscribe(set)` pattern so I think it is better this way. Besides both consumers of this already have `alertSpec`

web-common/src/components/tag/Tag.svelte
```diff
@@ -15,6 +15,7 @@
   export let color: Color = "gray";
   export let height = 21;
   export let text: string = "";
+  export let noSpan = false;
```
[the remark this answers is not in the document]
AdityaHegde: Seems like another merge issue

web-common/src/features/canvas/stores/filters.ts
```diff
@@ -556,10 +623,12 @@ export class Filters {
   };
 
   getFiltersFromText = (filterText: string) => {
-    let expr = convertFilterParamToExpression(filterText);
-    if (
-      expr?.cond?.op !== V1Operation.OPERATION_AND &&
-      expr?.cond?.op !== V1Operation.OPERATION_OR
+    let { expr } = convertFilterParamToExpression(filterText); // TODO: use dimensionsWithInlistFilter
```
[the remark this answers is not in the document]
AdityaHegde: yes, good catch

AdityaHegde:
Thanks @ericpgreen2 

For 1, I see its the case on prod as well. Lets address it as a bug fix in a follow up.
Addressed 2 and 3 but disabling apply. There is no design for error message. Since we disable apply it should be clear to the user. Lets take the improvement in a follow up.
Fixed 4 by using the correct time ranges.

web-common/src/features/dashboards/state-managers/selectors/dimension-filters.ts
```diff
@@ -21,22 +25,86 @@ export const selectedDimensionValues = (
...
+      enabled: true,
+    });
+  }
+
+  return readable({
+    isFetching: false,
+    isLoading: false,
+    error: undefined,
+    data: [],
+  });
```
ericpgreen2, line 105: When will this get hit? Should this return an error?

web-common/src/features/dashboards/state-managers/selectors/dimension-filters.ts
```diff
@@ -21,22 +25,86 @@ export const selectedDimensionValues = (
...
+  const dimExpr = whereFilter.cond?.exprs?.find((e) =>
+    matchExpressionByName(e, dimensionName),
+  );
+  if (!dimExpr?.cond?.op)
+    return readable({
+      isFetching: false,
+      isLoading: false,
+      error: undefined,
+      data: [],
+    });
```
ericpgreen2, line 72: Is this a type guard? Or should this return an error?

web-common/src/features/dashboards/state-managers/selectors/dimension-filters.ts
```diff
@@ -21,22 +25,86 @@ export const selectedDimensionValues = (
...
+  dimensionName: string,
+  timeStart?: string,
+  timeEnd?: string,
+): CompoundQueryResult<string[]> => {
+  // if it is a complex filter unsupported by UI then no values are selected
+  if (!whereFilter || isExpressionUnsupported(whereFilter))
+    return readable({
+      isFetching: false,
+      isLoading: false,
+      error: undefined,
```
ericpgreen2, line 59: Should there be an error here, or no?

web-common/src/features/dashboards/state-managers/actions/dimension-filters.ts
```diff
@@ -63,6 +103,55 @@ export function toggleDimensionValueSelection(
...
+  values: string[],
+) {
+  if (dashboard.temporaryFilterName !== null) {
+    dashboard.temporaryFilterName = null;
+  }
+
+  if (!dashboard.whereFilter.cond?.exprs) return;
+
+  const isInclude = !dashboard.dimensionFilterExcludeMode.get(dimensionName);
+  const expr = createInExpression(dimensionName, values, !isInclude);
```
ericpgreen2, line 118: Here there's still a double negation. Why not use `isExclude`?

web-common/src/features/dashboards/state-managers/actions/dimension-filters.ts
```diff
@@ -63,6 +103,55 @@ export function toggleDimensionValueSelection(
...
+  }
+
+  if (!dashboard.whereFilter.cond?.exprs) return;
+
+  const isInclude = !dashboard.dimensionFilterExcludeMode.get(dimensionName);
+  const expr = createLikeExpression(
+    dimensionName,
+    `%${searchText}%`,
+    !isInclude,
+  );
```
ericpgreen2, line 146: Double negation

web-common/src/features/dashboards/filters/dimension-filters/DimensionFilter.spec.ts
```diff
@@ -0,0 +1,252 @@
...
+  beforeAll(() => {
+    mocks.mockMetricsView(AD_BIDS_METRICS_NAME, AD_BIDS_METRICS_INIT);
+    mocks.mockMetricsExplore(
+      AD_BIDS_EXPLORE_NAME,
+      AD_BIDS_METRICS_INIT,
+      AD_BIDS_EXPLORE_INIT,
+    );
+  });
+
+  it("Select mode filter", async () => {
```
ericpgreen2, line 36: This is a bit hard to follow. At first, given the name of the test. I was confused when I saw "Contains" logic. It'd help if there were sectional comments at the _beginning_ of the "Contains" and "In list" parts of the test, not just at the end where those assertion happens.

web-common/src/features/dashboards/filters/dimension-filters/DimensionFilter.spec.ts
```diff
@@ -0,0 +1,252 @@
...
+    // "In List" mode does not persist since Apply was not clicked
+    await act(() => screen.getByLabelText("Open publisher filter").click());
+    await waitFor(() =>
+      expect(screen.getByLabelText("Open publisher filter")).toHaveTextContent(
+        "publisher Facebook +1 other",
+      ),
+    );
+  });
+
+  it("Search mode filter", async () => {
```
ericpgreen2, line 102: Outdated terminology – should be "contains"

web-common/src/features/dashboards/filters/dimension-filters/DimensionFilter.spec.ts
```diff
@@ -0,0 +1,252 @@
...
+      createAndExpression([
+        createLikeExpression(AD_BIDS_PUBLISHER_DIMENSION, "%oo%"),
+      ]),
+    );
+    expect(screen.getByLabelText("Open publisher filter")).toHaveTextContent(
+      "publisher Contains oo (3)",
+    );
+  });
+
+  it("Bulk mode filter using dropdown", async () => {
```
ericpgreen2, line 146: Outdated terminology – should be "in list"

web-common/src/features/dashboards/filters/dimension-filters/DimensionFilter.spec.ts
```diff
@@ -0,0 +1,252 @@
...
+    );
+    expect(
+      get(stateManagers.dashboardStore).dimensionsWithInlistFilter,
+    ).toEqual(["publisher"]);
+    expect(screen.getByLabelText("Open publisher filter")).toHaveTextContent(
+      "publisher In list (2 of 3)",
+    );
+  });
+
+  it("Bulk mode filter using search text", async () => {
```
ericpgreen2, line 213: Outdated terminology

web-common/src/features/dashboards/filters/dimension-filters/DimensionFilter.spec.ts
ericpgreen2: Sectional comments would help within each of this file's tests

web-common/src/features/dashboards/filters/dimension-filters/dimension-filter-values.ts
ericpgreen2, line 1: Can you please add function-level comments to this file? It'd help to more easily understand what they're doing.

web-common/src/features/dashboards/filters/dimension-filters/DimensionFilter.svelte
```diff
@@ -24,54 +40,237 @@
...
+  let curExcludeMode = excludeMode;
 
   $: ({ instanceId } = $runtime);
 
-  $: searchValues = useDimensionSearch(
+  $: resetFilterSettings(mode, sanitisedSearchText);
+
+  $: checkSearchText(curSearchText);
+
+  let searchedBulkValues: string[] =
```
ericpgreen2, line 60: Nit: Can you please add some sensible line breaks from lines 60 to 114? Without any line breaks, it's pretty overwhelming.

ericpgreen2 (changes requested):

web-common/src/features/dashboards/state-managers/selectors/dimension-filters.ts
AdityaHegde: It shouldnt be hit. Adding an error to make the code readable.

web-common/src/features/dashboards/state-managers/selectors/dimension-filters.ts
AdityaHegde: This is the case when a dimension has no filter. It should return empty array similar to how it was before.

web-admin/src/features/public-urls/CreatePublicURLForm.svelte
```diff
@@ -242,9 +245,12 @@
             <FilterChipsReadOnly
               exploreName={$exploreName}
               filters={$dashboardStore.whereFilter}
+              dimensionsWithInlistFilter={$dashboardStore.dimensionsWithInlistFilter}
               dimensionThresholdFilters={$dashboardStore.dimensionThresholdFilters}
               timeRange={undefined}
               comparisonTimeRange={undefined}
+              resolvedTimeStart={$timeControlStore.timeStart}
+              resolvedTimeEnd={$timeControlStore.timeEnd}
```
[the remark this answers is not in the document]
AdityaHegde: Renaming to `queryTime{Start,End}` and `displayTimeRange` to make it self-explanatory.

web-common/src/components/chip/removable-list-chip/RemovableListBody.svelte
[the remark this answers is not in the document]
AdityaHegde: Hmmm it is only used for dimension filter. It is probably simpler to just move it there.

web-common/src/features/dashboards/state-managers/selectors/dimension-filters.ts
AdityaHegde: No, this will break leaderboards when there is a complex filter present or no filter is present. Added a comment.

web-local/tests/explores/explores.spec.ts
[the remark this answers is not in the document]
AdityaHegde: Hmm the `publisher` come from the `name` of the dimension. Lets do this standardisation as our testing improvement effort.

web-common/src/features/dashboards/leaderboard/Leaderboard.svelte
```diff
@@ -53,7 +54,7 @@
   export let dimension: MetricsViewSpecDimensionV2;
   export let timeRange: V1TimeRange;
   export let comparisonTimeRange: V1TimeRange | undefined;
-  export let selectedValues: string[];
+  export let selectedValues: CompoundQueryResult<string[]>;
```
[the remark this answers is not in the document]
AdityaHegde: Hmmm this is not so trivial refactor. Since the Leaderboard component is called within a loop we cant just use `$` to get the values. Variables defined in a const statement cannot be used with a `$` either. Lets do this in a follow up.

ericpgreen2 (approved):
Nice work. I love the new "Contains" mode!

Given this is a huge PR, IMO cherry-picking it onto the release branch to be released in ~36 hours is a bit aggressive. Unless any of @mindspank, @royendo, @jkhwu, @ericokuma have time to sufficiently kick the tires.

PR #6774, opened 2025-02-26
feat: expose table size in information_schema

k-anshul:

**Checklist:**
- [x] Covered by tests
- [ ] I'm proud of this work!


runtime/drivers/clickhouse/information_schema.go
```diff
@@ -113,6 +113,74 @@ func (i informationSchema) Lookup(ctx context.Context, db, schema, name string)
...
+	}
+
+	queryBuilder.WriteString(strings.Join(placeholders, ", "))
+	queryBuilder.WriteString(") GROUP BY database, table")
+
+	rows, err := conn.QueryxContext(ctx, queryBuilder.String(), args...)
+	if err != nil {
+		return err
+	}
+	defer rows.Close()
```
begelundmuller, line 150: The caller may not have permission to access `system.parts`. To be consistent with the docstring, shouldn't it check for a permission error here and set `bytes_on_disk` to `-1` in that case?

runtime/drivers/olap.go
```diff
@@ -162,6 +162,9 @@ func (r *Result) Close() error {
 type InformationSchema interface {
 	All(ctx context.Context, like string) ([]*Table, error)
 	Lookup(ctx context.Context, db, schema, name string) (*Table, error)
+	// SizeOnDisk populates the BytesOnDisk field of the tables.
+	// It should be called after All or Lookup and not on manually created tables.
+	SizeOnDisk(ctx context.Context, tables []*Table) error
```
begelundmuller, line 167: It's a bit confusing to call it "size on disk" in some cases and "bytes on disk" in other cases. Their connected relationship is not immediately clear from the names, especially since the prefixes are different.

1. Maybe call it `LoadDiskSize` and the table property `DiskSizeBytes` (aligning with our practice in other places of putting the unit last in the name)?
2. Have you considered using "physical" instead of "disk"? Since in theory it doesn't have to be on a disk. For example, this is how BigQuery names these:

Screenshot 2025-03-03 at 14 07 16

runtime/drivers/pinot/olap.go
```diff
@@ -250,11 +249,71 @@ func (i informationSchema) Lookup(ctx context.Context, db, schema, name string)
 		View:            false,
 		Schema:          &runtimev1.StructType{Fields: schemaFields},
 		UnsupportedCols: unsupportedCols,
+		BytesOnDisk:     -1,
 	}
 
 	return table, nil
 }
 
+func (i informationSchema) SizeOnDisk(ctx context.Context, tables []*drivers.Table) error {
```
begelundmuller, line 258: Did you test this? And does it handle missing permissions correctly? Otherwise, IMO it's fine to put `-1` for Pinot since it's so rarely used right now.

begelundmuller (changes requested):

runtime/drivers/clickhouse/information_schema.go
```diff
@@ -113,6 +113,74 @@ func (i informationSchema) Lookup(ctx context.Context, db, schema, name string)
 	return tables[0], nil
 }
 
+func (i informationSchema) SizeOnDisk(ctx context.Context, tables []*drivers.Table) error {
```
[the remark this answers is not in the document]
k-anshul: Sure. I can name it `PopulateSizeOnDisk`.

runtime/drivers/clickhouse/information_schema.go
k-anshul: I expected the caller to take a call on whether to handle error or ignore. I imagined that we could ignore the error in OLAP APIs for example but that may not be correct to do in API exposed as part of `InformationSchema`. 
It returns -1 already if it is a table.

runtime/drivers/olap.go
k-anshul: I did not consider `physical`. It can also be used.

runtime/drivers/pinot/olap.go
k-anshul: I did not test this. I just looked around their API docs and manually checked API requests and responses via swagger. The docs aren't clear and IIRC doesn't mention anything about permissions.
We can remove it if you prefer that.

runtime/drivers/clickhouse/information_schema.go
begelundmuller: Since we may have several places that try to get table sizes (across UI, CLI, reconcilers) and checking the error message with string matching is cumbersome, I think returning `-1` for "unknown" is a reasonable thing to do (and it's already described in the docstring). It also propagates nicely, e.g. even if no handling for `-1` is done, it will often be self-explanatory for the end user to see that number.

runtime/server/connector_service.go
```diff
@@ -141,10 +143,16 @@ func (s *Server) OLAPListTables(ctx context.Context, req *runtimev1.OLAPListTabl
...
+	i := olap.InformationSchema()
+	tables, err := i.All(ctx, req.SearchPattern)
 	if err != nil {
 		return nil, err
 	}
+	err = i.LoadPhysicalSize(ctx, tables)
+	if err != nil {
+		// the size queries can fail due to permission erros so we log them and continue
+		s.logger.Warn("failed to populate table sizes", zap.Error(err))
+	}
```
begelundmuller, line 155: I'm a bit worried we'll see a lot of these, especially if the cloud UI starts fetching table sizes. I'd be up for just discarding the error (or turning it into an internal debug log), and just letting it return `-1`.

Also, if you intend to keep a log, should we add `observability.ZapCtx(ctx)`?

runtime/server/connector_service.go
```diff
@@ -173,11 +182,17 @@ func (s *Server) OLAPGetTable(ctx context.Context, req *runtimev1.OLAPGetTableRe
 	if err != nil {
 		return nil, err
 	}
+	err = olap.InformationSchema().LoadPhysicalSize(ctx, []*drivers.Table{table})
+	if err != nil {
+		// the size queries can fail due to permission erros so we log them and continue
+		s.logger.Warn("failed to populate table sizes", zap.Error(err))
+	}
```
begelundmuller, line 189: Same comment as above

begelundmuller (changes requested):

runtime/drivers/pinot/olap.go
begelundmuller: Up to you if you want to remove or keep

k-anshul: I can keep this for now but added a comment that this is untested code.

begelundmuller (approved):

PR #6782, opened 2025-02-27
feat: Set datadog trace id in response headers to correlate with logs 

k-anshul:

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/runtime/pkg/observability/middleware.go
+++ b/runtime/pkg/observability/middleware.go
@@ -35,6 +35,7 @@ func TracingMiddleware(next http.Handler, serviceName string) http.Handler {
 
 // LoggingUnaryServerInterceptor is a gRPC unary interceptor that logs requests.
 // It also recovers from panics and returns them as internal errors.
+// Also sends the trace ID in response headers to correlate with logs.
 func LoggingUnaryServerInterceptor(logger *zap.Logger) grpc.UnaryServerInterceptor {
 	logger = logger.WithOptions(zap.AddStacktrace(zapcore.InvalidLevel)) // Disable stacktraces for error logs in this interceptor
 	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (resp interface{}, err error) {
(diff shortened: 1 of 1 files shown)


runtime/pkg/observability/middleware.go
```diff
@@ -55,6 +56,14 @@ func LoggingUnaryServerInterceptor(logger *zap.Logger) grpc.UnaryServerIntercept
 		}
 
 		start := time.Now()
+
+		// Add datadog trace ID to response headers
+		traceID := DatadogTraceID(ctx)
+		if traceID != "" {
+			header := metadata.Pairs("dd-trace-id", traceID)
+			_ = grpc.SetHeader(ctx, header)
+		}
```
begelundmuller, line 65: Firstly, have you checked if there is a standard HTTP/gRPC header we can use for request/trace IDs? For example, I know `X-Request-Id` is a common header, but not sure if otel or gRPC has a different best practice.

Secondly, it would be nice to lean into native open telemetry handling if and when possible. Some ideas:
1. Can we configure DataDog to accept otel trace IDs?
2. Or, can we adapt otel to generate DataDog-friendly trace IDs?
3. Or, can we emit both IDs in one header for compatibility until (hopefully) DataDog someday supports native otel traces, e.g. `X-Request-Id: <otel ID>; <datadog ID>`?

begelundmuller (changes requested):

runtime/pkg/observability/middleware.go
k-anshul: Okay looks like [datadog](http://docs.datadoghq.com/tracing/guide/span_and_trace_id_format/) has started to accept 128bit unsigned IDs but the field names in logs should still be `dd.trace_id` and `dd.span_id`. We may need to upgrade our collectors and test this out.

My understanding(might be wrong) on `X-Request-Id` is that it is more used for auditing and not for distributed tracing so I would prefer to keep it `X-Trace-Id`. Another thing to note is that since it is custom metadata header grpc automatically adds `grpc-metadata` prefix and makes it `Grpc-Metadata-X-Trace-Id`.

k-anshul: It is something `Grpc-Metadata` adds to automatically pass http request headers to/from gRPC metadata.
We can disable this by adding a custom matching logic. 
My understanding is that in this case the client will also need to have the same logic if they want it added in grpc metadata automatically like accessing in rill cli.

Given our use case is mostly for UI we can add this. If this is required in CLI in future then we will need to add similar logic in CLI as well.

begelundmuller (approved):
LGTM

Issue #6607, opened 2025-02-06
Create Jobs to clean Dangling Resources


grahamplata:
Relates to: https://github.com/rilldata/rill/issues/6474 Support user deletion #6474

If the user is the last user in the organization and is deleted, hibernate the projects.

This would enable support to triage accidental deletions `rill sudo org join`

Create Two Jobs
1. To log orphaned orgs and connected projects
2. 30 Days of inactivity - drop org and projects - Manual execution first

PR #6785, opened 2025-02-27
Fix: Jobs to clean Dangling Resources

grahamplata:
This PR creates two jobs to assist with dangling resources. https://github.com/rilldata/rill/issues/6607

- **Job 1** Logs inactive organizations and hibernates projects. The criteria for the current logic is that an organization must  have no users.

- **Job 2** Enables inactive organizations to be deleted. This is only manual triggered at this time.

> Additionally adds pgweb to devtool containers

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!

### DIFF
--- a/admin/database/postgres/postgres.go
+++ b/admin/database/postgres/postgres.go
@@ -234,7 +234,9 @@ func (c *connection) FindInactiveOrganizations(ctx context.Context) ([]*database
 	// TODO: This definition may change, but for now, we are considering an organization as inactive if it has no users
 	res := []*database.Organization{}
 	err := c.getDB(ctx).SelectContext(ctx, &res, `
-		SELECT o.* FROM orgs o WHERE NOT EXISTS ( SELECT 1 FROM users_orgs_roles uor WHERE uor.org_id = o.id )
+		SELECT o.* FROM orgs o
+		WHERE now() - o.updated_on > INTERVAL '1 DAY'
+		AND NOT EXISTS ( SELECT 1 FROM users_orgs_roles uor WHERE uor.org_id = o.id )
 	`)
 	if err != nil {
 		return nil, parseErr("orgs", err)
--- a/admin/database/postgres/postgres.go
+++ b/admin/database/postgres/postgres.go
@@ -230,6 +230,21 @@ func (c *connection) DeleteOrganizationWhitelistedDomain(ctx context.Context, id
 	return checkDeleteRow("org whitelist domain", res, err)
 }
 
+func (c *connection) FindInactiveOrganizations(ctx context.Context) ([]*database.Organization, error) {
+	// TODO: This definition may change, but for now, we are considering an organization as inactive if it has no users
+	res := []*database.Organization{}
+	err := c.getDB(ctx).SelectContext(ctx, &res, `
+		SELECT o.* FROM orgs o
+		WHERE o.updated_on < now() - interval '30 days' AND NOT EXISTS (
+			SELECT 1 FROM users_orgs_roles uor WHERE uor.org_id = o.id
+		)
+	`)
+	if err != nil {
+		return nil, parseErr("orgs", err)
+	}
+	return res, nil
+}
+
 func (c *connection) FindProjects(ctx context.Context, afterName string, limit int) ([]*database.Project, error) {
 	var res []*projectDTO
 	err := c.getDB(ctx).SelectContext(ctx, &res, "SELECT p.* FROM projects p WHERE lower(name) > lower($1) ORDER BY lower(p.name) LIMIT $2", afterName, limit)
--- a/admin/database/postgres/postgres.go
+++ b/admin/database/postgres/postgres.go
@@ -234,10 +234,7 @@ func (c *connection) FindInactiveOrganizations(ctx context.Context) ([]*database
 	// TODO: This definition may change, but for now, we are considering an organization as inactive if it has no users
 	res := []*database.Organization{}
 	err := c.getDB(ctx).SelectContext(ctx, &res, `
-		SELECT o.* FROM orgs o
-		WHERE o.updated_on < now() - interval '30 days' AND NOT EXISTS (
-			SELECT 1 FROM users_orgs_roles uor WHERE uor.org_id = o.id
-		)
+		SELECT o.* FROM orgs o WHERE NOT EXISTS ( SELECT 1 FROM users_orgs_roles uor WHERE uor.org_id = o.id )
 	`)
 	if err != nil {
 		return nil, parseErr("orgs", err)
--- a/admin/jobs/river/org_jobs.go
+++ b/admin/jobs/river/org_jobs.go
@@ -163,67 +163,6 @@ func (w *DeleteOrgWorker) Work(ctx context.Context, job *river.Job[DeleteOrgArgs
 		}
 	}
 
-	// delete org, billing issues will be cascade deleted
-	err = w.admin.DB.DeleteOrganization(ctx, org.Name)
-	if err != nil {
-		return err
-	}
-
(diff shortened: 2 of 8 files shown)


admin/jobs/river/org_jobs.go
```diff
@@ -173,3 +173,77 @@ func (w *DeleteOrgWorker) Work(ctx context.Context, job *river.Job[DeleteOrgArgs
...
+	logger *zap.Logger
+}
+
+func (w *LogInactiveOrgsWorker) Work(ctx context.Context, job *river.Job[LogInactiveOrgsArgs]) error {
+	orgs, err := w.admin.DB.FindInactiveOrganizations(ctx)
+	if err != nil {
+		if errors.Is(err, database.ErrNotFound) {
+			return nil
+		}
+		return fmt.Errorf("failed to find inactive organizations: %w", err)
```
begelundmuller, line 193: No need for `database.ErrNotFound` because it only applies for DB calls that expect exactly one result (e.g. `FindOrganization`, but not `FindOrganizations`).

```suggestion
	if err != nil {
		return fmt.Errorf("failed to find inactive organizations: %w", err)
```

admin/jobs/river/org_jobs.go
```diff
@@ -173,3 +173,77 @@ func (w *DeleteOrgWorker) Work(ctx context.Context, job *river.Job[DeleteOrgArgs
...
+		if err != nil {
+			return fmt.Errorf("failed to find projects for organization %s: %w", org.Name, err)
+		}
+		for _, proj := range projects {
+			p, err := w.admin.HibernateProject(ctx, proj)
+			if err != nil {
+				return fmt.Errorf("failed to hibernate project %s: %w", proj.ID, err)
+			}
+			w.logger.Warn("hibernated project", zap.String("project_id", p.ID), zap.String("project_name", p.Name), zap.String("org_id", org.ID), zap.String("org_name", org.Name))
+		}
```
begelundmuller, line 207: On each run of this job, it will list and hibernate all projects in the inactive org. 

To prevent that, either:
1. Create a `DB.FindActiveProjectsForOrganization` that only returns projects that have one or more deployments.
2. Or perhaps just skip hibernation if `proj.ProdDeploymentID == nil`

admin/database/postgres/postgres.go
```diff
@@ -230,6 +230,18 @@ func (c *connection) DeleteOrganizationWhitelistedDomain(ctx context.Context, id
 	return checkDeleteRow("org whitelist domain", res, err)
 }
 
+func (c *connection) FindInactiveOrganizations(ctx context.Context) ([]*database.Organization, error) {
+	// TODO: This definition may change, but for now, we are considering an organization as inactive if it has no users
+	res := []*database.Organization{}
+	err := c.getDB(ctx).SelectContext(ctx, &res, `
+		SELECT o.* FROM orgs o WHERE NOT EXISTS ( SELECT 1 FROM users_orgs_roles uor WHERE uor.org_id = o.id )
```
begelundmuller, line 237: Not perfect, but maybe throw in something like a `now() - updated_on > INTERVAL '1 DAY'` to prevent too eager cleanup if some future refactors cause orgs to not have a member for a short while

admin/jobs/river/org_jobs.go
```diff
@@ -173,3 +173,77 @@ func (w *DeleteOrgWorker) Work(ctx context.Context, job *river.Job[DeleteOrgArgs
...
+	return nil
+}
+
+type DeleteInactiveOrgsArgs struct {
+	OrgID string
+}
+
+func (DeleteInactiveOrgsArgs) Kind() string { return "delete_inactive_orgs" }
+
+type DeleteInactiveOrgsWorker struct {
```
begelundmuller, line 220: 1. Since this takes an `OrgID`, shouldn't the job have a singular name, i.e. `DeleteInactiveOrg`?
2. Strictly speaking this job works for any org, there is nothing in it that has "inactive" handling

In light of the above, shouldn't this logic just be merged into the `DeleteOrg` job? I.e. adapt that job to check if the org has any projects, and then delete those (seems like an oversight in that job anyway).

admin/jobs/river/river.go
```diff
@@ -388,6 +391,28 @@ func (c *Client) PlanChanged(ctx context.Context, billingCustomerID string) (*jo
...
+		return nil, err
+	}
+	return &jobs.InsertResult{
+		ID:        res.Job.ID,
+		Duplicate: res.UniqueSkippedAsDuplicate,
+	}, nil
+}
+
+func (c *Client) DeleteInactiveOrgs(ctx context.Context) (*jobs.InsertResult, error) {
+	res, err := c.riverClient.Insert(ctx, DeleteInactiveOrgsArgs{}, nil)
```
begelundmuller, line 406: `DeleteInactiveOrgsArgs` expects an `OrgID`, but that doesn't get set here

admin/jobs/river/org_jobs.go
```diff
@@ -173,3 +173,77 @@ func (w *DeleteOrgWorker) Work(ctx context.Context, job *river.Job[DeleteOrgArgs
...
+
+func (LogInactiveOrgsArgs) Kind() string { return "log_inactive_orgs" }
+
+type LogInactiveOrgsWorker struct {
+	river.WorkerDefaults[LogInactiveOrgsArgs]
+	admin  *admin.Service
+	logger *zap.Logger
+}
+
+func (w *LogInactiveOrgsWorker) Work(ctx context.Context, job *river.Job[LogInactiveOrgsArgs]) error {
```
begelundmuller, line 187: Since this job actually hibernates projects, I think a better name would be `HibernateInactiveOrgs`

begelundmuller (changes requested):

begelundmuller (changes requested):

admin/jobs/river/river.go
```diff
@@ -105,6 +106,7 @@ func New(ctx context.Context, dsn string, adm *admin.Service) (jobs.Client, erro
 		newPeriodicJob(&TrialEndCheckArgs{}, "10 1 * * *", true),                 // daily at 1:10am UTC
 		newPeriodicJob(&TrialGracePeriodCheckArgs{}, "15 1 * * *", true),         // daily at 1:15am UTC
 		newPeriodicJob(&SubscriptionCancellationCheckArgs{}, "20 1 * * *", true), // daily at 1:20am UTC
+		newPeriodicJob(&HibernateInactiveOrgsArgs{}, "0 1 * * 0", true),          // weekly at 1am UTC on Sunday
```
begelundmuller, line 109: Can you make this 7am UTC on Monday instead? We don't want things happening automatically in the weekend if we can avoid it

begelundmuller (approved):

admin/jobs/river/org_jobs.go
```diff
@@ -173,3 +189,41 @@ func (w *DeleteOrgWorker) Work(ctx context.Context, job *river.Job[DeleteOrgArgs
...
+				continue
+			}
+
+			p, err := w.admin.HibernateProject(ctx, proj)
+			if err != nil {
+				return fmt.Errorf("failed to hibernate project %s: %w", proj.ID, err)
+			}
+			w.logger.Warn("hibernated project", zap.String("project_id", p.ID), zap.String("project_name", p.Name), zap.String("org_id", org.ID), zap.String("org_name", org.Name))
+		}
+		w.logger.Warn("inactive organization", zap.String("org_id", org.ID), zap.String("org_name", org.Name), zap.Time("last_updated_at", org.UpdatedOn), zap.Int("connected_projects", len(projects)))
```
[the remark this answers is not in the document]
begelundmuller: Lol, don't know how I missed that. Seems good then!

PR #6786, opened 2025-02-27
fix: sort for line and area charts in canvas

djbarnwal:
https://www.notion.so/rilldata/Some-Graphs-are-very-wrong-1a6ba33c8f578099a304d622b94b79ec?d=1a7ba33c8f5780699829001c3edea011#1a6ba33c8f57803eabe3f5655d6e7a91

Ignore sort for temporal field encoding.

Sorting on line and area chart is a bit more complex. This PR does not guarantee proper sorting in area and line chart but makes sure that atleast the charts are built properly.


briangregoryholmes (approved):
Looks good!

PR #6787, opened 2025-02-27
fix: support space in measure dimension names for charts

djbarnwal:
https://www.notion.so/rilldata/Expression-parse-error-Roy-1a7ba33c8f57806e98a4fd434e77083a

Sanitize measure name used for identifying formatter. Likely a bug in Vega Lite where it does not support space in `formatType`


briangregoryholmes (approved):
Weird bug.

PR #6791, opened 2025-02-28
fix: monthly parsing of report cron

AdityaHegde:
If monthly part of a cron is `*` we are incorrectly parsing it as NaN. So switching to monthly in a report edit will make it NaN from then on.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-common/src/features/scheduled-reports/time-utils.ts
+++ b/web-common/src/features/scheduled-reports/time-utils.ts
@@ -119,7 +119,9 @@ export function getTimeOfDayFromCronExpression(cronExpr: string): string {
 
 export function getDayOfMonthFromCronExpression(cronExpr: string): number {
   const [, , dayOfMonth] = cronExpr.split(" ");
-  return Number(dayOfMonth);
+  const dayOfMonthAsNum = Number(dayOfMonth);
+  if (Number.isNaN(dayOfMonthAsNum)) return 1;
+  return dayOfMonthAsNum;
 }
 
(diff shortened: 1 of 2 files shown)


web-common/src/features/scheduled-reports/time-utils.ts
```diff
@@ -119,7 +119,9 @@ export function getTimeOfDayFromCronExpression(cronExpr: string): string {
 
 export function getDayOfMonthFromCronExpression(cronExpr: string): number {
   const [, , dayOfMonth] = cronExpr.split(" ");
-  return Number(dayOfMonth);
+  const dayOfMonthAsNum = Number(dayOfMonth);
+  if (Number.isNaN(dayOfMonthAsNum)) return 1;
```
ericpgreen2, line 123: If `dayOfMonth === "*"`, which would resolve to `NaN`, then it feels wrong to consider it the same as `1`?

AdityaHegde: It doesnt really get used if the report is not monthly. Also since the Day-Of-Month selector is disabled there is no way to change it either. So lets change it when we do support Day-Of-Month selection.

ericpgreen2: Ok, yeah, mind just adding a code comment to explain this?

ericpgreen2 (approved):

Issue #6559, opened 2025-01-31
explore access should be > than metrics layer


royendo:
**Describe the bug**
If metrics layer is:
```
security:
  access: true
  row_filter: > 
        customer_domain = '{{ .user.domain }}' or {{ .user.admin }} 
  exclude:
    - if: "'{{.user.domain}}' != 'rilldata.com'"
      names:
        - transaction_id
        - sales_rep_email
        - sales_rep_domain
        - sales_team
        - internal_flag
        - sensitive_flag

```

and explore is 
```
security:
  access: "{{ .user.admin }} OR '{{ .user.domain }}' == 'rilldata.com'"
```

I would expect that explore dashboard would take the higher level of permissions and remove the dashboard however, that is not the case. 

**To Reproduce**
1. Set up metrics and explore as above
2. Test with a user that is not admin or rilldata.com 
3. see that metrics view gets applied

**Expected behavior**
The explore should be restricted on the access policy for dashboard. 
Use case is that im using the same metrics in a demo project to demonstrate metrics level permissions and explore and found that even setting explore level access is ignored. Will create unique metrics and explore as a workaround

**Screenshots**
![Image](https://github.com/user-attachments/assets/ab96191f-b73d-445c-813f-7dede6021909)

begelundmuller:
This certainly sounds like a bug – the explore should be hidden for this user. @grahamplata can you try to replicate and fix?

PR #6793, opened 2025-02-28
fix: Derived explore security policies

grahamplata:
Issue: https://github.com/rilldata/rill/issues/6559

https://github.com/user-attachments/assets/cf52f293-f3ed-4604-9b78-5cef1a114cd8
> Expected flow works the same in local and cloud

```yaml
# Explore YAML
# Reference documentation: https://docs.rilldata.com/reference/project-files/explores

type: explore

title: "Programmatic Ads Auction"
metrics_view: auction_metrics

dimensions: '*'
measures: '*'

security:
  access: "{{ .user.admin }} OR '{{ .user.domain }}' == 'rilldata.com'"

time_zones:
  - America/Los_Angeles
  - America/Chicago
  - America/New_York
  - Europe/London
  - Europe/Paris

embeds:
  hide_pivot: true
defaults:
  measures:
    - requests
    - avg_bid_floor
    - 1d_qps
  dimensions:
    - app_site_name
    - app_site_cat
    - ad_size
    - device_state
  time_range: P7D

---

# Metrics view YAML

version: 1
model: "auction_data_model"
type: metrics_view

timeseries: "__time"
smallest_time_grain: "hour"

security:
  access: true
  row_filter: >
    device_state = 'NY'
  exclude:
    - if: "'{{.user.domain}}' != 'rilldata.com'"
      names:
        - app_site_name

measures:
  - display_name: Requests
    name: requests
    expression: sum(bid_request_cnt)
    description: Total Requests
    format_preset: humanize
  - display_name: "Avg Bid Floor"
    name: avg_bid_floor
    expression: "sum(bid_floor) / sum(has_bid_floor_cnt) "
    description: "Average Bid Floor"
    format_preset: currency_usd
  - display_name: "1D QPS"
    name: 1d_qps
    expression: "sum(bid_request_cnt) / 86400 "
    description: "1D QPS"
    format_preset: humanize


dimensions:
  - name: app_site_name
    display_name: App Site Name
    column: app_site_name
    description: ""
  - name: app_site_cat
    display_name: App Site Cat
    column: app_site_cat
    description: ""
  - name: ad_size
    display_name: Ad Size
    column: ad_size
    description: ""
  - name: device_state
    display_name: Device State
    column: device_state
    description: ""

```

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

c.c. @royendo

### DIFF
--- a/runtime/reconcilers/explore.go
+++ b/runtime/reconcilers/explore.go
@@ -135,8 +135,8 @@ func (r *ExploreReconciler) validateAndRewrite(ctx context.Context, self *runtim
 			}
 		}
 	} else {
-		securityRules := append(mv.SecurityRules, spec.SecurityRules...)
-		access := mergeConditionRules(securityRules)
+		mv.SecurityRules = append(mv.SecurityRules, spec.SecurityRules...)
+		access := mergeConditionRules(mv.SecurityRules)
 		if access != nil {
 			spec.SecurityRules = []*runtimev1.SecurityRule{access}
 		}
--- a/runtime/reconcilers/explore.go
+++ b/runtime/reconcilers/explore.go
@@ -127,10 +127,54 @@ func (r *ExploreReconciler) validateAndRewrite(ctx context.Context, self *runtim
 		return nil, fmt.Errorf("parent metrics view %q is invalid", spec.MetricsView)
 	}
 
-	// Add the access and field access security rules from the parent metrics view.
-	for _, rule := range mv.SecurityRules {
-		if rule.GetAccess() != nil || rule.GetFieldAccess() != nil {
-			spec.SecurityRules = append(spec.SecurityRules, rule)
+	if len(spec.SecurityRules) == 0 {
+		for _, rule := range mv.SecurityRules {
+			if rule.GetAccess() != nil || rule.GetFieldAccess() != nil {
+				spec.SecurityRules = append(spec.SecurityRules, rule)
+			}
+		}
+	} else {
+		for _, rule := range mv.SecurityRules {
+			if rule.GetFieldAccess() != nil {
+				spec.SecurityRules = append(spec.SecurityRules, rule)
+			}
+		}
+
+		var exploreAccess *runtimev1.SecurityRule
+		var metricsViewAccess *runtimev1.SecurityRule
+		exploreAccessIndex := -1
+
+		for i, rule := range spec.SecurityRules {
+			if rule.GetAccess() != nil {
+				exploreAccess = rule
+				exploreAccessIndex = i
+				break
+			}
+		}
+
+		// Find metrics view's access rule
+		for _, rule := range mv.SecurityRules {
+			if rule.GetAccess() != nil {
+				metricsViewAccess = rule
+				break
+			}
+		}
+
+		if exploreAccess != nil && metricsViewAccess != nil {
+			condition := fmt.Sprintf("(%s) AND (%s)", metricsViewAccess.GetAccess().Condition, exploreAccess.GetAccess().Condition)
+
+			if exploreAccessIndex >= 0 {
+				spec.SecurityRules[exploreAccessIndex] = &runtimev1.SecurityRule{
+					Rule: &runtimev1.SecurityRule_Access{
+						Access: &runtimev1.SecurityRuleAccess{
+							Condition: condition,
+							Allow:     true,
+						},
+					},
+				}
+			}
+		} else if metricsViewAccess != nil && exploreAccess == nil {
+			spec.SecurityRules = append(spec.SecurityRules, metricsViewAccess)
 		}
 	}
 
--- a/runtime/reconcilers/explore.go
+++ b/runtime/reconcilers/explore.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 
 	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
 	"github.com/rilldata/rill/runtime"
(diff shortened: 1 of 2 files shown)


begelundmuller (changes requested):

runtime/reconcilers/explore.go
```diff
@@ -127,10 +127,11 @@ func (r *ExploreReconciler) validateAndRewrite(ctx context.Context, self *runtim
 		return nil, fmt.Errorf("parent metrics view %q is invalid", spec.MetricsView)
 	}
 
-	// Add the access and field access security rules from the parent metrics view.
-	for _, rule := range mv.SecurityRules {
-		if rule.GetAccess() != nil || rule.GetFieldAccess() != nil {
-			spec.SecurityRules = append(spec.SecurityRules, rule)
+	if len(spec.SecurityRules) == 0 {
```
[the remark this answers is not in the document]
royendo: my basic use case / expectation would be (with canvas emerging, also applies to APIs, etc.) is that the row access policies should be in some ways indepedent of each other. 

 in the example in [6559](https://github.com/rilldata/rill/issues/6559), youll see that a gmail user is able to access the dashboard even though in the explore level i have set (below) and it seems to be using the metrics level sec policy, ignoring the explore level one. 
 ``` 
  access: "{{ .user.admin }} OR '{{ .user.domain }}' == 'rilldata.com'"
```
project access:
Screenshot 2025-03-06 at 9 10 43


i can forsee examples where we might want to hide full explore dashboards from specific users but still be able to access the metrics-view to display data required on the canvas dashboard (can point to multiple metrics-views).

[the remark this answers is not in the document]
Comment (begelundmuller) [reply runtime/reconcilers/explore.go:130]: > is that the row access policies should be in some ways indepedent of each other.

@royendo Yeah agree and this is already how it works. I.e. direct metrics queries are only subject to the metrics view's security policy, not the security policies of any dashboards based on it. 

> we might want to hide full explore dashboards from specific users but still be able to access the metrics-view

This is the question – independent or DAG aware? Canvases are different, but for explores, which are always derived from one metrics view, I think there's an expectation that you can only access the explore if you have access to the underlying metrics view. I.e. the expect access logic is something like (pseudocode):
```
if eval(explore.security.access) && eval(explore.metricsView.security.access) {
  // allow
} else {
  // deny
}
```

Do you agree? (It _almost_ does this now, but not quite – the current implementation ends up behaving like there's an `||`, not an `&&`, in the `if` expression, which is not useful.)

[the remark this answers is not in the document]
Comment (royendo) [reply runtime/reconcilers/explore.go:130]: @begelundmuller hm does it? From my testing (maybe just in preview mock_users) but I can see an explore dashboard that is set to access:false if the metrics-view gives me access. see vid

https://www.loom.com/share/eb03418657594d88aeae7abe1267da0e

given the logic of the code, I would expect that jane@hotmail.com would also see a 403 when trying to visit the page. 

--- 

>  think there's an expectation that you can only access the explore if you have access to the underlying metrics view. I.e. the expected access logic is something like (pseudocode):

I agree on this, there's should never be a case where we allow access on the explore and block on the metrics, doesn't make sense.

File: runtime/reconcilers/explore.go [hunk 2]
```diff
@@ -127,10 +127,54 @@ func (r *ExploreReconciler) validateAndRewrite(ctx context.Context, self *runtim
...
```
begelundmuller, line 130: Explicitly handle if `len(mv.SecurityRules) == 0` but `len(spec.SecurityRules) != 0`

runtime/reconcilers/explore.go
```diff
@@ -127,10 +127,54 @@ func (r *ExploreReconciler) validateAndRewrite(ctx context.Context, self *runtim
...
+		for _, rule := range mv.SecurityRules {
+			if rule.GetAccess() != nil || rule.GetFieldAccess() != nil {
+				spec.SecurityRules = append(spec.SecurityRules, rule)
+			}
+		}
+	} else {
+		for _, rule := range mv.SecurityRules {
+			if rule.GetFieldAccess() != nil {
+				spec.SecurityRules = append(spec.SecurityRules, rule)
+			}
```
begelundmuller, line 140: We either need to:
1. Handle merge of `FieldAccess` rules
2. Or, return an error if someone uses a `FieldAccess` rule on an explore (can return an error from the reconciler here if we find one)
    - So basically we only allow a basic `Access` rule on explores

runtime/reconcilers/explore.go
```diff
@@ -201,3 +215,48 @@ func (r *ExploreReconciler) resolveFields(selected []string, selector *runtimev1
 	}
 	return res, nil
 }
+
+// Merges a list of security condition rules into a single rule.
+func mergeConditionRules(rules []*runtimev1.SecurityRule) *runtimev1.SecurityRule {
```
begelundmuller, line 220: It only merges `Access` rules, not `FieldAccess` or `RowFilter` or any other – that should probably be made clearer in the docstring and/or function name here

runtime/reconcilers/explore.go
```diff
@@ -127,10 +128,23 @@ func (r *ExploreReconciler) validateAndRewrite(ctx context.Context, self *runtim
...
+		mv.SecurityRules = append(mv.SecurityRules, spec.SecurityRules...)
+		access := mergeConditionRules(mv.SecurityRules)
+		if access != nil {
+			spec.SecurityRules = []*runtimev1.SecurityRule{access}
+		}
+
+		for _, rule := range mv.SecurityRules {
+			if rule.GetFieldAccess() != nil {
+				spec.SecurityRules = append(spec.SecurityRules, rule)
+			}
```
begelundmuller, line 147: This is a bit dangerous because it will fail silently if a `RowFilter` or `FieldAccess` (or some fourth type of security rule we might define in the future) is used in the explore spec's security policy.

Consider explicitly looping over the explore spec's security rules and returning an error if one of them is not an `Access` rule

begelundmuller (approved):

PR #6797, opened 2025-03-03
Add concurrent batch processing to `validate_deployments` job

kaspersjo:


admin/jobs/river/validate_deployments.go
```diff
@@ -27,8 +28,16 @@ func (w *ValidateDeploymentsWorker) Work(ctx context.Context, job *river.Job[Val
...
 
 func (w *ValidateDeploymentsWorker) validateDeployments(ctx context.Context) error {
-	// Iterate over batches of projects
+	// Resolve batch size from config
 	limit := 100
+	if w.admin.ProvisionerMaxConcurrency <= 100 {
+		limit = w.admin.ProvisionerMaxConcurrency
+	} else {
+		w.admin.Logger.Warn("validate deployments: provisioner max concurrency set too high, using maximum concurrency of 100", zap.Int("provisioner_max_concurrency", w.admin.ProvisionerMaxConcurrency), observability.ZapCtx(ctx))
+	}
```
begelundmuller, line 37: Is a hard cap necessary or can it just be:
```suggestion
	limit := 100
	if w.admin.ProvisionerMaxConcurrency > 0 {
		limit = w.admin.ProvisionerMaxConcurrency
	}
```

admin/jobs/river/validate_deployments.go
```diff
@@ -44,14 +53,20 @@ func (w *ValidateDeploymentsWorker) validateDeployments(ctx context.Context) err
 			afterName = projs[len(projs)-1].Name
 		}
 
-		// Process batch
+		// Process batch concurrently
```
begelundmuller, line 56: Not sure if this really matters, but a "problem" here is that if one or a few upgrades are slow, it holds up the entire batch. So potentially a lot of the time may be spent waiting for just a few slow upgrades.

If you want to ensure that it consistently runs `n` concurrent upgrades, you can consider
1. Using `sync.WaitGroup` to start `n` concurrent workers that consume projects to upgrade from a shared channel. When the channel is closed, it can return.
2. Having a separate goroutine that pages through `FindProjects`, sending each project to the worker's shared channel. Since channels are blocking, it will automatically never prefetch more projects than the page size.

admin/jobs/river/validate_deployments.go
```diff
@@ -44,16 +64,14 @@ func (w *ValidateDeploymentsWorker) validateDeployments(ctx context.Context) err
 			afterName = projs[len(projs)-1].Name
 		}
 
-		// Process batch
 		for _, proj := range projs {
```
k-anshul, line 67: Nit: Not required to change now but in future for these use cases also consider using `errgroup.Group`. It already has abstractions for limiting concurrent workers.
The code then could be simpler to understand:
```
g, ctx := errgroup.WithContext(it.ctx)
g.SetLimit(w.admin.ProvisionerMaxConcurrency)

for _, proj := range projs {
    g.Go(func() error {
        err := w.validateDeploymentsForProject(ctx, proj)
        ...   
     }
}

g.Wait()
```

k-anshul (approved):

admin/jobs/river/validate_deployments.go
kaspersjo: Updated to a shared channel and a worker pool.

admin/jobs/river/validate_deployments.go
```diff
@@ -27,7 +28,26 @@ func (w *ValidateDeploymentsWorker) Work(ctx context.Context, job *river.Job[Val
 const validateDeploymentsForProjectTimeout = 5 * time.Minute
 
 func (w *ValidateDeploymentsWorker) validateDeployments(ctx context.Context) error {
-	// Iterate over batches of projects
+	var wg sync.WaitGroup
+	ch := make(chan *database.Project)
+
+	// Setup concurrent workers
+	for range w.admin.ProvisionerMaxConcurrency {
```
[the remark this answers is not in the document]
kaspersjo: Thanks, added a check.

admin/jobs/river/validate_deployments.go
kaspersjo: Nice suggestion, was not aware that the `errgroup` package had the limiting concurrent workers functionality.

PR #6798, opened 2025-03-03
Canvas: KPI and time store fixes

djbarnwal:
- Do not show percentage change value when invalid
- Set initial comparison range for component time controls
- Move merge properties such as `showTimeComparison` and time states to `timeAndFilterStore`


briangregoryholmes:
Might be good to include the adjustment Janet mentioned for the comparison value toggle in this PR.

PR #6802, opened 2025-03-03
docs: Clarify browser support

ericpgreen2:
Clarifies that we do not support Safari & Firefox at the moment.

### DIFF
--- a/docs/docs/home/FAQ.md
+++ b/docs/docs/home/FAQ.md
@@ -4,45 +4,24 @@ sidebar_label: FAQ
 sidebar_position: 20
 ---
 
-## Getting Started
+## Technical requirements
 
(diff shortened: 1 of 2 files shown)


ericokuma (commented):
Based on the proposed changes, are we sure we want to remove the `Getting Started` section of the FAQ section?

cc @royendo

ericpgreen2:
> Based on the proposed changes, are we sure we want to remove the Getting Started section of the FAQ section?

I figured that that information was better suited (and already exists) elsewhere in the docs.

royendo:
@ericokuma I think this is fine, once i finish up metrics gonna do a full docs look through!

ericokuma (approved):
Looks good!

PR #6789, opened 2025-02-28
feat (Reports): UI for token-backed reports

ericpgreen2:
Frontend work for https://github.com/rilldata/rill/pull/6216. Closes https://github.com/rilldata/rill/issues/6708.

**Checklist:**
- [ ] Covered by tests – Public URLs are covered, but Reports are pending https://github.com/rilldata/rill/issues/6809
- [x] Ran it and it works as intended
- [x] Reviewed the diff before requesting a review
- [x] Checked for unhandled edge cases
- [x] Linked the issues it closes
- [ ] Checked if the docs need to be updated
- [ ] Intend to cherry-pick into the release branch
- [x] I'm proud of this work!

PR #6808, opened 2025-03-04
e2e: Add tests for Public URLs

ericpgreen2:
Contributes to https://github.com/rilldata/rill-private-issues/issues/523

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!


web-admin/tests/public-urls.spec.ts
```diff
@@ -0,0 +1,107 @@
+import { expect } from "@playwright/test";
+import { test } from "./setup/base";
+
+test.describe("Public URLs", () => {
+  let publicUrl: string;
+
+  test("should be able to create a public URL", async ({ page }) => {
```
AdityaHegde, line 7: Should this instead be a beforeAll? Feels like an anti-pattern for other tests depend on this, but not enforced to be dependant.

AdityaHegde (commented):
We should add some verification that the url params are populated. But we should add a separate test for it and reuse the code to assert.

web-admin/tests/public-urls.spec.ts
ericpgreen2: True. I want the "create a public URL" step to be an official "test", so I've marked the test suite as `serial`

ericpgreen2:
> We should add some verification that the url params are populated.

Good call. And it turns out that Public URLs are _not_ retaining original dashboard state at the moment. However, I did fix the state issue here: https://github.com/rilldata/rill/pull/6789.

I'll revert the URL state test from this PR and I'll add it to https://github.com/rilldata/rill/pull/6789, so we can at least merge the tests in this PR ASAP.

web-admin/tests/public-urls.spec.ts
AdityaHegde: This still doesnt fully fix the issue. If someone adds a new test, they have to make sure to run this and the new test all the time.

I think we can still have a test that asserts the contents of `publicUrl` but creation can happen in `beforeAll`. It is cleaner than having to remember to run both tests manually.

AdityaHegde: Found an issue talking about the exact thing: https://github.com/microsoft/playwright/issues/10114
Leading to https://playwright.dev/docs/api/class-test#test-step

Worth checking it out.

ericpgreen2: In this case, the `test.step` approach doesn't really work because each test operates on one `page`. Here, we need one test for an `adminPage` and one test for an `anonPage`.

ericpgreen2: I don't think there's a perfect solution. I added a `test.skip` note to make it a bit more clear that the `adminPage` and `anonPage` tests depend on the `publicURL` variable.

AdityaHegde (approved):

Issue #6713, opened 2025-02-18
Leaderboards: Multiple Measures


ericpgreen2:
https://www.notion.so/rilldata/Multiple-Metrics-in-Leaderboards-19aba33c8f57802f98aeefe569a81b39

PR #6816, opened 2025-03-04
Feat: Multiple Metrics in Leaderboards

lovincyrus:
This pull request introduces support for multiple metrics in leaderboards, enabling users to adjust the `leaderboard_measure_count` to control the number of visible measures. This featured is gated by feature flag `leaderboardMeasureCount`

```
features:
  - leaderboardMeasureCount
```

Closes https://github.com/rilldata/rill/issues/6713

https://www.notion.so/rilldata/Multiple-Metrics-in-Leaderboards-19aba33c8f57802f98aeefe569a81b39

**Checklist:**
- [x] Covered by tests
- [ ] I'm proud of this work!


web-common/src/features/dashboards/url-state/convertURLToExplorePreset.ts
```diff
@@ -414,6 +437,24 @@ function fromExploreUrlParams(
...
+    );
+
+    if (!leaderboardMeasureCount) {
+      // Gracefully remove sort_by from the URL
+      preset.exploreSortBy = undefined;
+    }
+
+    // Error because sort_by is not a valid option without
+    // leaderboard_measure_count
+    errors.push(new Error("sort_by requires leaderboard_measure_count"));
```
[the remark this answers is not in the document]
lovincyrus: Setting a default `leaderboard_measure_count` of 1 wouldn't make `sort_by=measure_2` valid. Imagine that the visible measures are `[pp, measure_1, measure_2]`. Having a default leaderboard measure count 1 would make the sort by measure 2 invalid, which is why I'd rather just remove the `sort_by` and it will default to the first visible measures internally even though it's not shown.

To repro:
1. Go to http://localhost:3001/explore/Bids_Sample_Dash?grain=day&sort_by=measure_2&leaderboard_measure_count=4
2. Remove `leaderboard_measure_count=4`

I think it makes more sense to remove the `sort_by` when `leaderboard_measure_count` is undefined.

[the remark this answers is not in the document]
ericpgreen2: Here's a case where I don't think that strategy works:
1. On the OpenRTB Auctions dashboard...
    image
2. I manually navigate to this URL: `http://localhost:3000/epg/openrtb/explore/auction_explore?sort_by=requests`.
3. I get this error message, but I would not expect to see it
    image

We should have sensible defaults for any state that's not provided explicitly (whether through user actions, URL parameters, bookmarks, or Explore YAML defaults).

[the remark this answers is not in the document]
ericpgreen2: > Setting a default leaderboard_measure_count of 1 wouldn't make sort_by=measure_2 valid
I agree, and that's why we should show an error in this case

web-common/src/features/dashboards/leaderboard/Leaderboard.svelte
```diff
@@ -184,18 +200,26 @@
   $: ({ data: sortedData, isFetching } = $sortedQuery);
   $: ({ data: totalsData } = $totalsQuery);
 
-  $: leaderboardTotal = totalsData?.data?.[0]?.[activeMeasureName] as
-    | number
-    | null;
+  // $: leaderboardTotal = totalsData?.data?.[0]?.[activeMeasureName] as
+  //   | number
+  //   | null;
```
ericpgreen2, line 205: Leftover code?

ericpgreen2 (changes requested):
Code looks clean! 

UXQA:
- (Blocker) When I toggle a leaderboard row on/off, the leaderboard width jumps wider for a brief moment: [Jam](https://jam.dev/c/00e37feb-fd07-4571-a812-2ef1ae001196)
- This shouldn't be a blocker because we need to get this out, but, if quick, it'd be good to add the `leaderboardMeasureCount` back to the URL
- I'm a bit worried about performance as the `leaderboardMeasureCount` increases, given we don't have column virtualization on these leaderboards. Let's watch it, and, maybe in a follow-up, add a limit to the `leaderboardMeasureCount`.

jkhwu:
@ericpgreen2 I disagree that leaderboard column width should be a blocker (it's a visual annoyance rather than functionality). Can we merge existing work so that I can test this feature in 0.58 before it goes out, and address layout shifts in a separate PR (that may or may not make it into 0.58 given time remaining)? cc @lovincyrus

web-common/src/features/dashboards/leaderboard/DelayedLoadingRows.svelte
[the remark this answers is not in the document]
lovincyrus: To avoid scope creep, let's tackle this next time

jkhwu (approved):
Looks good to me

web-common/src/features/dashboards/url-state/convertURLToExplorePreset.ts
```diff
@@ -151,6 +151,29 @@ export function convertURLToExplorePreset(
     }
   }
 
+  if (searchParams.has(ExploreStateURLParams.SortType)) {
```
AdityaHegde, line 154: Why do we need to copy this piece of code? `fromExploreUrlParams` should be enough no?

proto/gen/rill/runtime/v1/runtime.swagger.yaml
```diff
@@ -4340,6 +4340,8 @@ definitions:
         $ref: '#/definitions/v1ExploreSortType'
       exploreExpandedDimension:
         type: string
+      exploreLeaderboardMeasureCount:
```
AdityaHegde, line 4343: Was this changed by hand? I dont see the field in the proto.

Could you update `resources.proto` and add an entry there? Run `make proto.generate` after to generate these files.

AdityaHegde (commented):
Adding my UXQA around url params.
1. Refreshing the page clears the measure count along with the param in the url.
2. Pressing back button doesnt reset the measure count.
3. This field is not saved in bookmarks. We still use the legacy proto for this. You need to update `dashboard.proto`
4. Switching between pivot and explore tabs clears the param. (You will need to interact with pivot to trigger this). Add an entry here: https://github.com/rilldata/rill/blob/main/web-common/src/features/dashboards/url-state/explore-web-view-store.ts#L22 

Add a test to `url-state-variations.spec.ts`. Will catch 1-3 of the above issues. Sadly 4 doesnt have a test framework as of now.

proto/gen/rill/runtime/v1/runtime.swagger.yaml
lovincyrus: This was changed by `runtime.swagger.yaml` and I ran `npm run generate:runtime-client -w web-common`

lovincyrus: Added

lovincyrus:
> Adding my UXQA around url params.
> 
> 1. Refreshing the page clears the measure count along with the param in the url.
> 2. Pressing back button doesnt reset the measure count.
> 3. Measure count field is not saved in bookmarks. We still use the legacy proto for this. You need to update `dashboard.proto`
> 4. Switching between pivot and explore tabs clears the param. (You will need to interact with pivot to trigger this). Add an entry here: https://github.com/rilldata/rill/blob/main/web-common/src/features/dashboards/url-state/explore-web-view-store.ts#L22
> 
> Add a test to `url-state-variations.spec.ts`. Will catch 1-3 of the above issues. Sadly 4 doesnt have a test framework as of now.

1. Addressed (1), (2).
2. Added leaderboard_measure_count to `resources.proto` and `dashboard.proto`
3. Added a test to URL state variation.
4. (3) Aside from adding leaderboard count to `dashboard.proto`, do I need to do anything else here? @AdityaHegde

web-common/src/features/dashboards/url-state/url-state-variations.spec.ts
```diff
@@ -580,6 +586,7 @@ export function getCleanMetricsExploreForAssertion() {
   // TODO
   delete cleanedState.selectedScrubRange;
   delete cleanedState.leaderboardContextColumn;
+  delete cleanedState.leaderboardMeasureCount;
```
AdityaHegde, line 589: We should not be ignoring this. I see tests fail if this is removed.

`selectedScrubRange` is renamed but no tests are added for it yet, same for `leaderboardContextColumn`.

AdityaHegde (commented):
Back button is still not working all the time. From a fresh dashboard change the count and press back, doesnt reset the fields for me.

lovincyrus:
> Back button is still not working all the time. Start from a fresh dashboard change the count and press back, doesnt reset the fields for me. (Below changes should fix this)
> 
> You need a few more changes to get the tests working after commenting out the deletion of `leaderboardMeasureCount`.
> 
> 1. `getDashboardStateFromProto` in `fromProto.ts`
> 2. `getProtoFromDashboardState` in `toProto.ts`
> 3. `fromLegacyExploreFields` in `convertLegacyStateToExplorePreset.ts`
> 4. `getDefaultExplorePreset` in `getDefaultExplorePreset.ts` - This is for the back button to work when going back to blank url.

Wow, appreciate it! We need some documentation here. Added in https://github.com/rilldata/rill/pull/6816/commits/e015f029bd3b5a1cdb87406a1ae28b0f9ce71a24

ericpgreen2 (approved):
Nice implementation! I'll approve from my perspective, though I'd encourage you to wait for @AdityaHegde's sign-off on URL state.

AdityaHegde (approved):
Thanks for taking the effort. It is working well now!

Ya we need to 1st streamline the process of url params management. Docs before that probably would be redundant.

PR #6818, opened 2025-03-04
add structs to mv query cache key

pjain1:
Without the fix, test will use cached results from previous query and thus fail. Cause of this failure as well -https://github.com/rilldata/rill/actions/runs/13655204432/job/38178639449

### DIFF
--- a/runtime/resolvers/metrics.go
+++ b/runtime/resolvers/metrics.go
@@ -1,16 +1,13 @@
 package resolvers
 
 import (
-	"bytes"
 	"context"
-	"encoding/gob"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
 	"time"
 
-	"github.com/mitchellh/mapstructure"
 	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
 	"github.com/rilldata/rill/runtime"
 	"github.com/rilldata/rill/runtime/metricsview"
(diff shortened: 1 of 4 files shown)


runtime/resolvers/metrics.go
```diff
@@ -107,8 +109,44 @@ func (r *metricsResolver) CacheKey(ctx context.Context) ([]byte, bool, error) {
...
-	return bytes, true, err
+	if r.query.TimeRange != nil {
+		var buf bytes.Buffer
+		enc := gob.NewEncoder(&buf)
+		err = enc.Encode(r.query.TimeRange)
+		if err != nil {
+			return nil, false, err
+		}
+		queryMap["time_range"] = buf.Bytes()
+	}
```
begelundmuller, line 120: Why does the code on line 106 not suffice? I.e. this:
```
	err = mapstructure.Decode(r.query, &queryMap)
	if err != nil {
		return nil, false, err
	}
```

pjain1: clearly it does not work otherwise tests added here and https://github.com/rilldata/rill/actions/runs/13655204432/job/38178639449 would pass without this fix

pjain1: I think the issue might only be in decoding time types, we may add decoding config to handle time type or use gob encoder

pjain1: Relevant issue here - https://github.com/mitchellh/mapstructure/issues/270. We will need to use a decode hook, however since this is for cache key I directly encoded however if we want to use decode hook I can do that.

begelundmuller: Just pasting this from Slack for context for others:
> I think it might be better to add the decode hook for time. We anyway need to support creating a `metricsview.Query` from a map (for resolvers), so we anyway need a solid mapping from/to the `mapstructure` library.
> 
> Maybe consider adding a `.AsMap()` func on the `metricsview.Query` that configures the decode hook. Then we can also add a test for it in the `metricsview` package, so we ensure it keeps working.

begelundmuller (approved):

PR #6821, opened 2025-03-04
fix: canvas table stability for multiple table instances

djbarnwal:
For more than one table instance, the canvas table was unstable and behaved unexpectedly. This PR tries to address some of those concerns by unsubscribing to old table references, maintaining a component wise caching and softly failing when metrics view and measures are non-aligned.


briangregoryholmes (approved):
Looks good.

PR #6825, opened 2025-03-05
feat: URL params and bookmarks cloud E2E tests

AdityaHegde:
closes https://github.com/rilldata/rill-private-issues/issues/1375

Adds basic bookmark E2E test to create and use home, own and managed bookmarks. It also has a couple of fixes,
1. Home bookmark doesnt always apply.
2. Filter-only bookmark replaces current view.

Because of somewhat major issues fixed here. More advanced tests like editing will come as follow ups.

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!


web-common/tests/utils/assertUrlParams.ts
```diff
@@ -0,0 +1,7 @@
+import { expect, type Page } from "@playwright/test";
+
+export function assertUrlParams(page: Page, expectedParams: string) {
+  expect(new URL(page.url()).searchParams.toString()).toMatch(
+    new URL("http://localhost?" + expectedParams).searchParams.toString(),
```
ericpgreen2, line 5: I see `http://localhost?`, and it looks like a port is missing. But anyways, to test against `expectedParams`, what about just creating a `URLSearchParams` object rather than a full `URL` object?

web-common/tests/utils/assertUrlParams.ts
ericpgreen2, line 1: We have mostly kebab case, not camel case TS files

web-common/tests/utils/exploreInteractions.ts
ericpgreen2, line 1: This file too is camel case, not kebab case

web-admin/tests/bookmarks.spec.ts
```diff
@@ -0,0 +1,261 @@
+import { expect, type Page } from "@playwright/test";
+import { assertUrlParams } from "@rilldata/web-common/tests/utils/assertUrlParams";
+import { test } from "./setup/base";
+import { interactWithTimeRangeMenu } from "@rilldata/web-common/tests/utils/exploreInteractions";
+
+test.describe("Bookmarks", () => {
+  // TODO: use a separate explore to isolate bookmarks
```
ericpgreen2, line 7: Given you've edited the `explores.spec.ts` file, it looks like this comment is outdated?

web-admin/tests/bookmarks.spec.ts
```diff
@@ -0,0 +1,261 @@
...
+  test("Create and verify home bookmark", async ({ page }) => {
+    await page.goto("/e2e/openrtb/explore/auction_explore");
+
+    // Select "Last 14 days" as time range
+    await interactWithTimeRangeMenu(page, async () => {
```
ericpgreen2, line 12: I see you're just re-using an existing abstraction from `web-local`, but food for thought – Playwright seems to advocate for consolidating these sorts of abstractions into a [Page Object Model](https://playwright.dev/docs/pom).

web-admin/tests/bookmarks.spec.ts
```diff
@@ -0,0 +1,261 @@
+import { expect, type Page } from "@playwright/test";
+import { assertUrlParams } from "@rilldata/web-common/tests/utils/assertUrlParams";
+import { test } from "./setup/base";
+import { interactWithTimeRangeMenu } from "@rilldata/web-common/tests/utils/exploreInteractions";
+
+test.describe("Bookmarks", () => {
+  // TODO: use a separate explore to isolate bookmarks
+  test("Create and verify home bookmark", async ({ page }) => {
```
ericpgreen2, line 8: Nit: IMO it'd be easier to read the test cases if:
- "Home bookmark" was a test suite
- "Create home bookmark" was an independent test and the various verifications were discrete tests (run serially)

Similarly, "filters-only bookmark" and "standard bookmark" could be their own test suites.

ericpgreen2 (approved):

web-admin/tests/bookmarks.spec.ts
AdityaHegde: Good point. This what we had started with, I remember some devs had objections to this and we moved to a functional model. Lets do a greater refactor and move such methods in a separate PR.

web-admin/tests/bookmarks.spec.ts
AdityaHegde: I want to add `_bookmark` prefix to the dashboard so that it is clear what it is for. Temporarily I just moved the `explores.spec.ts` to use the other dashboard.

PR #6826, opened 2025-03-05
canvas: uxqa adjustments

briangregoryholmes:
- Adjusts KPI grid layout to accommodate more than four KPIs
- Addresses an issue where certain edits did not result in a reparsing of the file contents
- Adds try catch blocks around certain problematic edit points
- Fixes an issue with the `spreadEvenly` function
- Adjusts KPI rendering when the sparkling is positioned to the right
- Adjusts the minimum required KPI height
- Fixes an issue where images were being cut off when a title was present


djbarnwal (approved):
Looks good!

image

For the case when the sparkline is on the right. I would expect - 

1. The text on the left to be be vertically aligned in the middle
2. The sparkline should fill the rest of the whitespace instead it being 50% of the width. 
For the second point, I understand that it would be hard to estimate the max width of the text block as on hover, it expands a bit. I would suggest we let it overlap for now with the sparkline. 

Approving this PR, we can address the sparkline issue in a follow up

Issue #1366, opened 2022-12-05
Runtime: support metrics without a time dimension


begelundmuller:
To support metrics views without a time dimension:

- [x] Edit the reconcile logic to parse and migrate metrics views that do not have a time dimension – PR: #1459 (@AdityaHegde)
- [x] Audit/adapt the analytical APIs to support metrics views without a time dimension (just return an error if a metrics timeseries API is queried) (@egor-ryashin) (Handled by #1516)

Further, to support metrics views with no dimensions or only a time dimension:

- [x] Ensure the reconcile logic functions on such examples (add tests)
- [x] Ensure that non-dimension analytical APIs work for metrics views without dimensions (add tests)
- [x] Ensure that the big numbers API works for metrics views with neither dimensions or a time dimension

See [Notion](https://www.notion.so/New-Dashboard-View-0736643d9c51492f9b1f7517f58be152) for details.

PR #6855, opened 2025-03-10
Integration test for azure connector

NamanMahor:
Closed [#1366](https://github.com/rilldata/rill-private-issues/issues/1366)

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!


go.mod
```diff
@@ -362,6 +362,7 @@ require (
 	github.com/skeema/knownhosts v1.3.1 // indirect
 	github.com/skratchdot/open-golang v0.0.0-20200116055534-eef842397966 // indirect
 	github.com/spf13/cast v1.5.1 // indirect
+	github.com/testcontainers/testcontainers-go/modules/azurite v0.35.0 // indirect
```
k-anshul, line 365: I think you need to run `go mod tidy` on this branch.

k-anshul (changes requested):

go.mod
NamanMahor: done

runtime/testruntime/connectors.go
```diff
@@ -172,4 +174,76 @@ var Connectors = map[string]ConnectorAcquireFunc{
...
+		testdataPath := filepath.Join(currentFile, "..", "testdata")
+		azureInitData := filepath.Join(testdataPath, "init_data", "azure")
+		err = uploadDirectory(ctx, client, containerName, azureInitData)
+		require.NoError(t, err)
+
+		connectionString := fmt.Sprintf("DefaultEndpointsProtocol=http;AccountName=%s;AccountKey=%s;BlobEndpoint=%s;", azurite.AccountName, azurite.AccountKey, blobServiceURL)
+
+		ip, err := azuriteContainer.ContainerIP(context.Background())
+		require.NoError(t, err)
+		blobEndpointWithIP := fmt.Sprintf("http://%s:%d/%s", ip, 10000, azurite.AccountName)
```
k-anshul, line 210: Instead of a fixed port on host it should use `MappedPort` because 1000 may not be available all the time. See [mysql](https://github.com/rilldata/rill/blob/9c88e202fc7b3aedce2fccf3c9fab2fdca3acb47/runtime/testruntime/connectors.go#L184-L185) for reference.

NamanMahor: we are using mappedPort in azure_storage_connection_string which being used by duckdb model and source.
since clickhouse is in container it can't access host:mappedPort because it is mapped to host(localhost).

but to resolve this i m using  ip and direct port  as share the same network which is different from host. 
Same thing is happening in all the test for clickhouse.

k-anshul: Got it. Makes sense.

k-anshul (approved):

PR #6856, opened 2025-03-10
fix s3 integration test

NamanMahor:
fix s3 integration test

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!

### DIFF
--- a/runtime/resolvers/testdata/s3_connector.yaml
+++ b/runtime/resolvers/testdata/s3_connector.yaml
@@ -56,16 +56,17 @@ tests:
     properties:
       sql: "select * from all_datatypes_duckdb_source_csv"
     result_csv: |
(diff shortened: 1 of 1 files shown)


runtime/resolvers/testdata/s3_connector.yaml
```diff
@@ -265,26 +265,26 @@ tests:
   - name: query_all_result_duckdb_source_glob
     resolver: sql
     properties:
-      sql: "select * from all_datatypes_duckdb_source_glob"
+      sql: "select * from all_datatypes_duckdb_source_glob order by y,boolean_col"
```
k-anshul, line 268: If we ever add more rows in the data files then `y, boolen_col` can still lead to indeterministic behaviour. May be add a ID column to make it more deterministic.

k-anshul (changes requested):

runtime/resolvers/testdata/s3_connector.yaml
NamanMahor: done

PR #6859, opened 2025-03-10
Publish athena unload metrics 

k-anshul:


runtime/drivers/athena/warehouse.go
```diff
@@ -22,10 +22,15 @@ import (
...
 	"github.com/rilldata/rill/runtime/drivers"
 	rillblob "github.com/rilldata/rill/runtime/drivers/blob"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
 	"gocloud.dev/blob"
 	"gocloud.dev/blob/s3blob"
 )
 
+var tracer = otel.Tracer("github.com/rilldata/rill/runtime/drivers/unload")
```
begelundmuller, line 32: I think the tracer name is always supposed to match the package name
```suggestion
var tracer = otel.Tracer("github.com/rilldata/rill/runtime/drivers/athena")
```

runtime/drivers/athena/warehouse.go
```diff
@@ -151,6 +156,9 @@ func (c *Connection) awsConfig(ctx context.Context, awsRegion string) (aws.Confi
 }
 
 func (c *Connection) unload(ctx context.Context, client *athena.Client, conf *sourceProperties, unloadLocation string) error {
+	ctx, span := tracer.Start(ctx, "athenaUnload", trace.WithAttributes(attribute.String("sql", conf.SQL), attribute.String("location", unloadLocation)))
```
begelundmuller, line 159: It probably doesn't, but did you check if the AWS library supports otel tracing natively or with a lightweight third party library?

begelundmuller (approved):

runtime/drivers/athena/warehouse.go
k-anshul: Good point. Looks like it does. Let me reevaluate.

PR #6863, opened 2025-03-11
fix: default comparison when a non-standard time range is selected as default

AdityaHegde:
More details: https://www.notion.so/rilldata/Custom-default-time-range-returns-undefined-comparison-range-1b0ba33c8f5780c1aba1dedba800ecaf

When the selected default time range is non-standard then the comparison time range is not defined. It is ok to set it to previous period in this case.

**Checklist:**
- [x] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-common/src/features/dashboards/url-state/url-state-variations.spec.ts
+++ b/web-common/src/features/dashboards/url-state/url-state-variations.spec.ts
@@ -446,7 +446,6 @@ describe("Human readable URL state variations", () => {
         applyMutationsToDashboard(AD_BIDS_EXPLORE_NAME, mutations);
         const curState =
           getCleanMetricsExploreForAssertion() as MetricsExplorerEntity;
-        console.log(curState);
 
         const url = new URL("http://localhost");
         // load url with legacy protobuf state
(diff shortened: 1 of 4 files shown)


web-common/src/features/dashboards/url-state/url-state-variations.spec.ts
```diff
@@ -435,6 +446,7 @@ describe("Human readable URL state variations", () => {
         applyMutationsToDashboard(AD_BIDS_EXPLORE_NAME, mutations);
         const curState =
           getCleanMetricsExploreForAssertion() as MetricsExplorerEntity;
+        console.log(curState);
```
djbarnwal, line 449: Nit: Leftover log

web-common/src/features/dashboards/url-state/url-state-variations.spec.ts
```diff
@@ -447,6 +459,7 @@ describe("Human readable URL state variations", () => {
             explore,
             defaultExplorePreset,
           );
+        console.log(entityFromUrl);
```
djbarnwal, line 462: Nit: Leftover log

Issue #6579, opened 2025-02-04
Connectors UI: Provide an option to connect via DSN


ericpgreen2:
Start with:
- ClickHouse
- Druid
- Pinot

PR #6870, opened 2025-03-11
OLAP Connectors: Add DSN configuration option

ericpgreen2:
Closes https://github.com/rilldata/rill/issues/6579

image
image

**Checklist:**
- [x] Covered by tests
- [x] Checked if the docs need to be updated (FYI @royendo for docs updates, please!)
- [x] I'm proud of this work!


royendo:
definitely will need to look at revisiting
https://docs.rilldata.com/reference/olap-engines/clickhouse but that seems to be about it :)

k-anshul:
LGTM from an experience POV.

One minor thing that I noticed(exists on main not introduced in this PR) is that if clickhouse modeling is disabled then I can't add one more clickhouse connection from UI. The `data` option is hidden. I need to manually create another connection.yaml

ericpgreen2:
> One minor thing that I noticed(exists on main not introduced in this PR) is that if clickhouse modeling is disabled then I can't add one more clickhouse connection from UI. The `data` option is hidden. I need to manually create another connection.yaml

Thanks, Anshul. Addressed this comment here: https://github.com/rilldata/rill/pull/6904

web-local/tests/olap-connectors.spec.ts
```diff
@@ -0,0 +1,137 @@
+import { expect } from "@playwright/test";
+import { test } from "./setup/base";
+
+/**
+ * These tests focus on form validation and YAML generation for OLAP connectors.
+ * They don't test connections to live OLAP databases - that's handled by runtime integration tests.
```
AdityaHegde, line 6: Do we need E2E if we dont actually test the connection? We could have component tests instead?

web-common/src/features/sources/modal/AddDataForm.svelte
```diff
@@ -80,57 +98,103 @@
...
-          rpcError.message,
-        )}
-      />
-    {/if}
-
-    {#each properties as property (property.key)}
-      {#if property.key !== undefined && !property.noPrompt}
+  {/if}
+
+  {#if !useDsn}
```
AdityaHegde, line 155: Not sure if this is our standard, but it feels weird to have 2 significantly independent form in a single component. Would feel cleaner to me if they were isolated in separate components.

web-local/tests/olap-connectors.spec.ts
ericpgreen2: I added the connection test in this PR: https://github.com/rilldata/rill/pull/6892

web-common/src/features/sources/modal/AddDataForm.svelte
ericpgreen2: I agree it's not great, but I was hesitant to split them apart because:
1. We already have the "AddDataModal" wrapping the "AddDataForm". Adding another layer of abstraction would make it harder to follow.
2. The two forms share much in common, like the submission function, structure, and field types.
3. A change that applies to one of these forms likely applies to the other. So this follows "code that changes together lives together".

That being said, splitting them up still might be net-positive. Given there's changes coming to this form flow, I'd propose we keep as is for now and maybe split them up as the modal dialog evolves further.

AdityaHegde (approved):

Issue #6578, opened 2025-02-04
User management refactors


begelundmuller:
Introduce the guest role: 
- [x] Add `guest` boolean flags to org roles
- [x] Add a new org-level role `Guest` that is like `Viewer` but with `guest==true`
- [x] Grant the `Guest` org-level role to current and new project-level members (unless they already have another org-level role).
- [x] Update the membership APIs:
    - When adding a project member, also add them as an org-level guest.
    - When removing an org member, also remove them from all projects.
    - When adding someone under a domain whitelist

Add guest and non-guest groups:
- [x] Duplicate the current `all-users` group into separate `all-members` and `all-guests` groups.
- [x] Remove the special handling for the `all-users` group; replace it with a `managed` bool instead.

Other stand-alone changes:
- [x] Add an `Editor` org and project level role that can manage members and create resources, but can’t do things like manage env vars, billing, logos, or override security rules (e.g. mimic another user).
- [x] Add an org-level setting `default_project_role` that gives the role to grant the `all-members` group on new projects. It should default to `viewer` (current behavior) with support for specifying `none`.
- [x] Add an `autogroup:` prefix to system-managed group names
- [x] Update docs with the `editor` role and new autogroups
- [x] Change the `editor` role so they cannot add/remove `admin` roles
- [x] Add APIs for listing the groups and projects that a specific user belongs to
- [x] Change the member listing APIs to support filtering by role
- [x] Change the member listing APIs to return the count of groups and projects that each user belongs to
- [x] Change the group listing APIs to include the count of members
- [x] Unify project invites with org invites (so a project invite becomes an org guest invite)

PR #6877, opened 2025-03-12
APIs for new user management UI

begelundmuller:
**Changes:**
- Adds an API `ListProjectsForOrganizationAndUser` that an org member with `read_org_members` permission can use to list the projects another org member has access to.
- Adds an API `ListUsergroupsForOrganizationAndUser` that an org member with `read_org_members` permission can use to list the user groups another org member belongs to.
- Adds support for optionally filtering by role in the following APIs:
    - `ListOrganizationMemberUsers`
    - `ListOrganizationMemberUsergroups`
    - `ListProjectMemberUsers`
    - `ListProjectMemberUsergroups`
- Added an `include_counts` option to the `ListOrganizationMemberUsers` API. If set, the result will contain two additional fields `projects_count` and `usergroups_count` for each member.
- Added an `include_counts` option to the `ListOrganizationMemberUsergroups` API. If set, the result will contain a `users_count` field for each user group.

Contributes to https://github.com/rilldata/rill/issues/6578

**Frontend integration notes:**

- To show the number of projects and groups that org members belong to, pass `include_counts=true` when calling `ListOrganizationMemberUsers`.
- When you want to list the projects or usergroups a specific org member belongs to, use `ListProjectsForOrganizationAndUser` and `ListUsergroupsForOrganizationAndUser`. Make sure to call these lazily when the org admin drills-down for it (it's too heavy to load these eagerly for all members in the listing at once).

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!


pjain1 (approved):
Do we need cli for these APIs that works for superusers, looks convenient but not a blocker for this PR.

begelundmuller:
> Do we need cli for these APIs that works for superusers, looks convenient but not a blocker for this PR.

Good point – I think I'll wait and see if anyone requests it, then do it in a follow up PR if needed 😄

PR #6879, opened 2025-03-12
feat: Simplify rill time syntax

AdityaHegde:
The current syntax for rill time is not the most ergonomic. We are doing a second pass on the syntax: https://www.notion.so/rilldata/Rill-Time-Documentation-1b2ba33c8f57804fa147cff3cba30121

- [x] Shorten truncate and X-to-date formats.
- [x] Support links.
- [x] Support `First` (<) and `Last` (>) modifiers.
- [x] Support `Next` (+) modifier.
- [x] Support Ordinals. i.e. Week2 (W2) , Month5 (M5)
- [x] Absolute times.
- [x] Support `to` joiner.
- [x] Add timezone override option.
- [x] Support `PeriodToGrain`.
- [x] Adjust final grain based on the resolved time range.
- [x] Should last point be included if incomplete time range is requested? - Yes. A few tests verify this.

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!


runtime/pkg/rilltime/rilltime_test.go
```diff
@@ -4,121 +4,90 @@ import (
...
-		{`0W - 1D, 0W @ latest`, "2024-08-04T00:00:00Z", "2024-08-05T00:00:00Z"},
-		{`now/W, now/W+1D`, "2024-08-12T00:00:00Z", "2024-08-13T00:00:00Z"},
-		{`0W, 0W+1D @ 2024-02-08`, "2024-02-05T00:00:00Z", "2024-02-06T00:00:00Z"},
-		{`0D, 0D+5H @ 2024-02-08 21:34`, "2024-02-08T00:00:00Z", "2024-02-08T05:00:00Z"},
-
-		{"P2DT10H", "2024-08-03T20:00:00Z", "2024-08-06T07:32:36Z"},
-		{"rill-MTD", "2024-08-01T00:00:00Z", "2024-08-06T06:32:37Z"},
-		{"rill-PW", "2024-07-29T00:00:00Z", "2024-08-05T00:00:00Z"},
+		{"m", "2025-03-10T06:31:00Z", "2025-03-10T06:32:00Z", timeutil.TimeGrainSecond},
+		{"m~", "2025-03-10T06:32:00Z", "2025-03-10T06:32:37Z", timeutil.TimeGrainSecond},
```
briangregoryholmes, line 23: I can see both sides to this, but my thinking is that the API should _always_ return intervals that start and end on the colloquial boundaries. So, `m~` should actually end at `2025-03-10T06:33:00Z` even though that is "in the future". The reason for this is threefold 1) whether or not we should display the blank/null data that is in the future is more of a rendering concern and something we want to leave up to the user. 2) When the watermark is in the past, it feels odd to somewhat arbitrarily cut off a complete month. If I ask for `M~` and my data is from 6 years ago, it's probably helpful to see the full month even if the data ends halfway through. 3) When making comparisons to previous periods, it's again a determination of the user as to whether they want to compare partial period to full period or partial to partial. When `~` always returns "complete" periods, we don't need to make an unnecessary fetch request for, let's say, `M~ to +1M` or whatever it might be. The total and time series data would be same, so we can use the query cache for the primary data.

runtime/pkg/rilltime/rilltime_test.go
```diff
@@ -4,121 +4,90 @@ import (
...
+		{"m~ of +2d", "2025-03-12T06:32:00Z", "2025-03-12T06:32:37Z", timeutil.TimeGrainSecond},
+		{"<m of +2d", "2025-03-12T00:00:00Z", "2025-03-12T00:01:00Z", timeutil.TimeGrainSecond},
+		{">m of +2d", "2025-03-12T23:59:00Z", "2025-03-13T00:00:00Z", timeutil.TimeGrainSecond},
+
+		{"W1", "2025-03-03T00:00:00Z", "2025-03-10T00:00:00Z", timeutil.TimeGrainDay},
+		{"W1 by H", "2025-03-03T00:00:00Z", "2025-03-10T00:00:00Z", timeutil.TimeGrainHour},
+		{"W1 of -2M", "2024-12-30T00:00:00Z", "2025-01-06T00:00:00Z", timeutil.TimeGrainDay},
+		{"D3 of W1 of -3Y", "2022-01-05T00:00:00Z", "2022-01-06T00:00:00Z", timeutil.TimeGrainHour},
+		{"W2 of M11 of +3Y", "2028-11-06T00:00:00Z", "2028-11-13T00:00:00Z", timeutil.TimeGrainDay},
+		{"<3m of H2 of -6D of -1M", "2025-02-04T01:00:00Z", "2025-02-04T01:03:00Z", timeutil.TimeGrainSecond},
```
briangregoryholmes, line 47: I think the rule to follow in general with grains is that if you ask for a single unit of time (like W1, <m, -1H), the default grain should be the next lower order grain, as you have been doing. However, if you ask for values greater than a single unit, as in the case of <3m, the default grain should match that first token. So, I would expect this to default to minutes.

runtime/pkg/rilltime/rilltime_test.go
```diff
@@ -4,121 +4,90 @@ import (
...
+		{"2025-03", "2025-03-01T00:00:00Z", "2025-04-01T00:00:00Z", timeutil.TimeGrainWeek},
+		{"2025", "2025-01-01T00:00:00Z", "2026-01-01T00:00:00Z", timeutil.TimeGrainMonth},
+		{"D3 of W1 of 2022", "2022-01-05T00:00:00Z", "2022-01-06T00:00:00Z", timeutil.TimeGrainHour},
+		{"<3m of H2 of 2025-02-04", "2025-02-04T01:00:00Z", "2025-02-04T01:03:00Z", timeutil.TimeGrainSecond},
+
+		{"W1 to W3", "2025-03-03T00:00:00Z", "2025-03-24T00:00:00Z", timeutil.TimeGrainDay},
+		{"W1 to W3 by W", "2025-03-03T00:00:00Z", "2025-03-24T00:00:00Z", timeutil.TimeGrainWeek},
+		{"W1 of -2M to D", "2024-12-30T00:00:00Z", "2025-03-10T00:00:00Z", timeutil.TimeGrainDay},
+		{"W1 of -2M to D~", "2024-12-30T00:00:00Z", "2025-03-10T06:32:37Z", timeutil.TimeGrainDay},
+		{"-4D to -2D", "2025-03-06T00:00:00Z", "2025-03-09T00:00:00Z", timeutil.TimeGrainHour},
```
briangregoryholmes, line 62: So, my thinking here is that this should be a two day period and not a three day period. I put a poll up on slack, though, because others may not agree. I think people will naturally want the sort of quick mental math to match. If I ask for `-12M to -8M`, I think I want that return me a 4 month period (12 minus 8). The idea is {start} to {start}. `-4D to -2D~` would be a way to get three days.

runtime/pkg/rilltime/rilltime_test.go
AdityaHegde: Hmmm wont this will prevent users from comparing say week-to-date and the same period one week ago? It will end up being week-to-date vs full previous week.

briangregoryholmes: It's Thursday today. `W~` would give me March 10th to March 17th and seven total data points. The first three would be full days, the fourth would be partial data for today and the 5th, 6th and 7th would be null.

If I wanted no null data points, I would use `WTD~` or `WTD`.

When comparing to previous week with, for instance, `W~ of -1W` or just `-1W`, I would get seven data points as well, so I'd have a symmetric comparison to the current week. The time series graph would have two lines. One that extends the full interval and one that extends part way (the current period).

In the UI, when comparisons are off, we may give a way to visually truncate those nulls (extend the partial line to the full width of the timeseries graph), but this is a rendering concern and would not be a new API request.

runtime/metricsview/executor_rewrite_time.go
```diff
@@ -87,17 +87,15 @@ func (e *Executor) resolveTimeRange(ctx context.Context, tr *TimeRange, tz *time
 		return err
 	}
 
-	tr.Start, tr.End, err = rillTime.Eval(rilltime.EvalOptions{
+	// TODO: use grain when we have timeseries from metrics_view_aggregation
```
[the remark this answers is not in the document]
AdityaHegde: `rillTime.Eval` now returns grain. So once we use have timeseries queries moved over to using we will consume it.

briangregoryholmes (approved):

briangregoryholmes:
Approving based on intended syntax functionality.

runtime/server/queries_metrics.go
```diff
@@ -426,31 +427,39 @@ func (s *Server) MetricsViewTimeRanges(ctx context.Context, req *runtimev1.Metri
 		return nil, err
 	}
 
+	var tz *time.Location
+	if req.TimeZone != "" {
+		tz, err = time.LoadLocation(strings.Trim(req.TimeZone, "{}"))
```
begelundmuller, line 432: Why can it contain curly braces?

runtime/queries/timeutil.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/rilldata/rill/runtime/pkg/timeutil"
 )
 
-func convTimeGrain(tg runtimev1.TimeGrain) timeutil.TimeGrain {
+func convFromAPITimeGrain(tg runtimev1.TimeGrain) timeutil.TimeGrain {
```
begelundmuller, line 12: nit: `TimeGrainFromAPI` (also makes it public just for symmetry)

runtime/queries/timeutil.go
```diff
@@ -33,6 +33,32 @@ func convTimeGrain(tg runtimev1.TimeGrain) timeutil.TimeGrain {
 	return timeutil.TimeGrainUnspecified
 }
 
+func ConvToAPITimeGrain(tg timeutil.TimeGrain) runtimev1.TimeGrain {
```
begelundmuller, line 36: nit: call it `TimeGrainToAPI`

runtime/pkg/timeutil/timeutil.go
```diff
@@ -184,3 +184,69 @@ func AddTimeProto(to time.Time, tg runtimev1.TimeGrain, count int) time.Time {
...
+	return tm
+}
+
+func CopyTimeTill(from, to time.Time, tg TimeGrain, tz *time.Location) time.Time {
+	year := to.Year()
+	month := to.Month()
+	day := to.Day()
+	hour := to.Hour()
+	minute := to.Minute()
+	second := to.Second()
```
begelundmuller, line 221: Instead of accepting `tz` here, can it use `from.Location()` instead? And perhaps check that `from` and `to` have the same location? I'm asking since functions like `from.Hour()` and `to.Hour()` correct for the timestamp's location, so seems like if they are not the same, it could lead to some inconsistencies.

runtime/pkg/timeutil/timeutil.go
```diff
@@ -184,3 +184,69 @@ func AddTimeProto(to time.Time, tg runtimev1.TimeGrain, count int) time.Time {
...
+	case TimeGrainQuarter:
+		return tm.AddDate(0, n*3, 0)
+	case TimeGrainYear:
+		return tm.AddDate(n, 0, 0)
+	}
+
+	return tm
+}
+
+func CopyTimeTill(from, to time.Time, tg TimeGrain, tz *time.Location) time.Time {
```
begelundmuller, line 215: This function is slightly tricky to understand. Can you add a docstring? And maybe see if you can think of a more descriptive name? 

Not sure, but maybe `MergeComponents(target, source time.Time, until TimeGrain)`?

runtime/pkg/rilltime/rilltime.go
```diff
@@ -431,3 +672,22 @@ func getMinGrain(d duration.StandardDuration) string {
 	}
 	return ""
 }
+
+// truncateWithCorrection truncates time by a grain but corrects for https://en.wikipedia.org/wiki/ISO_week_date#First_week
+// TODO: will adding this directly to timeutil.TruncateTime break anything?
```
begelundmuller, line 677: Should this be addressed now?

runtime/pkg/rilltime/rilltime.go
```diff
@@ -196,149 +263,362 @@ func ParseCompatibility(timeRange, offset string) error {
...
+	if t.offset <= 1 && t.Grain != nil {
+		return lowerOrderMap[t.tg]
 	}
-	if e.End.Now {
-		// if end has now mentioned then start also should be relative to latest.
-		return evalOpts.Now, e.End
+	return t.tg
+}
+
+// TODO: reuse code from duration.ParseISO8601
```
begelundmuller, line 547: Should this be addressed now?

runtime/pkg/rilltime/rilltime_test.go
```diff
@@ -4,121 +4,152 @@ import (
...
+	now := parseTestTime(t, "2025-03-12T10:32:36Z")
+	minTime := parseTestTime(t, "2020-01-01T00:32:36Z")
+	maxTime := parseTestTime(t, "2025-03-11T06:32:36Z")
+	watermark := parseTestTime(t, "2025-03-10T06:32:36Z")
 	testCases := []struct {
 		timeRange string
 		start     string
 		end       string
+		grain     timeutil.TimeGrain
 	}{
```
begelundmuller, line 21: I again think about the case of rounding when the watermark is on a grain boundary. I think that requires test cases for the `s` grain. Can you try restoring these test cases in the new syntax?
```
		{`m : |s|`, "2024-08-05T06:32:00Z", "2024-08-05T06:32:36Z"},
		{`m : s`, "2024-08-05T06:32:00Z", "2024-08-05T06:32:37Z"},
```

begelundmuller (changes requested):

runtime/pkg/rilltime/rilltime.go
AdityaHegde: I dunno how stable this syntax will be. We can do a pass to cleanup and potentially merge this with duration package.

runtime/pkg/rilltime/rilltime.go
AdityaHegde: If we move the logic to TruncateTime alerts/reports might be affected. Mainly when duration package uses it it should perhaps not have this behaviour.

runtime/server/queries_metrics.go
AdityaHegde: Copy paste error, removing it

runtime/pkg/rilltime/rilltime_test.go
AdityaHegde: Added them here: https://github.com/rilldata/rill/pull/6879/commits/2efb5bf1292ae3a58ae81a21c35c704a2fc3d42c

The bucket doesnt impact the end time, it is rather the 1st term of the range that affects it.

runtime/pkg/timeutil/timeutil.go
begelundmuller: Yeah that name is also good. Although `from` and `to` sound like a time range, which might be confusing (hence my suggestion of `source` and `target`, but open to suggestions).

runtime/pkg/rilltime/rilltime_test.go
begelundmuller: What if you had `s by s` and `s~ by s`? Is that supported and what would it output?

Or alternatively, what about `m` and `m~` when the watermark is exactly on the boundary, i.e. `2025-03-10T06:32:00Z`?

begelundmuller:
@AdityaHegde I left some comments on the open threads above.

runtime/pkg/rilltime/rilltime_test.go
briangregoryholmes: `s by s` and `s~ by s` are both technically invalid. I would _prefer_ if they didn't error in the short term since we don't have the concept of "warning" when parsing YAMLs, but it's okay if they do.

We also don't have a concept of data granularity, only smallest time grain. So, the fact that a data point may fall on a boundary is somewhat irrelevant to us. A timestamp of `2025-03-10T06:32:00Z` could represent a full complete minute, a full complete second, a full complete millisecond, etc. We don't have any insight there. As such, we should always treat it as meaning that the grain has been started and is incomplete.

So, asking for `m` should return `2025-03-10T06:31:00Z` to `2025-03-10T06:31:00Z` and API requests will not include the data point represented by `2025-03-10T06:32:00Z`. 

`m~` should return  `2025-03-10T06:32:00Z` to `2025-03-10T06:33:00Z` and the API request will capture that data point.

briangregoryholmes: s by s and s~ by s are both technically invalid. I would prefer if they didn't error in the short term since we don't have the concept of "warning" when parsing YAMLs, but it's okay if they do.

How confident are we that data points that fall on a boundary are representative of that boundary? Do we have any insight into that without a `data_granularity` property? Let's say you have data that comes in with second granularity, but your `smallest_time_grain` is set to minute. A data point with `2025-03-10T06:32:00Z` is, in that case, not representative of a complete minute, so asking for `m` should give you `2025-03-10T06:31:00Z` to `2025-03-10T06:32:00Z`. I'm fine with establishing a rule that assumes `smallest_time_grain` is equivalent to data granularity, in which case `m` and `m~` would return the same thing: `2025-03-10T06:32:00Z` to `2025-03-10T06:33:00Z`

briangregoryholmes: Reflecting on this again. Is it as simple as just saying the anchor is actually timestamp of latest data point + duration of smallest time grain?

begelundmuller: > How confident are we that data points that fall on a boundary are representative of that boundary? 

Not very, it depends on the source/model logic, so I think we need to support a syntax that enables to either include or exclude it. For example, some customers have data that is rolled up in advance using `date_trunc` for the groupings – so for e.g. year-to-date, they want the latest day to be included even though it's time component may be `00:00:00`.

begelundmuller: Example test case to add:
- Watermark is at hour grain
- How to express "today" including the rows at the watermark?
- How to express "today" excluding the rows at the watermark?

begelundmuller: > Is it as simple as just saying the anchor is actually timestamp of latest data point + duration of smallest time grain?

Example:
```
smallest grain = m
watermark = '21:59:55'
```
then:
```
watermark+smallest grain = '22:00:55'
DTH = 00:00 up to 22:00
```

Unexpected because the `DTH` bin is supposed to be compete, but data for 21:59:56 may not have arrived yet.

------

This is due to smallest grain in this example being about presentation, not about data shape. We probably can't ask dashboard developers to know the smallest grain of the underlying data (since it may change dynamically if the source data changes).

runtime/pkg/rilltime/rilltime_test.go
```diff
@@ -4,121 +4,173 @@ import (
...
+		{"2D of -2D", "2025-03-02T00:00:00Z", "2025-03-04T00:00:00Z", timeutil.TimeGrainDay},
+		{"D", "2025-03-05T00:00:00Z", "2025-03-06T00:00:00Z", timeutil.TimeGrainHour},
+		{"D~", "2025-03-06T00:00:00Z", "2025-03-07T00:00:00Z", timeutil.TimeGrainHour},
+
+		{"MTD", "2025-03-01T00:00:00Z", "2025-03-06T00:00:00Z", timeutil.TimeGrainDay},
+		{"MTD by H", "2025-03-01T00:00:00Z", "2025-03-06T00:00:00Z", timeutil.TimeGrainHour},
+		{"MTD of -1M", "2025-02-01T00:00:00Z", "2025-02-06T00:00:00Z", timeutil.TimeGrainDay},
+		{"MTD of -1Y", "2024-03-01T00:00:00Z", "2024-03-06T00:00:00Z", timeutil.TimeGrainDay},
+		{"MTD~", "2025-03-01T00:00:00Z", "2025-03-07T00:00:00Z", timeutil.TimeGrainDay},
+		{"MTD~ by H", "2025-03-01T00:00:00Z", "2025-03-07T00:00:00Z", timeutil.TimeGrainHour},
```
begelundmuller, line 153: I wonder if some would want it to evaluate as `2025-03-06T01:00:00Z`, i.e. only go up by the smaller grain such that the range covers the watermark (which is `2025-03-06T00:00:00Z`)?

But I don't think so – it makes sense that `~` applies to the grain it's attached to. 

@briangregoryholmes Any thoughts on this?

begelundmuller (approved):

runtime/pkg/rilltime/rilltime_test.go
briangregoryholmes: These are two separate concepts. `MTH~` is what you're describing.

`MTH~`: all complete hours of the month plus the current incomplete hour (by hour)
`MTD~ by H`: all complete days of the month plus the current incomplete day (by hour)
`M~ by H`: all hours of the current month

These are all valid requests that have different presentational implications.

When comparing to previous periods, these are things to consider:

- Do the query responses have the same number of data points, regardless of nullity? **Always, yes**
- How many data points do I have (last data point was April 15th at 11:00AM)?
- - `MTH~`: 348
- - `MTD~ by H`: 360
- - `M~ by H`: 720
- Are all data points for both ranges not null (assuming no data gaps)? 
- - `MTH~`: Yes
- - `MTD~ by H`: Primary period hours after watermark until end of _day_ are null
- - `M~ by H`: Primary period hours after watermark until end of _month_ are null
- Where does the right bound of the time series chart fall?
- - `MTH~`: Latest data point + duration of grain
- - `MTD~ by H`: Start of following day
- - `M~ by H`: Start of following month
- Is the big number a symmetric comparison?
- - `MTH~`: Yes
- - `MTD~ by H`: No
- - `M~ by H`: No

If I wanted to track hourly sales and had a dashboard displayed permanently in my office, I would use the time range `D~ by H` with the comparison `-1D`. This would give me two time series lines: the full previous day and the current day up to the hour overlaid on top. This would allow me to track my progress along that line as data came in throughout the day. Other people may instead want to see `DTH~` so that the big number was a symmetric comparison. However, you would not be able to see your "target" for the upcoming hour because the time series chart would not show those data points.

PR #6892, opened 2025-03-14
OLAP Connectors: Test the connection before exiting the "Add Data" modal

ericpgreen2:
This PR ensures that a connection to the OLAP database is successfully established before finalizing the creation of a `connector.yaml` file and updating the `.env` file. Previously, submitting the ClickHouse, Druid, or Pinot connector forms would create these files regardless of the validity of the credentials.

Implementation details:
- Optimistically create a `connector.yaml` file.
- Optimistically create or update the `.env` file with the provided credentials.
- Call the `OLAPListTables` API to verify the connection.
- If the connection fails, remove the connector file, revert the changes made to the `.env` file, and display an error message.

Once https://github.com/rilldata/rill/pull/6671 lands, we can potentially switch the implementation to use a `tmp` directory and, in particular, a `tmp/.env` file to test credentials.

Builds on https://github.com/rilldata/rill/pull/7140

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!


web-common/src/features/connectors/olap/test-connection.ts
```diff
@@ -0,0 +1,66 @@
...
+} from "../../../runtime-client";
+import { fileArtifacts } from "../../entity-management/file-artifacts";
+import { humanReadableErrorMessage } from "../../sources/modal/errors";
+
+interface TestConnectorResult {
+  success: boolean;
+  error?: string;
+}
+
+export async function testOLAPConnector(
```
AdityaHegde, line 16: A method called `testOLAPConnector` seems to be also checking if the file has updated as well.

Perhaps we should split them out, update and confirming of reconcile can be in a method. We could also get rid of the hard wait for `1000s`. We could check for version change in ProjectParser instead?

web-common/src/features/sources/modal/submitAddDataForm.ts
```diff
@@ -141,14 +111,90 @@ export async function submitAddDataForm(
...
+   * Rollback all changes if the connection test fails
+   */
+  if (!result.success) {
+    // Clean-up the `connector.yaml` file
+    await runtimeServiceDeleteFile(instanceId, {
+      path: newConnectorFilePath,
+    });
+
+    // Clean-up the `.env` file
+    if (!envFileExisted) {
```
AdityaHegde, line 162: How long could the test take? Do we need to worry about overriding unintended `.env` changes if it takes a long time?

ericpgreen2: It could take 20-30s until the backend context deadline is exceeded. When this is happening, I've prevented the dialog from closing like so:
```
<Dialog.Root
    open
    onOpenChange={async (open) => {
      if (!open) {
        await onCancelDialog();
      }
    }}
    closeOnEscape={!isSubmittingForm}
    closeOnOutsideClick={!isSubmittingForm}
  >
```

web-common/src/features/connectors/olap/test-connection.ts
AdityaHegde: Ya this is better!

web-local/tests/olap-connectors.spec.ts
```diff
@@ -1,15 +1,21 @@
...
- * Future work: Add tests using ClickHouse testcontainers for end-to-end validation.
- */
+import { ClickHouseTestContainer } from "./utils/clickhouse";
 
 test.describe("ClickHouse connector", () => {
   test.use({ project: "Blank" });
 
+  const clickhouse = new ClickHouseTestContainer();
+
+  test.beforeAll(async () => {
```
AdityaHegde, line 10: Would be good to move this to a static method on ClickHouseTestContainer that also returns the instance.

web-common/src/features/sources/modal/submitAddDataForm.ts
```diff
@@ -142,3 +205,53 @@ export async function submitAddDataForm(
   // Go to the new connector file
   await goto(`/files/${newConnectorFilePath}`);
 }
+
+async function beforeSubmitForm(instanceId: string) {
+  // Emit telemetry
+  await behaviourEvent?.fireSourceTriggerEvent(
```
AdityaHegde, line 211: Do we await everywhere we use telemetry? Perhaps it would be good to revisit this, if the end point is unreachable all the actions will be blocked.

web-common/src/features/sources/modal/submitAddDataForm.ts
```diff
@@ -123,14 +115,85 @@ export async function submitAddDataForm(
...
+  /**
+   * Test the new OLAP connector:
+   * 1. Ensure the file has reconciled and has no errors
+   * 2. Test the connection to the OLAP database
+   */
+
+  // Wait until the project parser has reconciled the new connector file
+  await waitForProjectParserVersion(
+    instanceId,
+    projectParserStartingVersion + 1,
```
AdityaHegde, line 161: Saving `projectParserStartingVersion` just to wait for next version seems weird in the caller function.

Could we add a version of `runtimeServicePutFile` that hides the details from the caller?

web-common/src/features/entity-management/project-parser.ts
```diff
@@ -0,0 +1,45 @@
...
+    throw new Error("Project parser version not found");
+  }
+
+  if (Number(projectParserQuery.resource.meta.version) >= version) {
+    return;
+  }
+
+  await new Promise((resolve) => setTimeout(resolve, 300));
+
+  return waitForProjectParserVersion(instanceId, version);
```
AdityaHegde, line 44: A loop version might be better here than the recursion one. Will be more readable since recursion isnt really giving any advantage.

web-local/tests/olap-connectors.spec.ts
ericpgreen2: I can't quite tell exactly what you have in mind. But regarding another layer of abstraction– that might be valuable, but I'd prefer not to add extra abstractions here before we know how our usage of this class will evolve. For example, I could especially see the `seed` method being parameterized based on which dataset you want to use for a given test suite.

web-common/src/features/sources/modal/submitAddDataForm.ts
ericpgreen2: Good point. Removed.

web-local/tests/olap-connectors.spec.ts
AdityaHegde: We could do something like `ClickHouseTestContainer.use(<seed param>)`. But lets do that in a follow up when we use more of ClickHouseTestContainer

PR #6903, opened 2025-03-17
fix (Models): Escape non-standard table names

ericpgreen2:
Addresses [this Slack report](https://rilldata.slack.com/archives/C02T907FEUB/p1741981598994669)

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


AdityaHegde (changes requested):
We allow for a lot more chars in table names in backend https://github.com/rilldata/rill/blob/main/runtime/drivers/olap.go#L283

Perhaps we should do the same and always surround by quotes if it is not alpha-numeric.

AdityaHegde (approved):

Issue #6770, opened 2025-02-26
Timeseries API returns empty results when caching is enabled for clickhouse


k-anshul:
This is usually observed when running clickhouse in replication mode.

rillsidd:
hmm, I am seeing this with duckDB dashboards also - https://ui.rilldata.com/mobilefuse/mobile-fuse-test/explore/demand_v2_explore

PR #6927, opened 2025-03-19
fix: query cancellation leads to null results being cached

k-anshul:
closes https://github.com/rilldata/rill/issues/6770

**Checklist:**
- [x] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/runtime/queries/metricsview_timeseries.go
+++ b/runtime/queries/metricsview_timeseries.go
@@ -234,8 +234,9 @@ func (q *MetricsViewTimeSeries) populateResult(rows *drivers.Result, tsAlias str
 		})
 		start = addTo(t, dur, tz)
 	}
-	if rows.Err() != nil {
-		return rows.Err()
+	err := rows.Err()
+	if err != nil {
+		return err
 	}
 	if q.TimeEnd != nil && nullRecords != nil {
 		if start.Equal(zeroTime) && q.TimeStart != nil {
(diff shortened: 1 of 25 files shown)


runtime/queries/metricsview_timeseries.go
```diff
@@ -234,6 +234,9 @@ func (q *MetricsViewTimeSeries) populateResult(rows *drivers.Result, tsAlias str
 		})
 		start = addTo(t, dur, tz)
 	}
+	if rows.Err() != nil {
```
pjain1, line 237: nit: seems like a function call can we catch it in variable use it

runtime/queries/table_columns.go
```diff
@@ -117,6 +117,9 @@ func (q *TableColumns) Resolve(ctx context.Context, rt *runtime.Runtime, instanc
 				pcs = append(pcs, &pc)
 				i++
 			}
+			if rows.Err() != nil {
```
pjain1, line 120: nit: same as earlier

pjain1 (approved):
Good catch! not related to caching but there seem to be other places where we iterate over results but does not check error like in `SelectInlineResults` do we need change there as well.

runtime/queries/metricsview_timeseries.go
k-anshul: Done

PR #6935, opened 2025-03-20
reduce connection concurrency

pjain1:
In my observation, see 10-20% improvement for CH based page loads


k-anshul:
> In my observation, see 10-20% improvement for CH based page loads

Is this also for projects with too many dimensions and measures ?

pjain1:
> Is this also for projects with too many dimensions and measures ? 
Yes for those only

k-anshul (approved):

Issue #6765, opened 2025-02-26
improve formatting of Excel (XLSX) exports in Explore dashboards


medriscoll:
**Is your feature request related to a problem? Please describe.**

Currently, when data is exported to the Excel XLSX format from Pivot table views in Rill, two issues arise:

* Time dimension fields appears as strings such as `2025-01-23T00:00:00Z`
* Other values appear with quotation marks surrounding them

Additionally, in the time dimension detail view, date columns are suffixed with the name of the measure, which also makes them difficult to parse.  (This issue can be addressed at a later time).

**Describe the solution you'd like**

* Rill should export time fields that are recognized as date types in Excel rather than strings.
* Rill should not surround values with quotation marks.

**Additional context**

Rill currently uses the open-source Golang library excelize to handle these exports, which has support for these data types.

begelundmuller:
Additional context in this Slack thread: https://rilldata.slack.com/archives/C01A9DYP013/p1740508318827489?thread_ts=1740486655.142019&cid=C01A9DYP013

begelundmuller:
@NamanMahor I think a good first step here is to just try out DuckDB's Excel export support on DuckDB 1.2. It's quite likely it has improved a lot. See this code: https://github.com/rilldata/rill/blob/9af652b61fb54194ed9a22d0e474718b1195a0ba/runtime/drivers/duckdb/model_executor_self_file.go#L129-L150.

If that goes poorly, let's try to improve the Excel export support here: https://github.com/rilldata/rill/blob/9af652b61fb54194ed9a22d0e474718b1195a0ba/runtime/drivers/file/model_executor_olap_self.go#L175.

But if it goes well, we could try to use DuckDB for all Excel exports, even for ClickHouse/Druid data. We already use DuckDB to do pivots of ClickHouse/Druid data (since those databases don't natively support pivots), so doing the same for exports should be relatively straightforward. Here's a link to the code that does the handover of data from ClickHouse/Druid to DuckDB for pivots to look at as an example: https://github.com/rilldata/rill/blob/9af652b61fb54194ed9a22d0e474718b1195a0ba/runtime/metricsview/executor_pivot.go#L117-L139

NamanMahor:
The issue with quotes is not related to the Excel export. it occurs when we first export to Parquet for the PivotTable.

PR #6939, opened 2025-03-20
Excel export issue

NamanMahor:
> when data is exported to the Excel XLSX format from Pivot table views in Rill, two issues arise:
> 1. Time dimension fields appears as strings such as 2025-01-23T00:00:00Z

Actually it occurs when we export to csv so need fix in csv export(confirm from file share by Customer). 

> 2. Other values appear with quotation marks surrounding them

This was issue in both excel csv and parquet export. This issue was happening when we convert to parquet file in case of pivot table. 

Closed [#6765](https://github.com/rilldata/rill/issues/6765)

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!


runtime/drivers/file/model_executor_olap_self.go
```diff
@@ -334,12 +342,15 @@ func writeParquet(res *drivers.Result, fw io.Writer) error {
 					return err
 				}
 				recordBuilder.Field(i).(*array.TimestampBuilder).Append(tmp)
-			case runtimev1.Type_CODE_STRING, runtimev1.Type_CODE_INTERVAL, runtimev1.Type_CODE_DATE, runtimev1.Type_CODE_ARRAY, runtimev1.Type_CODE_STRUCT, runtimev1.Type_CODE_MAP, runtimev1.Type_CODE_JSON, runtimev1.Type_CODE_UUID:
+			case runtimev1.Type_CODE_STRING, runtimev1.Type_CODE_DATE:
+				v, _ := v.(string)
```
k-anshul, line 346: Datetype `date` will not be scanned to a string.
```
h, err := drivers.Open("clickhouse", "default", map[string]any{"dsn": "clickhouse://localhost:9000"}, storage.MustNew(os.TempDir(), nil), activity.NewNoopClient(), zap.NewNop())
	if err != nil {
		panic(err)
	}

	olap, _ := h.AsOLAP("default")

	result, err := olap.Execute(context.Background(), &drivers.Statement{
		Query: "SELECT '2024-01-01'::Date32",
	})
	if err != nil {
		panic(err)
	}
	var t any
	for result.Next() {
		if err := result.Scan(&t); err != nil {
			panic(err)
		}
		// panic: interface conversion: interface {} is time.Time, not string
		println(t.(string))
	}
```

k-anshul (changes requested):

runtime/drivers/file/model_executor_olap_self.go
NamanMahor: you are right about date type can't be converted to string but in this case  v is already string because v, err := jsonval.ToValue(v, res.Schema.Fields[i].Type) converted it to string.

k-anshul: Got it.

runtime/drivers/file/model_executor_olap_self.go
```diff
@@ -203,22 +216,34 @@ func writeXLSX(res *drivers.Result, fw io.Writer) error {
...
 
-			switch res.(type) {
-			case nil:
-				res = ""
-			case []any, map[string]any:
-				res, err = json.Marshal(res)
-				if err != nil {
-					return fmt.Errorf("failed to marshal JSON value: %w", err)
+			var s any
+			if jval != nil {
```
k-anshul, line 225: Instead of this shouldn't it check for for `nil` and write empty string as in the current code ?

k-anshul (approved):

runtime/drivers/file/model_executor_olap_self.go
NamanMahor: As discussed it is should be nil and empty string.

PR #6941, opened 2025-03-20
canvas: mar 20 uxqa

briangregoryholmes:
- Fixes a regression where clicking the canvas surface did not deselect the widget
- Reworks KPI tooltip to display values from all lines
- Fixes an issue where the KPI component would try to render out a sparkline even when the metrics view had no time series column set
- Removes `Generate chart with AI` dropdown menu item from metrics views

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!


djbarnwal (commented):
Flat table has padding but the pivot table doesn't seem to have any padding
image

image


Also, long term I would ideally prefer that we reuse our existing mouseover and chart components. Creating new component would lead to more maintenance overhead.

djbarnwal (approved):

PR #6956, opened 2025-03-24
feat: trace viewer for rill developer

k-anshul:
closes https://github.com/rilldata/rill-private-issues/issues/1338

1. Outputs all traces to a file `~/.rill/otel_traces.log`
2. A single trace can be visualised on static page `http://localhost:9009/trace-viewer/` either by trace_id or ingestion of a particular resource name.

image
 

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


cli/pkg/local/server.go
```diff
@@ -816,3 +819,114 @@ func (nameConflictRetryErrClassifier) Classify(err error) retrier.Action {
 
 	return retrier.Fail
 }
+
+// traceHandler returns trace information corresponding to the trace ID.
+func (s *Server) traceHandler() http.Handler {
```
begelundmuller, line 824: Maybe a brief comment about how this works? I.e. that it uses DuckDB to search the output of `observability.FileExporter`, which is used when `--debug` is passed, and returns a specific trace as JSON?

cli/pkg/local/server.go
```diff
@@ -816,3 +819,114 @@ func (nameConflictRetryErrClassifier) Classify(err error) retrier.Action {
...
+		if resourceName == "" && traceID == "" {
+			w.WriteHeader(http.StatusBadRequest)
+			return
+		}
+		if resourceName != "" && traceID != "" {
+			w.WriteHeader(http.StatusBadRequest)
+			return
+		}
+
+		db, err := sqlx.Open("duckdb", "")
```
begelundmuller, line 837: Consider moving the logic starting here to `runtime/pkg/observability/fileexporter.go` since it's coupled to the behavior of the file exporter. E.g. it could be `json, err := observability.SearchTracesFile (...)` or something like that. That would also enable having the `traceSpan` type in the `observability` package where it feels like it belongs.

runtime/pkg/observability/fileexporter.go
```diff
@@ -0,0 +1,81 @@
...
+func NewFileExporter() (*FileExporter, error) {
+	filepath, err := dotrill.ResolveFilename("otel_traces.log", true)
+	if err != nil {
+		return nil, err
+	}
+	// Setup lumberjack for log rotation
+	logWriter := &lumberjack.Logger{
+		Filename:   filepath, // File path
+		MaxSize:    10,       // Max file size in MB
+		MaxBackups: 3,        // Keep up to 3 old log files
```
begelundmuller, line 32: At such small sizes, does it even make sense to have three files? A single 30MB file would already be very small.

runtime/pkg/observability/observability.go
```diff
@@ -25,6 +25,7 @@ type Exporter string
 const (
 	NoopExporter       Exporter = ""
 	OtelExporter       Exporter = "otel"
+	FileBasedExporter  Exporter = "file"
```
begelundmuller, line 28: nit: can omit the "based": `FileExporter`

web-common/static/trace-viewer/index.html
```diff
@@ -0,0 +1,287 @@
+<!DOCTYPE html>
```
begelundmuller, line 1: Subjective, but I wonder if we should put this closer to the observability code, e.g. in `runtime/pkg/observability/embed/file-trace-viewer.html` or `cli/pkg/local/embed/file-trace-viewer.html` and mount it directly on the local server.

runtime/pkg/observability/fileexporter.go
k-anshul: Okay we can keep 1 backup which means max logs can be 60 MB.

runtime/pkg/observability/observability.go
begelundmuller: Ah I see – on a quick glance, the `FileExporter` type isn't used outside the package, so since its an implementation detail, could it be named `fileExporter` (lowercase) or `fileExporterImpl` instead? So that we can keep the naming of the external constants consistent.

PR #6965, opened 2025-03-25
throw internal server error if OLAP fails on dynamic runtimes

k-anshul:
closes https://github.com/rilldata/rill-private-issues/issues/1228

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/runtime/health.go
+++ b/runtime/health.go
@@ -46,7 +46,8 @@ func (r *Runtime) Health(ctx context.Context) (*Health, error) {
 
 	ih := make(map[string]*InstanceHealth, len(instances))
 	for _, inst := range instances {
-		ih[inst.ID], err = r.InstanceHealth(ctx, inst.ID)
+		// if there is a single instance hosted on this runtime then instead of returning error msgs throw error if OLAP/repo/controller are in error state
+		ih[inst.ID], err = r.InstanceHealth(ctx, inst.ID, len(instances) == 1)
 		if err != nil && !errors.Is(err, drivers.ErrNotFound) {
 			return nil, err
 		}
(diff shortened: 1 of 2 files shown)


runtime/health.go
```diff
@@ -46,7 +46,8 @@ func (r *Runtime) Health(ctx context.Context) (*Health, error) {
 
 	ih := make(map[string]*InstanceHealth, len(instances))
 	for _, inst := range instances {
-		ih[inst.ID], err = r.InstanceHealth(ctx, inst.ID)
+		// if there is a single instance hosted on this runtime then instead of returning error msgs throw error if OLAP/repo/controller are in error state
+		ih[inst.ID], err = r.InstanceHealth(ctx, inst.ID, len(instances) == 1)
```
begelundmuller, line 50: nit: the `retErr` option maybe saves a few lines of code, but I think it might be nicer to keep the signature and behavior simple, and explicitly handle the desired error cases here, e.g.:
```suggestion
		ih[inst.ID], err = r.InstanceHealth(ctx, inst.ID)
		if err != nil ... {}
		
		// If there is single instance ....
		if len(instances) == 1 {
			h := ih[inst.ID]
			if h.Controller != "" {
				return ...
			}
			if h.OLAP != "" {
				return ...
			}
			// etc.
		}
```

k-anshul: This is also fine just that need to duplicate this handling at two places. Here as well as API handler.

runtime/server/health.go
```diff
@@ -64,6 +65,16 @@ func (s *Server) InstanceHealth(ctx context.Context, req *runtimev1.InstanceHeal
...
+	// throw error if OLAP/repo/controller are in error state
+	if h.OLAP != "" {
+		return nil, errors.New(h.OLAP)
+	}
+	if h.Repo != "" {
+		return nil, errors.New(h.Repo)
+	}
+	if h.Controller != "" {
+		return nil, errors.New(h.Controller)
+	}
```
begelundmuller, line 77: Is it needed in `InstanceHealth`? It just seems weird because the point of `InstanceHealth` is to get a breakdown of which parts are healthy and which have errors. But if it returns the first error it finds, then you don't know what other parts are healthy/unhealthy.

k-anshul: I mean when we built the `v1/health` API we wanted it to return statuses and not fail like a liveness probe but we are changing the behaviour now. 

Maybe we can keep `InstanceHealth` to return detail statuses.

begelundmuller (approved):

runtime/server/health.go
begelundmuller: Yeah it still feels slightly weird to treat this as a liveness probe (since it needs to be live for the user to update the DSN and fix the error)... but open to whatever makes infra happy atm.

Issue #1456, opened 2022-12-15
refactor use of bold types throughout application


bcolloran:

bcolloran:
It may be the case that the text only looks too heavy when rendered on Windows. There is an optional branch here https://github.com/rilldata/rill-developer/pull/1449 that has lighter text throughout the app -- IMO it looks a bit better on windows than the slightly heavier default text in #1414, see screenshots here:
https://www.notion.so/rilldata/Migrating-Fonts-61999aa6d1b04b52ad7c6357d54f8cdd#d53c01118fde463991f0aa466baa7a80

It's a marginal improvement, not make or break, and obvi totally reasonable if we want to focus on Mac users, so NP if we want to "wontfix" the bold type refactor for now.

PR #6969, opened 2025-03-25
athena integration test

NamanMahor:
Closes [#1456](https://github.com/rilldata/rill-private-issues/issues/1456)

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!


.github/workflows/go-cover.yml
```diff
@@ -38,6 +38,8 @@ jobs:
         RILL_RUNTIME_GCS_TEST_HMAC_SECRET: ${{ secrets.RILL_RUNTIME_GCS_TEST_HMAC_SECRET }}
         RILL_RUNTIME_S3_TEST_AWS_ACCESS_KEY_ID: ${{ secrets.RILL_RUNTIME_S3_TEST_AWS_ACCESS_KEY_ID }}
         RILL_RUNTIME_S3_TEST_AWS_SECRET_ACCESS_KEY: ${{ secrets.RILL_RUNTIME_S3_TEST_AWS_SECRET_ACCESS_KEY }}
+        RILL_RUNTIME_ATHENA_TEST_AWS_ACCESS_KEY_ID: ${{ secrets.RILL_RUNTIME_S3_TEST_AWS_ACCESS_KEY_ID }}
```
k-anshul, line 41: The secret key seems to be same as `RILL_RUNTIME_S3_TEST_AWS_ACCESS_KEY_ID`. May be create separate keys even if values are same ?

k-anshul (approved):
So if it runs for 100 times a day what are the costs we are looking at ?

NamanMahor:
> So if it runs for 100 times a day what are the costs we are looking at ?

cost is $0.00005 per query for less the 10Mb query
I assume 500 query per day it would be 0.75$ per month 

but I will also confirm with @himadrisingh before merging the PR

.github/workflows/go-cover.yml
NamanMahor: sure

PR #6975, opened 2025-03-26
snowflake integration test

NamanMahor:
Closed [#1458](https://github.com/rilldata/rill-private-issues/issues/1458)

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!


runtime/resolvers/testdata/snowflake_connector.yaml
```diff
@@ -0,0 +1,141 @@
+connectors:
+  - clickhouse
+  - snowflake
+  - s3
+project_files:
+  all_datatypes_clickhouse.yaml:
+    type: model
+    connector: snowflake
+    # we are not doing * because TIMESTAMP_TZ and TIMESTAMP_LTZ is not support  while exporting to parquet
```
NamanMahor, line 9: @k-anshul  n @begelundmuller  you might be aware of this but commenting here so it not get missed in review.

runtime/resolvers/testdata/snowflake_connector.yaml
```diff
@@ -0,0 +1,141 @@
...
+    sql: "select id, boolean_col, tinyint_col, smallint_col, int32_col, int64_col, number_col, float_col, double_col, decimal_col, string_col, text_col, date_col, time_col, timestamp_ntz_col, variant_col, array_col, object_col, binary_col, geography_col, geometry_col from INTEGRATION_TEST.PUBLIC.all_datatypes"
+    stage:
+      connector: s3
+      path: s3://integration-test.rilldata.com/snowflake/
+    output:
+      connector: clickhouse
+  all_datatypes_duckdb.yaml:
+    type: model
+    connector: snowflake
+    # we are not doing * because time_col:TIME is not support https://github.com/rilldata/rill/blob/main/runtime/drivers/snowflake/warehouse.go#L192-L193
```
NamanMahor, line 19: this too.

k-anshul (approved):

runtime/resolvers/resolvers_test.go
```diff
@@ -84,6 +84,10 @@ func TestResolvers(t *testing.T) {
 		t.Run(fileutil.Stem(f), func(t *testing.T) {
 			// Load the test file.
 			data, err := os.ReadFile(f)
+			fileName := fileutil.Stem(f)
+			if isRestrictedConnector(fileName) && !isRestrictedTestEnabled(fileName) {
```
k-anshul, line 88: Instead of hardcoding this based on file name, can this check be run after parsing connectors at line 117 and checking if it is a restricted connector. 
Then we can just check on the connector name and the check still works if somebody creates two YAMLs for snowflake tests.

runtime/resolvers/resolvers_test.go
```diff
@@ -229,6 +233,29 @@ func TestResolvers(t *testing.T) {
...
+	switch connector {
+	case "snowflake_connector":
+		return true
+	default:
+		return false
+	}
+}
+
+func isRestrictedTestEnabled(connector string) bool {
+	env := os.Getenv("RILL_RUNTIME_RUN_RESTRICTED_RESOLVERS_TESTS")
```
k-anshul, line 246: This is somewhat confusing. Can this be called `RILL_RUNTIME_RESTRICTED_CONNECTORS` and just check for any connector to be present here then we can remove hardcoded `isRestrictedConnector` check as well and can enable/disable tests for any connector by changing envs. 
For local also you can set `RILL_RUNTIME_RESTRICTED_CONNECTORS` as snowflake in the `.env`.

k-anshul: Also add docs about why snowflake is restricted.

k-anshul (changes requested):

k-anshul (approved):

PR #6964, opened 2025-03-25
feat: Restore certain explore user state across page loads.

AdityaHegde:
closes [#6914](https://github.com/rilldata/rill/issues/6914)

- [x] Refactor url state management to a single component with a couple of classes.
- [x] Change the order of priority for dashboard state initialisation. https://www.notion.so/rilldata/Restore-users-to-a-most-recent-state-of-Rill-Cloud-dashboards-across-sessions-and-introduce-home-ico-5927980f4c6040538b4fd11cdfeb952e?pvs=4#1b4ba33c8f57802dabb4d60a425a44d3
- [x] Update explore state stored in sessionsStorage and localStorage to use url params format.
- [x] Add back tests for data store and load from sessionsStorage.
- [x] Add tests for data store and load from localStorage.
- [x] Save relevant fields as "most recent state" in views other than explore.
- [x] Cleanup references to old ExplorePreset based sessionStorage.
- [x] Fix `grain` being added in url in pivot view.

Note that we are getting rid of the old behaviour where clearing params in the address bar would take the user to empty dashboard. This will be replaced by a home button.

The order of loading of initial dashboard state is as follows,
1. If state is present in the url, use it.
2. If no url state, load from session storage (only persists within the tab)
3. If no url state, session storage, restore last user state (from local storage).
4. If no url state, session storage, user state, apply home bookmark.
5. If no url state, session storage, user state, home bookmark, apply explore.yaml defaults
6. If no url state, session storage, user state, home bookmark or defaults open as blank dashboard.

**Checklist:**
- [x] Covered by tests
- [x] Ran it and it works as intended
- [x] Reviewed the diff before requesting a review
- [x] Checked for unhandled edge cases
- [x] Linked the issues it closes
- [ ] Checked if the docs need to be updated
- [ ] Intend to cherry-pick into the release branch
- [x] I'm proud of this work!

PR #6979, opened 2025-03-26
feat: Move home bookmark to a separate button on the top bar

AdityaHegde:
- [x] Move the home bookmark out of bookmarks dropdown.
- [x] The button should be available even when there is no home bookmark. Should take the user to blank dashboard, i.e. a url without any params.
- [x] Fix issue where the button doesnt apply the state due to race condition. Not worth fixing on main's code since there are changes coming in https://github.com/rilldata/rill/pull/6964
- [ ] ~~Figure out what we will show on a shared url and rill-dev.~~ We will not show this for shared-url and rill-dev.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


ericpgreen2 (commented):
Let's get a Product review from @jkhwu. A few UXQA I see:
1. The Home button does not have the same hover state as the buttons it's next to
2. Should the Home button be disabled when either there's no Home bookmark or you're currently viewing the Home state? Disabled + tooltip to explain why.
3. It feels a bit off that the "Bookmark your current view as home" action exists in a dropdown that's anchored off a different button. Maybe, when there's no Home bookmark, the Home button should open a dropdown menu with the prompt?(CC @jkhwu)

Looks like the Home bookmark e2e test needs to be updated.

AdityaHegde:
Thanks @ericpgreen2 
1. Ya lemme fix that.
2. We can disable when the current state is equal to the button's link. But we need it to take the user to blank dashboard when there is no home bookmark. This i per requirement, cc @mindspank 
3. This is by design as well. cc @jkhwu 
4. Will fix the tests.

ericpgreen2:
Code looks good. I'll defer to @mindspank to approve as product owner, as I have reservations about this:

> It feels a bit off that the "Bookmark your current view as home" action exists in a dropdown that's anchored off a different button.

ericpgreen2 (changes requested):
Design-wise, IMO we need to make clear the difference between these two menu items. They look "equal", yet one should be clearly a navigation action and one should clearly be not. A couple ideas:
1. Adding a separator between the menu items
2. Editing the "Home" copy to be "Go to home"
3. Moving the "Go to home" menu item above the other, since it'll be the more common action. 

@jkhwu, any other thoughts on this?

image

ericpgreen2 (approved):

PR #6982, opened 2025-03-26
Test fixtures for `admin` and `cli`

begelundmuller:
**Changes:**
- Refactors test fixtures for the admin service into a dedicated `testadmin` package, enabling use of the fixtures outside the `server` package
- Introduces a `testcli` package with fixtures for easily testing CLI commands
- Makes various minor refactors to enable testing of CLI commands. Notably, it changes `dotrill` to use a struct scoped to an overridable home directory.
- Adds initial CLI tests for `rill org`
- Removes some old CLI tests which didn't work and have been disabled for a long time

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!


cli/cmd/org/delete.go
```diff
@@ -50,7 +47,7 @@ This operation cannot be undone. Use --force to skip confirmation.`,
 				}
 			}
 
-			if !force {
+			if ch.Interactive {
```
pjain1, line 50: nit: should this be controlled by the cmd own args instead of a general setting

pjain1 (approved):

admin/client/client.go
```diff
@@ -30,7 +30,8 @@ const retryPolicy = `{"methodConfig": [{
 type Client struct {
 	adminv1.AdminServiceClient
 	adminv1.AIServiceClient
-	conn *grpc.ClientConn
+	Token string
```
pjain1, line 33: exposing just for tests but I guess no other way around this

begelundmuller: Yeah, but could conceivably be useful for non-test use cases also in the future I guess.

PR #6995, opened 2025-03-28
refac: remove canvas StateManagersProvider pattern, consolidate ComponentRenderer and CanvasComponent

briangregoryholmes:
This PR is the first step in a larger refactor with the goal of cleaning up our state management in the Canvas surface and storing most state in the URL. The goal is to make things easier to follow by reducing the steps required to go from fetched resources to rendered dashboards and ensuring that those steps can be followed in sequence. The Canvas surface provides us an opportunity to establish new, clearer patterns for state management.

More immediately, this resolves the issue where components go blank and throw an error  (such as `Cannot destructure property 'canvasEntity' of 'ctx' as it is undefined.`) after hot module reloading in development.

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!


web-common/src/features/canvas/CanvasDashboardEmbed.svelte
```diff
@@ -35,33 +35,43 @@
...
+    {maxWidth}
+    filtersEnabled={canvas?.spec?.filtersEnabled}
+    {canvasName}
+  >
+    {#each rows as { items = [], height = MIN_HEIGHT, heightUnit = "px" }, rowIndex (rowIndex)}
+      {@const widths = normalizeSizeArray(items?.map((el) => el?.width ?? 0))}
+      {@const types = items?.map(
+        ({ component }) =>
+          canvasData?.resolvedComponents?.[component ?? ""]?.component?.spec
+            ?.renderer,
```
djbarnwal, line 49: For this component, we can simplify some of these references by using the `useCanvas` selector available in `features/canvas/selector.ts`. This also makes sure that we don't use different instances of the same query.

web-common/src/features/canvas/RowBasedCanvas.svelte
```diff
@@ -75,23 +71,27 @@
   let dragTimeout: ReturnType<typeof setTimeout> | null = null;
   let dragItemPosition = { top: 0, left: 0 };
   let dragItemDimensions = { width: 0, height: 0 };
-  let spec = $canvasSpec ?? {
+  let spec = canvasData?.canvas?.canvas?.spec ?? {
```
djbarnwal, line 74: I am wondering if we should use `validSpec` here?

web-common/src/features/canvas/RowBasedCanvas.svelte
```diff
@@ -75,23 +71,27 @@
...
     rows: [],
     filtersEnabled: false,
     maxWidth: DEFAULT_DASHBOARD_WIDTH,
   };
   let openSidebarAfterSelection = false;
 
   $: ({ instanceId } = $runtime);
+  $: metricsViews = Object.values(canvasData?.referencedMetricsViews ?? {});
```
djbarnwal, line 82: We should use the `useCanvas` selector upstream so these sort of references can be abstracted.

web-common/src/features/workspaces/CanvasWorkspace.svelte
```diff
@@ -65,6 +65,14 @@
 
   $: mainError = lineBasedRuntimeErrors?.at(0);
 
+  $: canvasResolverQuery = createQueryServiceResolveCanvas(
+    instanceId,
+    canvasName,
+    {},
+  );
+  $: canvasResolverQueryResult = $canvasResolverQuery;
+  $: canvasData = canvasResolverQueryResult.data;
```
djbarnwal, line 74: Use `useCanvas` selector

djbarnwal (commented):
Looks good! Thanks for making these changes.

One thing I am concerned about is prop drilling across canvas components. In this state model, we have to pass `canvasName` wherever we want to invoke the state managers. 

One of the main reasons we moved to `StateManagerProvider` in explore was to avoid this issue where we were prop drilling `metricViewName` across all our components. Is there a way we can solve for that?

cc @ericpgreen2 if you want to add your thoughts here.

djbarnwal (approved):

PR #7007, opened 2025-03-28
refac: kpi data provider

briangregoryholmes:
This PR refactors the KPI Canvas component to separate data fetching/preparation and rendering concerns. The rendering component, `KPI`, has also been updated with a more comprehensive loading skeleton.

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!


djbarnwal (changes requested):
Thanks for consolidating this. Since we are making this change and removing the KPI component altogether - 
- can we also make sure that we are not making individual aggregation queries for each KPI measure and instead batch them into a single query
- Remove all references of KPI component and it's index file?

briangregoryholmes:
Removed the unused references. Will follow up in a separate PR to address the batched queries.

PR #6986, opened 2025-03-26
chore: bump Orval and TanStack Query

briangregoryholmes:
Rill has been stuck on version 4.29 of TanStack query and version 6.12 of Orval for years. It appears that the primary hurdle was a small API change regarding the use of a custom `queryClient` in TanStack Query v5 that Orval was not accounting for.

A PR that I submitted against Orval (https://github.com/orval-labs/orval/pull/1992) to resolve that issue was released with version 7.8 https://github.com/orval-labs/orval/releases/tag/v7.8.0

- Removes dependency on Rill's fork of TanStack query that is pinned to an older version (repo to be deleted: https://github.com/rilldata/query)
- Bumps TanStack Query to version 5.69
- Bumps Orval to version 7.8
- Fixes deprecated API method of passing a custom `queryClient`
- Removes handrolled infinite queries for `ListMagicAuthTokens`, `ListOrganizationInvites`, and `ListOrganizationMemberUsers` and instead generates them with Orval
- Removes `add-signal-argument` transformer as this has been a native feature of Orval since 2022 https://github.com/orval-labs/orval/pull/357
- Fixes an API change with `invalidateQueries`, `refetchQueries` and related functions
- Fixes type issues related to the change of `any` to `unknown`

**Checklist:**
- [ ] Covered by tests
- [x] Ran it and it works as intended
- [x] Reviewed the diff before requesting a review
- [x] Checked for unhandled edge cases
- [ ] Linked the issues it closes
- [x] Checked if the docs need to be updated
- [ ] Intend to cherry-pick into the release branch
- [x] I'm proud of this work!

PR #7021, opened 2025-03-31
feat: improved leaderboard loading state

briangregoryholmes:
This implementation will change slightly after https://github.com/rilldata/rill/pull/6986 merges and we start to use TanStack Query v5's new features, but this is a stopgap solution to fix a few issues.

The primary issue is that `keepPreviousData` or, in v5, `placeholderData` does not work with reactive query assignment https://github.com/TanStack/query/issues/5913. We can address this once we upgrade to v5 and can pass the query options as a store, but, in the meantime, it means we should display some kind of skeleton loading state whenever the aggregation queries are running.

The DelayedLoadingRows implementation has changed to take both `isLoading` and `isFetching`. Loading rows should always be displayed if `isLoading` is true, but if only `isFetching` is true, then we should wait a certain amount of time before displaying the loading state. In v4 (or at least our pinned version), it appears that `isFetching` and `isLoading` do not have the same separation as they do in v5. They are usually the same.

Additionally, it supports rendering a variable number of rows based on the previous row count rather than always showing 7 rows.

Fixes a styling issue with the `LeaderboardHeader` loading indicator.

In general, the idea going forward will be this:

1. Interact with Leaderboard to trigger a new query
2. After ~150ms, display the refetching indicator in the LeaderboardHeader, but continue to display the previous data
3. If still loading after ~500ms (values tbd), display the `LoadingRows` component, which may be consolidated with `LeaderboardRow` itself

Screenshot 2025-03-31 at 3 23 21 PM

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-common/src/features/dashboards/leaderboard/DelayedLoadingRows.svelte
+++ b/web-common/src/features/dashboards/leaderboard/DelayedLoadingRows.svelte
@@ -1,27 +1,31 @@
-<!-- @component 
-  This component is used to only show loading rows if the loading state is true after a delay.
-  This is handy for preventing the loading rows from flickering.
--->
 <script lang="ts">
   import { onDestroy } from "svelte";
   import { writable } from "svelte/store";
   import LoadingRows from "./LoadingRows.svelte";
 
   export let isLoading: boolean;
+  export let isFetching: boolean;
   export let delay: number = 300;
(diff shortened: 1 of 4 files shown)


lovincyrus (approved):
ack

PR #7037, opened 2025-04-01
Fix leaderboard bg gradients

lovincyrus:
Fixes https://www.notion.so/rilldata/Background-bar-charts-seem-broken-1c8ba33c8f5780f69e8bf5c93447a315 and https://www.notion.so/rilldata/Blue-bars-showing-inaccurately-1c8ba33c8f57817e9411f82cc8d696fc

Also reported in https://rilldata.slack.com/archives/C087E4NS7FH/p1743541702913749

1 measure
CleanShot 2025-04-01 at 16 51 58@2x

1+ measures
CleanShot 2025-04-01 at 16 52 28@2x

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-common/src/features/dashboards/leaderboard/Leaderboard.svelte
+++ b/web-common/src/features/dashboards/leaderboard/Leaderboard.svelte
@@ -363,7 +363,6 @@
           <LeaderboardRow
             {suppressTooltip}
             {tableWidth}
-            {dimensionColumnWidth}
             {isBeingCompared}
             {filterExcludeMode}
             {atLeastOneActive}
@@ -382,7 +381,6 @@
         <LeaderboardRow
           {suppressTooltip}
           {itemData}
-          {dimensionColumnWidth}
           {tableWidth}
           {dimensionName}
           {isBeingCompared}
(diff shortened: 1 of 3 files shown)


ericpgreen2 (approved):
Nice, this makes the multiple measures case way better!

djbarnwal (changes requested):
Looks like it's still broken when comparisons are "on"
Single measure
image

Multiple measure
image

With context columns, conceptually we treat the whole row as a single element on which the bars are placed based on the percentages.

lovincyrus:
> With context columns, conceptually we treat the whole row as a single element on which the bars are placed based on the percentages.

Synced offline with @djbarnwal 

> Yeah, basically, the percentage of total bar spans across the dimension name and context columns. So for example is pct is 100%, the dimension column and all the context columns would have blue background underneath it.
I think before your work, we only supported either comparisons or percent of total as context columns and not both. To calculate the widths to be turned blue, we first calculated the total length to be colored based on leaderboard width. Once we had that, we would color the underlying cells one by one until that width was reached.

lovincyrus:
Latest:

Multiple measures, time comparison is off
CleanShot 2025-04-02 at 12 26 54@2x

Multiple measures, time comparison is on
CleanShot 2025-04-02 at 12 27 21@2x

PR #7039, opened 2025-04-02
fix: watermark field is converted to string when using `transform_sql`

k-anshul:
Fails the checks here : https://github.com/rilldata/rill/blob/12e9f976aa7e3272132823d6eda831b8c7e8cdae/runtime/reconcilers/model.go#L756-L759

Also it will be better to use duckdb to scan JSON file since we can control the datatype via `transform_sql`.

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!


begelundmuller (approved):

runtime/resolvers/glob.go
```diff
@@ -372,26 +376,27 @@ func (r *globResolver) writeTempNDJSONFile(rows []map[string]any) (string, error
...
+	}
+	defer rows.Close()
+	for rows.Next() {
+		row := make(map[string]any)
+		if err := rows.MapScan(row); err != nil {
 			return nil, err
 		}
-		rows = append(rows, row)
+		data = append(data, row)
 	}
```
begelundmuller, line 398: Could this be moved to around line 341, so it just reads the rows directly instead of doing `COPY` to JSON and calling `readNDJSONFile`?  That removes the need for `read_ndjson_auto` to infer the schema for the output of `transform_sql`.

Issue #1155, opened 2022-10-31
[stub] Models nav menu tool tip has click to open - Sources and Dashboard do not


magorlick:
![Image](https://user-images.githubusercontent.com/5587788/184549456-b122de25-1239-4f5f-afb8-dc6443a251e7.png)

![Image](https://user-images.githubusercontent.com/5587788/184549426-b3888299-4272-463a-8039-bebe2dccc489.png)

![Image](https://user-images.githubusercontent.com/5587788/184549443-032b225f-f15e-4990-963c-9e823b6ac8db.png)

hamilton:
I would be in favor of removing this "click to navigate" tooltip entirely. It's usually pretty self-explanatory that clicking does something.

magorlick:
I agree let’s remove it

PR #7040, opened 2025-04-02
user token expired warning

NamanMahor:
Closes [#1155](https://github.com/rilldata/rill-private-issues/issues/1155)

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!


cli/pkg/dotrill/dotrill.go
```diff
@@ -113,6 +118,16 @@ func SetDefaultOrg(orgName string) error {
 	return Set(ConfigFilename, DefaultOrgConfigKey, orgName)
 }
 
+// GetDefaultOrg loads the default org
```
k-anshul, line 121: The comment does not confirm to `GoDoc conventions`.

cli/cmd/root.go
```diff
@@ -107,7 +110,26 @@ func runCmd(ctx context.Context, ver cmdutil.Version) error {
 		ch.PrintfWarn("Could not parse representing user email\n\n")
 	}
 	if representingUser != "" {
-		ch.PrintfWarn("Warning: Running action as %q\n\n", representingUser)
+		tokenExpiryStr, err := dotrill.GetAccessTokenExpiry()
+		if err != nil {
+			ch.PrintfWarn("Could not parse token expiry\n\n")
```
k-anshul, line 115: Instead of erroring may be directly run unassume ? Just a suggestion.

cli/cmd/sudo/user/assume.go
```diff
@@ -18,6 +21,22 @@ func AssumeCmd(ch *cmdutil.Helper) *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			ctx := cmd.Context()
 
+			representingUser, err := dotrill.GetRepresentingUser()
+			if err != nil {
+				ch.PrintfWarn("Could not parse representing user email\n\n")
+			}
+			if representingUser != "" {
+				ch.Printf("You are already a assumed user %q. so unassuming it first.\n", representingUser)
```
k-anshul, line 29: Suggestion: From the issue it seems like we don't want assuming msg to be printed. Can we just print, `restoring user state` or a similar generic msg.

cli/pkg/dotrill/dotrill.go
NamanMahor: done

cli/cmd/sudo/user/assume.go
NamanMahor: remove the logs

cli/cmd/root.go
```diff
@@ -107,7 +110,25 @@ func runCmd(ctx context.Context, ver cmdutil.Version) error {
...
+		} else if tokenExpiryStr != "" {
+			expiryTime, err := strconv.ParseInt(tokenExpiryStr, 10, 64)
+			if err != nil {
+				ch.PrintfWarn("Could not parse token expiry time\n\n")
+			} else if time.Now().Unix() > expiryTime {
```
begelundmuller, line 120: 1. Make `GetAccessTokenExpiry` and `SetAccessTokenExpiry` accept a `time.Time`, and then do the serialization/deserialization inside those methods. That hides the complexity from upstream callers like here.
2. Use `time.RFC3339Nano` format instead of Unix time when serializing (it's easier to understand if looking directly at the files, and performance doesn't matter here)

cli/cmd/sudo/user/assume.go
```diff
@@ -18,6 +21,22 @@ func AssumeCmd(ch *cmdutil.Helper) *cobra.Command {
 		RunE: func(cmd *cobra.Command, args []string) error {
 			ctx := cmd.Context()
 
+			representingUser, err := dotrill.GetRepresentingUser()
+			if err != nil {
+				ch.PrintfWarn("Could not parse representing user email\n\n")
```
begelundmuller, line 26: Include the `err` in the error log for easier debugging
```suggestion
				ch.PrintfWarn("Could not parse representing user email: %v\n\n", err)
```

cli/cmd/sudo/user/unassume.go
```diff
@@ -17,59 +17,94 @@ func UnassumeCmd(ch *cmdutil.Helper) *cobra.Command {
...
+		return err
+	}
+	_, err = client.RevokeCurrentAuthToken(ctx, &adminv1.RevokeCurrentAuthTokenRequest{})
+	if err != nil {
+		ch.Printf("Failed to revoke token (it may have expired). Clearing local token anyway.\n")
+	}
+	return RestoreOriginalUserState(ctx, ch)
+}
+
+func RestoreOriginalUserState(ctx context.Context, ch *cmdutil.Helper) error {
```
begelundmuller, line 38: Why do these need to be separate functions? I think it can just be one combined function (where `UnassumeUser` only calls `RevokeCurrentAuthToken` if the token is not expired).

cli/pkg/dotrill/dotrill.go
```diff
@@ -21,13 +21,18 @@ const (
...
+	DefaultAdminURLConfigKey  = "api_url"
+	AnalyticsEnabledConfigKey = "analytics_enabled"
+	AccessTokenCredentialsKey = "token"
+	//nolint:gosec // This is a key name, not a credential
+	AccessTokenExpiryKey           = "token_expiry"
 	InstallIDStateKey              = "install_id"
 	RepresentingUserCredentialsKey = "representing_user"
 	BackupTokenCredentialsKey      = "backup_token"
+	//nolint:gosec // This is a key name, not a credential
+	BackupAccessTokenExpiryKey     = "backup_token_expiry"
```
begelundmuller, line 35: The user's main token never expires, so no need to save that. It just needs to track the expiry of the representing token (which has a TTL).

cli/pkg/dotrill/dotrill.go
```diff
@@ -21,13 +21,18 @@ const (
...
-	DefaultAdminURLConfigKey       = "api_url"
-	AnalyticsEnabledConfigKey      = "analytics_enabled"
-	AccessTokenCredentialsKey      = "token"
+	DefaultOrgConfigKey       = "org"
+	BackupDefaultOrgConfigKey = "backup_org"
+	DefaultAdminURLConfigKey  = "api_url"
+	AnalyticsEnabledConfigKey = "analytics_enabled"
+	AccessTokenCredentialsKey = "token"
+	//nolint:gosec // This is a key name, not a credential
+	AccessTokenExpiryKey           = "token_expiry"
```
begelundmuller, line 30: Notice all the keys have format `<name><file>Key`. So consider naming it `AccessTokenExpiryCredentialsKey` (since it's stored in the `CredentialsFilename`).

begelundmuller (changes requested):

cli/cmd/root.go
NamanMahor: sure.

cli/pkg/dotrill/dotrill.go
NamanMahor: done

cli/pkg/dotrill/dotrill.go
NamanMahor: done

cli/cmd/sudo/user/unassume.go
NamanMahor: done

NamanMahor:
closing it as rebased it :(

PR #7041, opened 2025-04-02
Simplify session storage to work off of url params

AdityaHegde:
As a precursor to merge url params per param instead of as state sources as a whole this PR moves to using url params in session storage. This will simplify it by storing in the format of url params already.

**Checklist:**
- [x] Covered by tests
- [ ] I'm proud of this work!


web-common/src/features/dashboards/state-managers/loaders/get-explore-state-from-session-storage.ts
ericpgreen2, line 1: I'd expect the "get state" (this file) and the "set state" (`explore-web-view-store.ts`?) to live in the same file. Can we move this to the `explore-web-view-store` file?

web-common/src/features/dashboards/state-managers/loaders/get-explore-state-from-session-storage.ts
```diff
@@ -0,0 +1,63 @@
...
+  const viewFromUrl = searchParams.get(ExploreStateURLParams.WebView) as
+    | ExploreUrlWebView
+    | undefined;
+  const exploreStateFromSessionStorage = getPartialExploreStateForWebView(
+    exploreName,
+    storageNamespacePrefix,
+    viewFromUrl ?? ExploreUrlWebView.Explore,
+    metricsViewSpec,
+    exploreSpec,
+  );
```
ericpgreen2, line 46: Maybe this function is an extra abstraction that could be removed?

web-common/src/features/dashboards/state-managers/loaders/get-explore-state-from-session-storage.ts
```diff
@@ -0,0 +1,63 @@
...
+  if (viewFromUrl === ExploreUrlWebView.TimeDimension) {
+    exploreStateFromSessionStorage.tdd ??= {
+      expandedMeasureName: "",
+      chartType: TDDChart.DEFAULT,
+      pinIndex: -1,
+    };
+    exploreStateFromSessionStorage.tdd.expandedMeasureName = searchParams.get(
+      ExploreStateURLParams.ExpandedMeasure,
+    ) as string;
+  }
```
ericpgreen2, line 60: Why do we have special handling for the `TDD` view, but not for the other views?

web-common/src/features/dashboards/state-managers/loaders/get-explore-state-from-session-storage.ts
AdityaHegde: Hmmm true. Earlier these were used separately, now we can merge them.

web-common/src/features/dashboards/state-managers/loaders/get-explore-state-from-session-storage.ts
AdityaHegde: It is because TDD has a parameter - expanded measure. Added some comments to clarify.

ericpgreen2 (changes requested):
Small bug. To reproduce:
1. Navigate from a stateful Explore to Pivot
2. Click on "Explore" to navigate back to your stateful Explore
3. Click on the "Explore" tab again. The URL parameters are removed.

AdityaHegde:
@ericpgreen2 That issue is on prod right now and unrelated to this change. It is most likely in the sync code. Lets not couple it with this PR

ericpgreen2 (approved):

PR #6930, opened 2025-03-19
chore: Update Go actions to use go.mod for versioning and adjust depth

grahamplata:
Investigate CI issues for Go resolves: https://github.com/rilldata/rill-private-issues/issues/1389

INSERT DESCRIPTION HERE

**Checklist:**
- [ ] Covered by tests
- [ ] Ran it and it works as intended
- [ ] Reviewed the diff before requesting a review
- [ ] Checked for unhandled edge cases
- [ ] Linked the issues it closes
- [ ] Checked if the docs need to be updated
- [ ] Intend to cherry-pick into the release branch
- [ ] I'm proud of this work!

PR #7049, opened 2025-04-02
[ENG-373] - Consolidate gRPC and HTTP server

grahamplata:
Resolves: https://github.com/rilldata/rill-private-issues/issues/1341

> Deprecates the current dual port set up using grpc-gateway in favor of a single port solution using vanguard.

Part of: https://github.com/rilldata/rill-helm-charts/pull/205

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


runtime/pkg/graceful/http.go
```diff
@@ -33,29 +36,30 @@ func ServeHTTP(ctx context.Context, server *http.Server, options ServeOptions) e
 		return err
 	}
 
-	// Channel to signal server has stopped
-	serveErrCh := make(chan error)
-	// Start server in a goroutine
+	cctx, cancel := context.WithCancel(ctx)
+	var serveErr error
```
begelundmuller, line 40: Doesn't this code re-introduce the race condition fixed here? https://github.com/rilldata/rill/pull/6930

Bad merge?

runtime/server/repos.go
```diff
@@ -227,21 +320,23 @@ func (s *Server) UploadMultipartFile(w http.ResponseWriter, req *http.Request, p
...
 
-	err = s.runtime.PutFile(ctx, pathParams["instance_id"], pathParams["path"], f, true, false)
+	err = s.runtime.PutFile(ctx, instanceID, path, f, true, false)
 	if err != nil {
+		s.logger.Error("failed to write file during upload",
+			zap.String("instanceID", instanceID),
+			zap.String("path", path),
+			zap.Error(err))
 		http.Error(w, fmt.Sprintf("failed to write file: %s", err), http.StatusBadRequest)
 		return
```
begelundmuller, line 341: Not sure this is worth an error log. The most likely error type is request cancellation, which isn't worth an alert.

If you want to improve this, then maybe lean into the middleware that logs response errors by customizing the response status code based on the `err` type? (5xx errors lead to `error`-level logs.)

runtime/server/server.go
```diff
@@ -146,80 +162,37 @@ func (s *Server) ServeGRPC(ctx context.Context) error {
...
-	gwMux := gateway.NewServeMux(
-		gateway.WithErrorHandler(HTTPErrorHandler),
-		gateway.WithOutgoingHeaderMatcher(func(s string) (string, bool) {
-			// grpc gateway adds gateway.MetadataHeaderPrefix to all outgoing headers
-			// we want to skip that for `x-trace-id` set in response
-			if s == observability.TracingHeader {
-				return s, true
-			}
-			// default matcher logic
-			return fmt.Sprintf("%s%s", gateway.MetadataHeaderPrefix, s), true
```
begelundmuller, line 182: Does the new implementation propagate `x-trace-id` correctly? Or do we need to restore this middleware somehow?

runtime/server/server.go
```diff
@@ -272,16 +245,6 @@ func (s *Server) HTTPHandler(ctx context.Context, registerAdditionalHandlers fun
 	return handler, nil
 }
 
-// HTTPErrorHandler wraps gateway.DefaultHTTPErrorHandler to map gRPC unknown errors (i.e. errors without an explicit
-// code) to HTTP status code 400 instead of 500.
-func HTTPErrorHandler(ctx context.Context, mux *gateway.ServeMux, marshaler gateway.Marshaler, w http.ResponseWriter, r *http.Request, err error) {
-	s := status.Convert(err)
-	if s.Code() == codes.Unknown {
-		err = &gateway.HTTPStatusError{HTTPStatus: http.StatusBadRequest, Err: err}
-	}
```
begelundmuller, line 281: Some comment here as in the `admin` service

runtime/server/server.go
```diff
@@ -146,80 +162,37 @@ func (s *Server) ServeGRPC(ctx context.Context) error {
...
+		httpMux.Handle("/v1/", transcoder)
 		registerAdditionalHandlers(httpMux)
 	}
 
-	// Add gRPC-gateway on httpMux
-	httpMux.Handle("/v1/", gwMux)
+	httpMux.Handle("/", transcoder)
+
+	// Add HTTP handler for
+	httpMux.Handle("/v1/instances/{instance_id}/files/watch", auth.HTTPMiddleware(s.aud, http.HandlerFunc(s.WatchFilesHandler)))
```
begelundmuller, line 182: nit: comment is broken

runtime/server/repos.go
```diff
@@ -52,6 +54,94 @@ func (s *Server) ListFiles(ctx context.Context, req *runtimev1.ListFilesRequest)
 	return &runtimev1.ListFilesResponse{Files: entries}, nil
 }
 
+// WatchFilesHandler is a HTTP handler for watching files like WatchFiles.
+func (s *Server) WatchFilesHandler(w http.ResponseWriter, req *http.Request) {
```
begelundmuller, line 58: 1. Can you add a comment about why this is needed instead of `WatchFiles`? (I.e. that vanguard doesn't currently map streaming RPCs to SSE, so we register this handler manually override the behavior.)
2. Instead of replicating the implementation of `WatchFiles`, would it be possible to wrap `WatchFiles` instead? You'll need to create a shim for `ss runtimev1.RuntimeService_WatchFilesServer`, but that might be as simple as something like this:
```go
type watchFilesServerShim struct {
	r *http.Request
	s *sse.Server
}

func (ss *watchFilesServerShim) Context() context.Context {
	return ss.r.Context()
}

func (ss *watchFilesServerShim) Send(e *runtimev1.WatchFilesResponse) error {
	data, err := protojson.Marshal(e)
	if err != nil {
		return err
	}

	ss.s.Publish("files", &sse.Event{Data: data})
	return nil
}

func (ss *watchFilesServerShim) SetHeader(metadata.MD) error {
}

func (ss *watchFilesServerShim) SendHeader(metadata.MD) error {
}

func (ss *watchFilesServerShim) SetTrailer(metadata.MD) {}

func (ss *watchFilesServerShim) SendMsg(m any) error {
}

func (ss *watchFilesServerShim) RecvMsg(m any) error {
}
```

runtime/server/controller.go
```diff
@@ -87,6 +90,102 @@ func (s *Server) ListResources(ctx context.Context, req *runtimev1.ListResources
 	return &runtimev1.ListResourcesResponse{Resources: rs}, nil
 }
 
+// WatchResourcesHandler implements an HTTP handler for runtimev1.RuntimeServiceServer
+func (s *Server) WatchResourcesHandler(w http.ResponseWriter, r *http.Request) {
```
begelundmuller, line 94: Same comments here as for `WatchFilesHandler`

begelundmuller (changes requested):

admin/server/server.go
```diff
@@ -399,16 +372,6 @@ func (s *Server) jwtAttributesForUser(ctx context.Context, userID, orgID string,
...
 }
 
-// httpErrorHandler wraps gateway.DefaultHTTPErrorHandler to map gRPC unknown errors (i.e. errors without an explicit
-// code) to HTTP status code 400 instead of 500.
-func httpErrorHandler(ctx context.Context, mux *gateway.ServeMux, marshaler gateway.Marshaler, w http.ResponseWriter, r *http.Request, err error) {
-	s := status.Convert(err)
-	if s.Code() == codes.Unknown {
-		err = &gateway.HTTPStatusError{HTTPStatus: http.StatusBadRequest, Err: err}
-	}
-	gateway.DefaultHTTPErrorHandler(ctx, mux, marshaler, w, r, err)
```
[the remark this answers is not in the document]
grahamplata: I believe Vanguard handles error translation automatically as part of its transcoder. It converts gRPC status codes to appropriate HTTP status codes https://github.com/connectrpc/vanguard-go/blob/main/protocol_http.go

runtime/server/server.go
grahamplata: It looks like it is successfully propagating the initial header via `LoggingUnaryServerInterceptor`

admin/server/server.go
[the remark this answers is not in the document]
begelundmuller: Thanks for the link. So the problem is they map `Unknown` to a 5xx error by default: https://github.com/connectrpc/vanguard-go/blob/4972a6c703d2c200cbe39cbc459d86847962df7c/protocol_http.go#L40C1-L40C46.

But we have a lot of code that returns Go errors as-is (which gets marked as unknown) and most of those are not internal errors. So we don't want 5xx status codes for those.

If Vanguard does not allow you to override it, consider changing the `mapGRPCError` middleware to rewrite unknown non-gRPC errors to `InvalidArgument` instead of `Unknown`.

runtime/server/server.go
begelundmuller: See response above

runtime/server/controller.go
```diff
@@ -116,72 +111,23 @@ func (s *Server) WatchResourcesHandler(w http.ResponseWriter, r *http.Request) {
...
-	}()
+	// Use the existing WatchResources implementation
+	err := s.WatchResources(&runtimev1.WatchResourcesRequest{
+		InstanceId: instanceID,
+		Kind:       kind,
+		Replay:     replay,
+	}, shim)
+	if err != nil {
+		s.logger.Info("watch resources ended with error", zap.Error(err))
+		http.Error(w, "internal server error", http.StatusInternalServerError)
```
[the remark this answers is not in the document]
grahamplata: That was a miss on my part. Moving it to the same pattern as WatchFilesHandler by using a goroutine

admin/server/server.go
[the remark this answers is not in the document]
grahamplata: Understood, that will likely be the case

cli/cmd/admin/start.go
```diff
@@ -58,11 +58,9 @@ type Config struct {
 	LogLevel                  zapcore.Level          `default:"info" split_words:"true"`
 	MetricsExporter           observability.Exporter `default:"prometheus" split_words:"true"`
 	TracesExporter            observability.Exporter `default:"" split_words:"true"`
-	HTTPPort                  int                    `default:"8080" split_words:"true"`
-	GRPCPort                  int                    `default:"8080" split_words:"true"`
+	Port                      int                    `default:"8080" split_words:"true"`
```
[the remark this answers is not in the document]
grahamplata: https://github.com/rilldata/rill-helm-charts/pull/205

admin/testadmin/testadmin.go
```diff
@@ -74,10 +73,9 @@ func New(t *testing.T) *Fixture {
 	require.NoError(t, err)
 
 	// Ports and external URLs
-	httpPort := findPort(t)
-	grpcPort := findPort(t)
-	externalURL := fmt.Sprintf("http://localhost:%d", grpcPort)
-	externalHTTPURL := fmt.Sprintf("http://localhost:%d", httpPort)
+	port := findPort(t)
+	externalURL := fmt.Sprintf("http://localhost:%d", port)
+	externalHTTPURL := fmt.Sprintf("http://localhost:%d", port)
```
begelundmuller, line 78: Can remove `externalHTTPURL`

go.mod
```diff
@@ -168,6 +169,7 @@ require (
 	cloud.google.com/go/auth v0.16.0 // indirect
 	cloud.google.com/go/auth/oauth2adapt v0.2.8 // indirect
 	cloud.google.com/go/monitoring v1.24.0 // indirect
+	connectrpc.com/vanguard v0.3.0
```
begelundmuller, line 172: Usually direct dependencies are separate from indirect ones. Try running `go mod tidy` or moving up manually

go.mod
```diff
@@ -413,7 +415,7 @@ require (
 	go.uber.org/mock v0.5.0 // indirect
 	golang.org/x/crypto v0.37.0 // indirect
 	golang.org/x/mod v0.24.0 // indirect
-	golang.org/x/net v0.39.0 // indirect
+	golang.org/x/net v0.39.0
```
begelundmuller, line 418: What made this a direct dependency?

runtime/server/controller.go
```diff
@@ -87,6 +91,47 @@ func (s *Server) ListResources(ctx context.Context, req *runtimev1.ListResources
...
+	// Use the existing WatchResources implementation in a goroutine
+	go func() {
+		err := s.WatchResources(&runtimev1.WatchResourcesRequest{
+			InstanceId: instanceID,
+			Kind:       kind,
+			Replay:     replay,
+		}, shim)
+		if err != nil {
+			s.logger.Error("watch resources ended with error", zap.Error(err))
+		}
```
begelundmuller, line 129: Don't remember if this is handled elsewhere, but you probably want to avoid error logs for `ctx` cancellations (the UI disconnects often, and we don't want ominous error logs in the console on local for that).

Related – did you do a thorough experience pass on the local experience as well (not cloud experience running on local)? E.g. `make cli` and `./rill start dev-project`?

runtime/server/repos.go
```diff
@@ -52,6 +55,57 @@ func (s *Server) ListFiles(ctx context.Context, req *runtimev1.ListFilesRequest)
...
+		watchReq := &runtimev1.WatchFilesRequest{
+			InstanceId: instanceID,
+			Replay:     replay,
+		}
+
+		// Call the existing WatchFiles implementation with our shim
+		err := s.WatchFiles(watchReq, shim)
+		if err != nil && !errors.Is(err, context.Canceled) {
+			s.logger.Info("watch error", zap.Error(err))
+		}
```
begelundmuller, line 102: Related to previous comment – here there's a `ctx` cancel check

runtime/server/server.go
```diff
@@ -266,22 +238,12 @@ func (s *Server) HTTPHandler(ctx context.Context, registerAdditionalHandlers fun
...
 
 	return handler, nil
 }
 
-// HTTPErrorHandler wraps gateway.DefaultHTTPErrorHandler to map gRPC unknown errors (i.e. errors without an explicit
-// code) to HTTP status code 400 instead of 500.
-func HTTPErrorHandler(ctx context.Context, mux *gateway.ServeMux, marshaler gateway.Marshaler, w http.ResponseWriter, r *http.Request, err error) {
-	s := status.Convert(err)
-	if s.Code() == codes.Unknown {
-		err = &gateway.HTTPStatusError{HTTPStatus: http.StatusBadRequest, Err: err}
```
begelundmuller, line 280: I saw you fixed this in `mapGRPCError` in `admin/server`, but the same handling needs to be copied to `mapGRPCError` in this file (`runtime/server`).

I notice that in this file, it's missing this check, which `admin/server` has. You'll probably need to add that here as well:
```
	if _, ok := status.FromError(err); ok {
		return err
	}
```

begelundmuller (changes requested):

runtime/server/controller.go
grahamplata:
```
go run ./cli devtool start local --refresh-dotenv=false
go run ./cli devtool start cloud --refresh-dotenv=false
```

For the most part

grahamplata: Validating this process now

grahamplata: I believe this is sorted now. The currently set cloud-e2e env is set outside of ci

runtime/server/server.go
```diff
@@ -272,6 +270,9 @@ func timeoutSelector(fullMethodName string) time.Duration {
 func errorMappingUnaryServerInterceptor() grpc.UnaryServerInterceptor {
 	return func(ctx context.Context, req interface{}, info *grpc.UnaryServerInfo, handler grpc.UnaryHandler) (interface{}, error) {
 		resp, err := handler(ctx, req)
+		if _, ok := status.FromError(err); ok {
+			return resp, err
+		}
 		return resp, mapGRPCError(err)
```
begelundmuller, line 276: nit: implement this check inside `mapGRPCError` – see how it's done in `admin/server/server.go`

runtime/server/server.go
```diff
@@ -280,6 +281,9 @@ func errorMappingUnaryServerInterceptor() grpc.UnaryServerInterceptor {
 func errorMappingStreamServerInterceptor() grpc.StreamServerInterceptor {
 	return func(srv interface{}, ss grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler) error {
 		err := handler(srv, ss)
+		if _, ok := status.FromError(err); ok {
+			return err
+		}
 		return mapGRPCError(err)
```
begelundmuller, line 287: Same comment here

runtime/server/server.go
```diff
@@ -168,52 +174,22 @@ func (s *Server) ServeHTTP(ctx context.Context, registerAdditionalHandlers func(
...
-	// One-off REST-only path for multipart file upload
-	// NOTE: It's local only and we should deprecate it in favor of a cloud-friendly alternative.
-	err = gwMux.HandlePath("POST", "/v1/instances/{instance_id}/files/upload/-/{path=**}", auth.GatewayMiddleware(s.aud, s.UploadMultipartFile))
-	if err != nil {
-		panic(err)
+	// Register the Vanguard handler for gRPC transcoding
+	if local {
+		httpMux.Handle("/v1/", transcoder)
+	} else {
+		httpMux.Handle("/", transcoder)
```
begelundmuller, line 184: Does this mean the gRPC endpoints won't route correctly on local? Won't that break e.g. `rill project refresh --local`?

Consider mounting the explicit gRPC paths on root instead. Like:
```go
httpMux.Handle("/v1/", transcoder)
httpMux.Handle("/rill.runtime.RuntimeService/", transcoder)
httpMux.Handle("/rill.runtime.QueryService/", transcoder)
httpMux.Handle("/rill.runtime.ConnectorService/", transcoder)
```

Btw, I'm not sure those are the correct paths, so please double check :)

begelundmuller (approved):

Issue #6909, opened 2025-03-18
Optimize Query time lookups for ClickHouse


nishantmonu51:
Our current approach to query-time lookups is suboptimal for ClickHouse, as it cannot effectively optimize filters even if the lookups is marked as injective. 

To enhance query performance, we need to refine our query generation logic.

PR #7051, opened 2025-04-02
optimize ch dicts

pjain1:
Fixes https://github.com/rilldata/rill/issues/6909

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


runtime/parser/parse_metrics_view.go
```diff
@@ -40,6 +40,11 @@ type MetricsViewYAML struct {
 		Ignore      bool   `yaml:"ignore"` // Deprecated
 		Unnest      bool
 		URI         string
+		Lookup      *struct {
```
begelundmuller, line 43: If you don't make this a pointer, the code gets simpler:
```suggestion
		Lookup      struct {
```

runtime/metricsview/query.go
```diff
@@ -186,6 +187,7 @@ type Condition struct {
 }
 
 type Subquery struct {
+	RawSQL    string      `mapstructure:"raw_sql"`
```
begelundmuller, line 190: Same comment as earlier – dangerous to expose `raw_sql` in the `Query` type where anyone with access can set it

runtime/metricsview/query.go
```diff
@@ -174,10 +174,11 @@ func (tr *TimeRange) IsZero() bool {
...
 type Expression struct {
-	Name      string     `mapstructure:"name"`
-	Value     any        `mapstructure:"val"`
-	Condition *Condition `mapstructure:"cond"`
-	Subquery  *Subquery  `mapstructure:"subquery"`
+	Name       string     `mapstructure:"name"`
+	Value      any        `mapstructure:"val"`
+	Condition  *Condition `mapstructure:"cond"`
+	Subquery   *Subquery  `mapstructure:"subquery"`
+	Identifier string     `mapstructure:"identifier"`
```
begelundmuller, line 181: Similar comment – I think it's unsafe even to expose `Identifier` in the metrics query API because it can leak underlying fields that are not exposed through dimension/measure expressions.

runtime/metricsview/executor_rewrite_clickhouse_dicts.go
```diff
@@ -0,0 +1,91 @@
...
+
+import (
+	"fmt"
+
+	"github.com/rilldata/rill/runtime/drivers"
+)
+
+// Rewrites filter on dictionary columns to use the key column instead of the value column. For example, if the dictionary value column is "country" and key column is "country_key",
+// then the filter will be rewritten to country_key = 'US' instead of country = 'USA'.
+func (e *Executor) rewriteClickhouseDictFilters(qry *Query) {
```
begelundmuller, line 11: Since lookups are introduced as a general concept in the metrics view spec, I wonder if it should be handled directly inside `astexpr.go` when converting an expression to SQL? 

I think that would also avoid the need for exposing internal fields in the `Query` type (which is supposed to be user-facing, as opposed to `AST` which is intended to be internal).

runtime/drivers/olap.go
```diff
@@ -766,6 +769,15 @@ func (d Dialect) GetTimeExpr(t time.Time) (bool, string) {
 	}
 }
 
+func (d Dialect) DictGetExpr(lookupTable, lookupColumn, lookupKey string) string {
+	switch d {
+	case DialectClickHouse:
+		return fmt.Sprintf("dictGet('%s', '%s', %s)", lookupTable, lookupColumn, lookupKey)
+	default:
+		return ""
```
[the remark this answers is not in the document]
pjain1: Probably yes for Druid we do not need `LookupValueColumn` we just need the lookup name and column that has values to be looked up. However we are doing this mainly for rewriting filter but Druid already does reverse lookup inherently so defining lookup expression directly in dimension expression should be ok. For Duckdb I think we should just avoid going into this complexity as it should not matter much at that scale.

runtime/metricsview/executor_rewrite_clickhouse_dicts.go
pjain1: I did it earlier as I was doing the group by optimization as well so thought it would be good to define both as standalone rewrite that be configured on/off.

runtime/drivers/olap.go
```diff
@@ -741,6 +744,15 @@ func (d Dialect) GetTimeExpr(t time.Time) (bool, string) {
 	}
 }
 
+func (d Dialect) DictGetExpr(lookupTable, lookupColumn, lookupKey string) string {
```
begelundmuller, line 747: nit: Should the function name use "lookup" instead of "dict", e.g. something like `LookupExpression`?

runtime/drivers/olap.go
```diff
@@ -741,6 +744,15 @@ func (d Dialect) GetTimeExpr(t time.Time) (bool, string) {
...
+	switch d {
+	case DialectClickHouse:
+		return fmt.Sprintf("dictGet('%s', '%s', %s)", lookupTable, lookupColumn, lookupKey)
+	default:
```
begelundmuller, line 751: Can you maybe add a comment about why it's not handling Druid or DuckDB? Like you wrote in the other PR comment:
> Druid already does reverse lookup inherently so defining lookup expression directly in dimension expression should be ok. For Duckdb I think we should just avoid going into this complexity as it should not matter much at that scale.

runtime/drivers/olap.go
```diff
@@ -326,6 +326,9 @@ func (d Dialect) AutoUnnest(expr string) string {
 }
 
 func (d Dialect) MetricsViewDimensionExpression(dimension *runtimev1.MetricsViewSpec_Dimension) string {
+	if dimension.LookupTable != "" {
+		return d.DictGetExpr(dimension.LookupTable, dimension.LookupValueColumn, dimension.Column)
+	}
 	if dimension.Expression != "" {
 		return dimension.Expression
```
begelundmuller, line 333: 1. Can it also support `dimension.Expression` here in addition to `dimension.Column`? Sometimes people use `expression` instead of `column:`, like `expression: country`.
2. The `DictGetExpr` returns `""` for DuckDB, Druid and Pinot, which might give weird errors. Can you think of some safer handling, e.g. a simple sub-select or returning an error?

runtime/metricsview/astexpr.go
```diff
@@ -210,7 +224,12 @@ func (b *sqlExprBuilder) writeBinaryCondition(exprs []*Expression, op Operator)
 
 		// If not unnested, write the expression as-is
 		if !unnest {
-			return b.writeBinaryConditionInner(nil, right, leftExpr, op)
+			if lm, ok := b.lookupMeta[left.Name]; ok {
```
begelundmuller, line 227: Inside `b.sqlForName` above, it calls `b.ast.lookupDimension`, which returns the dimension object that already has the lookup info. 

So is it possible to avoid the `lookupMeta` member and either:
1. compute the lookup expression inside `sqlForName`, or
2. return the lookup info from `sqlForName`

I think that would also mirror how unnest is handled (i.e. returned from `sqlForName` instead of being a meta object on the `ast`).

runtime/metricsview/astexpr.go
```diff
@@ -210,7 +224,12 @@ func (b *sqlExprBuilder) writeBinaryCondition(exprs []*Expression, op Operator)
...
+				leftExpr = fmt.Sprintf("%s IN (SELECT %s FROM dictionary(%s) WHERE %s ", b.ast.dialect.EscapeIdentifier(lm.keyExpr), b.ast.dialect.EscapeIdentifier(lm.keyCol), b.ast.dialect.EscapeIdentifier(lm.table), b.ast.dialect.EscapeIdentifier(lm.valueCol))
```
begelundmuller, line 228: Since this is ClickHouse-specific syntax, it should be implemented generically in a function on `drivers.Dialect`

runtime/metricsview/astexpr.go
```diff
@@ -210,7 +224,12 @@ func (b *sqlExprBuilder) writeBinaryCondition(exprs []*Expression, op Operator)
 
 		// If not unnested, write the expression as-is
 		if !unnest {
```
begelundmuller, line 226: 1. The comment here is no longer correct – because it is not returned "as-is" for lookups
2. What if a lookup is used together with `unnest`? Is it possible? If not, can it return an error if it happens?

runtime/metricsview/astexpr.go
```diff
@@ -316,6 +339,10 @@ func (b *sqlExprBuilder) writeBinaryConditionInner(left, right *Expression, left
 
 	b.writeByte(')')
 
+	if leftOverrideNeedsClosure {
+		b.writeByte(')')
+	}
```
begelundmuller, line 344: Can this be added around line 229 instead of being pushed into these functions? Or can you think of a refactor that would enable the open and close parenthesis to be written together?

It's just that adding the closing parenthesis in another scope than the open parenthesis makes it very difficult to understand if the parentheses are closed correctly or not in all cases.

runtime/parser/parse_metrics_view.go
```diff
@@ -820,14 +823,28 @@ func (p *Parser) parseMetricsView(node *Node) error {
 			continue
 		}
 
+		// all dict fields should be defined or none
+		if dim.LookupTable == "" && dim.LookKeyColumn == "" && dim.LookValueColumn == "" { // nolint:revive // avoids checking all permutations
+			// do nothing
+		} else if dim.LookupTable != "" && dim.LookKeyColumn != "" && dim.LookValueColumn != "" {
+			if dim.Column == "" {
+				return fmt.Errorf("column is required if lookup fields are defined, containing values to be looked up")
+			}
```
begelundmuller, line 832: Could we try to support `expression`? I may be wrong, but I think the UI sometimes defaults to `expression:` instead of `column:` now

begelundmuller (changes requested):
Can you also add a test for lookups in ClickHouse in `runtime/resolvers/testdata`?

runtime/metricsview/astexpr.go
pjain1: sure that can be done, I did this so lookup info is looked up only once instead of multiple times but looking at it now again it seems not necessary

runtime/parser/parse_metrics_view.go
pjain1: https://github.com/rilldata/rill/pull/7051#discussion_r2053346496

runtime/drivers/olap.go
begelundmuller: I think it'd be good to support `expression` for consistency. If they use `dictGet` in the expression, they should also get a validation error right? (Due to nested `dictGet`)

pjain1:
> I think it'd be good to support expression for consistency. If they use dictGet in the expression, they should also get a validation error right? (Due to nested dictGet)

No its not a validation error because dictGet support sql expressions so it will fail on runtime. We can validate manually during parsing like 
```
                         if dim.Expression != "" && strings.Contains(dim.Expression, "dictGet") {
				return fmt.Errorf("dictGet expression and lookup fields cannot be used together")
			}
```
but this check is specific to CH, we cannot delegate to dialect here as its not present in the context which should be ok. Another question that arises is since user can write any expression or use column name in the expression we may not know when to escape identifier, we cannot even lookup key dimension as it may not present in the metrics view. But atleast there will be validation error when there is a column that needs escaping but used in expression, in that case user can just use column field.

begelundmuller (approved):

PR #7053, opened 2025-04-03
fix: unstable e2e cloud setup

AdityaHegde:
This is an attempt to fix unstable e2e OpenRTB project deploy. Earlier it was stuck at getting dashboard listing. Now after adding a refresh there it gets stuck at getting the last refreshed which is available after reconcile is finished.

1. Locally the resources do reconcile after a long time. This PR stabilised the setup locally.
2. On CI there is still an issue with `sqlite_scanner` which seems to be blocking this. Pre-loading duckdb extensions as a temp fix.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-admin/src/components/layout/ContentContainer.svelte
+++ b/web-admin/src/components/layout/ContentContainer.svelte
@@ -9,7 +9,7 @@
 >
   <section class="w-full flex flex-col gap-y-3" style:max-width="{maxWidth}px">
     {#if title && showTitle}
-      <h1 class="text-2xl font-semibold">
+      <h1 class="text-2xl font-semibold" aria-label="Container title">
         {title}
       </h1>
     {/if}
--- a/web-admin/tests/setup/setup.ts
+++ b/web-admin/tests/setup/setup.ts
@@ -190,6 +190,21 @@ setup.describe("global setup", () => {
     // ).toBeVisible(); // Billing banner
     // await expect(adminPage.getByText("Free trial")).toBeVisible(); // Billing status
 
+    // There is a scenario where page loads before runtime can identify what files are present.
+    // This leads to a case where we never refresh the resources list.
+    // TODO: find a solution to refetch in the app itself
+    await expect
+      .poll(
+        async () => {
+          await adminPage.reload();
+          console.log("RELOADING...");
+          const title = adminPage.getByLabel("Container title");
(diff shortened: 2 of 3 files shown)


web-admin/tests/setup/setup.ts
```diff
@@ -208,7 +222,11 @@ setup.describe("global setup", () => {
           });
           return listing.textContent();
         },
-        { intervals: Array(6).fill(5_000), timeout: 30_000 },
+        {
+          // Increased timeout for the 1st dashboard to make sure sources are reconciled.
+          intervals: [10_000, 10_000, 20_000, 20_000, 30_000, 30_000],
+          timeout: 120_000,
+        },
```
ericpgreen2, line 229: I'm nervous about this part. Are you sure this is needed? Given the test project is so small, it's a problem if the project can't be reconciled & ingested in 30s.

web-admin/src/components/layout/ContentContainer.svelte
```diff
@@ -9,7 +9,7 @@
 >
   <section class="w-full flex flex-col gap-y-3" style:max-width="{maxWidth}px">
     {#if title && showTitle}
-      <h1 class="text-2xl font-semibold">
+      <h1 class="text-2xl font-semibold" aria-label="Container title">
```
ericpgreen2, line 12: Typically aria-labels are only applied on interactive elements. To access this element in Playwright, what about
```
await page.getByRole('heading', { level: 1, name: '<container-title>' })
```

web-admin/tests/setup/setup.ts
AdityaHegde: It times out locally for me. IMO 2 mins is not a long time considering tests take about 10mins total. Other tests dont run if setup fails so I think it is a good compromise to have.

ericpgreen2: Alright, let's try it

ericpgreen2 (approved):

web-admin/src/components/layout/ContentContainer.svelte
AdityaHegde: Hmmm I tried a bit to get this to work. But it always timed out for me. Lets not spend too much time here, we can get back to this some other time.

PR #7061, opened 2025-04-03
design: Polish the table toolbars

ericpgreen2:
This PR:
- Adds a common `toolbar` button type
- Adds icons to toolbar buttons
- Moves the "Flat" vs "Nested" pivot table button to the Pivot toolbar
- Moves common toolbar code into a `features/dashboards/toolbars` directory
- Improves the feel of the Search toolbar button
- Fixes the flickering of the TDD toolbar when adding a dimension filter

Dimension Table toolbar:
image

Time Dimension Detail toolbar:
image

Pivot toolbar:
image

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!


lovincyrus (changes requested):
The icons are looking good! A couple of notes:

1. There seems to be an unnecessary height in the toolbar. Can remove the h-9.

CleanShot 2025-04-16 at 22 49 18@2x

2. Could be out of scope, now that the toolbar buttons' hover styles are very distinctive, you can see the triple dot icon target area is relatively smaller.

CleanShot 2025-04-16 at 22 51 17@2x

ericpgreen2:
> 1. There seems to be an unnecessary height in the toolbar. Can remove the h-9.

I removed the `h-9` and I also decreased the Dimension Table Header's fixed height from 50px to 40px. This arbitrary fixed height feels wrong, but removing it entirely causes the toolbar+table to feel crowded. It's out-of-scope for now, but maybe we could give the Dimension Table a defined border in conjunction with a tighter toolbar.

> 2. Could be out of scope, now that the toolbar buttons' hover styles are very distinctive, you can see the triple dot icon target area is relatively smaller.

Agree. I removed the `compact` prop on that button.

lovincyrus (approved):
lgtm

PR #7071, opened 2025-04-04
Cleanup attached db

k-anshul:
closes https://github.com/rilldata/rill/issues/6998

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/runtime/pkg/rduckdb/db.go
+++ b/runtime/pkg/rduckdb/db.go
@@ -786,8 +786,12 @@ func (d *db) acquireWriteConn(ctx context.Context, dsn, table string, initQuerie
 		return rows.Err()
 	}
 
-	release := func() error {
-		err := dropViews()
+	release := func() (err error) {
+		defer func() {
+			// close the connection and db handle
+			err = errors.Join(err, conn.Close(), db.Close())
+		}()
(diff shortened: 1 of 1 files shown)


runtime/pkg/rduckdb/db.go
```diff
@@ -739,11 +739,69 @@ func (d *db) acquireWriteConn(ctx context.Context, dsn, table string, initQuerie
...
+		var name string
+		for rows.Next() {
+			err = rows.Scan(&name)
+			if err != nil {
+				return err
+			}
+			_, err = conn.ExecContext(ctx, "DROP VIEW IF EXISTS "+safeSQLName(name))
+			if err != nil {
+				return err
+			}
```
begelundmuller, line 762: Is this safe? I thought a `conn` is locked (so it can't be used) until the `rows` from a previous query on it are closed, which in this case would lead to a deadlock.

k-anshul: Yeah now when I think about it, yes it is not safe. I am not sure how it worked though. To be safe I will collect results first and drop later.

begelundmuller (approved):

PR #7074, opened 2025-04-04
Remove deprecated metrics view fields

begelundmuller:


web-common/src/features/alerts/extract-alert-form-values.ts
```diff
@@ -45,7 +45,7 @@ export function extractAlertFormValues(
     queryArgs.dimensions as V1MetricsViewAggregationDimension[];
 
   const timeRange = (queryArgs.timeRange as V1TimeRange) ?? {
-    isoDuration: metricsViewSpec.defaultTimeRange ?? TimeRangePreset.ALL_TIME,
+    isoDuration: TimeRangePreset.ALL_TIME,
```
begelundmuller, line 48: @AdityaHegde Does this look okay to you? The `defaultTimeRange` on the metrics view is deprecated anyway in favour of the explore spec field.

AdityaHegde: Hmmm it was a fallback anyway, so it should be fine to remove it.

begelundmuller: Great, thought so!

AdityaHegde (approved):

PR #7080, opened 2025-04-06
Enable bigquery connector to query google sheet backed external tables

### LINKED ISSUE #7070: Bigquery Driver Unable to query tables that are backed through google sheets
Google bigquery has the ability to create external tables that reference google sheets.
This is useful when data is inputted manually into sheets and we'd like to query it from bigquery.

Image

However when we query this from Rill we get the following error:

    BigQuery: Permission denied while getting Drive credentials

This issue is well documented in other systems as well:

    https://discourse.metabase.com/t/cant-access-a-bigquery-table-connected-to-gsheet/16920/8
    https://github.com/googleapis/google-auth-library-python/issues/1204

The solution is that the client needs to pass in the scope: 'https://www.googleapis.com/auth/drive' when connecting to bigquery.
I've tested this locally and confirmed it does indeed fail currently and work when this scope is given (fill in the ...)

```
package bigquery

import (
	"context"
	"encoding/json"
	"fmt"
	"os"
	"testing"

	"cloud.google.com/go/bigquery"
	"github.com/stretchr/testify/require"
	"golang.org/x/oauth2/google"
	"google.golang.org/api/iterator"
	"google.golang.org/api/option"
)

func TestBigQuerySelect(t *testing.T) {
	ctx := context.Background()

	// Read credentials file
	credsFile := "..."
	credsJSON, err := os.ReadFile(credsFile)
	require.NoError(t, err)

	// Get credentials with both Cloud Platform and Drive scopes
	creds, err := google.CredentialsFromJSON(ctx, credsJSON,
		"https://www.googleapis.com/auth/cloud-platform",
		"https://www.googleapis.com/auth/drive")
	require.NoError(t, err)

	// Create BigQuery client with the credentials
	client, err := bigquery.NewClient(ctx, "...", option.WithCredentials(creds))
	require.NoError(t, err)
	defer client.Close()

	// Run query
	q := client.Query("SELECT * FROM `...` LIMIT 10")
	it, err := q.Read(ctx)
	require.NoError(t, err)

	// Print schema
	fmt.Println("\nSchema:")
	for _, field := range it.Schema {
		fmt.Printf("%s: %s\n", field.Name, field.Type)
	}

	// Print results
	fmt.Println("\nResults:")
	for {
		var row map[string]bigquery.Value
		err := it.Next(&row)
		if err == iterator.Done {
			break
		}
		require.NoError(t, err)

		// Pretty print row as JSON
		jsonData, err := json.MarshalIndent(row, "", "  ")
		require.NoError(t, err)
		fmt.Printf("\nRow: %s\n", string(jsonData))
	}
}
```

A potential fix in rill itself would be, but should be tested end to end:

```
From e9021efbb31e06d8d5472f03f0f68a0a5c2790af Mon Sep 17 00:00:00 2001
From: Anil Vaitla <avaitla16@gmail.com>
Date: Fri, 4 Apr 2025 00:59:34 -0500
Subject: [PATCH] Bigquery Scopes

---
 runtime/drivers/bigquery/bigquery.go  | 11 ++++++++++-
 runtime/drivers/bigquery/warehouse.go |  1 +
 2 files changed, 11 insertions(+), 1 deletion(-)

diff --git a/runtime/drivers/bigquery/bigquery.go b/runtime/drivers/bigquery/bigquery.go
index f81afb318..0c9767136 100644
--- a/runtime/drivers/bigquery/bigquery.go
+++ b/runtime/drivers/bigquery/bigquery.go
@@ -15,6 +15,14 @@ import (
 	"google.golang.org/api/option"
 )
 
+var (
+	// Required scopes for BigQuery operations
+	requiredScopes = []string{
+		"https://www.googleapis.com/auth/cloud-platform",
+		"https://www.googleapis.com/auth/drive",
+	}
+)
+
 func init() {
 	drivers.Register("bigquery", driver{})
 	drivers.RegisterAsConnector("bigquery", driver{})
@@ -238,7 +246,8 @@ func parseSourceProperties(props map[string]any) (*sourceProperties, error) {
 }
 
 func (c *Connection) clientOption(ctx context.Context) ([]option.ClientOption, error) {
-	creds, err := gcputil.Credentials(ctx, c.config.SecretJSON, c.config.AllowHostAccess)
+	// Create credentials with required scopes
+	creds, err := gcputil.Credentials(ctx, c.config.SecretJSON, c.config.AllowHostAccess, requiredScopes...)
 	if err != nil {
 		return nil, err
 	}
diff --git a/runtime/drivers/bigquery/warehouse.go b/runtime/drivers/bigquery/warehouse.go
index 43d9d5beb..1db7aa048 100644
--- a/runtime/drivers/bigquery/warehouse.go
+++ b/runtime/drivers/bigquery/warehouse.go
@@ -163,6 +163,7 @@ func (c *Connection) QueryAsFiles(ctx context.Context, props map[string]any) (dr
 }
 
 func createClient(ctx context.Context, projectID string, opts []option.ClientOption) (*bigquery.Client, error) {
+	opts = append(opts, option.WithScopes("https://www.googleapis.com/auth/bigquery", "https://www.googleapis.com/auth/bigquerystorage"))
 	client, err := bigquery.NewClient(ctx, projectID, opts...)
 	if err != nil {
 		if strings.Contains(err.Error(), "unable to detect projectID") {
-- 
2.39.2 (Apple Git-143)

```

NamanMahor:
Closes #7070 

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!


k-anshul (approved):

runtime/drivers/bigquery/bigquery.go
```diff
@@ -238,7 +238,11 @@ func parseSourceProperties(props map[string]any) (*sourceProperties, error) {
 }
 
 func (c *Connection) clientOption(ctx context.Context) ([]option.ClientOption, error) {
-	creds, err := gcputil.Credentials(ctx, c.config.SecretJSON, c.config.AllowHostAccess)
+	scopes := []string{
+		"https://www.googleapis.com/auth/cloud-platform",
+		"https://www.googleapis.com/auth/drive.readonly",
```
begelundmuller, line 243: For non-drive users where the service account doesn't have Drive access, does this still work?

NamanMahor: yes.

begelundmuller: Perfect!

PR #7081, opened 2025-04-06
Mv rows export

pjain1:
Fixes https://github.com/rilldata/rill-private-issues/issues/1342

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


runtime/queries/metricsview_aggregation.go
```diff
@@ -227,6 +228,31 @@ func ResolveTimestampResult(ctx context.Context, rt *runtime.Runtime, instanceID
...
+		if len(q.Aliases) > 0 {
+			return nil, fmt.Errorf("aliases not supported when rows is set")
+		}
+		if q.Having != nil || q.HavingSQL != "" {
+			return nil, fmt.Errorf("having not supported when rows is set")
+		}
+		if len(q.PivotOn) > 0 {
+			return nil, fmt.Errorf("pivot_on not supported when rows is set")
+		}
+	}
```
begelundmuller, line 254: Can these checks be moved into the `runtime/metricsview` package? Since that package can (or will be) called from other places than just the `MetricsViewAggregation` API.

begelundmuller (commented):
1. Also needs to update the `MetricsViewRows` RPC to use the `metricsview` package (to fix the original bug)
2. Can you also add some tests in `runtime/resolvers/testdata` for the rows logic? Especially to check the measure filter support that was originally reported. I.e. have tests like:
```yaml
...
tests:
- name: metrics rows with a measure-based filter
  resolver: metrics
  properties:
    metrics_view: foo
    rows: true
    ...
```

runtime/server/queries_metrics_rows_test.go
```diff
@@ -102,9 +96,9 @@ func TestServer_MetricsViewRows_export_xlsx(t *testing.T) {
 	require.NoError(t, err)
 
 	require.Equal(t, 3, len(rows))
-	require.Equal(t, []string{"timestamp", "publisher", "domain", "bid_price", "volume", "impressions", "ad words", "clicks", "numeric_dim", "device"}, rows[0][2:])
-	require.Equal(t, []string{"2022-01-01T14:49:50.459Z", "", "msn.com", "2", "4", "2", "cars", "", "1", "iphone"}, rows[1][2:])
-	require.Equal(t, []string{"2022-01-02T11:58:12.475Z", "Yahoo", "yahoo.com", "2", "4", "1", "cars", "1", "1"}, rows[2][2:])
+	require.Equal(t, []string{"timestamp", "publisher", "domain", "bid_price", "volume", "impressions", "ad words", "clicks", "numeric_dim", "device"}, rows[0][1:])
+	require.Equal(t, []string{"1/1/22 14:49", "", "msn.com", "2", "4", "2", "cars", "", "1", "iphone"}, rows[1][1:])
```
pjain1, line 100: metricsview_rows export marshals values before writing like 
```
func convertToXLSXValue(pbvalue *structpb.Value) (interface{}, error) {
	switch pbvalue.GetKind().(type) {
	case *structpb.Value_StructValue:
		bts, err := protojson.Marshal(pbvalue)
		if err != nil {
			return "", err
		}

		return string(bts), nil
	case *structpb.Value_NullValue:
		return "", nil
	default:
		return pbvalue.AsInterface(), nil
	}
}
```
however mv agg export does not thats why the date format is of excel format.

begelundmuller (approved):
Nice, this cleans up a lot of old code as well!

runtime/server/queries_metrics_rows_test.go
pjain1: yes only for excel

runtime/metricsview/ast.go
```diff
@@ -212,6 +212,13 @@ func NewAST(mv *runtimev1.MetricsViewSpec, sec *runtime.ResolvedSecurity, qry *Q
 		ast.comparisonDimFields = append(ast.comparisonDimFields, cf)
 	}
 
+	if qry.Rows {
+		ast.dimFields = append(ast.dimFields, FieldNode{
+			Name: "*",
+			Expr: "*",
+		})
+	}
```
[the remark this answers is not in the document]
pjain1: added

PR #7082, opened 2025-04-07
fix: inference from uri scheme fails for `gs`

k-anshul:


runtime/reconcilers/model.go
```diff
@@ -719,7 +719,7 @@ func (r *ModelReconciler) resolveAndSyncPartitions(ctx context.Context, self *ru
 
 	// Log
 	count := batchStartIdx + len(batch)
-	defer r.C.Logger.Info("Resolved model partitions", zap.String("model", self.Meta.Name.Name), zap.Int("partitions", count), observability.ZapCtx(ctx))
+	defer r.C.Logger.Info("Resolved model partitions", zap.String("model", self.Meta.Name.Name), zap.Int("partitions", count), zap.Any("batch", batch), observability.ZapCtx(ctx))
```
k-anshul, line 722: I know this is potentially a very big log message but I find it extremely very helpful when building partitioned models.

runtime/connections.go
```diff
@@ -37,6 +37,9 @@ func (r *Runtime) AcquireSystemHandle(ctx context.Context, connector string) (dr
 
 // AcquireHandle returns instance specific handle
 func (r *Runtime) AcquireHandle(ctx context.Context, instanceID, connector string) (drivers.Handle, func(), error) {
+	if connector == "gs" {
```
k-anshul, line 40: In some places like `glob` resolver we rely on `uri.Scheme` to determine the connector which is mostly gs.

begelundmuller: I agree it's confusing. But I'm afraid this will break if someone creates a `gs.yaml` connector or configures `connector.gs.google_application_credentials`? It might need some more fuzzy/optimistic logic here.

runtime/reconcilers/model.go
begelundmuller: This kind of scares me 😅 Do you think it could be solved by:
1. Printing aggregate status info, e.g. things like `batch_size`, `partitions_count`, `"processing partition 1 of 200"`?
2. Using or improving `rill project partitions` to list the resolved partitions?
3. The `change_mode: analyze` idea shared here: https://www.notion.so/rilldata/ETL-usability-improvements-185ba33c8f5780b5b825eabed3caf9a5?pvs=4#185ba33c8f5780238eaaf3ea407a20b3

k-anshul: This is mostly for local development. 
Given there is no way to inspect what partition is going to output it is useful to inspect the results and having aggregated info is not that useful.

Will it scare you less if we just enable this on `rill-developer` ? I found it really very helpful in building partitioned models.

runtime/connections.go
k-anshul: Yeah that is one case though much less likely.

Do you think the below fuzzy logic is acceptable?

```
cfg, err := r.ConnectorConfig(ctx, instanceID, connector)
if err != nil {
  if strings.Contains(err.Error(), "unknown connector") && connector == "gs" {
	  var gcsErr error
	  cfg, gcsErr = r.ConnectorConfig(ctx, instanceID, "gcs")
	  if gcsErr != nil {
		  return nil, nil, err
	  }
  }
  return nil, nil, err
}
```

runtime/reconcilers/model.go
begelundmuller: So if it's for local, why is `rill project partitions --local` or the sidebar UI for viewing partitions not sufficient? Just want to understand other alternatives.

For first time users of incremental ingestion, seeing a massive JSON-encoded log could also feel more than a little daunting. Clean log messages give people confidence that things are working correctly.

runtime/connections.go
begelundmuller: Yeah it's probably better, but would be even better if it happened inside `ConnectorConfig` because it's called from multiple places.

I also wonder if we can avoid having this and implementing it upstream. Like when something sees `gs://`, it could just always map to the `gcs` connector? Not sure.

runtime/reconcilers/model.go
k-anshul: 1. The sidebar seems to be only available after the model has been built and not before.
image

2. I am not too comfortable with asking users to run too many commands when developing projects. Even as a platform developer I forget about them.   

Also both of these options do not list partitions if model has an error.

Given we want to promote users to use dev partitions it should not output a big log. Plus if the users see many partitions being logged then it will nudge them to use dev partitions.

In general, given, we already receive a lot of feedback even from internal users that runtime acts a black box most of the time, IMO, we should prefer logging information wherever it makes sense for the user to know about them.

runtime/connections.go
k-anshul: Sure doing this inside `ConnectorConfig` sounds good as well.

I am not too sure about handling it upstream. It is easy to miss that handling and is usually only added on discovery. 
For example we already handle gcs/gs in `object_store` but missed adding it in glob resolver.

begelundmuller: I think inside `ConnectorConfig` would be better. But I'm also good with just handling it in the `glob` resolver since hopefully this connector-inference-from-path isn't something we'll be doing more of 😓

runtime/reconcilers/model.go
begelundmuller: Thanks those are good points. Would it help if we:
1. Don't reset the partitions state when a model fails, so the UI and CLI can show the partitions even if the model failed.
2. Changed the UI to load partitions even if there are model errors.
3. Logged the batch when `--verbose` is passed, e.g. using a separate debug-level log.

I'm still not ready to dump the batch every time in the logs since it will can bury more important error logs and a large amount of daunting/dense logs may make people less likely to notice them. It's a goal that the normal (non verbose) logs are user-friendly.

k-anshul: I think just doing 1 and 2 should be good as well. I removed the batch for now.

runtime/connections.go
k-anshul: Sure handling it in `glob` is also fine. Though it will still have the problem that you mentioned above : `But I'm afraid this will break if someone creates a gs.yaml connector or configures connector.gs.google_application_credentials` 

Although the same issue exists today as well for normal GCS sources.

begelundmuller (approved):

PR #7083, opened 2025-04-07
Cleanup spans

k-anshul:
closes https://github.com/rilldata/rill-private-issues/issues/1371
closes https://github.com/rilldata/rill-private-issues/issues/1487

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!


runtime/resolvers/api.go
```diff
@@ -6,8 +6,11 @@ import (
 
 	"github.com/mitchellh/mapstructure"
 	"github.com/rilldata/rill/runtime"
+	"go.opentelemetry.io/otel"
 )
 
+var tracer = otel.Tracer("github.com/rilldata/rill/runtime/resolvers")
```
begelundmuller, line 12: nit: maybe put in a common file called `resolvers.go` instead of `api.go` in the package?

runtime/resolvers/glob.go
```diff
@@ -155,6 +157,15 @@ func (r *globResolver) Validate(ctx context.Context) error {
...
 
 func (r *globResolver) ResolveInteractive(ctx context.Context) (runtime.ResolverResult, error) {
+	ctx, span := tracer.Start(ctx, "glob.ResolveInteractive", trace.WithAttributes(
+		attribute.String("connector", r.props.Connector),
+		attribute.String("path", r.props.Path),
+		attribute.String("partition", string(r.props.Partition)),
+		attribute.Bool("rollup_files", r.props.RollupFiles),
+		attribute.String("transform_sql", r.props.TransformSQL),
+	))
+	defer span.End()
```
begelundmuller, line 167: Instead of starting spans in each resolver separately, could we do it one place in `runtime/resolver.go`? That would also enable spans to show up for resolvers that have a cache hit. (Which I _think_ would be nice to understand the code paths being taken, or do you disagree?)

It could either include all the props and args as attributes, or else maybe we could add a `SpanAttributes` func to the resolver interface to allow each resolver to sanitize/truncate its props/args first.

Or do you think it's nicer to do it separately in each resolver?

k-anshul: Yeah I think having `cache_hit` would be nice to track. The problem is that `SpanAttributes` does not take `map` input so need to parse and than add attributes. 
Keeping `SpanAttributes` in resolver interface does not seem very ideal to me.
One option is to get the parent span in individual resolvers and if valid add attributes in that. Of course that comes with an assumption that a span will always be added in `runtime/resolver.go`. 
If that is not fine I can add `SpanAttributes` func as well.

begelundmuller: I think the problem with the current approach of adding the props/args in `ResolveInteractive` is that they are not added for cache hits. So the trace doesn't show what query the cache hit was for, only that there was one. I thought that could be addressed with a new interface function, but also open to any other suggestions you have.

k-anshul: Actually we can populate args and props in the [initializer](https://github.com/rilldata/rill/blob/abd181b3533b0e7be5b9fb250201f3950b5acdea/runtime/resolver.go#L120-L127) itself. That will also save the trouble of setting the same span in the singleflight.

What do you think ?

begelundmuller: Yeah that seems good to me :)

runtime/resolver.go
```diff
@@ -179,18 +184,21 @@ func (r *Runtime) Resolve(ctx context.Context, opts *ResolveOptions) (ResolverRe
...
 		// Try cache again
 		if val, ok := r.queryCache.cache.Get(key); ok {
+			span.SetAttributes(attribute.Bool("cache_hit", true))
 			return val.(*cachedResolverResult), nil
 		}
 
 		// Resolve
 		// NOTE: We can under no circumstances return the res directly since we're in a singleflight and results can have iterator state.
+		span.SetAttributes(attribute.Bool("cache_hit", false))
+		ctx = trace.ContextWithSpan(ctx, span)
```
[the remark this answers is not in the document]
k-anshul: Actually I now realise that we don't need this. In the singleflight we copy the values of the parent ctx : https://github.com/rilldata/rill/blob/72ac43fbe30b433c6b43b2c76c6aaa699e047511/runtime/pkg/singleflight/singleflight.go#L159

runtime/resolvers/metrics.go
```diff
@@ -39,6 +41,11 @@ func newMetrics(ctx context.Context, opts *runtime.ResolverOptions) (runtime.Res
 		return nil, err
 	}
 
+	span := trace.SpanFromContext(ctx)
```
k-anshul, line 44: I am skipping setting the full query here since 
a) We can nearly rebuild it using the SQL query submitted to OLAP.
b) Need to find a way to serialize it to attributes. May be we can just convert it to JSON if it is required.

begelundmuller (approved):

PR #7084, opened 2025-04-07
refactor (state management): use explicit `activePage` state to route to Explore views

ericpgreen2:
There are various "views" within an Explore dashboard: the expanded dimension table, time dimension detail (TDD), and the pivot table. Our monolithic Explore state store (aka `MetricsExplorerEntity`) keeps state for each of these views _and_ keeps track of the active view via the `activePage` property.

While we've long had the `activePage` property, up until now we've chosen to route to the various views via this logic:
- If `exploreState.selectedDimensionName`, route to the expanded dimension table
- If `exploreState.tdd.expandedMeasureName`, route to TDD
- If `exploreState.pivot.active`, route to the pivot table

This PR updates our view routing logic so that we merely look at `activePage` to determine the view.

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!


AdityaHegde (approved):
We should get rid of pivot's active, I guess in a follow up to make sure we dont use it.

Issue #1484, opened 2022-12-19
Lower left panel weight of version next to blue info icon is confusing


magorlick:
- info icon should not be blue, it should be the same grey as the other icons.
- "version blah" should be in italics. 
- Don't put the info icon on the same line as the version information.  Go to the next line or right justify.
- Remove the item that says "Documentation" entirely. That is where the info icon points now.  
- Do _not_ add the text "documentation" back next to the info icon.

![Image](https://user-images.githubusercontent.com/5587788/208530108-6a2560fc-3f7a-4e78-bd9d-c596088b0f12.png)

PR #7089, opened 2025-04-07
adding retries for pinot sql driver

NamanMahor:
I have added the list of error from Pinot code which I think is should fail and everything else we will retry. 

Closes [#1484](https://github.com/rilldata/rill-private-issues/issues/1484)

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


runtime/drivers/pinot/sqldriver/driver.go
```diff
@@ -57,35 +74,44 @@ func (c *conn) Begin() (sqlDriver.Tx, error) {
...
-		resp, err = c.pinotConn.ExecuteSQL("", query)
-	}
-	if err != nil {
-		return nil, err
-	}
-	if len(resp.Exceptions) > 0 {
-		if len(resp.Exceptions) == 1 {
-			return nil, fmt.Errorf("query error: %q: %q", resp.Exceptions[0].ErrorCode, resp.Exceptions[0].Message)
+		if err != nil {
+			return nil, retrier.Fail, err
```
begelundmuller, line 92: I might be wrong, but isn't the `pinotConn` backed by a HTTP transport? So if the HTTP call itself fails (e.g. if there's a load balancer in front of Pinot that returns a retryable error due to the broker being unreachable), then won't it return a normal `net/http` error here? E.g. from here: https://github.com/startreedata/pinot-client-go/blob/1c314c96b95b74aeb1386473df4e079ec700efa1/pinot/jsonAsyncHTTPClientTransport.go#L68

If that's the case, then I think we also need to handle some retryable HTTP error codes here.

runtime/drivers/pinot/sqldriver/driver.go
```diff
@@ -57,35 +74,44 @@ func (c *conn) Begin() (sqlDriver.Tx, error) {
...
+				errMsg += fmt.Sprintf("\t%q: %q\n", e.ErrorCode, e.Message)
+			}
+			err := errors.New(errMsg)
+			for _, e := range resp.Exceptions {
+				switch e.ErrorCode {
+				case JSONParsing, SQLParsing, SQLRuntime, AccessDenied, TableDoesNotExist, TableIsDisabled, QueryExecution, QueryValidation, UnknownColumn, QueryPlanning:
+					return nil, retrier.Fail, err
+				default:
+					return nil, retrier.Retry, err
+				}
```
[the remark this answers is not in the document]
NamanMahor: done

runtime/drivers/pinot/sqldriver/driver.go
NamanMahor: done

runtime/drivers/pinot/sqldriver/driver.go
```diff
@@ -10,13 +10,65 @@ import (
...
 	"reflect"
+	"regexp"
+	"strconv"
 	"time"
 
+	"github.com/rilldata/rill/runtime/pkg/retrier"
 	"github.com/startreedata/pinot-client-go/pinot"
 )
 
+func isRetryableHTTPError(err error) bool {
```
begelundmuller, line 24: Move these util functions to the end of the file as per Uber's Go style guide, which we follow: https://github.com/uber-go/guide/blob/master/style.md#function-grouping-and-ordering

I know in some languages it's common to have utils before the main logic, but we (and most other Go projects) order files with the external/main/important logic first.

runtime/drivers/pinot/sqldriver/driver.go
```diff
@@ -10,13 +10,65 @@ import (
...
+		}
+	}
+	errStr := err.Error()
+	re := regexp.MustCompile(`Pinot: (\d{3})`)
+	matches := re.FindStringSubmatch(errStr)
+	if len(matches) == 2 {
+		if code, convErr := strconv.Atoi(matches[1]); convErr == nil {
+			return isRetryableHTTPCode(code)
+		}
+	}
```
begelundmuller, line 38: Not sure about the Pinot library internals, but does it have a structured way to get the HTTP error code other than using regexes (which might change without us noticing in new versions)?

Specifically I'm worried about the cases where the network request bounces at a load balancer and the underlying Pinot cluster is never reached. What does the error become there and will it correctly retry those?

begelundmuller (changes requested):

runtime/drivers/pinot/sqldriver/driver.go
NamanMahor: actually there are 3 type of error 
1. one http/net error [here](https://github.com/startreedata/pinot-client-go/blob/v0.5.0/pinot/jsonAsyncHTTPClientTransport.go#L58) which handled by errors.As(err, &netErr) in isRetryableHTTPError

2. one with response not 200 [here](https://github.com/startreedata/pinot-client-go/blob/v0.5.0/pinot/jsonAsyncHTTPClientTransport.go#L78)  which we need to parse using re := regexp.MustCompile(`Pinot: (\d{3})`)

3. exception in BrokerResponse [here](https://github.com/startreedata/pinot-client-go/blob/v0.5.0/pinot/jsonAsyncHTTPClientTransport.go#L76) which we are handling using isRetryablePinotErrorCode

NamanMahor: for case 1. code is changed. in latest version. I will raise PR in pinot for propagating the error no change required in our code. (also have't seen it in our logs)
for remaining case this PR is sufficient.

begelundmuller (approved):

pjain1:
@begelundmuller looks fine to me. @NamanMahor can you try against a live pinot cluster that basic functionality is working, will ping you the details on slack.

PR #7092, opened 2025-04-08
fix: get rid of the need for defaultExplorePreset pass 1

AdityaHegde:
We use a defaultExplorePreset to decide what needs to be shown in URL. We want to get away from ExplorePreset type as much as possible, it should only be to extracts dashboard config from yaml.

This is the one of the steps towards it. We are replacing `defaultExplorePreset` with `defaultExploreUrlParams` so that the decision to keep/remove params happens directly with url params.

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!


web-common/src/features/dashboards/url-state/convert-partial-explore-state-to-url-params.ts
```diff
@@ -0,0 +1,384 @@
...
+      const defaultValue = defaultExploreUrlParams.get(key);
+      if (
+        (defaultValue === null && value !== "") ||
+        (defaultValue !== null && value !== defaultValue)
+      ) {
+        return;
+      }
+      searchParams.delete(key);
+    });
+  }
```
ericpgreen2, line 123: Here, we should be reconciling two `Partial<ExploreState>` (our canonical data structure) rather than reconciling `URLSearchParams` (merely a user-facing encoding).

ericpgreen2 (changes requested):
Let's talk about this. This PR is not only attempting to quarantine the `V1ExplorePreset`, (which is good!) but it's also shifting how we transfer state around the application (which is a big change!). It looks like this PR moves toward a model where state is transferred through the application as `URLSearchParams`, rather than `Partial<MetricsExplorerEntity>`.

URL params are user-facing not developer-facing. E.g. when reasoning about time ranges as a developer, I would prefer to reason about and manipulate a `timeRange: TimeRange` object, rather than a `tr=<string>` data structure.

Happy to discuss, but IMO we should continue to use `MetricsExplorerEntity` (to be renamed to `ExploreState`) as the canonical data structure, not `URLSearchParams`. We should move toward an `ExploreState` class that has these methods called at the very edges of our application:
- `fromURL()`, `toURL()`
- `fromV1ExplorePreset()`
- `fromProtobufEncodedState()`, `toProtobufEncodedState()`
- `saveToLocalStorage()`, `loadFromLocalStorage()`, `saveToSessionStorage()`, `loadFromSessionStorage()`

With the above, we could decode the `V1ExplorePreset` as soon as we get it back from the API like so:
```
const explore = await runtimeServiceGetExplore(exploreName)
const defaultExploreState = exploreState.fromExplorePreset(explore.spec.defaultPreset)
// merge `defaultExploreState` with other `Partial<ExploreState>` as necessary
```

web-admin/src/features/bookmarks/selectors.ts
```diff
@@ -223,16 +230,18 @@ function parseBookmark(
 
   const url = new URL(get(page).url);
 
-  const searchParams = convertExploreStateToURLSearchParams(
-    finalExploreState,
+  // We need to check if the bookmark's url is equal to current url or not to show an "active" state.
```
AdityaHegde, line 233: This will come in a follow up PR

web-common/src/features/dashboards/state-managers/loaders/DashboardStateDataLoader.ts
```diff
@@ -100,84 +96,54 @@ export class DashboardStateDataLoader {
       },
     );
 
-    this.defaultExploreStateAndErrors = getCompoundQuery(
+    this.defaultExploreState = getCompoundQuery(
```
ericpgreen2, line 99: Do we need both `this.defaultExploreState` and `this.explorePresetFromYAMLConfig`? Can we remove `this.explorePresetFromYAMLConfig`? If you're keeping this to a separate PR, that's fine 👍.

web-common/src/features/dashboards/stores/get-blank-explore-state.ts
ericpgreen2, line 1: Naming nit– "blank" doesn't quite capture it. Especially for time ranges, we're applying some Rill-opinionated logic to initialize a sensible state. What about one of:
1. `getSystemDefaultExploreState`
2. `getRillDefaultExploreState`
3. `getBaselineExploreState`

^That's my order of preference, leaning towards 1 or 2.

web-common/src/features/dashboards/url-state/convert-partial-explore-state-to-url-params.ts
```diff
@@ -0,0 +1,385 @@
...
+  partialExploreState: Partial<MetricsExplorerEntity>,
+  // We have quite a bit of logic in TimeControlState to validate selections and update them
+  // Eg: if a selected grain is not applicable for the current grain then we change it
+  // But it is only available in TimeControlState and not MetricsExplorerEntity
+  timeControlsState: TimeControlState | undefined,
+  // Used to remove params that are equal to defaults. This helps keep the url shorter.
+  defaultExploreUrlParams?: URLSearchParams,
+  // Used to decide whether to compress or not based on the full url length
+  // TODO: move this to getCleanedUrlParamsForGoto
+  urlForCompressionCheck?: URL,
```
ericpgreen2, line 49: It looks like these aren't used anywhere, so we can rely on `getCleanedUrlParamsForGoto` to handle this

web-common/src/features/dashboards/url-state/get-cleaned-url-params-for-goto.ts
ericpgreen2, line 1: Can you please put this in the same file as `convertPartialExploreStateToURLParams()`? It'll be easier to reason about them if they're next to each other

web-common/src/features/dashboards/url-state/convert-partial-explore-state-to-url-params.ts
```diff
@@ -0,0 +1,385 @@
...
+  type TimeRange,
+  TimeRangePreset,
+} from "@rilldata/web-common/lib/time/types";
+import { copyParamsToTarget } from "@rilldata/web-common/lib/url-utils";
+import { DashboardState_ActivePage } from "@rilldata/web-common/proto/gen/rill/ui/v1/dashboard_pb";
+import { type V1ExploreSpec } from "@rilldata/web-common/runtime-client";
+
+export function convertPartialExploreStateToUrlParams(
+  exploreSpec: V1ExploreSpec,
+  partialExploreState: Partial<MetricsExplorerEntity>,
```
ericpgreen2, line 40: In the future, I think this should be a `exploreState.toURLSearchParams()` function that operates on a complete `ExploreState` not a partial one.

web-common/src/features/dashboards/stores/get-blank-explore-state.ts
AdityaHegde: This is actually not used. Removing it, might add it once we have cascading merge.

web-common/src/features/dashboards/url-state/convert-partial-explore-state-to-url-params.ts
AdityaHegde: It is currently used in `updateExploreSessionStore` where it could be a partial. So lets do this in a follow up.

web-common/src/features/dashboards/state-managers/loaders/DashboardStateDataLoader.ts
ericpgreen2: I see, got it

ericpgreen2 (approved):
Thank you for this step forward in reducing complexity!

PR #7097, opened 2025-04-08
druid version check

pjain1:
Fixes https://github.com/rilldata/rill-private-issues/issues/1166

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/runtime/drivers/druid/druid.go
+++ b/runtime/drivers/druid/druid.go
@@ -9,6 +9,7 @@ import (
 	"net/url"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/XSAM/otelsql"
 	"github.com/jmoiron/sqlx"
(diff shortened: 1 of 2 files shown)


runtime/drivers/druid/druid.go
```diff
@@ -162,6 +174,53 @@ func (d *driver) TertiarySourceConnectors(ctx context.Context, src map[string]an
...
+		return fmt.Errorf("druid version check failed with status code: %d", resp.StatusCode)
+	}
+
+	var statusResponse map[string]any
+	if err := json.NewDecoder(resp.Body).Decode(&statusResponse); err != nil {
+		return fmt.Errorf("failed to decode Druid status response: %w", err)
+	}
+
+	// Check if druid version is greater than or equal to 28.0.0, because of SQL NULL compatibility
+	if version, ok := statusResponse["version"]; ok {
```
begelundmuller, line 207: You can use an inline struct to make the JSON library decode the field into a string automatically:

```suggestion
	var statusResponse struct {
		Version string `json:"version"`
	}
	if err := json.NewDecoder(resp.Body).Decode(&statusResponse); err != nil {
		return fmt.Errorf("failed to decode Druid status response: %w", err)
	}
```

begelundmuller (approved):

PR #7109, opened 2025-04-09
feat : Support concurrent write queries in duckdb driver

k-anshul:
closes https://github.com/rilldata/rill/issues/6999

Reduces total reconcile time for `openrtb` example project from `25s` to `17s`.

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!


runtime/pkg/rduckdb/db.go
```diff
@@ -386,14 +415,21 @@ func (d *db) CreateTableAsSelect(ctx context.Context, name, query string, opts *
...
 		_ = release()
 	}()
 
+	// start running user queries
+	//
+	// given runnning user queries is the most intensive operation and assuming following constraints, we can release the lock and run user queries concurrently:
+	// 1. A table write attaches all other databases in read only mode and writes to a different database
+	// 2. Reconciler makes sure that dependent resources are not reconciled concurrently
+	d.writeSem.Release(1)
+	writeSemAcquired = false
```
begelundmuller, line 432: 1. If the write semaphore is released before the write queries, then is the write semaphore even needed?
2. Does this mean more concurrent writes can happen than the write semaphore says? IMO it would be convenient to be able to enforce a limit on the number of concurrent writes inside the driver as well (so we don't have to be super cautious about the concurrency of upstream logic). Could it keep enforcing the write semaphore limi if we changed the write semaphore to a higher default here? https://github.com/rilldata/rill/blob/2cb09ba3dd71ef85885b183a4e6e9c4201cfb365/runtime/pkg/rduckdb/db.go#L189

runtime/pkg/rduckdb/db.go
```diff
@@ -578,7 +654,23 @@ func (d *db) RenameTable(ctx context.Context, oldName, newName string) error {
...
+	}
+	err = d.trackWrite(newName)
+	if err != nil {
+		d.untrackWrite(oldName)
+		return err
+	}
+	defer func() {
+		d.untrackWrite(oldName)
+		d.untrackWrite(newName)
+	}()
```
begelundmuller, line 671: Would this be simpler?

```suggestion
	err := d.trackWrite(oldName)
	if err != nil {
		return err
	}
	defer d.untrackWrite(oldName)
	err = d.trackWrite(newName)
	if err != nil {
		return err
	}
	defer d.untrackWrite(newName)
```

runtime/pkg/rduckdb/db.go
```diff
@@ -712,112 +804,76 @@ func (d *db) Size() int64 {
...
-			_, err = conn.ExecContext(ctx, "DETACH DATABASE "+safeSQLName(name))
-			if err != nil {
-				return err
-			}
-		}
-		return nil
-	}
+		// detach can take long time if duckdb runs compaction so we need to set a timeout to not block the caller for too long
+		ctx, cancel := graceful.WithMinimumDuration(ctx, 5*time.Second)
+		defer cancel()
```
begelundmuller, line 871: I wonder if 5s is too little for some compactions. Since this is on the write path, a bit extra delay is perhaps okay (maybe 15s?).

The controller allows up to 30s for a cancellation: https://github.com/rilldata/rill/blob/2cb09ba3dd71ef85885b183a4e6e9c4201cfb365/runtime/controller.go#L302

runtime/pkg/rduckdb/db.go
```diff
@@ -183,15 +182,17 @@ func NewDB(ctx context.Context, opts *DBOptions) (DB, error) {
...
-		ctx:        bgctx,
-		cancel:     cancel,
+		opts:             opts,
+		localPath:        opts.LocalPath,
+		remote:           opts.Remote,
+		writeSem:         semaphore.NewWeighted(1),
+		metaSem:          semaphore.NewWeighted(1),
+		localDirty:       true,
+		writesInProgress: make(map[string]any),
+		progressMu:       sync.Mutex{},
```
begelundmuller, line 192: Since the number of members is growing big, consider naming it `writesInProgressMu` since it specifically relates to `writesInProgress`.

begelundmuller (changes requested):

runtime/pkg/rduckdb/db.go
k-anshul: 1. Yes so the idea is that only the ingestion in duckdb is allowed to run in parallel and rest of the stuff still works in serial like taking backups, pulling from backups etc. Those stuff are somewhat tricky to implement in parallel and given the ingestion takes up most of the latency so it seems better to just allow that to run concurrently.
2. Yes multiple ingestion runs can happen in parallel right now. I expected the database to take care of that. We can not increase `writeSem` to more than 1 for reasons stated above. 
We can set connection limit on the write handle to enforce not more than X ingestion queries running in paraellel.

runtime/pkg/rduckdb/db.go
k-anshul: Sure.

runtime/pkg/rduckdb/db.go
begelundmuller: > only the ingestion in duckdb is allowed to run in parallel and rest of the stuff still works in serial like taking backups, pulling from backups, directory cleanups etc

Could they be covered by the `metaSem`? Since they kind of sound like "meta" operations. Or should the `writeSem` be renamed to something else? (It's just a bit confusing that the `writeSem` is not actually acquired when writes happen.)

> we can set connection limit on the write handle to enforce not more than X ingestion queries running in paraellel.

Yeah that sounds fine. Should we add a config key for it and set some reasonable default limit for it as well? Like maybe 3 or 5? Maybe it's unwarranted, but I'm quite scared of too many parallel writes

k-anshul: > Could they be covered by the metaSem? 

The `metaSem` is acquired when generating a read connection as well and we don't want writes to block reads so we can keep separate locks. I renamed `metaSem` to `readMetaSem` and `writeSem` to `writeMetaSem`. 

> Should we add a config key for it and set some reasonable default limit for it as well?

Sure just added a configurable option with default as 5.

runtime/drivers/duckdb/config.go
```diff
@@ -23,6 +23,8 @@ type config struct {
 	MemoryLimitGB int `mapstructure:"memory_limit_gb"`
 	// ReadWriteRatio is the ratio of resources to allocate to the read DB. If set, CPU and MemoryLimitGB are distributed based on this ratio.
 	ReadWriteRatio float64 `mapstructure:"read_write_ratio"`
+	// WriteConnLimit is the maximum number of concurrent write connections allowed.
+	WriteConnLimit int `mapstructure:"write_conn_limit"`
```
begelundmuller, line 27: There's another config called `PoolSize` and it's unclear how this interacts with that. Can the two be refactored to have consistent naming that clarifies their relationship? Maybe `ReadPoolSize` and `WritePoolSize`?

begelundmuller (approved):

runtime/drivers/duckdb/config.go
k-anshul: Sure. Made the change.

PR #7129, opened 2025-04-10
hotfix: truncate long draggable list item names

lovincyrus:
Reported in https://rilldata.slack.com/archives/C02T907FEUB/p1744295510607029

This pull request fixes the styling tweak of long draggable list item names by truncation.

https://github.com/user-attachments/assets/2f4d42d3-54bf-437c-88d0-9b140b6a90ff

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!

### DIFF
--- a/web-common/src/components/menu/DashboardMetricsDraggableList.svelte
+++ b/web-common/src/components/menu/DashboardMetricsDraggableList.svelte
@@ -466,7 +466,7 @@
 
                     <button
                       class="hover:bg-slate-200 p-1 rounded-sm active:bg-slate-300"
-                      on:click={() => {
+                      on:click|stopPropagation={() => {
                         selectedItems = [...selectedItems, id];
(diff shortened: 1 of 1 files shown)


ericpgreen2 (changes requested):
Two UX improvements:
1. Can we add the measure display name to the tooltip (in addition to the description)?
2. Can we add the same tooltip to the hidden measures?

jkhwu:
@ericpgreen2 
1. I discussed this with @lovincyrus and given that we're trying to reduce our tooltip sizes, we decided not to add the display name. I think users can infer based on the space allotted what the full measure name is and once they make it visible, they can see the name on the Overview page.
2. Agree

ericpgreen2 (approved):
For a hidden measure that doesn't have a description– the tooltip still shows as a dark blob:
image

With that fixed, LGTM.

PR #7134, opened 2025-04-11
Simplify `OLAPStore` interface

begelundmuller:
**Changes:** 
- Removes the CRUD functions from the `OLAPStore` interface. This simplifies the OLAP interface and finalizes the move of CRUD-related logic to model executors.
- Removes the unused `longRunning` option from the external interface.
- Moves CRUD logic to its own file in the ClickHouse and DuckDB drivers.
- Adds docstrings throughout the `OLAPStore` interface

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!


k-anshul (approved):
Any particular reasons for keeping the CRUD APIs exported ?
They should ideally be just required in model executors.

begelundmuller:
> Any particular reasons for keeping the CRUD APIs exported ? They should ideally be just required in model executors.

I guess it's mainly laziness due to them being used in some tests with a different package name. But it could be addressed with a deeper refactor, which I just pushed.

PR #7140, opened 2025-04-14
fix: Regenerate `package-lock.json`

ericpgreen2:
In PR #6892, I ran into issues adding a couple dependencies. This PR re-generates the `package-lock.json` file to address those problems.

Changes:
- Removed the Docusaurus pin to `v3.1.1`, which was causing a `peerDependency` conflict
- Added an `.nvmrc` file to clarify that we're using Node v20 (with npm v10.8.2)
- Added an `overrides` block in the root `package.json` to ensure `@sveltejs/vite-plugin-svelte` is compatible with Svelte 4
- Updated the `tsconfig.json` in `web-common` to prevent it from inheriting from a missing SvelteKit config
- Modified the `svelte-check` ignore list to not use parentheses as a grouping operator, which is unsupported in the latest `svelte-check` version

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!


briangregoryholmes:
Functionally not super relevant, but the `tsconfig.json` should extend `./.svelte-kit/tsconfig.json` just to silence the warning. It'll be generated via sync, build or dev if it's not there. We're not using the generated types at the moment, but if we do build out some better component testing, we'd probably put some routes and load functions in that package eventually.

AdityaHegde (approved):

briangregoryholmes (approved):

ericpgreen2:
> Functionally not super relevant, but the web-common tsconfig.json should extend ./.svelte-kit/tsconfig.json just to silence the warning. You may see a warning if it isn't there the first time, but it'll be generated via sync, build or dev and won't warn on subsequent runs. We're not using the generated types at the moment, but if we do build out some better component testing, we'd probably put some routes and load functions in that package eventually.

Okay, I've added it back. The problem I hit was that `npm run build -w web-{admin,local}` was failing because the `web-common`'s `svelte-kit/.tsconfig` file hadn't yet been generated. Fixed like so:
image

PR #7142, opened 2025-04-14
feat: allow multiple chart types, add donut, heatmap

djbarnwal:
image

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!


web-common/src/features/canvas/components/charts/circular-charts/CircularChart.ts
```diff
@@ -0,0 +1,179 @@
...
+            limit: limit.toString(),
+          },
+          {
+            query: {
+              enabled,
+              placeholderData: keepPreviousData,
+            },
+          },
+          ctx.queryClient,
+        );
```
ericpgreen2, line 124: IMO it's quite hard to follow the multiple tiers of derived stores, and they can lead to performance issues. In this case, every change to the `timeAndFilterStore` will re-create a new `dataQuery` observer. These query observers need to be stable for `keepPreviousData` to take effect.

Here's the TanStack Query v5 pattern that I think we should follow:
```typescript
const xQueryOptionsStore = derived(...)
const xQuery = createQuery(xQueryOptionsStore)
return xQuery;
```

In cases where we need to make multiple queries:
```typescript
const xQueryOptionsStore = derived(...)
const yQueryOptionsStore = derived(...)
const zQuery = createQueries({ queries: [xQueryOptionsStore, yQueryOptionsStore], combine: ([xResponse, yResponse] => z))
return zQuery;
```

briangregoryholmes (approved):

briangregoryholmes:
I don't necessarily think this should hold anything up, but my impression was that we wanted to default to a donut chart visualization rather than a pie chart, which would mean our default inner radius should be something other than 0.

I also think inner radius is maybe the wrong control to expose. It should be relative to the outer radius of the pie chart, which is not user controllable and determined responsively by the size/positioning of the widget. It's more a "thickness" property that goes from, you know, 10% to 100%. You can see the issue when you set the inner radius to something large and then resize the widget container to something small.

djbarnwal:
> I also think inner radius is maybe the wrong control to expose. It should be relative to the outer radius of the pie chart, which is not user controllable and determined responsively by the size/positioning of the widget. It's more a "thickness" property that goes from, you know, 10% to 100%. You can see the issue when you set the inner radius to something large and then resize the widget container to something small.

This makes sense. Vega doesn't support percentage values for radius out of the box but I think there's a way to have relative inner radius using expression functions. I will take this as a follow up so that this and the legend PR are unblocked.

PR #7149, opened 2025-04-15
fix: dimension bulk search when both newline and comma is present

AdityaHegde:
When both newline and comma is present in the bulk search text then only split using newline. To allow for this we also need to support multiline in search input.

Background: https://rilldata.slack.com/archives/C088S9KJB17/p1744657879521399

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


web-common/src/components/search/Search.svelte
```diff
@@ -36,6 +37,27 @@
...
+    const correctedLines = Math.max(
+      // Show minimum of 1 line
+      1,
+      // We expand the input only till MaxLines
+      Math.min(lines, MaxLines),
+    );
+
+    height = `h-[${BaseHeight + (correctedLines - 1) * HeightPerLine}px]`;
+    console.log(lines, correctedLines, height);
+  }
```
ericpgreen2, line 58: Can we use CSS rather than JS for this?

web-common/src/components/search/Search.svelte
```diff
@@ -52,24 +74,42 @@
...
+      class="outline-none rounded-[2px] block w-full pl-8 p-1 {forcedInputStyle}"
+      class:h-full={large}
+      {disabled}
+      {placeholder}
+      bind:value
+      on:input
+      on:keydown={handleKeyDown}
+      aria-label={label}
+    />
+  {/if}
```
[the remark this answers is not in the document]
AdityaHegde: Yes, that was my 1st instinct as well. But I get `'value' is not a valid binding on <svelte:element> elements` when doing this since value is not available for every element. We could update value in `keydown` but that feels hacky compared to this.

web-common/src/components/search/Search.svelte
AdityaHegde: You mean something like `h-7` `h-10` etc? Not all height values we use here have it, otherwise it would have been perfect.

web-common/src/components/search/Search.svelte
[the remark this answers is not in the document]
ericpgreen2: IMO the first priority is to avoid the redundancy of the current approach. It's too bad that `bind` doesn't work here, but it's just syntactic sugar– updating the value using `on:<event>` is not the end of the world.

web-common/src/components/search/Search.svelte
ericpgreen2: I mean using a class something like this:
```css
.textarea-auto-height {
  height: auto;
  min-height: 28px;
  max-height: 92px; /* Equivalent to 5 lines: 28px + (4 * 16px) */
  line-height: 16px;
  overflow-y: auto;
  resize: none;
  box-sizing: border-box;
  padding: 6px 8px;
}
```

AdityaHegde: Updated to use both js and min/max height.

ericpgreen2 (approved):

PR #7156, opened 2025-04-15
feat: dark mode

briangregoryholmes:
INSERT DESCRIPTION HERE

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


web-common/src/components/button/classes.ts
```diff
@@ -6,7 +6,7 @@ export const levels = {
     secondary:
       "text-gray-800 border rounded-sm border-gray-300 shadow-sm hover:bg-gray-100 hover:text-gray-700 hover:border-gray-300 focus:ring-primary-300",
     highlighted:
-      "text-gray-500 border border-gray-200 hover:bg-gray-200 hover:text-gray-600 hover:border-gray-200 focus:ring-primary-300 shadow-lg rounded-sm h-8 ",
+      "text-gray-500 border   hover:bg-gray-200 hover:text-gray-600 hover:  focus:ring-primary-300 shadow-lg rounded-sm h-8 ",
```
djbarnwal, line 9: Seems like a bad find and replace with `border-gray-200`.

web-common/src/features/column-profile/ProfileContainer.svelte
```diff
@@ -39,7 +39,7 @@
         flex
         space-between
         gap-2
-        transition-colors
+
```
djbarnwal, line 42: nit: bad replace

web-common/src/components/tooltip/StackingWord.svelte
```diff
@@ -32,7 +32,5 @@
     @apply inline-block;
     border-radius: 2px;
     position: relative;
-    mix-blend-mode: screen;
-    background-blend-mode: screen;
```
djbarnwal, line 36: Not sure what this does but the Copy animation in light mode looks weird and has a momentarily dark bg

djbarnwal (changes requested):
Looks great overall.

- By default, without any feature flag, the application takes system theme settings (dark and light). Also since the theme toggle is behind a feature flag, it's not possible to explicitly change it in this case. I would expect the dark theme to be only available if I have set the feature flag explicitly. (**Blocker**)


- I see that `border-gray-200` has been removed from the whole project baring a couple of places. Any reason we would want to remove that instead replacing it with a class which changes based on theme?
- Similarly found `transition-color` to be removed from a lot of places.

- There are few occurrences of  bad find and replace leaving some whitespace in the class name strings, can we use a linter to fix class name strings in svelte files?

### UXQA light mode
Notifications blending with UI - 
image 

Tooltips do not retain their dark bg in light mode 
image

In both light and dark modes, I would expect a more contrasting color for tooltips and notifications background. So we should probably preserve the dark bg in light mode and light bg in dark mode for these two components.

### UXQA Dark mode 

I personally found the dark mode a bit too dark making components such as table hard to read
image
image

briangregoryholmes:
> Looks great! The text is much more legible now in dark mode.
> 
> I saw an issue with the feature flag though. On refreshing the page, the dropdown checks the right theme but it's not applied until you reselect it

> Approving so that you can merge this post fixing the above issue.

Haven't been able to reproduce this in Safari or Chrome. Looks like it may be isolated to Firefox. Will take a look.

djbarnwal:
> Haven't been able to reproduce this in Safari or Chrome. Looks like it may be isolated to Firefox. Will take a look.

I saw the issue on Chrome so could also be something to do with system setting perhaps?

PR #7160, opened 2025-04-16
Json schema for Yamls

NamanMahor:
Closes [#1489](https://github.com/rilldata/rill-private-issues/issues/1489)

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!


begelundmuller (commented):
This looks great! Some thoughts:
1. It would be convenient to have one combined schema file that can easily be referenced externally. This can perhaps be auto-generated with a CLI tool?
2. Could you try adding validation with the JSON schema in the parser code? That will help validate the schema because any unit test that doesn't match the schema would fail (revealing properties that may be missing). You can look at the currently commented-out component validation code [here](https://github.com/rilldata/rill/blob/2dca926416eef3c3769d0c3dd8bfd58cf503080a/runtime/parser/parse_component.go#L93) for an example.
3. Adding descriptions to the individual properties for models, metrics views, explores and canvases would be very useful. That will enable auto-generating useful docs based on the schema here.
4. Do you have any ideas for how we can incorporate model input/output properties in the schema here? It's fine to maintain them manually, but they have to be kind of loosely typed since the supported properties depends on the value of the `connector:` field.
5. For context, the next work after this schema is in place is to:
    - Sync up with the applications team to integrate it for auto-complete in the UI
    - Set up a script that uses it to auto-generate reference docs. It should work similar to the current `make docs.generate` make command, which outputs Markdown-formatted CLI usage docs in `docs/docs/reference/cli`.

runtime/parser/schema/rillyaml.schema.json
```diff
@@ -0,0 +1,129 @@
...
+                "type": "array",
+                "description": "An array of group names that the mock user is a member of",
+                "items": {
+                  "type": "string"
+                }
+              }
+            },
+            "required": [
+              "email"
+            ]
```
begelundmuller, line 123: These are the main properties, but we also allow arbitrary values attributes here. Does the schema fail if an arbitrary attribute is specified or does it gracefully handle unknown attributes? 

Is there a way to configure whether or not unknown properties are allowed on a per-type basis?

runtime/parser/schema/resource.schema.json
```diff
@@ -0,0 +1,3878 @@
...
+      "$ref": "#/$defs/connector"
+    },
+    {
+      "$ref": "#/$defs/explore"
+    },
+    {
+      "$ref": "#/$defs/metrics_view"
+    },
+    {
+      "$ref": "#/$defs/migration"
```
begelundmuller, line 29: Can you remove migration? We are going to remove it so don't want to document it.

begelundmuller (approved):
Looking forward to see the auto-generated docs!

runtime/parser/schema/resource.schema.json
NamanMahor: sure

runtime/parser/schema/rillyaml.schema.json
begelundmuller: Great! I agree that sounds safest

PR #7163, opened 2025-04-16
fix: default params strippings for non-explore views

AdityaHegde:
We have a default set of params that we strip from the url. But we do not take into account the difference in view. This is an issue only for pivot since there is an overlap in param name but they mean different things.

We should revisit the stripping and look at doing it using explore state directly.

**Checklist:**
- [x] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-common/src/features/dashboards/url-state/convert-partial-explore-state-to-url-params.ts
+++ b/web-common/src/features/dashboards/url-state/convert-partial-explore-state-to-url-params.ts
@@ -25,6 +25,7 @@ import {
   ExploreStateKeyToURLParamMap,
   ExploreStateURLParams,
 } from "@rilldata/web-common/features/dashboards/url-state/url-params";
+import { stripDefaultUrlParams } from "@rilldata/web-common/features/dashboards/url-state/url-params-strip-utils";
 import { arrayOrderedEquals } from "@rilldata/web-common/lib/arrayUtils";
 import {
   TimeComparisonOption,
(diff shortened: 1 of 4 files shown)


web-common/src/features/dashboards/url-state/convert-partial-explore-state-to-url-params.ts
```diff
@@ -53,30 +54,23 @@ export function getCleanedUrlParamsForGoto(
...
-    const defaultValue = defaultExploreUrlParams.get(key);
-    if (
-      (defaultValue === null && value !== "") ||
-      (defaultValue !== null && value !== defaultValue)
-    ) {
-      return;
-    }
-    stateParams.delete(key);
-  });
+  const strippedUrlParams = stripDefaultUrlParams(
```
ericpgreen2, line 57: Naming: I'd expect the "strippedURLParams" to be the URL params _removed_, not the ones that are retained

web-common/src/features/dashboards/url-state/url-params-strip-utils.ts
```diff
@@ -0,0 +1,38 @@
...
+    if (
+      // if there is no default value then skip setting if the value is empty then skip adding it.
+      (defaultValue === null && value === "") ||
+      // else if there is a default value,
+      (defaultValue !== null &&
+        // make sure it is not one of the common param name but different meaning
+        !isParamCommonButDifferentMeaning(currentView, defaultView, key) &&
+        // and values match then skip adding it
+        value === defaultValue)
+    ) {
```
ericpgreen2, line 31: Please use explanatory variables rather than have so many complex conditions + comments packed into the `if (...)` clause

ericpgreen2 (changes requested):

ericpgreen2 (approved):

PR #7165, opened 2025-04-17
fix: between measure filter converting to advance filter

AdityaHegde:
Closes https://github.com/rilldata/rill-private-issues/issues/1541

We had a regression where between measure filter was being identified as an advanced filter preventing user from editing it.

**Checklist:**
- [x] Covered by tests
- [ ] I'm proud of this work!


web-common/src/features/dashboards/stores/filter-utils.ts
```diff
@@ -359,6 +377,18 @@ export function isJoinerExpression(expression: V1Expression | undefined) {
...
+  expression: V1Expression | undefined,
+) {
+  if (
+    !expression ||
+    !isJoinerExpression(expression) ||
+    (expression.cond?.exprs?.length && expression.cond.exprs.length > 1)
+  )
+    return expression;
+  return expression.cond?.exprs?.[0];
+}
```
ericpgreen2, line 390: This is fairly clear, but I'm wondering if this implies that we have a problem elsewhere in the code? And what's a "joiner" expression?

web-common/src/features/dashboards/filters/measure-filters/measure-filter-utils.ts
```diff
@@ -39,21 +40,24 @@ export function splitWhereFilter(whereFilter: V1Expression | undefined) {
...
-        filters: filters.filter(Boolean) as MeasureFilterEntry[],
-      });
+    if (!subqueryExpr?.subquery?.having) {
+      dimensionFilters.cond?.exprs?.push(e);
       return;
     }
 
-    dimensionFilters.cond?.exprs?.push(e);
+    const isValidMeasureFilter =
+      !isJoinerExpression(subqueryExpr.subquery.having) ||
```
ericpgreen2, line 49: What's this condition mean? I'm not clear what a "joiner" expression is. Looking only at this line, I'm not sure it's right that a valid measure filter is equal to any expression that is not a "joiner"? 

What about including something like this:
```typescript
whereFilter?.cond?.exprs?.forEach((expr) => {
    if (isMeasureFilter(expr)) {
      dimensionThresholdFilters.push(createDimensionThresholdFilter(expr));
    } else {
      dimensionFilters.cond?.exprs?.push(expr);
    }
```

AdityaHegde: I have renamed "joiner" expression to clarify what it is. A measure filter supported in UI could be something like `mes > 10` this is definitely supported right now and it not a "joiner".

web-common/src/features/dashboards/stores/filter-utils.ts
AdityaHegde: The type supports multiple measure filter per dimension. That is why there is always an and at the top level. Moves things around and added comments to make things clearer.

ericpgreen2: If there's only 1 measure (or any other) filter, then we shouldn't have a wrapper AND expression. (That may be out-of-scope for this PR– if that's the case, so be it.)

web-common/src/features/dashboards/filters/measure-filters/measure-filter-entry.spec.ts
```diff
@@ -20,7 +20,7 @@ import { describe, expect, it } from "vitest";
 
 const TestCases: [
   title: string,
-  measureFilter: MeasureFilterEntry,
+  measureFilter: MeasureFilterEntry | undefined,
```
ericpgreen2, line 23: Why are there test cases where `MeasureFilterEntry` is `undefined`? Is the point to test arbitrary filters passed through the URL? Or something else?

web-common/src/features/dashboards/stores/filter-utils.ts
```diff
@@ -351,14 +369,26 @@ export function isExpressionIncomplete(expression: V1Expression): boolean {
...
-export function isJoinerExpression(expression: V1Expression | undefined) {
+export function isAndOrExpression(expression: V1Expression | undefined) {
   return (
     expression?.cond?.op &&
     (expression.cond.op === V1Operation.OPERATION_AND ||
       expression.cond.op === V1Operation.OPERATION_OR)
   );
 }
 
+export function unwrapRedundantAndOrExpression(
```
ericpgreen2, line 380: Naming nit: `removeWrapperAndOrExpression`

ericpgreen2 (approved):

web-common/src/features/dashboards/filters/measure-filters/measure-filter-entry.spec.ts
AdityaHegde: Ya there is one case where we return undefined when it is not supported.

PR #7172, opened 2025-04-18
Re-validate metrics on external tables when running `rill project refresh`

begelundmuller:
**Changes:**
- Running `rill project refresh [--all]` now also directly reconciles metrics and dashboards instead of only models. This means metrics based on external tables also get re-validated.
- Also fixes a bug where refreshing multiple metrics with `rill project refresh` in one go would only refresh the first one.

Closes https://github.com/rilldata/rill-private-issues/issues/1488

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!


runtime/server/controller.go
```diff
@@ -369,28 +369,33 @@ func (s *Server) CreateTrigger(ctx context.Context, req *runtimev1.CreateTrigger
...
-		}
-
-		// Add all models.
-		rs, err = ctrl.List(ctx, runtime.ResourceKindModel, "", false)
-		if err != nil {
-			return nil, status.Error(codes.InvalidArgument, fmt.Errorf("failed to list models: %w", err).Error())
+	// Handle the convenience flags for all (user-facing) resources.
+	// In practice, we only refresh the major user declared resources that impact serving.
+	// For example, we don't currently trigger alerts or reports.
+	if req.All || req.AllFull {
```
k-anshul, line 375: What about connectors ?

k-anshul (approved):

runtime/server/controller.go
begelundmuller: Good point, added them as well

PR #1553, opened 2023-01-03
Make time range filter inclusive for end part

primenumsdev:
Hi,

I like this product, and especially UI, looks very clean.

I noticed that time range filter doesn't include the very end date and filters out last row of data in the dashboard metrics.

See examples in screenshots:

image

PayOut shows -199.4 and it doesn't include the very last row item, that should substitute another -1000.
Here is the query that shows the correct sum:
image

This PR to fix the end date filter to be inclusive.

PR #7178, opened 2025-04-21
handle divide by zero causing null value to appear on top

NamanMahor:
Close [#1553](https://github.com/rilldata/rill-private-issues/issues/1553)

Actually, the null-like value appears at the top because it's not a real `NULL` — it's a `NaN`. We later convert it to `NULL` in the Go code. According to the DuckDB documentation:

`NaN compares equal to NaN and is greater than any other floating-point number.`

This means NaN will sort higher than any valid numeric value, which explains why it shows up first in sorted results.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


begelundmuller (approved):
LGTM

PR #7201, opened 2025-04-22
Support AWS STS in DUCKDB and CH connectors

grahamplata:
Support AWS STS in DUCKDB and CH connectors 

Resolves: https://github.com/rilldata/rill-private-issues/issues/1430

**Checklist:**
- [ ] Covered by tests
- [ ] Checked if the docs need to be updated - Will add after review
- [ ] I'm proud of this work!


runtime/drivers/clickhouse/clickhouse.go
```diff
@@ -124,6 +127,11 @@ type configProperties struct {
 	ConnMaxLifetime string `mapstructure:"conn_max_lifetime"`
 	// ReadTimeout is the maximum amount of time a connection may be reused. Default is 300s.
 	ReadTimeout string `mapstructure:"read_timeout"`
+
+	// AWS S3 related configuration
+	AWSRoleARN                string `mapstructure:"aws_role_arn"`
+	AWSRoleSessionName        string `mapstructure:"aws_role_session_name"`
+	AWSSessionDurationSeconds *int32 `mapstructure:"aws_session_duration_seconds"`
```
begelundmuller, line 134: 1. It feels wrong for these properties to be configured on the Clickhouse connector directly. Shouldn't they be implemented on the S3 connector (i.e. `runtime/drivers/s3.ConfigProperties{}`)? And then perhaps accessed from the Clickhouse driver in the `objectStoreToSelfExecutor`?
2. Why does Clickhouse have a different set of config properties for this than in the DuckDB driver?

runtime/drivers/clickhouse/clickhouse.go
```diff
@@ -310,6 +318,33 @@ func (d driver) Open(instanceID string, config map[string]any, st *storage.Clien
...
+	// If AWSRoleARN is provided, assume the role and cache credentials
+	if conf.AWSRoleARN != "" {
+		// Check cache first (lock, check, unlock pattern)
+		c.awsCredentialsCacheMu.Lock()
+		awsCreds := c.awsCredentials
+		c.awsCredentialsCacheMu.Unlock()
+
+		if awsCreds != nil && awsCreds.Expiration.After(time.Now().Add(5*time.Minute)) {
+			logger.Info("using cached AWS credentials for S3 access", zap.String("role_arn", conf.AWSRoleARN), zap.Time("expiration", *awsCreds.Expiration))
+		} else {
```
begelundmuller, line 330: This is inside the `Open` function and `c` was just instantiated in the statement above here. So `c.awsCredentials` will always be `nil` and there's no caching/refreshing happening?

runtime/drivers/duckdb/duckdb.go
```diff
@@ -487,6 +493,55 @@ func (c *connection) reopenDB(ctx context.Context) error {
...
+				fmt.Sprintf("SET s3_endpoint='%s';", escapeDuckDBString(c.config.AWSEndpoint)))
+		}
+		if c.config.AWSRegion != "" {
+			connInitQueries = append(connInitQueries,
+				fmt.Sprintf("SET s3_region='%s';", escapeDuckDBString(c.config.AWSRegion)))
+		}
+		if c.config.AWSS3UseSSL {
+			connInitQueries = append(connInitQueries,
+				fmt.Sprintf("SET s3_use_ssl=%t;", c.config.AWSS3UseSSL))
+		}
```
begelundmuller, line 542: This way of configuring S3 in DuckDB is deprecated – they now recommend using `CREATE SECRET` (e.g. search for `CREATE OR REPLACE TEMPORARY SECRET` in `runtime/drivers/duckdb/model_executor_objectstore_self.go`).

But I'm also a little unsure if we want to be creating/refreshing DuckDB secrets like this – at least if we do, I think it should be explicit not implicit (e.g. something like `create_secrets: [s3]` in the model YAML). I would consider it good enough if we "only" support OIDC for the `objectStoreToSelfExecutor` executor, not managing it for DuckDB's native connectors.

runtime/drivers/duckdb/duckdb.go
```diff
@@ -730,3 +786,8 @@ func (c *connection) periodicallyCheckConnDurations(d time.Duration) {
 		}
 	}
 }
+
+// escapeDuckDBString escapes single quotes for DuckDB string literals.
+func escapeDuckDBString(s string) string {
+	return strings.ReplaceAll(s, "'", "''")
+}
```
begelundmuller, line 793: Use `drivers.DialectDuckDB.EscapeStringValue`

runtime/pkg/credentials/credentials.go
```diff
@@ -0,0 +1,52 @@
+package credentials
```
begelundmuller, line 1: I think the package name/path should indicate that it's AWS related. Just `runtime/pkg/credentials` sounds like it's related to Rill authentication.

Also, maybe this logic just best belongs in a file in `runtime/drivers/s3` for now?

runtime/drivers/clickhouse/clickhouse.go
grahamplata: I missed the s3 driver on first pass. Agreed that makes the most sense.

runtime/drivers/duckdb/duckdb.go
grahamplata: The legacy handling better fit the need I hadn't realized it was fully deprecated. I found interacting with the additional manager middleman was awkward. I am fine with just objectStoreToSelfExecutor for the time being

begelundmuller (commented):
Questions:
1. How will this work in practice in Rill Cloud? The `config.LoadDefaultConfig` loads credentials from the environment. So the idea is that we configure an AWS service account managed by us on the runtimes?
2. How do we mitigate the [confused deputy problem](https://docs.aws.amazon.com/IAM/latest/UserGuide/confused-deputy.html)? As far as I can tell, the `aws_role_arn` and `aws_role_session_name` parameters aren't secrets, so they alone are not sufficient to mitigate the problem, right?
3. We'll need a section in the docs about how to set this up for the user, and I think also internal setup/runbook of how to configure the service account on the runtimes?

grahamplata:
> Questions:
> 
> 1. How will this work in practice in Rill Cloud? The `config.LoadDefaultConfig` loads credentials from the environment. So the idea is that we configure an AWS service account managed by us on the runtimes?
> 2. How do we mitigate the [confused deputy problem](https://docs.aws.amazon.com/IAM/latest/UserGuide/confused-deputy.html)? As far as I can tell, the `aws_role_arn` and `aws_role_session_name` parameters aren't secrets, so they alone are not sufficient to mitigate the problem, right?
> 3. We'll need a section in the docs about how to set this up for the user, and I think also internal setup/runbook of how to configure the service account on the runtimes?

Messaged out of band

docs/docs/reference/connectors/s3.md
```diff
@@ -74,6 +85,39 @@ If you've configured credentials locally already (in your `<RILL_PROJECT_DIRECTO
...
+- **AWS Role ARN** (`aws_role_arn`) - The Amazon Resource Name (ARN) of the role to assume
+- **AWS Role Session Name** (`aws_role_session_name`) - Optional name for the role session (defaults to "rill-session")
+- **AWS External ID** (`aws_external_id`) - Optional external ID for cross-account access
+
+### Setting Up an IAM Role for Rill
+
+1. **Create an IAM Role**:
+   - In the AWS Management Console, navigate to IAM > Roles > Create role
+   - Select "Another AWS account" as the trusted entity
+   - Enter your AWS account ID or the account ID that Rill will use
```
begelundmuller, line 105: Isn't this supposed to be an AWS service account that Rill manages, not an account managed/created by the user?

docs/docs/reference/connectors/s3.md
```diff
@@ -74,6 +85,39 @@ If you've configured credentials locally already (in your `<RILL_PROJECT_DIRECTO
...
+   - Optionally check "Require external ID" and provide a unique value that Rill will use
```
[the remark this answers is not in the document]
grahamplata: Ahh my apologies you are correct

docs/docs/reference/connectors/s3.md
```diff
@@ -14,6 +14,18 @@ sidebar_position: 2
 <img src = '/img/reference/connectors/s3/s3.png' class='centered' />
 <br />
 
+## Authentication Options
+
+Rill supports several authentication methods for connecting to AWS S3:
+
+1. **Local AWS credentials** - Uses your local AWS configuration (from `~/.aws/credentials`)
+   1. Note: This only a local only feature.
```
begelundmuller, line 22: nit: language

docs/docs/reference/connectors/s3.md
```diff
@@ -184,3 +206,37 @@ Here is a step-by-step guide on how to create an AWS service account with read-o
...
+
+2. Navigate to the [IAM dashboard](https://console.aws.amazon.com/iam).
+
+3. In the sidebar, select "Roles" and click the "Create role" button.
+
+4. For trust relationship, select "AWS account" and choose "Another AWS account".
+
+5. Enter Rill's AWS account ID that was provided to you by your Rill representative.
+
+6. For enhanced security: Select "Require external ID" and enter the External ID provided by Rill. This helps prevent the confused deputy problem.
```
begelundmuller, line 224: I think we should just always do this, not optionally "for enhanced security"

begelundmuller (approved):

PR #7213, opened 2025-04-24
feat: Admin service support for managed github repo APIs 

k-anshul:
Subtask for : https://github.com/rilldata/rill-private-issues/issues/1569

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!


admin/database/database.go
```diff
@@ -1261,3 +1268,22 @@ type UpdateProvisionerResourceOptions struct {
...
+	RepositoryID    int64     `db:"repository_id"`
+	CreatedByUserID string    `db:"created_by_user_id"`
+	CreatedOn       time.Time `db:"created_on"`
+	UpdatedOn       time.Time `db:"updated_on"`
+}
+
+type InsertManagedGithubRepoMetaOptions struct {
+	OrgID           string  `validate:"required"`
+	CreatedByUserID *string `validate:"required"`
+	HTMLURL         string  `validate:"required"`
```
begelundmuller, line 1287: I know we've previously used html URLs, but I regret it a bit and it feels fragile. For this use case, should we perhaps store HTTP remotes instead? E.g. `https://github.com/account/repo.git`. It would make configuring remotes simpler, and would be better compatible with other Git providers than Github (e.g. Gitlab allows extra levels of nesting for repository paths).

cli/cmd/admin/start.go
```diff
@@ -230,7 +231,7 @@ func StartCmd(ch *cmdutil.Helper) *cobra.Command {
 			emailClient := email.New(sender)
 
 			// Init github client
-			gh, err := admin.NewGithub(conf.GithubAppID, conf.GithubAppPrivateKey)
+			gh, err := admin.NewGithub(cmd.Context(), conf.GithubAppID, conf.GithubAppPrivateKey, conf.RillManagedGithubOrg)
```
begelundmuller, line 234: Not necessarily something to do now – but from an abstraction perspective, it feels weird to use the same client for user repositories and managed repositories. Would it make sense to initialize a separate client for managed Git repositories? E.g. `admin.Github` and `admin.ManagedGit` (incidentally backed by Github, but could be swapped for e.g. Google Cloud Source Repositories)?

Ideally we'd have an `admin/git/provider.go` interface with a common interface for Git providers. That'd also make supporting other providers easier.

begelundmuller (changes requested):

admin/database/postgres/migrations/0067.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE IF NOT EXISTS managed_github_repo_meta (
```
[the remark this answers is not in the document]
k-anshul: I just added to make it fail proof if we switch between versions. I can remove it as well.

proto/rill/admin/v1/api.proto
```diff
@@ -1914,6 +1922,18 @@ message ConnectProjectToGithubRequest {
...
+  string organization = 1;
+  // name of the repo to create. 
+  // Note: The final name will be suffixed with a random string to ensure uniqueness.
+  string name = 2;
+}
+
+message CreateManagedGithubRepoResponse {
+  string repo_html_url = 1;
+  string token = 2;
+}
```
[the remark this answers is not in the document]
k-anshul: The problem with remote is that it doesn't explicitly imply what URL we are returning whether `html_url`/`git_url`/`clone_url` etc.
But it is a minor detail and remote is fine as well.

cli/cmd/admin/start.go
```diff
@@ -77,6 +77,7 @@ type Config struct {
 	GithubAppWebhookSecret    string                 `split_words:"true"`
 	GithubClientID            string                 `split_words:"true"`
 	GithubClientSecret        string                 `split_words:"true"`
+	RillManagedGithubOrg      string                 `split_words:"true"`
```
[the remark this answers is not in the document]
k-anshul: 1. Our github app is already asking permissiong for creating/deleting repositories so it should be fine to use the same app to avoid maintaining two apps:
image

2. Sounds good.

admin/github.go
```diff
@@ -28,22 +32,38 @@ var (
...
 	appClient     *github.Client
+	managedOrg    string
+
+	// managedOrgInstallationID is usually populated when the client is created.
+	// But we do not return an error if there is any error in fetching the installation ID.
+	// This is to let admin server start even if there is an issue with Github service.
+	// If managedOrgFetchError is true then the installation ID is not valid and refetched again when needed.
+	// mu controls access to managedOrgInstallationID and managedOrgFetchError.
+	managedOrgInstallationID int64
+	managedOrgFetchError     error
```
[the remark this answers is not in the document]
k-anshul: > In what cases does this get set? Seems like a pretty serious error.

`managedOrgInstallationID` is fetched in server startup. If it fails `managedOrgFetchError` is set. If `managedOrgFetchError` is set to non nil then ID will be refetched on the next request. Once `managedOrgInstallationID` is set, `managedOrgFetchError` is set to nil and it is never refetched again. 

I can add a log on failure.

admin/server/assets.go
```diff
@@ -264,7 +264,7 @@ func (s *Server) UploadProjectAssets(ctx context.Context, req *adminv1.UploadPro
 		return nil, err
 	}
 
-	token, err := s.admin.Github.InstallationToken(ctx, *proj.GithubInstallationID)
+	token, err := s.admin.Github.InstallationToken(ctx, *proj.GithubInstallationID, 0)
```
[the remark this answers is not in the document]
k-anshul: Given the comments above we can avoid setting it 0 ever.

admin/database/database.go
```diff
@@ -1261,3 +1268,22 @@ type UpdateProvisionerResourceOptions struct {
 	State         map[string]any
 	Config        map[string]any
 }
+
+// ManagedGithubRepoMeta represents metadata about a Rill managed GitHub repository for projects deployed on Rill Cloud.
+type ManagedGithubRepoMeta struct {
```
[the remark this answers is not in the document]
k-anshul: `ManagedGitRepoMeta` indicates that it just a metadata about the managed git repo and the actual data is stored somewhere and not in this table/database. That is why I added the extra `meta` suffix.

`ManagedGitRepo` sounds like the managed git repository is stored here itself.
However I am fine with dropping the suffix to make it inline with other tables.

admin/database/database.go
```diff
@@ -1261,3 +1268,22 @@ type UpdateProvisionerResourceOptions struct {
...
+	ID              string    `db:"id"`
+	OrgID           string    `db:"org_id"`
+	ProjectID       *string   `db:"project_id"`
```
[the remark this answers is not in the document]
k-anshul: I kept `org_id` to easily fetch repo for an org(normalization vs simplcity). But I can remove it as well. Just need an extra join.

Yeah I think storing the relationship other way is also good. That way we no longer need to maintain a bool indicating that this project uses a managed git repo. Though we can continue to duplicate `remote` in both places to keep rest of the code consistent.

admin/database/database.go
k-anshul: In the follow up to this PR, we also need to set this in project URL and it may be inconvenient to see both `http_url` and `clone_url` in the same column. 
I can make it `remote` and set `clone_url` and then do handling at places wherever it is required.

cli/cmd/admin/start.go
k-anshul: I think we are in a complicated situation here. 
Github and other providers do not confirm to a common implementation. 
Things like installationID, installation token and several such constructs may not be applicable to other providers. So we have two choices here :
a) Either embrace the fact that github is going to be an important part of our workflows and make it strongly coupled within our code.
b) Try to find some middle ground for all these providers and keep it completely decoupled. But then we also need to make other changes in the code like saving `GithubInstallationID` with projects.

admin/database/database.go
```diff
@@ -308,6 +308,13 @@ type DB interface {
 	InsertProvisionerResource(ctx context.Context, opts *InsertProvisionerResourceOptions) (*ProvisionerResource, error)
 	UpdateProvisionerResource(ctx context.Context, id string, opts *UpdateProvisionerResourceOptions) (*ProvisionerResource, error)
 	DeleteProvisionerResource(ctx context.Context, id string) error
+
+	FindManagedGithubRepoMeta(ctx context.Context, htmlURL string) (*ManagedGithubRepoMeta, error)
```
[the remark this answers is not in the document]
k-anshul: I mentioned it in another reply but the problem is that a lot of code is already strongly coupled with Github and we don't know what common ground exists between Github and Gitlab and other providers so decoupling it completely is probably not as straightforward. 

Hence as an extension it just felt intuitive to mark it `Github` and not `Git` but I am fine with removing `Github` from managed and made changes in the PR.

cli/cmd/admin/start.go
k-anshul: Maybe lets evaluate option 2 in a separate task.

admin/github.go
```diff
@@ -54,12 +73,24 @@ func NewGithub(appID int64, appPrivateKey string) (Github, error) {
...
+	}
+
+	// Set the managed org installation client
+	i, _, err := appClient.Apps.FindOrganizationInstallation(ctx, managedGithubOrg)
+	if err != nil {
+		logger.Error("failed to get managed org installation ID", zap.Error(err), observability.ZapCtx(ctx))
+		g.managedOrgFetchError = err
+		return g, nil
+	}
+	g.managedOrgInstallationID = *i.ID
```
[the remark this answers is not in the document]
begelundmuller: Also, if `managedGithubOrg == ""`, i.e. is not configured, we should handle it gracefully by skipping the network call and saying "managed Git repositories are not configured for this environment"

admin/server/projects.go
```diff
@@ -1233,13 +1241,24 @@ func (s *Server) GetCloneCredentials(ctx context.Context, req *adminv1.GetCloneC
...
+	token, err := s.admin.Github.InstallationToken(ctx, *proj.GithubInstallationID, repoID)
 	if err != nil {
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
+	var cloneURL string
+	if strings.HasSuffix(*proj.GithubURL, ".git") {
+		cloneURL = *proj.GithubURL
+	} else {
+		cloneURL = *proj.GithubURL + ".git"
+	}
```
begelundmuller, line 1258: Would it make sense to run a migration that does this?

admin/database/postgres/migrations/0067.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE managed_git_repo (
+    id UUID DEFAULT uuid_generate_v4() PRIMARY KEY,
+    remote TEXT NOT NULL UNIQUE,
+    created_by_user_id UUID REFERENCES users (id) ON DELETE SET NULL,
+    created_on TIMESTAMP DEFAULT NOW(),
+    updated_on TIMESTAMP DEFAULT NOW()
+);
+
+ALTER TABLE PROJECTS ADD column managed_git_repo_id UUID REFERENCES managed_git_repo (id) ON DELETE RESTRICT;
```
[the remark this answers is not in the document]
k-anshul: Sure. But why do you think is leniency required here ? We should prevent deleting `managed_git_repo` if it is still being used in project.

admin/server/projects.go
k-anshul: We can do it in future but I think we still need to commit this change and run migration later ? Otherwise between the migration runs and code is deployed it will break.

admin/server/github.go
```diff
@@ -381,6 +381,49 @@ func (s *Server) ConnectProjectToGithub(ctx context.Context, req *adminv1.Connec
...
+	}
+	token, err := s.admin.Github.InstallationToken(ctx, id, *repo.ID)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
+
+	return &adminv1.CreateManagedGitRepoResponse{
+		Remote:   *repo.CloneURL,
+		Username: "x-access-token",
+		Password: token,
```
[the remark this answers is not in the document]
k-anshul: There are some minor variations here and there. Let us keep it hardcoded for now.

admin/database/postgres/postgres.go
```diff
@@ -388,9 +388,9 @@ func (c *connection) InsertProject(ctx context.Context, opts *database.InsertPro
 
 	res := &projectDTO{}
 	err := c.getDB(ctx).QueryRowxContext(ctx, `
-		INSERT INTO projects (org_id, name, description, public, created_by_user_id, provisioner, prod_olap_driver, prod_olap_dsn, prod_slots, subpath, prod_branch, archive_asset_id, github_url, github_installation_id, prod_ttl_seconds, prod_version)
-		VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10, $11, $12, $13, $14, $15, $16) RETURNING *`,
-		opts.OrganizationID, opts.Name, opts.Description, opts.Public, opts.CreatedByUserID, opts.Provisioner, opts.ProdOLAPDriver, opts.ProdOLAPDSN, opts.ProdSlots, opts.Subpath, opts.ProdBranch, opts.ArchiveAssetID, opts.GithubURL, opts.GithubInstallationID, opts.ProdTTLSeconds, opts.ProdVersion,
+		INSERT INTO projects (org_id, name, description, public, created_by_user_id, provisioner, prod_olap_driver, prod_olap_dsn, prod_slots, subpath, prod_branch, archive_asset_id, github_url, github_installation_id, github_repo_id, prod_ttl_seconds, prod_version)
```
begelundmuller, line 391: The `managed_git_repo_id` appears not to be set/used here or in the `UpdateProject` statements? If it's always empty, do we need it at all?

admin/database/postgres/postgres.go
```diff
@@ -2539,6 +2539,69 @@ func (c *connection) DeleteProvisionerResource(ctx context.Context, id string) e
...
+}
+
+func (c *connection) FindUnusedManagedGitRepos(ctx context.Context, pageSize int) ([]*database.ManagedGitRepo, error) {
+	// find managed github repos that are not associated with any project
+	// skip repos that are less than 7 days old to avoid deleting repos for projects
+	// that were accidentally deleted and may need to be restored
+	var res []*database.ManagedGitRepo
+	err := c.getDB(ctx).SelectContext(ctx, &res, `
+		SELECT * FROM managed_git_repos
+		WHERE project_id IS NULL
```
begelundmuller, line 2558: The `managed_git_repos` doesn't have a `project_id` column atm?

admin/database/postgres/migrations/0068.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE managed_git_repos (
+    id UUID DEFAULT uuid_generate_v4() PRIMARY KEY,
+    remote TEXT NOT NULL UNIQUE,
+    created_by_user_id UUID REFERENCES users (id) ON DELETE SET NULL,
+    created_on TIMESTAMP DEFAULT NOW(),
+    updated_on TIMESTAMP DEFAULT NOW()
+);
```
begelundmuller, line 7: Shouldn't it have either a `project_id` or `org_id` column? 

Maybe `org_id` would make sense? Then it would be similar to `assets`, i.e. the deploy process would consist of:
1. Creating a managed Git repo in the org and pushing to it
2. Creating a project that uses the managed Git repo

But open to better ideas if you have any?

admin/github.go
```diff
@@ -54,12 +70,27 @@ func NewGithub(appID int64, appPrivateKey string) (Github, error) {
...
+	// Set the managed org installation client
+	if githubManagedAcct == "" {
+		g.managedOrgFetchError = fmt.Errorf("managed Git repositories are not configured for this environment")
+	}
+	i, _, err := appClient.Apps.FindOrganizationInstallation(ctx, githubManagedAcct)
+	if err != nil {
+		logger.Error("failed to get managed org installation ID", zap.Error(err), observability.ZapCtx(ctx))
+		g.managedOrgFetchError = err
+		return g, nil
+	}
```
begelundmuller, line 90: Don't you need an `else` here? Otherwise `g.managedOrgFetchError` would always get overridden with the Github error for empty repo name?

begelundmuller (changes requested):

admin/database/postgres/migrations/0067.sql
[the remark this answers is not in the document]
begelundmuller: Leniency might make some handling when switching from managed to external Git, or since deletes happen in Github before they happen in Postgres. Honestly not sure about it, just wanted to ask.

admin/database/postgres/migrations/0068.sql
k-anshul: I implemented the review comment suggested [here](https://github.com/rilldata/rill/pull/7213#discussion_r2071764411) and that seems to have broken the flow.
a) Removed the `org_id` 
b) Made the relationship reverse i.e. projects now hold a reference to managed_git_repo

We need `org_id` here to check the count of repos.

admin/database/postgres/postgres.go
k-anshul: This will be used in subsequent PRs when we use the managed git repos.

admin/server/projects.go
```diff
@@ -1217,6 +1222,12 @@ func (s *Server) GetCloneCredentials(ctx context.Context, req *adminv1.GetCloneC
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 
+	claims := auth.GetClaims(ctx)
+	if !claims.Superuser(ctx) && !claims.ProjectPermissions(ctx, proj.OrganizationID, proj.ID).ManageProject {
```
begelundmuller, line 1226: Now that this endpoint is no longer just for superusers, can you add a `force_superuser_access` flag and check it similar to this: https://github.com/rilldata/rill/pull/7283

admin/server/projects.go
```diff
@@ -1859,6 +1883,53 @@ func (s *Server) hasAssetUsagePermission(ctx context.Context, id, orgID, ownerID
...
+// For older projects this may be nil since the github repo id was not stored.
+// This function fetches the ID from github API and stores it in the database.
+// It assumes that the project is connected to github and necessary checks have been done by the caller.
+func (s *Server) githubRepoIDForProject(ctx context.Context, p *database.Project) (int64, error) {
+	if p.GithubRepoID != nil {
+		return *p.GithubRepoID, nil
+	}
+
+	client, err := s.admin.Github.InstallationClient(*p.GithubInstallationID)
+	if err != nil {
```
begelundmuller, line 1896: Maybe just to be safe, can it return an error if `p.GithubInstallationID == nil` (sounds like it would be an internal error)?

begelundmuller (changes requested):

begelundmuller (approved):

Issue #6667, opened 2025-02-12
Cloud UI: apply `refetchInterval` function to all resource tables


ericpgreen2:
Apply the `refetchInterval` function created in [this PR](https://github.com/rilldata/rill/pull/6175) to all resources tables:
- The Dashboards table
- The Reports table
- The Alerts table 

After, edit [this e2e test](https://github.com/rilldata/rill/blob/main/web-admin/tests/setup/setup.ts#L177-L189) to remove the manual page reload.

PR #7216, opened 2025-04-24
Enable `refetchIntervalInBackground ` to all resource tables

lovincyrus:
Closes https://github.com/rilldata/rill/issues/6667

In [this pull request](https://github.com/rilldata/rill/pull/6175), we introduced a custom refetch interval to repeatedly reload a resource until its status was marked as complete. However, for dashboards, reports, and alerts tables, this approach isn't required. We can instead utilize a built-in TanStack function, `refetchIntervalInBackground`, which ensures data is refetched even when the tab is in the background.

The current changeset only includes updates to the `dashboards` query. Seeking @ericpgreen2 thoughts before implementing reports and alerts.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-admin/src/features/dashboards/listing/DashboardsTable.svelte
+++ b/web-admin/src/features/dashboards/listing/DashboardsTable.svelte
@@ -10,12 +10,18 @@
   import { useDashboardsV2, type DashboardResource } from "./selectors";
   import NoResourceCTA from "@rilldata/web-admin/features/projects/NoResourceCTA.svelte";
   import ResourceError from "@rilldata/web-admin/features/projects/ResourceError.svelte";
+  import { isDashboardPage } from "../../navigation/nav-utils";
+  import { page } from "$app/stores";
 
   export let isEmbedded = false;
 
   $: ({ instanceId } = $runtime);
 
-  $: dashboards = useDashboardsV2(instanceId);
+  $: onDashboardPage = isDashboardPage($page);
+
+  $: dashboards = useDashboardsV2(instanceId, {
+    enableBackgroundRefetch: onDashboardPage,
+  });
 
   /**
    * Table column definitions.
--- a/web-admin/src/features/dashboards/listing/selectors.ts
+++ b/web-admin/src/features/dashboards/listing/selectors.ts
@@ -3,7 +3,7 @@ import { useValidExplores } from "@rilldata/web-common/features/dashboards/selec
 import { getMapFromArray } from "@rilldata/web-common/lib/arrayUtils";
 import type { V1Resource } from "@rilldata/web-common/runtime-client";
 import { createRuntimeServiceListResources } from "@rilldata/web-common/runtime-client";
-import type { CreateQueryResult } from "@tanstack/svelte-query";
+import type { CreateQueryResult, Query } from "@tanstack/svelte-query";
 import { derived } from "svelte/store";
 
 export function useDashboardsLastUpdated(
@@ -49,6 +49,9 @@ export interface DashboardResource {
 // This iteration of `useDashboards` returns the above `DashboardResource` type, which includes `refreshedOn`
 export function useDashboardsV2(
   instanceId: string,
+  options?: {
+    enableBackgroundRefetch?: boolean;
+  },
 ): CreateQueryResult<DashboardResource[]> {
   return createRuntimeServiceListResources(instanceId, undefined, {
     query: {
@@ -79,6 +82,7 @@ export function useDashboardsV2(
         );
         return allDashboards;
       },
+      refetchIntervalInBackground: options?.enableBackgroundRefetch,
     },
   });
 }
--- a/web-admin/src/features/navigation/nav-utils.ts
+++ b/web-admin/src/features/navigation/nav-utils.ts
@@ -13,6 +13,10 @@ export function withinOrganization(page: Page): boolean {
   return !!page.route?.id?.startsWith("/[organization]");
 }
 
+export function isDashboardPage(page: Page): boolean {
+  return !!(page.route.id === "/[organization]/[project]");
+}
+
 export function isProjectPage(page: Page): boolean {
   return (
     page.route.id === "/[organization]/[project]" ||


briangregoryholmes (approved):

web-admin/tests/setup/setup.ts
ericpgreen2, line 1: What about the polling happening underneath the `// Check that the dashboards are listed` comment? The original motivation of the GitHub issue was to remove the need to poll there.

ericpgreen2 (changes requested):
> However, for dashboards, reports, and alerts tables, this approach isn't required

For dashboards especially, we _would_ like to poll the `ListResources` API when the resource has not finished reconciling. This will particularly help the UX when deploying projects (see my comment in the `setup.ts` file).

web-admin/tests/setup/setup.ts
```diff
@@ -232,35 +223,17 @@ setup.describe("global setup", () => {
...
-          return listing.textContent();
-        },
-        { intervals: Array(6).fill(5_000), timeout: 30_000 },
-      )
-      .toContain("Last refreshed");
+    await expect(
+      adminPage.getByRole("link", {
+        name: "Programmatic Ads Bids bids_explore",
+      }),
+    ).toContainText("Last refreshed");
```
ericpgreen2, line 236: I think these `expect` assertion timeouts are 5s. That might be a little tight. Can you increase them to 15s? E.g.:
```
).toContainText("Last refreshed", { timeout: 15_000 });
```

ericpgreen2 (changes requested):
Rather than repeating the refetch function across each resource file, can we put it in one central location? Maybe `features/projects` would be the right directory?

ericpgreen2:
Additionally, mind updating the PR description?

web-admin/src/features/alerts/selectors.ts
```diff
@@ -21,6 +27,19 @@ export function useAlerts(instanceId: string, enabled = true) {
...
+            query.state.data,
+            query,
+          );
+          if (newInterval === false) {
+            currentRefetchInterval = INITIAL_REFETCH_INTERVAL;
+            return false;
+          }
+          currentRefetchInterval = newInterval;
+          return newInterval;
+        },
```
ericpgreen2, line 42: Tracking state outside of the `calculateRefetchInterval` function makes this more complex than it feels like it should be. I'm wondering if there's a way `calculateRefetchInterval` could encompass all of the logic, so we could call it like this:
```
query: {
    ...
    refetchInterval: calculateRefetchInterval // more descriptive name could be something like `pollUntilResourcesReconciled`
}
```
One idea, could we add custom state to either `query.meta` or `query.state`, e.g. `query.state.currentRefetchInterval = X`?

ericpgreen2 (changes requested):
When I refetch a resource on the Project Status page, I see double network requests: one with `skipSecurityChecks`, one without. I'm expecting that only the `skipSecurityChecks` one would be refetched.
image

web-admin/src/features/alerts/selectors.ts
lovincyrus: Agreed that we should keep the refetchInterval state encapsulated within the component.

web-admin/src/features/shared/refetch-interval.ts
```diff
@@ -0,0 +1,79 @@
...
+): number | false {
+  if (query.state.error) return false;
+  if (!data?.resources) return false; // Stop polling if no resources
+
+  const hasErrors = data.resources.some(isResourceErrored);
+  const hasReconcilingResources = data.resources.some(isResourceReconciling);
+
+  if (hasErrors || !hasReconcilingResources) {
+    return false;
+  }
```
ericpgreen2, line 37: It looks like if any resource is errored, we stop polling. But, _if any resources are reconciling, we should keep polling_.

web-admin/src/features/shared/refetch-interval.ts
```diff
@@ -0,0 +1,79 @@
...
+  return Math.min(currentInterval * BACKOFF_FACTOR, MAX_REFETCH_INTERVAL);
+}
+
+function createRefetchIntervalStore() {
+  const store: Writable<number> = writable(INITIAL_REFETCH_INTERVAL);
+  let currentValue = INITIAL_REFETCH_INTERVAL;
+
+  store.subscribe((value) => {
+    currentValue = value;
+  });
```
ericpgreen2, line 48: This synchronization between the `store` and `currentValue` is a bit confusing. Can we just use one of these, so it's clear what's the source-of-truth?

web-admin/src/features/shared/refetch-interval.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  return {
+    subscribe: store.subscribe,
+    reset: () => {
+      store.set(INITIAL_REFETCH_INTERVAL);
+      return INITIAL_REFETCH_INTERVAL;
+    },
```
ericpgreen2, line 55: I don't see this used anywhere

web-admin/src/features/shared/refetch-interval.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  if (hasErrors || !hasReconcilingResources) {
+    return false;
+  }
+
+  return Math.min(currentInterval * BACKOFF_FACTOR, MAX_REFETCH_INTERVAL);
+}
+
+function createRefetchIntervalStore() {
+  const store: Writable<number> = writable(INITIAL_REFETCH_INTERVAL);
```
ericpgreen2, line 43: > One idea, could we add custom state to either query.meta or query.state, e.g. query.state.currentRefetchInterval = X?

I'm curious, did you consider this^ idea?

ericpgreen2 (changes requested):

web-admin/src/features/shared/refetch-interval.ts
lovincyrus: removed the store code and used the query's state directly https://github.com/rilldata/rill/pull/7216/commits/7c1017755ef048fd2b9ce6bb0adffcf5e8421ae0

web-admin/src/features/shared/refetch-interval.ts
```diff
@@ -0,0 +1,46 @@
...
+  if (!hasReconcilingResources) {
+    return false;
+  }
+
+  // Get the current interval from the query's state
+  const currentInterval =
+    query.state.dataUpdateCount === 0
+      ? INITIAL_REFETCH_INTERVAL
+      : INITIAL_REFETCH_INTERVAL *
+        Math.pow(BACKOFF_FACTOR, Math.min(query.state.dataUpdateCount, 5));
```
ericpgreen2: Unfortunately, I don't think this works the way you intend. For as long as the query is mounted, `dataUpdateCount` only gets incremented– it never resets to 0.

Here's a console log showing this when I repeatedly click "refresh" on a model on the project status page:

image

That's why I originally suggested using _custom_ state:
> One idea, could we add custom state to either query.meta or query.state, e.g. query.state.currentRefetchInterval = X?

Thoughts?

lovincyrus: hmm,
CleanShot 2025-06-11 at 22 08 38@2x

will explore the external store

lovincyrus: latest:

https://github.com/user-attachments/assets/7b46cee8-1e22-4750-ac60-627aaf7d6b32

lovincyrus: I left the dev logs in. Once this approach is approved, I'll remove them.

lovincyrus: removed the comments, you'll have to `git checkout e44873a6bbafbcb2f92615324cf4a0e85bb375a8` to review with logs.

web-admin/src/features/shared/refetchIntervalStore.ts
ericpgreen2, line 1: Please use kebab case for Typescript file names.

web-admin/src/features/shared/refetch-interval.ts
ericpgreen2, line 1: Can we consolidate this file with `shared/refetch-interval-store.ts`?

web-admin/src/features/scheduled-reports/selectors.ts
```diff
@@ -16,6 +21,11 @@ export function useReports(instanceId: string, enabled = true) {
       query: {
         enabled: enabled && !!instanceId,
         refetchOnMount: true,
+        select: (data) => {
+          updateSmartRefetchInterval(data?.resources);
+          return data;
+        },
+        refetchInterval: () => get(refetchInterval),
```
ericpgreen2: It'd be better if the refetch interval logic were all encapsulated in the `refetchInterval` parameter. It's not very clean to add (half of) the refetch logic as a side effect to the `select` function.

web-admin/src/features/shared/refetchIntervalStore.ts
```diff
@@ -0,0 +1,41 @@
...
+  INITIAL_REFETCH_INTERVAL,
+  BACKOFF_FACTOR,
+  MAX_REFETCH_INTERVAL,
+  isResourceReconciling,
+} from "./refetch-interval";
+
+// Store for the current refetch interval
+export const refetchInterval = writable<number | false>(
+  INITIAL_REFETCH_INTERVAL,
+);
```
ericpgreen2: I don't think one global store will work. This prohibits us from having two queries on the same page that use the smart refetch logic. For example, both the `ProjectGlobalStatusIndicator` and e.g. the `DashboardsTable` components are mounted at the same time and should use the smart refetch logic for their respective queries.

Can you please try the `query.meta` approach?

Per the docs, the `meta` field is intended for user-defined state. Yes, it's usually static state, but dynamic state might still work.
image

ericpgreen2 (changes requested):

web-admin/src/features/shared/refetch-interval.ts
ericpgreen2: See [comment here](https://github.com/rilldata/rill/pull/7216#discussion_r2154962707)

nishantmonu51:
@lovincyrus : Any updates on this PR ? Whats next here?

ericpgreen2:
@lovincyrus, checking in on this PR, have you given any thought to these two comments?

image

lovincyrus:
> @lovincyrus, checking in on this PR, have you given any thought to these two comments?
> 
> image

Ah, I responded in another thread https://github.com/rilldata/rill/pull/7216#discussion_r2141743261 We can certainly reconsider them

lovincyrus:
> @lovincyrus : Any updates on this PR ? Whats next here?

I plan to revisit this after completing the data connector UX sprints.

lovincyrus:
Added the `query.meta` approach in the recent commits.

web-admin/src/features/shared/refetch-interval-store.ts
ericpgreen2, line 1: Rather than a `web-admin/features/shared` directory, can we mimic `web-common` and create a `web-admin/lib` directory?

web-admin/src/features/shared/refetch-interval-store.ts
```diff
@@ -0,0 +1,88 @@
+import type { V1Resource } from "@rilldata/web-common/runtime-client";
+
+export const INITIAL_REFETCH_INTERVAL = 200; // Start at 200ms for immediate feedback
+export const MAX_REFETCH_INTERVAL = 2_000; // Cap at 2s
+export const BACKOFF_FACTOR = 1.5;
+
+export function isResourceErrored(resource: V1Resource) {
+  return !!resource?.meta?.reconcileError;
+}
```
ericpgreen2: I don't see this function used anywhere

web-admin/src/features/shared/refetch-interval-store.ts
```diff
@@ -0,0 +1,88 @@
...
+  );
+}
+
+/**
+ * Updates the meta object with a smart refetch interval based on resource states.
+ * @param resources Array of resources to check
+ * @param prevMeta Previous meta object (should contain refetchInterval and wasReconciling)
+ * @returns Updated meta object with new refetchInterval and wasReconciling state
+ */
+export function updateSmartRefetchMeta(
```
ericpgreen2: Nit: I don't see this used outside this file, so it need not be exported

web-admin/src/features/shared/refetch-interval-store.ts
```diff
@@ -0,0 +1,88 @@
...
+    // No reconciling resources - stop polling
+    return { refetchInterval: false, wasReconciling: false };
+  }
+
+  // Resources are reconciling
+  if (!wasReconciling) {
+    // NEW reconciliation cycle - reset to initial interval
+    return {
+      refetchInterval: INITIAL_REFETCH_INTERVAL,
+      wasReconciling: hasReconciling,
```
ericpgreen2: Nit: for readability, it looks like we can explicitly set `wasReconciling: true`

web-admin/src/features/shared/refetch-interval-store.ts
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  // CONTINUING reconciliation cycle - apply backoff
+  const current =
+    typeof prevMeta.refetchInterval === "number"
+      ? prevMeta.refetchInterval
+      : INITIAL_REFETCH_INTERVAL;
+  const next = Math.min(current * BACKOFF_FACTOR, MAX_REFETCH_INTERVAL);
+
+  return { refetchInterval: next, wasReconciling: hasReconciling };
```
ericpgreen2: Here too, `wasReconciling: true`

web-admin/src/features/shared/refetch-interval-store.ts
```diff
@@ -0,0 +1,88 @@
...
+>();
+
+/**
+ * Creates a smart refetch interval function that uses a WeakMap to store state.
+ * This approach keeps refetch state per query without mutating the query object.
+ *
+ * @param query The TanStack query object
+ * @returns The refetch interval (number in ms or false to disable)
+ */
+export function createSmartRefetchInterval(query: any): number | false {
```
ericpgreen2: Nit: it'd be nice if `query` were an explicit type, not an `any` type

web-admin/src/features/shared/refetch-interval-store.ts
lovincyrus: https://github.com/rilldata/rill/pull/7216/commits/86f095f4b0a775c62930fd2a45afe9eafee15773

web-admin/src/lib/refetch-interval-store.ts
```diff
@@ -0,0 +1,96 @@
...
+>();
+
+/**
+ * Creates a smart refetch interval function that uses a WeakMap to store state.
+ * This approach keeps refetch state per query without mutating the query object.
+ *
+ * @param query The TanStack query object
+ * @returns The refetch interval (number in ms or false to disable)
+ */
+export function createSmartRefetchInterval(
```
AdityaHegde, line 73: The usage of this is just, `refetchInterval: createSmartRefetchInterval`
Perhaps the name create would have matched if it was `refetchInterval: createSmartRefetchInterval(...)`
Just `smartRefetchInterval` might be better?

web-admin/tests/setup/setup.ts
```diff
@@ -222,19 +222,10 @@ setup.describe("global setup", () => {
...
-        async () => {
-          await adminPage.reload();
-          const title = adminPage.getByLabel("Container title");
-          return title.textContent();
-        },
-        { intervals: Array(5).fill(1_000), timeout: 5_000 },
-      )
-      .toEqual("Project dashboards");
+    // Wait for the project to be ready
+    await expect(adminPage.getByLabel("Container title")).toHaveText(
```
AdityaHegde, line 226: Great to see these fixed!

Could you check `web-integration/tests/deploy.spec.ts` as well? It has a similar reloading assertion at the bottom.

web-admin/src/lib/refetch-interval-store.ts
```diff
@@ -0,0 +1,96 @@
...
+ */
+function updateSmartRefetchMeta(
+  resources: any[] | undefined,
+  prevMeta: { refetchInterval?: number | false; wasReconciling?: boolean } = {},
+): { refetchInterval: number | false; wasReconciling: boolean } {
+  if (!resources) {
+    return { refetchInterval: false, wasReconciling: false };
+  }
+
+  const hasReconciling = resources.some(isResourceReconciling);
```
AdityaHegde, line 33: There is a case where during initial creation of a `Deployment` parser has not run yet. So there could be no resources returned here.

So would be good to do a fixed refetch if there are no resources at all. An empty project will have the project parser resource so there wont be unnecessary refetches there.

lovincyrus: https://github.com/rilldata/rill/pull/7216/commits/673bd4638fc70c30d132a31c7fd83a7dea29486f

web-admin/src/lib/refetch-interval-store.ts
lovincyrus: let's keep it as `createSmartRefetchInterval` for now

web-admin/tests/setup/setup.ts
lovincyrus: I suppose, we can remove all the manual page reloads. Can we explore this as a follow-up

AdityaHegde: Ya that should be fine.

ericokuma:
@lovincyrus, is this ready to be merged?

PR #7222, opened 2025-04-28
Use rill managed git repositories instead of zip-ship

k-anshul:
closes https://github.com/rilldata/rill-private-issues/issues/1569

1. Uses rill managed git repositories instead of zip-ship
2. Existing projects using zip-ship will automatically migrate to managed git after first push.
3. Added a `archive` flag to fallback to older zip-ship flow for testing purposes. 

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!


admin/server/github.go
```diff
@@ -413,12 +423,96 @@ func (s *Server) CreateManagedGitRepo(ctx context.Context, req *adminv1.CreateMa
...
+		if err != nil {
+			return err
+		}
+
+		return copyFromSrcGit(path, *proj.GithubURL, proj.ProdBranch, proj.Subpath, token)
+	}
+	err = s.pushToGit(ctx, copyData, *repo.CloneURL, *repo.DefaultBranch, "", mgdRepoToken, true)
+	if err != nil {
+		return nil, err
+	}
```
begelundmuller, line 497: I'm wondering if this logic could be decoupled/simplified some more. Perhaps moved to a Git utils package.

For example, could we separate the clone and re-upload logic into separate utils? Like have `gitutils.ShallowClone` and `gitutils.ForcePush`, which we connect together here, or something like that?

admin/server/github.go
```diff
@@ -413,12 +423,96 @@ func (s *Server) CreateManagedGitRepo(ctx context.Context, req *adminv1.CreateMa
...
+	subpath := ""
+	_, err = s.UpdateProject(ctx, &adminv1.UpdateProjectRequest{
+		OrganizationName: req.Organization,
+		Name:             req.Project,
+		GithubUrl:        repo.CloneURL,
+		ProdBranch:       &branch,
+		Subpath:          &subpath,
+		ArchiveAssetId:   nil,
+	})
+	if err != nil {
```
begelundmuller, line 509: It's not too good to call RPCs with direct function calls. It may work here, but just has some code smell. (I notice it comes from the old `UploadProjectAssets` RPC, but would be good to take this chance to fix it.)

If shared logic is needed by multiple RPCs, then ideally it should be refactored into a common function that both RPCs call. Usually such "service layer" functions go on `s.admin`, but could also be an internal function on `s` direectly.

admin/server/github.go
```diff
@@ -1069,20 +1163,37 @@ func (s *Server) pushToGit(ctx context.Context, copyData func(projPath string) e
...
+		// fallback to username and email in Rill
+		claims := auth.GetClaims(ctx)
+		if claims.OwnerType() != auth.OwnerTypeUser {
+			// ideally should never reach here
+			return fmt.Errorf("failed to get current user: %w", err)
+		}
+		u, err := s.admin.DB.FindUser(ctx, claims.OwnerID())
+		if err != nil {
+			return err
+		}
```
begelundmuller, line 1179: Consider doing this in the RPC implementation and passing the desired name/email as parameters to this function. That way `pushToGit` is not coupled to server middleware, and can be considered a standalone util function, which may in the future e.g. be tested stand-alone or invoked from a background job instead.

admin/server/projects.go
```diff
@@ -554,21 +554,11 @@ func (s *Server) CreateProject(ctx context.Context, req *adminv1.CreateProjectRe
...
 		return nil, status.Error(codes.InvalidArgument, "cannot set both github_url and archive_asset_id")
 	} else if req.GithubUrl != "" {
-		// Github projects must be configured by a user so we can ensure that they're allowed to access the repo.
-		if userID == nil {
-			return nil, status.Error(codes.Unauthenticated, "not authenticated as a user")
-		}
-
-		// Check Github app is installed and caller has access on the repo
-		repoID, installationID, err := s.getAndCheckGithubInstallationID(ctx, req.GithubUrl, *userID)
+		opts.GithubRepoID, opts.GithubInstallationID, opts.ManagedGitRepoID, opts.ProdBranch, err = s.githubOptsForGithubURL(ctx, org.ID, req.ProdBranch, userID, req.GithubUrl)
```
begelundmuller, line 557: Not sure, but you could consider having the util function write to the `opts` in-place with something like `s.applyGithubOpts(ctx, opts, ...)` to avoid too many return values.

cli/cmd/deploy/deploy.go
```diff
@@ -66,6 +70,11 @@ func DeployCmd(ch *cmdutil.Helper) *cobra.Command {
 	}
 
 	deployCmd.Flags().BoolVar(&upload, "upload", false, "Create project using rill managed repo")
+	deployCmd.Flags().BoolVar(&zipship, "zipship", false, "Create project using the zip and ship flow(for testing only)")
+	err := deployCmd.Flags().MarkHidden("zipship")
+	if err != nil {
+		panic(err)
+	}
 	deployCmd.Flags().BoolVar(&github, "github", false, "Use github repo to create the project")
```
begelundmuller, line 78: `upload` sounds like a tarball upload as opposed to a managed Git, I'm wondering if we should not use that for managed Git. What do you think about:
- `--github` for user-managed Github
- `--managed` or `--managed-git` for Rill-managed Git
- `--archive` for tarball uploads (I prefer `archive` over `zipship` because that's what we call it internally)

begelundmuller (changes requested):

admin/github.go
```diff
@@ -301,8 +301,8 @@ func (s *Service) ProcessGithubEvent(ctx context.Context, rawEvent any) error {
 func (s *Service) processGithubPush(ctx context.Context, event *github.PushEvent) error {
 	// Find Rill project matching the repo that was pushed to
 	repo := event.GetRepo()
-	githubURL := *repo.HTMLURL
-	projects, err := s.DB.FindProjectsByGithubURL(ctx, githubURL)
+	// We stored HTML URL earlier, but now we are using clone URL
+	projects, err := s.DB.FindProjectsByGithubURL(ctx, []string{*repo.HTMLURL, *repo.CloneURL})
```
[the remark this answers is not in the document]
k-anshul: Ohh I thought it will require building script. I can simply add a column and that will a better solution. I can make that change.

admin/server/github.go
```diff
@@ -413,12 +423,96 @@ func (s *Server) CreateManagedGitRepo(ctx context.Context, req *adminv1.CreateMa
...
+	observability.AddRequestAttributes(ctx,
+		attribute.String("args.organization", req.Organization),
+		attribute.String("args.project", req.Project),
+	)
+
+	// Check the request is made by a user
+	claims := auth.GetClaims(ctx)
+	if claims.OwnerType() != auth.OwnerTypeUser {
+		return nil, status.Error(codes.Unauthenticated, "not authenticated as a user")
+	}
```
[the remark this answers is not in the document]
k-anshul: Yes. To get the git commit signature(`name` and `email`) we need this operation to be done by a user.

admin/server/projects.go
k-anshul: This is used in `create` and `update` both which have different opts so can't modify opts in place. 

Few other options: 
1. I can return a struct if that makes it more clearer. I did not add it earlier since it is unlikely that this will be used in more places.
2. I can take in both `opts` and set to whatever is not nil.

cli/cmd/deploy/deploy.go
k-anshul: Yeah I agree `upload` sounds somewhat `flawed`. 
The suggestions sounds good to me but do you think we should still keep `--upload`, mark it as deprecated and reroute it to `--managed-git` for backward compatibility.

admin/server/github.go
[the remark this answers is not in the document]
k-anshul: This is anyways just invoked from UI but if we want to let a service account make these changes then we can set service name and constant email. Since this commit is present in rill managed git repo so it should be fine to not set actual user's email.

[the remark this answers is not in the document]
begelundmuller: I see. If we could fall back to `noreply@rilldata.com` or something like that, that'd be nice. But otherwise it's also fine to require user auth.

admin/server/projects.go
begelundmuller: I see, probably it's simplest just to keep what you have now then.

cli/cmd/deploy/deploy.go
begelundmuller: Up to you if you prefer `--upload` or `--archive`, I'm good with both

admin/server/projects.go
k-anshul: Keeping it same for now.

cli/cmd/deploy/deploy.go
```diff
@@ -10,7 +10,7 @@ import (
 
 // DeployCmd is the guided tour for deploying rill projects to rill cloud.
 func DeployCmd(ch *cmdutil.Helper) *cobra.Command {
-	var upload, github bool
+	var upload, github, zipship bool
```
begelundmuller, line 13: Have the variables reflect the non-deprecated flag names:

```suggestion
	var managed, github, archive bool
```

cli/cmd/deploy/deploy.go
```diff
@@ -65,7 +69,25 @@ func DeployCmd(ch *cmdutil.Helper) *cobra.Command {
...
+	// --managed and --upload are mutually exclusive and do the same thing i.e. create a project using rill managed git repo
+	// --archive is hidden and create a project using tarballs. It is for testing only.
+	// --github creates a project using user managed github repo
+	deployCmd.Flags().BoolVar(&upload, "managed", false, "Create project using rill managed repo")
 	deployCmd.Flags().BoolVar(&upload, "upload", false, "Create project using rill managed repo")
+	err := deployCmd.Flags().MarkDeprecated("upload", "Use --managed instead")
+	if err != nil {
+		panic(err)
+	}
+	deployCmd.MarkFlagsMutuallyExclusive("managed", "upload")
```
begelundmuller, line 83: I think we should just remove the `upload` flag. It's more okay to change CLI behavior than APIs because compatibility is not a problem.

One worry I had was for folks deploying from Gitlab, but I checked and our docs for Gitlab deploys don't use the `--upload` flag, so I don't think there's anything to be worried about here.

begelundmuller (approved):

PR #7223, opened 2025-04-28
feat: improve canvas chart legend

djbarnwal:
Add multiple improvements to chart legends - 
1. Add selector for choosing legend orientation
2. Add symbol limit when legend is placed vertically
3. Add column limit when legend is placed horizontally
4. Center align all legend types for circular charts
5. Add a config wrapper util method.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


briangregoryholmes (approved):

djbarnwal:
> Is there a limitation in place on the width of the legend? Seems like this should be two or even three columns wide.

Yeah, this is more of a stopgap solution. The current number of columns is calculated using an estimation which works for most cases but breaks when the legend labels are too small or too large. Vega doesn't provide a way for us to read the label lengths. We might have a write a workaround for it. For now I am thinking of exposing the `column` configuration to the end user as well. The config will be set to `auto` and then the user can tweak it. 

I will merge this PR and create a follow up to address the above issue. I think we should fix it before the next release.

PR #7235, opened 2025-04-29
Motherduck Integration test

NamanMahor:
Closes [#1642](https://github.com/rilldata/rill-private-issues/issues/1642)

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!


.github/workflows/go-test.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     - name: Go test
       run: go test -short -v ./...
       env:
-        RILL_RUNTIME_RESOLVERS_TEST_RESTRICTED_CONNECTORS: "snowflake"
+        RILL_RUNTIME_RESOLVERS_TEST_RESTRICTED_CONNECTORS: "snowflake,motherduck"
```
k-anshul, line 24: Why is `motherduck` restricted ?

k-anshul (approved):

.github/workflows/go-test.yml
NamanMahor: it free upload 10 hour cpu. but I think we will not hit the limit even if we run with ever pr. I think I should make it unrestricted.

Issue #5964, opened 2024-10-23
Security policies: when no rows are returned, the UI hangs


ericpgreen2:
- When a security policy filters out all rows, the UI hangs
- The call to `MetricsViewTimeRange` succeeds, but no follow up requests for metrics are made
- The `MetricsViewTimeRange` response probably contains null start/end times because the security policy didn’t match any rows.
- The null start/end times likely cause the UI to hang, but we should instead show a message that the dashboard has no rows

ericpgreen2:
Re-opening this issue as we're currently observing problems with empty models, which lead the `TimeRangeSummary` API to return `null`.

PR #7236, opened 2025-04-30
fix: Add back the no-data check for dashboards

AdityaHegde:
closes https://github.com/rilldata/rill/issues/5964

The no-data check from here https://github.com/rilldata/rill/pull/6079/files was removed when we moved to human readable url state.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


web-common/src/features/dashboards/selectors.ts
```diff
@@ -122,6 +126,79 @@ export function useMetricsViewTimeRange(
   );
 }
 
+/**
+ * Wrapper function that fetches full time range.
+ * Uses useExploreValidSpec unlike the above useMetricsViewTimeRange since it is more widely used.
+ *
+ * Does an additional validation where null min and max returned throws an error instead.
+ *
+ * TODO: replace usages of useMetricsViewTimeRange with this eventually
```
ericpgreen2, line 135: I'm not sure– do we want to use this everywhere? The Luxon settings should only be set once– in the entrypoint to the dashboard. And as long as we throw the `null` error at the top-level, then we're good.

web-common/src/features/dashboards/selectors.ts
```diff
@@ -122,6 +126,79 @@ export function useMetricsViewTimeRange(
...
+  options?: {
+    query?: CreateQueryOptions<V1MetricsViewTimeRangeResponse>;
+  },
+  queryClient?: QueryClient,
+): CompoundQueryResult<V1MetricsViewTimeRangeResponse> {
+  const { query: queryOptions } = options ?? {};
+
+  return derived(
+    useExploreValidSpec(instanceId, exploreName, undefined, queryClient),
+    (validSpecResp, set) => {
```
ericpgreen2, line 149: For dependent queries, can we move away from this pattern, which leads to unstable query observers, and toward this one made possible by TanStack Query v5:
```typescript
const query1OptionsStore = derived(...)
const query1 = createQuery(query1OptionsStore)

const query2OptionsStore = derived([..., query1], (..., query1) => {
  const query2Options = {
    ...
    enabled: !!query1.data
  }

  return query2Options;
})
const query2 = createQuery(query2OptionsStore)
```
As an example, Dhiraj did this recently [here](https://github.com/rilldata/rill/blob/fix/empty-dashboard-source/web-common/src/features/canvas/components/charts/cartesian-charts/CartesianChart.ts#L74-L198).

It not only gives us stable query observers (so we can use `keepPreviousData`), but it also reduces nesting, which makes the code easier to read.

ericpgreen2 (changes requested):

web-common/src/features/dashboards/selectors.ts
AdityaHegde: Hmmm true. I can remove the TODO

PR #7239, opened 2025-04-30
Support "Change Modes" on models enabling for controlled model refreshes

grahamplata:
Currently a small SQL change in a model automatically triggers a full refresh. This PR supports specifying a "change mode" on models, which governs behavior when the model logic changes.

Resolves: https://github.com/rilldata/rill-private-issues/issues/1635

**It supports these options**:
`reset`: changing the model automatically leads to a full refresh (default, current behavior)
`manual`: changing the model stops refreshes until a manual incremental or full refresh is run
`patch`: changing the model automatically changes to the new logic without a reset

```yaml
type: model
change_mode: {reset|manual|patch}
```

**Checklist:**
- [x] Covered by tests
- [ ] I'm proud of this work!


runtime/reconcilers/model.go
```diff
@@ -1521,3 +1515,31 @@ func md5Hash(val []byte) (string, error) {
...
+		!exists {
+		return true
+	}
+
+	// Handle different change modes
+	switch model.Spec.ChangeMode {
+	case "manual":
+		// In manual mode, spec hash changes don't trigger automatic resets
+		// Only explicit trigger or missing output will cause a reset
+		return false
```
begelundmuller, line 1534: 1. It should return an error informing the user that the model is stopped and that they should manually trigger it. Something like `execution paused because the model definition was changed and change_mode=manual: you must manually trigger either a full or incremental refresh` (feel free to suggest something better).
2. When the user next sends a manual trigger (i.e. sets `Spec.Trigger` or `Spec.TriggerFull`), it should stop blocking reconciliation.

runtime/reconcilers/model.go
```diff
@@ -1521,3 +1515,31 @@ func md5Hash(val []byte) (string, error) {
...
+	case "patch":
+		// In patch mode, silently update the hash but don't trigger a reset
+		if model.State.SpecHash != specHash {
+			model.State.SpecHash = specHash
+		}
+		return false
```
begelundmuller, line 1540: 1. It should error if the `patch` change mode is used with a non-incremental model.
2. Let's emit an info log that the definition was changed and that we're changing to it without a full refresh. 
3. It needs to call `r.C.UpdateState` for changes to `model.State` to take effect

begelundmuller (changes requested):

grahamplata:
Screenshot 2025-05-09 at 10 49 43 AM

Example warn

begelundmuller (changes requested):

begelundmuller (approved):

PR #7198, opened 2025-04-22
Share project popover improvements

lovincyrus:
[INSERT DESCRIPTION HERE](https://www.notion.so/rilldata/User-Management-Role-Based-Access-Control-RBAC-Enhancements-8d331b29d9b64d87bca066e06ef87f54?pvs=4#1a6ba33c8f5780349c22fde23ddeee60)

**Checklist:**
- [ ] Covered by tests
- [ ] Ran it and it works as intended
- [ ] Reviewed the diff before requesting a review
- [ ] Checked for unhandled edge cases
- [ ] Linked the issues it closes
- [ ] Checked if the docs need to be updated
- [ ] Intend to cherry-pick into the release branch
- [ ] I'm proud of this work!

PR #7242, opened 2025-04-30
Fix org permissions for org tabs

lovincyrus:
This pull request ensures the correctness of org roles' permissions to the organization tabs. Closes https://github.com/rilldata/rill-private-issues/issues/1651

- `/users` can be accessed by org admin + editor
- `/settings` can be accessed by org admin

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-admin/src/features/organizations/users/permissions.ts
+++ b/web-admin/src/features/organizations/users/permissions.ts
@@ -1,12 +1,41 @@
-import type { V1OrganizationPermissions } from "@rilldata/web-admin/client";
+import type {
+  V1OrganizationPermissions,
+  V1OrganizationRole,
+} from "@rilldata/web-admin/client";
+
+/**
+ * Gets the organization role from the list of roles based on the permissions
+ */
+export function getOrgRole(
+  permissions?: V1OrganizationPermissions,
+  roles?: V1OrganizationRole[],
+): V1OrganizationRole | undefined {
+  if (!permissions || !roles) return undefined;
+
+  // Find the role that matches the user's permissions
+  return roles.find((role) => {
+    const rolePerms = role.permissions;
+    if (!rolePerms) return false;
+
+    // Check if all permissions in the role match the user's permissions
+    return Object.entries(rolePerms).every(
+      ([key, value]) =>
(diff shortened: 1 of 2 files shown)


web-admin/src/features/organizations/users/permissions.ts
```diff
@@ -0,0 +1,43 @@
+import type { V1OrganizationPermissions } from "@rilldata/web-admin/client";
+
+/**
+ * @source https://docs.rilldata.com/manage/roles-permissions#organization-level-permissions
+ *
+ * Checks if a user has admin permissions for an organization
+ */
+export function isOrgAdmin(permissions?: V1OrganizationPermissions): boolean {
+  return !!permissions?.admin;
```
lovincyrus, line 9: Just curious - any reason why we have a separate admin and guest prop but not for editor and viewer? @begelundmuller

begelundmuller: Yeah, I understand it seems a bit weird, but the idea is that `admin` and `guest` in this case are flags with a special internal meaning.

You can imagine in the future, we might introduce more roles, like "owner" and "billing admin", which would also have the `admin=true` despite having a different names.

lovincyrus: Any chance that we can have roles `editor` and `viewer` in the props? This is so we can have a reliable role source and don't need to maintain a separate org role with the granular permissions for the UI.

nishantmonu51: maybe expose role as a separate field that the UI can use and not rely on these flags ?

ericpgreen2: > Any chance that we can have roles `editor` and `viewer` in the props? 

@lovincyrus, I think the mapping that you're looking for comes from the `adminServiceListRoles` API. It looks like it's the backend's mapping from role->permissions.

lovincyrus:
> https://github.com/rilldata/rill/pull/7198#discussion_r2079720876

Continue using permissions, not roles

PR #1652, opened 2023-01-25
Add UX support for globs

djbarnwal:
#1618

- [x] Sanitize the * in the source name inferred from path
- [x]  The info tool tip needs to say we DO support glob patterns
- [x]  Add UI support for local file globs (setting a path instead of uploading a file)

cc @magorlick 
We might need to think more on how to handle UX for Local files glob support here. The current implementation adds a secondary button to show the detailed options.

PR #7251, opened 2025-05-01
Integration test for redshift

NamanMahor:
Closed [#1652](https://github.com/rilldata/rill-private-issues/issues/1652)

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!


runtime/drivers/redshift/warehouse.go
```diff
@@ -240,8 +240,8 @@ type sourceProperties struct {
 	OutputLocation    string `mapstructure:"output_location"`
 	Workgroup         string `mapstructure:"workgroup"`
 	Database          string `mapstructure:"database"`
-	ClusterIdentifier string `mapstructure:"cluster.identifier"`
-	RoleARN           string `mapstructure:"role.arn"`
+	ClusterIdentifier string `mapstructure:"cluster_identifier"`
```
k-anshul, line 243: This will make it backward incompatible. Why is this change necessary ?

k-anshul (changes requested):

runtime/drivers/redshift/warehouse.go
NamanMahor: it is mentioned as `cluster_identifier` and `role_arn`  in SourceProperties which UI uses to generate it yaml. 
So I think user were not able to redshift as source unless we told them to use role.arn

k-anshul (approved):

PR #7252, opened 2025-05-01
Fix: Canvas Metrics View Time Consistency flaky test

grahamplata:
Fix: Canvas Metrics View Time Consistency flaky test

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/runtime/reconcilers/canvas_test.go
+++ b/runtime/reconcilers/canvas_test.go
@@ -2,6 +2,7 @@ package reconcilers_test
 
 import (
 	"testing"
+	"time"
 
 	"github.com/rilldata/rill/runtime"
 	"github.com/rilldata/rill/runtime/testruntime"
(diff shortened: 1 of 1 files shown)


runtime/reconcilers/canvas_test.go
```diff
@@ -192,6 +193,13 @@ first_month_of_year: 1
 	// Reconcile and verify that the metrics view gets updated
 	testruntime.ReconcileParserAndWait(t, rt, id)
 
+	// Add polling logic to wait for expected state
+	require.Eventually(t, func() bool {
+		mv2 = testruntime.GetResource(t, rt, id, runtime.ResourceKindMetricsView, "mv2")
+		return mv2.GetMetricsView().State.ValidSpec != nil &&
+			mv2.GetMetricsView().State.ValidSpec.FirstDayOfWeek == 1
+	}, 5*time.Second, 100*time.Millisecond, "metrics view spec not updated in time")
```
begelundmuller, line 201: Not sure, but I don't think you need to poll. The `ReconcileParserAndWait` call will wait until nothing is reconciling.

I think the problem is just that the order iteration is non-deterministic, so the check on line `212` sometimes have the names the other way around.

begelundmuller (changes requested):

runtime/reconcilers/canvas_test.go
grahamplata: Ahh yep

begelundmuller (approved):

PR #7264, opened 2025-05-02
feat : Adding a clone command

k-anshul:
Subtask for https://github.com/rilldata/rill-private-issues/issues/1569

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/cli/cmd/project/clone.go
+++ b/cli/cmd/project/clone.go
@@ -1,0 +1,62 @@
+package project
+
+import (
+	"fmt"
+
+	"github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/plumbing"
+	githttp "github.com/go-git/go-git/v5/plumbing/transport/http"
+	"github.com/rilldata/rill/cli/pkg/cmdutil"
+	adminv1 "github.com/rilldata/rill/proto/gen/rill/admin/v1"
+	"github.com/spf13/cobra"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+func CloneCmd(ch *cmdutil.Helper) *cobra.Command {
+	var path string
+
+	showCmd := &cobra.Command{
+		Use:   "clone [<project-name>]",
+		Args:  cobra.ExactArgs(1),
+		Short: "Clone Project",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			client, err := ch.Client()
+			if err != nil {
+				return err
+			}
+			name := args[0]
+
+			proj, err := client.GetCloneCredentials(cmd.Context(), &adminv1.GetCloneCredentialsRequest{
+				Organization: ch.Org,
+				Project:      name,
+			})
+			if err != nil {
+				st, ok := status.FromError(err)
+				if !ok {
+					return err
+				}
+				if st.Code() == codes.InvalidArgument {
+					return fmt.Errorf("project %q not found: %w", name, err)
+				}
+				if st.Code() == codes.PermissionDenied {
+					return fmt.Errorf("you do not have permission to clone project %q", name, err)
+				}
+				return err
+			}
+
+			_, err = git.PlainClone(path, false, &git.CloneOptions{
+				URL:           proj.GitRepoUrl,
+				Auth:          &githttp.BasicAuth{Username: proj.GitUsername, Password: proj.GitPassword},
+				ReferenceName: plumbing.NewBranchReferenceName(proj.GitProdBranch),
+			})
+
+			return err
+		},
+	}
+
+	showCmd.Flags().SortFlags = false
+	showCmd.Flags().StringVar(&path, "path", ".", "Project path to clone to")
+
+	return showCmd
+}
--- a/cli/cmd/project/project.go
+++ b/cli/cmd/project/project.go
@@ -30,6 +30,7 @@ func ProjectCmd(ch *cmdutil.Helper) *cobra.Command {
 	projectCmd.AddCommand(RefreshCmd(ch))
 	projectCmd.AddCommand(JwtCmd(ch))
 	projectCmd.AddCommand(GitPushCmd(ch))
+	projectCmd.AddCommand(CloneCmd(ch))
 	projectCmd.AddCommand(DeployCmd(ch))
 
 	return projectCmd


cli/cmd/project/clone.go
```diff
@@ -0,0 +1,153 @@
...
+			}
+
+			// get project
+			res, err := client.GetProject(cmd.Context(), &adminv1.GetProjectRequest{OrganizationName: ch.Org, Name: name})
+			if err != nil {
+				return err
+			}
+
+			if res.Project.ArchiveAssetId != "" {
+				return fmt.Errorf("project is not connected to a git repository, please redeploy the project to connect it to a git repository")
```
k-anshul, line 42: If required we can auto migrate the project to managed git repos as well.

cli/cmd/project/clone.go
```diff
@@ -0,0 +1,153 @@
...
+	adminv1 "github.com/rilldata/rill/proto/gen/rill/admin/v1"
+	"github.com/spf13/cobra"
+)
+
+func CloneCmd(ch *cmdutil.Helper) *cobra.Command {
+	var path string
+
+	cloneCmd := &cobra.Command{
+		Use:   "clone [<project-name>]",
+		Args:  cobra.ExactArgs(1),
```
begelundmuller, line 22: Square parentheses means optional:

```suggestion
		Use:   "clone <project-name>",
		Args:  cobra.ExactArgs(1),
```

cli/cmd/project/clone.go
```diff
@@ -0,0 +1,153 @@
...
+				return fmt.Errorf("project is not connected to a git repository, please redeploy the project to connect it to a git repository")
+			}
+
+			// check if dir is empty
+			empty, err := isDirAbsentOrEmpty(path)
+			if err != nil {
+				return err
+			}
+			if !empty {
+				ok, err := cmdutil.ConfirmPrompt(fmt.Sprintf("Directory %q is not empty. Do you want to overwrite it?", path), "", false)
```
begelundmuller, line 51: We don't know that it is a directory, it could also be a file:
```suggestion
				ok, err := cmdutil.ConfirmPrompt(fmt.Sprintf("There are files at path %q. Do you want to overwrite?", path), "", false)
```

cli/cmd/project/clone.go
```diff
@@ -0,0 +1,153 @@
...
+				return err
+			}
+
+			// clone repository
+			_, err = git.PlainCloneContext(cmd.Context(), path, false, &git.CloneOptions{
+				URL:           config.Remote,
+				Auth:          &githttp.BasicAuth{Username: config.Username, Password: config.Password},
+				ReferenceName: plumbing.NewBranchReferenceName(config.DefaultBranch),
+				SingleBranch:  true,
+			})
```
begelundmuller, line 78: Why not have cloning as a function on the Git helper? Might be nice to keep Git logic in a single place.

cli/cmd/project/clone.go
```diff
@@ -0,0 +1,153 @@
...
+
+			repo, _, err := cmdutil.RepoForProjectPath(path)
+			if err != nil {
+				return err
+			}
+
+			_, err = cmdutil.EnsureGitignoreHasDotRillCloud(cmd.Context(), repo)
+			if err != nil {
+				return fmt.Errorf("failed to update .gitignore: %w", err)
+			}
```
begelundmuller, line 107: Should we put `EnsureGitignoreHasDotRillCloud` inside `dotrillcloud.SetAll`?

cli/cmd/project/clone.go
```diff
@@ -0,0 +1,153 @@
...
+
+	// Check if the directory is empty
+	return len(entries) == 0, nil
+}
+
+func recreateDir(path string) error {
+	// Remove directory and its contents if exists
+	err := os.RemoveAll(path)
+	if err != nil {
+		// NOTE :: does not return an error if the directory does not exist
```
begelundmuller, line 142: nit: it's English, not C++ 😄 
```suggestion
		// NOTE: does not return an error if the directory does not exist
```

cli/cmd/project/clone.go
begelundmuller: I would not do that automatically, some people may want to use archives, like deploys from Gitlab or internal/test projects

cli/cmd/project/project.go
```diff
@@ -31,6 +31,7 @@ func ProjectCmd(ch *cmdutil.Helper) *cobra.Command {
 	projectCmd.AddCommand(JwtCmd(ch))
 	projectCmd.AddCommand(GitPushCmd(ch))
 	projectCmd.AddCommand(DeployCmd(ch))
+	projectCmd.AddCommand(CloneCmd(ch))
```
begelundmuller, line 34: Perhaps move it up above `JwtCmd`?

cli/pkg/cmdutil/githelper.go
```diff
@@ -0,0 +1,200 @@
...
+
+var (
+	gitignoreHasDotenvRegexp       = regexp.MustCompile(`(?m)^\.env$`)
+	gitignoreHasDotRillCloudRegexp = regexp.MustCompile(`(?m)^\s*\.rillcloud/`)
+)
+
+// GitHelper manages git operations for a project
+// It also caches the git credentials for the project
+type GitHelper struct {
+	c         *client.Client
```
begelundmuller, line 29: I think it should hold a reference to `cmdutil.Helper` instead here, and then use it to get the client when it needs it.

It isn't safe to keep the admin client cached/open outside of `cmdutil.Helper`, see the logic here: https://github.com/rilldata/rill/blob/7e9ec7c223146c0f13211de822467183d7bb4ef2/cli/pkg/cmdutil/helper.go#L142

cli/pkg/cmdutil/githelper.go
```diff
@@ -0,0 +1,200 @@
...
+	if err != nil {
+		return err
+	}
+	defer g.gitConfigMu.Release(1)
+
+	g.gitConfig = c
+	return nil
+}
+
+func AutoCommitGitSignature(ctx context.Context, c adminv1.AdminServiceClient, path string) (*object.Signature, error) {
```
begelundmuller, line 140: Instead of passing an `AdminServiceClient`, it should probably be a member function of `Helper` or `GitHelper` and get the client lazily.

cli/pkg/cmdutil/githelper.go
```diff
@@ -0,0 +1,200 @@
...
+	return &GitHelper{
+		c:           adminClient,
+		org:         org,
+		project:     project,
+		localPath:   localPath,
+		gitConfigMu: semaphore.NewWeighted(1),
+	}
+}
+
+func (g *GitHelper) FetchGitConfig(ctx context.Context) (*gitutil.Config, error) {
```
begelundmuller, line 49: nit: I suggest avoiding the word "fetch" here because it has a specific meaning in Git.
```suggestion
func (g *GitHelper) GitConfig(ctx context.Context) (*gitutil.Config, error) {
```

cli/pkg/cmdutil/githelper.go
```diff
@@ -0,0 +1,200 @@
...
+	org       string
+	project   string
+	localPath string
+
+	// do not access gitConfig directly, use FetchGitConfig and setGitConfig
+	gitConfig   *gitutil.Config
+	gitConfigMu *semaphore.Weighted
+}
+
+func NewGitHelper(adminClient *client.Client, org, project, localPath string) *GitHelper {
```
begelundmuller, line 39: Consider refactoring this to `(h *Helper) GitHelper(project, localPath string)` (org is already configured on the `Helper` and the client should be retrieved lazily from the helper).

cli/pkg/cmdutil/githelper.go
```diff
@@ -0,0 +1,200 @@
...
+	"github.com/rilldata/rill/admin/client"
+	"github.com/rilldata/rill/cli/pkg/gitutil"
+	adminv1 "github.com/rilldata/rill/proto/gen/rill/admin/v1"
+	"github.com/rilldata/rill/runtime/drivers"
+	"golang.org/x/sync/semaphore"
+)
+
+var (
+	gitignoreHasDotenvRegexp       = regexp.MustCompile(`(?m)^\.env$`)
+	gitignoreHasDotRillCloudRegexp = regexp.MustCompile(`(?m)^\s*\.rillcloud/`)
```
begelundmuller, line 23: I don't think `.gitignore` requries a trailing slash? I.e. `.rillcloud` is also valid, doesn't have to be `.rillcloud/`

cli/pkg/cmdutil/helper.go
```diff
@@ -439,6 +443,13 @@ func (h *Helper) OpenRuntimeClient(ctx context.Context, org, project string, loc
 	return rt, instanceID, nil
 }
 
+func (h *Helper) GitHelper(client *client.Client, org, project, localPath string) *GitHelper {
```
begelundmuller, line 446: Ah I see you already added this – do you need the `NewGitHelper` to be exported then? It'd be great if this was the only entrypoint for getting a Git helper.

cli/pkg/cmdutil/helper.go
```diff
@@ -439,6 +443,13 @@ func (h *Helper) OpenRuntimeClient(ctx context.Context, org, project string, loc
...
+	h.gitHelperOnce.Do(func() {
+		h.gitHelper = NewGitHelper(client, org, project, localPath)
```
begelundmuller, line 448: This is inconsistent – if you invoke it for another org/project/localPath, then it will return a stale helper.

Even if such invocations don't happen today, it's confusing and could conceivably lead to bugs in the future as the local UI builds out project-switching capabilities.

cli/cmd/project/clone.go
k-anshul: Yes we can do that though I would keep it here as well since older projects may not have it.

k-anshul: Actually I now realise that this is a mistake.

IIRC it was a requirement to commit `.rillcloud` to github.

cli/pkg/cmdutil/githelper.go
k-anshul: In certain code flows the project name is not known. I can pass `Helper` to create the client lazily.

cli/cmd/project/clone.go
begelundmuller: Yeah I was a little stumped by this, but I don't clearly remember the requirements here.

cli/pkg/cmdutil/helper.go
begelundmuller: Is a check sufficient though? What about if someone deploys fron the local UI? Then we will go from not knowing the project name to suddenly knowing it?

It's fine to cache the `GitHelper`, but maybe it needs to be a map or to do do some invalidation checks similar to `h.Client()`?

cli/cmd/project/clone.go
k-anshul: Yeah reverted to the original behaviour.

cli/pkg/cmdutil/helper.go
k-anshul: > Then we will go from not knowing the project name to suddenly knowing it?

Which is fine I guess we will just recreate the `GitHelper`.

```
func (h *Helper) GitHelper(project, localPath string) *GitHelper {
	h.gitHelperMu.Lock()
	defer h.gitHelperMu.Unlock()

	// If the git helper is nil or the org, project, or local path has changed, create a new one.
	if h.gitHelper == nil || h.gitHelper.project != project || h.gitHelper.localPath != localPath {
		h.gitHelper = newGitHelper(h, h.Org, project, localPath)
	}
	return h.gitHelper
}
```

> but maybe it needs to be a map or to do do some invalidation checks similar to h.Client()

Map is unnecessary because it should be fine to recreate it when project changes. Happen only once in running server right now.

Checking for project and localPath is the invalidation check similar to `h.Client()`. There the state depends on admin token and admin URL and here the state depends on project and org and path.

begelundmuller: Yeah that seems fine – just as long as it responds to project name changes

cli/pkg/cmdutil/githelper.go
```diff
@@ -0,0 +1,207 @@
...
+	if err != nil {
+		return err
+	}
+	defer g.gitConfigMu.Release(1)
+
+	g.gitConfig = c
+	return nil
+}
+
+func AutoCommitGitSignature(ctx context.Context, h *Helper, path string) (*object.Signature, error) {
```
begelundmuller, line 147: Maybe `h.GitSignature(ctx, path)`? The `*Helper` is usually accessed, not passed.

begelundmuller (approved):

PR #1640, opened 2023-01-21
Create `features/entity-management` dir

ericpgreen2:
This PR continues the refactor to `web-common/features/` directories by consolidating the entity management code that relates to _all_ entities (sources, models, and dashboards) into `web-common/src/features/entity-management`. This new directory does not contain code that applies to _individual_ entities; that code should be placed into `features/sources`, `features/models`, etc.

Notably, entity management has much to do with our Rill Code artifacts. You'll see various utils and Svelte Query convenience functions that apply to code artifacts generally.

PR #7275, opened 2025-05-06
Generate doc from json schema

NamanMahor:
Closes [#1640](https://github.com/rilldata/rill-private-issues/issues/1640)

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!


Makefile
```diff
@@ -40,9 +40,10 @@ coverage.go:
 docs.generate:
 	# Temporarily replaces ~/.rill/config.yaml to avoid including user-defined defaults in generated docs.
 	# Sets version to the latest tag to simulate a production build, where certain commands are hidden.
-	rm -rf docs/docs/reference/cli
+	rm -rf docs/docs/reference/cli docs/docs/reference/project-files
```
begelundmuller, line 43: We don't want to delete the old manually maintained files just yet. Can we output to a separate directory in this PR? For example, could be `docs/docs/reference/yaml` or something like that.

And then when everyone's happy with the new docs, we can do a separate PR that does the full switch.

Makefile
```diff
@@ -40,9 +40,10 @@ coverage.go:
...
 	if [ -f ~/.rill/config.yaml ]; then mv ~/.rill/config.yaml ~/.rill/config.yaml.tmp; fi;
-	go run -ldflags="-X main.Version=$(shell git describe --tags `git rev-list --tags --max-count=1`)" ./cli docs generate docs/docs/reference/cli/
+	go run -ldflags="-X main.Version=$(shell git describe --tags `git rev-list --tags --max-count=1`)" ./cli docs generate-cli docs/docs/reference/cli/
+	go run -ldflags="-X main.Version=$(shell git describe --tags `git rev-list --tags --max-count=1`)" ./cli docs generate-project --project runtime/parser/schema/project.schema.json --rillyaml runtime/parser/schema/rillyaml.schema.json  --out docs/docs/reference/project-files
```
begelundmuller, line 46:
```
--project runtime/parser/schema/project.schema.json --rillyaml runtime/parser/schema/rillyaml.schema.json
```
Can you hard-code the schema paths paths in the command instead of passing here? They are unlikely to change, so better keep the config in the code

docs/docs/reference/cli/docs/docs.md
```diff
@@ -22,4 +22,5 @@ rill docs [flags]
 ### SEE ALSO
 
 * [rill](../cli.md)	 - A CLI for Rill
+* [rill docs generate-project](generate-project.md)	 - Generate Markdown docs from JSON Schemas for Project files
```
begelundmuller, line 25: This command is internal so should not be in the CLI usage docs. Maybe the command definition is missing `Hidden: !ch.IsDev()`?

begelundmuller (changes requested):
Can you also:
1. Fix the lint errors on the PR
2. Make sure the docs compile correctly so I can check how it renders 😄 You test with `npm run dev --workspace docs`

Screenshot 2025-05-07 at 12 14 36
Screenshot 2025-05-07 at 12 15 07

cli/cmd/docs/generate_project.go
```diff
@@ -0,0 +1,451 @@
...
+	for _, key := range m.Keys() {
+		value, _ := m.Get(key)
+
+		// Check if the key is a comment (starts with "_comment_")
+		if strings.HasPrefix(key, "_comment_") {
+			if strVal, ok := value.(string); ok {
+				pendingComment = strVal
+			}
+			continue
+		}
```
begelundmuller, line 264: Is this a normal way of doing examples in JSON Schema? It seems a little hard to read/edit when looking in the JSON. 

We don't necessarily need examples in the JSON schema itself. We can have them in separate example docs instead.

docs/docs/reference/new-project-files/alert.md
```diff
@@ -0,0 +1,163 @@
...
+  - **`type`**  - _[string]_ - type of resource 
+
+  - **`name`**  - _[string]_ - name of resource  _(required)_
+
+  *option 2* - _[string]_ - A string reference like `<resource-name>` or `<type/resource-name>`.
+
+**`dev`**  - _[object]_ - Overrides properties in development 
+
+**`prod`**  - _[object]_ - Overrides properties in production 
+
```
begelundmuller, line 29: Would it be possible to have the common properties in a separate `## Common properties` section of each page?

docs/docs/reference/new-project-files/alert.md
```diff
@@ -0,0 +1,163 @@
...
+**`refs`**  - _[array of oneOf]_ - List of resource references, each as a string or map. 
+
+  *option 1* - _[object]_ - An object reference with at least a `<name>` and `<type>`.
+
+  - **`type`**  - _[string]_ - type of resource 
+
+  - **`name`**  - _[string]_ - name of resource  _(required)_
+
+  *option 2* - _[string]_ - A string reference like `<resource-name>` or `<type/resource-name>`.
```
begelundmuller, line 24: Can we make option 2 the only documented option in the schema? And keep option 1 as an internal implementation detail. Every time the feature has been used, we have used option 2.

docs/docs/reference/new-project-files/alert.md
```diff
@@ -0,0 +1,163 @@
...
+
+**`prod`**  - _[object]_ - Overrides properties in production 
+
+**`display_name`**  - _[string]_ - Refers to the display name for the alert 
+
+**`refresh`**  - _[object]_ - Specifies the refresh schedule that Rill should follow to re-ingest and update the underlying data  _(required)_
+
+  - **`cron`**  - _[string]_ - A cron expression that defines the execution schedule 
+
+  - **`every`**  - _[string]_ - Run at a fixed interval using a Go duration string (e.g., '1h', '30m', '24h'). See: https://pkg.go.dev/time#ParseDuration
```
begelundmuller, line 36: Let's remove `every` from the docs – we're probably going to deprecate it, and always use `cron`

docs/docs/reference/new-project-files/model.md
```diff
@@ -0,0 +1,541 @@
...
+
+**`dsn`**  - _[string]_ - DSN(Data Source Name) for the ClickHouse connection 
+
+**`username`**  - _[string]_ - Username for authentication 
+
+**`password`**  - _[string]_ - Password for authentication 
+
+**`host`**  - _[string]_ - Host where the ClickHouse instance is running 
+
+**`port`**  - _[integer]_ - Port where the ClickHouse instance is accessible
```
begelundmuller, line 247: The properties listed here are for `type: connector`, not `type: model`. Models have fields like `sql:` and `output: incremental_strategy`, etc.

docs/docs/reference/new-project-files/explore-dashboard.md
```diff
@@ -0,0 +1,206 @@
+---
+note: GENERATED. DO NOT EDIT.
+title: Explore Dashboard YAML
```
begelundmuller, line 3: Just call the file `explore.yaml` – every file name should match the `type: <name>` identifier so it's predictable to link to

begelundmuller (commented):
I added some inline comments below. But as an overall comment when looking at the docs: can we use some kind of headings for the different properties? Especially the top-level properties. Two reasons for this:

- It will be very useful to be able to link directly to a property, so we want to output headings with an `id="property_name"` tag so we can get `<url>#property_name` links directly to the properties.
- When properties have nested types or options, it can be difficult to find out what they relate to because the design becomes quite flag. I'm wondering if different sizes, spacing or indentation could somehow make this easier to read. For example, see this screenshot:



Screenshot 2025-05-08 at 21 51 45

docs/docs/reference/new-project-files/index.md
```diff
@@ -0,0 +1,38 @@
+---
+note: GENERATED. DO NOT EDIT.
+title: New Project files
+sidebar_position: 30
+---
+## Overview
```
begelundmuller, line 6: Can you call this `YAML Syntax` and the path `docs/reference/yaml/index.md`? So it doesn't conflict with the current "Project files" section

cli/cmd/docs/generate_project.go
NamanMahor: Yes. below is the ways in json schema for examples
```
"examples": [
]
```

json donot support comment so i have to tweak it to support \_comment\_* as comment.

we can use this as below if we want separate example docs instead
```
"examples": [
{
 "title": "Example one for api",
 "refrence_urls": "https://github.com/rilldata/rill-examples/blob/main/my-rill-tutorial/2.%20Rill%20Advanced/apis/metrics_view_api.yaml"
},
{
 "title": "Example two for api",
 "refrence_urls": "https://github.com/rilldata/rill-examples/blob/main/my-rill-tutorial/2.%20Rill%20Advanced/apis/metrics_view_api.yaml"
}
]
```
And can list he example reference doc in end of md file

docs/docs/reference/new-project-files/alert.md
NamanMahor: removing option 1 then

docs/docs/reference/new-project-files/alert.md
NamanMahor: remove it

docs/docs/reference/new-project-files/explore-dashboard.md
NamanMahor: done

docs/docs/reference/new-project-files/index.md
NamanMahor: done

docs/docs/reference/new-project-files/alert.md
NamanMahor: actually these properties are defined at different places so can't combined them in one common properties header

e.g 
`type` is defined differently every resource schem with different constant value.

`dev` and `prod` are also defined differently in every resource schema because it point to different properties depending upon resource type

only name and ref is defined as common_properties definitions which is being used by every resource.  


do can't mark them in single heading as common_properties only name and ref can be in common_properties.

docs/docs/reference/new-project-files/model.md
NamanMahor: done

runtime/parser/schema/project.schema.json
```diff
@@ -203,54 +226,55 @@
...
+      "title": "API YAML",
+      "description": "In your Rill project directory, create a new file name `<api-name>.yaml` in the `apis` directory containing a custom API definition. See comprehensive documentation on how to define and use [custom APIs](/integrate/custom-apis/index.md)",
       "examples": [
         {
+          "_comment_heading": "Example: This api returns the top 10 authors by net line changes since the specified date provided in the arguments.",
           "type": "api",
-          "connector": "clickhouse",
-          "sql": "select * from my_model"
+          "name": "metrics_view_api",
+            "metrics_sql": "SELECT author_name, net_line_changes\nFROM advanced_metrics_view\n  where author_date > '{{ .args.date }}'\n  order by net_line_changes DESC\n  limit 10"
```
begelundmuller, line 237: I still don't understand this way of writing examples in JSON schema. It feels quite weird to break out a YAML snippet in separate entries like this. I would have expected just a single string with line breaks and similar using escape characters like `\n`.

If you are sure this is the recommended way, can you share a link to the place in the JSON schema spec that recommends this?

begelundmuller (changes requested):

docs/docs/reference/new-project-files/alert.md
begelundmuller: Could it be hard-coded? We know the common properties are limited to a few known names.

begelundmuller (approved):

PR #7283, opened 2025-05-07
Require `force_superuser_access` flag for every sudo request

begelundmuller:
We've had a lot of issues where internal superusers see weird/broken screens due to some endpoints allowing direct superuser access without being in a superuser context. 

This PR fixes it by adding and requiring a `force_superuser_access` in all APIs when superuser access override is possible. It also updates the CLI to set this flag when using `rill sudo` commands. 

This PR breaks `rill sudo` support in older versions of Rill.

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!


admin/server/deployment.go
```diff
@@ -124,7 +124,7 @@ func (s *Server) GetDeploymentCredentials(ctx context.Context, req *adminv1.GetD
 	permissions := claims.ProjectPermissions(ctx, proj.OrganizationID, proj.ID)
 
 	// If the user is not a superuser, they must have ManageProd permissions
-	if !permissions.ManageProd && !claims.Superuser(ctx) {
```
pjain1, line 127: we are disallowing deploy access to superusers now, if yes then comment above should be updated

admin/server/deployment.go
```diff
@@ -219,7 +219,7 @@ func (s *Server) GetIFrame(ctx context.Context, req *adminv1.GetIFrameRequest) (
...
```
pjain1, line 222: same as above

admin/server/projects.go
```diff
@@ -1203,15 +1205,16 @@ func (s *Server) SetProjectMemberUserRole(ctx context.Context, req *adminv1.SetP
...
-	if !claims.Superuser(ctx) {
-		return nil, status.Error(codes.PermissionDenied, "superuser permission required to get clone credentials")
-	}
 	observability.AddRequestAttributes(ctx,
 		attribute.String("args.org", req.Organization),
 		attribute.String("args.project", req.Project),
 	)
 
+	claims := auth.GetClaims(ctx)
+	if !claims.Superuser(ctx) {
```
pjain1, line 1214: here force access flag is not added because its just a cli only API?

admin/server/bookmarks.go
```diff
@@ -60,7 +60,7 @@ func (s *Server) GetBookmark(ctx context.Context, req *adminv1.GetBookmarkReques
 		permissions.ReadProd = true
 	}
 
-	if !permissions.ReadProject && !claims.Superuser(ctx) {
```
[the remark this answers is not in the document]
begelundmuller: Yes, seems like it was a copy-paste error when added – because no `sudo` commands use bookmarks

admin/server/projects.go
begelundmuller: It's because it's a sudo-only API (no normal users can call it). The `force_superuser_access` flag only makes sense for APIs that allow calls from normal users as well as superusers.

pjain1 (approved):

PR #7289, opened 2025-05-08
Prevent alerts/reports from disappearing when changing the default Git branch

begelundmuller:
**Changes:**
- Fixes an issue where alerts/reports disappear when changing the default Git branch, including when connecting/disconnecting Git.
- Changes the `virtual_files` APIs to center around environments instead of branches. This will also better integrate with the upcoming cloud editing changes.
- Also improves the `admin` driver to better return/retry clone errors.

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!


runtime/drivers/admin/admin.go
```diff
@@ -287,7 +287,23 @@ func (h *Handle) cloneOrPull(ctx context.Context) error {
 		defer cancel()
 
 		r := retrier.New(retrier.ExponentialBackoff(pullRetryN, pullRetryWait), retryErrClassifier{})
-		h.syncErr = r.Run(func() error { return h.cloneOrPullInner(ctx) })
+		h.syncErr = r.Run(func() error {
+			h.syncErr = h.cloneOrPullInner(ctx)
```
k-anshul, line 291: I do not understand this change.
1. Setting `h.syncErr` to err inside the `Run` will mean that temporarily `syncErr` can be set to nil. If the first run fails but the second run passes ideally it should not set `syncErr` to nil ever.
2. Resetting everything will mean that we can't server from existing state as well (in case of pulls).

k-anshul (commented):
BTW did you also consider changing `branch` in `virtual_files` when prod branch is changed. The net effect will be the same but logically it maintains the concept that all project files are tied to a branch.

runtime/drivers/admin/admin.go
begelundmuller: 1. Yeah it looks weird. I think there was a reason I did it, but don't remember. The caller here holds `h.repoMu.Lock` for the full duration of the retries, so the `syncErr` is never read while this code runs. I've just pushed a change to use a local scope err instead.
2. Yeah it's not great to reset, but I looked at the code under `cloneOrPullInner` and realized there are several error scenarios that are not recoverable – so it just felt safer to block repo reads and wait until a new successful clone happens.

Also note that repo reads are assumed to be unavailable sometimes – they should not block serving, and as long as it is not cloned successfully, the parser just blocks and doesn't update the catalog.

What do you think?

k-anshul (approved):

PR #7296, opened 2025-05-12
Clickhouse: partition overwrite by default + refactors

begelundmuller:
This PR makes partitioned Clickhouse models use the `partition_overwrite` strategy by default. It does so by automatically injecting a column named `__rill_partition` with the current partition ID into the model's `SELECT` statement. After this change, the simplest possible incremental and partitioned Clickhouse model looks like this:
```yaml
type: model
incremental: true
partitions:
  glob: s3://bucket/path/**
sql: SELECT * FROM s3('{{.partition.uri}}')
```

Contributes to https://github.com/rilldata/rill-private-issues/issues/1632

**Other changes:**
- Fixes an issue where some config overrides where not handled when DSN was passed.
- Uses `ReplicatedMergeTree` by default instead of `MergeTree` for Clickhouse tables on clusters.
- Renames `connector.settings_override` to `connector.query_settings` (breaking change).
- Renames `distributed.settings` and `distributed.sharding_key` to `distributed_settings` and `distributed_sharding_key` (breaking change; confirmed not used in any production projects).
- Refactors validation and rewriting of input/output properties into a single function `validateAndApplyDefaults`.
- Moves the resolver test utilities into `testruntime` package so they can be used for driver tests.
- Various other minor refactors in the Clickhouse driver.

**Upcoming changes:**
The changes in this PR were made in preparation for a few other improvements, which I will be submitting as separate PRs. Sharing them here for context:
- [ ] Use `mapstructure.WeakDecodeMetadata` in `validateAndApplyDefaults` to error for unknown or unsupported properties
- [ ] Support partition overwrite by default in DuckDB

**Checklist:**
- [x] Covered by tests
- [x] Checked if the docs need to be updated (note: not yet)
- [x] I'm proud of this work!

### DIFF
--- a/runtime/drivers/clickhouse/model_manager.go
+++ b/runtime/drivers/clickhouse/model_manager.go
@@ -15,15 +15,17 @@ type ModelInputProperties struct {
 	SQL string `mapstructure:"sql"`
 }
 
-func (p *ModelInputProperties) Validate() error {
-	return nil
-}
-
 type ModelOutputProperties struct {
-	Table               string                      `mapstructure:"table"`
-	Materialize         *bool                       `mapstructure:"materialize"`
-	UniqueKey           []string                    `mapstructure:"unique_key"`
+	// Table is the name of the table to create. If not specified, the model name is used.
+	Table string `mapstructure:"table"`
+	// Materialize is a flag to indicate if the model should be materialized as a physical table or dictionary.
+	// If false, the model will be created as a view.
+	// If unspcecified, a default is selected based on context.
+	Materialize *bool `mapstructure:"materialize"`
+	// IncrementalStrategy is the strategy to use for incremental inserts.
 	IncrementalStrategy drivers.IncrementalStrategy `mapstructure:"incremental_strategy"`
+	// UniqueKey is the unique key for the model. This is used for the incremental strategy "merge".
+	UniqueKey []string `mapstructure:"unique_key"`
 	// Typ to materialize the model into. Possible values include `TABLE`, `VIEW` or `DICTIONARY`. Optional.
 	Typ string `mapstructure:"type"`
 	// Columns sets the column names and data types. If unspecified these are detected from the select query by clickhouse.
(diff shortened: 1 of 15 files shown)


runtime/drivers/clickhouse/model_manager.go
```diff
@@ -55,97 +57,142 @@ type ModelOutputProperties struct {
...
+
+	// We want to use partition_overwrite as the default incremental strategy for models with partitions.
+	// This requires us to inject the partition key into the SQL query, so this only works for SQL models.
+	if op.IncrementalStrategy == drivers.IncrementalStrategyUnspecified && opts.PartitionRun && ip != nil && ip.SQL != "" {
+		if op.EngineFull != "" {
+			return fmt.Errorf("you must provide an explicit `incremental_strategy` when using `engine_full` with a partitioned model")
+		}
+		ip.SQL = fmt.Sprintf("SELECT %s AS __rill_partition, * FROM (%s\n)", safeSQLString(opts.PartitionKey), ip.SQL)
+		op.IncrementalStrategy = drivers.IncrementalStrategyPartitionOverwrite
+		op.PartitionBy = "__rill_partition"
```
k-anshul, line 151: Nit: As a best practice clickhouse prefixes virtual columns with underscore so we should avoid adding `_` prefix in this column.

runtime/drivers/clickhouse/model_manager.go
```diff
@@ -173,8 +220,13 @@ func (p *ModelOutputProperties) tblConfig() string {
 
 	// settings
 	if p.TableSettings != "" {
-		fmt.Fprintf(&sb, " %s", p.TableSettings)
+		if strings.HasPrefix(p.TableSettings, "SETTINGS") || strings.HasPrefix(p.TableSettings, "settings") {
```
k-anshul, line 223: 1. Nit: `strings.EqualFold(p.TableSettings, "SETTINGS")` ?
2. Also add a comment that this is required because of an earlier bug because such a check is not required for other props?

runtime/drivers/clickhouse/olap.go
```diff
@@ -125,8 +122,8 @@ func (c *Connection) Query(ctx context.Context, stmt *drivers.Statement) (res *d
 		return nil, err
 	}
 
-	if c.config.SettingsOverride != "" {
-		stmt.Query += "\n SETTINGS " + c.config.SettingsOverride
+	if c.config.QuerySettings != "" {
```
k-anshul, line 125: I preferred `SettingsOverride`. It makes it explicit that it will override the settings set by Rill internally.

k-anshul (approved):

runtime/drivers/clickhouse/olap.go
k-anshul: `QuerySettingsOverride` is also fine. Want to emphasize on `override`.

runtime/drivers/clickhouse/model_manager.go
begelundmuller: 1. EqualFold checks the whole string, not the prefix, so I don't think that would work?
2. Good idea, just added that

runtime/drivers/clickhouse/model_manager.go
begelundmuller: I double checked and Clickhouse doesn't appear to have issues with such columns unless they conflict with built-in virtual column names. So as long as we have a `__rill` prefix, we should be good.

PR #7297, opened 2025-05-12
feat: Add a mirroring GetProject API to local service to get project permissions during deploy

AdityaHegde:
During deploy we need to prefetch project permissions and show a modal for requesting access.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


proto/rill/admin/v1/api.proto
```diff
@@ -971,6 +971,8 @@ message GetProjectRequest {
   uint32 access_token_ttl_seconds = 3;
   bool superuser_force_access = 5;
   bool issue_superuser_token = 4;
+  // Skips fetching deployment and jwt related to querying deployment
+  bool skip_deployment = 6;
```
begelundmuller: Are you going to be calling this endpoint in a poll loop? Otherwise, I'd prefer to just keep the API simple and always return the deployment info (it's relatively cheap to generate).

AdityaHegde: Right now it is used only when we select a project to re-deploy to. I can remove this if it is cheap.

PR #7299, opened 2025-05-12
feat: Rill time final syntax tweak

AdityaHegde:
INSERT DESCRIPTION HERE

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/runtime/pkg/rilltime/rilltime.go
+++ b/runtime/pkg/rilltime/rilltime.go
@@ -31,12 +31,16 @@ var (
 		// this has to be at the end
 		{"TimeZone", `{.+?}`},
 		{"ISOTime", isoTimePattern},
+		{"Prefix", `[+\-]`},
+		{"Suffix", `[\^\$]`},
 		{"AnchorPrefix", `[+\-<>]`},
 		{"Current", "[~]"},
 		{"Number", `\d+`},
+		{"Snap", `[/]`},
 		{"To", `(?i)to`},
 		{"By", `(?i)by`},
 		{"Of", `(?i)of`},
+		{"As", `(?i)as`},
 		// needed for misc. direct character references used
 		{"Punct", `[-[!@#$%^&*()+_={}\|:;"'<,>.?/]]`},
 		{"Whitespace", `[ \t]+`},
--- a/runtime/pkg/rilltime/rilltimefinal.go
+++ b/runtime/pkg/rilltime/rilltimefinal.go
@@ -1,0 +1,259 @@
+package rilltime
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/alecthomas/participle/v2"
+	"github.com/rilldata/rill/runtime/pkg/timeutil"
+)
+
+var (
+	rillTimeParserFinal = participle.MustBuild[ExpressionFinal](
+		participle.Lexer(rillTimeLexer),
+		participle.Elide("Whitespace"),
+	)
+	snapToStart = "^"
+	snapToEnd   = "$"
+)
+
+type ExpressionFinal struct {
+	Interval *Interval `parser:"@@"`
+
+	timeZone *time.Location
+}
+
+type Interval struct {
+	Ordinal  *OrdinalInterval  `parser:"( @@"`
+	StartEnd *StartEndInterval `parser:"| @@)"`
+}
+
+type OrdinalInterval struct {
+	Parts             []*OrdinalIntervalPart `parser:"@@ (Of @@)*"`
+	IntervalAnchor    *StartEndInterval      `parser:"( Of @@"`
+	PointInTimeAnchor *PointInTime           `parser:"| As Of @@)"`
+}
+
+type OrdinalIntervalPart struct {
+	Grain string `parser:"@Grain"`
+	Num   int    `parser:"@Number"`
+}
+
+type StartEndInterval struct {
+	Start      *PointInTime `parser:"@@"`
+	EndOfStart bool         `parser:"( '!'"`
+	End        *PointInTime `parser:"| To @@)"`
+}
+
+type PointInTime struct {
+	Relative *RelativePointInTime `parser:"( @@"`
+	Labelled *LabelledPointInTime `parser:"| @@)"`
+}
+
+type RelativePointInTime struct {
+	Prefix   *string   `parser:"@Prefix?"`
+	Duration *Duration `parser:"@@"`
+	Snap     *string   `parser:"(Snap @Grain)?"`
+	Suffix   string    `parser:"@Suffix"`
+}
+
+type LabelledPointInTime struct {
+	Earliest  bool `parser:"( @Earliest"`
+	Now       bool `parser:"| @Now"`
+	Latest    bool `parser:"| @Latest"`
+	Watermark bool `parser:"| @Watermark)"`
+}
+
+type Duration struct {
+	Num   *int   `parser:"@Number?"`
+	Grain string `parser:"@Grain"`
+}
+
+func ParseFinal(from string, parseOpts ParseOptions) (*ExpressionFinal, error) {
+	//tokens, err := rillTimeParserFinal.Lex("", strings.NewReader(from))
+	//if err != nil {
+	//	return nil, err
+	//}
+	//for _, token := range tokens {
+	//	fmt.Println(token.Type, token.Value)
+	//}
+
+	rt, err := rillTimeParserFinal.ParseString("", from)
+	if err != nil {
+		return nil, err
+	}
+
+	err = rt.parse(parseOpts)
+	if err != nil {
+		return nil, err
+	}
+
+	return rt, nil
+}
+
+func (e *ExpressionFinal) parse(parseOpts ParseOptions) error {
+	e.timeZone = time.UTC
+	if parseOpts.TimeZoneOverride != nil {
+		e.timeZone = parseOpts.TimeZoneOverride
+	} else if parseOpts.DefaultTimeZone != nil {
+		e.timeZone = parseOpts.DefaultTimeZone
+	}
+
+	if e.Interval != nil {
+		return e.Interval.parse()
+	}
+	return nil
+}
+
+func (e *ExpressionFinal) Eval(evalOpts EvalOptions) (time.Time, time.Time, timeutil.TimeGrain) {
+	if evalOpts.FirstDay == 0 {
+		evalOpts.FirstDay = 1
+	}
+	if evalOpts.FirstMonth == 0 {
+		evalOpts.FirstMonth = 1
+	}
+
+	cur := evalOpts.Watermark
+
+	if e.Interval != nil {
+		return e.Interval.eval(evalOpts, cur, e.timeZone)
+	}
+
+	return cur, cur, timeutil.TimeGrainUnspecified
+}
+
+func (i *Interval) parse() error {
+	if i.StartEnd != nil {
+		return i.StartEnd.parse()
+	}
+	return nil
+}
+
+func (i *Interval) eval(evalOpts EvalOptions, tm time.Time, tz *time.Location) (time.Time, time.Time, timeutil.TimeGrain) {
+	if i.StartEnd != nil {
+		return i.StartEnd.eval(evalOpts, tm, tz)
+	} else if i.Ordinal != nil {
+		return i.Ordinal.eval(evalOpts, tm, tz)
+	}
+	return tm, tm, timeutil.TimeGrainUnspecified
+}
+
+func (o *OrdinalInterval) eval(evalOpts EvalOptions, tm time.Time, tz *time.Location) (time.Time, time.Time, timeutil.TimeGrain) {
+	if o.PointInTimeAnchor != nil {
+		tm, _ = o.PointInTimeAnchor.eval(evalOpts, tm, tz)
+	} else if o.IntervalAnchor != nil {
+		tm, _, _ = o.IntervalAnchor.eval(evalOpts, tm, tz)
+	}
+
+	start := tm
+	for _, part := range o.Parts {
+		start = part.eval(evalOpts, start, tz)
+	}
+	i := len(o.Parts) - 1
+	for i >= 0 {
+		start = o.Parts[i].eval(evalOpts, start, tz)
+		i--
+	}
+
+	tg := grainMap[o.Parts[0].Grain]
+
+	end := timeutil.OffsetTime(start, tg, 1)
+
+	return start, end, lowerOrderMap[tg]
+}
+
+func (o *OrdinalIntervalPart) eval(evalOpts EvalOptions, tm time.Time, tz *time.Location) time.Time {
+	tg := grainMap[o.Grain]
+	offset := o.Num - 1
+
+	tm = timeutil.OffsetTime(tm, tg, offset)
+	tm = truncateWithCorrection(tm, tg, tz, evalOpts.FirstDay, evalOpts.FirstMonth)
+
+	return tm
+}
+
+func (s *StartEndInterval) parse() error {
+	if !s.EndOfStart {
+		return nil
+	}
+
+	if s.Start.Relative == nil {
+		return fmt.Errorf("start must be relative")
+	}
+
+	s.Start.Relative.Snap = &snapToStart
+	s.End = &PointInTime{
+		Relative: &RelativePointInTime{
+			Prefix:   s.Start.Relative.Prefix,
+			Duration: s.Start.Relative.Duration,
+			Snap:     &snapToEnd,
+			Suffix:   s.Start.Relative.Suffix,
+		},
+	}
+
+	return nil
+}
+
+func (s *StartEndInterval) eval(evalOpts EvalOptions, tm time.Time, tz *time.Location) (time.Time, time.Time, timeutil.TimeGrain) {
+	start, startTg := s.Start.eval(evalOpts, tm, tz)
+	end, endTg := s.End.eval(evalOpts, tm, tz)
+	tg := endTg
+	if endTg == timeutil.TimeGrainUnspecified {
+		tg = startTg
+	}
+	return start, end, tg
+}
+
+func (p *PointInTime) eval(evalOpts EvalOptions, tm time.Time, tz *time.Location) (time.Time, timeutil.TimeGrain) {
+	if p.Relative != nil {
+		return p.Relative.eval(evalOpts, tm, tz)
+	} else if p.Labelled != nil {
+		return p.Labelled.eval(evalOpts, tm), timeutil.TimeGrainUnspecified
+	}
+	return tm, timeutil.TimeGrainUnspecified
+}
+
+func (r *RelativePointInTime) eval(evalOpts EvalOptions, tm time.Time, tz *time.Location) (time.Time, timeutil.TimeGrain) {
+	sign := -1
+	if r.Prefix != nil && *r.Prefix == "+" {
+		sign = 1
+	}
+	tm = r.Duration.Offset(tm, sign)
+
+	tg := grainMap[r.Duration.Grain]
+	if r.Snap != nil {
+		tg = grainMap[*r.Snap]
+	}
+
+	// `$` suffix means snap to end. So add 1 to the offset before truncating.
+	if r.Suffix == "$" {
+		tm = timeutil.OffsetTime(tm, tg, 1)
+	}
+	tm = truncateWithCorrection(tm, tg, tz, evalOpts.FirstDay, evalOpts.FirstMonth)
+
+	return tm, tg
+}
+
+func (l *LabelledPointInTime) eval(evalOpts EvalOptions, tm time.Time) time.Time {
+	if l.Earliest {
+		return evalOpts.MinTime
+	} else if l.Now {
+		return evalOpts.Now
+	} else if l.Latest {
+		return evalOpts.MaxTime
+	} else if l.Watermark {
+		return evalOpts.Watermark
+	}
+	return tm
+}
+
+func (d *Duration) Offset(tm time.Time, sign int) time.Time {
+	n := 0
+	if d.Num != nil {
+		n = *d.Num
+	}
+	n *= sign
+
+	tg := grainMap[d.Grain]
+	return timeutil.OffsetTime(tm, tg, n)
+}
--- a/runtime/pkg/rilltime/rilltimefinal_test.go
+++ b/runtime/pkg/rilltime/rilltimefinal_test.go
@@ -1,0 +1,52 @@
+package rilltime
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/rilldata/rill/runtime/pkg/timeutil"
+	"github.com/stretchr/testify/require"
+)
+
+func Test_EvalFinal(t *testing.T) {
+	now := parseTestTime(t, "2025-03-12T10:32:36Z")
+	minTime := parseTestTime(t, "2020-01-01T00:32:36Z")
+	maxTime := parseTestTime(t, "2025-03-11T06:32:36Z")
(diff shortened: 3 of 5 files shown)


briangregoryholmes (approved):

briangregoryholmes:
Added a couple edge cases to our Luxon based tests and fixed one that was failing. I don't necessarily mind keeping this file around.

AdityaHegde:
Ya sounds good. The file will be helpful. But i will remove the auto trigger from rill-dev route before merging.

runtime/pkg/rilltime/rilltime.go
```diff
@@ -271,277 +338,303 @@ func (e *Expression) Eval(evalOpts EvalOptions) (time.Time, time.Time, timeutil.
...
+	end := start
 
 	tg := timeutil.TimeGrainUnspecified
-	if len(l.anchorParts) > 0 {
-		tg = l.anchorParts[0].grain()
+	if o.Starting {
+		end, tg = o.Duration.offset(start, 1)
+	} else if o.Ending {
+		start, tg = o.Duration.offset(end, -1)
 	}
```
begelundmuller, line 404: What is the thinking around adding a grain here?

runtime/pkg/rilltime/rilltime.go
```diff
@@ -271,277 +338,303 @@ func (e *Expression) Eval(evalOpts EvalOptions) (time.Time, time.Time, timeutil.
...
 
-	err := lpa.parse()
-	if err != nil {
-		return nil, err
+func (o *StartEndInterval) eval(evalOpts EvalOptions, tm time.Time, tz *time.Location) (time.Time, time.Time, timeutil.TimeGrain) {
+	start, startTg := o.Start.eval(evalOpts, tm, tz)
+	end, endTg := o.End.eval(evalOpts, tm, tz)
+	// Correction for labeled ends points. We need to add +1ms to make sure the final point is included as our end time is exclusive.
+	if o.End.Labeled != nil {
+		end = timeutil.OffsetTime(end, timeutil.TimeGrainMillisecond, 1)
```
begelundmuller, line 446: This really isn't safe – we don't know what the smallest time granularity is. It can vary between nanosecond and second granularity depending on the timestamp type used.

runtime/pkg/rilltime/rilltime.go
```diff
@@ -271,277 +338,303 @@ func (e *Expression) Eval(evalOpts EvalOptions) (time.Time, time.Time, timeutil.
...
+		secondarySnap = grainMap[tgs[1]]
 	}
 
-	if t.Prefix == nil {
-		// Without a prefix of either +/- we actually need a time range.
-		// EG: 7D is 7 day period.
+	// `$` suffix means snap to end. So add 1 to the offset before truncating.
+	if *g.Suffix == "$" {
+		tm = timeutil.OffsetTime(tm, tg, 1)
+	}
```
begelundmuller, line 575: What if it's already at the boundary? Then it will snap to the boundary that is one grain higher. Is that expected?

runtime/pkg/rilltime/rilltime_test.go
```diff
@@ -1,176 +1,576 @@
...
+	for i, grain := range grains {
+		tmGrain := timeutil.TimeGrain(i + 2)
+
+		// Current period
+		testCases = append(testCases, testCase{
+			timeRange: fmt.Sprintf("%[1]s^ to %[1]s$", grain),
+			start:     curPeriodStarts[i],
+			end:       curPeriodEnds[i],
+			grain:     expectedGrains[i],
+		})
```
begelundmuller, line 95: The automation here has made these test cases hard to follow. They are also no longer useful as examples we can easily reference when writing docs or doing support.

Even if it means some repetition, I favor the previous test syntax that clearly shows how an expression evaluates:
```
{"W1 to W3", "2025-03-03T00:00:00Z", "2025-03-17T00:00:00Z", timeutil.TimeGrainDay},
{"W1 to W3 by W", "2025-03-03T00:00:00Z", "2025-03-17T00:00:00Z", timeutil.TimeGrainWeek},
...
```

runtime/pkg/rilltime/rilltime_test.go
```diff
@@ -1,176 +1,576 @@
...
+	minTime := "2020-01-01T00:32:36Z"
+	maxTime := "2025-07-01T00:00:00Z"   // month and quarter boundary
+	watermark := "2025-05-12T00:00:00Z" // day and week boundary
+	testCases := []testCase{
+		{"-2D^ to D^", "2025-05-10T00:00:00Z", "2025-05-12T00:00:00Z", timeutil.TimeGrainDay, 1, 1},
+		{"-2D$ to D$", "2025-05-11T00:00:00Z", "2025-05-13T00:00:00Z", timeutil.TimeGrainDay, 1, 1},
+		{"-4D^ to -2D^", "2025-05-08T00:00:00Z", "2025-05-10T00:00:00Z", timeutil.TimeGrainDay, 1, 1},
+		{"-1D^ to D^", "2025-05-11T00:00:00Z", "2025-05-12T00:00:00Z", timeutil.TimeGrainHour, 1, 1},
+		{"D^ to D$", "2025-05-12T00:00:00Z", "2025-05-13T00:00:00Z", timeutil.TimeGrainHour, 1, 1},
+		{"-2D^ to watermark", "2025-05-10T00:00:00Z", "2025-05-12T00:00:00.001Z", timeutil.TimeGrainDay, 1, 1},
```
begelundmuller, line 445: The watermark is `2025-05-12T00:00:00Z`, so why does this end at 1ms past that? I'd have expected it to either "round up" by a whole grain (`D` in this case?) or to stay at the boundary.

runtime/pkg/rilltime/rilltime_test.go
```diff
@@ -1,176 +1,576 @@
...
+		{"2025-02-20T01", "2025-02-20T01:00:00Z", "2025-02-20T02:00:00Z", timeutil.TimeGrainMinute, 1, 1},
+		{"2025-02-20", "2025-02-20T00:00:00Z", "2025-02-21T00:00:00Z", timeutil.TimeGrainHour, 1, 1},
+		{"2025-02", "2025-02-01T00:00:00Z", "2025-03-01T00:00:00Z", timeutil.TimeGrainDay, 1, 1},
+		{"2025", "2025-01-01T00:00:00Z", "2026-01-01T00:00:00Z", timeutil.TimeGrainMonth, 1, 1},
+	}
+
+	runTests(t, testCases, now, minTime, maxTime, watermark, nil)
 }
 
 func Test_Eval_watermark_on_boundary(t *testing.T) {
```
begelundmuller, line 434: Test function names should use pascal case, optionally using underscores for namespacing, e.g. `TestEval_WatermarkOnBoundary`

runtime/pkg/rilltime/rilltime_test.go
```diff
@@ -1,176 +1,576 @@
...
+			timeRange: fmt.Sprintf("2%[1]s starting -2%[2]s/%[3]s^", grainPair.grain, grainPair.higherGrain, snapGrain),
+			start:     higherPeriod,
+			end:       firstEnd,
+			grain:     tmGrain,
+		})
+		testCases = append(testCases, testCase{
+			timeRange: fmt.Sprintf("-2%[2]s/%[3]s^ to -2%[2]s/%[3]s^+2%[1]s", grainPair.grain, grainPair.higherGrain, snapGrain),
+			start:     higherPeriod,
+			end:       firstEnd,
+			//grain:     tmGrain, // TODO: certain cases dont correct identify grain
```
begelundmuller, line 319: Does this need to be addressed?

runtime/pkg/rilltime/rilltime_test.go
```diff
@@ -1,176 +1,576 @@
...
+		{"rill-QTD", "2025-04-01T00:00:00Z", "2025-05-14T00:00:00Z", timeutil.TimeGrainDay, 1, 1},
+		{"rill-YTD", "2025-01-01T00:00:00Z", "2025-05-14T00:00:00Z", timeutil.TimeGrainMonth, 1, 1},
+
+		{"rill-PDC", "2025-05-12T00:00:00Z", "2025-05-13T00:00:00Z", timeutil.TimeGrainHour, 1, 1},
+		{"rill-PWC", "2025-05-05T00:00:00Z", "2025-05-12T00:00:00Z", timeutil.TimeGrainDay, 1, 1},
+		{"rill-PMC", "2025-04-01T00:00:00Z", "2025-05-01T00:00:00Z", timeutil.TimeGrainDay, 1, 1},
+		{"rill-PQC", "2025-01-01T00:00:00Z", "2025-04-01T00:00:00Z", timeutil.TimeGrainMonth, 1, 1},
+		{"rill-PYC", "2024-01-01T00:00:00Z", "2025-01-01T00:00:00Z", timeutil.TimeGrainMonth, 1, 1},
+
+		{"inf", "2020-01-01T00:32:36Z", "2025-05-14T06:32:36.001Z", timeutil.TimeGrainUnspecified, 1, 1},
```
begelundmuller, line 494: Same comment – this can end up excluding the latest rows for timestamp types that use second granularity.

begelundmuller (dismissed):

runtime/pkg/rilltime/rilltime.go
briangregoryholmes: It doesn't snap to the boundary that is one grain higher, but I may misunderstand what you mean. If your latest timestamp is `2025-05-20T00:00:00Z`, then `d^ to d$` gives you the full day and `h^ to h$` would give you the full first hour.

begelundmuller (commented):
It would also be nice with some test cases for broken/malformed time expressions? I.e. to think about common syntax errors and ensuring that we return legible error messages for them.

runtime/pkg/rilltime/rilltime.go
```diff
@@ -263,6 +299,37 @@ func ParseCompatibility(timeRange, offset string) error {
 	return nil
 }
 
+func (e *Expression) parse(parseOpts ParseOptions) error {
```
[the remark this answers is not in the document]
AdityaHegde: It calls `parse` on components under it, mainly the `AbsoluteTime`.

runtime/pkg/timeutil/timeutil.go
```diff
@@ -186,6 +186,7 @@ func AddTimeProto(to time.Time, tg runtimev1.TimeGrain, count int) time.Time {
 }
 
 func OffsetTime(tm time.Time, tg TimeGrain, n int) time.Time {
+	// TODO: edge case where offsetting by a month to a date that is not valid should snap to the last day of the month
```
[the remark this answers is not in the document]
AdityaHegde: Yes. Lemme address it

runtime/pkg/rilltime/rilltime.go
AdityaHegde: It will offset by 1 for the time grain specified. Added a test for `h^ to h$` in `Test_Eval_watermark_on_boundary` to make it clear. Will also add more comments.

runtime/pkg/rilltime/rilltime_test.go
AdityaHegde: This returns the grain we show by default. I think this is not that big a deal right now. There will definitely be a follow up once users start using this, we can address this then.

runtime/pkg/rilltime/rilltime_test.go
AdityaHegde: Sounds good. Lemme add explicit entries.

runtime/pkg/rilltime/rilltime_test.go
briangregoryholmes: When you ask for the last 7 days, there are four things people could mean:

1. `-7d to now`, meaning return all of the data points within the last 7 day period as a duration
2. `-7d^ to d^`, meaning 7 days, but exclude the "current" day, which may be incomplete
3. `-7d$ to d$`, meaning 6 complete days and the full "current" day, regardless of completeness
4. 7 day duration, but exclude/include some other period: i.e. `7d ending h$`

At this point, I think 1 is the most common meaning. It's the one most platforms seem to choose, including Grafana, Google's metrics explorer and pretty much any stock trading platform. It's also one we don't currently support without making people jump through hoops like `PT168H`, which is only an approximation of what it should be.

However, there are two problems. One is about how you define `now` when you have lag between the wall clock time and your data points (which other platforms don't have) and the other is that people intuitively think `-7d to now` when what they mean is something more like `7d ending {g}$` where {g} is the granularity of the underlying data (which may actually be not definable).

If we had insight into the underlying data granularity, this wouldn't really be a problem. In that case, we could define `now` as the last timestamp rounded forward to the inclusive end of the underlying granularity. So, hourly data with a last data point of `2025-05-20T10:00:00Z` would mean that our `now` anchor would be `2025-05-20T10:59:59Z` (plus whatever milli/nano seconds). That would mean both `h^ to h$` and `-7d to now` would return the correct values.

Following from that, I think we can all agree that something needs to be added to the `watermark` when making a request like `-7d to watermark`. The questions are:

1. What is that value?
2. Does it get added before or after resolving the time range?

My argument is that until we have insight about the underlying data granularity, the least problematic option for the value is one millisecond. Here are the options currently available to us:

1. One millisecond
2. `smallest_time_grain`
3. Grain of syntax string

1. May not be technically correct, but will return the correct data.
2. May not be defined, may not match the underlying data granularity, may return incorrect data
3. Is not what is being asked for and may be ambiguous . `-2d^ to watermark` is not asking for `-2d^ to d$` `-2d3h to watermark` is ambiguous.

That said, I think the actual issue is that we should be adding the millisecond before resolving the time range. If you have daily data and a latest timestamp of `2025-05-20T00:00:00Z`, the correct range from `-7d to watermark` is:

`2025-05-13T00:00:01Z` to `2025-05-20T00:00:01Z`

AND NOT

`2025-05-13T00:00:00Z` to `2025-05-20T00:00:01Z`

The first one would return 7 data points and the second one would return 8. This works with smaller grains as well. If you have hourly data and your latest timestamp is `2025-05-20T10:00:00Z`, the correct request range would be:

`2025-05-13T11:00:00Z` to `2025-05-20T11:00:00Z`

But our approximation using the one millisecond approach is:

`2025-05-13T10:00:01Z` to `2025-05-20T10:00:01Z`

However, with hourly data, they both return the same data points, which is the critical part.

runtime/pkg/rilltime/rilltime_test.go
briangregoryholmes: Can you follow up on this? I assumed (clearly incorrectly) that we were normalizing timestamp columns into a standard format and that as long as the API request had an endpoint ahead of the latest data point, it would be returned.

That's how "all time" requests have always worked. We request the max time range and then add one millisecond to the endpoint.

runtime/pkg/rilltime/rilltime.go
briangregoryholmes: Two uses cases are getting the first 6 hours of the day and getting last n periods, but excluding some other period.

So, `d^ to d^+6h` can be `6h starting d^` instead and `h$-7d to h$` can be `7d ending h$` instead.

runtime/pkg/rilltime/rilltime_test.go
begelundmuller: > 1. `-7d to now`, meaning return all of the data points within the last 7 day period as a duration

Would that be including or excluding `now`? If we keep our inclusive-start-exclusive-end semantics, users who internalize that property might rightfully think it'd be exclusive `now`.

> If we had insight into the underlying data granularity, this wouldn't really be a problem.

It's a tough ask, especially because its easy to mis-infer or misconfigure.

> My argument is that until we have insight about the underlying data granularity, the least problematic option for the value is one millisecond. 

Unfortunately I think it's too problematic. Clickhouse and Pinot often use second precision for timestamps, so the millisecond would get truncated during casting and the result would exclude the latest rows.

I think the smallest contextual grain of the expression is a safer choice, with an option to ceil by another grain (how would I do that?).

> Which means the approximation should be:
> 
> 2025-05-13T00:00:00.001Z to 2025-05-20T00:00:00.001Z

I don't completely follow this – for example, if data is not rolled up already and you're aggregating by a day grain, you'd get missing data in the first bin and last bin?

briangregoryholmes: > Would that be including or excluding now? If we keep our inclusive-start-exclusive-end semantics, users who internalize that property might rightfully think it'd be exclusive now.

To answer this, I should first say that I think we should be using the generic term `now` instead of watermark/latest/etc. And that you can redefine `now` to be either wall clock time or based off the data set.

So, assuming that, `now`, in my view, is always after the timestamp of the latest data point. By definition, you can never have a data point for `now` because it is in progress. So, when I imagine the behavior of `-7d to now`, I expect it to return the latest data point. I understand that this doesn't feel like it has to apply to the term `watermark`, but I think the intended behavior is the same. It's very unlikely that people would ever want to exclude the latest data point when requesting data this way.

> Unfortunately I think it's too problematic. Clickhouse and Pinot often use second precision for timestamps, so the millisecond would get truncated during casting and the result would exclude the latest rows.

I edited one of the points here to suggest that it should instead be the duration of smallest possible grain based on timestamp column or engine. Do you think that would be feasible?

> I think the smallest contextual grain of the expression is a safer choice, with an option to ceil by another grain (how would I do that?).

While `-7d to watermark` and `-7d to d$` return the same data points for the primary period, they are not the same when considering comparisons.

That said, the way you would do this would be `h^-7d to h^` or `7d ending h^` or `-7d/h^ to h^`. That's essentially saying go to the start of the hour, then subtract 7 days or subtract seven days and then go to the start of the hour. That's all well and good, but we need a general solution for the UI where the granularity is not known.

> I don't completely follow this – for example, if data is not rolled up already and you're aggregating by a day grain, you'd get missing data in the first bin and last bin?

Let's assume you have an event based data set with no regular intervals between points. Your latest data point just so happens to fall on `2025-05-20T00:00:00.000Z`, but doesn't represent a fundamental period of time. It's just an event.

If you request, `-7d to now`, what you're asking for is all of the data points that have occurred between `2025-05-13T00:00:00.001Z` and `2025-05-20T00:00:00.001Z`. If you then aggregate those data points by day, you would have buckets for 2025-05-13 and 2025-05-20 that would represent partial days worth of data. That's totally fine and the exact intention of this particular syntax option. We are completely disentangling aggregation from relative range declaration. They are independent. If you are requesting a range with endpoints that fall in the middle of whatever you're aggregating on, that's okay. The fact that we haven't always been doing that is what has caused issues with users finding our handling of time and time series unintuitive.

It makes more sense if you imagine a data point in the middle of the day. If my timestamp is `2025-05-20T11:32:46.245Z` and I ask for `-7d to now`, I am asking for all of the data points between`2025-05-13T11:32:46.246Z` and `2025-05-20T11:32:46.246Z`. Whether I want those aggregated within the time series chart by minute, hour, day, week or maybe not aggregated at all is a secondary question and not relevant to how the range is resolved.

That said, these are not our defaults. If you select our Rill default `Last 7 days` option, you get `-7d$ to d$`, which means the range will fall on day boundaries.

runtime/pkg/rilltime/rilltime.go
```diff
@@ -17,54 +16,65 @@ import (
...
 	isoTimeRegex    = regexp.MustCompile(isoTimePattern)
 	// nolint:govet // This is suggested usage by the docs.
 	rillTimeLexer = lexer.MustSimple([]lexer.SimpleRule{
+		{"Ref", "ref"},
 		{"Earliest", "earliest"},
 		{"Now", "now"},
 		{"Latest", "latest"},
 		{"Watermark", "watermark"},
+		{"Starting", "starting"},
+		{"Ending", "ending"},
```
nishantmonu51: Do we still need these ?  appears unused..
```
{"Starting", "starting"},
{"Ending", "ending"},
{"SnapPrefix", `[<>]`},
{"Interval", `[#]`},
{"Ceil", `[!]`},
```

runtime/pkg/rilltime/rilltime.go
```diff
@@ -85,7 +95,7 @@ var (
 		timeutil.TimeGrainSecond:  timeutil.TimeGrainMinute,
 		timeutil.TimeGrainMinute:  timeutil.TimeGrainHour,
 		timeutil.TimeGrainHour:    timeutil.TimeGrainDay,
-		timeutil.TimeGrainDay:     timeutil.TimeGrainWeek,
+		timeutil.TimeGrainDay:     timeutil.TimeGrainMonth,
```
nishantmonu51, line 90: this seems unintentional ?

runtime/pkg/rilltime/rilltime.go
```diff
@@ -96,75 +106,94 @@ var (
 		timeutil.TimeGrainHour:    timeutil.TimeGrainMinute,
 		timeutil.TimeGrainDay:     timeutil.TimeGrainHour,
 		timeutil.TimeGrainWeek:    timeutil.TimeGrainDay,
-		timeutil.TimeGrainMonth:   timeutil.TimeGrainWeek,
+		timeutil.TimeGrainMonth:   timeutil.TimeGrainDay,
```
nishantmonu51, line 101: unintentional?

runtime/pkg/rilltime/rilltime.go
```diff
@@ -96,75 +106,94 @@ var (
...
-	Grain          *string        `parser:"(By @Grain)?"`
-	AnchorOverride *LabeledAnchor `parser:"('@' @@)?"`
-	TimeZone       *string        `parser:"('@' @TimeZone)?"`
+	Interval        *Interval      `parser:"@@"`
+	AnchorOverrides []*PointInTime `parser:"(As Of @@)*"`
+	Grain           *string        `parser:"(By @Grain)?"`
+	TimeZone        *string        `parser:"('@' @TimeZone)?"`
 
 	isNewFormat bool
 	timeZone    *time.Location
```
nishantmonu51: Expression has two timeZone fields ? Shall we merge them into one or rename to make it clear which one to use when ?

runtime/pkg/rilltime/rilltime.go
```diff
@@ -580,6 +627,15 @@ func (a *AbsoluteTime) parse() error {
...
+			a.nano = val * 1000 * 1000
+			a.tg = timeutil.TimeGrainMillisecond
+		case "micro":
+			a.nano = val * 1000
+			a.tg = timeutil.TimeGrainMillisecond // We dont go below milli
+		case "nano":
+			a.nano = val
+			a.tg = timeutil.TimeGrainMillisecond // We dont go below milli
 		default:
 			return fmt.Errorf("unexpected field %q in duration", name)
```
nishantmonu51: nit: use time instead of duration in error message?

runtime/pkg/rilltime/rilltime.go
AdityaHegde: This is intentional. We never implicitly use week for anything. `higherOrderMap` is used to snap ordinals for example, there when nothing is explicitly specified, we snap to month for day. EG: `D5` is day 5 of month.

runtime/pkg/rilltime/rilltime.go
AdityaHegde: Same here, we dont implicitly use week. `lowerOrderMap` is used to return the default grain when the duration is exactly 1 month. EG: `-2M to -1M` should default to day grain.

runtime/pkg/rilltime/rilltime.go
AdityaHegde: One is the raw timezone, the other is parsed timezone. Lemme rename the other to `tz`. We use `tz` for `*time.Location` across the codebase.

nishantmonu51 (approved):

PR #7300, opened 2025-05-12
Return Creator email in project status

NamanMahor:
Printing creator id

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


admin/server/projects.go
```diff
@@ -85,7 +85,7 @@ func (s *Server) ListProjectsForOrganization(ctx context.Context, req *adminv1.L
 
 	dtos := make([]*adminv1.Project, len(projs))
 	for i, p := range projs {
-		dtos[i] = s.projToDTO(p, org.Name)
+		dtos[i] = s.projToDTO(ctx, p, org.Name)
 	}
```
begelundmuller, line 89: It's not really safe to do this in a loop, can lead to a lot of additional lookups (N+1 query problem)

admin/server/projects.go
```diff
@@ -1828,30 +1828,40 @@ func (s *Server) TriggerRedeploy(ctx context.Context, req *adminv1.TriggerRedepl
...
-func (s *Server) projToDTO(p *database.Project, orgName string) *adminv1.Project {
+func (s *Server) projToDTO(ctx context.Context, p *database.Project, orgName string) *adminv1.Project {
+	var createdByUserEmail string
+	if p.CreatedByUserID != nil {
+		u, err := s.admin.DB.FindUser(ctx, *p.CreatedByUserID)
+		// ignoring err will be empty in case error in getting user
+		if err == nil {
+			createdByUserEmail = u.Email
+		}
+	}
```
begelundmuller, line 1839: It should only return the creator email for admin users, not safe to show to public/reader users

begelundmuller (changes requested):

cli/cmd/project/status.go
```diff
@@ -55,8 +55,9 @@ func StatusCmd(ch *cmdutil.Helper) *cobra.Command {
 			fmt.Printf("  Organization: %v\n", proj.Project.OrgName)
 			fmt.Printf("  Public: %v\n", proj.Project.Public)
 			fmt.Printf("  Github: %v\n", githubURL)
-			fmt.Printf("  Created: %s\n", proj.Project.CreatedOn.AsTime().Local().Format(time.RFC3339))
-			fmt.Printf("  Updated: %s\n", proj.Project.UpdatedOn.AsTime().Local().Format(time.RFC3339))
+			fmt.Printf("  Created by: %s\n", proj.Project.CreatedByUserId)
```
begelundmuller, line 58: This user ID will be meaningless to normal users though? It's only us internally who are able to lookup by user ID, but this command is an end-user command

begelundmuller (changes requested):

PR #7305, opened 2025-05-13
Improve org/project name validation errors

begelundmuller:
**Changes:**
- Allow two-character org names (down from three)
- Allow single character project names (down from three)
- Validate name length separate from the `slug` rule so error messages are clearer
- Rewrite the `slug` error message to:
    - `Validation failed for field '%s': must use only letters, numbers, underscores and dashes`
- Rewrite the length error messages to:
    - `Validation failed for field '%s': must be at least %s characters`
    - `Validation failed for field '%s': must be at most %s characters`

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!


admin/database/database.go
```diff
@@ -373,7 +373,7 @@ type InsertOrganizationOptions struct {
 
 // UpdateOrganizationOptions defines options for updating an existing org
 type UpdateOrganizationOptions struct {
-	Name                                string `validate:"slug"`
+	Name                                string `validate:"min=2,max=40,slug"`
```
pjain1, line 376: hopefully we should not be having any org name greater than 40 characters otherwise during update it will fail

admin/database/database.go
```diff
@@ -446,7 +446,7 @@ type InsertProjectOptions struct {
 
 // UpdateProjectOptions defines options for updating a Project.
 type UpdateProjectOptions struct {
-	Name                 string `validate:"slug"`
+	Name                 string `validate:"min=1,max=40,slug"`
```
pjain1, line 449: same as above

pjain1 (approved):

admin/database/database.go
begelundmuller: The previous `slug` rule already had an upper limit of 40 characters, so that should not be a problem 🙂

PR #7311, opened 2025-05-14
removing lock configuration setting in duckdb

k-anshul:


runtime/pkg/rduckdb/db.go
```diff
@@ -949,13 +949,6 @@ func (d *db) openDBAndAttach(ctx context.Context, uri, ignoreTable string, initQ
 			return nil, err
 		}
 	}
-	if !read {
-		// at the end disable any more configuration changes on the write handle via pre_exec sql
-		_, err = db.ExecContext(ctx, "SET lock_configuration TO true", nil)
```
k-anshul, line 954: 1. Some settings are useful for users to set.
2. Security settings in duckdb like `secret_directory` once set can't be reset.
3. This can anyways be bypassed via `boot_queries` in duckdb.yaml.

begelundmuller (approved):

PR #7319, opened 2025-05-15
Skip organization check in local mode

grahamplata:
Resolves: Support --local in CLI even when not logged in: https://github.com/rilldata/rill-private-issues/issues/1711

Before
```bash
➜  rill git:(gplata/local-flag) ✗ rill project refresh --local --model auction_data_model
Error: no organization is set, pass `--org` or run `rill org switch`
```

After
```
➜  rill git:(gplata/local-flag) ✗ rill project refresh --local --model auction_data_model
Running in local mode, skipping organization checkRefresh initiated. Check the project logs for status updates.
```

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!

### DIFF
--- a/cli/pkg/cmdutil/check.go
+++ b/cli/pkg/cmdutil/check.go
@@ -33,6 +33,11 @@ func CheckAuth(ch *Helper) PreRunCheck {
 
 func CheckOrganization(ch *Helper) PreRunCheck {
 	return func(cmd *cobra.Command, args []string) error {
+		// If the command is run in local mode, skip the check.
+		if cmd.Flags().Lookup("local").Changed {
+			return nil
+		}
+
 		if ch.Org != "" {
 			return nil
 		}


begelundmuller (approved):
LGTM

begelundmuller (changes requested):
Actually, isn't this also needed in the `CheckAuth` middleware? Try `rill logout` and then verifying if it works.

grahamplata:
> > Actually, isn't this also needed in the `CheckAuth` middleware? Try `rill logout` and then verifying if it works.
> 
> Oh let me verify, one moment

I don't believe it is required, tested with prod cli

cli/pkg/cmdutil/check.go
```diff
@@ -20,23 +26,30 @@ func CheckChain(chain ...PreRunCheck) PreRunCheck {
...
 	return func(cmd *cobra.Command, args []string) error {
 		// This will just check if token is present in the config
-		if ch.IsAuthenticated() {
-			return nil
+		if !ch.IsAuthenticated() {
+			return fmt.Errorf("command '%s': %w", cmd.Name(), ErrNotAuthenticated)
 		}
-
-		return fmt.Errorf("not authenticated, please run 'rill login'")
+		return nil
```
begelundmuller, line 36: I guess I still don't understand why this doesn't need to check for the `--local` flag because this is what e.g. the `rill project refresh` command looks like:
```
	refreshCmd := &cobra.Command{
		...
		PersistentPreRunE: cmdutil.CheckChain(cmdutil.CheckAuth(ch), cmdutil.CheckOrganization(ch)),
```

Is something else broken somewhere? Because it looks like `cmdutil.CheckAuth` should bounce here for a user who isn't logged in.

cli/pkg/cmdutil/check.go
```diff
@@ -6,6 +6,12 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var (
+	ErrNotAuthenticated = fmt.Errorf("not authenticated, please run 'rill login'")
+	ErrNoOrganization   = fmt.Errorf("no organization is set, pass `--org` or run `rill org switch`")
+)
```
begelundmuller, line 12: Can we keep the inline errors? Generally we only extract errors to variables if a) they are returned from multiple functions, or b) we need to be able to check them using `errors.Is` somewhere upstream. But neither of those conditions apply here.

grahamplata: Fair enough 👍

PR #7331, opened 2025-05-19
Move embed navigation to url params based navigation

AdityaHegde:
Url params sync assumes that the current url params are for the dashboard. For embed since we do not actually navigate the url params from previous dashboard will be loaded into the new navigated dashboard.

Adding basic url based navigation for embed. This cleans the url and retains just the resource and type.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-admin/src/routes/-/embed/+page.svelte
+++ b/web-admin/src/routes/-/embed/+page.svelte
@@ -31,6 +31,12 @@
       name: resourceName,
       kind: resourceType,
     };
+  } else {
+    // Important! Do not set `activeResource` to `null` here.
+    // In `DashboardStateDataLoader` we are currently clearing the url of non-dashboard params since it will interfere with session storage extraction logic.
+    // So setting `activeResource` to `null` will make the navigation logic in this file to think there is no active resource everytime `DashboardStateDataLoader` cleaned the url.
+    // For going to home we manually set `activeResource` to `null` in `handleGoHome`.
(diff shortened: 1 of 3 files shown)


web-admin/src/routes/-/embed/+page.svelte
```diff
@@ -25,18 +26,24 @@
...
   // Manage active resource
   let activeResource: V1ResourceName | null = null;
-  if (initialResourceName && initialResourceType) {
+  $: if (resourceName && resourceType) {
     activeResource = {
-      name: initialResourceName,
-      kind: initialResourceType,
+      name: resourceName,
+      kind: resourceType,
     };
```
ericpgreen2, line 33: Can we do `else { activeResource = null }`, so we can avoid manually setting `activeResource=null` in `handleGoHome()`?

ericpgreen2: Can you think of ways we can make this more clear? This feels brittle (someone could easily mistakenly make this change when modifying the code in the future).

One thing I'm noticing: the `newUrl` in `handleGoHome()` will be the URL of the last dashboard (just without the search parameters), right? That doesn't seem semantically correct.

ericpgreen2:
It'd be awesome if you wouldn't mind adding a test for embeds with `navigation=true`. Given the unique routing implementation, I've generally been worried about it's stability.

ericpgreen2 (approved):

PR #7337, opened 2025-05-20
fix: home button when home bookmark is not present not applying yaml config.

AdityaHegde:
We use an url with empty params for home button when home bookmark is not present. Since params for yaml config are explicitly added this means we do not apply any settings.

Updating the home button to add url params from yaml config.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-admin/src/features/bookmarks/Bookmarks.svelte
+++ b/web-admin/src/features/bookmarks/Bookmarks.svelte
@@ -26,6 +26,7 @@
   import { useMetricsViewTimeRange } from "@rilldata/web-common/features/dashboards/selectors";
   import { getStateManagers } from "@rilldata/web-common/features/dashboards/state-managers/state-managers";
   import { useExploreState } from "@rilldata/web-common/features/dashboards/stores/dashboard-stores";
+  import { createYAMLConfigExploreUrlSearch } from "@rilldata/web-common/features/dashboards/stores/get-explore-state-from-yaml-config";
   import { ResourceKind } from "@rilldata/web-common/features/entity-management/resource-selectors";
   import { useExploreValidSpec } from "@rilldata/web-common/features/explores/selectors";
   import { eventBus } from "@rilldata/web-common/lib/event-bus/event-bus";
@@ -62,6 +63,11 @@
   $: schemaResp = createQueryServiceMetricsViewSchema(
     instanceId,
     metricsViewName,
+  );
+
+  $: exploreUrlParamsFromYAMLConfig = createYAMLConfigExploreUrlSearch(
+    validExploreSpec,
+    metricsViewTimeRange,
   );
 
   $: categorizedBookmarks = categorizeBookmarks(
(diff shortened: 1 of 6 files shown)


web-admin/src/features/bookmarks/HomeBookmarkButton.svelte
```diff
@@ -19,13 +19,12 @@
   export let project: string;
   export let exploreName: string;
   export let homeBookmark: BookmarkEntry | undefined;
+  export let defaultExploreUrl: string;
```
ericpgreen2, line 22: Because our concept of a "default" Explore is so confusing (which default is it?), we need to be clear about where this state is coming from. How about `urlForExploreYAMLDefaultState`?

If our state system allowed for it, another option would be to pass the `exploreState` object into this component and call `.toURL()`
```
export let exploreStatefromExploreYAMLDefaults: ExploreState

$: homeBookmarkUrl = homeBookmark?.url ?? exploreStatefromExploreYAMLDefaults.toURL();
```

web-common/src/features/dashboards/stores/get-explore-state-from-yaml-config.ts
```diff
@@ -31,6 +37,39 @@ export function getExploreStateFromYAMLConfig(
...
+
+      const urlParams = convertPartialExploreStateToUrlParams(
+        exploreSpec,
+        exploreStateFromYAMLConfig,
+        timeControlState,
+      );
+      return `?${urlParams.toString()}`;
+    },
+  );
+}
```
ericpgreen2, line 71: We have the `exploreStateFromYAMLConfig` in the `DashboardStateDataLoader`. Can we use that, rather than create the same object again?

ericpgreen2 (changes requested):
Are all the `| undefined` edits intended to be in this PR?

web-common/src/features/dashboards/stores/get-explore-state-from-yaml-config.ts
AdityaHegde: Bookmark is outside of dashboard component tree. So we would have to create `DashboardStateDataLoader` and pass it down. This would mean we would have to either refactor `DashboardStateManager` to optionally get it from parent or always get from parent. Both feels unnecessary, since `getExploreStateFromYAMLConfig` is pretty lightweight.

web-admin/src/features/bookmarks/HomeBookmarkButton.svelte
AdityaHegde: `exploreState` is a plain object. If and when we move it a class we can add methods like these.

AdityaHegde:
Yes those are needed to make sure `getTimeControlState` and methods within it are aware that it is passed a partial instead of full explore state.

web-admin/src/features/bookmarks/Bookmarks.svelte
```diff
@@ -64,6 +65,11 @@
     metricsViewName,
   );
 
+  $: urlForExploreYAMLDefaultState = createYAMLConfigExploreUrlSearch(
```
ericpgreen2, line 68: Nit: it'd be nice if the function name more closely matched the name of the resultant object

ericpgreen2 (approved):

PR #7341, opened 2025-05-20
feat (AI): Add YAML properties for user-provided context

ericpgreen2:
This PR adds support for user-provided LLM context via an `ai_context` YAML property in:
- `metrics_view.yaml` files
- The `rill.yaml` file

A future PR will leverage this user-provided context in our MCP server.

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!


ericpgreen2:
> I think I prefer `ai_context`:, but just want to ask if considered dropping "AI" and having the property name be just `context:` or `instructions:` instead?

I also lean towards including the "AI" prefix. We specify in the property description that "this information is used to give LLMs more guidance", and I think this is critical enough information that the property name should also hint at it. "Context" is a generic enough term that, at least today, I don't think everyone exclusively thinks of it as AI-related.

PR #7343, opened 2025-05-20
feat: Generic external duckDB OLAP connector

k-anshul:
Sample project that ingests data from `postgres` to `motherduck` :
```
connectors/duckdb.yaml
---
type: connector
driver: duckdb

path: "md:my_db"

boot_queries: |
  INSTALL 'motherduck';
  LOAD 'motherduck';
  SET motherduck_token= {{ .env.motherduck_token }} 

sources/users.yaml
---
type: source

connector: "postgres"
sql: |
  select * from users
```

**Callout** : **The user needs to make sure that same database is not used for modeling on local vs on cloud otherwise one will overwrite other.**

closes https://github.com/rilldata/rill-private-issues/issues/1753

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!

### DIFF
--- a/runtime/drivers/duckdb/duckdb.go
+++ b/runtime/drivers/duckdb/duckdb.go
@@ -505,6 +505,7 @@ func (c *connection) reopenDB(ctx context.Context) error {
 		maps.Copy(settings, c.config.writeSettings())
 		c.db, err = rduckdb.NewGeneric(ctx, &rduckdb.GenericOptions{
 			Path:               c.config.Path,
+			Attach:             c.config.Attach,
 			DBName:             c.config.DatabaseName,
 			LocalDataDir:       dataDir,
 			LocalCPU:           c.config.CPU,
--- a/runtime/pkg/rduckdb/generic.go
+++ b/runtime/pkg/rduckdb/generic.go
@@ -37,6 +37,7 @@ type GenericOptions struct {
 	// Path to the external DuckDB database.
 	Path string
 	// Attach string allows user to directly pass a DuckDB attach string.
+	//  Example syntax : "'ducklake:metadata.ducklake' AS my_ducklake(DATA_PATH 'datafiles1')"
 	Attach string
 	// DBName is set to the name of the database identified by the Path.
 	DBName string
--- a/runtime/drivers/duckdb/config.go
+++ b/runtime/drivers/duckdb/config.go
@@ -32,6 +32,9 @@ type config struct {
 
 	// Path switches the implementation to use a generic rduckdb implementation backed by the db used in the Path
 	Path string `mapstructure:"path"`
+	// Attach allows user to pass a full ATTACH statement to attach a DuckDB database.
+	// Example YAML syntax : attach: "'ducklake:metadata.ducklake' AS my_ducklake(DATA_PATH 'datafiles1')"
+	Attach string `mapstructure:"attach"`
 	// DatabaseName is the name of the attached DuckDB database specified in the Path.
 	// This is usually not required but can be set if our auto detection of name fails.
 	DatabaseName string `mapstructure:"database_name"`


avaitla:
Will this mean that we don't need DuckDB files locally? We can have sources ingested into MD and models in a MD database? Or is it simply for the serving layer like clickhouse?

k-anshul:
> Will this mean that we don't need DuckDB files locally? We can have sources ingested into MD and models in a MD database? Or is it simply for the serving layer like clickhouse?

Yes we will be able to ingest data directly into motherduck and run models directly in motherduck.
However if you want you can also operate it just as a serving layer where data is ingested externally and just dashboards are built in Rill.

runtime/pkg/rduckdb/motherduck.go
```diff
@@ -0,0 +1,458 @@
...
+		attribute.Bool("view", opts.View),
+	))
+	defer func() {
+		if err != nil {
+			span.SetStatus(codes.Error, err.Error())
+		}
+		span.End()
+	}()
+
+	m.logger.Debug("create: create motherduck table", zap.String("name", name), zap.Bool("view", opts.View), observability.ZapCtx(ctx))
```
begelundmuller, line 193: nit: debug logs should prefix with the file/package name, not function name (to make them easier to find in a log file with debug logs from many different packages). E.g. `motherduck: creating table`.

runtime/pkg/rduckdb/motherduck.go
```diff
@@ -0,0 +1,458 @@
...
+	writeSem *semaphore.Weighted
+
+	opts *MotherDuckDBOptions
+}
+
+type MotherDuckDBOptions struct {
+	// Database is the name of motherduck database to connect to. Accepts both with md: prefix and without.
+	Database string
+	// Token is the token to connect to motherduck.
+	Token string
```
begelundmuller, line 37: Instead of a Motherduck-specific implementation, did you consider a generic DuckDB implementation? It would be nice because it would also enable use of any local DB path or other DuckDB-compatible service. I.e. basically just a basic DuckDB driver without external table storage or something like that.

From a config perspective, I would expect it to work something like this for Motherduck:
```yaml
# duckdb.yaml
type: connector
driver: duckdb

path: md:my_motherduck_db

init_sql: SET motherduck_token = '{{ .env.motherduck_token }}'
-- perhaps support this as syntactic sugar:
-- motherduck_token: {{ .env.motherduck_token }}
```

but could just as well be a local path like:
```yaml
# duckdb.yaml
type: connector
driver: duckdb
path: my_duckdb.db
```

The idea would just be that if `path` is explicitly set, then we skip our managed driver (which has replication and external table storage), and call `rduckdb.NewGeneric` or something like that instead.

runtime/pkg/rduckdb/motherduck.go
```diff
@@ -0,0 +1,458 @@
...
+
+	// LocalPath is the path to the local DuckDB database file.
+	LocalPath string
```
begelundmuller, line 40: Does it need a local path? Can't it connect directly to the Motherduck path? (I guess there are some spill-to-disk temp files to think about, maybe needs to point to a temp dir?)

runtime/pkg/rduckdb/motherduck.go
```diff
@@ -0,0 +1,458 @@
...
+	// LocalCPU cores available for the local instance of DB.
+	LocalCPU int
+	// LocalMemoryLimitGB is the amount of memory available for the local instance of DB.
+	LocalMemoryLimitGB int
+	// Settings are additional query parameters to be passed when creating local DuckDB instance.
+	Settings map[string]string
```
begelundmuller, line 46: I think these could be provided with `path:` as well using query args?

runtime/parser/parse_model.go
```diff
@@ -245,7 +245,7 @@ func (p *Parser) parseModel(ctx context.Context, node *Node) error {
 func (p *Parser) inferSQLRefs(node *Node) ([]ResourceName, error) {
 	// Currently only supports DuckDB.
 	driver, _, _ := p.driverForConnector(node.Connector)
-	if driver != "duckdb" {
+	if driver != "duckdb" && driver != "motherduck" {
```
begelundmuller, line 248: As per the other comment, can we avoid even having a separate `motherduck` driver name?

begelundmuller (changes requested):

runtime/pkg/rduckdb/motherduck.go
k-anshul: The problem with a generic implementation will be that an external duckdb will not be much useful for cloud operations. A user will not be able to push a big duckdb file to github.
We might end up supporting configuration that works on developer but not on cloud.

If we do want to support a generic implementation than I think having a `DSN` is recommended in place of `path` then a user will not require to set any init_sql and can just pass `motherduck_token`. It will also allow a user to pass other settings if required.

```
type: connector
driver: duckdb

dsn:  md:my_motherduck_db?motherduck_token=<>
```

k-anshul: I am up for the generic implementation but just wanted to highlight issues with generic duckdb for cloud

k-anshul: > the extension gets autoloaded when connecting with a path starting `md:`

Unfortunately it only works for CLI and not with go driver. Probably some missing implementation. However I would still prefer to have DSN and pass `motherduck_token` in connection string. 

Setting it separately has issues which are handled right now but will not be possible with a generic implementation.
See comments in current code :
```
// ignore `can only be set during initialization` error
// it is also returned when database is reopened
// looks like the driver or the extension is caching some state
```

k-anshul: Also the connection to `md:my_db` only works if `motherduck_token` is set as env or passed as query param so I would recommend to use `dsn` instead of `path`. The definition would then be :

```
type: connector
driver: duckdb

dsn: md:my_db?motherduck_token=<token>
```

runtime/pkg/rduckdb/motherduck.go
k-anshul: Unfortunately with the generic approach I can't find a way to set where `wal` file is created. I can control where tmp files are created using config `temp_directory` but no such config exists for where `wal` is created. It is most likely to get created at the root in rill cloud.

avaitla:
Would ducklake be an alternative to motherduck? We still would prefer motherduck as it is more mature and user friendly, but I can imagine a hypothetical where one is using ducklake.

runtime/pkg/rduckdb/generic.go
```diff
@@ -0,0 +1,469 @@
...
+	for _, qry := range opts.DBInitQueries {
+		_, err := db.ExecContext(ctx, qry)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// attach the passed external db
+	if opts.Path != "" {
+		// NOTE: This does not wrap with safeSQLString to allow user to pass a path with other properties.
```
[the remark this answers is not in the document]
begelundmuller: I added some thoughts on this in the comments below

runtime/pkg/rduckdb/generic.go
```diff
@@ -0,0 +1,471 @@
...
+func (m *generic) MutateTable(ctx context.Context, name string, initQueries []string, mutateFn func(ctx context.Context, conn *sqlx.Conn) error) (res *TableWriteMetrics, resErr error) {
+	ctx, span := tracer.Start(ctx, "MutateTable", trace.WithAttributes(attribute.String("name", name)))
+	defer func() {
+		if resErr != nil {
+			span.SetStatus(codes.Error, resErr.Error())
+		}
+		span.End()
+	}()
+
+	m.logger.Debug("GenericDuckDB: mutating table", zap.String("name", name), observability.ZapCtx(ctx))
```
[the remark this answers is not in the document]
k-anshul: This was specifically for rill developer but yeah given we have traces viewer now, I agree with you that it is redundant.

runtime/pkg/rduckdb/generic.go
```diff
@@ -0,0 +1,471 @@
...
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// attach the passed external db
+	if opts.Path != "" {
+		// NOTE: This does not wrap with safeSQLString to allow user to pass a path with other properties.
+		// Example: path: "'ducklake:metadata.ducklake' AS my_ducklake (DATA_PATH 'gs://my-bucket/data_files/')"
+		_, err = db.ExecContext(ctx, fmt.Sprintf("ATTACH %s", opts.Path))
```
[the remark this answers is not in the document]
k-anshul: Yeah in past we have started with decomposed properties and then introduced direct SQL expression since we ended up creating a wrapper over SQL.
Also the property that duckdb calls `database-path` is also a SQL expression so I kept it as an expression.

However I am fine with a `path` and an `attach`. Hopefully very few users will ever need `attach`.

runtime/pkg/rduckdb/generic.go
```diff
@@ -0,0 +1,471 @@
...
+	Logger         *zap.Logger
+	OtelAttributes []attribute.KeyValue
+}
+
+func (d *GenericDBOptions) setDefaultSettings() {
+	if d.Settings == nil {
+		d.Settings = make(map[string]string)
+	}
+	if d.LocalMemoryLimitGB > 0 {
+		d.Settings["memory_limit"] = fmt.Sprintf("%d bytes", d.LocalMemoryLimitGB*1000*1000*1000)
```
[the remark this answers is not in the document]
k-anshul: I borrowed it from the existing driver which uses `1000`. I don't know remember why we explicitly did `1000` there. We can reevaluate separately and make changes at both places if that's a mistake.

runtime/pkg/rduckdb/generic.go
```diff
@@ -0,0 +1,471 @@
...
+	}
+	query := dsn.Query()
+	for k, v := range opts.Settings {
+		query.Set(k, v)
+	}
+
+	// Rebuild DuckDB DSN (which should be "path?key=val&...")
+	// this is required since spaces and other special characters are valid in db file path but invalid and hence encoded in URL
+	dsn.RawQuery = ""
+	localFileName := filepath.Join(opts.LocalDataDir, "main"+uuid.NewString()[:8]+".db")
```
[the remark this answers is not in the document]
k-anshul: The problem is that motherduck has some instance caching logic because of which connecting to the same database again fails. Consider the script below:
```
func main() {
	test := func() {
		db, err := sql.Open("duckdb", "duck.db")
		panicIF(err)
		defer db.Close()

		_, err = db.Exec("LOAD motherduck;")
		panicIF(err)

		_, err = db.Exec("SET motherduck_token='<token>'")
		panicIF(err)

		_, err = db.Exec("ATTACH 'md:my_db'")
		panicIF(err)

		_, err = db.Exec("DETACH my_db")
		panicIF(err)
	}

	for i := 0; i < 2; i++ {
		fmt.Printf("iteration: %v\n", i)
		test()
	}
}
```

The second iteration will fail with an error : `token can only be set during initialization`. Even the `attach` also fails.
I posted it on motherduck channel but did not get satisfactory answers. They mostly expect users to directly connect to motherduck and not attach.
My understanding is that `.wal` of local db is not useful since nothing is being written in that database so it should be okay to clear it between restarts. 
However if you think this is problematic we can explore following solutions:
 
1. Explicitly check for `motherduck_token` failure and `attach` failure msg and ignore but it feels weird to have motherduck specific handling in a generic implementation. 
2. Motherduck docs list some options for removing `WAL` so we can explore them especially the `motherduck_dbinstance_inactivity_ttl` but I am not sure if those are fullproof.

runtime/pkg/rduckdb/generic.go
```diff
@@ -0,0 +1,471 @@
...
+	connector, err := duckdb.NewConnector(generateDSN(localFileName, query.Encode()), func(execer driver.ExecerContext) error {
```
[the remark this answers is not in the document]
k-anshul: I can simplify parts of it especially the `dsn, err := url.Parse("")` part by calling `make` directly.

I think we can't directly use `url.String()` and must use the `generateDSN` helper based on the comment above.

avaitla:
Would be good to know how we configure this so that only rill local uses this stuff, but rill cloud can just use its own duckdb as well, so when we move to motherduck it doesn't break our cloud dashboards.

begelundmuller:
> Would be good to know how we configure this so that only rill local uses this stuff, but rill cloud can just use its own duckdb as well, so when we move to motherduck it doesn't break our cloud dashboards.

That'd be possible like this:
```
```

avaitla:
Anything we can help do to get this out?

runtime/pkg/rduckdb/generic.go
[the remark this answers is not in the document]
begelundmuller: What comment is that? Perhaps then pass `url` to `generateDSN` for the final assembly?

runtime/pkg/rduckdb/generic.go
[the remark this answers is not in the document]
begelundmuller: Okay, maybe a little effort to see if we can clear the setting and make it work with a stable DB would be nice. Otherwise it's fine with temp DBs.

Btw, my memory is poor, why is it we can't connect directly? Was it extension installation? Wondering if there's anything we could do to work around that.

begelundmuller:
> Anything we can help do to get this out?

Hey Anil, Anshul is OOO at the moment, but will be back in a couple days. This PR is his top priority so we should be able to get it merged and released next week.

runtime/pkg/rduckdb/generic.go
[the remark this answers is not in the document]
k-anshul: Okay I did more reading on the motherduck setting and I think there will be [performance implication](https://motherduck.com/docs/key-tasks/authenticating-and-connecting-to-motherduck/connecting-to-motherduck/#setting-custom-database-instance-cache-time-ttl) of setting it to 0.
From the docs : `By default, connections to MotherDuck established through the database instance caching supporting DuckDB APIs will reuse the same database instance for 15 minutes after the last connection is closed.`

> Btw, my memory is poor, why is it we can't connect directly? Was it extension installation? Wondering if there's anything we could do to work around that.

1. Since this is generic, ducklake only allows an [attach](https://duckdb.org/2025/05/27/ducklake.html#usage) syntax and not connecting directly.
2. The `wal` file will be created at the root. Though I think in our case the WAL file may not be created at all since we don't write anything in the local db.

I think at this point using a temp db seems to be an okayish solution. If it leads to issues we can explore following solutions:
1. Explicitly check for `token already set` and `motherduck already attached` error and ignore.
2. Create a `newMotherduck` constructor that directly connects to motherduck but returns `generic` struct and uses the generic implementation of the APIs. I am planning to do some followups where I will add some motherduck specific code that will auto install motherduck extension so we can explore this path in the follow up.

runtime/pkg/rduckdb/generic.go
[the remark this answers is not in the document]
k-anshul: > What comment is that?

```
// Rebuild DuckDB DSN (which should be "path?key=val&...")
// this is required since spaces and other special characters are valid in db file path but invalid and hence encoded in URL
```

> Perhaps then pass url to generateDSN for the final assembly?

I made some changes now and it feels much cleaner now but please let me know if you think otherwise.

runtime/pkg/rduckdb/generic.go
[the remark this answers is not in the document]
begelundmuller: Alright, using temp DB sounds alright for now then

runtime/drivers/duckdb/duckdb.go
```diff
@@ -495,6 +499,27 @@ func (c *connection) reopenDB(ctx context.Context) error {
 	connInitQueries = append(connInitQueries, "SET max_expression_depth TO 250")
 
 	// Create new DB
+	if c.config.Path != "" {
+		settings := make(map[string]string)
+		maps.Copy(settings, c.config.readSettings())
+		maps.Copy(settings, c.config.writeSettings())
+		c.db, err = rduckdb.NewGeneric(ctx, &rduckdb.GenericOptions{
+			Path:               c.config.Path,
```
begelundmuller, line 507: `Attach` is not passed here?

runtime/pkg/rduckdb/generic.go
```diff
@@ -0,0 +1,479 @@
...
+	writeSem      *semaphore.Weighted
+
+	opts *GenericOptions
+}
+
+type GenericOptions struct {
+	// Path to the external DuckDB database.
+	Path string
+	// Attach string allows user to directly pass a DuckDB attach string.
+	Attach string
```
begelundmuller: Can you add an example here (and in the main DuckDB config)?

runtime/drivers/duckdb/duckdb.go
k-anshul: Good catch !!!

PR #7344, opened 2025-05-20
Add SQL model validation for full and incremental models

grahamplata:
Add SQL model validation in the form of `assert`. Assertions are SQL queries that SHOULD NOT return any rows.

**Example table** 
`SELECT * FROM range(5)` results in a range column containing `0,1,2,3,4`

**Example model.yaml** 
This example contains a mixture of both `sql` and `assert` syntax

```yaml
# model.yaml
type: model
sql: SELECT * FROM range(5)

tests:
  # Test that all values are in valid range
  - name: Valid Range
    assert: range >= 0 AND range <= 4

  # Test that all values are in valid range (sql)
  - name: Valid Range SQL
    sql: SELECT * FROM model WHERE range < 0 OR range > 4

  # Test row count is exactly 5
  - name: Exact Row Count
    sql: SELECT 'Wrong row count' as error WHERE (SELECT COUNT(*) FROM model) != 5

  # Test no null values exist
  - name: No Nulls
    assert: range IS NOT NULL

  # Test all values are non-negative
  - name: Non-negative Values
    assert: range >= 0

  # Test maximum value doesn't exceed 4
  - name: Max Value Check
    assert: range <= 4

  # Test using BETWEEN syntax
  - name: Range Between
    assert: range BETWEEN 0 AND 4

  # Test that specific values exist
  - name: Value 0 Exists
    sql: SELECT 'Value 0 missing' WHERE (SELECT COUNT(*) FROM model WHERE range = 0) = 0

  - name: Value 4 Exists
    sql: SELECT 'Value 4 missing' WHERE (SELECT COUNT(*) FROM model WHERE range = 4) = 0

  # Test no duplicates (each value appears exactly once)
  - name: No Duplicates
    sql: SELECT range, COUNT(*) as count FROM model GROUP BY range HAVING COUNT(*) > 1

  # Test arithmetic properties
  - name: Sum Check
    sql: SELECT 'Sum should be 10' WHERE (SELECT SUM(range) FROM model) != 10

  - name: Average Check
    sql: SELECT 'Average should be 2' WHERE (SELECT AVG(range) FROM model) != 2.0

  # Test min/max values
  - name: Min Value Check
    sql: SELECT 'Min should be 0' WHERE (SELECT MIN(range) FROM model) != 0

  - name: Max Value Check
    sql: SELECT 'Max should be 4' WHERE (SELECT MAX(range) FROM model) != 4

  # Test even numbers exist
  - name: Even Numbers Present
    assert: range % 2 = 0 OR range % 2 = 1

  # Test specific range exclusions
  - name: No Negative Values
    assert: range >= 0

  - name: No Large Values
    assert: range < 100

  # Test using IN clause
  - name: Valid Values Only
    assert: range IN (0, 1, 2, 3, 4)

  # Test data completeness
  - name: All Expected Values Present
    sql: |
      SELECT missing_value FROM (
        VALUES (0), (1), (2), (3), (4)
      ) AS expected(missing_value)
      WHERE missing_value NOT IN (SELECT range FROM model)

```

**Checklist:**
- [x] Covered by tests
- [x] Linked the issues it closes [PLAT-3](https://linear.app/rilldata/issue/PLAT-3/tests-in-models)
- [ ] Checked if the docs need to be updated - I plan to update these once we collect feedback
- [x] I'm proud of this work!


runtime/parser/parse_model.go
```diff
@@ -17,17 +17,23 @@ import (
...
-	PartitionsConcurrency uint                                    `yaml:"partitions_concurrency"`
-	InputProperties       map[string]any                          `yaml:",inline" mapstructure:",remain"`
+	commonYAML  `yaml:",inline" mapstructure:",squash"` // Only to avoid loading common fields into InputProperties
+	Refresh     *ScheduleYAML                           `yaml:"refresh"`
+	Timeout     string                                  `yaml:"timeout"`
+	Incremental bool                                    `yaml:"incremental"`
+	ChangeMode  string                                  `yaml:"change_mode"`
+	State       *DataYAML                               `yaml:"state"`
+	Partitions  *DataYAML                               `yaml:"partitions"`
+	// TODO: Consider changing to (data_tests, assertions) in the future
```
begelundmuller, line 27: I like `tests:`, I'm good with removing this todo

runtime/parser/parse_model.go
```diff
@@ -17,17 +17,23 @@ import (
...
+	Tests []struct {
+		Name        string `yaml:"name"`
+		Description string `yaml:"description"`
+		DataYAML    `yaml:",inline"`
+	} `yaml:"tests"`
+	Splits                *DataYAML      `yaml:"splits"` // Deprecated: use "partitions" instead
+	PartitionsWatermark   string         `yaml:"partitions_watermark"`
+	PartitionsConcurrency uint           `yaml:"partitions_concurrency"`
```
begelundmuller, line 35: These properties go together with `Partitions`. Maybe move `Tests` below `Stage` or `Output`?

runtime/parser/parse_model.go
```diff
@@ -237,6 +243,25 @@ func (p *Parser) parseModel(ctx context.Context, node *Node) error {
...
 	r.ModelSpec.OutputProperties = outputPropsPB
 
+	if r.ModelSpec != nil && len(tmp.Tests) > 0 {
+		var sqlTests []*runtimev1.ModelTest
+		for i := range tmp.Tests {
+			t := &tmp.Tests[i]
+			resolver, props, refs, err := p.parseDataYAML(&t.DataYAML, outputConnector)
+			if err != nil {
+				return fmt.Errorf(`failed to parse test %q: %w`, t.Name, err)
+			}
```
begelundmuller, line 253: You can't return errors here – look for this comment above:
```
```

So move construction of `var sqlTests` up before that, and then just do assignment here

runtime/parser/parse_model.go
```diff
@@ -237,6 +243,25 @@ func (p *Parser) parseModel(ctx context.Context, node *Node) error {
 	r.ModelSpec.OutputConnector = outputConnector
 	r.ModelSpec.OutputProperties = outputPropsPB
 
+	if r.ModelSpec != nil && len(tmp.Tests) > 0 {
```
begelundmuller, line 246: No need to check `!= nil`, it's guaranteed and if it wasn't, the earlier lines would panic anyway

runtime/parser/parse_model.go
```diff
@@ -237,6 +243,25 @@ func (p *Parser) parseModel(ctx context.Context, node *Node) error {
...
+		var sqlTests []*runtimev1.ModelTest
+		for i := range tmp.Tests {
+			t := &tmp.Tests[i]
+			resolver, props, refs, err := p.parseDataYAML(&t.DataYAML, outputConnector)
+			if err != nil {
+				return fmt.Errorf(`failed to parse test %q: %w`, t.Name, err)
+			}
+			node.Refs = append(node.Refs, refs...)
```
begelundmuller, line 254: This also needs to happen before calling `insertResource`

runtime/parser/parse_model.go
```diff
@@ -17,17 +17,23 @@ import (
...
+	Refresh     *ScheduleYAML                           `yaml:"refresh"`
+	Timeout     string                                  `yaml:"timeout"`
+	Incremental bool                                    `yaml:"incremental"`
+	ChangeMode  string                                  `yaml:"change_mode"`
+	State       *DataYAML                               `yaml:"state"`
+	Partitions  *DataYAML                               `yaml:"partitions"`
+	// TODO: Consider changing to (data_tests, assertions) in the future
+	Tests []struct {
+		Name        string `yaml:"name"`
+		Description string `yaml:"description"`
```
begelundmuller, line 30: Open to input, but keep only `name` and remove `description`? The names can anyway be long if you want, and additional info can be put in YAML comments.

runtime/reconcilers/model.go
```diff
@@ -314,6 +314,13 @@ func (r *ModelReconciler) Reconcile(ctx context.Context, n *runtimev1.ResourceNa
 		} else {
 			model.State.TotalExecutionDurationMs = model.State.LatestExecutionDurationMs
 		}
+
+		// Run model tests if no partitions are configured
```
begelundmuller, line 318: This will also run if partitions are configured (this code path is shared for both partitioned and non-partitioned models, right?)

runtime/reconcilers/model.go
```diff
@@ -1123,6 +1130,14 @@ func (r *ModelReconciler) executePartition(ctx context.Context, catalog drivers.
 		logArgs = append(logArgs, zap.Error(err))
 	}
 
+	// Run partition-level tests after each partition is executed
+	if err == nil && mdl.Spec.PartitionsResolver != "" {
+		testErr := r.runModelTests(ctx, self)
+		if testErr != nil {
+			return nil, false, testErr
+		}
+	}
```
begelundmuller, line 1139: If you want to support partition-level tests, it'd need to pass the `data` variable in here and make it available in templating. Otherwise checks like `where uri = '{{.partition.uri}}'` wouldn't work.

The problem then is that if partitions can be referenced, then it will only work if partition context is provided. I wonder if a separate `partition_tests:` is the only option or if we can think of another way.

runtime/reconcilers/model.go
```diff
@@ -1547,6 +1562,59 @@ func (r *ModelReconciler) shouldTrigger(ctx context.Context, self *runtimev1.Res
...
+				break
+			}
+			if row != nil {
+				r.C.Logger.Warn("Model test failed", zap.String("model", self.Meta.Name.Name), zap.String("test", test.Name), observability.ZapCtx(ctx))
+				errs = append(errs, fmt.Errorf("model test: %s: failed (returned rows)", test.Name))
+				testFailed = true
+				break
+			}
+		}
+		result.Close()
```
begelundmuller, line 1605: It's best to `defer result.Close()` right after `result` is created to avoid leaks in error cases (although there aren't any here). Consider pulling the actual test check into a separate util function so you can use `defer` (not suitable in a loop).

runtime/reconcilers/model.go
```diff
@@ -1547,6 +1562,59 @@ func (r *ModelReconciler) shouldTrigger(ctx context.Context, self *runtimev1.Res
...
+			errs = append(errs, fmt.Errorf("%s: %w", test.Name, err))
+			continue
+		}
+
+		var testFailed bool
+		for {
+			row, err := result.Next()
+			if errors.Is(err, io.EOF) {
+				break
+			}
```
begelundmuller, line 1591: You actually don't need a `for` loop here because just one row is enough to consider the test failed. Compare to how the alert checks work: https://github.com/rilldata/rill/blob/644ee987a7961b84a23328b212a412a43e1f72b1/runtime/reconcilers/alert.go#L579

runtime/reconcilers/model.go
```diff
@@ -1547,6 +1562,59 @@ func (r *ModelReconciler) shouldTrigger(ctx context.Context, self *runtimev1.Res
...
+				testFailed = true
+				break
+			}
+		}
+		result.Close()
+
+		if testFailed {
+			continue
+		}
+	}
```
begelundmuller, line 1610: This doesn't do anything? Was it mean to be a `break`? I think it's should `break`, let's not be too eager about running all tests when we already know some are failing.

runtime/reconcilers/model.go
```diff
@@ -1547,6 +1562,59 @@ func (r *ModelReconciler) shouldTrigger(ctx context.Context, self *runtimev1.Res
...
+// runModelTests executes the user defined tests for the model
+func (r *ModelReconciler) runModelTests(ctx context.Context, self *runtimev1.Resource) error {
+	if len(self.GetModel().Spec.Tests) == 0 {
+		r.C.Logger.Debug("No tests defined for model", zap.String("model", self.Meta.Name.Name), observability.ZapCtx(ctx))
+		return nil
+	}
+	var errs []error
+
+	for _, test := range self.GetModel().Spec.Tests {
+		result, err := r.C.Runtime.Resolve(ctx, &runtime.ResolveOptions{
```
begelundmuller, line 1574: Add `Args: map[string]any{"limit": 1}` as well here, several resolvers support that to optimize to not return too many rows (since just one row already means the test failed). Example here: https://github.com/rilldata/rill/blob/644ee987a7961b84a23328b212a412a43e1f72b1/runtime/reconcilers/alert.go#L570

begelundmuller (changes requested):

runtime/parser/parse_model.go
grahamplata: Fair enough, I can drop description

avaitla:
If a test fails, the resulting table view is empty, this is a bit painful if you want to have a TDD approach, where you first write a sql model and then start to add tests onto it and when a test fails you can iteratively fix your model until the test passes, however when the data goes blank because the test failed you can't iterate effectively (as you can no longer see whats going on)

Screenshot 2025-05-23 at 2 34 16 PM

Ideally the tests could have a warning at the bottom, but still allow the model to exist.

runtime/reconcilers/model.go
avaitla: In my opinion, it would be nice to see all tests that fail, not just a random one, in case they are related and help give insight into what went wrong with the model since often one test could correlate with another signal to ultimately fix the model.

avaitla:
Longer term would be nice if the SQL test can also display what the failure was (the returned rows can include concatenated strings on what broke to help us debug further).

runtime/reconcilers/model.go
```diff
@@ -314,6 +314,12 @@ func (r *ModelReconciler) Reconcile(ctx context.Context, n *runtimev1.ResourceNa
 		} else {
 			model.State.TotalExecutionDurationMs = model.State.LatestExecutionDurationMs
 		}
+
+		testErr := r.runModelTests(ctx, self)
+		if testErr != nil {
+			execErr = testErr
+		}
```
begelundmuller, line 321: I think executing tests like this may have some problems:
1. The execution has already completed, so blocking execution doesn't prevent the result from showing up.
2. Even in Rill Developer (where `modelEnv.StageChanges` is false), you probably want to keep the ingested model data so you can iterate on test definitions without re-running the model each time?
3. Changing the test definition currently doesn't trigger a condition in the `specHash`, so it won't automatically run the updated tests. (And as per 2., you probably don't want it to match the `specHash` since that would mean changing a test would lead to a full refresh of the data.)

One option to remedy this is to separate out test execution with a `testSpecHash` and caching test errors in the `model.State`. However that would perhaps not align with partition-level tests...

runtime/reconcilers/model.go
```diff
@@ -1547,6 +1561,75 @@ func (r *ModelReconciler) shouldTrigger(ctx context.Context, self *runtimev1.Res
 	}
 }
 
+// execModelTest runs a single model test and returns an error if it fails. Accepts partitionData for partition-level tests.
+func (r *ModelReconciler) execModelTest(ctx context.Context, self *runtimev1.Resource, test *runtimev1.ModelTest, partitionData map[string]any) error {
+	args := map[string]any{"limit": 1}
+	if partitionData != nil {
+		args["partition"] = partitionData
+		args["split"] = partitionData // Deprecated: use "partition" instead
```
begelundmuller, line 1569: Can remove this line. No need to keep backwards compatibility here because tests are a new feature.

avaitla:
Anything left to do here?

avaitla:
Wanted to bump here @begelundmuller

begelundmuller:
> Wanted to bump here @begelundmuller

@avaitla We've run into some tricky edge cases around how tests interact with model execution that are taking a while to get right. We're hoping to get this shipped soon!

avaitla:
I like the mix of sql and assert, it makes the simple cases easier but still gives the power of flexible queries. Is there a simple for assert saying "user_id" is unique, or ("user_id", "date") is unique composite, thats another common case.

Also curious when a few tests failed where they will appear and what the message says. Will the dashboard viewer see it somewhere to know the dashboard they're looking at has errors in the underlying model.

grahamplata:
> I like the mix of sql and assert, it makes the simple cases easier but still gives the power of flexible queries. Is there a simple for assert saying "user_id" is unique, or ("user_id", "date") is unique composite, thats another common case.
> 
> Also curious when a few tests failed where they will appear and what the message says. Will the dashboard viewer see it somewhere to know the dashboard they're looking at has errors in the underlying model.

Initially it will be surfaced through the existing "error handling popover" that appears in the lower third of the table results. I will be setting up some time with the frontend folks. Happy to capture your thoughts

avaitla:
If an error is shown for tests, will the end dashboard still be up for end users? Ideally I imagine it could be.

Would this be the best way to encode a unique column test?

    SELECT 1 FROM table GROUP BY uniq_key HAVING COUNT(*) > 1

This is a good reference for some similar tests: https://cloud.google.com/bigquery/docs/reference/standard-sql/debugging-statements

grahamplata:
> If an error is shown for tests, will the end dashboard still be up for end users? Ideally I imagine it could be.
> 
> Would this be the best way to encode a unique column test?
> 
> ```
> SELECT 1 FROM table GROUP BY uniq_key HAVING COUNT(*) > 1
> ```
> 
> This is a good reference for some similar tests: https://cloud.google.com/bigquery/docs/reference/standard-sql/debugging-statements

The model test errors will not prevent dashboards from being available to end users. If model tests fail, the reconciler returns  an error, but it does not delete or roll back the model output.

PR #7347, opened 2025-05-20
Refactor object store utilities

begelundmuller:
**Changes:**
- Avoid partial batches when using `partition_overwrite` in DuckDB
- Prepare for changes to `glob:` resolver
- Remove legacy code for object stores that predates good native support in DuckDB
- Simplify the object store interfaces

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


runtime/drivers/bigquery/warehouse.go
```diff
@@ -38,6 +42,9 @@ var _ drivers.Warehouse = &Connection{}
 
 // QueryAsFiles implements drivers.SQLStore
 func (c *Connection) QueryAsFiles(ctx context.Context, props map[string]any) (drivers.FileIterator, error) {
+	ctx, span := tracer.Start(ctx, "Connection.QueryAsFiles")
+	defer span.End()
```
k-anshul, line 46: Same as `runtime/drivers/athena/warehouse.go`

runtime/drivers/blob/blobdownloader.go
```diff
@@ -3,166 +3,112 @@ package blob
...
+// The downloaded objects will be stored in a temporary directory with the same file hierarchy the bucket, enabling parsing of hive partitioning on the downloaded files.
 // The client should call Close() once done to release all resources.
 // Calling Close() on the iterator will also close the bucket.
-func NewIterator(ctx context.Context, bucket *blob.Bucket, opts Options, l *zap.Logger) (drivers.FileIterator, error) {
-	opts.validate()
-
+func (b *Bucket) Download(ctx context.Context, opts *DownloadOptions) (drivers.FileIterator, error) {
 	tempDir, err := os.MkdirTemp(opts.TempDir, "blob_ingestion")
 	if err != nil {
+		if opts.CloseBucket {
```
k-anshul, line 58: This is usually very easy to miss when adding new if conditions. I would suggest to run a defer at top and check for `error` returned from function. Sample here : https://github.com/rilldata/rill/blob/ac3caa7c912520c87668275adaf73fcea4421b4c/runtime/drivers/snowflake/warehouse.go#L56-L64

runtime/drivers/blob/blobdownloader.go
```diff
@@ -3,166 +3,112 @@ package blob
...
+type DownloadOptions struct {
+	// Glob is the pattern to match files in the bucket
+	Glob string
+	// Format is the format of the files (e.g. "csv", "json", "parquet")
 	Format string
 	// TempDir where temporary files should be stored
 	TempDir string
+	// KeepFilesUntilClose if true, files will not be deleted until Close() is called.
+	KeepFilesUntilClose bool
+	// BatchSizeBytes is the size of the batch to download before sending it to the client.
```
k-anshul, line 45: Not necessarily to be done in this PR but we don't seem to be using batching now. Given we also want users to use partitions we should probably deprecate all batching code.

k-anshul (approved):

PR #7349, opened 2025-05-21
allow default value for dict get expr

pjain1:
lookup can return NULL and cause issues in places like search where (I)LIKE operator with CH where we cast it to TEXT. Using default value in this case can be helpful, this can be helpful in other cases as well. 

Specifically solving the cast to TEXT issue can be solved just with coalescing the expression as user might still want to keep NULL value for toplist etc. However it can be tricky to wrap expression. 

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


proto/rill/runtime/v1/resources.proto
```diff
@@ -187,6 +187,7 @@ message MetricsViewSpec {
     string lookup_table = 8;
     string lookup_key_column = 9;
     string lookup_value_column = 10;
+    optional string lookup_default_value = 11; // optional because need to distinguish between not set and empty value
```
begelundmuller, line 190: Can it be an expression instead, like `lookup_default_expression`? Then it could work for non-string values and wouldn't need `optional`

begelundmuller (changes requested):

PR #7350, opened 2025-05-21
feat: Add back archive based deploy for UI deploy flow testing

AdityaHegde:
Managed github based deploy for E2E could cause issues with rate limiting. So until we figure that out we need the old archive based deploy. This code is already there for cli deploys, adding it back for UI based deploy endpoints. 

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


proto/rill/local/v1/api.proto
```diff
@@ -102,6 +104,8 @@ message DeployProjectResponse {
 message RedeployProjectRequest {
   string project_id = 1;
   bool reupload = 2;
+  // temporarily used for testing
+  bool rearchive = 5;
```
k-anshul, line 108: Nit: field 3 ?

k-anshul (approved):

proto/rill/local/v1/api.proto
AdityaHegde: Ah yes. Copy paste error, updating.

PR #7354, opened 2025-05-22
fix: switch to created org via local API

k-anshul:
**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/cli/pkg/local/server.go
+++ b/cli/pkg/local/server.go
@@ -309,6 +309,15 @@ func (s *Server) DeployProject(ctx context.Context, r *connect.Request[localv1.D
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


cli/pkg/local/server.go
```diff
@@ -309,6 +309,15 @@ func (s *Server) DeployProject(ctx context.Context, r *connect.Request[localv1.D
 		}
 	}
 
+	if s.app.ch.Org != r.Msg.Org {
+		// Switching to passed org
+		s.app.ch.Org = r.Msg.Org
```
AdityaHegde, line 314: Should this be a method on `cmdutil.Helper`? That way we can update any state around org in a single place.

PR #7357, opened 2025-05-22
Serve MCP from the runtime

begelundmuller:
Implements an MCP server in the Rill runtime, enabling MCP clients to connect directly to Rill without starting a separate process.

**Features:**
- Supports MCP in Rill Developer on `http://localhost:9009/mcp/sse`
- Supports MCP for Rill Cloud projects on `https://api.rilldata.com/v1/organizations/{org}/projects/{proj}/runtime/mcp/sse`
- Supports public/unauthenticated access to public Rill Cloud projects
- Supports authenticated access with [user tokens](https://github.com/rilldata/rill/pull/7359) to private Rill Cloud projects

**Examples:**

Claude config for connecting to Rill Developer (must have `rill start` running in a terminal):
```json
{
    "mcpServers": {
        "rill": {
            "command": "npx",
            "args": [
                "mcp-remote",
                "http://localhost:9009/mcp/sse"
            ]
        }
    }
}
```

Claude config for connecting to a public Rill Cloud project:
```json
{
    "mcpServers": {
        "rill": {
            "command": "npx",
            "args": [
                "mcp-remote",
                "https://api.rilldata.com/v1/organizations/demo/projects/rill-github-analytics/runtime/mcp/sse"
            ]
        }
    }
}
```

Claude config for connecting to a private Rill Cloud project (use `rill token issue` to get a personal access token):
```json
{
    "mcpServers": {
        "rill": {
            "command": "npx",
            "args": [
                "mcp-remote",
                "https://api.rilldata.com/v1/organizations/demo/projects/rill-openrtb-prog-ads/runtime/mcp/sse",
                "--header",
                "Authorization:${AUTH_HEADER}"
            ],
            "env": {
                "AUTH_HEADER": "Bearer <Rill access token>"
            }
        }
    }
}
```

**Future work:**
- Implement a unified, multi-project MCP server for Rill Cloud (e.g. `api.rilldata.com/v1/mcp`)
- Implement the MCP OAuth flows so you don't need to manually create a personal access token
- QA, prompt tuning, more endpoints (user management, etc.)

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!

### DIFF
--- a/runtime/server/mcp.go
+++ b/runtime/server/mcp.go
@@ -220,6 +220,7 @@ func (s *Server) mcpQueryMetricsView() (mcp.Tool, server.ToolHandlerFunc) {
 	description := `
 Perform an arbitrary aggregation on a metrics view.
 Tip: Use the 'sort' and 'limit' parameters for best results and to avoid large, unbounded result sets.
+Important note: The 'time_range' parameter is inclusive of the start time and exclusive of the end time.
 
 Example: Get the total revenue by country and product category for 2024:
     {
--- a/runtime/metricsview/query.go
+++ b/runtime/metricsview/query.go
@@ -389,7 +389,7 @@ const QueryJSONSchema = `
     },
     "time_range": {
       "$ref": "#/$defs/TimeRange",
-      "description": "Time range filter for the query. Note that for large datasets, querying shorter and/or more recent time ranges has significant performance benefits."
+      "description": "Time range filter for the query. Time ranges are inclusive of start time and exclusive of end time. Note that for large datasets, querying shorter and/or more recent time ranges has significant performance benefits."
     },
     "comparison_time_range": {
       "$ref": "#/$defs/TimeRange",


begelundmuller:
> Any reason for using SSE vs Streamable HTTP? Streamable HTTP would allow multiple clients to connect, but I haven't had the same luck with SSE, maybe I'm doing something incorrectly.

The `mcp-go` library that we're using only just merged Streamable HTTP [two days ago](https://github.com/mark3labs/mcp-go/pull/273). Unfortunately I wasn't able to get it working with Claude Desktop (even the example project didn't work). So we're going with the currently released SSE implementation for now, but definitely want to switch to the Streamable HTTP implementation when it gets a bit more stable.

avaitla:
Gotcha, do you think this will be available on rill cloud or just for rill developer? If on rill cloud would be curious if this MCP will support multiple clients (claude desktop from different users) simultaneously?

begelundmuller:
> Gotcha, do you think this will be available on rill cloud or just for rill developer? If on rill cloud would be curious if this MCP will support multiple clients (claude desktop from different users) simultaneously?

Yeah this will also be availability in Rill Cloud (see the third example Claude config in the PR note). And yeah it will support concurrent users/clients! Each user will authenticate with their own access token (initially you'll have to create it manually with [this new CLI command](https://github.com/rilldata/rill/pull/7359), but we also plan on adding support for the OAuth -based redirects for the local client to get an access token).

runtime/pkg/middleware/timeout.go
```diff
@@ -48,3 +51,25 @@ func TimeoutUnaryServerInterceptor(fn func(method string) time.Duration) grpc.Un
...
+			duration := fn(req.Params.Name)
+			if duration == 0 {
+				return next(ctx, req)
+			}
+
+			ctx, cancel := context.WithTimeout(ctx, duration)
+			defer cancel()
+
+			res, err := next(ctx, req)
+			if err != nil && errors.Is(ctx.Err(), context.DeadlineExceeded) {
```
k-anshul, line 67: Just wondering if it will not be handled in `mcp/mcpErrorFromGRPC`

runtime/server/mcp.go
```diff
@@ -0,0 +1,380 @@
...
+		),
+	)
+
+	handler := func(ctx context.Context, req mcp.CallToolRequest) (*mcp.CallToolResult, error) {
+		name, err := req.RequireString("metrics_view")
+		if err != nil {
+			return mcpNewToolError(err)
+		}
+
+		resp, err := s.MetricsViewTimeRange(ctx, &runtimev1.MetricsViewTimeRangeRequest{
```
k-anshul, line 205: Not necessary to handle now but why not directly call the resolver ?

k-anshul (approved):

nishantmonu51 (approved):
👍

runtime/pkg/middleware/timeout.go
begelundmuller: It won't because this happens in the middleware stack, whereas the error mapping currently happens in the individual handlers. But it's a good idea, I've refactored it to handle the error mapping in a middleware instead.

runtime/server/mcp.go
begelundmuller: Done!

avaitla:
Do you think the mcp will ever support the ability to generate a rill url the end user can go to as partial evidence of the answer? My concern is it will spit out some numbers but it's really hard to say whether that is right or wrong, being able to corroborate it with a rill dashboard helps check the answer is correct but also continue to answer in the rill ui for the human operator.

PR #7364, opened 2025-05-24
Git sync APIs

k-anshul:
INSERT DESCRIPTION HERE

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


cli/pkg/gitutil/gitcmdwrapper.go
```diff
@@ -0,0 +1,153 @@
+// gitutil package provides utility functions for working with git repositories.
+// To execute git operations we use the go-git library.
+// However the library does not support all git operations and in those cases we directly run git commands.
+// The utility functions in this file directly run git commands.
+package gitutil
```
begelundmuller, line 5: It'd be great with some tests in this package. Maybe there's a `testcontainer` that launches a Git server?

cli/pkg/local/git.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/rilldata/rill/cli/pkg/gitutil"
+	localv1 "github.com/rilldata/rill/proto/gen/rill/local/v1"
+	"go.uber.org/zap"
+)
+
+func (s *Server) WatchGitStatus(ctx context.Context, r *connect.Request[localv1.WatchGitStatusRequest], stream *connect.ServerStream[localv1.WatchGitStatusResponse]) error {
+	// Get authenticated admin client
+	if !s.app.ch.IsAuthenticated() {
+		return errors.New("must authenticate before performing this action")
+	}
```
begelundmuller, line 17: Why? I think the expectation is that we can still check status, push and pull if you run `rill start` in a self-managed Git.

cli/pkg/local/git.go
```diff
@@ -0,0 +1,101 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = gitutil.GitFetch(ctx, s.app.ProjectPath, remote)
+	if err != nil {
+		return err
+	}
+
+	return gitutil.PollGitStatus(ctx, s.app.ProjectPath, remote, func(gs gitutil.GitStatus) {
```
begelundmuller, line 38: I'm not sure I like the streaming API, but if necessary, then can we move all polling to the API handler itself? Since the logic here doesn't seem stateful, it doesn't seem necessary to push the polling logic into nested libraries.

begelundmuller (changes requested):

cli/pkg/gitutil/gitcmdwrapper.go
```diff
@@ -0,0 +1,153 @@
...
+	if err != nil {
+		return err
+	}
+	args = append(args, u)
+
+	st, err := RunGitStatus(path)
+	if err != nil {
+		return err
+	}
+	// TODO : handle detached state
```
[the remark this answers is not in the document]
k-anshul: Ideally not. Unless the user makes some git operations externally. I will throw an error.

cli/pkg/gitutil/gitutil.go
```diff
@@ -312,3 +330,45 @@ func Clone(ctx context.Context, path string, c *Config) (*git.Repository, error)
...
+	for {
+		select {
+		case <-time.After(time.Minute):
+			err = fetchAndCheckStatus()
+			if err != nil {
+				return err
+			}
+		case <-ctx.Done():
+			return ctx.Err()
+		}
```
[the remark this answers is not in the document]
k-anshul: This is not required based on other comments.

cli/pkg/gitutil/gitcmdwrapper.go
k-anshul: Yeah I added an issue to handle this later : https://github.com/rilldata/rill-private-issues/issues/1825

cli/pkg/local/git.go
begelundmuller: What if it is a user-managed Git repository and the user has their Git credentials configured locally (with normal Git) such that they can push/pull?

cli/pkg/local/git.go
```diff
@@ -0,0 +1,99 @@
...
+	"connectrpc.com/connect"
+	"github.com/rilldata/rill/cli/pkg/gitutil"
+	localv1 "github.com/rilldata/rill/proto/gen/rill/local/v1"
+)
+
+func (s *Server) GitStatus(ctx context.Context, r *connect.Request[localv1.GitStatusRequest]) (*connect.Response[localv1.GitStatusResponse], error) {
+	// Get authenticated admin client
+	if !s.app.ch.IsAuthenticated() {
+		return nil, errors.New("must authenticate before performing this action")
+	}
```
begelundmuller: Since it's a status API, even if the user is not logged in (or there is no configured Rill Cloud project), might it still be relevant to report a) if there is a Git remote, b) if the user can push/pull to it?

cli/pkg/local/git.go
k-anshul: Maybe we need some product inputs here but I think it will be much simpler to just ask user to do login and then do `push`/`pull`/`status`. 
If a user want to use their native credentials then they can anyways use git commands.

Also the `rill project clone` command does not use user's credentials even for cloning a self managed git projects so it is possible that a user does not have git configured/credentials set up/or access not available but they will still be able to clone the project if they are project admins.

k-anshul: I think it will make it difficult to think and reason about the implementation if we start mixing native credentials. Even from user's POV it will be easier to explain them that if you want to push/pull/status using native credentials use git commands. If you want to do it within Rill you must login.

Some questions that will need to answered :
1. If the git credentials are not found for a self managed repo, should we ask user to setup git credentials or login ?
2. Should we mandatorily use `rill project clone` with native credentials for self managed repos ?

k-anshul: cc : @ericokuma @AdityaHegde for more inputs.

cli/pkg/local/git.go
k-anshul: It will depend on how UI integrates these APIs but I think these APIs have a prerequisite of the repo being connected to a Project then 

a) will no longer be required.
b) It will depend on whether we use native credentials or not and whether we allow user to clone a project without native credentials set up. If we use permissions setup in Rill then push/pull check is just the `ProjectPermissions` check from `GetProject`.

cli/pkg/local/git.go
begelundmuller: > it will be easier to explain them that if you want to push/pull/status using native credentials use git commands

I agree and I don't think we should automate credentials for user-managed repositories. At least not in the first 
iteration. If they cloned the repository themselves, they probably have credentials for it 😄 

My comment here is just to ensure that the APIs also support this workflow, i.e.:
1. Allow the UI to get the Git status to find out if it's a self-managed Git repository, even when the user is not logged into Rill.
2. Allow the UI to trigger a commit/push for a self-managed Git repository (with the user's self-managed credentials). I'm pretty sure there's a requirement to support this.

cli/pkg/local/git.go
begelundmuller: Let me know if I answered this in my comment in the other thread

cli/pkg/local/git.go
```diff
@@ -0,0 +1,150 @@
...
+		RemoteCommits: gs.RemoteCommits,
+	}), nil
+}
+
+func (s *Server) GitPull(ctx context.Context, r *connect.Request[localv1.GitPullRequest]) (*connect.Response[localv1.GitPullResponse], error) {
+	_, err := gitutil.GitPull(ctx, s.app.ProjectPath, r.Msg.DiscardLocal, "")
+	if err == nil {
+		return connect.NewResponse(&localv1.GitPullResponse{}), nil
+	}
+	// if native git pull fails, try with ephemeral token - this may be a managed git project
```
AdityaHegde, line 83: If we want to support unauthenticated pull we would need to catch conflict error and return to UI no?

This is the error I get from native pull with conflicts,
```
git pull failed: error: Your local changes to the following files would be overwritten by merge:
	dashboards/AdBids_metrics_explore.yaml
Please commit your changes or stash them before you merge.
Aborting
```

cli/pkg/local/git.go
```diff
@@ -0,0 +1,150 @@
...
+	err = gitutil.CommitAndForcePush(ctx, s.app.ProjectPath, config, r.Msg.CommitMessage, author, true)
+	if err != nil {
+		return nil, err
+	}
+
+	return connect.NewResponse(&localv1.GitPushResponse{}), nil
+}
+
+func (s *Server) gitRemoteForProject(ctx context.Context, project *adminv1.Project) (string, error) {
+	config, err := s.app.ch.GitHelper(project.OrgName, project.Name, s.app.ProjectPath).GitConfig(ctx)
```
AdityaHegde, line 162: If a user is logged in and the github repo is not deployed then it panics here. (project is nil?)

k-anshul: Oh okay. I will handle the case. I thought these APIs will only be hit when it is connected to a project.

cli/pkg/local/git.go
k-anshul: Oh okay. I missed setting the error in `Pull` response. Let me set the error in `GitPullResponse`

k-anshul: Now setting the following msg in response:
```
git pull failed: From https://github.com/managed-rill-projects/test-anshul1-git-test-8f3e9d1a\n * branch            main       -> FETCH_HEAD\nerror: Your local changes to the following files would be overwritten by merge:\n\tREADME.md\nPlease commit your changes or stash them before you merge.\nAborting\n
```

begelundmuller (approved):

PR #7366, opened 2025-05-26
support alternate time dimensions

pjain1:
Fixes https://github.com/rilldata/rill-private-issues/issues/1641

In addition to primary timeseries, another dimension of type `timestamp` can be specified in the list of dimensions. This dimension can be sent as request argument to aggregation, timerange and timeseries APIs.  Dimension list in the `ValidSpec` of the metrics view is populated with types and application can use it show alternate time dimension columns.

Currently limitations
- Expressions are not supported
- Column should be of timestamp type
- Column used in `timeseries` of the mv cannot be defined in the dimension list

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


proto/rill/runtime/v1/queries.proto
```diff
@@ -426,6 +426,8 @@ message MetricsViewAggregationRequest {
   bool fill_missing = 21;
   // Optional. Defaults to false. Used to fetch rows from underlying model
   bool rows = 22;
+  // Optional. If not specified, falls back to the primary time dimension in the metrics view spec
+  string time_column = 23;
```
begelundmuller: I assume it's the dimension name, not the underlying column name? So then maybe call it `time_dimension`?

runtime/metricsview/ast.go
```diff
@@ -121,7 +122,7 @@ const (
 // This is due to NewAST not being able (or intended) to resolve external time anchors such as watermarks.
 //
 // The qry's PivotOn must be empty. Pivot queries must be rewritten/handled upstream of NewAST.
-func NewAST(mv *runtimev1.MetricsViewSpec, sec *runtime.ResolvedSecurity, qry *Query, dialect drivers.Dialect) (*AST, error) {
```
begelundmuller: Why pass it separately here instead of accessing `qry.TimeDimension`?

runtime/metricsview/ast.go
```diff
@@ -160,6 +167,22 @@ func NewAST(mv *runtimev1.MetricsViewSpec, sec *runtime.ResolvedSecurity, qry *Q
...
+		// This is necessary to ensure that the time dimension can be used in the query.
+		t, err := ast.lookupDimension(timeDim, true)
+		if err != nil {
+			return nil, fmt.Errorf("time dimension %q not found: %w", timeDim, err)
+		}
+
+		ast.timeField = &FieldNode{
+			Name:        t.Column, // use actual column name instead of dim name as the time dim may not be present in the select clause, so using alias won't work
+			DisplayName: t.DisplayName,
+			Expr:        t.Expression,
```
begelundmuller: 1. Does this mean the result will use the underlying column name instead of the dimension name? That would not be good. 
2. Shouldn't it populate `Expr:` by calling `dialect.MetricsViewDimensionExpression`? So it becomes the escaped version of `t.Column` if it is set? Like [here](https://github.com/rilldata/rill/blob/d24cf0179d58b96af81ac256c282751bb925fe98/runtime/metricsview/ast.go#L181)

runtime/metricsview/ast.go
```diff
@@ -1182,18 +1211,21 @@ func (a *AST) generateIdentifier() string {
 }
 
 // sqlForTimeRange builds a SQL expression and query args for filtering by a time range.
-func (a *AST) sqlForTimeRange(timeCol string, start, end time.Time) (string, []any) {
+func (a *AST) sqlForTimeRange(timeCol *FieldNode, start, end time.Time) (string, []any) {
+	col := a.dialect.EscapeIdentifier(timeCol.Name)
+	if timeCol.Expr != "" {
+		col = timeCol.Expr
+	}
```
begelundmuller: If you use `dialect.MetricsViewDimensionExpression`, then I think this would just be `timeCol.Expr`

runtime/metricsview/ast.go
```diff
@@ -1182,18 +1211,21 @@ func (a *AST) generateIdentifier() string {
 }
 
 // sqlForTimeRange builds a SQL expression and query args for filtering by a time range.
-func (a *AST) sqlForTimeRange(timeCol string, start, end time.Time) (string, []any) {
+func (a *AST) sqlForTimeRange(timeCol *FieldNode, start, end time.Time) (string, []any) {
```
begelundmuller: nit: call it `timeDimension` or `timeField`, not `timeCol` (since it may not be a column)

runtime/metricsview/executor.go
```diff
@@ -30,6 +30,7 @@ type Executor struct {
 	streaming   bool
 	security    *runtime.ResolvedSecurity
 	priority    int
+	timeColumn  string // it's needed here mainly to resolve Timestamps
```
begelundmuller: Any way we can avoid having it here? Feels like it breaks some abstractions to require query-specific fields in the executor.

Could it be solved by changing the type of `timestamps` to `map[string]TimestampsResult`, so it can cache timestamps for different time dimensions if necessary?

runtime/metricsview/executor_timestamps.go
```diff
@@ -19,7 +19,8 @@ func (e *Executor) resolveDuckDBClickHouseAndPinot(ctx context.Context) (Timesta
 	if filter != "" {
 		filter = fmt.Sprintf(" WHERE %s", filter)
 	}
-	timeDim := e.olap.Dialect().EscapeIdentifier(e.metricsView.TimeDimension)
+
+	timeDim := e.timeColumnOrExpr()
```
begelundmuller: Could it take the desired time dimension as input in `e.Timestamps(ctx, timeDim)` instead and pass it through? And allow the empty string `""` to default to the default time dimension?

runtime/metricsview/executor_timestamps.go
```diff
@@ -217,6 +219,24 @@ func (e *Executor) resolveDruid(ctx context.Context) (TimestampsResult, error) {
...
+	for _, dim := range e.metricsView.Dimensions {
+		if dim.Name == timeDim {
+			if dim.Expression != "" {
+				return dim.Expression
+			}
+			return e.olap.Dialect().EscapeIdentifier(dim.Column)
+		}
+	}
+	return e.olap.Dialect().EscapeIdentifier(timeDim) // fallback to the time dimension if not found in dimensions
+}
```
begelundmuller: Can it use `dialect.MetricsViewDimensionExpression`?

runtime/metricsview/executor_validate.go
```diff
@@ -138,9 +128,22 @@ func (e *Executor) ValidateMetricsView(ctx context.Context) (*ValidateMetricsVie
...
+				if typ, ok := mvSchema[d.Name]; ok {
+					d.Type = typ
+				} else {
+					res.DimensionErrs = append(res.DimensionErrs, IndexErr{
+						Idx: -1, // -1 indicates no specific index for dimensions without a column
+						Err: fmt.Errorf("dimension %q not found in schema", d.Name),
+					})
+				}
+			}
+			e.validateTimeTypes(cols, mvSchema, res)
```
begelundmuller: It feels unexpected that it rewrites the metrics view in place. It's also not safe to do that unless it was cloned when being read from the catalog.

I wonder if the code flows could be simplified with something like `mv, err := e.NormalizeMetricsView(ctx context.Context)` that clones the metrics view and adds schema info (and in the future, maybe it does other rewrites).

Then if that happens before validation, the validation can just assume that `dim.Type` has been populated instead of computing the schema again.

What do you think?

runtime/metricsview/executor_validate.go
```diff
@@ -272,6 +275,47 @@ func (e *Executor) validateIndividualDimensionsAndMeasures(ctx context.Context,
...
+	for i, d := range e.metricsView.Dimensions {
+		if d.Type == nil || (d.Type.Code != runtimev1.Type_CODE_TIMESTAMP && d.Type.Code != runtimev1.Type_CODE_DATE) {
+			continue
+		}
+		if d.Column != "" && d.Column != d.Name {
+			res.DimensionErrs = append(res.DimensionErrs, IndexErr{
+				Idx: i,
+				Err: fmt.Errorf("dimension %q must have the same name as the column as it is of type %s", d.Name, d.Type.Code),
+			})
+		}
```
begelundmuller: Why can't we allow a different name than the underlying column? At least when the time dimension is explicitly defined

runtime/metricsview/query.go
```diff
@@ -26,6 +26,7 @@ type Query struct {
 	TimeZone            string      `mapstructure:"time_zone"`
 	UseDisplayNames     bool        `mapstructure:"use_display_names"`
 	Rows                bool        `mapstructure:"rows"`
+	TimeColumn          string      `mapstructure:"time_column"` // optional time column to use for time-based operations, if not specified, the default time column in the metrics view is used
```
begelundmuller: First, can we avoid using the word "column" since that refers to underlying table columns, which doesn't match dimension/field, which can also have expressions?

Second, instead of passing the time dimension as a root property, would it be possible to take it in the individual constructs instead, specifically in `Query.Dimensions[*].Compute.Dimension` (already exists) and `TimeRange.Dimension` (would need to be added)?

That would enable e.g. aggregating by two different time dimensions, or filtering by a different time dimension than you are aggregating by. I'm not sure, but it _might_ also make the code a bit simpler.

Only thing is for comparisons, I think we'd have to require that `Query.TimeRange.Dimension == Query.ComparisonTimeRange.Dimension`.

runtime/parser/parse_metrics_view.go
```diff
@@ -307,7 +307,13 @@ func (p *Parser) parseMetricsView(node *Node) error {
...
 		lower := strings.ToLower(dim.Name)
 		if _, ok := names[lower]; ok {
-			return fmt.Errorf("found duplicate dimension or measure name %q", dim.Name)
+			// allow time dimension to be defined in the dimensions list if it refers to a column then the name should be equal to the column name
+			if !strings.EqualFold(lower, tmp.TimeDimension) {
+				return fmt.Errorf("found duplicate dimension or measure name %q", dim.Name)
+			}
+			if dim.Expression == "" && dim.Column != dim.Name {
+				return fmt.Errorf("cannot use alias for time dimension, use same name as column name")
+			}
```
begelundmuller: Why not?

runtime/reconcilers/metrics_view.go
```diff
@@ -103,7 +103,7 @@ func (r *MetricsViewReconciler) Reconcile(ctx context.Context, n *runtimev1.Reso
 	// NOTE: Not checking refs for errors since they may still be valid even if they have errors. Instead, we just validate the metrics view against the table name.
 
 	// Validate the metrics view and update ValidSpec
-	e, err := metricsview.NewExecutor(ctx, r.C.Runtime, r.C.InstanceID, mv.Spec, !hasInternalRef, runtime.ResolvedSecurityOpen, 0)
+	e, err := metricsview.NewExecutor(ctx, r.C.Runtime, r.C.InstanceID, mv.Spec, !hasInternalRef, runtime.ResolvedSecurityOpen, 0, "")
```
begelundmuller: In this section it implicitly adds the types, but it's not very clear. So that's why I think it might be better with an explicit step for rewriting the metrics view with the types.

runtime/metricsview/ast.go
pjain1: Hmm good suggestion, I think we can handle this while writing time filter expression.

runtime/metricsview/query.go
pjain1: > instead of passing the time dimension as a root property, would it be possible to take it in the individual constructs instead, specifically in Query.Dimensions[*].Compute.Dimension (already exists) and TimeRange.Dimension (would need to be added)?

I understand that a `Dimension` field can be added to `TimeRange` that specifies which time dim to use for filtering instead of passing `time_dimension` as a separate argument in the API. I don't understand what you mean by passing time dim in `Query.Dimensions[*].Compute.Dimension` is it so that it can be rolled up and used in the group by? thats already possible even when using more than time dims.

runtime/parser/parse_metrics_view.go
pjain1: I wanted to disallow redefining same time column multiple times but I guess its also allowed for other dimensions so I can remove the restriction.

runtime/metricsview/executor_validate.go
pjain1: I wanted to disallow redefining same time column multiple times but I guess its also allowed for other dimensions so I can remove the restriction.

runtime/metricsview/query.go
begelundmuller: >  I don't understand what you mean by passing time dim in Query.Dimensions[*].Compute.Dimension
Currently it works like this:
```
{"dimensions": {"name": "__time__hour", "compute":{"time_floor": {"grain": "hour", "dimension":"__time"}}}
```

Notice how the first `name` field is just an alias, and the actual time dimension is given in the nested `query.dimensions[*].compute.time_floor.dimension` property?

So what I mean is, the time dimension name is explicitly given here, so a base `query.time_column` can perhaps be avoided.

This also allows a query to filter the time range by another time dimension than they are doing the grouping by.

runtime/metricsview/executor_validate.go
pjain1: > It feels unexpected that it rewrites the metrics view in place. It's also not safe to do that unless it was cloned when being read from the catalog.


its just appending types not deleting or changing anything which feels safe

> I wonder if the code flows could be simplified with something like mv, err := e.NormalizeMetricsView(ctx context.Context) that clones the metrics view and adds schema info (and in the future, maybe it does other rewrites).

then how will the state be updated in catalog as the application should get the types, should we update the cloned spec in state?

begelundmuller: If you look at the metrics view reconciler, it already updates `State.ValidSpec` to the rewritten spec and calls `r.C.UpdateState` to save it to the catalog. So it will get persisted there.

pjain1: yes its done [here](https://github.com/rilldata/rill/blob/main/runtime/reconcilers/metrics_view.go#L142) but it updates with the same [spec](https://github.com/rilldata/rill/blob/main/runtime/reconcilers/metrics_view.go#L136) sent to executor so my point is if we clone the spec then we will need to return cloned spec and use that?

pjain1: Doing the normalization after validation is done as we need schema for normalization and schema resolution will fail if mv is not valid. Also doing it in place in a separate method so caller can choose whether to use it or not and since its done after validation it returns a cloned mv spec.

proto/rill/runtime/v1/resources.proto
```diff
@@ -188,6 +188,8 @@ message MetricsViewSpec {
     string lookup_key_column = 9;
     string lookup_value_column = 10;
     string lookup_default_expression = 11;
+    // The data type of the dimension
+    Type data_type = 12;
```
begelundmuller: nit:
```suggestion
    // The data type of the dimension. Only populated in ValidSpec.
    Type data_type = 12;
```

proto/rill/runtime/v1/resources.proto
```diff
@@ -227,6 +229,8 @@ message MetricsViewSpec {
     google.protobuf.Struct format_d3_locale = 13;
     bool valid_percent_of_total = 6;
     string treat_nulls_as = 14;
+    // The data type of the measure
```
begelundmuller:
```suggestion
    // The data type of the measure. Only populated in ValidSpec.
```

runtime/metricsview/ast.go
```diff
@@ -130,6 +131,12 @@ func NewAST(mv *runtimev1.MetricsViewSpec, sec *runtime.ResolvedSecurity, qry *Q
 		return nil, fmt.Errorf("must specify at least one dimension or measure")
 	}
 
+	// Use provided time column if available, otherwise fall back to TimeDimension
+	timeDim := mv.TimeDimension
+	if qry.TimeRange != nil && qry.TimeRange.TimeDimension != "" {
+		timeDim = qry.TimeRange.TimeDimension
+	}
```
begelundmuller, line 137: Maybe add validation that `ComparisonTimeRange` uses the same time dimension? Something like:
```go
if qry.ComparisonTimeRange != nil && qry.ComparisonTimeRange.TimeDimension != timeDim {
  return nil, fmt.Errorf(...)
}
```

runtime/metricsview/ast.go
```diff
@@ -26,6 +26,7 @@ type AST struct {
 	comparisonDimFields []FieldNode
 	unnests             []string
 	nextIdentifier      int
+	timeField           *FieldNode
```
begelundmuller: Maybe not possible, but looking at the diff here, does the time field even need to be stored as an instance variable? Since it seems like it's contextually available most places, like you could do `timeColumnOrExpr(TimeRange.TimeDimension)` or something like that most places.

Seems like it's not available for `TimeSpine`, but probably it should also be added there? Like `TimeSpine.TimeDimension`.

runtime/metricsview/executor.go
```diff
@@ -130,36 +131,55 @@ func (e *Executor) CacheKey(ctx context.Context) ([]byte, bool, error) {
...
-	e.timestamps.Now = time.Now()
-	return e.timestamps, nil
+	res.Now = time.Now()
+	e.timestamps[timeDim] = res
+
+	return res, nil
 }
 
 // BindQuery allows to set min, max and watermark from a cache.
 func (e *Executor) BindQuery(ctx context.Context, qry *Query, timestamps TimestampsResult) error {
```
begelundmuller, line 176: Consider explicitly passing the `timeDim` name to add it for, similar to in `Timestamps`

runtime/metricsview/executor.go
```diff
@@ -676,6 +701,37 @@ func (e *Executor) executeSearchInDruid(ctx context.Context, qry *SearchQuery, e
...
 
+func (e *Executor) validateQuery(qry *Query) error {
+	err := qry.Validate()
+	if err != nil {
+		return err
+	}
+
+	if e.metricsView.TimeDimension == "" && qry.TimeRange != nil && qry.TimeRange.TimeDimension != "" {
+		return fmt.Errorf("time_dimension cannot be used with metrics views that does not have a primary timeseries defined")
+	}
```
begelundmuller: Is this necessary? In theory, if there's a time type dimension it could be used even if no primary time series is explicitly defined.

runtime/metricsview/executor_validate.go
```diff
@@ -52,8 +53,8 @@ func (r *ValidateMetricsViewResult) Error() error {
 	return errors.Join(errs...)
 }
 
-// ValidateMetricsView validates the dimensions and measures in the executor's metrics view.
-func (e *Executor) ValidateMetricsView(ctx context.Context) (*ValidateMetricsViewResult, error) {
+// ValidateMetricsView validates the dimensions and measures in the executor's metrics view and returns a ValidateMetricsViewResult and the schema of the metrics view.
+func (e *Executor) ValidateMetricsView(ctx context.Context) (*ValidateMetricsViewResult, map[string]*runtimev1.Type, error) {
```
begelundmuller: Can it use the standard `*runtimev1.StructType` type instead of a `map`? And then convert to map if needed elsewhere (it's just validation, not hot path so doesn't matter with a few extra allocations).

runtime/metricsview/executor_validate.go
```diff
@@ -136,12 +130,18 @@ func (e *Executor) ValidateMetricsView(ctx context.Context) (*ValidateMetricsVie
...
+		schema, err := e.Schema(ctx)
 		if err != nil {
-			res.OtherErrs = append(res.OtherErrs, fmt.Errorf("failed to validate metrics view schema: %w", err))
+			return nil, nil, fmt.Errorf("failed to resolve metrics view schema: %w", err)
 		}
+		for _, f := range schema.Fields {
+			mvSchema[f.Name] = f.Type
+		}
+
+		e.validateSchema(mvSchema, res)
```
begelundmuller: Can it require `NormalizeMetricsView` to have been called before validation? Or call it as part of validation? And then check the field types directly in the metrics view properties instead of computing the schema separately?

runtime/metricsview/executor_validate.go
```diff
@@ -312,20 +378,10 @@ func (e *Executor) validateMeasure(ctx context.Context, t *drivers.Table, m *run
...
-	types := make(map[string]*runtimev1.Type, len(schema.Fields))
-	for _, f := range schema.Fields {
-		types[f.Name] = f.Type
-	}
-
+func (e *Executor) validateSchema(mvTypes map[string]*runtimev1.Type, res *ValidateMetricsViewResult) {
 	// Check that the measures are not strings
 	for i, m := range e.metricsView.Measures {
-		typ, ok := types[m.Name]
+		typ, ok := mvTypes[m.Name]
```
begelundmuller: Like here, if normalize had been called, it could perhaps just be `m.DataType` instead?

runtime/parser/parse_metrics_view.go
```diff
@@ -262,6 +262,7 @@ func (p *Parser) parseMetricsView(node *Node) error {
 
 	names := make(map[string]uint8)
 	names[strings.ToLower(tmp.TimeDimension)] = nameIsDimension
+	timeSeen := false
```
begelundmuller, line 265: Instead of tracking `timeSeen` separately, I think the same can be achieved by moving line 264 to after all the other dimensions are added, like:
```go
if _, ok := names[strings.ToLower(tmp.TimeDimension)]; !ok {
  	names[strings.ToLower(tmp.TimeDimension)] = nameIsDimension
}
```

runtime/reconcilers/alert_test.go
```diff
@@ -67,7 +67,13 @@ email:
...
 	testruntime.RequireReconcileState(t, rt, id, 4, 0, 0)
 
-	_, metricsRes := newMetricsView("mv1", "bar", "__time", []string{"count(*)"}, []string{"country"})
+	_, metricsRes := newMetricsView("mv1", "bar", "__time",
+		[]fieldWithType{
+			{Name: "count(*)", Type: &runtimev1.Type{Code: runtimev1.Type_CODE_INT64, Nullable: true}},
+		},
+		[]fieldWithType{
+			{Name: "country", Type: &runtimev1.Type{Code: runtimev1.Type_CODE_STRING, Nullable: true}},
+		})
```
begelundmuller: nit, but the point of the helper function here is to make the tests easy to read/write, not it looks quite complex. Can you think of some other option that's easier to read? Probably there are better options, but one idea is using `any` and mixing like:
```go
_, metricsRes := newMetricsView("mv1", "bar", "__time", []string{"count(*)", runtimev1.Type_CODE_INT64}, []string{"country", runtimev1.Type_CODE_STRING})
```

runtime/reconcilers/metrics_view.go
```diff
@@ -132,8 +132,11 @@ func (r *MetricsViewReconciler) Reconcile(ctx context.Context, n *runtimev1.Reso
 		return runtime.ReconcileResult{Err: validateErr}
 	}
 
+	// If the validation was successful, update the data types in the spec.
+	normalizedMVSpec := e.NormalizeMetricsView(mvSchema)
+
```
begelundmuller: Mentioned this elsewhere, but what about running `NormalizeMetricsView` before `ValidateMetricsView`? So that `ValidateMetricsView` can just inspect the `DataType` property in the spec instead of needing to pass the schema around. 

Or perhaps having a combined `NormalizeAndValidateMetricsView` function instead that clones and validates `e.metricsView`?

runtime/metricsview/ast.go
pjain1: its already checked [here](https://github.com/rilldata/rill/pull/7366/files#diff-5ecb74cb513e8c0b0c107d39a82782a0c4a51f19e487277c33fcf586ea7b41deR109)

runtime/metricsview/executor.go
pjain1: Either the `qry` will have the time dimension in time range otherwise metrics view will have it, why it needs to explicitly passed?

runtime/metricsview/executor.go
pjain1: Yes in theory its possible. I think the main reason is if user is going to use a time dimension for filtering then let them define a primary time dimension so that watermark can be calculated for cache invalidation for rill managed models. 

May be deprecate `timeseries` field but no way to know which is the default time field for watermarks.

runtime/metricsview/ast.go
pjain1: Yes Its possible

runtime/metricsview/executor_validate.go
pjain1: The main issue is `NormalizeMetricsView` needs full schema and we cannot get full schema until we have a valid metrics view otherwise calling e.Schema may fail with various errors. Thats why doing normalization after validation and it expects a valid schema. Also kept it as a separate step as discussed earlier so that caller can choose to just call `ValidateMetricsView` without getting mv spec mutated.

runtime/metricsview/executor_validate.go
pjain1: https://github.com/rilldata/rill/pull/7366#discussion_r2145892550

runtime/parser/parse_metrics_view.go
pjain1: Yes I did that earlier but it was allowing the case of `timeseries` to be different from dimension name which can cause confusion to users. They may think they are actually using the defined dimension
```
timeseries: time,

dimension:
  - name: Time
     expression: <some sql expr>
```
Other case like below where it refers to an alias with different case, sql error will be thrown from olap instead of a clear validation error.
```
timeseries: Primary_time,

dimension:
  - name: primary_time
     column: time
```
Both are edge cases, let me know what you think.

runtime/reconcilers/metrics_view.go
pjain1: https://github.com/rilldata/rill/pull/7366#discussion_r2145892550

runtime/reconcilers/alert_test.go
pjain1: changed

runtime/metricsview/executor.go
pjain1: We may just ignore cache key if no primary defined, may just lift this restriction. Let me check.

pjain1: removed

runtime/metricsview/executor.go
begelundmuller: Probably not important, can just keep as it is

runtime/parser/parse_metrics_view.go
begelundmuller: The case insensitivity is fine. As for the expression to use, in these examples, will it use the `expression:` or will it use a column with the name of the dimension? I'd say the former is expected, basically if the name in `timeseries:` matches a name under `dimensions:`, it should use the column/expression defined under `dimensions:`. Only if it does not match a name there should it fall back to using a column with the same name as in `timeseries:`.

runtime/metricsview/executor_validate.go
begelundmuller: Okay, so it needs to validate some things before it's safe to call `e.Schema`? That makes sense, but then once it calls `e.Schema`, can it then add the types to the metrics view instead of building `mvSchema`? Just seems like it might be simpler to access the types on the dims/measures instead of passing `mvSchema` around.

Maybe it should all be a combined function called `ValidateAndNormalizeMetricsView`?

pjain1: Yes it can populate the schema, thats what was done in first version. I can revert to that and just change the method name to `ValidateAndNormalizeMetricsView`. Are we okay with mutation then with a note on the method definition or should `ValidateAndNormalizeMetricsView` return a clone of mv spec?

pjain1: BTW the diff is outdated.

runtime/parser/parse_metrics_view.go
pjain1: It will using the column with name of the dimension like `Primary_time` and not the expression, however it may not work with all olap engines I think as they may be case sensitive. Also the schema query fails with duplicate alias error if the dimension definition contains an expression. To avoid these confusion shouldn't we just disallow doing this.

runtime/metricsview/executor_validate.go
begelundmuller: I think that sounds good – if it has "normalize" in the name, then mutation is probably fine. Your decision :)

runtime/parser/parse_metrics_view.go
begelundmuller: Oh, it should use the expression. Maybe I'm not fully following this, but I'd say the expected behavior is:
1. `timeseries: <dim name>` should set the default time dimension to be the dimension named `<dim name>` in the `dimensions:` list.
2. For backwards compatibility, if `timeseries: <dim name>` references a `<dim name>` that doesn't exist in the `dimensions:` list, it should default to a dimension with `column: <dim name>`.

The lookups should either be case insensitive or fail if the case is wrong – it shouldn't be possible for both `Primary_time` and `primary_time` to exist and have different definitions.

pjain1: The behaviour you want is what is happening now. But stepping back, do you want allow this
```
timeseries: time,

dimension:
  - name: Time
     expression: <some sql expr>
```
If yes then `timeseries` should refer to `<some sql expr>` instead `time` column in the model?

runtime/metricsview/ast.go
```diff
@@ -783,21 +810,47 @@ func (a *AST) buildSpineSelect(alias string, spine *Spine, tr *TimeRange) (*Sele
...
+	expr, args := a.sqlForTimeRange(timeDimExpr, tr.Start, tr.End)
 	n.TimeWhere = &ExprNode{
 		Expr: expr,
 		Args: args,
 	}
+
+	return nil
+}
+
+func (a *AST) getTimeDimensionExpression(tr *TimeRange) (string, error) {
```
begelundmuller: nits: maybe add a little docstring? it's a big file and most functions in it have a little docstring

runtime/server/mcp.go
```diff
@@ -283,6 +294,25 @@ Example: Get the total revenue by country and month for 2024:
...
+        "metrics_view": "ecommerce_financials",
+        "dimensions": [
+            {"name": "event_time", "compute": {"time_floor": {"dimension": "event_time", "grain": "month"}}},
+            {"name": "country"}
+        ],
+        "measures": [{"name": "total_revenue"}],
+        "time_range": {
+            "start": "2024-01-01T00:00:00Z",
+            "end": "2025-01-01T00:00:00Z",
+			"time_dimension"": "order_shipped_time",
```
begelundmuller: Indentation is off here

runtime/server/mcp.go
```diff
@@ -226,6 +235,8 @@ func (s *Server) mcpQueryMetricsView() (mcp.Tool, server.ToolHandlerFunc) {
 Perform an arbitrary aggregation on a metrics view.
 Tip: Use the 'sort' and 'limit' parameters for best results and to avoid large, unbounded result sets.
 Important note: The 'time_range' parameter is inclusive of the start time and exclusive of the end time.
+Note: 'time_dimension' is an optional parameter under "time_range" that can be used to specify the time dimension to use for the time range. 
+If not provided, the default time column of the metrics view will be used.
```
begelundmuller: Keep it as one line so AI knows the second sentence is related to the first
```suggestion
Note: 'time_dimension' is an optional parameter under "time_range" that can be used to specify the time dimension to use for the time range.  If not provided, the default time column of the metrics view will be used.
```

begelundmuller (approved):

runtime/server/mcp.go
pjain1: Interesting I used tabs and it looks correct on IDE, will change to use spaces

PR #7367, opened 2025-05-26
feat: ui integration of report metadata

mindspank:
UI integration for https://github.com/rilldata/rill/pull/7318

Adds the ability to include metadata for exports and reports.
In Rill Developer exports exclude the dashboard url and other rill cloud metadata.

Closes ENG-613

TODO: Add dashboard url to reports. Will handle in a separate PR.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-common/src/features/exports/ExportMenu.svelte
+++ b/web-common/src/features/exports/ExportMenu.svelte
@@ -104,7 +104,7 @@
     >
       Export as CSV
     </DropdownMenu.Item>
-    {#if !workspace && exportHeader}
+    {#if !workspace && $exportHeader}
       <DropdownMenu.Item
         on:click={() =>
           handleExport({
@@ -131,7 +131,7 @@
     >
       Export as XLSX
     </DropdownMenu.Item>
-    {#if !workspace && exportHeader}
+    {#if !workspace && $exportHeader}
       <DropdownMenu.Item
         on:click={() =>
           handleExport({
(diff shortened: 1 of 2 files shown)


web-common/src/features/exports/ExportMenu.svelte
```diff
@@ -37,14 +38,21 @@
   }
 
   const exportDash = createQueryServiceExport();
-  const { reports } = featureFlags;
+  const { reports, adminServer, exportMetadata } = featureFlags;
 
-  async function handleExport(format: V1ExportFormat) {
+  async function handleExport(format: V1ExportFormat, includeHeader = false) {
```
ericpgreen2, line 43: I don't like boolean flag function parameters. It makes the call site hard to read. E.g. when I see `handleExport(V1ExportFormat.EXPORT_FORMAT_CSV, true)}`, I have no way of knowing what `true` indicates unless I read the function signature. Instead, how about we use an options object here:

```ts
async function handleExport(options: { format: V1ExportFormat; includeHeader?: boolean }) {
  const { format, includeHeader = false } = options;
  // ...rest of the code
}
```
...which is more clear at the call site:
```ts
handleExport({ format: V1ExportFormat.EXPORT_FORMAT_CSV, includeHeader: true });
```

ericpgreen2 (changes requested):

web-common/src/features/scheduled-reports/BaseScheduledReportForm.svelte
```diff
@@ -108,6 +109,16 @@
...
+  <Checkbox
+    bind:checked={$data["exportIncludeHeader"]}
+    id="exportIncludeHeader"
+    onCheckedChange={(checked) => {
+      $data["exportIncludeHeader"] = checked;
+    }}
+    inverse
+    disabled={$data["exportFormat"] === V1ExportFormat.EXPORT_FORMAT_PARQUET}
+    label="Include metadata"
+  />
```
ericpgreen2, line 121: Should this be gated by the feature flag too? "Include metadata" doesn't feel like a sufficiently clear explanation. A tooltip could help.

ericpgreen2 (approved):
Looks good. Thanks for doing this. I like that it's behind a feature flag for now, until we get the UI right.

PR #7379, opened 2025-05-28
feat: Add an AI page with MCP instructions 

ericpgreen2:
Private projects:
image

Public projects:
image

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-admin/src/features/ai/MCPConfigSection.svelte
+++ b/web-admin/src/features/ai/MCPConfigSection.svelte
@@ -1,0 +1,55 @@
+<script lang="ts">
+  import { ADMIN_URL } from "@rilldata/web-admin/client/http-client";
+  import CodeBlock from "@rilldata/web-common/components/code-block/CodeBlock.svelte";
+
+  export let organization: string;
+  export let project: string;
+  export let isPublic: boolean;
+
+  // Construct the API URL for the MCP server
+  $: apiUrl = `${ADMIN_URL}/v1/organizations/${organization}/projects/${project}/runtime/mcp/sse`;
+
+  // Config snippets with exact formatting
+  $: publicConfig = `{
+  "mcpServers": {
+    "rill": {
+      "command": "npx",
+      "args": [
+        "mcp-remote",
+        "${apiUrl}"
+      ]
+    }
+  }
+}`;
+
+  $: privateConfig = `{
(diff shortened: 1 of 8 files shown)


web-admin/src/features/ai/MCPConfigSection.svelte
```diff
@@ -0,0 +1,49 @@
+<script lang="ts">
+  import { ADMIN_URL } from "@rilldata/web-admin/client/http-client";
+  import CodeBlock from "@rilldata/web-common/components/code-block/CodeBlock.svelte";
+
+  export let organization: string;
+  export let project: string;
+  export let isPublic: boolean;
+
+  // Construct the API URL for the MCP server
+  $: apiUrl = `${ADMIN_URL}/v1/organizations/${organization}/projects/${project}/runtime/mcp/sse`;
```
nishantmonu51, line 10: Do we want to do a replace for `admin.rilldata` with `api.rilldata`. The recommended urls when proxying are https://api.rilldata.com/.... instead of https://admin.rilldata.com/....

nishantmonu51 (commented):
LGTM, overall.

web-admin/src/features/ai/MCPConfigSection.svelte
ericpgreen2: Yeah, what is the right logic here? Is it just a find/replace? What is recommended for custom domains?

begelundmuller: `api.rilldata.com` is the best one to use for user-facing endpoints (eventually we should also switch away from `admin.rilldata.com` internally, but that'll require some extra QA). We don't support MCP on custom domains at the moment.

begelundmuller:
Not blockers for this, but two thoughts:
1. If the user clicks "create token", it'd be nice if it also templated it into the Claude config so you could one-click copy instead of copy/pasting twice and maybe messing up.
2. If the project is public and the user is not logged in, hide the "create token" section and remove the `Authorization` header fields from the Claude config.

web-admin/src/features/ai/MCPConfigSection.svelte
ericpgreen2: Thanks, updated

PR #7384, opened 2025-05-29
using display name instead of resource name

NamanMahor:
using display name instead of resource name

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


proto/rill/runtime/v1/queries.proto
```diff
@@ -301,9 +301,9 @@ message ExportRequest {
   string baked_query = 5 [deprecated = true];
   // If true, the export will include header comments with metadata about the export.
   bool include_header = 6;
-  // Optional name of the dashboard the export originates from.
+  // Optional display name of the dashboard the export originates from.
   // Only used if include_header is true.
-  ResourceName origin_dashboard = 7;
+  string origin_dashboard = 7;
```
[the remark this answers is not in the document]
NamanMahor: make sense.

runtime/queries/metricsview_aggregation.go
```diff
@@ -446,8 +448,9 @@ func (q *MetricsViewAggregation) generateExportHeaders(ctx context.Context, rt *
 	if org != "" && project != "" {
 		parts = append(parts, org, project)
 	}
-	if opts.OriginDashboard != nil {
-		parts = append(parts, opts.OriginDashboard.Name)
+	dashboardDisplayName := q.getDisplayName(ctx, rt, instanceID, opts.OriginDashboard)
+	if dashboardDisplayName != "" {
+		parts = append(parts, dashboardDisplayName)
```
begelundmuller: The origin dashboard property is optional (some exports may be created outside dashboards), so handle it gracefully:

```suggestion
	if opts.OriginDashboard != nil {
		dashboardDisplayName := q.getDisplayName(ctx, rt, instanceID, opts.OriginDashboard)
		if dashboardDisplayName != "" {
			parts = append(parts, dashboardDisplayName)
		}
```

runtime/queries/metricsview_aggregation.go
```diff
@@ -479,6 +482,35 @@ func (q *MetricsViewAggregation) generateExportHeaders(ctx context.Context, rt *
...
+	if err != nil {
+		rt.Logger.Warn("getDisplayName: failed to get controller", zap.Error(err), zap.String("instance_id", instanceID), observability.ZapCtx(ctx))
+		return ""
+	}
+
+	res, err := c.Get(ctx, resourceName, false)
+	if err != nil {
+		rt.Logger.Warn("getDisplayName: failed to get resource", zap.Error(err), zap.String("instance_id", instanceID), zap.String("resource_name", resourceName.Name), zap.String("resource_kind", resourceName.Kind), observability.ZapCtx(ctx))
+		return ""
+	}
```
begelundmuller: Fall back to `resourceName.Name` only `if errors.Is(err, drivers.ErrResourceNotFound)`. In other cases, return the error and bubble it up.

runtime/queries/metricsview_aggregation.go
```diff
@@ -479,6 +482,35 @@ func (q *MetricsViewAggregation) generateExportHeaders(ctx context.Context, rt *
...
+		if explore != nil && explore.Spec != nil {
+			return explore.Spec.DisplayName
+		}
+	case runtime.ResourceKindCanvas:
+		canvas := res.GetCanvas()
+		if canvas != nil && canvas.Spec != nil {
+			return canvas.Spec.DisplayName
+		}
+	}
+	return ""
```
begelundmuller: 1. Use `State.ValidSpec` instead of `Spec` directly
2. Sometimes `spec.DisplayName` is empty – if that is the case, it should return `resourceName.Name` as a fallback.

PR #7386, opened 2025-05-29
[ENG-769] Do not parse string with leading zeros as Number in Cell Inspector

lovincyrus:
This pull request provides a better default state handling and improves the parsing of values with leading zeros.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-common/src/components/CellInspector.svelte
+++ b/web-common/src/components/CellInspector.svelte
@@ -7,7 +7,7 @@
   import { cubicOut } from "svelte/easing";
   import Kbd from "./Kbd.svelte";
 
-  export let value: any = "";
+  export let value: any = null;
   export let isOpen: boolean = false;
 
   let hovered = false;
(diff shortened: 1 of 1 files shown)


avaitla:
Should rill parse any strings as integers at all? Why not just display it as a string since that's what the model says? Otherwise sometimes it displays with commas and sometimes not which is inconsistent behavior.

lovincyrus:
> Should rill parse any strings as integers at all? Why not just display it as a string since that's what the model says? Otherwise sometimes it displays with commas and sometimes not which is inconsistent behavior.

Checking if a value is string or number allows us to parse the number in a comma-separated format.

CleanShot 2025-05-30 at 09 35 36@2x
CleanShot 2025-05-30 at 09 35 22@2x

avaitla:
I see because the measure rollups are all strings, and you're using the same code for inspecting dimensions as well.

djbarnwal (approved):
Looks good!

PR #7389, opened 2025-05-30
Improve token management CLI

grahamplata:
Closes: https://linear.app/rilldata/issue/ENG-766/improve-token-management-cli

```bash
rill token issue --display-name nullstring
Token: rill_usr_4I6tSnrUZ7EDTKNOyiq5a6pWC1W45OeU4WFw6yY0MNhcxj34w0YZAu

rill token list
  ID (5)                                 DESCRIPTION       TOKEN                 CLIENT             CREATED               EXPIRES   LAST USED
 -------------------------------------- ----------------- --------------------- ------------------ --------------------- --------- ---------------------
  380bdf31-17ea-4aaa-811b-78ec3d0c21bd   nullstring        rill_usr_4I6tSnrUZ7   Created manually   2025-05-29 17:55:49             2025-05-29 17:55:49
  008e7ed2-b049-4755-8b9c-48fe9b37010a                     rill_usr_2TRxOIOB9b   Rill CLI           2025-05-29 14:12:16             2025-05-29 17:55:07
  09f00cdd-22bc-48ca-8774-dc5e152ac049   Browser session   rill_usr_PNcMSU0pVX   Rill Web           2025-05-29 14:12:10             2025-05-29 14:12:35
```

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!


admin/database/postgres/postgres.go
```diff
@@ -961,10 +961,10 @@ func (c *connection) FindUserAuthTokens(ctx context.Context, userID, afterID str
 	`)
 	args := []any{userID}
 	if afterID != "" {
-		qry.WriteString(" AND t.id > $2 ORDER BY t.id LIMIT $3")
+		qry.WriteString(" AND t.id > $2 ORDER BY t.created_on DESC, t.id DESC LIMIT $3")
```
begelundmuller: Won't this skip rows? The filter does not match the order by.

E.g. if you have `ORDER BY created_on DESC, id DESC`, your filter needs to be `created_on > ? OR (created_on = ? AND id > ?)`.

On reflection, timestamps are a very annoying pagination key. I suggest keeping ID pagination on the backend. Then in the CLI, use a large page size (e.g. 1000) and IF the result size is smaller than the page size, sort by timestamp in memory there. 

That will use creation time sorting for small lists, and switch to ID based pagination for huge lists. Should be good in practice.

admin/database/postgres/postgres.go
```diff
@@ -1020,6 +1020,12 @@ func (c *connection) DeleteExpiredUserAuthTokens(ctx context.Context, retention
 	return parseErr("auth token", err)
 }
 
+// DeleteInactiveUserAuthTokens deletes user authentication tokens that have not been used within the specified retention period.
+func (c *connection) DeleteInactiveUserAuthTokens(ctx context.Context, retention time.Duration) error {
+	_, err := c.getDB(ctx).ExecContext(ctx, "DELETE FROM user_auth_tokens WHERE used_on IS NULL OR used_on < now() - $1", retention)
```
begelundmuller: In theory `used_on` might be null right after creation (although I think currently we always populate it). But best to be safe here

admin/database/postgres/postgres.go
```diff
@@ -1162,6 +1168,12 @@ func (c *connection) DeleteExpiredServiceAuthTokens(ctx context.Context, retenti
 	return parseErr("service auth token", err)
 }
 
+// DeleteInactiveServiceAuthTokens deletes service authentication tokens that have not been used within the specified retention period.
+func (c *connection) DeleteInactiveServiceAuthTokens(ctx context.Context, retention time.Duration) error {
+	_, err := c.getDB(ctx).ExecContext(ctx, "DELETE FROM service_auth_tokens WHERE used_on IS NULL OR used_on < now() - $1", retention)
```
begelundmuller: Same comment

admin/pkg/authtoken/authtoken.go
```diff
@@ -131,6 +131,14 @@ func (t *Token) SecretHash() []byte {
 	return hashed[:]
 }
 
+// Prefix returns a safe, partial display string for the token, e.g., "rill_usr_abcdefghij"
+func (t *Token) Prefix() string {
```
begelundmuller: The thing is this needs to work even if `t.Secret` is empty (as long as `t.ID` is set). Can you:
1. Note that in the docstring.
2. Add a test case that loops e.g. 100 times and checks that the prefix is correct for random tokens where the `Secret` is empty?

admin/pkg/authtoken/authtoken.go
```diff
@@ -131,6 +131,14 @@ func (t *Token) SecretHash() []byte {
...
+	payload := make([]byte, 40)
+	copy(payload[0:11], t.ID[:])
```
begelundmuller: Just copy the whole ID – even if not used, makes it easier to change the prefix length and reduces chance of edge cases in base62 we maybe didn't think about
```suggestion
	payload := make([]byte, 40)
	copy(payload[0:16], t.ID[:])
```

admin/server/users.go
```diff
@@ -321,6 +332,48 @@ func (s *Server) RevokeUserAuthToken(ctx context.Context, req *adminv1.RevokeUse
 	return &adminv1.RevokeUserAuthTokenResponse{}, nil
 }
 
+// findUserAuthTokenFuzzy attempts to find a user auth token by exact ID, full token string, or unique prefix.
+func (s *Server) findUserAuthTokenFuzzy(ctx context.Context, input string) (*database.UserAuthToken, error) {
```
begelundmuller: Internal functions come after exported functions: https://github.com/uber-go/guide/blob/master/style.md#function-grouping-and-ordering

admin/server/users.go
```diff
@@ -321,6 +332,48 @@ func (s *Server) RevokeUserAuthToken(ctx context.Context, req *adminv1.RevokeUse
...
+	}
+
+	tokens, err := s.admin.DB.FindUserAuthTokens(ctx, "", "", 100)
+	if err != nil {
+		return nil, err
+	}
+
+	var matches []*database.UserAuthToken
+	for _, t := range tokens {
+		if strings.HasPrefix(t.ID, input) || strings.HasPrefix(t.ID, strings.TrimPrefix(input, "rill_usr_")) {
```
begelundmuller: Not sure this works – `t.ID` is a UUID (basically hex encoded), but `strings.TrimPrefix(input, "rill_usr_")` is base62 encoded.

admin/server/users.go
```diff
@@ -321,6 +332,48 @@ func (s *Server) RevokeUserAuthToken(ctx context.Context, req *adminv1.RevokeUse
...
+		if err == nil {
+			return token, nil
+		}
+	}
+
+	if len(input) < 6 {
+		return nil, status.Error(codes.InvalidArgument, "token prefix too short (min 6 chars)")
+	}
+
+	tokens, err := s.admin.DB.FindUserAuthTokens(ctx, "", "", 100)
```
begelundmuller: Does this work? It's passing `""` as the user ID. Would have to be `claims.OwnerID()`.

The fuzzy logic here is a bit hacky, I'm okay with just supporting full IDs or full tokens if you prefer.

If you want to keep it, then let's add tests for the various match cases.

admin/worker/delete_unsued_tokens.go
```diff
@@ -0,0 +1,23 @@
+package worker
```
begelundmuller: Can you create the job in `admin/jobs/river` instead? We're migrating to that for background jobs and eventually will remove the `worker` package.

https://github.com/rilldata/rill/blob/d24cf0179d58b96af81ac256c282751bb925fe98/admin/jobs/river/river.go#L96

cli/cmd/token/issue.go
```diff
@@ -21,10 +21,17 @@ func IssueCmd(ch *cmdutil.Helper) *cobra.Command {
 				return err
 			}
 
+			if displayName == "" {
```
begelundmuller: Respect the `--interactive=false` flag:

```suggestion
			if ch.Interactive && displayName == "" {
```

cli/pkg/printer/resources.go
```diff
@@ -281,16 +281,18 @@ func toServiceTokenRow(s *adminv1.ServiceToken) *serviceToken {
...
+		ExpiresOn:   expiresOn,
 	}
 }
 
 type serviceToken struct {
-	ID        string `header:"id" json:"id"`
-	CreatedOn string `header:"created_on,timestamp(ms|utc|human)" json:"created_on"`
-	ExpiresOn string `header:"expires_on,timestamp(ms|utc|human)" json:"expires_on"`
+	ID          string `header:"id" json:"id"`
+	TokenPrefix string `header:"token" json:"token_prefix"`
```
begelundmuller: Put `header:"prefix"` or `header:"token prefix"` since it's not the full token. Since this is in context of a token listing command, putting "token" is perhaps not needed

proto/rill/admin/v1/api.proto
```diff
@@ -2640,6 +2640,7 @@ message Bookmark {
 
 message ServiceToken {
   string id = 1;
+  string token_prefix = 4;
```
begelundmuller: I suggest just naming it `prefix`, not `token_prefix`.

The type serves as scope to provide context. Like `ServiceToken.prefix` is probably clear enough. Similar to how the other fields are not `ServiceToken.token_expires_on`, but just `ServiceToken.expires_on`.

proto/rill/admin/v1/api.proto
```diff
@@ -2650,6 +2651,7 @@ message UserAuthToken {
   string auth_client_id = 3;
   string auth_client_display_name = 4;
   string representing_user_id = 5;
+  string token_prefix = 9;
```
begelundmuller: Same comment as above

admin/worker/delete_unsued_tokens.go
grahamplata: Can do, want me to create an issue for that?

admin/database/postgres/postgres.go
```diff
@@ -1020,6 +1020,12 @@ func (c *connection) DeleteExpiredUserAuthTokens(ctx context.Context, retention
 	return parseErr("auth token", err)
 }
 
+// DeleteInactiveUserAuthTokens deletes user authentication tokens that have not been used within the specified retention period.
+func (c *connection) DeleteInactiveUserAuthTokens(ctx context.Context, retention time.Duration) error {
+	_, err := c.getDB(ctx).ExecContext(ctx, "DELETE FROM user_auth_tokens WHERE used_on IS NULL OR used_on < (now() - $1)", retention)
```
begelundmuller: > In theory used_on might be null right after creation (although I think currently we always populate it). But best to be safe here

This review comment is still not addressed. If `used_on` is `null` right after creation and this query runs, it deletes a token that was literally just created (and hence shouldn't be considered inactive).

admin/database/postgres/postgres.go
```diff
@@ -1162,6 +1168,12 @@ func (c *connection) DeleteExpiredServiceAuthTokens(ctx context.Context, retenti
 	return parseErr("service auth token", err)
 }
 
+// DeleteInactiveServiceAuthTokens deletes service authentication tokens that have not been used within the specified retention period.
+func (c *connection) DeleteInactiveServiceAuthTokens(ctx context.Context, retention time.Duration) error {
+	_, err := c.getDB(ctx).ExecContext(ctx, "DELETE FROM service_auth_tokens WHERE used_on IS NULL OR used_on < (now() - $1)", retention)
```
begelundmuller: Same comment

admin/jobs/river/river.go
```diff
@@ -101,6 +105,8 @@ func New(ctx context.Context, dsn string, adm *admin.Service) (jobs.Client, erro
 		newPeriodicJob(&TrialEndCheckArgs{}, "10 1 * * *", true),                 // daily at 1:10am UTC
 		newPeriodicJob(&TrialGracePeriodCheckArgs{}, "15 1 * * *", true),         // daily at 1:15am UTC
 		newPeriodicJob(&SubscriptionCancellationCheckArgs{}, "20 1 * * *", true), // daily at 1:20am UTC
+		newPeriodicJob(&DeleteUnusedUserTokenArgs{}, "25 1 * * *", true),         // daily at 1:25am UTC
+		newPeriodicJob(&DeleteUnusedServiceTokenArgs{}, "30 1 * * *", true),      // daily at 1:30am UTC
```
begelundmuller: Just do a simple cron tab. The offset-by-5 mins thing is just the billing jobs.

admin/jobs/river/token_jobs.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// NewDeleteUnusedUserTokenWorker creates a new worker that deletes unused user tokens.
+func (w *DeleteUnusedUserTokenWorker) Work(ctx context.Context, job *river.Job[DeleteUnusedUserTokenArgs]) error {
+	err := w.admin.DB.DeleteInactiveUserAuthTokens(ctx, tokenRetentionPeriod)
+	if err != nil {
+		if errors.Is(err, database.ErrNotFound) {
+			// No tokens to delete, this is not an error.
+			return nil
+		}
```
begelundmuller: Don't need to check this – it's a multi-row delete, so doesn't return single row not found errors

admin/jobs/river/token_jobs.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// NewDeleteUnusedServiceTokenWorker creates a new worker that deletes unused service tokens.
+func (w *DeleteUnusedServiceTokenWorker) Work(ctx context.Context, job *river.Job[DeleteUnusedServiceTokenArgs]) error {
+	err := w.admin.DB.DeleteInactiveServiceAuthTokens(ctx, tokenRetentionPeriod)
+	if err != nil {
+		if errors.Is(err, database.ErrNotFound) {
+			// No tokens to delete, this is not an error.
+			return nil
+		}
```
begelundmuller: Same comment

admin/pkg/authtoken/authtoken.go
```diff
@@ -131,13 +131,27 @@ func (t *Token) SecretHash() []byte {
...
 // marshalBase62 marshals a byte slice to a string of [0-9A-Za-z] characters.
 func marshalBase62(val []byte) string {
 	var i big.Int
 	i.SetBytes(val)
 	return i.Text(62)
 }
 
+func UnmarshalBase62(s string) ([]byte, bool) {
+	return unmarshalBase62(s)
+}
```
begelundmuller: 1. This breaks the style guide – exported functions come before internal functions + should have a docstring.
2. Base62 is an implementation detail, as is the token format in general. Any parsing/serialization of tokens should happen inside this package. So this package shouldn't really expose this kind of function.

admin/pkg/authtoken/authtoken_test.go
```diff
@@ -98,3 +99,21 @@ func TestMany(t *testing.T) {
...
+}
+
+func TestPrefixWithEmptySecret(t *testing.T) {
+	for i := 0; i < 100; i++ {
+		tkn := NewRandom(TypeUser)
+		tkn.Secret = [24]byte{}
+		prefix := tkn.Prefix()
+		require.True(t, strings.HasPrefix(prefix, "rill_usr_"), "Prefix should start with 'rill_usr_'")
+		require.Len(t, prefix, len("rill_usr_")+10, "Prefix length should be correct")
+		require.True(t, strings.HasPrefix(tkn.String(), prefix), "Full token should start with prefix")
```
begelundmuller: This test doesn't check that the token without a secret yields the same prefix as the token with the secret

admin/server/service.go
```diff
@@ -169,10 +171,18 @@ func (s *Server) ListServiceAuthTokens(ctx context.Context, req *adminv1.ListSer
 
 	dtos := make([]*adminv1.ServiceToken, len(tokens))
 	for i, token := range tokens {
+		id, err := uuid.Parse(token.ID)
+		if err != nil {
+			return nil, status.Error(codes.Internal, "invalid token ID format")
+		}
+
+		t := authtoken.Token{Type: authtoken.TypeService, ID: id}
```
begelundmuller: Consider adding a function for this in the `authtoken` package, e.g.:
```
prefix := authtoken.FromID(authtoken.TypeService, id).Prefix()
```

admin/server/users.go
```diff
@@ -215,6 +218,13 @@ func (s *Server) ListUserAuthTokens(ctx context.Context, req *adminv1.ListUserAu
 			expiresOn = timestamppb.New(*t.ExpiresOn)
 		}
 
+		id, err := uuid.Parse(t.ID)
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "invalid token ID %q: %v", t.ID, err)
+		}
+
+		token := &authtoken.Token{Type: authtoken.TypeUser, ID: id}
```
begelundmuller: Same comment

admin/server/users.go
```diff
@@ -618,3 +629,66 @@ func projInviteToPB(i *database.ProjectInviteWithRole) *adminv1.ProjectInvite {
...
+// findUserAuthTokenFuzzy attempts to find a user auth token by exact ID, full token string, or unique prefix.
+func (s *Server) findUserAuthTokenFuzzy(ctx context.Context, input string) (*database.UserAuthToken, error) {
+	claims := auth.GetClaims(ctx)
+	userID := claims.OwnerID()
+
+	// Try exact ID match
+	token, err := s.admin.DB.FindUserAuthToken(ctx, input)
+	if err == nil {
+		return token, nil
+	}
```
begelundmuller: Return the error unless it's a not found error

admin/server/users.go
```diff
@@ -618,3 +629,66 @@ func projInviteToPB(i *database.ProjectInviteWithRole) *adminv1.ProjectInvite {
...
+
+	// Try full token string
+	tokenStr, err := authtoken.FromString(input)
+	if err == nil {
+		token, err := s.admin.DB.FindUserAuthToken(ctx, tokenStr.ID.String())
+		if err == nil {
+			return token, nil
+		}
```
begelundmuller, line 657: Return the error unless it's a not found error

admin/server/users.go
```diff
@@ -618,3 +629,66 @@ func projInviteToPB(i *database.ProjectInviteWithRole) *adminv1.ProjectInvite {
...
+
+		const prefix = "rill_usr_"
+		if strings.HasPrefix(input, prefix) {
+			base62 := strings.TrimPrefix(input, prefix)
+			// Base62-encoded UUIDs are always 22 characters long.
+			// This allows us to match tokens by their base62 prefix.
+			if len(base62) >= 22 {
+				if uuidBytes, ok := authtoken.UnmarshalBase62(base62[:22]); ok && len(uuidBytes) == 16 {
+					uuidStr := uuid.Must(uuid.FromBytes(uuidBytes)).String()
```
begelundmuller: We don't want token parsing logic outside the `authtoken` package. Try pushing it into that package. E.g.:
```go
dbTokens, err := s.admin.DB.FindUserAuthTokens(ctx, userID, "", 1000)
...

tokens := make([]authtoken.Token{}, len(dbTokens))
for i, dbToken := range dbTokens {
  tokens[i] = ...
}

matches := authtoken.MatchByPrefix(input, tokens)
...
```

admin/server/users.go
```diff
@@ -618,3 +629,66 @@ func projInviteToPB(i *database.ProjectInviteWithRole) *adminv1.ProjectInvite {
...
+				}
+			}
+		}
+	}
+
+	if len(matches) == 1 {
+		return matches[0], nil
+	}
+	if len(matches) > 1 {
+		return nil, status.Error(codes.InvalidArgument, "multiple tokens match the given prefix (please use a more specific prefix)")
```
begelundmuller:
```suggestion
		return nil, status.Error(codes.InvalidArgument, "multiple tokens match the given prefix (please use the full ID)")
```

cli/cmd/token/list.go
```diff
@@ -34,6 +36,13 @@ func ListCmd(ch *cmdutil.Helper) *cobra.Command {
 				return nil
 			}
 
+			// If the result set is smaller than the page size, sort by creation time (newest first).
+			if uint32(len(res.Tokens)) < pageSize {
```
begelundmuller: Only if not during pagination:

```suggestion
			// If the result set is smaller than the page size, sort by creation time (newest first).
			if pageToken == "" && uint32(len(res.Tokens)) < pageSize {
```

admin/server/service.go
```diff
@@ -169,10 +171,18 @@ func (s *Server) ListServiceAuthTokens(ctx context.Context, req *adminv1.ListSer
...
+
 		dtos[i] = &adminv1.ServiceToken{
 			Id:        token.ID,
 			CreatedOn: timestamppb.New(token.CreatedOn),
 			ExpiresOn: timestamppb.New(safeTime(token.ExpiresOn)),
+			Prefix:    t.Prefix(),
```
begelundmuller: nit: `Prefix` comes after `Id` and before `CreatedOn` in the protobuf definition, so use the same ordering here

admin/database/postgres/postgres.go
```diff
@@ -1016,7 +1016,13 @@ func (c *connection) DeleteUserAuthToken(ctx context.Context, id string) error {
 }
 
 func (c *connection) DeleteExpiredUserAuthTokens(ctx context.Context, retention time.Duration) error {
-	_, err := c.getDB(ctx).ExecContext(ctx, "DELETE FROM user_auth_tokens WHERE expires_on IS NOT NULL AND expires_on + $1 < now()", retention)
+	_, err := c.getDB(ctx).ExecContext(ctx, `DELETE FROM user_auth_tokens WHERE expires_on IS NOT NULL AND expires_on + $1 < now() AND created_on < (now() - $1)`, retention)
```
begelundmuller: Did you change the wrong function? The `DeleteExpiredUserAuthTokens` function is not related to this PR.

admin/database/postgres/postgres.go
begelundmuller: The comment isn't addressed here – `used_on IS NULL` doesn't work since it has a `NOT NULL` constraint (checking `created_on` like in `DeleteInactiveUserAuthTokens` is good)

admin/pkg/authtoken/authtoken.go
```diff
@@ -131,6 +140,30 @@ func (t *Token) SecretHash() []byte {
...
 }
 
+// Prefix returns a safe, partial display string for the token, e.g., "rill_usr_abcdefghij".
+// This works even if t.Secret is empty (all zeroes), as long as t.ID is set.
+func (t *Token) Prefix() string {
+	payload := make([]byte, 40)
+	copy(payload[0:16], t.ID[:])
+
+	// The remaining 24 bytes are already zero-initialized (Go default for slices)
+	return fmt.Sprintf("%s_%s_%s", Prefix, t.Type, marshalBase62(payload)[:10])
```
begelundmuller: Consider this to avoid duplicating logic in `String()`:
```
n := len(Prefix) + len(t.Type) + 12
return t.String()[0:n]
```

admin/pkg/authtoken/authtoken.go
```diff
@@ -131,6 +140,30 @@ func (t *Token) SecretHash() []byte {
...
+}
+
+// MatchByPrefix attempts to match a decoded token by its prefix.
+func MatchByPrefix(input string, tokens []Token) []*Token {
```
begelundmuller: Inconsistent use of `[]Token` and `[]*Token` here. Since the constructors in this package return `*Token`, you should probably use `[]*Token` everywhere. 
```suggestion
func MatchByPrefix(input string, tokens []*Token) []*Token {
```

admin/server/users.go
```diff
@@ -618,3 +629,68 @@ func projInviteToPB(i *database.ProjectInviteWithRole) *adminv1.ProjectInvite {
...
+// findUserAuthTokenFuzzy attempts to find a user auth token by exact ID, full token string, or unique prefix.
+func (s *Server) findUserAuthTokenFuzzy(ctx context.Context, input string) (*database.UserAuthToken, error) {
+	claims := auth.GetClaims(ctx)
+	userID := claims.OwnerID()
+
+	// Try exact ID match
+	token, err := s.admin.DB.FindUserAuthToken(ctx, input)
+	if err != nil && status.Code(err) != codes.NotFound {
+		return nil, err
+	}
```
begelundmuller: `s.admin.DB` doesn't return errors that use `status.Code`, which is a gRPC error library and only used for server errors, not DB errors.

Can you add a test that would have caught this error?

You need to check the `database.ErrNotFound` type from the `database` package: https://github.com/rilldata/rill/blob/a8be29bd35868ba08c33364133c3a7235ec3e620/admin/database/errors.go#L6

admin/server/users.go
```diff
@@ -618,3 +629,68 @@ func projInviteToPB(i *database.ProjectInviteWithRole) *adminv1.ProjectInvite {
...
+	if err == nil {
+		return token, nil
+	}
+
+	// Try full token string
+	tokenStr, err := authtoken.FromString(input)
+	if err == nil {
+		token, err := s.admin.DB.FindUserAuthToken(ctx, tokenStr.ID.String())
+		if err != nil && status.Code(err) != codes.NotFound {
```
begelundmuller: Same coment

admin/server/users.go
```diff
@@ -618,3 +629,68 @@ func projInviteToPB(i *database.ProjectInviteWithRole) *adminv1.ProjectInvite {
...
+	}
+
+	// Validate input length and prefix
+	if len(input) < 10 || !strings.HasPrefix(input, "rill_usr_") {
+		return nil, status.Error(codes.InvalidArgument, "invalid token ID (must be at least 10 characters and start with 'rill_usr_')")
+	}
+
+	// Find all tokens for the user and match by prefix
+	dbTokens, err := s.admin.DB.FindUserAuthTokens(ctx, userID, "", 1000)
+	if err != nil && status.Code(err) != codes.NotFound {
```
begelundmuller: Same comment

admin/server/users.go
```diff
@@ -618,3 +630,68 @@ func projInviteToPB(i *database.ProjectInviteWithRole) *adminv1.ProjectInvite {
...
+	}
+
+	// Validate input length and prefix
+	if len(input) < 10 || !strings.HasPrefix(input, "rill_usr_") {
+		return nil, status.Error(codes.InvalidArgument, "invalid token ID (must be at least 10 characters and start with 'rill_usr_')")
+	}
+
+	// Find all tokens for the user and match by prefix
+	dbTokens, err := s.admin.DB.FindUserAuthTokens(ctx, userID, "", 1000)
+	if err != nil && !errors.Is(err, database.ErrNotFound) {
```
begelundmuller: This error doesn't happen for listings (not found only happens for single lookups)

```suggestion
	dbTokens, err := s.admin.DB.FindUserAuthTokens(ctx, userID, "", 1000)
	if err != nil {
```

admin/server/users.go
```diff
@@ -618,3 +630,68 @@ func projInviteToPB(i *database.ProjectInviteWithRole) *adminv1.ProjectInvite {
...
+		return nil, err
+	}
+
+	tokens := make([]*authtoken.Token, len(dbTokens))
+	for i, dbToken := range dbTokens {
+		id, err := uuid.Parse(dbToken.ID)
+		if err != nil {
+			continue // skip invalid UUIDs
```
begelundmuller: It'd be an internal error so might as well report it:

```suggestion
		id, err := uuid.Parse(dbToken.ID)
		if err != nil {
			return nil, status.Errorf(codes.Internal, "invalid token ID %q: %v", t.ID, err)
```

PR #7390, opened 2025-05-30
Add event truncation for oversized telemetry data

grahamplata:
https://linear.app/rilldata/issue/ENG-758/truncate-large-events-before-emitting-metrics
Fixes events are too large errors

**Checklist:**
- [x] Covered by tests - added test cases.
- [x] I'm proud of this work!

### DIFF
--- a/runtime/pkg/activity/client.go
+++ b/runtime/pkg/activity/client.go
@@ -121,7 +121,7 @@ func (c *Client) RecordMetric(ctx context.Context, name string, value float64, a
 	})
 }
 
-// EmitBehavioral sends a telemetry event of type "behavioral" with the provided name and attributes.
+// RecordBehavioralLegacy sends a telemetry event of type "behavioral" with the provided name and attributes.
 // The event additionally has all the attributes associated with out legacy behavioral events.
 // It will panic if all of WithServiceName, WithServiceVersion, WithInstallID and WithUserID have not been called on the client.
 func (c *Client) RecordBehavioralLegacy(name string, extraAttrs ...attribute.KeyValue) {
@@ -246,25 +246,6 @@ func (c *Client) RecordRaw(data map[string]any) error {
 	return nil
 }
 
-func truncateEvent(e *Event) {
-	if e == nil || e.Data == nil {
-		return
-	}
-
-	b, err := json.Marshal(e)
-	if err != nil {
-		return
-	}
-	if len(b) <= maxSize {
-		return
-	}
-
-	e.Data = map[string]any{
-		"truncated": true,
-		"reason":    "event data exceeded 1MB and was truncated",
-	}
-}
-
 // emitRaw sends an event to the sink.
 func (c *Client) emitRaw(e Event) {
 	truncateEvent(&e)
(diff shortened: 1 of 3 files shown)


runtime/pkg/activity/client.go
```diff
@@ -245,8 +248,28 @@ func (c *Client) RecordRaw(data map[string]any) error {
 	return nil
 }
 
+func truncateEvent(e *Event) {
```
begelundmuller: Util functions should come at the end of the file: https://github.com/uber-go/guide/blob/master/style.md#function-grouping-and-ordering

runtime/pkg/activity/client.go
```diff
@@ -245,8 +248,28 @@ func (c *Client) RecordRaw(data map[string]any) error {
...
+		return
+	}
+
+	b, err := json.Marshal(e)
+	if err != nil {
+		return
+	}
+	if len(b) <= maxSize {
+		return
+	}
```
begelundmuller: This means it'll be doing two `json.Marshal` for every event being emitted – since this is very hot path (sometimes 100s or 1000s of events per second), I don't think that's a good idea.

Two things you can consider (maybe both):
1. Find the source of the large event you saw, and fix it so it doesn't output so much data
2. Apply this check in `Event.MarshalJSON`, which the sink calls, then it'll only be done once

grahamplata: I may have to wait until it happens again to get an accurate source of the event. I moved the validation into Event.MarshalJSON and added a test

runtime/pkg/activity/client.go
```diff
@@ -245,8 +246,28 @@ func (c *Client) RecordRaw(data map[string]any) error {
 	return nil
 }
 
+func truncateEvent(e *Event) {
```
begelundmuller: > Util functions should come at the end of the file: https://github.com/uber-go/guide/blob/master/style.md#function-grouping-and-ordering

This comment is still not addressed

runtime/pkg/activity/event.go
```diff
@@ -16,17 +18,33 @@ type Event struct {
 }
 
 func (e Event) MarshalJSON() ([]byte, error) {
-	// Add the common fields to the map.
-	if e.Data == nil {
-		e.Data = make(map[string]any)
+	data := make(map[string]any, len(e.Data)+4)
+	for k, v := range e.Data {
+		data[k] = v
 	}
```
begelundmuller: It should reuse `e.Data` if its already allocated as it did before. We emit a lot of events, so we need to be mindful of allocations.

runtime/pkg/activity/event.go
```diff
@@ -5,6 +5,8 @@ import (
 	"time"
 )
 
+const maxSize = 1048576 // 1MB in bytes
```
begelundmuller: I'd make it much smaller – maybe 32 KB?

runtime/pkg/activity/event.go
```diff
@@ -20,13 +22,30 @@ func (e Event) MarshalJSON() ([]byte, error) {
...
+	}
+
+	// Truncate if too large
+	truncated := map[string]any{
+		"event_id":   e.EventID,
+		"event_time": e.EventTime.UTC().Format(time.RFC3339Nano),
+		"event_type": e.EventType,
+		"event_name": e.EventName,
+		"truncated":  true,
+		"reason":     "event data exceeded 1MB and was truncated",
```
begelundmuller: Number doesn't match the constant

runtime/pkg/activity/client.go
```diff
@@ -247,6 +248,7 @@ func (c *Client) RecordRaw(data map[string]any) error {
 
 // emitRaw sends an event to the sink.
 func (c *Client) emitRaw(e Event) {
+	truncateEvent(&e)
```
begelundmuller: Why is this still needed here if it does truncation in `Event.MarshalJSON`?

begelundmuller (approved):

PR #7392, opened 2025-05-30
feat: add canvas and api to defaults

mindspank:
Adds canvas and api resource types to rill.yaml defaults

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/runtime/parser/parse_rillyaml.go
+++ b/runtime/parser/parse_rillyaml.go
@@ -90,7 +90,7 @@ type rillYAML struct {
 	// Default YAML values for canvases
 	Canvases yaml.Node `yaml:"canvases"`
 	// Default YAML values for custom apis
-	CustomAPIs yaml.Node `yaml:"custom_apis"`
+	APIs yaml.Node `yaml:"apis"`
 	// Default YAML values for migrations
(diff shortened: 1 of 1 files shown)


runtime/parser/parse_rillyaml.go
```diff
@@ -87,6 +87,10 @@ type rillYAML struct {
 	MetricsViewsLegacy yaml.Node `yaml:"dashboards"`
 	// Default YAML values for explores
 	Explores yaml.Node `yaml:"explores"`
+	// Default YAML values for canvases
+	Canvases yaml.Node `yaml:"canvases"`
+	// Default YAML values for custom apis
+	CustomAPIs yaml.Node `yaml:"custom_apis"`
```
begelundmuller: Can we call it `apis:`? Also a little ugly, but since the resource type is `type: api`, not `type: custom_api` I think it's more consistent.

mindspank: Sure thing, and I agree. We will get a miss match between docs and code but maybe let's correct docs instead

ericokuma:
Fixes ENG-770

PR #7399, opened 2025-06-02
Add support for metadata in metrics SQL

grahamplata:
The `QueryResolver` API now returns a `meta` field that contains metadata about the referenced dimensions and measures when the resolver is `metrics_sql` or `metrics`.

```json
{
    "meta": {
        "fields": [
            {
                "type": "measure",
                "name": "total_requests",
                "display_name": "Total Requests",
                "description": "",
                "format_preset": "humanize",
                "format_d3": null,
                "format_d3_locale": {},
                "valid_percent_of_total": false,
                "treat_nulls_as": null
            },
            ...
        ]
    },
    "schema": [
        {
            "name": "avg_bid_floor",
            "type": "CODE_FLOAT64",
            "nullable": true
        },
        ...
    ],
    "data": [
        ...
    ]
}

```

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

Closes https://linear.app/rilldata/issue/PLAT-24/platform-support-for-metadata-in-metrics-sql


begelundmuller (changes requested):
This PR returns metadata from the metrics APIs, but the goal is to return metadata for metrics SQL, not the metrics APIs. Metrics SQL has its entrypoint in `runtime/resolvers/metrics_sql.go`. See details in the issue: https://linear.app/rilldata/issue/PLAT-24/platform-support-for-metadata-in-metrics-sql

runtime/resolver.go
```diff
@@ -290,17 +302,19 @@ func (r *driverResolverResult) MarshalJSON() ([]byte, error) {
...
 		return nil, r.rows.Err()
 	}
-	if out == nil { // fixes 'null' output when there are no rows
+	if out == nil {
 		out = []map[string]any{}
 	}
+	if r.meta != nil {
+		return json.Marshal(map[string]any{"data": out, "meta": r.meta})
+	}
 	return json.Marshal(out)
```
begelundmuller: `MarshalJSON` is supposed to always return a JSON array (equivalent to paging through rows individually with `Next` and then serializing with JSON).

You should add a new `Meta` function on the `ResolverResult` interface for returning it instead.

runtime/resolvers/metrics.go
```diff
@@ -142,9 +146,42 @@ func (r *metricsResolver) ResolveInteractive(ctx context.Context) (runtime.Resol
...
+
 	return runtime.NewDriverResolverResult(res), nil
 }
 
 func (r *metricsResolver) ResolveExport(ctx context.Context, w io.Writer, opts *runtime.ResolverExportOptions) error {
 	return errors.New("not implemented")
 }
+
+// buildMetricsViewMeta builds metadata for the measures and dimensions defined in the MetricsViewSpec.
+func buildMetricsViewMeta(spec *runtimev1.MetricsViewSpec) []map[string]any {
```
begelundmuller: This just returns a description of all the fields in the metrics view. It needs to return a description of only the fields in the result (i.e. the fields selected in `r.query` and returned from `r.executor.Query`).

begelundmuller (changes requested):

runtime/resolver.go
```diff
@@ -304,6 +307,34 @@ func (r *driverResolverResult) MarshalJSON() ([]byte, error) {
...
+		name, ok := m["name"].(string)
+		if !ok {
+			continue
+		}
+		if _, exists := fieldNames[name]; exists {
+			filtered = append(filtered, m)
+		}
+	}
+	return filtered
+}
```
begelundmuller: Not sure I follow this, seems like this is some logic specific to the metrics resolver? It should just return `r.meta` I think, and any logic to transform/enrich the `meta` object should be done in the specific resolver implementation before calling `NewDriverResolverResult`.

runtime/resolver.go
```diff
@@ -345,6 +377,31 @@ func (r *mapsResolverResult) MarshalJSON() ([]byte, error) {
 	return json.Marshal(r.rows)
 }
 
+// Meta implements ResolverResult.
+func (r *mapsResolverResult) Meta() []map[string]any {
```
begelundmuller: Same comment

runtime/resolver.go
```diff
@@ -366,6 +424,7 @@ type cachedResolverResult struct {
 
 	// Iterator fields. Should only be populated on short-lived copies obtained via copy().
 	rows []map[string]any
+	meta []map[string]any
```
begelundmuller: The `meta` field belongs up alongside `data` and `schema`, not here under "Iterator fields"

runtime/resolvers/metrics.go
```diff
@@ -27,6 +27,7 @@ type metricsResolver struct {
 	query          *metricsview.Query
 	args           *metricsResolverArgs
 	claims         *runtime.SecurityClaims
+	meta           []map[string]any
```
begelundmuller: This doesn't need to be a struct field. I think it can just be computed inside the `ResolveInteractive` function right before calling `runtime.NewDriverResolverResult`?

runtime/resolvers/metrics.go
```diff
@@ -142,9 +146,37 @@ func (r *metricsResolver) ResolveInteractive(ctx context.Context) (runtime.Resol
...
 }
+
+// metaFromMetricsView extracts metadata from the metrics view.
+func metaFromMetricsView(spec *runtimev1.MetricsViewSpec) []map[string]any {
+	if spec == nil {
+		return nil
+	}
+
+	details := make([]map[string]any, 0)
+	for _, m := range spec.Measures {
```
begelundmuller: This function should take the `r.query` into account. I.e. it should only include those dimensions and measures that are present in `r.query.Dimensions` and `r.query.Measures`.

begelundmuller (changes requested):

begelundmuller (approved):

PR #7400, opened 2025-06-03
fix (Alerts UI): Handle spaces in Explore names

ericpgreen2:
Closes [ENG-784](https://linear.app/rilldata/issue/ENG-784/spaces-in-dashboard-name-causes-alert-to-not-link-back)

Note: I've checked Scheduled Reports, and we do not have this same issue.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-admin/src/features/alerts/metadata/AlertMetadata.svelte
+++ b/web-admin/src/features/alerts/metadata/AlertMetadata.svelte
@@ -141,7 +141,7 @@
               </div>
             {:else}
               <a
-                href={`/${organization}/${project}/explore/${$dashboardName.data}`}
+                href={`/${organization}/${project}/explore/${encodeURIComponent($dashboardName.data)}`}
               >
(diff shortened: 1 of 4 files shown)


ericpgreen2:
> Curious why this is not an issues for reports
 
- For Alerts, we provide a `web_open_path` annotation. Confusingly, we don't directly navigate to the `web_open_path`, but instead we parse it for the `exploreName`, which we use to compose a different URL (one with URL state). The bug was that we weren't correctly parsing the `web_open_path` for the `exploreName` when the `exploreName` included spaces.
- For Reports, we no longer provide a `web_open_path` annotation. Instead, we explicitly provide the `explore` resource name. So, there's no bug with Reports, as they don’t rely on `web_open_path` at all.

It'd be a good clean-up if we used the Report approach for Alerts too.

PR #7408, opened 2025-06-04
Include version in admin service ping

NamanMahor:
Include version in admin service ping

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


admin/server/server.go
```diff
@@ -332,8 +332,14 @@ func (s *Server) AwaitServing(ctx context.Context) error {
...
 func (s *Server) Ping(ctx context.Context, req *adminv1.PingRequest) (*adminv1.PingResponse, error) {
+	var version string
+	if s.admin.VersionNumber == "" {
+		version = "unknown (built from source)"
+	} else {
+		version = fmt.Sprintf("%s (build commit: %s)", s.admin.VersionNumber, s.admin.VersionCommit)
+	}
 	resp := &adminv1.PingResponse{
-		Version: "", // TODO: Return version
+		Version: version,
```
begelundmuller: This duplicates the logic in `cli/pkg/version/version.go`. Can you change it to pass the `version.Version` struct through instead and then use `version.String()`? That's what the runtime server does: https://github.com/rilldata/rill/blob/3c92e3106f8b37b019d0e9b2cb622fe3e24269f0/runtime/server/server.go#L116

begelundmuller (changes requested):

PR #7422, opened 2025-06-05
Add additional WHERE clause support for metrics SQL resolver

grahamplata:
Enable the application to merge global filter state into the metrics SQL queries.

**Checklist:**
- [x] Covered by tests
- [x] [Linked](https://linear.app/rilldata/issue/PLAT-23/platform-support-for-merging-filters-into-metrics-sql) the issues it closes 
- [x] Checked if the docs need to be updated. Added `additional_where` entry
- [x] I'm proud of this work!

### DIFF
--- a/runtime/resolvers/metrics_sql.go
+++ b/runtime/resolvers/metrics_sql.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/mitchellh/mapstructure"
 	"github.com/rilldata/rill/runtime"
+	"github.com/rilldata/rill/runtime/metricsview"
 	metricssqlparser "github.com/rilldata/rill/runtime/pkg/metricssql"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/trace"
@@ -16,7 +17,10 @@ func init() {
 }
 
 type metricsSQLProps struct {
+	// SQL is the metrics SQL to evaluate.
 	SQL string `mapstructure:"sql"`
+	// AdditionalWhere is a filter to apply to the metrics SQL. (additional WHERE clause)
+	AdditionalWhere *metricsview.Expression `mapstructure:"additional_where"`
 }
 
 type metricsSQLArgs struct {
@@ -39,6 +43,7 @@ func newMetricsSQL(ctx context.Context, opts *runtime.ResolverOptions) (runtime.
 	span := trace.SpanFromContext(ctx)
 	if span.SpanContext().IsValid() {
 		span.SetAttributes(attribute.String("metrics_sql", props.SQL))
+		span.SetAttributes(attribute.Bool("has_additional_where", props.AdditionalWhere != nil))
 	}
 
 	instance, err := opts.Runtime.Instance(ctx, opts.InstanceID)
@@ -46,7 +51,6 @@ func newMetricsSQL(ctx context.Context, opts *runtime.ResolverOptions) (runtime.
 		return nil, err
 	}
 
-	// todo handle refs
 	props.SQL, _, err = resolveTemplate(props.SQL, opts.Args, instance, opts.Claims.UserAttributes, opts.ForExport)
 	if err != nil {
 		return nil, err
@@ -66,6 +70,21 @@ func newMetricsSQL(ctx context.Context, opts *runtime.ResolverOptions) (runtime.
 	query, err := compiler.Rewrite(ctx, props.SQL)
 	if err != nil {
 		return nil, err
+	}
+
+	// Inject the additional where clause if provided
+	if props.AdditionalWhere != nil {
+		expr := props.AdditionalWhere
+		if query.Where != nil {
+			query.Where = &metricsview.Expression{
+				Condition: &metricsview.Condition{
+					Operator:    metricsview.OperatorAnd,
+					Expressions: []*metricsview.Expression{query.Where, expr},
+				},
+			}
+		} else {
+			query.Where = expr
+		}
 	}
 
 	// Build the options for the metrics resolver
--- a/runtime/resolvers/metrics_sql_test.go
+++ b/runtime/resolvers/metrics_sql_test.go
@@ -126,3 +126,66 @@ func TestPolicyMetricsSQLAPI(t *testing.T) {
 	require.Equal(t, "msn.com", resp[0]["domain"])
 	require.Equal(t, nil, resp[0]["publisher"])
 }
+
+func TestMetricsSQLWithAdditionalWhere(t *testing.T) {
+	rt, instanceID := testruntime.NewInstanceForProject(t, "ad_bids")
+
+	testruntime.RequireParseErrors(t, rt, instanceID, nil)
+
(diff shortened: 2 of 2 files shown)


runtime/parser/parse_partial_data.go
```diff
@@ -10,13 +10,14 @@ import (
...
-	SQL            string         `yaml:"sql"`
-	MetricsSQL     string         `yaml:"metrics_sql"`
-	API            string         `yaml:"api"`
-	Args           map[string]any `yaml:"args"`
-	Glob           yaml.Node      `yaml:"glob"` // Path (string) or properties (map[string]any)
-	ResourceStatus map[string]any `yaml:"resource_status"`
+	Connector       string         `yaml:"connector"`
+	SQL             string         `yaml:"sql"`
+	MetricsSQL      string         `yaml:"metrics_sql"`
+	AdditionalWhere string         `yaml:"additional_where"` // Optional additional WHERE clause for metrics SQL
```
begelundmuller: Should not be an option in `DataYAML`, only in the `metrics_sql.go` resolver (since this is something that shouldn't be used from YAML, only in requests).

runtime/resolvers/metrics_sql.go
```diff
@@ -16,7 +17,8 @@ func init() {
 }
 
 type metricsSQLProps struct {
-	SQL string `mapstructure:"sql"`
+	SQL             string `mapstructure:"sql"`              // SQL is the metrics SQL to evaluate.
+	AdditionalWhere string `mapstructure:"additional_where"` // AdditionalWhere is a filter to apply to the metrics SQL. (additional WHERE clause)
```
begelundmuller: The frontend will want to provide this as a `metricsview.Expression` type, not a `string` (that's why they need this and can't just put it in the metrics SQL).

I think you can just use the `metricsview.Expression` type directly here, it's serializable/deserializable.

runtime/resolvers/metrics_sql.go
```diff
@@ -39,14 +41,14 @@ func newMetricsSQL(ctx context.Context, opts *runtime.ResolverOptions) (runtime.
 	span := trace.SpanFromContext(ctx)
 	if span.SpanContext().IsValid() {
 		span.SetAttributes(attribute.String("metrics_sql", props.SQL))
+		span.SetAttributes(attribute.String("metrics_sql.additional_where", props.AdditionalWhere))
```
begelundmuller:
```suggestion
		span.SetAttributes(attribute.String("additional_where", props.AdditionalWhere))
```

runtime/testruntime/testdata/ad_bids/apis/simple_mv_sql_api_global.yaml
```diff
@@ -0,0 +1,4 @@
+kind : api
+
+metrics_sql: select dom from ad_bids_metrics LIMIT 10
+additional_where: dom != 'google.com'
```
begelundmuller: We don't use the `testruntime/testdata` projects any more (coupled tests got annoying past a certain amount of tests), instead create a project directly in your test cases with `testruntime.NewInstanceWithOptions` by providing files in the options.

begelundmuller (commented):
Can you also update the `OpenAPIRequestSchema` in `runtime/resolvers/builtin_metrics_sql.go`?

runtime/resolvers/metrics_sql.go
```diff
@@ -68,6 +70,21 @@ func newMetricsSQL(ctx context.Context, opts *runtime.ResolverOptions) (runtime.
 		return nil, err
 	}
 
+	// Inject the additional where clause if provided
+	if props.AdditionalWhere.Name != "" || props.AdditionalWhere.Condition != nil {
```
begelundmuller: If you make `AdditionalWhere` a pointer (`*metricsview.Expression`), then you can just check if it is `props.AdditionalWhere != nil`. That's safer since this might miss future cases if we add more subfields to `metricsview.Expression`.

runtime/resolvers/metrics_sql.go
```diff
@@ -39,14 +41,14 @@ func newMetricsSQL(ctx context.Context, opts *runtime.ResolverOptions) (runtime.
 	span := trace.SpanFromContext(ctx)
 	if span.SpanContext().IsValid() {
 		span.SetAttributes(attribute.String("metrics_sql", props.SQL))
+		span.SetAttributes(attribute.String("additional_where", props.AdditionalWhere.Name))
```
begelundmuller: `Name` is just one of many fields in there, not really useful. Something like this might be sufficient (builds on previous comment):
```suggestion
		span.SetAttributes(attribute.String("has_additional_where", props.AdditionalWhere != nil))
```

runtime/resolvers/metrics_sql.go
```diff
@@ -16,7 +17,8 @@ func init() {
 }
 
 type metricsSQLProps struct {
-	SQL string `mapstructure:"sql"`
+	SQL             string                 `mapstructure:"sql"`              // SQL is the metrics SQL to evaluate.
+	AdditionalWhere metricsview.Expression `mapstructure:"additional_where"` // AdditionalWhere is a filter to apply to the metrics SQL. (additional WHERE clause)
```
begelundmuller: Add the docstrings above the field names instead of inline (inline is useful for short comments only)

runtime/resolvers/builtin_metrics_sql.go
```diff
@@ -16,18 +17,25 @@ func init() {
...
-			"type":"object",
+			"type": "object",
 			"properties": {
-				"sql": {"type":"string"}
+				"sql": {"type": "string"},
+				"additional_where": {
+					"type": "object",
+					"description": "Additional filter expression for the metrics SQL query. See metricsview.Expression schema.",
+					"additionalProperties": true
+				},
```
begelundmuller: It'd be great to use a actual `Expression` type here from `metricsview.QueryJSONSchema`. Maybe add a util function for it in the `openapiutil` that extracts a definition from a schema into a standalone schema. Claude should can probably generate that quickly. Something like:
```go
// runtime/metricsivew/query.go
var ExpressionJSONSchema = must(openapituil.ExtractDefAsSchema(QueryJSONSchema, "Expression"))
```

grahamplata: This is a great idea, I will add this in another effort

runtime/resolvers/builtin_metrics_sql.go
```diff
@@ -18,16 +19,22 @@ func init() {
...
 			"required":["sql"]
 		}`,
 	})
 }
 
 type builtinMetricsSQLArgs struct {
-	SQL      string `mapstructure:"sql"`
-	Priority int    `mapstructure:"priority"`
+	SQL             string                 `mapstructure:"sql"`
+	AdditionalWhere metricsview.Expression `mapstructure:"additional_where"`
```
begelundmuller: This isn't used? Also think it should be a pointer, i.e. `*metricsview.Expression`

PR #7423, opened 2025-06-06
feat: Add in-filter limits

AdityaHegde:
As a temporary fix, we are limiting users from adding filters that increase the url size a lot. The long term fix would be to keep the query in the backend.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-common/src/features/dashboards/state-managers/loaders/DashboardStateManager.svelte
+++ b/web-common/src/features/dashboards/state-managers/loaders/DashboardStateManager.svelte
@@ -92,6 +92,7 @@
     const changedDashboard =
       !from || !to || from.params?.dashboard !== to.params?.dashboard;
     // Clear out any dashboard banners
+    // Note: we still have this on top of the above reactive statement to handle cases where navigation is to a non-dashboard route.
     if (changedDashboard) {
       eventBus.emit("remove-banner", ExploreUrlLimitWarningBannerID);
     }
--- a/web-common/src/features/dashboards/url-state/url-length-limits.ts
+++ b/web-common/src/features/dashboards/url-state/url-length-limits.ts
@@ -1,7 +1,9 @@
 const HEADER_SIZE_LIMIT = 16 * 1024; // 16KB
-const URL_LENGTH_THRESHOLD = HEADER_SIZE_LIMIT * 0.9; // -10% of the actual limit. This is arbitrary
+const URL_LENGTH_THRESHOLD = HEADER_SIZE_LIMIT * 0.8; // -20% of the actual limit. This is arbitrary to include cookies and other metadata filled in by backend for tokens.
 
 export function isUrlTooLong(url: URL) {
+  // Use TextEncoder to estimate the size of the url in bytes.
+  // `encode` here returns bytes array, so its length is what we want.
   const length = new TextEncoder().encode(url.toString()).length;
   return length > URL_LENGTH_THRESHOLD;
 }


avaitla:
Won't this break users who need those filters?

AdityaHegde:
@avaitla The UI can be unpredictable when the url size is too large. We will have a more permanent solution in the future with large filter support, but this is just to prevent user from getting into bad state.

cc @mindspank

mindspank:
Shared comments in Slack otherwise LGTM as a temporary fix until we get the long term solution in place.

web-common/src/features/dashboards/filters/dimension-filters/DimensionFilter.svelte
```diff
@@ -158,6 +162,11 @@
       ? selectedValues
       : (correctedSearchResults ?? []);
 
+  $: disableApplyButton =
+    curMode !== DimensionFilterMode.Select
+      ? !enableSearchCountQuery || inListTooLong
+      : true;
```
ericpgreen2:
```suggestion
$: disableApplyButton =
  (curMode === DimensionFilterMode.Select)
    || !enableSearchCountQuery
    || inListTooLong;
```

web-common/src/features/dashboards/filters/dimension-filters/DimensionFilter.svelte
```diff
@@ -394,6 +407,10 @@
         </div>
       {:else if error}
         <div class="min-h-9 p-3 text-center text-red-600 text-xs">error</div>
+      {:else if inListTooLong}
+        <div class="min-h-9 p-3 text-center text-red-600 text-xs">
+          Selected in list is too long. Please trim some values.
```
ericpgreen2:
```suggestion
          List is too long. Please remove some values.
```

web-common/src/features/dashboards/state-managers/loaders/DashboardStateManager.svelte
```diff
@@ -59,10 +65,37 @@
     });
   }
 
+  let hasBanner = false;
```
ericpgreen2: This feels like a code smell. Can we avoid tracking banner state like this?

web-common/src/features/dashboards/state-managers/loaders/DashboardStateManager.svelte
ericpgreen2, line 1: Might another approach be a reactive statement like this:
```typescript
$: showUrlWarning = currentUrl ? isUrlTooLong(currentUrl) : false;

$: if (showUrlWarning) {
    eventBus.emit("add-banner", {
      id: ExploreUrlLimitWarningBannerID,
      priority: ExploreUrlLimitWarningBannerPriority,
      message: {
        type: "warning",
        message:
          "URL is too long. Some features like export will not work. Please remove some filters.",
        iconType: "alert",
      },
    });
  } else {
    eventBus.emit("remove-banner", ExploreUrlLimitWarningBannerID);
  }
```

web-common/src/features/dashboards/url-state/url-length-limits.ts
```diff
@@ -0,0 +1,7 @@
+const HEADER_SIZE_LIMIT = 16 * 1024; // 16KB
+const URL_LENGTH_THRESHOLD = HEADER_SIZE_LIMIT * 0.9; // -10% of the actual limit. This is arbitrary
+const HEADER_WORD_SIZE = 4; // Word size for headers is 4
```
ericpgreen2: I'm confused by this. What's this mean?

web-common/src/features/dashboards/state-managers/loaders/DashboardStateSync.ts
```diff
@@ -75,13 +82,39 @@ export class DashboardStateSync {
...
     this.unsubExploreState?.();
   }
 
+  public getUrlForExploreState(exploreState: ExploreState) {
+    const { data: validSpecData } = get(this.dataLoader.validSpecQuery);
+    const exploreSpec = validSpecData?.explore ?? {};
+    const { data: rillDefaultExploreURLParams } = get(
+      this.rillDefaultExploreURLParams,
+    );
+    if (!rillDefaultExploreURLParams) return undefined;
```
ericpgreen2: It looks like it'd clean up the code (all code that calls `getUrlForExploreState()`) if this function only returned ever returned a `URL` not `undefined`.

AdityaHegde: Dont think we could remove all instances of `!rillDefaultExploreURLParams`. There are comments in relevant methods. But I can make this return a URL for type safety here.

web-common/src/features/dashboards/url-state/url-length-limits.ts
AdityaHegde: Update to avoid referencing word size directly.

web-common/src/features/dashboards/url-state/url-length-limits.ts
```diff
@@ -0,0 +1,7 @@
+const HEADER_SIZE_LIMIT = 16 * 1024; // 16KB
+const URL_LENGTH_THRESHOLD = HEADER_SIZE_LIMIT * 0.9; // -10% of the actual limit. This is arbitrary
+
+export function isUrlTooLong(url: URL) {
+  const length = new TextEncoder().encode(url.toString()).length;
```
ericpgreen2, line 7: Nit: a comment could help explain the use of `TextEncoder`

web-common/src/features/dashboards/state-managers/loaders/DashboardStateManager.svelte
```diff
@@ -59,12 +66,37 @@
...
   });
 
+  onNavigate(({ from, to }) => {
+    const changedDashboard =
+      !from || !to || from.params?.dashboard !== to.params?.dashboard;
+    // Clear out any dashboard banners
+    if (changedDashboard) {
+      eventBus.emit("remove-banner", ExploreUrlLimitWarningBannerID);
+    }
+  });
```
ericpgreen2: Do we need this, given the reactive statement above?

web-common/src/features/dashboards/filters/Filters.svelte
```diff
@@ -412,6 +433,8 @@
                   applyDimensionInListMode(name, values)}
                 onApplyContainsMode={(searchText) =>
                   applyDimensionContainsMode(name, searchText)}
+                isUrlTooLongAfterInListFilter={(values) =>
+                  isUrlTooLongAfterInListFilter(name, values)}
```
ericpgreen2, line 437: This makes me think that we should prevent submission of the "InList" filter if it would cause the URL to be too long. Prevent the error from ever happening.

AdityaHegde: Thats what this does no? We show an error when the input in the filter dropdown would cause the url to be too long. It also prevents submitting the filter.

We still need the banner because there are other sources where this could exceed. Since this is a temporary solution adding upfront error everywhere is overkill.

web-common/src/features/dashboards/state-managers/loaders/DashboardStateManager.svelte
AdityaHegde: Good question! When the navigate is to non-dashboard route then the reactive statement doesnt trigger. But onNavigate will run, so it is a safeguard. Lemme add a comment to clalrify.

PR #7426, opened 2025-06-09
PLAT-83:  Connector reconciler should test basic credentials

NamanMahor:
Closes [PLAT-83](https://linear.app/rilldata/issue/PLAT-83/test-connection-functionality)

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!

### DIFF
--- a/runtime/reconcilers/connector.go
+++ b/runtime/reconcilers/connector.go
@@ -95,7 +95,7 @@ func (r *ConnectorReconciler) Reconcile(ctx context.Context, n *runtimev1.Resour
 	// Test the connector configuration
 	err = r.testConnector(ctx, self.Meta.Name.Name)
 	if err != nil {
-		return runtime.ReconcileResult{Err: err}
+		return runtime.ReconcileResult{Err: fmt.Errorf("validation failed: %w", err)}
 	}
 
 	t.State.SpecHash = specHash


runtime/reconcilers/connector.go
```diff
@@ -86,6 +86,21 @@ func (r *ConnectorReconciler) Reconcile(ctx context.Context, n *runtimev1.Resour
...
+		TemplatedProperties: t.Spec.TemplatedProperties,
+		Provision:           t.Spec.Provision,
+		ProvisionArgs:       t.Spec.ProvisionArgs,
+	}
+
+	// Test the connector configuration
+	err = r.testConnector(ctx, connector)
+	if err != nil {
+		return runtime.ReconcileResult{Err: err}
+	}
```
begelundmuller: I don't think we should block the call to `UpdateInstanceConnector` if connector testing fails. We just want to return a reconcile error if the ping fails. I think that will also make the code simpler because you'll be able to use the normal `AcquireHandle` function after the connector is added, i.e. around line 116 you can have `r.C.Runtime.AcquireHandle(ctx, r.C.InstanceID, self.Meta.Name.Name)` to get the connector and then return the `Ping` result as the reconcile error.

runtime/drivers/azure/azure.go
```diff
@@ -116,7 +125,23 @@ var _ drivers.Handle = &Connection{}
...
+	}
+
+	// Try to open the bucket to verify connection
+	bucket, err := c.openBucket(ctx, c.config.Bucket, false)
+	if err != nil {
+		return fmt.Errorf("failed to open bucket: %w", err)
+	}
+	defer bucket.Close()
+
+	_, err = bucket.ListObjects(ctx, "*")
```
begelundmuller: A full listing can be quite expensive – since these pings can run somewhat frequently, something more lightweight would be preferable (like listing with a limit of 1).

runtime/drivers/azure/azure.go
```diff
@@ -204,3 +229,146 @@ func (c *Connection) AsWarehouse() (drivers.Warehouse, bool) {
 func (c *Connection) AsNotifier(properties map[string]any) (drivers.Notifier, error) {
 	return nil, drivers.ErrNotNotifier
 }
+
+// newClient returns a new azure blob client.
+func (c *Connection) newClient(bucket string) (*container.Client, error) {
```
begelundmuller: Can we keep these util functions in the `object_store.go` function? They are object-store specific, so make more sense there.

runtime/drivers/gcs/gcs.go
```diff
@@ -63,6 +66,7 @@ type ConfigProperties struct {
 	// When working in s3 compatible mode
 	KeyID  string `mapstructure:"key_id"`
 	Secret string `mapstructure:"secret"`
+	Bucket string `mapstructure:"bucket"`
```
begelundmuller: It doesn't really work to require the bucket as part of the config properties because that's a source property (usually provided as part of a glob with `gs://bucket/path...`). The ping will need to check the credentials without knowing which bucket it is for. The bucket existence will be validated as part of source/model creation.

runtime/drivers/azure/azure.go
```diff
@@ -72,6 +80,7 @@ type ConfigProperties struct {
 	Key              string `mapstructure:"azure_storage_key"`
 	SASToken         string `mapstructure:"azure_storage_sas_token"`
 	ConnectionString string `mapstructure:"azure_storage_connection_string"`
+	Bucket           string `mapstructure:"azure_storage_bucket"`
```
begelundmuller: Same comment as for GCS

runtime/drivers/gcs/gcs.go
```diff
@@ -115,7 +119,23 @@ var _ drivers.Handle = &Connection{}
 
 // Ping implements drivers.Handle.
 func (c *Connection) Ping(ctx context.Context) error {
-	return drivers.ErrNotImplemented
+	// TODO: also handle if hmac key_id and secret is provided
```
begelundmuller: This is important to handle – we now recommend using HMAC instead in certain cases

runtime/drivers/gcs/object_store.go
```diff
@@ -62,17 +61,3 @@ func (c *Connection) parseBucketURL(path string) (*globutil.URL, error) {
 	}
 	return url, nil
 }
-
-func (c *Connection) openBucket(ctx context.Context, bucket string) (*blob.Bucket, error) {
```
begelundmuller: This function really belongs in `object_store.go` – buckets are specific to object stores

runtime/drivers/postgres/postgres.go
```diff
@@ -76,8 +82,25 @@ func (d driver) Open(instanceID string, config map[string]any, st *storage.Clien
...
+	dsn := conf.ResolveDSN()
+	if dsn == "" {
+		return nil, fmt.Errorf("database_url or dsn not provided")
+	}
+
+	// Open DB handle
+	db, err := sqlx.Open("pgx", dsn)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open connection: %w", err)
+	}
```
begelundmuller: Keeping a permanently open handle only for pings seems excessive. Postgres struggles with many concurrent open connections, so we should be economical with it. Better to just open the handle as part of ping until we have other use cases that require a persistently open connection.

runtime/drivers/mysql/mysql.go
```diff
@@ -70,8 +76,24 @@ func (d driver) Open(instanceID string, config map[string]any, st *storage.Clien
...
+
+	if conf.DSN == "" {
+		return nil, fmt.Errorf("dsn not provided")
+	}
+
+	// Open DB handle
+	db, err := sqlx.Open("mysql", conf.DSN)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open connection: %w", err)
+	}
```
begelundmuller: Same comment as for postgres – avoid a permanently open handle since it's just for pings.

runtime/drivers/s3/s3.go
```diff
@@ -106,6 +117,7 @@ type ConfigProperties struct {
 	SessionToken    string `mapstructure:"aws_access_token"`
 	Region          string `mapstructure:"region"`
 	Endpoint        string `mapstructure:"endpoint"`
+	Bucket          string `mapstructure:"bucket"`
```
begelundmuller: Same comment as for GCS

runtime/drivers/slack/slack.go
```diff
@@ -72,7 +72,22 @@ var _ drivers.Handle = &handle{}
 
 // Ping implements drivers.Handle.
 func (h *handle) Ping(ctx context.Context) error {
-	return drivers.ErrNotImplemented
+	if h.config.BotToken == "" {
+		return fmt.Errorf("bot token not configured")
+	}
```
begelundmuller: I think it's optional? Not required for Slack webhooks I believe. So it should just `return nil` if no token is set.

runtime/drivers/snowflake/snowflake.go
```diff
@@ -80,8 +82,18 @@ func (d driver) Open(instanceID string, config map[string]any, st *storage.Clien
...
+	}
+
+	// Open DB handle
+	db, err := sqlx.Open("snowflake", conf.DSN)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open connection: %w", err)
+	}
+
 	return &connection{
+		db:               db,
```
begelundmuller: Keeping the connection permanently open may not be desirable since it's used so rarely. Or at least also use this connection in `warehouse.go`

runtime/drivers/snowflake/snowflake.go
```diff
@@ -80,8 +82,18 @@ func (d driver) Open(instanceID string, config map[string]any, st *storage.Clien
 		return nil, err
 	}
 
-	// actual db connection is opened during query
+	if conf.DSN == "" {
+		return nil, fmt.Errorf("dsn not provided")
+	}
```
begelundmuller: Doesn't this break backwards compatibility with projects that configure the DSN in the sources (legacy way of configuring the DSN)? We definitely can't break backwards compatibility for any commonly used connectors.

It will need to gracefully handle the case, e.g. by returning early from pings if no DSN is set. 

Also please check that no other change in this PR breaks backwards compatibility for other connectors.

begelundmuller (changes requested):

runtime/drivers/azure/azure.go
NamanMahor: Instead of List using below method for lightweight. also this does not require bucket
for Azure
```
_, err = client.GetAccountInfo(ctx, nil)
```
for gcp 
```
ts := gcp.CredentialsTokenSource(creds)
_, err = ts.Token()
```

For s3
```
_, err = stsClient.GetCallerIdentityWithContext(ctx, &sts.GetCallerIdentityInput{})
```

runtime/drivers/gcs/gcs.go
NamanMahor: We currently don't have a good way to handle this without specifying a bucket. One possible approach is to use ListBuckets, but that lists all buckets.

If ListBuckets succeeds, we'll get the full list. If it fails, we can interpret the error:

AccessDenied: The credentials are valid, but the user doesn't have permission to list all buckets.

SignatureDoesNotMatch or similar: The credentials are incorrect, and the request has failed.

runtime/drivers/azure/azure.go
```diff
@@ -204,3 +219,58 @@ func (c *Connection) AsWarehouse() (drivers.Warehouse, bool) {
 func (c *Connection) AsNotifier(properties map[string]any) (drivers.Notifier, error) {
 	return nil, drivers.ErrNotNotifier
 }
+
+// newStorageClient returns a service client.
+func (c *Connection) newStorageClient() (*service.Client, error) {
```
k-anshul: There is already a `newClient` in `object_store.go`. Is it not useful ?

runtime/drivers/bigquery/bigquery.go
```diff
@@ -128,7 +128,31 @@ var _ drivers.Handle = &Connection{}
...
+
+	client, err := createClient(ctx, projectID, opts)
+	if err != nil {
+		return fmt.Errorf("failed to create client: %w", err)
+	}
+	defer client.Close()
+
+	// Run a simple query to verify connection
+	q := client.Query("SELECT 1")
+	_, err = q.Read(ctx)
```
k-anshul, line 150: In bigquery, the basic query can succeed with limited permissions however we need wider permissions for real world datasets. 
Is this sufficient to handle all ?
If not, can you think of an easy way to test for all permissions ?

runtime/drivers/gcs/gcs.go
```diff
@@ -115,7 +116,25 @@ var _ drivers.Handle = &Connection{}
...
+
+		ts := gcp.CredentialsTokenSource(creds)
+		_, err = ts.Token()
+		if err != nil {
+			return fmt.Errorf("failed to retrieve access token: %w", err)
+		}
+	}
+
+	if c.config.KeyID != "" && c.config.Secret != "" {
+		// TODO: handle in case of HMAC key secret
```
k-anshul, line 133: Does this require handling before merging ?

runtime/drivers/s3/s3.go
```diff
@@ -154,7 +155,32 @@ var _ drivers.Handle = &Connection{}
...
+		cfg = cfg.WithEndpoint(c.config.Endpoint).WithS3ForcePathStyle(true)
+	}
+
+	sess, err := session.NewSession(cfg)
+	if err != nil {
+		return fmt.Errorf("failed to create AWS session: %w", err)
+	}
+
+	stsClient := sts.New(sess)
+	_, err = stsClient.GetCallerIdentityWithContext(ctx, &sts.GetCallerIdentityInput{})
```
k-anshul, line 178: Does this work for static creds as well ?

runtime/drivers/snowflake/snowflake.go
```diff
@@ -108,7 +109,16 @@ type connection struct {
...
+	if c.configProperties.DSN == "" {
+		// backwards compatibility: return early can't ping because dsn can be define in source.
+		return nil
+	}
+	db, err := sqlx.Open("snowflake", c.configProperties.DSN)
+	if err != nil {
+		return fmt.Errorf("failed to open connection: %w", err)
+	}
+	defer db.Close()
+	return db.PingContext(ctx)
```
k-anshul, line 121: Have you confirmed that snowflake driver implements `driver.Pinger` interface ? It is not mandatory for drivers to implement that.

runtime/controller_test.go
```diff
@@ -1292,59 +1292,6 @@ func newMetricsView(name, model string, measures, dimensions []string) (*runtime
 	return metrics, metricsRes
 }
 
-func TestDedicatedConnector(t *testing.T) {
```
k-anshul, line 1295: I think this is an important test. If you think this is failing due to s3/gcs configs absent then you can make it work by setting creds from envs/ using some other connectors etc.

k-anshul (changes requested):

runtime/drivers/bigquery/bigquery.go
NamanMahor: as discussed we are only testing cred here to keep it lightweight.

runtime/drivers/gcs/gcs.go
NamanMahor: As discussed not handling it for now.

runtime/drivers/snowflake/snowflake.go
NamanMahor: Yes it does. tested it already

runtime/drivers/s3/s3.go
NamanMahor: yes i tested with access_id and secret.

runtime/controller_test.go
NamanMahor: fixed it.

runtime/drivers/azure/azure.go
NamanMahor: It is different client. but found the way to convert service.Client to container.Client. so now there is no duplicate code.

runtime/reconcilers/connector.go
```diff
@@ -92,6 +92,12 @@ func (r *ConnectorReconciler) Reconcile(ctx context.Context, n *runtimev1.Resour
 		return runtime.ReconcileResult{Err: err}
 	}
 
+	// Test the connector configuration
+	err = r.testConnector(ctx, self.Meta.Name.Name)
+	if err != nil {
+		return runtime.ReconcileResult{Err: err}
```
begelundmuller: nit: consider adding some extra context to the error here:
```suggestion
		return runtime.ReconcileResult{Err: fmt.Errorf("validation failed: %w", err)}
```

begelundmuller (approved):

PR #7432, opened 2025-06-09
Fix href hover icon in leaderboard row

lovincyrus:
INSERT DESCRIPTION HERE

Closes https://linear.app/rilldata/issue/APP-138/investigate-uri-functionality-issue-in-v64-update

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-common/src/features/dashboards/leaderboard/LeaderboardRow.svelte
+++ b/web-common/src/features/dashboards/leaderboard/LeaderboardRow.svelte
@@ -19,6 +19,8 @@
     deltaColumn,
   } from "./leaderboard-widths";
 
+  const DOMAIN_REGEX = /^[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$/;
+
   export let itemData: LeaderboardItemData;
   export let dimensionName: string;
   export let tableWidth: number;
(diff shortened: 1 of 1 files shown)


web-common/src/features/dashboards/leaderboard/LeaderboardRow.svelte
```diff
@@ -227,17 +233,18 @@
       </span>
     {/if}
 
-    {#if hovered && href}
+    <span class="external-link-wrapper">
```
AdityaHegde: Most of the time this wont be a uri. So feels wasteful to have the element around. Any reason to remove the `if` around this?

web-common/src/features/dashboards/leaderboard/LeaderboardRow.svelte
```diff
@@ -85,7 +87,11 @@
         )
       : undefined;
 
-  $: href = makeHref(uri, dimensionValue);
+  $: href = uri
+    ? makeHref(uri, dimensionValue)
+    : DOMAIN_REGEX.test(dimensionValue)
```
AdityaHegde: What is the exact requirement here? Feel like a hacky with a regex check, will also have performance impact since we are checking for every dimension values.

lovincyrus: The ask is to reinstate the external URL icon on hover when the dimension value is in URI form. Without the regex check on the dimension value, the href maker will be undefined even after passing in the URI. Anything changed on the URI recently?

https://linear.app/rilldata/issue/APP-138/investigate-uri-functionality-issue-in-v64-update

AdityaHegde (approved):

PR #7440, opened 2025-06-10
return error message in http health check

pjain1:
Fixes https://linear.app/rilldata/issue/PLAT-53/investigate-and-fix-health-check-failure-on-moloco

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/runtime/server/health.go
+++ b/runtime/server/health.go
@@ -2,10 +2,12 @@ package server
 
 import (
 	"context"
+	"errors"
 	"net"
 	"net/http"
 
 	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
+	"github.com/rilldata/rill/runtime/pkg/httputil"
 	"github.com/rilldata/rill/runtime/pkg/observability"
 	"github.com/rilldata/rill/runtime/server/auth"
 	"go.opentelemetry.io/otel/attribute"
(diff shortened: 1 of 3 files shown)


runtime/server/health.go
```diff
@@ -69,37 +71,33 @@ func (s *Server) InstanceHealth(ctx context.Context, req *runtimev1.InstanceHeal
...
 	if status.Registry != nil {
-		w.WriteHeader(http.StatusInternalServerError)
-		return
+		return httputil.Error(http.StatusInternalServerError, status.Registry)
 	}
 	for _, h := range status.InstancesHealth {
-		if h.Controller != "" || h.Repo != "" {
-			w.WriteHeader(http.StatusInternalServerError)
-			return
+		if h.Controller != "" || h.Repo != "" || h.OLAP != "" {
```
pjain1, line 95: I added `h.OLAP` here, not sure why it was not present earlier?

k-anshul: This was to handle for static runtimes where it is possible that OLAP for one instance is failing while it is up and running for other instances. Also OLAP errors were expected to be recoverable.
But if we want to change the behaviour then it is fine to also add OLAP errors here given we just run on dynamic runtimes now. In that case please change comment on line 86.

pjain1: changed comment, lets add it.

PR #7442, opened 2025-06-11
PLAT-16: partition overwrite as default incremental strategy for duckdb

k-anshul:
Ported from clickhouse

**Checklist:**
- [x] Covered by tests
- [ ] I'm proud of this work!


begelundmuller (commented):
Did you confirm this also works for warehouse and object store executors? I think there were previously some edge cases around these, but those edge cases may no longer apply after the recent refactors.

k-anshul:
> Did you confirm this also works for warehouse and object store executors? I think there were previously some edge cases around these, but those edge cases may no longer apply after the recent refactors.

I tested for GCS and BigQuery and worked fine so looks like the edge cases no longer apply :)

PR #7444, opened 2025-06-11
feat: AI chatbot v0

ericpgreen2:
This PR introduces an AI chatbot to Rill Cloud focused on data exploration.

### What's in this PR
- **Admin service**:  
  - Adds tool-calling support to the OpenAI driver
- **Runtime service**: 
  - Persists conversations in new SQLite tables (`conversations` and `messages`)
  - Adds an agentic completion loop that invokes the MCP tools and submits their results back to the LLM
  - Exposes gRPC/HTTP endpoints: `ListConversations`, `GetConversation`, and `Complete`
  - Adds a `UseAI` permission for configurable user-level access
- **Frontend**: 
  - Adds a chatbot sidebar UI with conversation history navigation

### How to enable
The chatbot is behind a feature flag:
```yaml
// rill.yaml
features:
  - chat
```

### Immediate follow-up work
- Seed conversations with page-specific context
- Stream responses to the client
- Provide Explore URLs for each `MetricsViewQuery` tool call
- Create smart (AI-generated) conversation titles
- Track token usage


proto/rill/admin/v1/ai.proto
```diff
@@ -23,5 +26,38 @@ message CompleteResponse {
 
 message CompletionMessage {
   string role = 1;
-  string data = 2;
+  repeated ContentBlock content = 2;
```
begelundmuller: Old Rill Developer versions (on local) will be expecting field 2 to be a string. Maybe keep it for backwards compatibility and move the new `content` to field 3?

admin/ai/openai.go
```diff
@@ -24,30 +27,181 @@ func NewOpenAI(apiKey string) (Client, error) {
...
+				"input": toolCall.Input.AsMap(),
+			})
+			if err == nil {
+				content += string(toolCallJSON)
+			}
+		} else if toolResult := block.GetToolResult(); toolResult != nil {
+			// Add tool results directly as text content
+			content += toolResult.Content
+		}
+	}
```
begelundmuller: Is this the usual way to do it? Just wondering if it would help to use some separators between messages, or alternatively use JSON for it all

admin/ai/openai.go
```diff
@@ -24,30 +27,181 @@ func NewOpenAI(apiKey string) (Client, error) {
...
+
+	var schemaMap map[string]interface{}
+	if err := json.Unmarshal([]byte(schemaJSON), &schemaMap); err != nil {
+		// Fallback to basic schema if parsing fails
+		return map[string]interface{}{
+			"type":       "object",
+			"properties": map[string]interface{}{},
+		}
+	}
```
begelundmuller: Would be better to return and surface the error

proto/rill/runtime/v1/api.proto
```diff
@@ -864,6 +886,91 @@ message ListNotifierConnectorsResponse {
...
+  string name = 2;
+  google.protobuf.Struct input = 3;
+}
+
+// Tool execution result
+message ToolResult {
+  string id = 1;
+  string content = 2;
+  bool is_error = 3;
+}
```
begelundmuller: Consider importing these types from `proto/rill/admin/v1/ai.proto`. Ideally we move `proto/rill/admin/v1/ai.proto` to a dedicated `proto/rill/ai/v1` package, but we don't have to do that in this PR.

runtime/completion.go
```diff
@@ -0,0 +1,512 @@
...
+}
+
+// CompleteWithTools runs a conversational AI completion with tool calling support using the provided tool service
+func (r *Runtime) CompleteWithTools(ctx context.Context, ownerID string, req *CompletionRequest, toolService ToolService) (*CompletionResult, error) {
+	// Get instance-specific logger
+	logger, err := r.InstanceLogger(ctx, req.InstanceID)
+	if err != nil {
+		r.Logger.Error("failed to get instance logger", zap.Error(err), zap.String("instance_id", req.InstanceID), observability.ZapCtx(ctx))
+		logger = r.Logger.With(zap.String("instance_id", req.InstanceID))
+	}
```
begelundmuller: Return the error instead of logging

runtime/completion.go
```diff
@@ -0,0 +1,512 @@
...
+			zap.Duration("duration", time.Since(start)),
+			observability.ZapCtx(ctx))
+	}()
+
+	// 1. Determine conversation ID (create if needed)
+	conversationID, err := r.ensureConversation(ctx, req.InstanceID, ownerID, req.ConversationID, req.Messages)
+	if err != nil {
+		logger.Error("failed to ensure conversation", zap.Error(err), observability.ZapCtx(ctx))
+		return nil, err
+	}
```
begelundmuller: No need to log efter error case, the API middleware already logs any returned errors

runtime/completion.go
```diff
@@ -0,0 +1,512 @@
...
+		zap.String("conversation_id", req.ConversationID),
+		zap.Int("message_count", len(req.Messages)),
+		observability.ZapCtx(ctx))
+
+	start := time.Now()
+	defer func() {
+		logger.Info("completed AI completion",
+			zap.Duration("duration", time.Since(start)),
+			observability.ZapCtx(ctx))
+	}()
```
begelundmuller: If you use named return values, you can check if an error was returned and adapt the log here accordingly (e.g. so it says `failed AI completion` and includes the error)

runtime/completion.go
```diff
@@ -0,0 +1,512 @@
...
+	}
+
+	// 3. Load complete conversation context from database (single DB call)
+	allMessages, err := r.loadConversationContext(ctx, req.InstanceID, conversationID)
+	if err != nil {
+		return nil, err
+	}
+
+	// 4. Execute AI completion with database-backed context
+	logger.Debug("executing AI completion", zap.Int("context_messages", len(allMessages)), observability.ZapCtx(ctx))
```
begelundmuller: Consider dropping this log, there's already one `starting AI completion` log – for detailed step by step breakdown, we have traces

runtime/completion.go
```diff
@@ -0,0 +1,512 @@
...
+}
+
+// executeAICompletion runs the AI completion loop with tool calling support
+func (r *Runtime) executeAICompletion(ctx context.Context, instanceID string, allMessages []*drivers.CompletionMessage, toolService ToolService) ([]*runtimev1.ContentBlock, error) {
+	// Get instance-specific logger
+	logger, err := r.InstanceLogger(ctx, instanceID)
+	if err != nil {
+		r.Logger.Error("failed to get instance logger in executeAICompletion", zap.Error(err), zap.String("instance_id", instanceID), observability.ZapCtx(ctx))
+		logger = r.Logger.With(zap.String("instance_id", instanceID))
+	}
```
begelundmuller: Same comment, return errors

runtime/completion.go
```diff
@@ -0,0 +1,512 @@
...
+
+	// Apply timeout
+	ctx, cancel := context.WithTimeout(ctx, aiGenerateTimeout)
+	defer cancel()
+
+	// Get available tools and convert to drivers.Tool format (once per conversation)
+	tools, err := toolService.ListTools(ctx)
+	if err != nil {
+		logger.Error("failed to list tools", zap.Error(err), observability.ZapCtx(ctx))
+		return nil, err
```
begelundmuller: Mentioned before, but generally remove error logs from cases where the error is also returned.

(Also for context, we reserve `error` level logs for internal errors that we want alerts for, not user errors. Our API middleware has mappings of known errors to different log levels, so e.g. it avoids logging context cancellation at error level.)

runtime/completion.go
```diff
@@ -0,0 +1,512 @@
...
+
+			var toolResult *runtimev1.ToolResult
+			if err != nil {
+				logger.Error("tool execution failed", zap.Error(err), zap.String("tool_name", toolCall.Name), zap.String("tool_id", toolCall.Id), observability.ZapCtx(ctx))
+				toolResult = &runtimev1.ToolResult{
+					Id:      toolCall.Id,
+					Content: fmt.Sprintf("Error executing tool %s: %v", toolCall.Name, err),
+					IsError: true,
+				}
+			} else {
```
begelundmuller: It needs to handle context errors differently – it would signal e.g. a timeout or that the user closed a tab. It should probably return early instead of adding the errors as messages and continuing to make new calls.

You can check for context errors with `errors.Is(err, ctx.Err())`

runtime/completion.go
```diff
@@ -0,0 +1,512 @@
...
+				var result string
+				if respContent, ok := resp.([]mcp.Content); ok && len(respContent) > 0 {
+					if textContent, ok := respContent[0].(mcp.TextContent); ok {
+						result = textContent.Text
+					} else {
+						result = fmt.Sprintf("%v", respContent[0])
+					}
+				} else {
+					result = fmt.Sprintf("%v", resp)
+				}
```
begelundmuller, line 308: Using `mcp` types here kind of breaks the point of a generic `ToolService` interface

runtime/completion.go
```diff
@@ -0,0 +1,512 @@
...
+	ConversationID string
+	Messages       []*runtimev1.Message
+}
+
+// Tool represents a tool that can be called by the AI
+type Tool struct {
+	Name        string
+	Description string
+	InputSchema string // TODO: Use a better type
+}
```
begelundmuller: Consider using the proto instead – will be more natural especially if we add a dedicated `ai` proto package

runtime/completion.go
```diff
@@ -0,0 +1,512 @@
...
+	InstanceID     string
+	ConversationID string // Empty string means create new conversation
+	Messages       []*runtimev1.Message
+}
+
+// CompletionResult represents the output of AI completion
+type CompletionResult struct {
+	ConversationID string
+	Messages       []*runtimev1.Message
+}
```
begelundmuller: Consider inlining these, e.g.:
```go
Complete(ctx, instanceID, conversationID, ownerID, messages) (string, []*runtimev1.Message, error)
```

Having dedicated input/output types is also fine, but then I suggest:
1. The input/output types should have the format `FunctionNameOptions` and `FunctionNameResult`
2. Removing the shorthand `Complete` function and just having `CompleteWithTools` where not all options are required
3. Covering all parameters in the options type, e.g.:
```go
type CompleteWithToolsOptions {
  // ...
}

func (r *Runtime) CompleteWithTools(ctx context.Context, opts *CompleteWithToolsOptions) (*CompleteWithToolsResult, error) {
  /// ...
}
```

runtime/completion.go
```diff
@@ -0,0 +1,512 @@
...
+	Name        string
+	Description string
+	InputSchema string // TODO: Use a better type
+}
+
+// ToolService interface for managing and executing tools - will be implemented by server layer
+type ToolService interface {
+	ListTools(ctx context.Context) ([]Tool, error)
+	ExecuteTool(ctx context.Context, toolName string, toolArgs map[string]any) (any, error)
+}
```
begelundmuller: Needing an interface here doesn't feel great – ideally we move the MCP tools into the base `runtime` package and reference it directly on e.g. `r.tools`, or we move the logic in this file into the server package.

runtime/drivers/catalog.go
```diff
@@ -50,6 +52,12 @@ type CatalogStore interface {
 
 	FindInstanceHealth(ctx context.Context, instanceID string) (*InstanceHealth, error)
 	UpsertInstanceHealth(ctx context.Context, h *InstanceHealth) error
+
+	FindConversations(ctx context.Context, ownerID string) ([]*runtimev1.Conversation, error)
+	FindConversation(ctx context.Context, conversationID string) (*runtimev1.Conversation, error)
+	InsertConversation(ctx context.Context, ownerID, title string) (string, error)
+	FindMessages(ctx context.Context, conversationID string) ([]*runtimev1.Message, error)
+	InsertMessage(ctx context.Context, conversationID, role string, content []*runtimev1.ContentBlock, parentMessageID *string) (string, error)
```
begelundmuller: - It would be good to add dedicated `Conversation` and `Message` types here instead of using the `runtimev1` types.
    - While it can generally be helpful to re-use protos in different places, we generally don't do it for database interfaces and I think we should also avoid it here.
    - At this level, it's nice to use types that more closely reflect the underlying table schemas, potentially even using `db:"field_name"` struct annotations.
- The `runtimev1.ContentBlock` type here feels a little out of place in a catalog context. For example, if someone renames a field in the API, they may not realize that now requires a migration in the DB. I'm wondering if we could either:
    - Ideally, deconstruct it into different columns (e.g. `content_text TEXT`, `content_tool_call TEXT`, `content_tool_result JSON`) and not have more than one content column populated per message?
    - Have a flattened representation of it here (not using `oneof`) that is JSON serializable, and then store it as JSON in the DB?

runtime/drivers/ai.go
```diff
@@ -2,13 +2,24 @@ package drivers
...
+// Drivers are responsible for parsing their API responses into this structure
 type CompletionMessage struct {
-	Role string
-	Data string
+	Role    string
+	Content []*runtimev1.ContentBlock
+}
+
+type AIService interface {
+	Complete(ctx context.Context, msgs []*CompletionMessage, tools []Tool) (*CompletionMessage, error)
```
begelundmuller: Conversely here, since it's essentially an API proxy, if it'd be easier to just use the proto types, I think that's fine 🙃

proto/rill/runtime/v1/api.proto
```diff
@@ -864,6 +886,91 @@ message ListNotifierConnectorsResponse {
...
+// AI
+// **********
+
+// Message exchanged in a conversation
+message Message {
+  string id = 1;
+  string role = 2;
+  repeated ContentBlock content = 3;
+  string created_on = 4;
+  string updated_on = 5;
```
begelundmuller: Use `google.protobuf.Timestamp` instead of `string`

proto/rill/runtime/v1/api.proto
```diff
@@ -864,6 +886,91 @@ message ListNotifierConnectorsResponse {
...
+  bool is_error = 3;
+}
+
+// Conversation metadata and messages
+message Conversation {
+  string id = 1;
+  string owner_id = 2;
+  string title = 3;
+  string created_on = 4;
+  string updated_on = 5;
```
begelundmuller: Use `google.protobuf.Timestamp`

runtime/drivers/sqlite/catalog.go
```diff
@@ -357,3 +362,149 @@ func (c *catalogStore) UpsertInstanceHealth(ctx context.Context, h *drivers.Inst
...
+	if err := rows.Err(); err != nil {
+		return nil, err
+	}
+
+	// Ensure we return an empty slice instead of nil
+	if result == nil {
+		result = []*runtimev1.Conversation{}
+	}
+
+	return result, nil
```
begelundmuller: Should not be needed – `nil` works similar to an empty slice (with some small exceptions that shouldn't be relied upon by upstream logic)

runtime/drivers/sqlite/catalog.go
```diff
@@ -357,3 +362,149 @@ func (c *catalogStore) UpsertInstanceHealth(ctx context.Context, h *drivers.Inst
...
+
+	conv.CreatedOn = createdOn.Format(time.RFC3339)
+	conv.UpdatedOn = updatedOn.Format(time.RFC3339)
+
+	// Fetch messages for this conversation
+	messages, err := c.FindMessages(ctx, conversationID)
+	if err != nil {
+		return nil, err
+	}
+	conv.Messages = messages
```
begelundmuller: Ideally this should happen in the upstream logic. From the DB implementation perspective, we strive for 1:1 mapping between function calls and DB queries. Then it falls on the service/API layer to piece together multiple database calls to get the data it needs.

runtime/drivers/sqlite/catalog.go
```diff
@@ -357,3 +362,149 @@ func (c *catalogStore) UpsertInstanceHealth(ctx context.Context, h *drivers.Inst
...
+	return &conv, nil
+}
+
+// InsertConversation inserts a new conversation.
+func (c *catalogStore) InsertConversation(ctx context.Context, ownerID, title string) (string, error) {
+	conversationID := uuid.NewString()
+	_, err := c.db.ExecContext(ctx, `
+        INSERT INTO conversations (instance_id, conversation_id, owner_id, title, created_on, updated_on)
+        VALUES (?, ?, ?, ?, CURRENT_TIMESTAMP, CURRENT_TIMESTAMP)
+    `, c.instanceID, conversationID, ownerID, title)
```
begelundmuller: nit: indentation here – same applies for the other multi-line SQL calls in this file

runtime/drivers/sqlite/catalog.go
```diff
@@ -357,3 +362,149 @@ func (c *catalogStore) UpsertInstanceHealth(ctx context.Context, h *drivers.Inst
...
+		var createdOn, updatedOn time.Time
+		var seqNum int
+		if err := rows.Scan(&msg.Id, &msg.Role, &contentJSONStr, &createdOn, &updatedOn, &seqNum); err != nil {
+			return nil, err
+		}
+
+		// Parse content JSON into ContentBlock array using protojson
+		if contentJSONStr != "" {
+			// Create a temporary message with the content array wrapped
+			wrappedJSON := fmt.Sprintf(`{"content": %s}`, contentJSONStr)
```
begelundmuller: Since `contentJSONStr` is valid JSON, is this needed?

runtime/server/chat.go
```diff
@@ -0,0 +1,117 @@
...
+}
+
+// Complete runs a conversational AI completion with tool calling support.
+func (s *Server) Complete(ctx context.Context, req *runtimev1.CompleteRequest) (resp *runtimev1.CompleteResponse, err error) {
+	if !auth.GetClaims(ctx).CanInstance(req.InstanceId, auth.UseAI) {
+		return nil, ErrForbidden
+	}
+
+	// Add basic validation - fail fast for invalid requests
+	if req.Messages == nil {
```
begelundmuller:
```suggestion
	if len(req.Messages) == 0 {
```

runtime/server/generate_chart.go
```diff
@@ -91,17 +112,25 @@ func (s *Server) generateRendererWithAI(ctx context.Context, instanceID, userPro
 	defer cancel()
 
 	// Call AI service to infer a metrics view YAML
-	res, err := ai.Complete(ctx, msgs)
+	res, err := ai.Complete(ctx, msgs, []drivers.Tool{})
```
begelundmuller:
```suggestion
	res, err := ai.Complete(ctx, msgs, nil)
```

runtime/server/generate_chart.go
```diff
@@ -91,17 +112,25 @@ func (s *Server) generateRendererWithAI(ctx context.Context, instanceID, userPro
...
 	if err != nil {
 		return "", nil, err
 	}
 
+	// Extract text from content blocks
+	var responseText string
+	for _, block := range res.Content {
+		if text := block.GetText(); text != "" {
```
begelundmuller: Could it possibly return some other content type (even if we don't want it to)? If so, how can we handle it?

runtime/server/mcp_client.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/mark3labs/mcp-go/mcp"
+	"github.com/mark3labs/mcp-go/server"
+	"github.com/rilldata/rill/runtime"
+)
+
+func (s *Server) newMCPClient(mcpServer *server.MCPServer) *client.Client {
+	client, err := client.NewInProcessClient(mcpServer)
+	if err != nil {
+		panic(err)
+	}
```
begelundmuller: Return the error instead (panics generally should not be used) – applies a few places in this file, some of them it might just be a `ctx` error like timeout.

runtime/server/mcp_server.go
```diff
@@ -36,7 +36,7 @@ In the workflow, do not proceed with the next step until the previous step has b
 If a response contains an "ai_instructions" field, you should interpret it as additional instructions for how to behave in subsequent responses that relate to that tool call.
 `
 
-func (s *Server) newMCPHandler() http.Handler {
+func (s *Server) createMCPServer() *server.MCPServer {
```
begelundmuller: nit: `newMCPServer` – it's just convention, but "new" implies creating an in-memory object, "create" implies creating a persistent (DB-backed) object

runtime/server/mcp_server.go
```diff
@@ -63,6 +63,10 @@ func (s *Server) newMCPHandler() http.Handler {
 	mcpServer.AddTool(s.mcpQueryMetricsViewTimeRange())
 	mcpServer.AddTool(s.mcpQueryMetricsView())
 
+	return mcpServer
+}
+
+func (s *Server) createMCPHTTPHandler(mcpServer *server.MCPServer) http.Handler {
```
begelundmuller: nit: `newMCPHTTPHandler`

admin/ai/openai.go
ericpgreen2: > Is this the usual way to do it? 

No... 

I've updated to follow the OpenAI SDK's [completion-with-tool example](https://github.com/sashabaranov/go-openai/blob/master/examples/completion-with-tool/main.go)

proto/rill/runtime/v1/api.proto
ericpgreen2: Moved to a dedicated `proto/rill/ai/v1` package.

runtime/completion.go
ericpgreen2: Yeah, it felt wrong! This should be fixed in the refactor that brings the core of the MCP server into the `runtime` package.

runtime/completion.go
ericpgreen2: I went with the dedicated input/output types

runtime/completion.go
ericpgreen2: Yeah, as discussed, we'll move the MCP tools into the base `runtime` package in a follow-up PR

runtime/drivers/catalog.go
ericpgreen2: How's this?

```golang
FindConversations(ctx context.Context, ownerID string) ([]*Conversation, error)
FindConversation(ctx context.Context, conversationID string) (*Conversation, error)
InsertConversation(ctx context.Context, ownerID, title string) (string, error)
FindMessages(ctx context.Context, conversationID string) ([]*Message, error)
InsertMessage(ctx context.Context, conversationID, role string, content []MessageContent) (string, error)

// Conversation represents a conversation entity in the catalog
type Conversation struct {
	ID        string    `db:"conversation_id"`
	OwnerID   string    `db:"owner_id"`
	Title     string    `db:"title"`
	CreatedOn time.Time `db:"created_on"`
	UpdatedOn time.Time `db:"updated_on"`
}

// Message represents a message entity in the catalog
type Message struct {
	ID             string    `db:"message_id"`
	ConversationID string    `db:"conversation_id"`
	SeqNum         int       `db:"seq_num"`
	Role           string    `db:"role"`
	ContentJSON    []byte    `db:"content_json"`
	CreatedOn      time.Time `db:"created_on"`
	UpdatedOn      time.Time `db:"updated_on"`
}

// GetContent returns the parsed message content
func (m *Message) GetContent() ([]MessageContent, error) {
	if len(m.ContentJSON) == 0 {
		return []MessageContent{}, nil
	}
	var content []MessageContent
	err := json.Unmarshal(m.ContentJSON, &content)
	return content, err
}

// SetContent sets the message content by marshaling to JSON
func (m *Message) SetContent(content []MessageContent) error {
	contentJSON, err := json.Marshal(content)
	if err != nil {
		return err
	}
	m.ContentJSON = contentJSON
	return nil
}

// MessageContent represents a flattened, JSON-serializable content block
type MessageContent struct {
	// Content type indicator
	Type string `json:"type"` // "text", "tool_call", "tool_result"

	// Text content (when Type == "text")
	Text string `json:"text,omitempty"`

	// Tool call content (when Type == "tool_call")
	ToolCallID    string                 `json:"tool_call_id,omitempty"`
	ToolCallName  string                 `json:"tool_call_name,omitempty"`
	ToolCallInput map[string]interface{} `json:"tool_call_input,omitempty"`

	// Tool result content (when Type == "tool_result")
	ToolResultID      string `json:"tool_result_id,omitempty"`
	ToolResultContent string `json:"tool_result_content,omitempty"`
	ToolResultIsError bool   `json:"tool_result_is_error,omitempty"`
}

```

runtime/drivers/ai.go
ericpgreen2: Yeah, using the `proto/rill/ai/v1` types really simplified this

runtime/server/generate_chart.go
ericpgreen2: Given we're not calling the AI service with any tools, I _hope_ it won't return another content type (e.g. Tool Call or Tool Result). But I've added error handling in case it does.

runtime/drivers/sqlite/migrations/0032.sql
```diff
@@ -0,0 +1,24 @@
+CREATE TABLE IF NOT EXISTS conversations (
```
k-anshul, line 1: Also add an index on `instance_id, owner_id`

runtime/drivers/catalog.go
```diff
@@ -97,3 +104,62 @@ type InstanceHealth struct {
 	HealthJSON []byte    `db:"health_json"`
 	UpdatedOn  time.Time `db:"updated_on"`
 }
+
+// Conversation represents a conversation entity in the catalog
+type Conversation struct {
+	ID        string    `db:"conversation_id"`
```
k-anshul, line 110: Nit: Just `id` ? `conversation` prefix is implicit from the table name ?

runtime/drivers/catalog.go
```diff
@@ -97,3 +104,62 @@ type InstanceHealth struct {
...
+	OwnerID   string    `db:"owner_id"`
+	Title     string    `db:"title"`
+	CreatedOn time.Time `db:"created_on"`
+	UpdatedOn time.Time `db:"updated_on"`
+}
+
+// Message represents a message entity in the catalog
+type Message struct {
+	ID             string    `db:"message_id"`
```
k-anshul, line 119: Nit: Same as above. `message` is implicit from the table name ?

runtime/drivers/catalog.go
```diff
@@ -97,3 +104,62 @@ type InstanceHealth struct {
...
+	Role           string    `db:"role"`
+	ContentJSON    []byte    `db:"content_json"`
+	CreatedOn      time.Time `db:"created_on"`
+	UpdatedOn      time.Time `db:"updated_on"`
+}
+
+// GetContent returns the parsed message content
+func (m *Message) GetContent() ([]MessageContent, error) {
+	if len(m.ContentJSON) == 0 {
+		return []MessageContent{}, nil
```
k-anshul:
```suggestion
		return nil, nil
```

k-anshul: Can iterate over nil slice.

runtime/drivers/sqlite/catalog.go
```diff
@@ -357,3 +359,106 @@ func (c *catalogStore) UpsertInstanceHealth(ctx context.Context, h *drivers.Inst
...
+	msg := &drivers.Message{
+		ID:             messageID,
+		ConversationID: conversationID,
+		Role:           role,
+	}
+	if err := msg.SetContent(content); err != nil {
+		return "", fmt.Errorf("failed to marshal content: %w", err)
+	}
+
+	// Auto-calculate seq_num using a subquery - this is atomic and race-condition safe
```
k-anshul, line 456: Curious, why not directly use `auto_increment` on `seq_num`

k-anshul (changes requested):
I am curious why aren't we using the official SDK : https://github.com/openai/openai-go
Maybe the unofficial SDK became popular because the official SDK did not exist previously.

runtime/drivers/catalog.go
ericpgreen2: It's more clear in the migrations file, but IMO the prefix helps disambiguate between the other kinds of IDs (`instance_id`, `owner_id`) in the database table:

image

runtime/drivers/catalog.go
ericpgreen2: Same as above

runtime/drivers/sqlite/catalog.go
ericpgreen2: Because that would lead to global sequence numbers, but this leads to conversation-specific sequence numbers. Maybe what I have is over-engineered? Happy to do whatever is conventional.

ericpgreen2:
> I am curious why aren't we using the official SDK : https://github.com/openai/openai-go
> Maybe the unofficial SDK became popular because the official SDK did not exist previously.

Yeah, we originally adopted the 3rd-party SDK before the original was released. I haven't done a feature deep-dive, but it still seems like the 3rd-party SDK is most prominent. I think we can defer the decision to migrate for now.

image

k-anshul (approved):

PR #7453, opened 2025-06-13
emit rcus for clickhouse

pjain1:
Part of https://linear.app/rilldata/issue/PLAT-11/setup-and-audit-billing-metrics-for-druid-and-clickhouse

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/runtime/drivers/clickhouse/clickhouse.go
+++ b/runtime/drivers/clickhouse/clickhouse.go
@@ -329,7 +329,7 @@ func (d driver) Open(instanceID string, config map[string]any, st *storage.Clien
 	}
 
 	c.used()
-	go c.periodicallyEmitStats(time.Minute)
+	go c.periodicallyEmitStats(time.Minute, 10*time.Minute)
 
 	return c, nil
 }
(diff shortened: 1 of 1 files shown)


runtime/drivers/clickhouse/clickhouse.go
```diff
@@ -507,39 +507,70 @@ func (c *Connection) lastUsedOn() time.Time {
...
+	// Emit non-sensitive stats periodically
+	go func() {
+		for {
+			select {
+			case <-regularTicker.C:
+				// Emit the latest RCU per service.
+				latestRCU, err := c.latestRCUPerService(c.ctx)
+				if err == nil {
+					for service, value := range latestRCU {
+						c.activity.RecordMetric(c.ctx, "clickhouse_rcu", value, attribute.String("service", service))
```
pjain1: I believe this adds org and project metadata to the event

pjain1: checked, it will come from instance annotation

runtime/drivers/clickhouse/clickhouse.go
```diff
@@ -552,3 +583,33 @@ func (c *Connection) estimateSize(ctx context.Context) (int64, error) {
...
 	return size, nil
 }
+
+// latestRCUPerService returns the sum latest RCU value reported for nodes in each service i.e. read/write.
+func (c *Connection) latestRCUPerService(ctx context.Context) (map[string]float64, error) {
+	var query string
+	if c.config.Cluster == "" {
+		query = "SELECT service as billing_service, anyLast(value) as latest_value from billing.events WHERE event_name = 'rcu' GROUP BY service"
+	} else {
+		query = fmt.Sprintf(`SELECT service as billing_service, sum(value) AS latest_value FROM (SELECT service, anyLast(value) as value FROM clusterAllReplicas('%s', billing.events) WHERE event_name = 'rcu' GROUP BY hostName(), service) GROUP BY billing_service`, c.config.Cluster)
```
begelundmuller, line 600: Is `billing.events` a Clickhouse system table? This code also runs for self-managed Clickhouses

runtime/drivers/clickhouse/clickhouse.go
```diff
@@ -507,39 +507,70 @@ func (c *Connection) lastUsedOn() time.Time {
...
-				c.activity.RecordMetric(c.ctx, "clickhouse_estimated_size_bytes", float64(size))
-			} else if !errors.Is(err, c.ctx.Err()) {
-				lvl := zap.WarnLevel
-				if c.config.Managed {
-					lvl = zap.ErrorLevel
+		}
+	}()
+
+	// Emit sensitive stats periodically
+	go func() {
```
begelundmuller: It can use the same loop and just have multiple `case` statements in the `select` for the different tickers. The call to `periodicallyEmitStats` is already running in the background (invoked with `go`), so ideally it should avoid internally creating more background tasks

runtime/drivers/clickhouse/clickhouse.go
```diff
@@ -507,39 +507,70 @@ func (c *Connection) lastUsedOn() time.Time {
 }
 
 // Periodically collects stats about the database and emit them as activity events.
-func (c *Connection) periodicallyEmitStats(d time.Duration) {
+func (c *Connection) periodicallyEmitStats(sensitive, regular time.Duration) {
```
begelundmuller: Now that there are multiple, maybe just drop the params here and hard-code in the loop

begelundmuller (changes requested):

runtime/drivers/clickhouse/clickhouse.go
pjain1: No its a table we create for our internal cluster, in case its not present it will just print a warning and do nothing. Should we just ignore printing error of move it to debug level?

pjain1: or another way is to have a config on connector like `rill_managed` or something like that, when set we will emit otherwise not.

pjain1: or one more way is to first do a query like `SELECT 1 FROM system.tables WHERE database = 'billing' AND name = 'events'` and then only query for rcu. Let me know what you think is best.

begelundmuller: > or one more way is to first do a query like SELECT 1 FROM system.tables WHERE database = 'billing' AND name = 'events' and then only query for rcu. Let me know what you think is best.

This sounds good to me – just make sure it only checks this initially, not every iteration (to prevent excessive queries on user-managed clusters).

PR #7243, opened 2025-05-01
feat: deploy journey org selection improvements

AdityaHegde:
Updates the deploy journey to ask for explicit org name for new users instead of autogenerating using user's email.

Note: this does not have project selection changes or the github workflow changes.

Pre-req: https://github.com/rilldata/rill/pull/7244 https://github.com/rilldata/rill/pull/7262

**Checklist:**
- [ ] Covered by tests
- [x] Ran it and it works as intended
- [x] Reviewed the diff before requesting a review
- [x] Checked for unhandled edge cases
- [x] Linked the issues it closes
- [ ] Checked if the docs need to be updated
- [ ] Intend to cherry-pick into the release branch
- [ ] I'm proud of this work!

PR #7457, opened 2025-06-16
feat: Integrate github sync API in rill-dev

AdityaHegde:
Prompts user when there are changes in the remote git repo. This works for rill-managed projects as well since it will have a hidden git repo.

Downloading the diff is pending a backend change. Will be addressed in a follow up once that is ready.

Some suggestions on testing, (Make sure to use the binary built from this branch)
1. Build a binary (`rill`) from this branch to use for rill dev, `make cli`.
2. Copy the binary to home folder, `cp ./rill ~/rill`
3. Start cloud from this branch. `~/rill devtool start cloud`
4. Download smee cli here: https://github.com/probot/smee-client . Run the smee command from slack. This is needed for git events to be sent to the local cloud instance.
5. Create a project and deploy using UI.
6. Run `~/rill project clone <project_name> --path <new_path>`. 
7. Now that there are 2 folders for the projects, you can change things in one and update. The other folder should get prompts for updates.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


AdityaHegde:
1. Couldnt reproduce. Perhaps fixing 2 fixed this as well?
2. Fixed
3. I guess this is expected? cc @k-anshul
4. Will defer to @ericokuma 
5. Seems to be the issue with `.rillcloud` folder. @k-anshul since we stopped adding the folder in clone it doesnt work if the file is not checked in. Not sure if there is a good solution here. @ericpgreen2 for now you could deploy twice before cloning a separate folder.

k-anshul:
> I guess this is expected? cc @k-anshul

Yes if the repo is not connected to git then it will report an error. Though maybe from an implementation POV this can avoid a 5xx. I will raise a fix.

> Seems to be the issue with .rillcloud folder. @k-anshul since we stopped adding the folder in clone it doesnt work if the file is not checked in. Not sure if there is a good solution here. @ericpgreen2 for now you could deploy twice before cloning a separate folder.

Yeah the problem is `.rillcloud` does not gets checked-in in the first call. I see following solutions:
1. We start looking for projects built using the github remote instead of the `.rillcloud` folder.
2. During first deploy, we do a second commit that checks-in `.rillcloud` as well. That will handle this systematically instead of relying on user to deploy twice. @AdityaHegde I can make this change if it sound acceptable.

k-anshul:
> (@ericokuma) I'm a bit confused by the terminology on the status page to "Connect to GitHub". Now, if deployed projects are indeed already backed by GitHub, then it feels this copy needs updating.

I will also defer this to EricO for final thoughts but from an user POV they are not aware that rill managed repositories are backed by Github and this is an indication for them to connect to self managed Github instead.

ericokuma:
Sorry just saw this thread. Yes, @k-anshul is right. Users should not be aware that they are connected to a rill-managed repo. "Connect to GitHub" is an indication for them to connect to self managed GitHub instead. Lovable has a similar model

AdityaHegde:
@k-anshul Yes that sounds good. I was planning to do that in the update flow PR using the name of the project. But it is a bad experience here itself. Also we should do it based on git remote instead and fallback to project name.

ericokuma:
Some UXQA feedback:

* The "Skip changes" and "Fetch and merge changes" buttons seem to be really narrow. Make the buttons more like this:



* Let's actually remove the "x" from the modals so users have to select one of the two options
* When I click on "Fetch and Merge Changes", if it's successful, let's display a success toast when the modal closes

For the project update modal that appears after user tries to update/deploy, here's what the modal should look like:



* Clicking on deploy to another project will launch the "which project would you like to overwrite page" in a new browser tab i.e. same behavior as clicking on the "deploy to another project' button in this modal

AdityaHegde:
`Deploy to another project` flow is part of https://github.com/rilldata/rill/pull/7243 . So the modal show during `Update` will not show that in this PR.

web-common/src/features/dashboards/workspace/DeployProjectCTA.svelte
```diff
@@ -83,17 +103,39 @@
...
+      eventBus.emit("notification", {
+        message:
+          "Remote project changes fetched and merged. Your changes have been stashed.",
+      });
+      return;
+    }
+
+    errorFromGitCommand = {
+      message: resp.output,
+    } as ConnectError;
```
ericpgreen2: Rather than type casting, what about creating a proper Error type, e.g. `errorFromGitCommand = new Error(resp.output)`?

AdityaHegde:
Regarding tests, these need github for all the flows. We have not figured out a good method to run automated tests for github without running into rate limiting issues. We can revisit this once have figured that out.

PR #7465, opened 2025-06-16
fix: min time grain derivation

briangregoryholmes:
Fixes an issue with appropriately determining the minimum time grain for multiple metrics views.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-common/src/features/canvas/stores/time-control.ts
+++ b/web-common/src/features/canvas/stores/time-control.ts
@@ -100,13 +100,14 @@ export class TimeControls {
         (min: V1TimeGrain, metricsView) => {
           const timeGrain = metricsView?.state?.validSpec?.smallestTimeGrain;
 
+          console.log({ timeGrain });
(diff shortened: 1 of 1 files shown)


briangregoryholmes:
I had the logic flipped, so it was pulling the smallest of the smallest time grains rather than the largest. That shouldn't have caused a direct regression, but may have had similar effects.

To sum:

The original issue was created because there could be a mismatch between two smallest_time_grain values. There should be no regression with that issue because of this PR.

This new PR was created because when all upstream metrics views didn't have a `smallest_time_grain` set, it was being set to `TIME_GRAIN_YEAR`

There's _another_ issue that should not be addressed in this PR where when you are looking at a range that is shorter than your smallest_time_grain, we don't show any options in the dropdown. This is true on Explore as well. The solutions to that should not be part of this PR.

djbarnwal (approved):

ericpgreen2:
Can/should this PR be merged @briangregoryholmes?

PR #7471, opened 2025-06-17
Redirect to signup page on deploy if not logged in

lovincyrus:
This PR adds `screen_hint=signup` parameter to loginUrl for non-logged-in users.

```
User clicks deploy → loginUrl?screen_hint=signup → adminUrl OAuth → /auth/signup → Auth0 signup page
```

Closes https://linear.app/rilldata/issue/APP-28/make-sure-that-sign-up-is-the-default-auth0-universal-login-page-when

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-integration/tests/deploy.spec.ts
+++ b/web-integration/tests/deploy.spec.ts
@@ -203,7 +203,7 @@ async function login(deployPage: Page) {
   // Login 1st to start deploy.
 
   // Fill in the email
-  const emailInput = deployPage.locator('input[name="email"]');
+  const emailInput = deployPage.locator('input[name="username"]');
   await emailInput.waitFor({ state: "visible" });
   await emailInput.click();
   await emailInput.fill(process.env.RILL_DEVTOOL_E2E_ADMIN_ACCOUNT_EMAIL);


web-common/src/features/dashboards/workspace/DeployProjectCTA.svelte
```diff
@@ -53,7 +53,9 @@
   $: deployPageUrl = `${$page.url.protocol}//${$page.url.host}/deploy`;
 
   $: if (userNotLoggedIn && $metadata.data) {
-    deployCTAUrl = `${$metadata.data.loginUrl}?redirect=${deployPageUrl}`;
+    // For first-time users, redirect to signup flow
+    // Use admin server's signup endpoint (/auth/signup) for proper signup experience
+    deployCTAUrl = `${$metadata.data.adminUrl}/auth/signup?redirect=${deployPageUrl}`;
```
lovincyrus: Curious for everyone's thoughts on this - I've appended /signup with adminUrl to this line here when user isn't logged in. Eric O and Mike requested that we land users on the sign-up page in the Rill Developer when they deploy the project.

lovincyrus: I could turn it into a condition statement to only apply adminUrl with /signup in the cloud environments and remain using loginUrl in the dev environment.

ericokuma:
So if i understand this correctly, we will re-direct all non-logged-in users to the signup page during the deploy journey regardless of whether they are first time users or returning users right? 

I think that's okay for now. According to the team, most returning users will be logged in so they won't even see the Auth0 page. 

In the near future, it might be good to build that intermediary page with both login and signup options

ericokuma (approved):
lgtm

lovincyrus:
> So if i understand this correctly, we will re-direct all non-logged-in users to the signup page during the deploy journey regardless of whether they are first time users or returning users right?

Yes!

> I think that's okay for now. According to the team, most returning users will be logged in so they won't even see the Auth0 page.
> 
> In the near future, it might be good to build that intermediary page with both login and signup options

Okay!

web-common/src/features/dashboards/workspace/DeployProjectCTA.svelte
AdityaHegde: Login doesn't actually work during deploy. Logout from everything, cli and browser, clicking deploy doesnt really redirect back with login status after logging in.

The local server needs to be hit before the admin endpoint. It needs to get the callback 1st to actually get the token as well. `loginUrl` has the local endpoint here, so we cant just use `adminUrl `. Check the handler for more info: https://github.com/rilldata/rill/blob/main/cli/pkg/local/server.go#L768

@k-anshul Any quick path here?

AdityaHegde (changes requested):
Check my other comment. This breaks login flow during deploy.

web-common/src/features/dashboards/workspace/DeployProjectCTA.svelte
```diff
@@ -53,7 +53,8 @@
   $: deployPageUrl = `${$page.url.protocol}//${$page.url.host}/deploy`;
 
   $: if (userNotLoggedIn && $metadata.data) {
-    deployCTAUrl = `${$metadata.data.loginUrl}?redirect=${deployPageUrl}`;
+    // FIXME: implement signupUrl
+    deployCTAUrl = `${$metadata.data.signupUrl}?redirect=${deployPageUrl}`;
```
lovincyrus: I'm curious if we could use a `$metadata.data.signupUrl` here

lovincyrus:
Just pushed some changes to address the comment @AdityaHegde. I reverted to loginUrl and included `screen_hint=signup` in the OAuth flow to ensure it gets forwarded through the local server → admin server → Auth0 chain properly.

I would say this is a non-breaking change. If it fails, users get the login page anyway (existing behavior).

PR #7474, opened 2025-06-18
Plat 84 list and get api for connector

NamanMahor:
Fixes [PLAT-84](https://linear.app/rilldata/issue/PLAT-84/listapi-and-getable-api)

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


runtime/drivers/admin/admin.go
```diff
@@ -103,7 +103,7 @@ func (d driver) TertiarySourceConnectors(ctx context.Context, src map[string]any
 	return nil, fmt.Errorf("not implemented")
 }
 
-type Handle struct {
+type Connection struct {
```
begelundmuller: Keep this as `Handle`. It's a more correct term and aligns with the interface name (`drivers.Handle`). If anything should be changed, it's other drivers that call it `Connection` (which is not a good word since they often wrap a connection pool, not the other way around).

If you want to make refactors like this in the future, let's discuss first and do it in a dedicated PR to keep concerns separated.

runtime/drivers/drivers.go
```diff
@@ -84,6 +87,10 @@ type Handle interface {
 	// MigrationStatus returns the handle's current and desired migration version (if applicable).
 	MigrationStatus(ctx context.Context) (current int, desired int, err error)
 
+	// InformationSchema returns an InformationSchema if the handle can serve as such, otherwise returns nil.
+	// Information schema provides metadata about existing tables in a driver.
+	InformationSchema() InformationSchema
```
begelundmuller: Should be `AsInformationSchema() (InformationSchema, bool)` to align with the other `As...` functions. The `bool` should indicate if the driver can serve as an information schema or not – e.g. many drivers like `admin`, `slack`, etc. will not be valid information schemas.

runtime/drivers/file/file.go
```diff
@@ -132,7 +132,7 @@ func parseSourceProperties(props map[string]any) (*sourceProperties, error) {
 	return conf, nil
 }
 
-type connection struct {
```
begelundmuller: If anything, call it `Handle` – but let's keep such large refactors in dedicated PRs. These kinds of large diffs makes it difficult to review the PR because it's not clear which changes are functional and which changes are simple refactors.

runtime/drivers/mysql/information_schema.go
```diff
@@ -0,0 +1,236 @@
...
+				IsDefaultDatabaseSchema: isDefaultDatabaseSchema,
+				Name:                    name,
+				View:                    tableType == "VIEW",
+				Schema:                  &runtimev1.StructType{},
+				PhysicalSizeBytes:       -1,
+			}
+			res = append(res, t)
+		}
+
+		// parse column type
```
k-anshul: Do we really need this information for drivers other than OLAP drivers ?
If no, then we can skip a lot of this schema conversions for other drivers?

runtime/drivers/redshift/information_schema.go
```diff
@@ -0,0 +1,226 @@
...
+	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
+	"github.com/rilldata/rill/runtime/drivers"
+)
+
+var errUnsupportedType = errors.New("encountered unsupported redshift type")
+
+func (c *Connection) All(ctx context.Context, like string) ([]*drivers.Table, error) {
+	var likeClause string
+	if like != "" {
+		likeClause = "AND (LOWER(T.table_name) LIKE LOWER('" + like + "') OR CONCAT(T.table_schema, '.', T.table_name) LIKE LOWER('" + like + "'))"
```
k-anshul: What about the case when the search pattern is like `catalog.schema.table` ?
Also applies to several other drivers.

runtime/drivers/bigquery/bigquery.go
```diff
@@ -167,6 +167,11 @@ func (c *Connection) Config() map[string]any {
 	return m
 }
 
+// InformationSchema implements drivers.Handle.
+func (c *Connection) AsInformationSchema() (drivers.InformationSchema, bool) {
```
k-anshul: We will need this for BigQuery as well. The dataset can be mapped to a schema, project_id to database.

k-anshul (changes requested):
Few overall comments:

1. Do we really need column type information for drivers other than OLAP drivers ? If no, then it will be better to not query for them and avoid join of system.table with system.column. This will make the queries lightweight especially for cases when there are many databases and tables.
2. Even if we need column type information, I think the mapping of db column types to runtime types should only be applicable for OLAP drivers.
3. It will be good to add tests for information_schema. Can be done in a separate PR to not block integrations

In future, for such changes, it will be good to raise separate PRs for separate drivers for faster iteration and code reviews.

runtime/drivers/bigquery/bigquery.go
NamanMahor: I will handle it in separate PR.it was not part of the requirement. these were required Snowflake, Postgres, MySQL, Athena, Redshift.

runtime/drivers/redshift/information_schema.go
NamanMahor: Currently we are only using current catalog to list table we need to handle passing catalog and schema here. we already have ticket for it. So will handle it there.

runtime/drivers/mysql/information_schema.go
NamanMahor: GetTable API was requirement in the ticket.

runtime/drivers/bigquery/api.go
```diff
@@ -40,7 +40,7 @@ func (c *Connection) ListDatasets(ctx context.Context, req *runtimev1.BigQueryLi
 	return names, nextToken, nil
 }
 
-func (c *Connection) ListTables(ctx context.Context, req *runtimev1.BigQueryListTablesRequest) ([]string, string, error) {
+func (c *Connection) ListBigQueryTables(ctx context.Context, req *runtimev1.BigQueryListTablesRequest) ([]string, string, error) {
```
k-anshul, line 43: Any particular reason for this change ? `BigQuery` is implicit from the driver name.

runtime/drivers/bigquery/bigquery.go
```diff
@@ -167,6 +167,11 @@ func (c *Connection) Config() map[string]any {
...
+	return c, true
```
k-anshul: May be return false till you actually implement in subsequent PRs? Also applies to other drivers.

runtime/drivers/information_schema.go
```diff
@@ -0,0 +1,58 @@
...
+type TableInfo struct {
+	Name string
+	View bool
+}
+
+type TableMetadata struct {
+	Schema map[string]string
+}
+
+// OlapInformationSchema contains information about existing tables in an OLAP driver.
```
k-anshul: Should this be part of `runtime/drivers/olap.go` ?

runtime/drivers/information_schema.go
```diff
@@ -0,0 +1,58 @@
...
+	// ListDatabaseSchemas returns all schemas across databases
+	ListDatabaseSchemas(ctx context.Context) ([]*DatabaseSchemaInfo, error)
+	// ListTables returns all tables in a schema.
+	ListTables(ctx context.Context, database, databaseSchema string) ([]*TableInfo, error)
+	// GetTable returns metadata about a specific table.
+	GetTable(ctx context.Context, database, databaseSchema, table string) (*TableMetadata, error)
+}
+
+// SchemaInfo represents a schema in an information schema.
+type DatabaseSchemaInfo struct {
```
k-anshul, line 17: May be drop `info` suffix? Same for `TableInfo`.

runtime/drivers/pinot/information_schema.go
```diff
@@ -0,0 +1,304 @@
...
+	"github.com/c2h5oh/datasize"
+	"github.com/jmoiron/sqlx"
+	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
+	"github.com/rilldata/rill/runtime/drivers"
+	"github.com/rilldata/rill/runtime/pkg/observability"
+	"go.uber.org/zap"
+	"golang.org/x/sync/errgroup"
+)
+
+func (c *connection) ListDatabaseSchemas(ctx context.Context) ([]*drivers.DatabaseSchemaInfo, error) {
```
k-anshul, line 21: Do you plan to actually implement for OLAP drivers ?
Do you plan to deprecate OLAPInformationSchema completely ?

k-anshul (changes requested):

runtime/drivers/bigquery/api.go
NamanMahor: Yes we have same name method in biquery/Information_schema.go

runtime/drivers/information_schema.go
NamanMahor: <olap>/information_schema.go will implement both  OlapInformationSchema and InformationSchema. so that's why moved to information_schema.go.

runtime/drivers/pinot/information_schema.go
NamanMahor: Yes will implement for Olap drivers too.
Not sure we can do that right now but Yes my idea is to use single api for all for them. Eric also want single type of api.

runtime/drivers/bigquery/bigquery.go
NamanMahor: sure.

runtime/drivers/information_schema.go
k-anshul: But given `OlapInformationSchema` is specific to OLAP it should be kept in `olap.go` ?

runtime/drivers/drivers.go
```diff
@@ -84,6 +84,10 @@ type Handle interface {
 	// MigrationStatus returns the handle's current and desired migration version (if applicable).
 	MigrationStatus(ctx context.Context) (current int, desired int, err error)
 
+	// AsInformationSchema returns a InformationSchema if the handle can serve as such, otherwise returns false.
+	// InformationSchema provides metadata about existing tables in a driver.
+	AsInformationSchema() (InformationSchema, bool)
```
begelundmuller: nit: This function should come later in the interface, i.e. notice all the functions prefixed with `As` come last in the interface. I think moving this down below `AsOLAP` might be a suitable place for this.

runtime/drivers/information_schema.go
```diff
@@ -0,0 +1,58 @@
...
+	View bool
+}
+
+type TableMetadata struct {
+	Schema map[string]string
+}
+
+// OlapInformationSchema contains information about existing tables in an OLAP driver.
+// Table lookups should be case insensitive.
+type OlapInformationSchema interface {
```
begelundmuller: 1. nit: OLAP is an abbreviation, so it should be `OLAPInformationSchema`
2. Can you think of a way to unify `InformationSchema` and `OLAPInformationSchema`? E.g. `ListTables(ctx, "", "")` could be used for listing the default database and schema. And some fields like `Schema` could be made optional and only populated by the OLAP drivers.

runtime/drivers/drivers.go
NamanMahor: done

runtime/drivers/information_schema.go
NamanMahor: 1. done
2. It's not straight forward to merge them. may be we should deprecate it. UI can single new apis for listing.

k-anshul (approved):

PR #7480, opened 2025-06-18
PLAT-50: support roles and attrs on services

pjain1:
Fixes https://linear.app/rilldata/issue/PLAT-50/support-roles-and-attributes-for-services

```
// ways to create a service with optional org, project role and attributes
rill service create test-svc  --org test --org-role viewer 
rill service create test-svc  --org test --org-role viewer --project test --project-role admin 
rill service create test-svc  --org test --org-role viewer --project test --project-role admin --attributes '{"k": "v"}'

// to set attributes on service, this overrides the previous as of now
rill service edit test-svc --attributes '{"k": "v1"}' 

// to set org role on a service
rill service set-role  
rill service set-role --project test // to add or change project role on service

// remove org level role for service
rill service remove test-svc 
// remove project level role for service
rill service remove test-svc --project test 

// list org level services
rill service list 
rill service list --project test // list project level services

// show service details including projects its part of
rill service show test-svc
```

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


admin/database/postgres/migrations/0074.sql
```diff
@@ -0,0 +1,18 @@
+CREATE TABLE service_orgs_roles (
+    service_id UUID PRIMARY KEY NOT NULL REFERENCES service (id) ON DELETE CASCADE,
+    org_id UUID NOT NULL REFERENCES orgs (id) ON DELETE CASCADE,
+    org_role_id UUID NOT NULL REFERENCES org_roles (id) ON DELETE CASCADE
+);
+
+CREATE INDEX service_orgs_roles_org_service_idx ON service_orgs_roles (org_id, service_id);
```
begelundmuller: For backwards compatibility, should there be a migration that populates this table such that all existing services get the `admin` role?

admin/permissions.go
```diff
@@ -36,8 +37,33 @@ func (s *Service) OrganizationPermissionsForUser(ctx context.Context, orgID, use
...
+			ReadProjects:     role.ReadProjects,
+			CreateProjects:   role.CreateProjects,
+			ManageProjects:   role.ManageProjects,
+			ReadOrgMembers:   role.ReadOrgMembers,
+			ManageOrgMembers: role.ManageOrgMembers,
+			ManageOrgAdmins:  role.ManageOrgAdmins,
+		}, nil
+	}
+
+	// Fall back to legacy behavior for backward compatibility
```
begelundmuller: Would it be better to do a migration? I'm thinking what if a service does not have an org role, then it might mean it should only have project access instead of meaning it should have admin access.

admin/permissions.go
```diff
@@ -167,6 +194,63 @@ func (s *Service) ProjectPermissionsForService(ctx context.Context, projectID, s
...
+		return nil, err
+	}
+
+	// Get org project belongs to
+	proj, err := s.DB.FindProject(ctx, projectID)
+	if err != nil {
+		return nil, err
+	}
+
+	// Services get read permissions on the projects of the org they belong to
```
begelundmuller: If we add a migration that makes existing services admins, then can this be removed? Since `orgPerms.ManageProjects` would get hit above?

admin/server/service.go
```diff
@@ -31,24 +34,66 @@ func (s *Server) CreateService(ctx context.Context, req *adminv1.CreateServiceRe
...
+		OrgID:      org.ID,
+		Name:       req.Name,
+		Attributes: req.Attributes,
 	})
 	if err != nil {
 		return nil, err
 	}
 
+	// If org role is specified, assign it
+	if req.OrgRoleName != "" {
```
begelundmuller, line 70: Since older CLI's won't set this and would expect the service to become an admin, should we require people to upgrade by adding this method in the `_minCliVersionByMethod` const?

admin/server/service.go
```diff
@@ -31,24 +34,66 @@ func (s *Server) CreateService(ctx context.Context, req *adminv1.CreateServiceRe
...
+			return nil, status.Error(codes.InvalidArgument, "invalid organization role")
+		}
+		err = s.admin.DB.InsertOrganizationMemberService(ctx, service.ID, org.ID, orgRole.ID)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// If project role is specified, assign it
+	if req.ProjectName != "" && req.ProjectRoleName != "" {
```
begelundmuller, line 82: 1. If a project name or role is set but not the other, I guess it should error?
2. If neither `OrgRoleName` nor `ProjectRoleName` is set, should it also error? Since it would not have any access (assuming we replace the legacy handling code with a migration)

admin/server/service.go
```diff
@@ -57,26 +102,87 @@ func (s *Server) ListServices(ctx context.Context, req *adminv1.ListServicesRequ
...
+	}
+
 	return &adminv1.ListServicesResponse{
-		Services: servicesToPB(services, req.OrganizationName),
+		Services: servicesPB,
+	}, nil
+}
+
+// ListProjectServices lists all service accounts for a project.
+func (s *Server) ListProjectServices(ctx context.Context, req *adminv1.ListProjectServicesRequest) (*adminv1.ListProjectServicesResponse, error) {
```
begelundmuller: To align with `ListProjectMemberUsers` and `ListProjectMemberUsergroups`, should we call this `ListProjectMemberServices` instead?

admin/server/service.go
```diff
@@ -57,26 +102,87 @@ func (s *Server) ListServices(ctx context.Context, req *adminv1.ListServicesRequ
...
+
+	org, err := s.admin.DB.FindOrganizationByName(ctx, req.OrganizationName)
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+
+	project, err := s.admin.DB.FindProjectByName(ctx, req.OrganizationName, req.ProjectName)
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
```
begelundmuller: Return `err` directly, then the middleware will map context errors and not found errors to the right codes automatically

admin/server/service.go
```diff
@@ -57,26 +102,87 @@ func (s *Server) ListServices(ctx context.Context, req *adminv1.ListServicesRequ
...
+
+	claims := auth.GetClaims(ctx)
+	if !claims.OrganizationPermissions(ctx, org.ID).ManageOrg {
+		return nil, status.Error(codes.PermissionDenied, "not allowed to list project services")
+	}
+
+	services, err := s.admin.DB.FindProjectMemberServices(ctx, project.ID)
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
+	}
```
begelundmuller: Return `err` directly so context cancels don't get marked as internal errors

admin/server/service.go
```diff
@@ -98,18 +204,139 @@ func (s *Server) UpdateService(ctx context.Context, req *adminv1.UpdateServiceRe
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 
-	updatedService, err := s.admin.DB.UpdateService(ctx, service.ID, &database.UpdateServiceOptions{
-		Name: valOrDefault(req.NewName, service.Name),
```
begelundmuller, line 104: Is there a reason it doesn't use `valOrDefault` for `NewName` and `Attributes` now?

cli/cmd/service/create.go
```diff
@@ -17,10 +23,39 @@ func CreateCmd(ch *cmdutil.Helper) *cobra.Command {
 				return err
 			}
 
-			res1, err := client.CreateService(cmd.Context(), &adminv1.CreateServiceRequest{
+			// Parse attributes if provided
+			var attrs map[string]string
+			if attributes != "" {
+				if err := json.Unmarshal([]byte(attributes), &attrs); err != nil {
+					return err
```
begelundmuller: nit: people will probably mess up here, so some context might be nice:
```suggestion
					return fmt.Errorf("failed to parse --attributes as JSON: %w", err)
```

cli/cmd/service/create.go
```diff
@@ -17,10 +23,39 @@ func CreateCmd(ch *cmdutil.Helper) *cobra.Command {
...
+			req := &adminv1.CreateServiceRequest{
 				Name:             args[0],
 				OrganizationName: ch.Org,
-			})
+			}
+
+			// Set org role if provided
+			if orgRole != "" {
+				req.OrgRoleName = orgRole
+			}
```
begelundmuller: If org role is not set and if project role is not set, should we error? Or prompt for role?

cli/cmd/service/delete.go
```diff
@@ -17,6 +19,20 @@ func DeleteCmd(ch *cmdutil.Helper) *cobra.Command {
...
+				// If projectName is provided, delete the service from the specified project
+				_, err = client.RemoveProjectMemberService(cmd.Context(), &adminv1.RemoveProjectMemberServiceRequest{
+					Name:             args[0],
+					OrganizationName: ch.Org,
+					ProjectName:      projectName,
+				})
+				if err != nil {
+					return err
+				}
+				ch.PrintfSuccess("Removed service: %q from project: %q\n", args[0], projectName)
```
begelundmuller: nit:
```suggestion
				ch.PrintfSuccess("Removed service %q from project %q\n", args[0], projectName)
```

cli/cmd/service/set_role.go
```diff
@@ -0,0 +1,71 @@
...
+			}
+
+			err := cmdutil.StringPromptIfEmpty(&name, "Enter service name")
+			if err != nil {
+				return err
+			}
+			err = cmdutil.SelectPromptIfEmpty(&role, "Select role", roleOptions, "")
+			if err != nil {
+				return err
+			}
```
begelundmuller: Should check `ch.Interactive`. Although I also think it's fine not to have prompting in `set-role` since it's kind of rare to use it (prompts are mainly important for common commands)

cli/cmd/service/update.go
```diff
@@ -0,0 +1,62 @@
...
+	adminv1 "github.com/rilldata/rill/proto/gen/rill/admin/v1"
+	"github.com/spf13/cobra"
+)
+
+func UpdateCmd(ch *cmdutil.Helper) *cobra.Command {
+	var newName string
+	var attributes string
+
+	updateCmd := &cobra.Command{
+		Use:   "update <service-name>",
```
begelundmuller: Call the command `edit` instead of `update` to align with other commands for updates/edits

cli/cmd/service/update.go
```diff
@@ -0,0 +1,62 @@
...
+			client, err := ch.Client()
+			if err != nil {
+				return err
+			}
+
+			// Parse attributes if provided
+			var attrs map[string]string
+			if attributes != "" {
+				if err := json.Unmarshal([]byte(attributes), &attrs); err != nil {
+					return err
```
begelundmuller: Nicer error message would also be nice here

cli/cmd/service/list.go
```diff
@@ -16,18 +17,33 @@ func ListCmd(ch *cmdutil.Helper) *cobra.Command {
...
+				})
+				if err != nil {
+					return err
+				}
+
+				ch.PrintProjectMemberServices(res.Services)
+				return nil
+			}
+
 			res, err := client.ListServices(cmd.Context(), &adminv1.ListServicesRequest{
```
begelundmuller, line 34: So with the changes inside `ListServices` this now only lists services with an org role, right? I think people might expect `rill service list` to list all services in the org, both those with an org role and those that are project-level. 

Could it maybe list all services and have a) the org role column say "none" if there isn't one, b) maybe have a bool column for "has project roles"?

cli/cmd/service/list.go
```diff
@@ -16,18 +17,33 @@ func ListCmd(ch *cmdutil.Helper) *cobra.Command {
...
 			}
 
-			ch.PrintServices(res.Services)
+			ch.PrintOrganizationMemberServices(res.Services)
 
 			return nil
 		},
 	}
 
+	listCmd.Flags().StringVar(&project, "project", "", "Project name to filter services")
```
begelundmuller, line 47: How can a user look up which projects a service has access to?

begelundmuller (changes requested):

admin/database/postgres/migrations/0074.sql
pjain1: Its not needed, code assumes if no org role set then its admin while resolving permissions etc. But for better clarity we can do that.

admin/permissions.go
pjain1: Sure can be done

admin/permissions.go
pjain1: Actually the comment is wrong `// Fall back to legacy behavior for backward compatibility` this is for handling new behaviour of project level permissions. If a service has viewer role on org and no role on project then it should still get read permission on the project, this code handles that.

admin/server/service.go
pjain1: If we remove assumption no role means admin and migrate all legacy svc to admin role explicitly then yes we will need to enforce ugprade.

admin/server/service.go
pjain1: 1. Yes we can add that
2. After migration it should error

admin/server/service.go
pjain1: it was kind of defensive if the map parameter remains `nil` instead of empty map then we can use it. Also map type cannot have optional in proto so it cannot be used with `valOrDefault` as it expects a pointer.

cli/cmd/service/create.go
pjain1: sure we can error since we are going with not assuming no role means admin role

cli/cmd/service/set_role.go
pjain1: Kept it same as user set-role will add check there as well.

cli/cmd/service/list.go
pjain1: Yeah that does not exists, we want to support that in cli? May be we need a separate cmd for it something like show which lists projects a service has access to or we add another boolean flag like `projects` if set we list projects service has access to?

admin/server/service.go
pjain1: Interesting thing is if I set `'{}'` in the cli as attributes, cmd initializes the map but in the admin service the request attribute is nil so hard to distinguish between empty and no response. I was thinking if user sets `'{}'` then we might delete the attribute. One way is to accept attrs as string in the rpc and deserialize there, other is to add another method to clear all attributes.

pjain1: may be add a field in `UpdateServiceRequest` like `bool clear_attributes` and if user sets `'{}'` then set this

admin/database/postgres/postgres.go
```diff
@@ -1028,36 +1028,89 @@ func (c *connection) DeleteInactiveUserAuthTokens(ctx context.Context, retention
...
+func (c *connection) FindOrganizationMemberServices(ctx context.Context, orgID string) ([]*database.OrganizationMemberService, error) {
+	var services []*organizationMemberServiceDTO
+	query := `
+		SELECT s.id, s.name, COALESCE(r.name, '') as role_name, EXISTS (
+           SELECT 1
+           FROM service_projects_roles spr
+           JOIN projects p ON p.id = spr.project_id
+           WHERE spr.service_id = s.id AND p.org_id = $1
+       	) AS has_project_roles, 
+		s.attributes, s.created_on, s.updated_on
```
begelundmuller: nit: indentation

admin/permissions.go
```diff
@@ -167,6 +172,62 @@ func (s *Service) ProjectPermissionsForService(ctx context.Context, projectID, s
...
+		return nil, err
+	}
+
+	proj, err := s.DB.FindProject(ctx, projectID)
+	if err != nil {
+		return nil, err
+	}
+
+	// Services get read permissions on the projects of the org they belong to
+	if proj.OrganizationID == service.OrgID {
```
begelundmuller: I think this needs to be removed now? Otherwise if it has no project roles, it would get access to all projects?

For backwards compatibility for the services with the `admin` org role, I believe the `if orgPerms.ManageProjects` case above covers them.

cli/cmd/service/create.go
```diff
@@ -17,10 +23,43 @@ func CreateCmd(ch *cmdutil.Helper) *cobra.Command {
...
+
+			req := &adminv1.CreateServiceRequest{
 				Name:             args[0],
 				OrganizationName: ch.Org,
-			})
+			}
+
+			if orgRole == "" && projectRole == "" {
+				return fmt.Errorf("either --org-role or --project-role must be specified")
+			}
```
begelundmuller: It's not ideal to require a flag (since flags are usually optional). What do you think would be a nice behavior here?

cli/cmd/service/create.go
```diff
@@ -17,10 +23,43 @@ func CreateCmd(ch *cmdutil.Helper) *cobra.Command {
...
+
+			// Set org role if provided
+			if orgRole != "" {
+				req.OrgRoleName = orgRole
+			}
+
+			// Set project role if provided
+			if projectRole != "" {
+				if projectName == "" {
+					return fmt.Errorf("project name is required when project role is set")
```
begelundmuller: Better to be clear about the flag that's missing:

```suggestion
					return fmt.Errorf("you must set --project when using --project-role")
```

cli/cmd/service/delete.go
```diff
@@ -17,6 +19,20 @@ func DeleteCmd(ch *cmdutil.Helper) *cobra.Command {
 				return err
 			}
 
+			if projectName != "" {
+				// If projectName is provided, delete the service from the specified project
+				_, err = client.RemoveProjectMemberService(cmd.Context(), &adminv1.RemoveProjectMemberServiceRequest{
```
begelundmuller: It feels a little weird that the `rill service delete` does removal when `--project` is passed. Also, is it correct that there's currently no way to remove a service's org role without deleting it?

For consistency with `rill user remove`, what do you think about adding a `rill service remove` command for removing service roles from orgs and projects, and keeping `rill service delete` only for deleting services?

cli/cmd/service/set_role.go
```diff
@@ -0,0 +1,75 @@
...
+				roleOptions = projectRoles
+			} else {
+				roleOptions = orgRoles
+			}
+
+			if ch.Interactive {
+				err := cmdutil.StringPromptIfEmpty(&name, "Enter service name")
+				if err != nil {
+					return err
+				}
```
begelundmuller, line 31: Consider making it a required positional arg for consistency with the other commands, e.g. `"edit <service-name>"` has it as a required positional arg.

cli/cmd/service/set_role.go
```diff
@@ -0,0 +1,75 @@
...
+				ch.PrintfSuccess("Updated role of service %q to %q in the organization \"%s\"\n", name, role, ch.Org)
+			}
+
+			return nil
+		},
+	}
+	setRoleCmd.Flags().SortFlags = false
+	setRoleCmd.Flags().StringVar(&projectName, "project", "", "Project")
+	setRoleCmd.Flags().StringVar(&name, "name", "", "Name of the service")
```
begelundmuller: See comment above

cli/cmd/service/create.go
```diff
@@ -33,12 +72,17 @@ func CreateCmd(ch *cmdutil.Helper) *cobra.Command {
...
-			ch.Printf("Created service %q in org %q.\n", res1.Service.Name, res1.Service.OrgName)
+			ch.Printf("Created service %q in org %q.\n", res1.Service.Name, ch.Org)
 			ch.Printf("Access token: %s\n", res2.Token)
 
 			return nil
 		},
 	}
 
+	createCmd.Flags().StringVar(&orgRole, "org-role", "", "Organization role to assign to the service")
+	createCmd.Flags().StringVar(&projectRole, "project-role", "", "Project role to assign to the service")
```
begelundmuller, line 111: It should probably list the available options

proto/rill/admin/v1/api.proto
```diff
@@ -1287,12 +1313,25 @@ message ListServicesRequest {
...
+  repeated ProjectMemberService services = 1;
 }
 
 message CreateServiceRequest {
   string name = 1 [(validate.rules).string.min_len = 1];
   string organization_name = 2 [(validate.rules).string.min_len = 1];
+  string org_role_name = 3; // Optional org role to assign
+  string project_name = 4; // Optional project to assign role to
+  string project_role_name = 5; // Optional project role to assign
+  map<string, string> attributes = 6;
```
begelundmuller: Would about non-string types, like `admin` (bool) or `groups` (array)? Consider using either `google.protobuf.Struct` passing it as JSON, e.g. `string attributes_json`.

If you use `google.protobuf.Struct`, I think you can add `optional` in front – maybe that solves the issue about ability to clear attributes?

begelundmuller (changes requested):

admin/server/service.go
begelundmuller: I addressed this in a comment below – I think try `optional google.protobuf.Struct attributes` and if that doesn't work, maybe `optional string attributes_json`?

cli/cmd/service/list.go
begelundmuller: I definitely think we need this. Agree that `rill service show` that shows info about a service including its projects sounds like a good idea.

Btw, do we also need a command for adding a service to another project? E.g. `rill service add` similar to `rill user add`?

pjain1: Added `rill service show`. `rill service set-role` adds or updates service to a project with a role so need for a dedicated add cmd.

admin/permissions.go
pjain1: > Otherwise if it has no project roles, it would get access to all projects?

Yes this is intended behaviour, if a service is added as a viewer or any other role than admin to org then it should have read permission on project. Otherwise whats the use of org level roles?

cli/cmd/service/create.go
pjain1: Should I just include a prompt saying if no orgRole or projectRole is provide then service will be created with admin role on org level otherwise use either --org-role or --project-role to set a role? Use can accept or deny the prompt.

admin/permissions.go
pjain1: Although this should change to check if any org role exists for service if not then return empty permissions

cli/cmd/service/delete.go
pjain1: Added `rill user remove` cmd

cli/cmd/service/create.go
pjain1: Another option is to to prompt for whether they want to add an org role or not and give role options to select from, similar for project name and roles. That will solve [this](https://github.com/rilldata/rill/pull/7480#discussion_r2178048137) as well.

admin/permissions.go
pjain1: made the change

proto/rill/admin/v1/api.proto
pjain1: changed to use `google.protobuf.Struct`

cli/cmd/service/create.go
pjain1: changed the flow please check

cli/cmd/service/remove.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/rilldata/rill/cli/pkg/cmdutil"
+	adminv1 "github.com/rilldata/rill/proto/gen/rill/admin/v1"
+	"github.com/spf13/cobra"
+)
+
+func RemoveCmd(ch *cmdutil.Helper) *cobra.Command {
+	var service, projectName string
+
+	removeCmd := &cobra.Command{
+		Use:   "remove",
```
pjain1: Not taking service name here as arg to keep it same as user remove cmd however it feels more consistent to take name as arg in the command

begelundmuller: I agree it would feel more consistent – I think it's okay to diverge from the `rill user` behavior here since this case is a bit different

pjain1: added

pjain1: please review again

cli/cmd/service/create.go
```diff
@@ -17,10 +24,65 @@ func CreateCmd(ch *cmdutil.Helper) *cobra.Command {
...
+				}
+				if ok {
+					err = cmdutil.SelectPromptIfEmpty(&orgRole, "Select role", orgRoles, "")
+					if err != nil {
+						return err
+					}
+				}
+			}
+
+			if ch.Interactive && projectRole == "" {
```
pjain1: This means that user will have to set `--interactive=false` if they want to use this cmd programatically as just setting only `org-role` will also lead to this prompt.

admin/database/postgres/postgres.go
```diff
@@ -1028,36 +1028,142 @@ func (c *connection) DeleteInactiveUserAuthTokens(ctx context.Context, retention
...
 
-// FindServicesByOrgID returns a list of services in an org.
-func (c *connection) FindServicesByOrgID(ctx context.Context, orgID string) ([]*database.Service, error) {
-	var res []*database.Service
+// FindOrganizationMemberServices returns a list of services in an org.
+func (c *connection) FindOrganizationMemberServices(ctx context.Context, orgID string) ([]*database.OrganizationMemberService, error) {
+	var services []*organizationMemberServiceDTO
+	query := `
+       SELECT s.id, s.name, COALESCE(r.name, '') as role_name, EXISTS (
+           SELECT 1
```
begelundmuller: nit: indentation here

admin/permissions.go
```diff
@@ -167,6 +172,51 @@ func (s *Service) ProjectPermissionsForService(ctx context.Context, projectID, s
...
+		composite := &adminv1.ProjectPermissions{}
+		for _, role := range roles {
+			composite = UnionProjectRoles(composite, role)
+		}
+		return composite, nil
+	}
+
+	// If no roles are found, check if the service has read projects permission on the org
+	if orgPerms.ReadProjects {
+		return &adminv1.ProjectPermissions{
```
begelundmuller: Actually `ReadProjects` should not give access to the actual projects, only access to listing projects. For example, the `guest` role also has `ReadProjects`, but they can only list/access the projects they are members of. See `ProjectPermissionsForUser` for comparison.

I think you can remove this case and things will work as expected:
1. If org role is `admin`, it can do everything.
2. If org role is not `admin`, it can only access org-level APIs, not project-level APIs unless it has project-level permissions.

cli/cmd/service/create.go
begelundmuller: > This means that user will have to set --interactive=false if they want to use this cmd programatically as just setting only org-role will also lead to this prompt.

I think that's fine, but if you don't think it's ideal, you can use the `Changed` function to check if `--org-role` was explicitly provided like this:
```suggestion
			if ch.Interactive && projectRole == "" && !cmd.Flags().Changed("org-role") {
```

cli/cmd/service/create.go
```diff
@@ -17,10 +24,65 @@ func CreateCmd(ch *cmdutil.Helper) *cobra.Command {
...
+				ok, err := cmdutil.ConfirmPrompt("Do you want to assign a project role to the service?", "", false)
+				if err != nil {
+					return err
+				}
+				if ok {
+					err = cmdutil.StringPromptIfEmpty(&projectName, "Enter project name")
+					if err != nil {
+						return err
+					}
```
begelundmuller: Should it maybe use a select prompt with a list of projects in the current org instead of asking the user to enter the name? Since it can be hard to remember the names

proto/rill/admin/v1/api.proto
```diff
@@ -662,13 +662,24 @@ service AdminService {
...
     option (google.api.http) = {
       post: "/v1/organizations/{organization_name}/services",
+      body: "*"
     };
   }
 
+  // ShowService returns information about a specific service
+  rpc ShowService(ShowServiceRequest) returns (ShowServiceResponse) {
+    option (google.api.http) = {get: "/v1/organizations/{organization_name}/services/{name}"};
+  }
```
begelundmuller: Should this be called `GetService` to align with other RPC names, like `GetProject` or `GetOrganization`?

proto/rill/admin/v1/api.proto
```diff
@@ -1287,28 +1324,84 @@ message ListServicesRequest {
...
 }
 
+message ShowServiceRequest {
+  string name = 1 [(validate.rules).string.min_len = 1];
+  string organization_name = 2 [(validate.rules).string.min_len = 1];
+}
+
+message ShowServiceResponse {
+  OrganizationMemberService org_service = 1;
+  repeated ProjectMemberService project_services = 2; // All projects that the service is a member of
```
begelundmuller: Not sure, but does it feel a little off to have a `_service` suffix here since its implicit from the type name and request name? Not sure about good alternatives, but maybe something like:

```suggestion
  OrganizationMemberService service = 1;
  repeated ProjectMemberService project_memberships = 2; // All projects that the service is a member of
```

begelundmuller (approved):

cli/cmd/service/create.go
pjain1: Yeah I don't like that much since we can just show limited options and user gets confused about missing projects from the list. But anyways changed.

PR #7481, opened 2025-06-18
Limit PostHog error tracking ( events) to ui.rilldata.com and localho…

ericokuma:
Limiting exception events to just ui.rilldata.com and localhost

INSERT DESCRIPTION HERE

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-common/src/lib/analytics/posthog.ts
+++ b/web-common/src/lib/analytics/posthog.ts
@@ -30,6 +30,20 @@ export function initPosthog(rillVersion: string, sessionId?: string | null) {
         "Rill version": rillVersion,
       });
     },
+    before_send: (event) => {
+      // Only filter $exception (error tracking) events by URL
+      if (!event) return null;
+      if (event.event === "$exception") {
+        const url = window.location.href;
+        if (url.includes("ui.rilldata.com") || url.includes("localhost")) {
+          return event;
+        }
+        // Drop $exception events for all other URLs
+        return null;
+      }
+      // Allow all other events
+      return event;
+    },
   });
 }
 


web-common/src/lib/analytics/posthog.ts
```diff
@@ -30,6 +30,24 @@ export function initPosthog(rillVersion: string, sessionId?: string | null) {
...
+          hostname === "localhost" ||
+          hostname === "127.0.0.1"
+        ) {
+          return event;
+        }
+        // Drop $exception events for all other URLs
+        return null;
+      }
+      // Allow all other events
+      return event;
```
ericpgreen2: A little clean-up suggestion:
```suggestion
      if (event?.event !== "$exception") {
        return event;
      }

      // For $exception events, only send them from specific hostnames.
      // TODO: A ONE-LINE COMMENT THAT EXPLAINS WHY WE NEED TO DO THIS
      const hostname = window.location.hostname;
      const allowedHostnames = ["ui.rilldata.com", "localhost", "127.0.0.1"];
      if (allowedHostnames.includes(hostname)) {
        return event;
      }

      // Drop $exception events for all other hostnames.
      return null;
```

ericpgreen2: Additionally, would you mind adding a comment that explains the motivation for this code? E.g. does this avoid collecting exceptions thrown by Chrome extensions?

ericpgreen2 (changes requested):

web-common/src/lib/analytics/posthog.ts
ericokuma: @ericpgreen2, this was motivated by the desire to decrease our potential for large event tracking bills from PostHog. <br><br>I'm not sure if this will avoid collecting exceptions thrown by Chrome extensions but it'll certainly avoid errors from our marketing and doc pages

royendo:
Do we still need this @ericokuma , @ericpgreen2 ?

royendo:
bump @ericokuma

web-common/src/lib/analytics/posthog.ts
ericpgreen2: I meant a code comment. In the above snippet, I've added the "TODO" as a place for you to write the explanation :)

PR #7482, opened 2025-06-18
Enable PostHog session redaction for non-US and California users only

ericokuma:
Mike's request to remove redaction of data in Session Replays for non-US and California users

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-common/src/lib/analytics/posthog.ts
+++ b/web-common/src/lib/analytics/posthog.ts
@@ -4,33 +4,65 @@ const POSTHOG_API_KEY = import.meta.env.RILL_UI_PUBLIC_POSTHOG_API_KEY;
 
 export function initPosthog(rillVersion: string, sessionId?: string | null) {
   // No need to proceed if PostHog is already initialized
-  if (posthog.__loaded) return;
+  if ((posthog as any).__loaded) return;
 
   if (!POSTHOG_API_KEY) {
     console.warn("PostHog API Key not found");
     return;
   }
 
-  // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-member-access
-  posthog.init(POSTHOG_API_KEY, {
-    api_host: "https://us.i.posthog.com", // TODO: use a reverse proxy https://posthog.com/docs/advanced/proxy
-    session_recording: {
-      maskAllInputs: true,
-      maskTextSelector: "*",
-      recordHeaders: true,
-      recordBody: false,
-    },
-    autocapture: true,
-    enable_heatmaps: true,
-    bootstrap: {
-      sessionID: sessionId ?? undefined,
-    },
-    loaded: (posthog) => {
-      posthog.register_for_session({
-        "Rill version": rillVersion,
+  fetch("https://ipapi.co/json/")
+    .then((res) => res.json())
+    .then((data) => {
+      const isUS = data && data.country_code === "US";
+      const isCalifornia =
+        isUS && (data.region_code === "CA" || data.region === "California");
+      const shouldRedact = !isUS || isCalifornia;
+
+      posthog.init(POSTHOG_API_KEY, {
+        api_host: "https://us.i.posthog.com",
+        session_recording: shouldRedact
+          ? {
+              maskAllInputs: true,
+              maskTextSelector: "*",
+              recordHeaders: true,
+              recordBody: false,
+            }
+          : {}, // No redaction for non-California US users
+        autocapture: true,
+        enable_heatmaps: true,
+        bootstrap: {
+          sessionID: sessionId ?? undefined,
+        },
+        loaded: (posthog) => {
+          posthog.register_for_session({
+            "Rill version": rillVersion,
+          });
+        },
       });
-    },
-  });
+    })
+    .catch(() => {
+      // Fallback: initialize with redaction if GeoIP fails
+      posthog.init(POSTHOG_API_KEY, {
+        api_host: "https://us.i.posthog.com",
+        session_recording: {
+          maskAllInputs: true,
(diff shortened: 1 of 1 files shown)


web-common/src/lib/analytics/posthog.ts
```diff
@@ -4,33 +4,65 @@ const POSTHOG_API_KEY = import.meta.env.RILL_UI_PUBLIC_POSTHOG_API_KEY;
...
-    },
-    autocapture: true,
-    enable_heatmaps: true,
-    bootstrap: {
-      sessionID: sessionId ?? undefined,
-    },
-    loaded: (posthog) => {
-      posthog.register_for_session({
-        "Rill version": rillVersion,
+  fetch("https://ipapi.co/json/")
```
ericpgreen2, line 14: I'm uneasy about calling this API on every app load. Here's a convo that suggests using PostHog's server-side abilities to determine location & to conditionally set session settings:
image

What are your thoughts on that approach?

ericokuma:
i'll all for using PostHog's server-side features!

ericokuma:
From PostHog:<br><br>Unfortunately, **no** - there's no server-side implementation option for selective session replay masking, and here's why:

## **Why masking must be client-side**

From the [session replay privacy documentation](https://posthog.com/docs/session-replay/privacy):

> "PostHog offers a range of controls to limit what data is captured by session recordings. **Our privacy controls run in the browser or mobile app. So, masked data is never sent over the network to PostHog.**"

This is by design for several important reasons:

1. **Privacy by default**: Sensitive data should never travel over the network unmasked, even if it would be masked later
2. **Compliance**: GDPR, CCPA, and other regulations require that you don't collect sensitive data in the first place - not just that you don't store it
3. **Security**: If unmasked data reaches servers, it could be intercepted, logged, or cached somewhere in the pipeline
4. **Irreversible**: Once data is masked client-side, it cannot be unmasked server-side (there's no original data to recover)

## **What PostHog doesn't offer server-side**

* No server-side masking/unmasking controls
* No conditional masking based on user location after data arrives
* No way to "unmask" data when viewing replays for certain admins/viewers

## **Your only option: Client-side conditional masking**

The client-side approach I outlined above is the **only** way to achieve your goal. The key insight is:

**You're not "unmasking" data when viewing - you're choosing not to mask it during capture** for non-EU/CA users.

### **Alternative consideration: Project-based separation**

If you absolutely need server-side control, you could:

1. Create separate PostHog projects for regulated (EU/CA) vs. non-regulated regions
2. Route users to different projects based on their location (server-side routing decision)
3. Configure different masking rules per project

But this adds significant complexity and management overhead compared to the client-side conditional approach.

Would the client-side solution I provided earlier work for your needs, or is there a specific reason you need server-side control?

ericokuma:
Here's their recommendation:<br>If your goal is to **view** unredacted session replays for non-EU/CA users, you need to configure masking at the **capture time** (not viewing time). Once data is masked during capture, it cannot be unmasked later - the original data is gone.

## **Simpler solution: Only mask at capture for EU/CA users**

Here's the straightforward approach:

### **Implementation**

```
posthog.init('<ph_project_api_key>', {
    api_host: 'https://us.i.posthog.com',
    session_recording: {
        // Disable default masking
        maskAllInputs: false,
        maskAllText: false,
        
        // Use a custom masking function that only masks for EU/CA users
        maskInputFn: (text, element) => {
            // You'll need to set a person property indicating EU/CA location
            // This gets set once based on their GeoIP on first visit
            const isRegulatedUser = posthog.getPersonProperty('is_regulated_region');
            
            if (isRegulatedUser) {
                return '*'.repeat(text.length);
            }
            
            // Still mask passwords for everyone
            if (element?.type === 'password') {
                return '*'.repeat(text.length);
            }
            
            return text; // No masking for non-EU/CA users
        },
    }
});
```

### **Better approach: Set a person property on first visit**

```
posthog.init('<ph_project_api_key>', {
    api_host: 'https://us.i.posthog.com',
    loaded: (posthog) => {
        // Get GeoIP data from PostHog's auto-captured properties
        posthog.onFeatureFlags(() => {
            const country = posthog.getPersonProperty('$geoip_country_code');
            const subdivision = posthog.getPersonProperty('$geoip_subdivision_1_code');
            
            // Check if user is in EU or California
            const euCountries = ['AT', 'BE', 'BG', 'HR', 'CY', 'CZ', 'DK', 'EE', 'FI', 'FR', 'DE', 'GR', 'HU', 'IE', 'IT', 'LV', 'LT', 'LU', 'MT', 'NL', 'PL', 'PT', 'RO', 'SK', 'SI', 'ES', 'SE'];
            const isEU = euCountries.includes(country);
            const isCA = country === 'US' && subdivision === 'CA';
            
            // Set person property for future use
            posthog.setPersonProperties({
                is_regulated_region: isEU || isCA
            });
            
            // Configure session recording based on location
            if (isEU || isCA) {
                posthog.sessionRecording.startRecording({
                    maskAllInputs: true,
                    maskAllText: false,
                });
            } else {
                posthog.sessionRecording.startRecording({
                    maskAllInputs: false,
                    maskInputOptions: {
                        password: true, // Still mask passwords
                    },
                });
            }
        });
    }
});
```

### **Key points:**

1. **Masking is permanent**: Once data is masked during capture, you cannot unmask it when viewing the replay (see [privacy docs](https://posthog.com/docs/session-replay/privacy))
2. **Capture-time decision**: The masking decision must be made when the user's session is being recorded, not when you're viewing it later
3. **Result**: Non-EU/CA user sessions will be recorded with full data visibility, so when you view those replays, you'll see everything
4. **Privacy maintained**: EU/CA users' data is still properly masked and compliant

PR #7501, opened 2025-06-23
Transaction API for file creations

NamanMahor:
Closed [PLAT-85](https://linear.app/rilldata/issue/PLAT-85/transactional-api-for-rollback-and-creation)

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


proto/rill/runtime/v1/api.proto
```diff
@@ -109,6 +109,31 @@ service RuntimeService {
...
+    };
+  }
+
+  // RollbackFileTransaction discards all staged files for the given transaction ID.
+  rpc RollbackFileTransaction(RollbackFileTransactionRequest) returns (RollbackFileTransactionResponse) {
+    option (google.api.http) = {
+      post: "/v1/instances/{instance_id}/files/transaction/rollback",
+      body: "*"
+    };
+  }
```
begelundmuller, line 135: This isn't quite the expectation for this API. The expectation is rather that the UI needs a single API that only saves files if they *reconcile* successfully. E.g. something like:
```proto
service RuntimeService {
  // ...

  rpc PutFilesIfValid(PutFilesIfValidRequest) returns (PutFilesIfValidResponse) {}
}

message PutFilesIfValidRequest {
  string instance_id = 1;
  repeated File files = 2;
}

message PutFilesIfValidResponse {
  bool success = 1;
  map <string, string> errors = 2;
}
```

The logic here can get quite tricky as it'll have to a) save the files, b) watch the reconciler state for specifically those paths, c) remove the files again if reconciles for any of the paths fail.

I think it may make sense to close this PR and hold off on implementing this feature at the moment. I wonder if the new AI codegen work will mean we don't need an API like this after all.

NamanMahor:
Closing this PR as Requirement was misunderstood and we might not need it as Ben mentioned.

PR #7502, opened 2025-06-23
PLAT-15: api to list group and project membership for a user in the org

pjain1:
I have modified API `ListOrganizationMemberUsergroups` and `ListProjectMemberUsergroups` to accept optional user email to filter groups for that user. The code already has this API

```
  // ListProjectsForOrganizationAndUser lists all the projects that an organization member user has access to.
  // It does not include projects that the user has access to through a usergroup.
  rpc ListProjectsForOrganizationAndUser(ListProjectsForOrganizationAndUserRequest) returns (ListProjectsForOrganizationAndUserResponse) {
    option (google.api.http) = {get: "/v1/organizations/{organization}/projects-for-user"};
  }
```

Does this not work to get list of projects user has access to in the org?

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


admin/database/postgres/postgres.go
```diff
@@ -1754,9 +1754,15 @@ func (c *connection) FindOrganizationMemberUsergroups(ctx context.Context, orgID
 		LEFT JOIN org_roles r ON uor.org_role_id = r.id
 		WHERE ug.org_id=$1
 	`)
+	paramIdx := 4
 	if filterRoleID != "" {
-		qry.WriteString(" AND uor.org_role_id=$4")
+		qry.WriteString(fmt.Sprintf(" AND uor.org_role_id=$%d", paramIdx))
```
begelundmuller, line 1759: nit: consider using `len(args) + 1` instead of a dedicated variable

proto/rill/admin/v1/api.proto
```diff
@@ -1622,6 +1622,7 @@ message ListOrganizationMemberUsergroupsRequest {
   string organization = 1 [(validate.rules).string.min_len = 1];
   string role = 4; // Optionally filter by role
   bool include_counts = 5; // Optionally include counts
+  string email = 6; // Optionally filter by user email
```
begelundmuller, line 1625: How is this different from `rpc ListUsergroupsForOrganizationAndUser`?

proto/rill/admin/v1/api.proto
```diff
@@ -1635,6 +1636,7 @@ message ListProjectMemberUsergroupsRequest {
   string organization = 1 [(validate.rules).string.min_len = 1];
   string project = 2 [(validate.rules).string.min_len = 1];
   string role = 5; // Optionally filter by role
+  string email = 6; // Optionally filter by user email
```
begelundmuller, line 1639: Would this return the usergroups in a project that a user is part of? I thought the requirement was to list the projects in the org that the user had access to. Not sure though

pjain1:
Closing the PR as we have the required APIs - discussion [here](https://linear.app/rilldata/issue/PLAT-15/create-api-service-to-support-listing-group-and-project-membership-for#comment-4a0831da)

PR #7506, opened 2025-06-24
feat: Add support for admin request access

AdityaHegde:
During the deploy flow we need a way to trigger request access flow for admin access. This PR adds support for this by adding args to request access landing pages.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


runtime/pkg/email/email.go
```diff
@@ -334,15 +334,24 @@ type ProjectAccessRequest struct {
...
 	OrgName     string
+	Role        string
 	ProjectName string
 	ApproveLink string
 	DenyLink    string
 }
 
 func (c *Client) SendProjectAccessRequest(opts *ProjectAccessRequest) error {
-	subject := fmt.Sprintf("%s would like to view %s/%s", opts.Email, opts.OrgName, opts.ProjectName)
+	accessPrefix := "to view"
```
k-anshul: 1. I know a lot of code already uses hardcoded string but it will be better to use constants defined in https://github.com/rilldata/rill/blob/eda2564e105a394a3f769e9b343acc81d79d2104/admin/database/database.go#L822-L823
2. Nit: Having a separate case for `viewer` improves readability IMO.

admin/server/projects.go
k-anshul, line 1: What happens if a user previously requested access for some role but later wants access for another role?
Currently that request is denied with error `have already requested access to project`. Is that an expected behaviour ?

k-anshul:
> [Backend] RequestProjectAccess API now has role param. The validation here is also updated to make sure a viewer can request access as an admin.

Looks like this is not implemented on backend. I still see the check in the API handler:
https://github.com/rilldata/rill/blob/cf6d309463a3a21c13d1a0efc1bca11925c64f2f/admin/server/projects.go#L1288-L1291

admin/server/projects.go
AdityaHegde: Yes. I saw this as well. Pushing a commit.

runtime/pkg/email/email.go
AdityaHegde: Hmm this would be a big refactor. Feels like there already is a lot of refactor in this PR. Lets do the enum refactor for the backend code in a separate PR.

AdityaHegde: I misunderstood. Update to use existing enums.

web-common/src/features/users/roles.ts
```diff
@@ -0,0 +1,6 @@
+export enum UserRoles {
```
lovincyrus: I can see that most of the changes are using an enum for roles. The reason it wasn't consolidated was that the project doesn't have a guest role. We can perhaps split into `OrgUserRoles` and `ProjectUserRoles` for clarity. A single generic enum will introduce an extra mental overload in understanding if they were for org or project.

lovincyrus (commented):
Left a comment on `UserRoles`

web-common/src/features/users/roles.ts
AdityaHegde: There is quite a bit of mixing of roles. Lemme know if the changes are correct.

lovincyrus: Took a look at the refactors with OrgUsersRoles and ProjectUserRoles. Thanks for the change!

PR #7507, opened 2025-06-24
Fix: Remote commits not reported for managed repositories

k-anshul:
INSERT DESCRIPTION HERE

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


cli/pkg/gitutil/gitcmdwrapper.go
```diff
@@ -106,8 +97,9 @@ func GitFetch(ctx context.Context, path, remote string) error {
 	return nil
 }
 
-func GitPull(ctx context.Context, path string, discardLocal bool, remote string) (string, error) {
-	st, err := RunGitStatus(path)
+// GitPull runs a git pull command in the specified path.
+func GitPull(ctx context.Context, path string, discardLocal bool, remote, remoteName string) (string, error) {
```
AdityaHegde: Should this be `RunGitPull` to match the other methods?

cli/pkg/gitutil/gitutil.go
```diff
@@ -52,6 +53,13 @@ func (g *Config) FullyQualifiedRemote() (string, error) {
 	return u.String(), nil
 }
 
+func (g *Config) RemoteName() string {
+	if g.ManagedRepo {
+		return "rill"
```
AdityaHegde: Should we name this something a bit more obscure? `__rill_remote` or something? That way there is a lower probability of a clash.

cli/pkg/gitutil/gitutil.go
```diff
@@ -309,3 +310,65 @@ func NativeGitSignature(ctx context.Context, path string) (*object.Signature, er
...
+	}
+	if remote != nil {
+		if remote.Config().URLs[0] == config.Remote {
+			// remote already exists with the same URL, no need to create it again
+			return nil
+		}
+		// if the remote already exists with a different URL, delete it
+		err = repo.DeleteRemote(config.RemoteName())
+		if err != nil {
+			return fmt.Errorf("failed to delete existing remote: %w", err)
```
AdityaHegde, line 365: Feels like this needs a warning and a confirmation. Perhaps if we rename the remote name to something obscure its not necessary.

cli/pkg/local/git.go
```diff
@@ -3,79 +3,96 @@ package local
...
-			}
-			return connect.NewResponse(&localv1.GitStatusResponse{
-				Branch:     gs.Branch,
-				GithubUrl:  gs.RemoteURL,
-				ManagedGit: true, // We assumed managed git but it can also be a native git project
-			}), nil
-		}
-
-		project, err := s.app.ch.LoadProject(ctx, s.app.ProjectPath)
+	err := gitutil.GitFetch(ctx, s.app.ProjectPath, nil)
```
AdityaHegde: Since the remotes are different now should we check for authentication upfront? Same applies to other endpoints.

cli/pkg/gitutil/gitutil.go
k-anshul: Actually there are genuine cases when `rill` will be already set (by partial deploy commands). 
But I can rename it to clash less with user specified remotes.

cli/pkg/gitutil/gitutil.go
k-anshul: Yeah renaming it to `__rill_remote` will make it less likely to clash with user set remote.
However we can not set warning because there are genuine cases where we will overwrite this.

cli/pkg/local/git.go
k-anshul: Added a check on presence of `origin` remote before making actual calls.

AdityaHegde (approved):

PR #7510, opened 2025-06-24
Cursor Rules

lovincyrus:
https://docs.cursor.com/context/rules

https://docs.cursor.com/context/rules#team-rules

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!


nishantmonu51 (approved):

ericpgreen2 (dismissed):
1. Can you please do a pass to make the rules more mutually exclusive? I see `frontend-development` includes naming conventions and component guidelines, which both have their own dedicated pages. Within the `frontend-development` rule, there's also duplication between the Data Fetching and the API Integration sections.
2. Generally, I'm a little skeptical that some of these rules are needed. I'd rather we create these rules from experience, rather than pull them from generic online guides. For example: "Use semantic HTML elements"– have we actually observed the LLMs _not_ using semantic HTML? "Use **camelCase" for function names"– have we observed the LLM not doing this for Typescript? Ideally, we have as few rules as are needed, given the costs of adding to context.

lovincyrus:
> 2\. Generally, I'm a little skeptical that some of these rules are needed. I'd rather we create these rules from experience, rather than pull them from generic online guides. For example: "Use semantic HTML elements"– have we actually observed the LLMs _not_ using semantic HTML? "Use **camelCase" for function names"– have we observed the LLM not doing this for Typescript? Ideally, we have as few rules as are needed, given the costs of adding to context.

Yes to the latter. I've observed LLM adding different naming conventions. The same goes for the true constant - it often uses camelCase. Did a pass on the naming conventions.

lovincyrus:
> 1. Can you please do a pass to make the rules more mutually exclusive? I see `frontend-development` includes naming conventions and component guidelines, which both have their own dedicated pages. Within the `frontend-development` rule, there's also duplication between the Data Fetching and the API Integration sections.

Addressed and did a pass on the frontend development rule.

PR #7512, opened 2025-06-25
fix : deploy to managed repo fails using a svc account

k-anshul:
INSERT DESCRIPTION HERE

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/cli/pkg/cmdutil/helper.go
+++ b/cli/pkg/cmdutil/helper.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"crypto/md5"
 	"encoding/hex"
-	"errors"
 	"fmt"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


cli/pkg/cmdutil/helper.go
```diff
@@ -500,10 +499,22 @@ func (h *Helper) GitSignature(ctx context.Context, path string) (*object.Signatu
...
+				Email: "service-account@rilldata.com", // not an actual email
+				When:  time.Now(),
+			}, nil
+		}
 		return nil, err
 	}
+
 	if userResp.User == nil {
-		return nil, errors.New("failed to get current user")
+		return &object.Signature{
```
pjain1: even when no user is logged in we want to allow this, is this the case for deployment from rill developer but how was it working till now?

k-anshul: User has to be log in to deploy a project. This is when authenticating with a service account token.

pjain1: lets revert this as if user is logged in, an empty user should never be returned

pjain1 (approved):

PR #7514, opened 2025-06-25
PLAT-25: support individual_credentials for snowflake

NamanMahor:
Closes [PLAT-25](https://linear.app/rilldata/issue/PLAT-25/support-configuring-snowflake-with-individual-properties-instead-of-a)

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/runtime/drivers/snowflake/snowflake.go
+++ b/runtime/drivers/snowflake/snowflake.go
@@ -2,16 +2,19 @@ package snowflake
 
 import (
 	"context"
+	"crypto/rsa"
+	"crypto/x509"
+	"encoding/base64"
+	"encoding/pem"
 	"errors"
+	"fmt"
 
 	"github.com/mitchellh/mapstructure"
 	"github.com/rilldata/rill/runtime/drivers"
 	"github.com/rilldata/rill/runtime/pkg/activity"
 	"github.com/rilldata/rill/runtime/storage"
+	"github.com/snowflakedb/gosnowflake"
 	"go.uber.org/zap"
-
-	// Load database/sql driver
-	_ "github.com/snowflakedb/gosnowflake"
 )
 
 func init() {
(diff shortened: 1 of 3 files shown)


runtime/drivers/snowflake/snowflake.go
```diff
@@ -65,8 +68,67 @@ var spec = drivers.Spec{
 type driver struct{}
 
 type configProperties struct {
```
k-anshul, line 70: You will also need to set these new props in `spec`

runtime/drivers/snowflake/warehouse.go
```diff
@@ -46,18 +46,20 @@ func (c *connection) QueryAsFiles(ctx context.Context, props map[string]any) (ou
 		span.End()
 	}()
 
-	srcProps, err := parseSourceProperties(props)
+	sourceProperties, err := parseSourceProperties(props)
```
k-anshul: Nit: Unlike other programming languages, it is idiomatic in golang to use smaller variable names for small scopes.

runtime/drivers/snowflake/snowflake.go
```diff
@@ -205,3 +267,41 @@ func (c *connection) AsWarehouse() (drivers.Warehouse, bool) {
 func (c *connection) AsNotifier(properties map[string]any) (drivers.Notifier, error) {
 	return nil, drivers.ErrNotNotifier
 }
+
+// parseRSAPrivateKey parses a private key string that is either:
```
k-anshul: Incomplete godoc

runtime/drivers/snowflake/snowflake.go
```diff
@@ -205,3 +267,41 @@ func (c *connection) AsWarehouse() (drivers.Warehouse, bool) {
...
+
+	// Try PKCS#8
+	if key, err := x509.ParsePKCS8PrivateKey(keyBytes); err == nil {
+		if rsaKey, ok := key.(*rsa.PrivateKey); ok {
+			return rsaKey, nil
+		}
+		return nil, errors.New("unsupported key type: not RSA (PKCS#8)")
+	}
+
+	return nil, errors.New("failed to parse RSA private key (neither PKCS#1 nor PKCS#8)")
```
k-anshul: May be I am missing something but we don't seem to be parsing PKCS#1 ?
Also are there any recommendations from the driver/snowflake to use PKCS#1 or PKCS#8, then we could just support one ?

k-anshul (changes requested):

runtime/drivers/snowflake/snowflake.go
NamanMahor: https://github.com/snowflakedb/gosnowflake/blob/master/dsn.go#L863
snowflake driver only support PKCS#8 
Previous I handle PKCS#1 code too later removed it. because snowflake support PKCS#8 in DSN. so I think we support also only support PKCS#8. So removing the remaining trace of PKCS#1 in code.

runtime/drivers/snowflake/snowflake.go
NamanMahor: Does want to break the UI. may be UI team need to add it.

k-anshul (approved):

PR #7517, opened 2025-06-25
Deprecate type source

lovincyrus:
Closes https://linear.app/rilldata/issue/APP-86/deprecate-type-sources-when-adding-connectors

This pull request completes the phase 1 of https://www.notion.so/rilldata/Deprecate-type-sources-when-adding-connectors-move-towards-connectors-models-only-1ebba33c8f57808da3e9cdc158357c98

The key change is replacing `type: source` → `type: model`. We handled this in the backend where we treat source files as models. Sources: https://github.com/rilldata/rill/blob/main/runtime/parser/parse_source.go#L34-L36 & https://github.com/rilldata/rill/blob/main/runtime/parser/parse_source.go#L66

- [x] in addition to type: model, we should also add materialize: true
- [x] fix e2e

CleanShot 2025-07-15 at 09 27 54@2x

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!


web-local/tests/data/projects/AdBids/sources/AdBids.yaml
```diff
@@ -1,7 +1,7 @@
 # Source YAML
 # Reference documentation: https://docs.rilldata.com/reference/project-files/sources
 
-type: source
+type: model
```
ericpgreen2: We should add `materialize: true` here too

ericpgreen2 (changes requested):
FYI, given all the e2e tests that are failing, I'll wait to review this until you have e2e passing

web-local/tests/data/projects/AdBids/sources/AdBids.yaml
lovincyrus: https://github.com/rilldata/rill/pull/7517/commits/eff81506e64121a2d95e4bdec9c77b00a1fa4aba

ericpgreen2:
How's this going @lovincyrus?

lovincyrus:
Fixed isNewSource to trigger SourceImportedModal when adding a new source

CleanShot 2025-07-15 at 09 27 54@2x

lovincyrus:
Ready for +1

ericpgreen2 (approved):
Looks good. Only thing: the new "source models" have an extra line break in their code (see lines 6 & 7):
    image

PR #7518, opened 2025-06-25
Add ai_instructions details and examples

TheRillJon:
- Update to MCP page (also removed all Claude Styles information, we should push customers towards using ai_instructions)
- Update to Metrics View Customization page


royendo (approved):
LGTM

PR #7520, opened 2025-06-25
Stable keys for draggable list

lovincyrus:
This pull request addresses a bug where hiding or showing a dimension or measure would not immediately update the list in the UI.

When toggling the visibility of dimensions or measures in the "Shown dimensions" or "Hidden dimensions" lists, the UI would sometimes not update correctly. This was due to the use of array indices as keys in Svelte's {#each} blocks, which can cause DOM update issues when the list changes.

- Changed the {#each} blocks for `filteredSelectedItems` and `filteredHiddenItems` in `DashboardMetricsDraggableList.svelte` to use the unique dimension/measure ID as the key instead of the index.
- This ensures Svelte can correctly track and update each item, so the UI always reflects the current state after toggling.

Resolves https://linear.app/rilldata/issue/APP-131/ui-bug-hide-dimension-in-preview-not-working

---

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-common/src/components/menu/DashboardMetricsDraggableList.svelte
+++ b/web-common/src/components/menu/DashboardMetricsDraggableList.svelte
@@ -268,7 +268,7 @@
               : `No ${type === "measure" ? "measures" : "dimensions"} shown`}
           </div>
         {:else}
-          {#each filteredSelectedItems as id, i (i)}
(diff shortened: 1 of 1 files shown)


ericpgreen2:
After merging, can you please cherry-pick this to the release branch?

PR #7526, opened 2025-06-27
pipe embedded clickhouse logs to zap

k-anshul:
1. By default the warning and above logs are passed to zap. **Note**: The logs are very chatty. Even user errors report big stack traces.
2. Debug logs can be enabled and passed to zap by starting Rill with `--verbose` flag.

Subtask for https://linear.app/rilldata/issue/PLAT-26/qa-of-the-embedded-clickhouse

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!


runtime/drivers/clickhouse/embed.go
```diff
@@ -115,7 +116,31 @@ func (e *embedClickHouse) start() (*clickhouse.Options, error) {
 	// We read StderrPipe initially to check for clickhouse running status.
 	// Once the process is closed the stderr pipe will be closed too, io.Copy will return EOF and the goroutine will exit.
 	go func() {
-		_, _ = io.Copy(io.Discard, stderr)
+		decoder := json.NewDecoder(stderr)
+
+		for {
+			var log clickhouseLog
+			if err := decoder.Decode(&log); err != nil {
```
k-anshul, line 125: If we are worried about decoding every debug log line before rejecting it then we can add a config prop called `embed_server_log_level` and use it to set the log level to `debug` when verbose is true.
Just wanted to keep code changes minimal

runtime/drivers/clickhouse/embed.go
```diff
@@ -115,7 +116,31 @@ func (e *embedClickHouse) start() (*clickhouse.Options, error) {
...
+				if err == io.EOF {
+					// EOF means the process has exited and the stderr pipe is closed.
+					break
+				}
+				e.logger.Error("Failed to decode ClickHouse log", zap.Error(err))
+			}
+
+			switch log.Level {
+			case "Fatal":
+				e.logger.Fatal("ClickHouse embedded server: fatal log received, restart server", zap.String("logger_name", log.LoggerName), zap.String("message", log.Message))
```
begelundmuller: On the Zap side, should we map this to `error` level? We generally avoid fatal everywhere since it will not even run deferred code in any upstream logic (leading to an ungraceful shutdown).

runtime/drivers/clickhouse/embed.go
```diff
@@ -263,8 +288,16 @@ func (e *embedClickHouse) getConfigContent() ([]byte, error) {
...
-        <level>information</level>
+        <level>debug</level>
         <console>true</console>
+		<formatting>
+			<type>json</type>
+			<names>
+				<level>level</level>
+				<logger_name>logger_name</logger_name>
+				<message>message</message>
+        	</names>
```
begelundmuller: nit: indentation

runtime/drivers/clickhouse/embed.go
```diff
@@ -317,18 +350,20 @@ func (e *embedClickHouse) startAndWaitUntilReady(stderr io.Reader) error {
...
-					e.logger.Error(line)
-				} else if strings.Contains(line, "Application: Ready for connections") {
-					return nil
-				}
-			} else {
-				if err := scanner.Err(); err != nil {
-					return fmt.Errorf("error reading clickhouse logs: %w", err)
-				}
-				return fmt.Errorf("clickhouse is not ready")
+			if !scanner.Scan() {
```
begelundmuller, line 366: Since this is the `default:` case, doesn't it create a chance for an infinite loop if the scanner is closed? It will just keep calling `scanner.Scan` until the timer runs out.

begelundmuller (commented):
> Note: The logs are very chatty. Even user errors report big stack traces.

Is it also very chatty when not using `--verbose`? If yes, any thoughts on how we can reduce it?

We should generally treat log output on `localhost` as a product surface area and make sure it stays informative. If it becomes noisy or confusing, people will stop looking and miss any actually important error messages.

runtime/drivers/clickhouse/embed.go
k-anshul: Good catch. Yeah reverted to the original implementation of returning err if scan returns false

k-anshul:
> Is it also very chatty when not using `--verbose`? If yes, any thoughts on how we can reduce it?
> 
> We should generally treat log output on `localhost` as a product surface area and make sure it stays informative. If it becomes noisy or confusing, people will stop looking and miss any actually important error messages.

Unfortunately clickhouse logging is too verbose, too chatty even when only error logs are logged.
A SQL error in a model reports following two logs:
image

Any query cancellation produces a client disconnected error log trace:
image

In the current state it may be not acceptable to pipe error logs to out stdout.
I see following ways now(though none fully satisfactory):
1. Output to a different file and output a log saying clickhouse logs are being stored at this path. It is likely users will not even look at that file.
2. Keep filtering log messages in code on case by case basis. In sometime we can end up generating an exhaustive list that filters out all these unwanted error msgs.

begelundmuller:
I see. These basically don't seem like service issues, just stuff that happens during normal usage... Do you know if there are any logger names or codes that specifically indicate issues with the service itself? Maybe ask @kaspersjo , he's spent some time setting up Clickhouse monitoring. Otherwise, perhaps building up a list of codes for filtering out errors that happen in normal usage is the best path forward?

I'm not too enthusiastic about piping logs to a separate file because it doesn't quite address the goal here. The main goal is that in normal usage (not verbose), when unexpected/unhandled errors happen in the embedded Clickhouse, we want to log it so the user knows about it and can report it instead of failing silently.

begelundmuller (approved):

PR #7529, opened 2025-06-27
Support Full and Incremental Model Refresh in Rill Cloud

lovincyrus:
Closes https://linear.app/rilldata/issue/APP-47/project-status-page-add-menu-item-for-incremental-model-refresh

References
https://github.com/rilldata/rill/blob/main/cli/cmd/project/refresh.go
https://docs.rilldata.com/reference/cli/project/refresh#rill-project-refresh

This pull request adds support for both "full refresh" and "incremental refresh" actions for models in Rill Cloud. The model refresh logic in the project resources status table is updated to use the correct API payload: instead of incorrectly using the all or allFull flags for individual model refreshes, the code now sends a models array with the model name and a full flag to indicate the refresh type. This ensures only the selected model is refreshed, matching backend expectations. Additionally, the refresh confirmation dialog title now clearly indicates whether a full or incremental refresh is being performed.

Option | Efficient | Use Case | Backend Flag
-- | -- | -- | --
Incremental | Yes | Routine updates, new data only | full: false
Full | No | Rebuild, schema/logic changes | full: true

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!

### DIFF
--- a/web-admin/src/features/projects/status/ActionsCell.svelte
+++ b/web-admin/src/features/projects/status/ActionsCell.svelte
@@ -106,6 +106,7 @@
 <RefreshResourceConfirmDialog
   bind:open={isConfirmDialogOpen}
   name={resourceName}
+  {refreshType}
   onRefresh={() => {
     void refresh();
     isConfirmDialogOpen = false;
--- a/web-admin/src/features/projects/status/ActionsCell.svelte
+++ b/web-admin/src/features/projects/status/ActionsCell.svelte
@@ -25,8 +25,7 @@
     if (resourceKind === ResourceKind.Model) {
       await $createTrigger.mutateAsync({
         instanceId: $runtime.instanceId,
-        // NOTE: `RuntimeServiceCreateTriggerBody` allFull is only supported for models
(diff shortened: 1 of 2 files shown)


ericokuma:
When I select Incremental refresh, I see this confirmation message. It read "Incremental refresh {model_name}"? <br><br>Does an incremental refresh update all dependent resources too?

lovincyrus:
> 
> 
> When I select Incremental refresh, I see this confirmation message. It read "Incremental refresh {model_name}"? Does an incremental refresh update all dependent resources too?

Yes, an incremental refresh will update all dependent resources.

How it works:
- When you trigger an incremental refresh for a model, Rill will refresh that model by processing only new or changed data (not a full rebuild).
- After the model is refreshed, any resources that depend on that model (such as downstream models, metrics views, or reports) will also be refreshed as part of the dependency graph.
- This is true for both incremental and full refreshes—the difference is only in how the selected model itself is refreshed (incremental vs. full), not in the propagation to dependents.

Update: Just pushed a fix to update the copy

ericokuma (approved):
lgtm

PR #7532, opened 2025-06-30
Fix: leaderboard context dropdown allowing advanced measures

AdityaHegde:
Leaderboard context allows to select advanced measures that cant be queried in a leaderboard. Also these measures are not filtered out in dimension display.

This PR updates the leaderboard context dropdown to filter out unsupported measures. Also filters out unsupported measures from queries from dimension display.

Closes APP-179

**Checklist:**
- [x] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-common/src/features/dashboards/filters/measure-filters/measure-filter-entry.ts
+++ b/web-common/src/features/dashboards/filters/measure-filters/measure-filter-entry.ts
@@ -36,7 +36,7 @@ export const ComparisonDeltaPreviousSuffix = "_prev";
 export const ComparisonDeltaAbsoluteSuffix = "_delta";
 export const ComparisonDeltaRelativeSuffix = "_delta_perc";
 export const ComparisonPercentOfTotal = "_percent_of_total";
-export const ComparisonSuffixRegex = /_delta(?:_perc)?|_percent_of_total/;
+export const MeasureModifierSuffixRegex = /_delta(?:_perc)?|_percent_of_total/;
 
 export function mapExprToMeasureFilter(
   expr: V1Expression | undefined,
@@ -103,7 +103,7 @@ export function mapExprToMeasureFilter(
   }
 
   return {
-    measure: field.replace(ComparisonSuffixRegex, ""),
+    measure: field.replace(MeasureModifierSuffixRegex, ""),
     operation,
     type,
     value1: value1.toString(),
@@ -163,8 +163,8 @@ export function mapMeasureFilterToExpr(
 }
 
 export function measureHasSuffix(measureName: string) {
-  return ComparisonSuffixRegex.test(measureName);
+  return MeasureModifierSuffixRegex.test(measureName);
 }
 export function stripMeasureSuffix(measureName: string) {
-  return measureName.replace(ComparisonSuffixRegex, "");
+  return measureName.replace(MeasureModifierSuffixRegex, "");
 }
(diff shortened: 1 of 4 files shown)


web-common/src/features/dashboards/state-managers/selectors/measures.ts
```diff
@@ -101,56 +102,98 @@ export const filteredSimpleMeasures = ({
...
+ *
+ * This is a variant of the above but works on V1MetricsViewAggregationMeasure.
+ * Once we move all queries to MetricsViewAggregation we dont need the above method.
+ */
+export const filterOutSomeAdvancedAggregationMeasures = (
+  exploreState: ExploreState,
+  metricsViewSpec: V1MetricsViewSpec,
+  measures: V1MetricsViewAggregationMeasure[],
+  includeWindowMeasures: boolean,
+  includeTimeDependantMeasures: boolean,
```
ericpgreen2: It looks like there are no instances of this function where these flags are not `false`, so we can define them as constants inside this function.

web-common/src/features/dashboards/filters/measure-filters/measure-filter-entry.ts
```diff
@@ -36,7 +36,7 @@ export const ComparisonDeltaPreviousSuffix = "_prev";
 export const ComparisonDeltaAbsoluteSuffix = "_delta";
 export const ComparisonDeltaRelativeSuffix = "_delta_perc";
 export const ComparisonPercentOfTotal = "_percent_of_total";
-const HasSuffixRegex = /_delta(?:_perc)?|_percent_of_total/;
+export const ComparisonSuffixRegex = /_delta(?:_perc)?|_percent_of_total/;
```
ericpgreen2: Nit: "percent of total" does not indicate a comparison

web-common/src/features/dashboards/state-managers/selectors/measures.ts
ericpgreen2: (and, in this case, it's "dependent" not "dependant")

web-common/src/features/dashboards/state-managers/selectors/measures.ts
```diff
@@ -101,56 +102,98 @@ export const filteredSimpleMeasures = ({
...
+/**
+ * Selects measure valid for current dashboard selections. We filter out advanced measures that are,
+ * 1. Of type MEASURE_TYPE_TIME_COMPARISON.
+ * 2. Dependent on a time dimension with a defined grain and not equal to the current selected grain.
+ * 3. Window measures if includeWindowMeasures=false. Right now totals query does not support these.
+ *
+ * This is a variant of the above but works on V1MetricsViewAggregationMeasure.
+ * Once we move all queries to MetricsViewAggregation we dont need the above method.
+ */
+export const filterOutSomeAdvancedAggregationMeasures = (
```
ericpgreen2, line 140: Would it simpler for this function to wrap the function above?
```ts
export const filterOutSomeAdvancedAggregationMeasures = (
  // 1. Get measure names from `V1MetricsViewAggregationMeasure[]
  // 2. Call `filterOutSomeAdvancedMeasures`
)
```

AdityaHegde: Hmmm I want to get rid of `filterOutSomeAdvancedMeasures` eventually. So it would be redundant to have it.

PR #7546, opened 2025-07-01
Move alert forms to superforms

AdityaHegde:
We are bringing some common features to report and alert forms. Since reports are already on superforms it is easier to share form elements if we move alerts as well.

One thing we needed to change was the notification section. `MultiInput` only supports the old lib. `InputArray` supports superforms. So we need to switch to `InputArray`, it is a design difference since it is inline pill based input vs separate input for each entry.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


web-common/src/features/alerts/AlertForm.svelte
ericpgreen2, line 1: Nice to see the consistency with Scheduled Reports!

AdityaHegde:
Will add tests in follow up PRs for editing filters in alerts/reports and frequency in alerts.

PR #7551, opened 2025-07-02
Scaffolding for Git flows in dev deployments

begelundmuller:
This PR refactors the `drivers.RepoStore` interface and `drivers/admin` implementation of it in preparation for supporting Git commits and merge resolution for dev deployments.

Contributes to PLAT-96.

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!

### DIFF
--- a/runtime/drivers/admin/repo.go
+++ b/runtime/drivers/admin/repo.go
@@ -23,9 +23,10 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
-const repoPullTimeout = 10 * time.Minute
-
-const repoCheckpointTimeout = 30 * time.Second
+const (
+	repoPullTimeout       = 10 * time.Minute
+	repoCheckpointTimeout = 30 * time.Second
+)
 
 // repo implements the drivers.RepoStore interface.
 // It does a handshake using GetRepoMeta on the admin service to discover code files (Git, tarball archive, and/or virtual files).


runtime/drivers/admin/repo.go
```diff
@@ -23,7 +23,9 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
-const repoSyncTimeout = 10 * time.Minute
+const repoPullTimeout = 10 * time.Minute
```
k-anshul: Nit : Combine constants ?

runtime/drivers/admin/repo.go
```diff
@@ -480,42 +526,43 @@ func (r *repo) roots() []string {
...
+	// We know it's ready now. Take read lock and return.
 	return r.mu.RLock(ctx)
 }
 
-// sync clones or pulls/updates the repo with the latest code files.
+// pull clones/pulls the repo with the latest code files.
 // It is safe for concurrent use and deduplicates concurrent calls (using a singleflight).
-func (r *repo) sync(ctx context.Context) error {
-	ctx, span := tracer.Start(ctx, "r.sync")
+func (r *repo) pull(ctx context.Context, discardChanges, forceHandshake bool) error {
```
k-anshul, line 558: Is it a problem if concurrent calls to `pull` with different `discardChanges` args runs ?

runtime/drivers/admin/repo.go
```diff
@@ -419,14 +421,38 @@ func (r *repo) Watch(ctx context.Context, cb drivers.WatchCallback) error {
...
-	return r.sync(ctx)
+// Pull implements drivers.RepoStore.
+func (r *repo) Pull(ctx context.Context, discardChanges, forceHandshake bool) error {
+	return r.pull(ctx, discardChanges, forceHandshake)
+}
+
+// CommitAndPush implements drivers.RepoStore.
+func (r *repo) CommitAndPush(ctx context.Context, message string, force bool) error {
+	// Get a write lock.
+	// NOTE: Not using rlockEnsureReady here because we need to exclude reads while the commit is happening.
```
k-anshul, line 433: Why is that a requirement i.e. exclude reads while the commit is happening ?

runtime/drivers/admin/repo_git.go
```diff
@@ -119,12 +126,28 @@ func (r *gitRepo) syncInner(ctx context.Context) error {
...
 
+	// If not editable, we stay on the default branch.
+	if !r.editable() {
+		return nil
+	}
+
+	// We're in editable mode, so r.editBranch is set. We want to pull/create it and switch to it.
+	// The edit branch enables us to commit progress when closing (e.g. due to hibernation) without affecting the default branch.
+	// When pulling the editBranch, we should force pull if there are conflicts even if `force` is false (to bring us in sync with changes made in a split-brain scenario).
+	// To reduce the chance of conflicts, we should also try to merge the default branch into the edit branch (but only force merge if `force` is true).
```
k-anshul, line 146: > To reduce the chance of conflicts, we should also try to merge the default branch into the edit branch (but only force merge if `force` is true).

How and when will this flow be executed ?

runtime/drivers/admin/repo_git.go
```diff
@@ -142,6 +165,32 @@ func (r *gitRepo) root() string {
 	return r.repoDir
 }
 
+// commitToEditBranch auto-commits any current changes to the edit branch of the repository.
+// This is done to checkpoint progress when the handle is closed.
+// If there are conflicts, it should drop any local changes.
+func (r *gitRepo) commitToEditBranch(ctx context.Context) error {
```
k-anshul, line 171: Maybe I am forgetting, is this just commit or commit and push. If this is commitAndPush then 30s timeout seems somewhat less.

runtime/drivers/admin/repo_git.go
k-anshul: I see code for merging changes from edit branch to default branch but not the other way.

runtime/drivers/admin/repo.go
begelundmuller: Done

runtime/drivers/admin/repo.go
begelundmuller: It should not be – at least I tried to add logic to handle that case. Specifically, notice the singleflight has this key:
```go
key := fmt.Sprintf("pull(_, %v, %v)", discardChanges, forceHandshake)
```

So different singleflights will start for different args. Each singleflight however internally acquires the write lock, which only one can have at a time. This means that if pulls with different args are called, they will end up running once per combination of args in rough call order sequence.

runtime/drivers/admin/repo.go
begelundmuller: I think the logic in `commitAndPushToDefaultBranch` may need to mutate files (to do the pull + merge + push), hence the write lock. But if we can implement it without any file writes, then we can change this to a read lock.

runtime/drivers/admin/repo_git.go
begelundmuller: We don't have code for either of these operations right now. I'm hoping you can implement it :) 

Specifically, I think the operations should happen like this:
- In `pullInner` (referenced in your comment), it merges the default branch into the edit branch.
- In `commitAndPushToDefaultBranch` (called from `CommitAndPush`), it merges the edit branch into the default branch (only in the `admin` driver, in the `file` driver it just does a normal commit and push).

But I encourage you to think critically about it and change the behavior if you think there's a better way. This is just a proposal.

runtime/drivers/admin/repo_git.go
begelundmuller: This needs to be commit and push. Because it will be called in `Close()` when the dev deployment is hibernated. 

I agree 30s is a bit little and we can try increasing it. The problem is that shutdowns don't have long grace periods, so we may end up having it get force killed instead.

Potentially we can move it out of `Close` and instead call a separate function when hibernating the deployment to give it more time. IMO it should be fine to do that in follow up work (I'm taking a todo about that).

PR #7567, opened 2025-07-04
Implementing ListDatabaseSchemas

NamanMahor:
INSERT DESCRIPTION HERE

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/runtime/drivers/postgres/information_schema.go
+++ b/runtime/drivers/postgres/information_schema.go
@@ -7,7 +7,36 @@ import (
 )
 
 func (c *connection) ListDatabaseSchemas(ctx context.Context) ([]*drivers.DatabaseSchemaInfo, error) {
-	return nil, nil
+	q := `SELECT datname FROM pg_database 
+	WHERE has_database_privilege(datname, 'CONNECT') AND datname NOT IN ('template0', 'template1')
+	ORDER BY datname
+	`
+	db, err := c.getDB("")
+	if err != nil {
+		return nil, err
+	}
+	defer db.Close()
+
+	rows, err := db.QueryxContext(ctx, q)
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+
+	var databaseSchemas []*drivers.DatabaseSchemaInfo
+	for rows.Next() {
+		var database string
+		if err := rows.Scan(&database); err != nil {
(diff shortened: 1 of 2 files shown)


k-anshul (changes requested):
I think we should not query across all databases by making changes to the connection string.
1. Since postgres has a notion of database isolation then we must honour that and not change connection string.
2. The rest of the code will still use the default database so it will not work if user chooses table from other database.

We must restrict to the database provided in the connection string.

NamanMahor:
> I think we should not query across all databases by making changes to the connection string.
> 
> 1. Since postgres has a notion of database isolation then we must honour that and not change connection string.
> 2. The rest of the code will still use the default database so it will not work if user chooses table from other database.
> 
> We must restrict to the database provided in the connection string.

I think we can have a config like `allow_cross_database_metadata` which is false by default. it useful to see the databases user has access to. 
e.g in case of clickhouse user can do 
`  sql: "SELECT * FROM postgresql('{{.env.connector.postgres.ip}}:5432', 'dbname', 'tablename', 'user', 'pass')"`
user can se different db in different source. 

Edit: 
Discussed with Eric too. now only query the current database.

runtime/drivers/postgres/postgres.go
```diff
@@ -208,13 +208,15 @@ func (c *connection) AsNotifier(properties map[string]any) (drivers.Notifier, er
 // getDB opens a new sqlx.DB connection using the config.
 func (c *connection) getDB() (*sqlx.DB, error) {
 	conf := &ConfigProperties{}
-	if err := mapstructure.WeakDecode(c.config, conf); err != nil {
+	var err error
```
k-anshul: Nit: No need to initialize separately ?

runtime/drivers/postgres/information_schema.go
```diff
@@ -7,13 +7,107 @@ import (
 )
 
 func (c *connection) ListDatabaseSchemas(ctx context.Context) ([]*drivers.DatabaseSchemaInfo, error) {
-	return nil, nil
+	q := `
+	SELECT 
+	    current_database() AS database_name,
```
k-anshul: Nit: Formatting

PR #7568, opened 2025-07-04
PLAT-103 Implement ListDatabaseSchemas, ListTables and GetTable Apis for Mysql

NamanMahor:
Closes [#PLAT-103](https://linear.app/rilldata/issue/PLAT-103/implement-listdatabaseschemas-listtables-and-gettable-apis-for-mysql)

Implemented all 3 Apis ListDatabaseSchemas ListTables and GetTable Apis

Tested: 
1. All 3 api working 
2. Gettable api for tablename with ' in it.
3. User can only see the schema they have access to. 

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!

### DIFF
--- a/runtime/drivers/mysql/information_schema.go
+++ b/runtime/drivers/mysql/information_schema.go
@@ -11,7 +11,7 @@ func (c *connection) ListDatabaseSchemas(ctx context.Context) ([]*drivers.Databa
 	SELECT
 		schema_name
 	FROM information_schema.schemata
-	WHERE schema_name not in ('information_schema', 'performance_schema', 'sys')
+	WHERE schema_name not in ('information_schema', 'performance_schema', 'sys') OR schema_name = DATABASE()
 	ORDER BY schema_name
 	`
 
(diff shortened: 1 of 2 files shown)


runtime/drivers/mysql/information_schema.go
```diff
@@ -7,13 +7,123 @@ import (
 )
 
 func (c *connection) ListDatabaseSchemas(ctx context.Context) ([]*drivers.DatabaseSchemaInfo, error) {
-	return nil, nil
+	q := `
+	SELECT
+		schema_name
+	FROM information_schema.schemata
+	WHERE schema_name not in ('information_schema', 'performance_schema', 'sys')
```
k-anshul: This reminds me that at some point we had a requirement that some body wanted to build a dashboard on top of tables in `sys` in clickhouse.
We had modified the conditions as follows : https://github.com/rilldata/rill/blob/9165ee13c7e47199d1b6f2278b2fbcb0fc07c423/runtime/drivers/clickhouse/information_schema.go#L54-L55

If in MySQL it is possible to use one of these schemas as `current_schema` then need to modify this condition here as well. Please also make changes in other drivers.

k-anshul (changes requested):

runtime/drivers/mysql/information_schema.go
NamanMahor: sure

PR #7569, opened 2025-07-04
PLAT-104 Implement ListDatabaseSchemas, ListTables and GetTable Apis for Athena

NamanMahor:
Closes [#PLAT-104](https://linear.app/rilldata/issue/PLAT-104/implement-listdatabaseschemas-listtables-and-gettable-apis-for-athena)

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!

### DIFF
--- a/runtime/drivers/athena/athena.go
+++ b/runtime/drivers/athena/athena.go
@@ -17,6 +17,7 @@ import (
 	"github.com/mitchellh/mapstructure"
 	"github.com/rilldata/rill/runtime/drivers"
 	"github.com/rilldata/rill/runtime/pkg/activity"
+	"github.com/rilldata/rill/runtime/pkg/graceful"
 	"github.com/rilldata/rill/runtime/storage"
 	"go.opentelemetry.io/otel"
 	"go.uber.org/zap"
@@ -315,18 +316,20 @@ func (c *Connection) executeQuery(ctx context.Context, client *athena.Client, sq
 		return "", err
 	}
 
-	queryID := *queryExecutionOutput.QueryExecutionId
-
+	ticker := time.NewTicker(time.Second)
+	defer ticker.Stop()
 	for {
 		select {
 		case <-ctx.Done():
+			ctx, cancel := graceful.WithMinimumDuration(ctx, 15*time.Second)
 			_, stopErr := client.StopQueryExecution(ctx, &athena.StopQueryExecutionInput{
-				QueryExecutionId: aws.String(queryID),
+				QueryExecutionId: queryExecutionOutput.QueryExecutionId,
 			})
+			cancel()
 			return "", errors.Join(ctx.Err(), stopErr)
-		default:
+		case <-ticker.C:
 			status, err := client.GetQueryExecution(ctx, &athena.GetQueryExecutionInput{
-				QueryExecutionId: aws.String(queryID),
+				QueryExecutionId: queryExecutionOutput.QueryExecutionId,
 			})
 			if err != nil {
 				return "", err
@@ -334,7 +337,7 @@ func (c *Connection) executeQuery(ctx context.Context, client *athena.Client, sq
 
 			switch status.QueryExecution.Status.State {
 			case types2.QueryExecutionStateSucceeded:
-				return queryID, nil
+				return *queryExecutionOutput.QueryExecutionId, nil
 			case types2.QueryExecutionStateCancelled:
 				return "", fmt.Errorf("Athena query execution cancelled")
 			case types2.QueryExecutionStateFailed:
--- a/runtime/drivers/athena/athena.go
+++ b/runtime/drivers/athena/athena.go
@@ -157,7 +157,8 @@ func (c *Connection) Ping(ctx context.Context) error {
 	})
 
 	// Execute a simple query to verify connection
-	return c.executeQuery(ctx, client, "SELECT 1", c.config.Workgroup, c.config.OutputLocation)
+	_, err = c.executeQuery(ctx, client, "SELECT 1", c.config.Workgroup, c.config.OutputLocation)
+	return err
 }
 
 // Driver implements drivers.Connection.
@@ -209,7 +210,7 @@ func (c *Connection) AsOLAP(instanceID string) (drivers.OLAPStore, bool) {
 
 // AsInformationSchema implements drivers.Connection.
 func (c *Connection) AsInformationSchema() (drivers.InformationSchema, bool) {
-	return nil, false
+	return c, true
 }
 
 // Migrate implements drivers.Connection.
@@ -293,7 +294,7 @@ func (c *Connection) awsConfig(ctx context.Context, awsRegion string) (aws.Confi
 	return awsConfig, nil
 }
 
-func (c *Connection) executeQuery(ctx context.Context, client *athena.Client, sql, workgroup, outputLocation string) error {
+func (c *Connection) executeQuery(ctx context.Context, client *athena.Client, sql, workgroup, outputLocation string) (string, error) {
 	executeParams := &athena.StartQueryExecutionInput{
 		QueryString: aws.String(sql),
 	}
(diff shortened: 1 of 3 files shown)


runtime/drivers/athena/athena.go
```diff
@@ -311,31 +312,33 @@ func (c *Connection) executeQuery(ctx context.Context, client *athena.Client, sq
...
 	}
 
+	queryID := *queryExecutionOutput.QueryExecutionId
+
 	for {
 		select {
 		case <-ctx.Done():
-			_, err = client.StopQueryExecution(ctx, &athena.StopQueryExecutionInput{
-				QueryExecutionId: queryExecutionOutput.QueryExecutionId,
+			_, stopErr := client.StopQueryExecution(ctx, &athena.StopQueryExecutionInput{
```
k-anshul, line 329: I know this is existing code but given `ctx` is already completed this `StopQueryExecution` will not work.
You can use `graceful.WithMinimumDuration(ctx, 10*time.Second)`

runtime/drivers/athena/athena.go
```diff
@@ -311,31 +312,33 @@ func (c *Connection) executeQuery(ctx context.Context, client *athena.Client, sq
 
 	queryExecutionOutput, err := client.StartQueryExecution(ctx, executeParams)
 	if err != nil {
-		return err
+		return "", err
 	}
 
+	queryID := *queryExecutionOutput.QueryExecutionId
```
k-anshul: Can this be nil ?

runtime/drivers/athena/information_schema.go
```diff
@@ -2,18 +2,224 @@ package athena
...
+	input := &athena.ListDataCatalogsInput{}
+	paginator := athena.NewListDataCatalogsPaginator(client, input)
+	for paginator.HasMorePages() {
+		page, err := paginator.NextPage(ctx)
+		if err != nil {
+			var ae smithy.APIError
+			if errors.As(err, &ae) {
+				switch ae.ErrorCode() {
+				case "AccessDeniedException", "NotAuthorized":
+					// Return whatever empty catalogs
```
k-anshul: Maybe just return `nil, nil` to make it clear that this will fail at the start itself. Currently it gives an impression that it can fail after listing some catalogs.

runtime/drivers/athena/information_schema.go
```diff
@@ -2,18 +2,224 @@ package athena
...
-	return nil, nil
+	var res []*drivers.DatabaseSchemaInfo
+
+	catalogs, err := c.listCatalogs(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list catalogs: %w", err)
+	}
+	// if no catalogs query current catalog by passing empty string
+	if len(catalogs) == 0 {
+		return c.listSchemasForCatalog(ctx, "")
```
k-anshul: Will the current catalog be returned when `listCatalogs` return non empty list ?

runtime/drivers/athena/information_schema.go
```diff
@@ -2,18 +2,224 @@ package athena
...
+
+func (c *Connection) listSchemasForCatalog(ctx context.Context, catalog string) ([]*drivers.DatabaseSchemaInfo, error) {
+	// Get AWS config with configured region
+	awsConfig, err := c.awsConfig(ctx, c.config.AWSRegion)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get AWS config: %w", err)
+	}
+
+	// Create Athena client
+	client := athena.NewFromConfig(awsConfig, func(o *athena.Options) {
```
k-anshul: Reuse client in all calls ?

runtime/drivers/athena/information_schema.go
```diff
@@ -2,18 +2,224 @@ package athena
...
+
+	// Fetch results
+	results, err := client.GetQueryResults(ctx, &athena.GetQueryResultsInput{
+		QueryExecutionId: aws.String(queryID),
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to get query results: %w", err)
+	}
+
+	res := make([]*drivers.DatabaseSchemaInfo, 0, len(results.ResultSet.Rows)-1)
```
k-anshul, line 199: May be add a comment on why the first row is being skipped ?

runtime/drivers/athena/information_schema.go
```diff
@@ -2,18 +2,224 @@ package athena
...
+	for _, row := range results.ResultSet.Rows[1:] {
+		if len(row.Data) < 2 || row.Data[0].VarCharValue == nil || row.Data[1].VarCharValue == nil {
```
k-anshul, line 201: If you query for two columns in what condition can a row have less than 2 columns ?

runtime/drivers/athena/information_schema.go
```diff
@@ -2,18 +2,224 @@ package athena
...
+	"github.com/aws/aws-sdk-go-v2/service/athena"
+	"github.com/aws/smithy-go"
+	"github.com/aws/smithy-go/tracing/smithyoteltracing"
 	"github.com/rilldata/rill/runtime/drivers"
+	"go.opentelemetry.io/otel"
 )
 
 func (c *Connection) ListDatabaseSchemas(ctx context.Context) ([]*drivers.DatabaseSchemaInfo, error) {
-	return nil, nil
+	var res []*drivers.DatabaseSchemaInfo
```
k-anshul: Nit: initialize variables closer to the actual usage. Here above the for loop on line 29.

runtime/drivers/athena/information_schema.go
```diff
@@ -2,18 +2,224 @@ package athena
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to list catalogs: %w", err)
+	}
+	// if no catalogs query current catalog by passing empty string
+	if len(catalogs) == 0 {
+		return c.listSchemasForCatalog(ctx, "")
+	}
+
+	for _, catalog := range catalogs {
+		schemas, err := c.listSchemasForCatalog(ctx, catalog)
```
k-anshul: Fetch in batch as suggested in other PR.

runtime/drivers/athena/information_schema.go
NamanMahor: i think it should always be 2 but just for safe side.

runtime/drivers/athena/athena.go
NamanMahor: changed it pass directly to aws api. check they will error out id its null.

runtime/drivers/athena/athena.go
```diff
@@ -311,31 +313,35 @@ func (c *Connection) executeQuery(ctx context.Context, client *athena.Client, sq
...
+				return queryExecutionOutput.QueryExecutionId, nil
 			case types2.QueryExecutionStateCancelled:
-				return fmt.Errorf("Athena query execution cancelled")
+				return nil, fmt.Errorf("Athena query execution cancelled")
 			case types2.QueryExecutionStateFailed:
-				return fmt.Errorf("Athena query execution failed %s", *status.QueryExecution.Status.AthenaError.ErrorMessage)
+				return nil, fmt.Errorf("Athena query execution failed: %s", aws.ToString(status.QueryExecution.Status.AthenaError.ErrorMessage))
 			}
 		}
 		time.Sleep(time.Second)
```
k-anshul, line 351: If you are using a ticker than `time.Sleep` is not necessary.

IMO `time.Sleep` seems okay here:
1. The query cancelation will get delayed by max 1 second(plus API call to be precise).
2. If the API latency is high like a second or so we will effectively keep on calling the API without any delay. Ticker will not account for the time spent in API call.

runtime/drivers/athena/information_schema.go
```diff
@@ -2,18 +2,232 @@ package athena
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to list catalogs: %w", err)
+	}
+	// if no catalogs query current catalog by passing empty string
+	if len(catalogs) == 0 {
+		return c.listSchemasForCatalog(ctx, client, "")
+	}
+	var (
+		mu  sync.Mutex
+		res []*drivers.DatabaseSchemaInfo
```
k-anshul, line 33: Nit: can initialize with length of catalogs

runtime/drivers/athena/information_schema.go
```diff
@@ -2,18 +2,232 @@ package athena
...
+	"golang.org/x/sync/errgroup"
 )
 
 func (c *Connection) ListDatabaseSchemas(ctx context.Context) ([]*drivers.DatabaseSchemaInfo, error) {
-	return nil, nil
+	awsConfig, err := c.awsConfig(ctx, c.config.AWSRegion)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get AWS config: %w", err)
+	}
+	client := athena.NewFromConfig(awsConfig, func(o *athena.Options) {
```
k-anshul: Maybe extract the client creation in a small helper. Seems repeated a few times.

runtime/drivers/athena/athena.go
NamanMahor: oh forgot to remove it.

runtime/drivers/athena/information_schema.go
NamanMahor: we can't because we have unknown number of schemas

k-anshul (approved):

PR #7571, opened 2025-07-04
PLAT-105 Implement ListDatabaseSchemas, ListTables and GetTable Apis for Redshift

NamanMahor:
Closes [#PLAT-105](https://linear.app/rilldata/issue/PLAT-105/implement-listdatabaseschemas-listtables-and-gettable-apis-for)

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!

### DIFF
--- a/runtime/drivers/redshift/information_schema.go
+++ b/runtime/drivers/redshift/information_schema.go
@@ -56,7 +56,8 @@ func (c *Connection) ListTables(ctx context.Context, database, databaseSchema st
 	END AS view
 	FROM svv_all_tables
 	WHERE database_name = '%s'
-	AND schema_name = '%s';
+	AND schema_name = '%s' 
+	ORDER BY table_name;
 	`, database, databaseSchema)
 
 	awsConfig, err := c.awsConfig(ctx, c.config.AWSRegion)
@@ -88,6 +89,45 @@ func (c *Connection) ListTables(ctx context.Context, database, databaseSchema st
 	return res, nil
 }
 
-func (c *Connection) GetTable(ctx context.Context, database, schema, table string) (*drivers.TableMetadata, error) {
-	return nil, nil
+func (c *Connection) GetTable(ctx context.Context, database, databaseSchema, table string) (*drivers.TableMetadata, error) {
+	// Query to get column name and data type
+	q := fmt.Sprintf(`
+	SELECT column_name, data_type
+	FROM svv_all_columns
+	WHERE database_name = '%s'
+	AND schema_name = '%s'
+	AND table_name = '%s'
+	ORDER BY ordinal_position;
+	`, database, databaseSchema, table)
+
+	awsConfig, err := c.awsConfig(ctx, c.config.AWSRegion)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get AWS config: %w", err)
+	}
+
+	client := redshiftdata.NewFromConfig(awsConfig)
+
+	queryExecutionID, err := c.executeQuery(ctx, client, q, c.config.Database, c.config.Workgroup, c.config.ClusterIdentifier)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get table metadata: %w", err)
+	}
+
+	result, err := client.GetStatementResult(ctx, &redshiftdata.GetStatementResultInput{
+		Id: aws.String(queryExecutionID),
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to get query results: %w", err)
+	}
+
+	// Build schema map
+	schemaMap := make(map[string]string)
+	for _, record := range result.Records {
+		column := record[0].(*types.FieldMemberStringValue).Value
+		dataType := record[1].(*types.FieldMemberStringValue).Value
+		schemaMap[column] = dataType
+	}
+
+	return &drivers.TableMetadata{
+		Schema: schemaMap,
+	}, nil
 }
--- a/runtime/drivers/redshift/information_schema.go
+++ b/runtime/drivers/redshift/information_schema.go
@@ -17,7 +17,7 @@ func (c *Connection) ListDatabaseSchemas(ctx context.Context) ([]*drivers.Databa
 		database_name, 
 		schema_name 
 	FROM svv_all_tables 
-	WHERE schema_name NOT IN ('information_schema', 'pg_catalog') 
+	WHERE schema_name NOT IN ('information_schema', 'pg_catalog') OR schema_name = current_schema()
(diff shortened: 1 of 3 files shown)


runtime/drivers/redshift/information_schema.go
```diff
@@ -2,18 +2,132 @@ package redshift
...
+	"github.com/aws/aws-sdk-go-v2/service/redshiftdata/types"
 	"github.com/rilldata/rill/runtime/drivers"
 )
 
 func (c *Connection) ListDatabaseSchemas(ctx context.Context) ([]*drivers.DatabaseSchemaInfo, error) {
-	return nil, nil
+	q := `
+	SELECT database_name,schema_name 
+	FROM svv_all_tables 
+	WHERE schema_name NOT IN ('information_schema','pg_catalog')
```
k-anshul: Nit: Space after `,`

runtime/drivers/redshift/information_schema.go
```diff
@@ -2,18 +2,132 @@ package redshift
...
 	"github.com/rilldata/rill/runtime/drivers"
 )
 
 func (c *Connection) ListDatabaseSchemas(ctx context.Context) ([]*drivers.DatabaseSchemaInfo, error) {
-	return nil, nil
+	q := `
+	SELECT database_name,schema_name 
+	FROM svv_all_tables 
+	WHERE schema_name NOT IN ('information_schema','pg_catalog') 
+	GROUP BY database_name,schema_name`
```
k-anshul: Why is the `GROUP BY` required ? It may be better to do an order by `database_name` though ?

runtime/drivers/redshift/information_schema.go
```diff
@@ -2,18 +2,132 @@ package redshift
...
+		Id: aws.String(queryExecutionID),
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to get query results: %w", err)
+	}
+
+	var res []*drivers.DatabaseSchemaInfo
+	for _, record := range result.Records {
+		res = append(res, &drivers.DatabaseSchemaInfo{
+			Database:       record[0].(*types.FieldMemberStringValue).Value,
```
k-anshul: May be do safe cast just to be safe ?
```
v, ok := record[0].(*types.FieldMemberStringValue)
if !ok {
  return some error
}
```

runtime/drivers/redshift/information_schema.go
```diff
@@ -2,18 +2,132 @@ package redshift
...
+func (c *Connection) ListTables(ctx context.Context, database, databaseSchema string) ([]*drivers.TableInfo, error) {
+	q := fmt.Sprintf(`
+	SELECT
+	table_name,
+	CASE
+		WHEN table_type = 'VIEW' THEN true
+		ELSE false
+	END AS view
+	FROM svv_all_tables
+	WHERE database_name = '%s'
```
k-anshul: Either use sql args if the database/driver supports it or use `safeSQLString`.
Also the SQL needs formatting ? Example formatting : 
https://github.com/rilldata/rill/blob/e79809370c1c31018ee6167935d5d6e4f870ec3b/runtime/drivers/clickhouse/information_schema.go#L59-L75

Both applies to all the APIs.

runtime/drivers/redshift/information_schema.go
```diff
@@ -2,18 +2,132 @@ package redshift
...
+		Id: aws.String(queryExecutionID),
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to get query results: %w", err)
+	}
+
+	// Build schema map
+	schemaMap := make(map[string]string)
+	for _, record := range result.Records {
+		column := record[0].(*types.FieldMemberStringValue).Value
```
k-anshul: Nit: Maybe initialize column and datatype once outside of the loop ?

runtime/drivers/redshift/information_schema.go
```diff
@@ -2,18 +2,132 @@ package redshift
...
+
+	result, err := client.GetStatementResult(ctx, &redshiftdata.GetStatementResultInput{
+		Id: aws.String(queryExecutionID),
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to get query results: %w", err)
+	}
+
+	// Build schema map
+	schemaMap := make(map[string]string)
```
k-anshul: Nit: can set initial capacity equal to `len(result.Records)`

k-anshul (changes requested):

runtime/drivers/redshift/information_schema.go
NamanMahor: we are using `svv_all_tables` instead of `svv_all_schemas` because svv_all_schemas showing all the schema whether user have  permission or not. but  svv_all_tables only show where user has permission to. so we need group by because its on svv_all_tables.

k-anshul: > svv_all_schemas because svv_all_schemas showing all the schema whether user have permission or not. but svv_all_tables only show where user has permission to. 

This seems like a weird implementation from redshift. Curious do you have a link with this information ?
I see this information in the docs: [Superusers can see all rows; regular users can see only their own data.](https://docs.aws.amazon.com/redshift/latest/dg/r_SVV_ALL_SCHEMAS.html)

NamanMahor: Sharing the test result on chat. May be it's because the redshift does not have CONNECT Grant, any user can login to any db and see the schemas but can't query the table.

k-anshul: Thanks for the analysis. Sounds good.

runtime/drivers/redshift/information_schema.go
```diff
@@ -119,15 +136,27 @@ func (c *Connection) GetTable(ctx context.Context, database, databaseSchema, tab
...
+		dataType = typeField.Value
 		schemaMap[column] = dataType
 	}
 
 	return &drivers.TableMetadata{
 		Schema: schemaMap,
 	}, nil
 }
+
+func safeSQLString(input string) string {
```
k-anshul: There is nothing wrong with current implementation but just to consistent with every implementation of `safeSQLString` in current code : https://github.com/rilldata/rill/blob/9165ee13c7e47199d1b6f2278b2fbcb0fc07c423/runtime/drivers/olap.go#L220-L222

Wrap output in single quotes and where you are using this you can directly use `%s` and no need to do `'%s'`

k-anshul (approved):

runtime/drivers/redshift/information_schema.go
NamanMahor: done

PR #7580, opened 2025-07-07
PLAT-90: implement listdatabaseschemas listtables and gettable apis for biquery

NamanMahor:
Closes [#PLAT-90](https://linear.app/rilldata/issue/PLAT-90/implement-listdatabaseschemas-listtables-and-gettable-apis-for)

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!

### DIFF
--- a/runtime/drivers/bigquery/information_schema.go
+++ b/runtime/drivers/bigquery/information_schema.go
@@ -2,18 +2,117 @@ package bigquery
 
 import (
 	"context"
+	"errors"
+	"fmt"
 
+	"cloud.google.com/go/bigquery"
 	"github.com/rilldata/rill/runtime/drivers"
+	"google.golang.org/api/iterator"
 )
 
 func (c *Connection) ListDatabaseSchemas(ctx context.Context) ([]*drivers.DatabaseSchemaInfo, error) {
-	return nil, nil
+	client, err := c.createClient(ctx, "")
+	if err != nil {
+		return nil, fmt.Errorf("failed to get BigQuery client: %w", err)
+	}
+	defer client.Close()
+
+	var allSchemas []*drivers.DatabaseSchemaInfo
+	it := client.Datasets(ctx)
+	for {
+		ds, err := it.Next()
+		if errors.Is(err, iterator.Done) {
+			break
+		}
+		if err != nil {
+			return nil, fmt.Errorf("error listing datasets: %w", err)
+		}
+		allSchemas = append(allSchemas, &drivers.DatabaseSchemaInfo{
+			Database:       ds.ProjectID,
+			DatabaseSchema: ds.DatasetID,
+		})
+	}
+
+	return allSchemas, nil
 }
 
-func (c *Connection) ListTables(ctx context.Context, database, schema string) ([]*drivers.TableInfo, error) {
-	return nil, nil
+func (c *Connection) ListTables(ctx context.Context, database, databaseSchema string) ([]*drivers.TableInfo, error) {
+	client, err := c.createClient(ctx, "")
+	if err != nil {
+		return nil, fmt.Errorf("failed to get BigQuery client: %w", err)
+	}
+	defer client.Close()
+
+	dataset := client.DatasetInProject(database, databaseSchema)
+	it := dataset.Tables(ctx)
+
+	var res []*drivers.TableInfo
+	for {
+		tbl, err := it.Next()
+		if err != nil {
+			if errors.Is(err, iterator.Done) {
+				break
+			}
+			return nil, fmt.Errorf("error listing tables in dataset %s: %w", databaseSchema, err)
+		}
+
+		meta, err := tbl.Metadata(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("error fetching table metadata: %w", err)
+		}
+
+		res = append(res, &drivers.TableInfo{
+			Name: tbl.TableID,
+			View: meta.Type == bigquery.ViewTable,
+		})
+	}
+
+	return res, nil
 }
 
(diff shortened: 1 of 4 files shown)


runtime/drivers/bigquery/information_schema.go
```diff
@@ -2,18 +2,117 @@ package bigquery
...
 func (c *Connection) ListDatabaseSchemas(ctx context.Context) ([]*drivers.DatabaseSchemaInfo, error) {
-	return nil, nil
+	client, err := c.createClient(ctx, "")
+	if err != nil {
+		return nil, fmt.Errorf("failed to get BigQuery client: %w", err)
+	}
+	defer client.Close()
+
+	var allSchemas []*drivers.DatabaseSchemaInfo
+	it := client.Datasets(ctx)
```
k-anshul, line 21: Just to confirm this will only datasets in current project. This is expected right? Sounds good to me.

runtime/drivers/bigquery/information_schema.go
```diff
@@ -2,18 +2,117 @@ package bigquery
...
+	for {
+		ds, err := it.Next()
+		if errors.Is(err, iterator.Done) {
```
k-anshul: Nit: Move this in the error block below

runtime/drivers/bigquery/information_schema.go
```diff
@@ -2,18 +2,117 @@ package bigquery
...
+	for {
+		tbl, err := it.Next()
+		if err != nil {
+			if errors.Is(err, iterator.Done) {
+				break
+			}
+			return nil, fmt.Errorf("error listing tables in dataset %s: %w", databaseSchema, err)
+		}
+
+		meta, err := tbl.Metadata(ctx)
```
k-anshul: How fast are these calls ? If slow then might make sense to call them in batch.

runtime/drivers/bigquery/information_schema.go
```diff
@@ -2,18 +2,117 @@ package bigquery
...
+		})
+	}
+
+	return res, nil
 }
 
-func (c *Connection) GetTable(ctx context.Context, database, schema, table string) (*drivers.TableMetadata, error) {
-	return nil, nil
+func (c *Connection) GetTable(ctx context.Context, database, databaseSchema, table string) (*drivers.TableMetadata, error) {
+	query := fmt.Sprintf(`
```
k-anshul: Nit: format query

runtime/drivers/bigquery/information_schema.go
```diff
@@ -2,18 +2,117 @@ package bigquery
...
+	}
+
+	schemaMap := make(map[string]string)
+	var row struct {
+		ColumnName string `bigquery:"column_name"`
+		DataType   string `bigquery:"data_type"`
+	}
+	for {
+		err := it.Next(&row)
+		if errors.Is(err, iterator.Done) {
```
k-anshul, line 114: Nit: Move in the if error block below:

k-anshul (changes requested):

runtime/drivers/bigquery/information_schema.go
NamanMahor: changes in to query to  INFORMATION_SCHEMA.TABLES which will be fasters.

runtime/drivers/bigquery/information_schema.go
NamanMahor: We provide project_id in connector yaml and also if not provided it should be in cred to detect so I assume we want to show in current project.
also user can create another connector yaml for another project. kind of same as postgres.

NamanMahor: Discussed with @ericpgreen2 he agree to only show project id mentioned in connector yaml.

runtime/drivers/bigquery/information_schema.go
```diff
@@ -37,61 +37,69 @@ func (c *Connection) ListDatabaseSchemas(ctx context.Context) ([]*drivers.Databa
 }
 
 func (c *Connection) ListTables(ctx context.Context, database, databaseSchema string) ([]*drivers.TableInfo, error) {
-	client, err := c.createClient(ctx, "")
+	q := fmt.Sprintf(`
+	SELECT
+		table_name,
+		table_type
+	FROM `+"`%s.%s.INFORMATION_SCHEMA.TABLES`"+`
```
k-anshul, line 44: 1. Does it need `safeSQLName` for database and databaseSchema?
2. Nit: Initialize the variable closest to its usage i.e. move this query to above line 54.

runtime/drivers/bigquery/information_schema.go
```diff
@@ -37,61 +37,69 @@ func (c *Connection) ListDatabaseSchemas(ctx context.Context) ([]*drivers.Databa
...
+			View: row.TableType == "VIEW",
 		})
 	}
 
 	return res, nil
 }
 
 func (c *Connection) GetTable(ctx context.Context, database, databaseSchema, table string) (*drivers.TableMetadata, error) {
-	query := fmt.Sprintf(`
+	q := fmt.Sprintf(`
```
k-anshul, line 83: Same as `ListTables`

k-anshul (approved):

runtime/drivers/bigquery/information_schema.go
NamanMahor: 1. safeSqlName is not required as projectid and dataset name can't have special characters other then colon. for which we have added \`\`  

2. there are two variable q and client both are used in 54. I placed q first then client. that's same as every other  connector.

PR #7584, opened 2025-07-07
Add mutable option to allow overwriting existing tables in ClickHouse

grahamplata:
Add a permissive flag to enable Rill to overwrite any tables with the same name in the external ClickHouse database.

**Checklist:**
- [x] Covered by tests
- [x] Linked the [issues it closes](https://linear.app/rilldata/issue/PLAT-108/disable-clickhouse-mutations-by-default-for-byodb)
- [x] I'm proud of this work!


runtime/drivers/clickhouse/clickhouse.go
```diff
@@ -44,6 +44,16 @@ var spec = drivers.Spec{
 			Default:     "false",
 			NoPrompt:    true,
 		},
+		{
+			Key:         "mutable",
```
nishantmonu51: Would prefer to use term `readonly` instead of mutable. It clears the meaning more clearly and also a term used by clickhouse already.

grahamplata: What about `mode: readwrite`

runtime/drivers/clickhouse/clickhouse.go
```diff
@@ -464,6 +491,10 @@ func (c *Connection) AsObjectStore() (drivers.ObjectStore, bool) {
 
 // AsModelExecutor implements drivers.Handle.
 func (c *Connection) AsModelExecutor(instanceID string, opts *drivers.ModelExecutorOptions) (drivers.ModelExecutor, bool) {
+	if c.config.Mode != modeReadWrite {
```
k-anshul: Clickhouse should still be able to write to other drivers in `read` mode.
As of now there is no model executor that outputs data from clickhouse but such model executor can be added in future. Thus this warning should only be triggered if output is `clickhouse` i.e. below the first if check.

runtime/drivers/clickhouse/clickhouse.go
```diff
@@ -481,6 +512,10 @@ func (c *Connection) AsModelExecutor(instanceID string, opts *drivers.ModelExecu
 
 // AsModelManager implements drivers.Handle.
 func (c *Connection) AsModelManager(instanceID string) (drivers.ModelManager, bool) {
+	if c.config.Mode != modeReadWrite {
+		c.logger.Warn("Model management is disabled. To enable modeling on this ClickHouse database, set 'mode: readwrite' in your connector configuration. WARNING: This will allow Rill to create and overwrite tables in your database.")
```
k-anshul:
```suggestion
		c.logger.Warn("Model execution is disabled. To enable modeling on this ClickHouse database, set 'mode: readwrite' in your connector configuration. WARNING: This will allow Rill to create and overwrite tables in your database.")
```
Model management is internal construct.

k-anshul (changes requested):

k-anshul (approved):

PR #7587, opened 2025-07-08
feat: add grain dropdown to time pivot chip

djbarnwal:
Closes https://linear.app/rilldata/issue/APP-151/add-time-pill-to-tables
This also builds some capability for adding multiple time dimensions later.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


mindspank:
One nit bug: if the metrics view has a smallest_time_grain that is larger then the available data we show "No time column" in the pivot time list. E.g metrics view has data for 1-6th of June and smallest_time_grain is set to Month.

Other then that, approve from product.

djbarnwal:
> One nit bug: if the metrics view has a smallest_time_grain that is larger then the available data we show "No time column" in the pivot time list. E.g metrics view has data for 1-6th of June and smallest_time_grain is set to Month.
> 
> Other then that, approve from product.

I am wondering if we should show the time grains here. Even for explore, the grain list is empty when `smallest_time_grain` is larger than selected range. This seems like a bad config issue. Do we want to have a fix for it?

mindspank:
As you point out, I think we can leave it as is. Explore in general behaves strangely in that scenario

ericpgreen2:
@claude

web-common/src/features/dashboards/pivot/time-pill-store.ts
ericpgreen2, line 1: Thoughts on encapsulating the state+selectors+actions in a class? It could be more clean to expose the API via a class, rather than exporting everything here individually.

ericpgreen2 (approved):
Code is clean! Left one comment, but feel free to merge at your discretion.

mindspank:
Approved for cherry picking onto branch `release-0.68`

web-common/src/features/dashboards/pivot/time-pill-store.ts
djbarnwal: Yes, plan is to move to class based approach once we introduce multiple time dimensions. Thought of keeping it simpler for now.

PR #7591, opened 2025-07-08
Upgrade snowflake to v1.15.0 and arrow to arrow-go/v18 v18.3.0

NamanMahor:
Closes [PLAT-98: Upgrate snowflake driver because of security issue](https://linear.app/rilldata/issue/PLAT-98/upgrate-snowflake-driver-because-of-security-issue)

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!


runtime/drivers/snowflake/model_executor_self_objectstore.go
```diff
@@ -71,7 +71,7 @@ func (e *selfToObjectStoreExecutor) export(ctx context.Context, props map[string
 		dsn = dsnResolved
 	}
 
-	db, err := otelsql.Open("snowflake", dsn)
+	db, err := sql.Open("snowflake", dsn)
```
k-anshul, line 74: Why this change ?

k-anshul (changes requested):

runtime/drivers/snowflake/model_executor_self_objectstore.go
NamanMahor: `otelsql.Open` uses `db, err := sql.Open(driverName, "")` with empty dsn. newer version for snowflake parse the dsn on Open so erroring out as account is empty

k-anshul: Looks like they fixed it in latest versions : https://github.com/XSAM/otelsql/pull/359/files
But I think it is okay to just use `sql.Open`. The main query is anyways instrumented in the code.

PR #7595, opened 2025-07-09
null value for ch watermark

pjain1:
Fixes https://linear.app/rilldata/issue/ENG-838/syft-data-ch-based-dashboard-when-no-data-returns-weird-dashboard

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/runtime/metricsview/executor_timestamps.go
+++ b/runtime/metricsview/executor_timestamps.go
@@ -36,6 +36,11 @@ func (e *Executor) resolveDuckDBClickHouseAndPinot(ctx context.Context, timeExpr
 		filter,
 	)
 
+	// if datetime column is not nullable then ch returns 0 value instead of NULL when there are no rows so set this
+	if e.olap.Dialect() == drivers.DialectClickHouse {
+		rangeSQL += " SETTINGS aggregate_functions_null_for_empty = 1"
+	}
+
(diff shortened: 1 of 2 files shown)


runtime/metricsview/executor_timestamps.go
```diff
@@ -36,6 +36,11 @@ func (e *Executor) resolveDuckDBClickHouseAndPinot(ctx context.Context, timeExpr
 		filter,
 	)
 
+	// if datetime column is not nullable then ch returns 0 value instead of NULL when there are no rows so set this
+	if e.olap.Dialect() == drivers.DialectClickHouse {
```
k-anshul: 1. Not necessary to be handled in this PR but this function has few dialect specific checks. May be we just create separate functions for clickhouse and pinot.
2. Should we consider setting this by default ? Default to 0 behaviour can be problematic in other cases as well ?

pjain1: 1. Feels like a lot of duplication since they vary only slightly but can do if you strongly feel so
2. Don't want to change behaviour for all queries as user may start seeing different values on upgrade, since this is default CH behaviour so lets keep that, if user complains can be set using connector configs.

pjain1: I have made the change for 1. @k-anshul can you have a look again

PR #7602, opened 2025-07-09
Add summary statistics for metrics views via MCP

grahamplata:
Add summary statistics for metrics views via MCP server tool

Screenshot 2025-07-09 at 3 00 31 PM

`query_metrics_view_summary` mcp tool 

> request
```json
{
  "metrics_view": "product_metrics_pg_snapshot_date_2025_03_21_orgs_metrics"
}
```

> response
```json
    [{
        "max": "2025-03-21T15:06:06.104879Z",
        "min": "2023-05-04T12:24:25.106518Z",
        "watermark": "2025-03-21T15:06:06.104879Z",
        "summary": {
            "_all_usergroup_id": {
                "DimensionName": "_all_usergroup_id",
                "DataType": "VARCHAR",
                "SampleValue": "5bd92381-6a5b-4858-a9ec-7cc056ec82f3"
            },
            "_billing_customer_id": {
                "DimensionName": "_billing_customer_id",
                "DataType": "VARCHAR",
                "SampleValue": "24d49175-8de2-428e-9ea3-0b01db4724a5"
            },
            "_created_by_user_id": {
                "DimensionName": "_created_by_user_id",
                "DataType": "VARCHAR",
                "SampleValue": "d991b5f5-fde2-4c15-b30d-169f030b27fa"
            },
            "_id": {
                "DimensionName": "_id",
                "DataType": "VARCHAR",
                "SampleValue": "24d49175-8de2-428e-9ea3-0b01db4724a5"
            },
            "_payment_customer_id": {
                "DimensionName": "_payment_customer_id",
                "DataType": "VARCHAR",
                "SampleValue": "cus_RIEnoGvpb0AUlS"
            },
            "billing_customer_id": {
                "DimensionName": "billing_customer_id",
                "DataType": "VARCHAR",
                "SampleValue": "24d49175-8de2-428e-9ea3-0b01db4724a5"
            },
// ...snip
```
Sample execution from Claude Desktop

**Checklist:**
- [x] Covered by tests
- [x] Linked [the issues it closes](https://linear.app/rilldata/issue/PLAT-74/provide-summary-statistics-for-metrics-views-over-mcp)
- [ ] I'm proud of this work!

### DIFF
--- a/runtime/metricsview/executor_summary.go
+++ b/runtime/metricsview/executor_summary.go
@@ -12,10 +12,12 @@ import (
 )
 
 const (
-	// DefaultSummaryLimit is the maximum number of dimensions to summarize.
-	DefaultSummaryLimit = 15
-	// DefaultSampleTimeWindow is the time window used for sampling values
-	DefaultSampleTimeWindow = 24 * time.Hour
+	// SummaryDimensionsLimit is the maximum number of dimensions to summarize.
+	SummaryDimensionsLimit = 15
+	// SummaryTimeDimensionsLimit is the maximum number of time dimensions to summarize.
+	SummaryTimeDimensionsLimit = 2
+	// SummarySampleInterval is the time offset from the max timestamp used for sampling values.
+	SummarySampleInterval = 24 * time.Hour
 )
 
 // SummaryResult is the statistics for a single metrics view.
@@ -40,21 +42,12 @@ func (e *Executor) Summary(ctx context.Context) (*SummaryResult, error) {
 		return nil, runtime.ErrForbidden
 	}
 
-	dimensions := make([]*runtimev1.MetricsViewSpec_Dimension, 0, DefaultSummaryLimit)
-	timeDimensions := make([]*runtimev1.MetricsViewSpec_Dimension, 0, 2)
-
-	// Track the default time dimension name to ensure it's included
-	defaultTimeDimName := e.metricsView.TimeDimension
-
+	// Gather the categorical and time dimensions
+	var dimensions, timeDimensions []*runtimev1.MetricsViewSpec_Dimension
 	for _, dim := range e.metricsView.Dimensions {
 		// Skip dimensions that the user cannot access first
 		if !e.security.CanAccessField(dim.Name) {
 			continue
-		}
-
-		// If we have reached the limit, stop adding dimensions
-		if len(dimensions)+len(timeDimensions) >= DefaultSummaryLimit {
-			break
 		}
 
 		// Check if this dimension is a time dimension (has timestamp data type)
@@ -73,136 +66,115 @@ func (e *Executor) Summary(ctx context.Context) (*SummaryResult, error) {
 		}
 	}
 
-	// Ensure the default time dimension is included even if it's not in the dimensions list
-	defaultTimeDimFound := false
-	if defaultTimeDimName != "" {
+	// Add the default time dimension if it wasn't in the list (it's not currently guaranteed to be included in the dimensions list).
+	if e.metricsView.TimeDimension != "" {
+		var found bool
 		for _, dim := range timeDimensions {
-			if dim.Name == defaultTimeDimName {
-				defaultTimeDimFound = true
+			if dim.Name == e.metricsView.TimeDimension {
+				found = true
 				break
 			}
 		}
-	}
 
-	// Get the default time range for the metrics view using the default time dimension
-	// This will be used for time range filtering in other queries
-	var defaultTimeRange TimestampsResult
-	var err error
-	if defaultTimeDimName != "" {
-		defaultTimeRange, err = e.Timestamps(ctx, defaultTimeDimName)
-		if err != nil {
-			return nil, fmt.Errorf("failed to get default time range for dimension %q: %w", defaultTimeDimName, err)
+		if !found {
+			// Prepend it so it doesn't get truncated
+			head := []*runtimev1.MetricsViewSpec_Dimension{{
+				Name:     e.metricsView.TimeDimension,
+				DataType: &runtimev1.Type{Code: runtimev1.Type_CODE_TIMESTAMP},
+			}}
+			timeDimensions = append(head, timeDimensions...)
 		}
 	}
 
-	timeDimensionSummaries := make([]DimensionSummary, 0, len(timeDimensions)+1)
-	if !defaultTimeDimFound && defaultTimeDimName != "" && e.security.CanAccessField(defaultTimeDimName) {
-		timeDimensions = append(timeDimensions, &runtimev1.MetricsViewSpec_Dimension{
-			Name:     defaultTimeDimName,
-			DataType: &runtimev1.Type{Code: runtimev1.Type_CODE_TIMESTAMP},
-		})
+	// Apply the default dimension limits
+	if len(dimensions) > SummaryDimensionsLimit {
+		dimensions = dimensions[0:SummaryDimensionsLimit]
+	}
+	if len(timeDimensions) > SummaryTimeDimensionsLimit {
+		timeDimensions = timeDimensions[0:SummaryTimeDimensionsLimit]
 	}
 
+	// Compute the time dimension summaries
+	var summaries []DimensionSummary
+	var defaultTimeDimensionSummary DimensionSummary
 	for _, dim := range timeDimensions {
-		var timeRange TimestampsResult
-		if dim.Name == defaultTimeDimName {
-			timeRange = defaultTimeRange
-		} else {
-			timeRange, err = e.Timestamps(ctx, dim.Name)
-			if err != nil {
-				return nil, fmt.Errorf("failed to get time range for dimension %s: %w", dim.Name, err)
-			}
+		timeRange, err := e.Timestamps(ctx, dim.Name)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get time range for dimension %q: %w", dim.Name, err)
 		}
-
-		var dataType string
-		if dim.DataType != nil {
-			dataType = dim.DataType.Code.String()
-		}
-
 		summary := DimensionSummary{
-			Name:     dim.Name,
-			DataType: dataType,
+			Name: dim.Name,
 		}
-
-		// Only populate min/max if we have time data
+		if dim.DataType != nil {
+			summary.DataType = dim.DataType.Code.String()
+		}
 		if !timeRange.Min.IsZero() {
 			summary.MinValue = timeRange.Min
 		}
 		if !timeRange.Max.IsZero() {
 			summary.MaxValue = timeRange.Max
 		}
-
-		timeDimensionSummaries = append(timeDimensionSummaries, summary)
+		summaries = append(summaries, summary)
+		if dim.Name == e.metricsView.TimeDimension {
+			defaultTimeDimensionSummary = summary
+		}
 	}
 
 	// Create summaries for normal dimensions
 	if len(dimensions) == 0 {
 		return &SummaryResult{
-			Dimensions:           timeDimensionSummaries,
-			DefaultTimeDimension: createDefaultTimeDimensionSummary(defaultTimeDimName, defaultTimeRange),
+			Dimensions:           summaries,
+			DefaultTimeDimension: defaultTimeDimensionSummary,
 		}, nil
 	}
 
 	// Build and execute the summary query
 	selectClauses := make([]string, 0, len(dimensions)*4)
-	var failedDimensions []string
-
 	for _, dim := range dimensions {
-		dimName := dim.Name
 		expr, err := e.olap.Dialect().MetricsViewDimensionExpression(dim)
 		if err != nil {
-			return nil, fmt.Errorf("failed to get expression for dimension %s: %w", dimName, err)
+			return nil, fmt.Errorf("failed to get expression for dimension %s: %w", dim.Name, err)
 		}
 
-		dialect := e.olap.Dialect()
-
 		// Add select clauses for min, max, has_nulls, and sample value
+		dialect := e.olap.Dialect()
 		selectClauses = append(selectClauses,
-			fmt.Sprintf("MIN(%s) AS %s", expr, dialect.EscapeIdentifier(dimName+"__min")),
-			fmt.Sprintf("MAX(%s) AS %s", expr, dialect.EscapeIdentifier(dimName+"__max")),
-			fmt.Sprintf("COUNT(*) - COUNT(%s) > 0 AS %s", expr, dialect.EscapeIdentifier(dimName+"__has_nulls")),
-			fmt.Sprintf("ANY_VALUE(%s) AS %s", expr, dialect.EscapeIdentifier(dimName+"__example")),
+			fmt.Sprintf("MIN(%s) AS %s", expr, dialect.EscapeIdentifier(dim.Name+"__min")),
+			fmt.Sprintf("MAX(%s) AS %s", expr, dialect.EscapeIdentifier(dim.Name+"__max")),
+			fmt.Sprintf("COUNT(*) - COUNT(%s) > 0 AS %s", expr, dialect.EscapeIdentifier(dim.Name+"__has_nulls")),
+			fmt.Sprintf("ANY_VALUE(%s) AS %s", expr, dialect.EscapeIdentifier(dim.Name+"__example")),
 		)
 	}
 
-	if len(selectClauses) == 0 {
-		if len(failedDimensions) > 0 {
-			return nil, fmt.Errorf("failed to get expressions for dimensions: %v", failedDimensions)
-		}
-		return nil, fmt.Errorf("no dimensions to summarize")
-	}
-
+	// Determine the default time dimension expression
 	var timeDimExpr string
-	if defaultTimeDimName != "" {
-		timeDimExpr = e.olap.Dialect().EscapeIdentifier(defaultTimeDimName)
-		for _, dim := range e.metricsView.Dimensions {
-			if dim.Name == defaultTimeDimName {
-				expr, err := e.olap.Dialect().MetricsViewDimensionExpression(dim)
-				if err == nil {
-					timeDimExpr = expr
-				}
-				break
+	for _, dim := range timeDimensions {
+		if dim.Name == e.metricsView.TimeDimension {
+			expr, err := e.olap.Dialect().MetricsViewDimensionExpression(dim)
+			if err == nil {
+				timeDimExpr = expr
 			}
+			break
 		}
 	}
 
-	// Use a recent time window with proper timestamp formatting
-	// Use the default time dimension for filtering to ensure it's likely indexed
+	// Create a where clause that applies the SummarySampleInterval to the default time dimension
 	var whereClause string
 	var args []any
-	if defaultTimeDimName != "" && !defaultTimeRange.Max.IsZero() {
-		whereClause = fmt.Sprintf("WHERE %s >= ?", timeDimExpr)
-		args = []any{defaultTimeRange.Max.Add(-DefaultSampleTimeWindow)}
-	} else {
-		whereClause = "" // No time filtering if no default time dimension
+	if timeDimExpr != "" && defaultTimeDimensionSummary.MaxValue != nil {
+		maxTime, _ := defaultTimeDimensionSummary.MaxValue.(time.Time)
+		if !maxTime.IsZero() {
+			whereClause = fmt.Sprintf("WHERE %s >= ?", timeDimExpr)
+			args = []any{maxTime.Add(-SummarySampleInterval)}
+		}
 	}
 
-	escapedTableName := e.olap.Dialect().EscapeTable(e.metricsView.Database, e.metricsView.DatabaseSchema, e.metricsView.Table)
+	// Build the SQL query
 	var sqlBuilder strings.Builder
 	sqlBuilder.WriteString("SELECT ")
 	sqlBuilder.WriteString(strings.Join(selectClauses, ", "))
 	sqlBuilder.WriteString(" FROM ")
-	sqlBuilder.WriteString(escapedTableName)
+	sqlBuilder.WriteString(e.olap.Dialect().EscapeTable(e.metricsView.Database, e.metricsView.DatabaseSchema, e.metricsView.Table))
 	if whereClause != "" {
 		sqlBuilder.WriteString(" ")
 		sqlBuilder.WriteString(whereClause)
@@ -210,6 +182,7 @@ func (e *Executor) Summary(ctx context.Context) (*SummaryResult, error) {
 	sqlBuilder.WriteString(" LIMIT 1")
 	sql := sqlBuilder.String()
 
+	// Execute the SQL query
 	rows, err := e.olap.Query(ctx, &drivers.Statement{
 		Query:            sql,
 		Args:             args,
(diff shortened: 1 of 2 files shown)


avaitla:
One thing to consider later is if SampleValue can be [SampleValue], with top five sample values, I suspect this will get better results (low cardinality columns will show their full set of values) - then again on the other hand might be too much context. A human generally sees this immediately in the ui and benefits from it.

grahamplata:
> One thing to consider later is if SampleValue can be [SampleValue], with top five sample values, I suspect this will get better results (low cardinality columns will show their full set of values) - then again on the other hand might be too much context. A human generally sees this immediately in the ui and benefits from it.

Done and done, added more core sample
```json
[
    {
        "max": "2025-03-21T15:06:06.104879Z",
        "min": "2023-05-04T12:24:25.106518Z",
        "watermark": "2025-03-21T15:06:06.104879Z",
        "summary": {
            "custom_domain": {
                "DimensionName": "custom_domain",
                "DataType": "CODE_STRING",
                "Values": [
                    "reports.gdb.net",
                    "",
                    "rill.romen.ai",
                    "lens.digiadglobal.com",
                    "nebula.boldcollective.co"
                ]
            },
            "description": {
                "DimensionName": "description",
                "DataType": "CODE_STRING",
                "Values": [
                    "A sandbox for me",
                    "Audience-driven revenue data provided by Auth",
                    "BlueCargo",
                    "\"\"",
                    ""
                ]
            },
            "display_name": {
                "DimensionName": "display_name",
                "DataType": "CODE_STRING",
                "Values": [
                    "iansilva2305",
                    "audiohook",
                    "emodo",
                    "djkim",
                    "networkcraze"
                ]
            },
            "favicon_asset_id": {
                "DimensionName": "favicon_asset_id",
                "DataType": "CODE_UUID",
                "Values": [
                    null,
                    "60ec1638-dc36-4c2f-b4fb-56074f2aab57"
                ]
            },
            "id": {
                "DimensionName": "id",
                "DataType": "CODE_UUID",
                "Values": [
                    "78599ac2-c458-453c-86bd-1cd4aa1711c7",
                    "2ca72c5b-e5d7-48ff-8ef8-d235e493f47e",
                    "61cfb1bb-8426-468e-b023-94685cd015ae",
                    "ed5226e0-4b61-4c43-b315-db5ec6b464af",
                    "54c47392-b6ec-4dac-a9f6-db436bbf51da"
                ]
            },
            "logo_asset_id": {
                "DimensionName": "logo_asset_id",
                "DataType": "CODE_UUID",
                "Values": [
                    "54652be3-1ef5-4009-b4ff-e06f3aec2945",
                    "d9526a4f-176a-482b-ae5f-ecfe881e9280",
                    "106d3b6f-f66f-42a0-96a7-6feb3dda97db",
                    "3ec454c6-2fec-40b0-bfe8-57f37425abaf",
                    "70fa445b-596f-49a5-b6a4-41dd499ef2ae"
                ]
            },
            "name": {
                "DimensionName": "name",
                "DataType": "CODE_STRING",
                "Values": [
                    "kiwialec",
                    "syft_data",
                    "BillingTest",
                    "TTI",
                    "kensuke-b-sasaki"
                ]
            },
            "payment_customer_id": {
                "DimensionName": "payment_customer_id",
                "DataType": "CODE_STRING",
                "Values": [
                    "",
                    "cus_RnO55p6OIzqR7u",
                    "cus_RIEnwjUggkNuuB",
                    "cus_RIEnaX6gP5KpAl",
                    "cus_RIEnPcDZ10wDMq"
                ]
            },
            "updated_on": {
                "DimensionName": "updated_on",
                "DataType": "CODE_TIMESTAMP",
                "Values": [
                    "2025-01-06 01:16:48.392374+00",
                    "2025-02-25 15:32:28.471094+00",
                    "2024-12-09 06:55:56.456532+00",
                    "2025-03-17 01:15:48.121846+00",
                    "2025-01-06 01:18:41.720782+00"
                ]
            }
        }
    }
]
```

avaitla:
Nice. I think Examples or Samples is still a better name than just Values as the latter indicates it's the only set of values but the former indicates it's just an example subset. You can also ask gpt itself what it thinks the best naming is for it to best use it in context window for the use case of an AI Analyst trying to better understand the metrics view of interest.

ericpgreen2:
Tagged @k-anshul for the backend code review, given Benjamin's OOO.

Some notes about the data structure:
- The current summary structure has redundancy: dimension names serve as both object keys and object properties.
- It's unclear how to extend this pattern to include measure summaries.

What about something like this (doesn't all need implementing now, but demonstrates a cleaner, more extensible structure):

```json
{
  "time_range": {
    "min": "2023-05-04T12:24:25.106518Z",
    "max": "2025-03-21T15:06:06.104879Z",
    "watermark": "2025-03-21T15:06:06.104879Z"
  },
  "dimensions": [
    {
      "name": "custom_domain",
      "data_type": "CODE_STRING",
      "sample_values": ["reports.foodie.com", "", "rill.groupie.ai", ...],
      "null_count": 42,
      "distinct_count": 5
    },
    {
      "name": "description", 
      "data_type": "CODE_STRING",
      "sample_values": ["A sandbox for me", "Audience-driven...", ...],
      "null_count": 15,
      "distinct_count": 100
    }
  ],
  "measures": [
    {
      "name": "revenue",
      "data_type": "CODE_FLOAT64",
      "min": 0.0,
      "max": 1000000.0,
      "avg": 50000.0,
      "null_count": 0,
      "zero_count": 150
    }
  ],
  "metadata": {
    "total_rows": 1000000,
    "dimensions_count": 9,
    "measures_count": 3,
    "last_refreshed": "2025-03-21T15:06:06.104879Z"
  }
}
```

avaitla:
Curious if you run both in your own set of basic eval queries on your demo dashboard which performs better. I think the suggestion from Eric is good, not having dim name as the key for the dict but rather name: user_id, the null and total counts are also good signals (but it's worthwhile to build eval queries to avoid regression)

grahamplata:
Screenshot 2025-07-11 at 12 08 23 PM

runtime/server/mcp_server.go
```diff
@@ -177,30 +177,32 @@ func (s *Server) mcpGetMetricsView() (mcp.Tool, server.ToolHandlerFunc) {
 	return tool, handler
 }
 
-func (s *Server) mcpQueryMetricsViewTimeRange() (mcp.Tool, server.ToolHandlerFunc) {
-	tool := mcp.NewTool("query_metrics_view_time_range",
+func (s *Server) mcpQueryMetricsViewSummary() (mcp.Tool, server.ToolHandlerFunc) {
+	tool := mcp.NewTool("query_metrics_view_summary",
 		mcp.WithDescription(`
-            Retrieve the total time range available for a given metrics view.
-            Note: All subsequent queries of the metrics view should be constrained to this time range to ensure accurate results.
```
begelundmuller, line 185: Any reason for removing this note?

runtime/server/mcp_server.go
```diff
@@ -177,30 +177,32 @@ func (s *Server) mcpGetMetricsView() (mcp.Tool, server.ToolHandlerFunc) {
...
+			Retrieve summary statistics for a metrics view including:
+			- Total time range available
+			- Sample values and data types for each dimension
+		`),
 		mcp.WithString("metrics_view",
 			mcp.Required(),
 			mcp.Description("Name of the metrics view"),
 		),
 		mcp.WithString("time_dimension",
-			mcp.Description("Optional time dimension to use for resolving the time range. If not provided, the default time dimension defined under timeseries field of the metrics view will be used."),
```
begelundmuller, line 192: Any reason for removing this note?

runtime/server/mcp_server.go
```diff
@@ -177,30 +177,32 @@ func (s *Server) mcpGetMetricsView() (mcp.Tool, server.ToolHandlerFunc) {
...
+			mcp.Description("Optional time dimension to use for resolving the time range"),
 		),
+		// mcp.WithBoolean("include_measures",
+		// 	mcp.Description("Whether to include basic statistics for measures (default: false)"),
+		// ),
```
begelundmuller: Can be removed

runtime/server/mcp_server.go
```diff
@@ -209,12 +211,12 @@ func (s *Server) mcpQueryMetricsViewTimeRange() (mcp.Tool, server.ToolHandlerFun
...
 			InstanceID: instanceID,
-			Resolver:   "metrics_time_range",
+			Resolver:   "metrics_summary",
 			ResolverProperties: map[string]any{
-				"metrics_view": name,
+				"metrics_view": metricsViewName,
 			},
 			Args: map[string]any{
-				"time_dimension": timeDim,
+				"time_dimension": timeDimension,
```
begelundmuller: With the richer output here, I don't think it makes sense to pass a time dimension as input. Rather, the summary should include the min/max for each timestamp dimension present in the metrics view.

runtime/server/mcp_server.go
begelundmuller: See other comment, I'd consider not taking a time dimension at all

runtime/resolvers/metricsview_summary.go
```diff
@@ -0,0 +1,138 @@
...
+type metricsViewSummaryResolverArgs struct {
+	Priority      int    `mapstructure:"priority"`
+	TimeDimension string `mapstructure:"time_dimension"` // if empty, the default time dimension in mv is used
+}
+
+type metricsViewSummary struct {
+	MetricsView string `mapstructure:"metrics_view"`
+}
+
+func newMetricsViewSummaryResolver(ctx context.Context, opts *runtime.ResolverOptions) (runtime.Resolver, error) {
```
begelundmuller: In our terminology, you make a "metrics query" to a "metrics view". So consider calling these types `metricsSummaryResolver`, `metricsSummaryProps`, `metricsSummaryArgs`, etc.

runtime/resolvers/metricsview_summary.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	mv := res.GetMetricsView().State.ValidSpec
+	if mv == nil {
+		return nil, fmt.Errorf("metrics view %q is invalid", res.Meta.Name.Name)
+	}
+
+	if mv.TimeDimension == "" && args.TimeDimension == "" {
+		return nil, fmt.Errorf("no time dimension specified for metrics view %q", tr.MetricsView)
+	}
```
begelundmuller: For a summary, I think it should gracefully handle metrics view time dimensions

runtime/metricsview/executor_summary.go
```diff
@@ -0,0 +1,124 @@
...
+	nullCount := totalCount - nonNullCount
+
+	return &DimensionSummary{
+		Name:          dim.Name,
+		DataType:      dim.DataType.Code.String(),
+		SampleValues:  sampleValues,
+		NullCount:     nullCount,
+		DistinctCount: distinctCount,
+	}, nil
+}
```
begelundmuller: I don't think this is safe for large scale datasets like some of those we have in Druid or Clickhouse – if there are many dimensions, won't it lead to potentially 100s of separate "all time" queries? Making queries without a constrained time range can be particularly dangerous since older time ranges might be in cold storage.

I'm also a bit worried about the `DISTINCT` as it might lead to large intermediate results. Could it maybe take the `MIN` and `MAX` as samples instead?

I don't have an answer here, but some ideas:
1. Loop over the time dimensions and use `e.Timestamps` to get the time range for each of them. `e.Timestamps` is optimized to hit metadata in Druid to prevent loading data from cold storage.
2. For normal dimensions, consider if we can construct a single query that e.g. hits a small time range and gets samples for all dimensions in one go.  For example, it could be something like this:
```sql
SELECT
  MIN(<expr>) AS dim1__min, MAX(<expr>) AS dim1__max, <null check> AS dim1__hasnull,
  MIN(<expr>) AS dim2__min, MAX(<expr>) AS dim2__max, <null check> AS dim2__hasnull,
  ...
FROM <table>
WHERE <default time dimension> >= <small recent time range, e.g. max time minus 1 day>
```

You might also consider starting out simple by just getting one sample value per dimension (e.g. using just `LIMIT 1`). I realize multiple samples are nice, just wondering about the trade-offs of query performance vs. how much extra value the info gives the LLM.

Another option is looking into DB-specific support for descriptive column info, however it's probably best to start simple and branch out into such advanced work in a follow up if we run into issues.

runtime/metricsview/executor_summary.go
```diff
@@ -0,0 +1,250 @@
...
+	DataType  string           `json:"data_type"`
+	Value     any              `json:"value,omitempty"`
+	HasNulls  bool             `json:"has_nulls,omitempty"`
+	MinValue  any              `json:"min_value,omitempty"`
+	MaxValue  any              `json:"max_value,omitempty"`
+	TimeRange TimestampsResult `json:"time_range,omitempty"`
+}
+
+// Summary provides statistics for all dimensions and measures in the metrics view.
+func (e *Executor) Summary(ctx context.Context, timeDimension string) (*SummaryResult, error) {
```
begelundmuller: I don't think it should accept a `timeDimension` here – we support multiple time dimensions, but the AI likely doesn't know which to use in advance. So the summary should give it info about every time dimension available.

I.e. it should return min/max for every dimension where `dim.DataType` is a timestamp type, as well as the default `mv.TimeDimension` (note that `mv.TimeDimension` may or may not be present in `mv.Dimensions` – yeah that ambiguity is a bit confusing).

As for the time range filter that is used to limit other data scans, I suggest doing that using `mv.TimeDimension` as it's most likely to be indexed.

runtime/metricsview/executor_summary.go
```diff
@@ -0,0 +1,250 @@
...
+	selectClauses := make([]string, 0, len(dimensions)*4)
+	var failedDimensions []string
+
+	for _, dim := range dimensions {
+		dimName := dim.Name
+		expr, err := e.olap.Dialect().MetricsViewDimensionExpression(dim)
+		if err != nil {
+			failedDimensions = append(failedDimensions, dimName)
+			continue // Skip this dimension instead of failing entirely
+		}
```
begelundmuller, line 137: You can return the error here immediately. It would be an internal error because this stuff is already validated upstream (`e.metricsView` comes from `MetricsView.State.ValidSpec`)

runtime/metricsview/executor_summary.go
```diff
@@ -0,0 +1,250 @@
...
+					timeDimExpr = expr
+				}
+				break
+			}
+		}
+	}
+
+	// Use a recent time window with proper timestamp formatting
+	dimensionTimeRange := defaultTimeRange.Max.Add(-DefaultSampleTimeWindow)
+	whereClause := fmt.Sprintf("WHERE %s >= '%s'", timeDimExpr, dimensionTimeRange.Format(time.RFC3339))
```
begelundmuller: The metrics view may not have a time dimension. So make these checks conditional on whether a time range/expression was found.

runtime/metricsview/executor_summary.go
```diff
@@ -0,0 +1,250 @@
...
+
+const (
+	// DefaultSummaryLimit is the maximum number of dimensions to summarize.
+	DefaultSummaryLimit = 15
+	// DefaultSampleTimeWindow is the time window used for sampling values
+	DefaultSampleTimeWindow = 24 * time.Hour
+	// statsPerDimension is the number of statistics returned per dimension
+	statsPerDimension = 4
+	// summaryQueryLimit is the maximum number of rows to return in the summary query
+	summaryQueryLimit = 1
```
begelundmuller: The logic is hard-coded for these values (e.g. only reads on row, scans four values per dimension), they are not configurable. So I'd recommend removing them as top-level constants and inlining them.

runtime/metricsview/executor_summary.go
```diff
@@ -0,0 +1,250 @@
...
+		return nil, fmt.Errorf("failed to scan dimension summary results: %w", err)
+	}
+
+	// Build the result summaries
+	summaries := make([]DimensionSummary, len(dimensions))
+
+	for i, dim := range dimensions {
+		dataType := getDimensionDataType(dim)
+
+		minValue, maxValue, sampleValue, hasNulls, err := extractDimensionStats(values, i)
```
begelundmuller: You know they'll exist, otherwise the logic is very broken and a panic is warranted. So would suggest just:
```suggestion
		minValue, maxValue, sampleValue, hasNulls := vals[i], vals[i+1], vals[i+2], vals[i+3]
```

runtime/metricsview/executor_summary.go
```diff
@@ -0,0 +1,250 @@
...
+
+	maxPtr, ok := values[baseIndex+1].(*interface{})
+	if !ok {
+		return nil, nil, nil, false, fmt.Errorf("invalid max value type")
+	}
+
+	hasNullsPtr, ok := values[baseIndex+2].(*bool)
+	if !ok {
+		return nil, nil, nil, false, fmt.Errorf("invalid has_nulls type")
+	}
```
begelundmuller: Does this work? I wonder if it might always be a `*interface{}` because that's how it's created in the loop that creates `values`. The pointer may need to be dereferenced before you can cast the `interface{}` to a `bool`. Not sure though.

runtime/resolvers/metricsview_summary.go
```diff
@@ -0,0 +1,136 @@
...
+	runtime    *runtime.Runtime
+	instanceID string
+	mvName     string
+	executor   *metricsview.Executor
+	args       *metricsSummaryArgs
+}
+
+type metricsSummaryArgs struct {
+	Priority      int    `mapstructure:"priority"`
+	TimeDimension string `mapstructure:"time_dimension"` // if empty, the default time dimension in mv is used
```
begelundmuller: As per comment above, I think it should not accept any time dimension. It should summarize them all to help inform the AI about which to use.

begelundmuller (changes requested):

runtime/server/mcp_server.go
```diff
@@ -177,44 +177,37 @@ func (s *Server) mcpGetMetricsView() (mcp.Tool, server.ToolHandlerFunc) {
...
+func (s *Server) mcpQueryMetricsViewSummary() (mcp.Tool, server.ToolHandlerFunc) {
+	tool := mcp.NewTool("query_metrics_view_summary",
 		mcp.WithDescription(`
-            Retrieve the total time range available for a given metrics view.
+           	Retrieve the total time range available for a given metrics view.
             Note: All subsequent queries of the metrics view should be constrained to this time range to ensure accurate results.
-        `),
+			Retrieve summary statistics for a metrics view including:
+			- Total time range available
+			- Sample values and data types for each dimension
```
begelundmuller, line 186: This should be incorporated into the note above – it still says `"Retrieve the total time range available for a given metrics view."`

runtime/resolvers/metricsview_summary.go
```diff
@@ -0,0 +1,147 @@
...
+				switch dim.DataType.Code {
+				case runtimev1.Type_CODE_TIMESTAMP, runtimev1.Type_CODE_DATE, runtimev1.Type_CODE_TIME:
+					hasTimeDimensions = true
+				}
+			}
+		}
+		if !hasTimeDimensions {
+			return nil, fmt.Errorf("no time dimensions found for metrics view %q", tr.MetricsView)
+		}
+	}
```
begelundmuller: Ideally validation should be inside the `r.executor.Summary` implementation. But is it even needed? Or can it handle it gracefully and just return dimension summaries and not any time range summaries?

runtime/metricsview/executor_summary.go
```diff
@@ -0,0 +1,285 @@
...
+type SummaryResult struct {
+	Dimensions []DimensionSummary `json:"dimensions"`
+	TimeRange  TimestampsResult   `json:"time_range"`
+}
+
+// DimensionSummary provides statistics for a single dimension in the metrics view.
+type DimensionSummary struct {
+	Name      string           `json:"name"`
+	DataType  string           `json:"data_type"`
+	Value     any              `json:"value,omitempty"`
```
begelundmuller: Make it clear it's an example value
```suggestion
	Value     any              `json:"example_value,omitempty"`
```

runtime/metricsview/executor_summary.go
```diff
@@ -0,0 +1,285 @@
...
+	HasNulls  bool             `json:"has_nulls,omitempty"`
+	MinValue  any              `json:"min_value,omitempty"`
+	MaxValue  any              `json:"max_value,omitempty"`
+	TimeRange TimestampsResult `json:"time_range,omitempty"`
```
begelundmuller: It'd be better to only return min and max for the time ranges as well. That maps to the APIs accepting start/end timestamps, so less likely to confuse the LLM (`now` and `watermark` are not explained anywhere). 

So for time dimensions, just populate `min_value` and `max_value` with the time range's min and max.

runtime/metricsview/executor_summary.go
```diff
@@ -0,0 +1,285 @@
...
+	// DefaultSummaryLimit is the maximum number of dimensions to summarize.
+	DefaultSummaryLimit = 15
+	// DefaultSampleTimeWindow is the time window used for sampling values
+	DefaultSampleTimeWindow = 24 * time.Hour
+)
+
+// SummaryResult is the statistics for a single metrics view.
+type SummaryResult struct {
+	Dimensions []DimensionSummary `json:"dimensions"`
+	TimeRange  TimestampsResult   `json:"time_range"`
```
begelundmuller: Consider calling this `DefaultTimeDimension` and using `DimensionSummary` as per the comment about only returning min/max, not other fields. That will also ensure we return the actual name of the default time dimension, which may be useful context.

runtime/metricsview/executor_summary.go
```diff
@@ -0,0 +1,285 @@
...
+				Name:     defaultTimeDimName,
+				DataType: "TIMESTAMP", // Assume timestamp for default time dimension
+			})
+		}
+
+		return &SummaryResult{
+			Dimensions: timeDimensionSummaries,
+			TimeRange:  defaultTimeRange,
+		}, nil
+	}
```
begelundmuller: This case seems a little weird. If the range is empty, I think it can just keep it empty and keep going (not return early).

(And not add a time filter in the dimension profiling.)

runtime/metricsview/executor_summary.go
```diff
@@ -0,0 +1,285 @@
...
+
+	// Include default time dimension if it wasn't found in explicit dimensions
+	if !defaultTimeDimFound && defaultTimeDimName != "" && e.security.CanAccessField(defaultTimeDimName) {
+		timeDimensionSummaries = append(timeDimensionSummaries, DimensionSummary{
+			Name:      defaultTimeDimName,
+			TimeRange: defaultTimeRange,
+			DataType:  "TIMESTAMP", // Assume timestamp for default time dimension
+		})
+	}
```
begelundmuller: Can this be moved up below where `defaultTimeDimFound` is defined since they seem related?

grahamplata:
Sample

```json
[
    {
        "dimensions": [
            {
                "name": "__time",
                "data_type": "TIMESTAMP",
                "min_value": "2025-07-15T18:24:27.8Z",
                "max_value": "2025-07-22T18:24:27.8Z"
            },
            {
                "name": "adomain",
                "data_type": "CODE_STRING",
                "example_value": "clorox.com",
                "min_value": "1800petmeds.com",
                "max_value": "zyp.test"
            },
            {
                "name": "advertiser_name",
                "data_type": "CODE_STRING",
                "example_value": "Clorox",
                "min_value": "6sense",
                "max_value": "vertex"
            },
            {
                "name": "app_or_site",
                "data_type": "CODE_STRING",
                "example_value": "App",
                "min_value": "App",
                "max_value": "Site"
            }
        ],
        "time_range": {
            "name": "__time",
            "data_type": "TIMESTAMP",
            "min_value": "2025-07-15T18:24:27.8Z",
            "max_value": "2025-07-22T18:24:27.8Z"
        }
    }
]
```

runtime/metricsview/executor_summary.go
```diff
@@ -0,0 +1,302 @@
...
+
+	timeDimensionSummaries := make([]DimensionSummary, 0, len(timeDimensions)+1)
+
+	// Process each time dimension found in the dimensions list
+	for _, dim := range timeDimensions {
+		timeRange, err := e.Timestamps(ctx, dim.Name)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get time range for dimension %s: %w", dim.Name, err)
+		}
```
begelundmuller: If the `defaultTimeDimName` is present in `timeDimensions`, then it will end up calling `e.Timestamps` twice for it.

runtime/metricsview/executor_summary.go
```diff
@@ -0,0 +1,302 @@
...
+		// Only populate min/max if we have time data
+		if !defaultTimeRange.Min.IsZero() {
+			summary.MinValue = defaultTimeRange.Min
+		}
+		if !defaultTimeRange.Max.IsZero() {
+			summary.MaxValue = defaultTimeRange.Max
+		}
+
+		timeDimensionSummaries = append(timeDimensionSummaries, summary)
+	}
```
begelundmuller: Since this relates to the logic that builds `defaultTimeDimFound` on lines 77-85, I suggest moving this block up right below that code.

runtime/metricsview/executor_summary.go
```diff
@@ -0,0 +1,302 @@
...
+
+	// Include default time dimension if it wasn't found in explicit dimensions
+	if !defaultTimeDimFound && defaultTimeDimName != "" && e.security.CanAccessField(defaultTimeDimName) {
+		summary := DimensionSummary{
+			Name:     defaultTimeDimName,
+			DataType: "TIMESTAMP", // Assume timestamp for default time dimension
```
begelundmuller: Consider using `runtimev1.Type_CODE_TIMESTAMP.String()` to align with how it serializes other data types

runtime/metricsview/executor_summary.go
```diff
@@ -0,0 +1,302 @@
...
+			}
+		}
+	}
+
+	// Use a recent time window with proper timestamp formatting
+	// Use the default time dimension for filtering to ensure it's likely indexed
+	var whereClause string
+	if defaultTimeDimName != "" && !defaultTimeRange.Max.IsZero() {
+		dimensionTimeRange := defaultTimeRange.Max.Add(-DefaultSampleTimeWindow)
+		whereClause = fmt.Sprintf("WHERE %s >= '%s'", timeDimExpr, dimensionTimeRange.Format(time.RFC3339))
```
begelundmuller: I'm not sure this works in Clickhouse syntax. E.g. see syntax differences here: https://github.com/rilldata/rill/blob/fe929272f0aadd7512875b3d2afaeb2c63494ec4/runtime/drivers/olap.go#L541.

However, I don't even think you even need to put the literal time string into the query. Just put a `?` in the query and pass the timestamp as a positional argument when querying (i.e. in `drivers.Statement.Args`)

runtime/metricsview/executor_summary.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+	return ""
+}
+
+func createDefaultTimeDimensionSummary(defaultTimeDimName string, timeRange TimestampsResult) DimensionSummary {
+	summary := DimensionSummary{}
+
+	if defaultTimeDimName != "" {
+		summary.Name = defaultTimeDimName
+		summary.DataType = "TIMESTAMP"
```
begelundmuller: Same comment as above

begelundmuller (changes requested):
Can you also add a new test fixture for the `metricsview_summary` resolver in `runtime/resolvers/testdata`? Ideally test against DuckDB, Clickhouse and Pinot (see other test fixtures in that directory for examples, there are built-in utils for testing against different connectors).

PR #7604, opened 2025-07-10
feat (chatbot): Seed conversation with context from the active page

ericpgreen2:
This PR makes the chatbot *context-aware* by detecting what page/dashboard users are viewing and seeding new conversations with relevant system messages.

Note: As a short-term measure, this PR knowingly adds complexity to `runtime/completion.go`. Going forward, the plan is to extract context-specific prompting and tooling into a dedicated "agent" abstraction. This will be implemented in a subsequent PR.

image

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!


proto/rill/runtime/v1/api.proto
```diff
@@ -898,6 +898,12 @@ message Message {
   google.protobuf.Timestamp updated_on = 5;
 }
 
+// Application context for AI conversations
+message AppContext {
+  string context_type = 1;
```
k-anshul: It will be good to add an ENUM for this ?

runtime/completion.go
```diff
@@ -145,6 +168,115 @@ func (r *Runtime) ensureConversation(ctx context.Context, instanceID, ownerID, c
...
+		return nil, nil
+	}
+
+	switch appContext.ContextType {
+	case "general_chat":
+		return r.processGeneralChatContext(ctx, instanceID, appContext.ContextMetadata, toolService)
+	case "explore_dashboard":
+		return r.processExploreDashboardContext(ctx, instanceID, appContext.ContextMetadata, toolService)
+	default:
+		return nil, nil // Unknown context type, no processing
```
k-anshul: Maybe return an error ?

runtime/completion.go
```diff
@@ -145,6 +168,115 @@ func (r *Runtime) ensureConversation(ctx context.Context, instanceID, ownerID, c
...
+		Content: []*aiv1.ContentBlock{{
+			BlockType: &aiv1.ContentBlock_Text{
+				Text: fmt.Sprintf("Available metrics views: %s", result),
+			},
+		}},
+	}}, nil
+}
+
+// processExploreDashboardContext provides specific dashboard context
+func (r *Runtime) processExploreDashboardContext(ctx context.Context, instanceID string, metadata *structpb.Struct, toolService ToolService) ([]*runtimev1.Message, error) {
```
k-anshul, line 185: What about canvas ?

runtime/completion.go
```diff
@@ -145,6 +168,115 @@ func (r *Runtime) ensureConversation(ctx context.Context, instanceID, ownerID, c
...
+	}
+
+	explore := exploreResource.GetExplore()
+	if explore == nil || explore.State == nil || explore.State.ValidSpec == nil {
+		return nil, fmt.Errorf("explore '%s' does not have a valid spec", dashboardName)
+	}
+
+	metricsViewName := explore.State.ValidSpec.MetricsView
+
+	// Verify that the MetricsView resource exists before calling tools
```
k-anshul: I think this is not necessary. If a explore has a valid spec then metrics view must exist ?

proto/rill/runtime/v1/api.proto
```diff
@@ -936,6 +943,8 @@ message ListConversationsResponse {
 message GetConversationRequest {
   string instance_id = 1 [(validate.rules).string = {pattern: "^[_\\-a-zA-Z0-9]+$"}];
   string conversation_id = 2;
+  // Whether to include system messages in the response (defaults to false for UI use)
+  optional bool include_system_messages = 3;
```
k-anshul: Optional is not necessary and usually only used when `nil` is different from default value. In this case the default value of false is same as nil.

runtime/drivers/catalog.go
```diff
@@ -107,11 +107,13 @@ type InstanceHealth struct {
...
 type Conversation struct {
-	ID        string    `db:"conversation_id"`
-	OwnerID   string    `db:"owner_id"`
-	Title     string    `db:"title"`
-	CreatedOn time.Time `db:"created_on"`
-	UpdatedOn time.Time `db:"updated_on"`
+	ID                     string    `db:"conversation_id"`
+	OwnerID                string    `db:"owner_id"`
+	Title                  string    `db:"title"`
+	AppContextType         *string   `db:"app_context_type"`
```
k-anshul: So far in the code we try to avoid using nullable columns unless nullable has some meaning different from empty string and try to look for empty string to check the presence of the field.

runtime/drivers/catalog.go
```diff
@@ -107,11 +107,13 @@ type InstanceHealth struct {
...
+	AppContextMetadataJSON *string   `db:"app_context_metadata_json"`
```
k-anshul: If we are saving the app context as system msg then why is this still required ?
Doesn't seem to be getting used anywhere as well.

runtime/completion.go
```diff
@@ -145,6 +168,115 @@ func (r *Runtime) ensureConversation(ctx context.Context, instanceID, ownerID, c
...
 }
 
+// processAppContext processes the app context and generates contextual system messages
+func (r *Runtime) processAppContext(ctx context.Context, instanceID string, appContext *runtimev1.AppContext, toolService ToolService) ([]*runtimev1.Message, error) {
+	if appContext == nil {
+		return nil, nil
+	}
+
+	switch appContext.ContextType {
+	case "general_chat":
```
k-anshul: Is this necessary? Listing of all metrics views is something that will anyways happen ?

k-anshul (changes requested):

runtime/drivers/catalog.go
ericpgreen2: It's a bit of future-proofing, but a couple reasons come to mind:
1. When listing conversations in the UI, there's a chance we only show conversations from the current app context
2. When doing evals, it'll be helpful to filter/segment conversations by their app context

runtime/completion.go
ericpgreen2: I've just removed this case for now. I'll add it back when we actually add UI for a project-level chat. (We _will_ want to have a custom system message for this case.)

runtime/completion.go
ericpgreen2: Currently I have `app_context` as an optional parameter in the API request. Given it's optional, this shouldn't error. (However, debatably `AppContext` shouldn't be optional.)
image

runtime/completion.go
ericpgreen2: At the moment we don't show the chatbot on the Canvas page. I'll add it to the UI & add a Canvas-specific system message in a later PR.

runtime/drivers/catalog.go
```diff
@@ -110,8 +110,8 @@ type Conversation struct {
 	ID                     string    `db:"conversation_id"`
 	OwnerID                string    `db:"owner_id"`
 	Title                  string    `db:"title"`
-	AppContextType         *string   `db:"app_context_type"`
-	AppContextMetadataJSON *string   `db:"app_context_metadata_json"`
+	AppContextType         string    `db:"app_context_type"`
```
k-anshul, line 113: Given these are not pointers now it will be good to also include not null clause in the migration sql.

runtime/completion.go
k-anshul: But that is handled with `nil` check a few lines above this code. 
But I think it is fine to skip context processing now since client won't be able to set any random type.

PR #7612, opened 2025-07-10
chore(docs): Remove unused `sidebar` dependency

ericpgreen2:
Removes the `sidebar` package and its dependency the `bootstrap` package, which [Dependabot called out as having a Moderate vulnerability](https://github.com/rilldata/rill/security/dependabot/187). The `sidebar` package was not actually used anywhere in our docs.

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -44,8 +44,5 @@
       "last 1 safari version"
     ]
   },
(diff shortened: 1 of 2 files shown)


royendo (approved):
👍

PR #7619, opened 2025-07-11
copied example from explore/

royendo:
INSERT DESCRIPTION HERE

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/docs/docs/build/rill-project-file.md
+++ b/docs/docs/build/rill-project-file.md
@@ -23,10 +23,15 @@ models:
 When adding an OLAP connector to your project, this will automatically populate with the new OLAP connector. IE: `ClickHouse`, `Druid`
 
 ## `ai_instructions`
-With our new MCP feature, you may want to pass some project context to the Agent so that it can understand better what this project is used for. A classic example is that this project is a QA project and you dont want to include it in the analysis. In this case, you can add something like below.
+With our new MCP feature, you may want to pass some project context to the Agent so that it can understand better what this project is used for. A classic example is that this project is a QA set of dashboard and metrics internally and you don't want to include it in the analysis. In this case, you can add something like below.
 
 ```
-ai_instructions: This project is a QA project, do not include this project in any of the MCP calls. If a user asks for this project specifically, always indicate in the answer's header and footer, "This project is a QA project and its number's are not verified. You should not be using this for any purposes other than QA."
+ai_instructions: |
+  You are a data analyst, responding to questions from business users with precision, clarity, and concision.
+
+  Any time you are asked about metrics or business data: First use list_metrics, then use get_metrics_view and query_metrics_view_time_range to get the latest information about what dimensions, measures and time ranges are available.
+
+  When you run queries for actual data, run up to three queries in a row, and then provide the user with a summary, any insights you can see in the data, and suggest up to three things to investigate as a next step.
 ```
 
 ## Mock Users 


docs/docs/build/rill-project-file.md
```diff
@@ -23,10 +23,15 @@ models:
 When adding an OLAP connector to your project, this will automatically populate with the new OLAP connector. IE: `ClickHouse`, `Druid`
 
 ## `ai_instructions`
-With our new MCP feature, you may want to pass some project context to the Agent so that it can understand better what this project is used for. A classic example is that this project is a QA project and you dont want to include it in the analysis. In this case, you can add something like below.
+With our new MCP feature, you may want to pass some project context to the Agent so that it can understand better what this project is used for. A classic example is that this project is a QA set of dashboard and metrics internally and you don't want to include it in the analysis. In this case, you can add something like below.
```
ericpgreen2: > this project is a QA set of dashboard and metrics internally and you don't want to include it in the analysis

This doesn't make sense and does not map to the instructions below :)

docs/docs/build/rill-project-file.md
```diff
@@ -23,10 +23,15 @@ models:
...
 
 ```
-ai_instructions: This project is a QA project, do not include this project in any of the MCP calls. If a user asks for this project specifically, always indicate in the answer's header and footer, "This project is a QA project and its number's are not verified. You should not be using this for any purposes other than QA."
+ai_instructions: |
+  You are a data analyst, responding to questions from business users with precision, clarity, and concision.
+
+  Any time you are asked about metrics or business data: First use list_metrics, then use get_metrics_view and query_metrics_view_time_range to get the latest information about what dimensions, measures and time ranges are available.
+
+  When you run queries for actual data, run up to three queries in a row, and then provide the user with a summary, any insights you can see in the data, and suggest up to three things to investigate as a next step.
```
ericpgreen2: These instructions are mostly already included in our "system" instructions, so the user doesn't have to specify this themselves. The one thing here that our system instructions don't include is "suggest up to three things to investigate as a next step", so that is value-add.

docs/docs/build/rill-project-file.md
royendo: oops. also going to push a few menu item fixes to this branch too

docs/docs/build/rill-project-file.md
royendo: Made some changes :)

docs/docs/build/rill-project-file.md
```diff
@@ -23,11 +23,22 @@ models:
...
 
+```yaml
+ai_instructions: |
+  You are a data analyst supporting business users with questions about metrics and business data. Your responses should be precise, clear, and actionable.
+
+  When running data queries:
+    - Run up to three queries in sequence to explore the data from different perspectives or to answer follow-up aspects of the question.
+    - After running the queries, provide a summary of the results, highlight any notable trends or insights, and suggest up to three additional questions or areas that could be explored further based on the findings.
+
+  Your goal is to guide users to actionable insights, ensure transparency in your process, and help them make informed decisions based on the data. Instead of assuming what the user might want to see, ask the user with follow up questions on exactly what they are looking for.
```
ericpgreen2: This kind of "system" instruction is baked into our backend, so the user need not provide it in `rill.yaml`. The user should provide information that's unique to their project in particular, like guidance on which specific metrics views are most important.

royendo: updated

ericpgreen2 (approved):
LGTM!

PR #7626, opened 2025-07-14
fix: editing report/alert for broken dashboards

AdityaHegde:
We need a lot of info from dashboard when editing report or alert. So if the dashboard is broken then we need to disable editing.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-admin/src/features/alerts/metadata/AlertMetadata.svelte
+++ b/web-admin/src/features/alerts/metadata/AlertMetadata.svelte
@@ -23,6 +23,7 @@
   import ThreeDot from "@rilldata/web-common/components/icons/ThreeDot.svelte";
   import Tooltip from "@rilldata/web-common/components/tooltip/Tooltip.svelte";
   import TooltipContent from "@rilldata/web-common/components/tooltip/TooltipContent.svelte";
+  import { hasValidMetricsViewTimeRange } from "@rilldata/web-common/features/dashboards/selectors.ts";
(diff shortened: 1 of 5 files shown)


web-admin/src/features/alerts/metadata/AlertMetadata.svelte
```diff
@@ -105,7 +111,9 @@
         </h1>
         <div class="grow" />
         {#if !$isAlertCreatedByCode.data}
-          <EditAlert {alertSpec} />
+          {#if $exploreIsValid}
+            <EditAlert {alertSpec} />
+          {/if}
```
ericpgreen2: Nit: this should be `disabled` rather than hidden.

web-admin/src/features/scheduled-reports/metadata/ReportMetadata.svelte
```diff
@@ -118,9 +124,11 @@
...
             </DropdownMenu.Trigger>
             <DropdownMenu.Content align="start">
-              <DropdownMenu.Item on:click={handleEditReport}>
-                Edit report
-              </DropdownMenu.Item>
+              {#if $exploreIsValid}
+                <DropdownMenu.Item on:click={handleEditReport}>
+                  Edit report
+                </DropdownMenu.Item>
+              {/if}
```
ericpgreen2: Same here

ericpgreen2 (approved):
👍

PR #7628, opened 2025-07-14
Connector preview

lovincyrus:
This pull request displays a preview of the form state on the right panel. Closes https://linear.app/rilldata/issue/APP-180/add-connector-preview

CleanShot 2025-07-15 at 12 11 44@2x

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!

### DIFF
--- a/runtime/drivers/clickhouse/clickhouse.go
+++ b/runtime/drivers/clickhouse/clickhouse.go
@@ -61,7 +61,7 @@ var spec = drivers.Spec{
 		{
 			Key:         "dsn",
 			Type:        drivers.StringPropertyType,
-			Required:    false,
+			Required:    true,
 			DisplayName: "Connection string",
 			Placeholder: "clickhouse://localhost:9000?username=default&password=password",
 			Secret:      true,
--- a/runtime/drivers/clickhouse/clickhouse.go
+++ b/runtime/drivers/clickhouse/clickhouse.go
@@ -65,7 +65,6 @@ var spec = drivers.Spec{
 			DisplayName: "Connection string",
 			Placeholder: "clickhouse://localhost:9000?username=default&password=password",
 			Secret:      true,
-			NoPrompt:    true,
 		},
 		{
 			Key:         "host",
--- a/runtime/drivers/clickhouse/clickhouse.go
+++ b/runtime/drivers/clickhouse/clickhouse.go
@@ -65,6 +65,7 @@ var spec = drivers.Spec{
 			DisplayName: "Connection string",
 			Placeholder: "clickhouse://localhost:9000?username=default&password=password",
 			Secret:      true,
+			NoPrompt:    true,
 		},
 		{
 			Key:         "host",
--- a/runtime/drivers/druid/druid.go
+++ b/runtime/drivers/druid/druid.go
@@ -42,7 +42,6 @@ var spec = drivers.Spec{
 			DisplayName: "Connection string",
 			Placeholder: "https://example.com/druid/v2/sql/avatica-protobuf?authentication=BASIC&avaticaUser=username&avaticaPassword=password",
 			Secret:      true,
-			NoPrompt:    true,
 		},
 		{
 			Key:         "host",
--- a/runtime/drivers/druid/druid.go
+++ b/runtime/drivers/druid/druid.go
@@ -42,6 +42,7 @@ var spec = drivers.Spec{
 			DisplayName: "Connection string",
 			Placeholder: "https://example.com/druid/v2/sql/avatica-protobuf?authentication=BASIC&avaticaUser=username&avaticaPassword=password",
 			Secret:      true,
+			NoPrompt:    true,
 		},
 		{
 			Key:         "host",
--- a/web-common/src/components/forms/Tabs.svelte
+++ b/web-common/src/components/forms/Tabs.svelte
@@ -12,12 +12,13 @@
 
   const dispatch = createEventDispatcher();
 
-  function handleChange(e: CustomEvent) {
-    dispatch("change", e.detail);
+  function handleValueChange(newValue: string) {
+    value = newValue;
+    dispatch("value", newValue);
   }
 </script>
 
-<Tabs {value} on:change={handleChange} class="w-full">
+<Tabs {value} onValueChange={handleValueChange} class="w-full">
   <TabsList
     class="bg-muted h-9 rounded-[10px] w-full border border-gray-200 flex my-4 {disableMarginTop
       ? 'mt-0'
--- a/web-common/src/components/forms/Tabs.svelte
+++ b/web-common/src/components/forms/Tabs.svelte
@@ -1,5 +1,4 @@
 <script lang="ts">
-  import { createEventDispatcher } from "svelte";
   import {
     Tabs,
     TabsList,
@@ -10,11 +9,8 @@
   export let options: { value: string; label: string }[];
   export let disableMarginTop: boolean = false;
 
-  const dispatch = createEventDispatcher();
-
   function handleValueChange(newValue: string) {
     value = newValue;
-    dispatch("value", newValue);
   }
 </script>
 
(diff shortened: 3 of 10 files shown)


runtime/drivers/clickhouse/clickhouse.go
```diff
@@ -62,7 +62,7 @@ var spec = drivers.Spec{
 		{
 			Key:         "dsn",
 			Type:        drivers.StringPropertyType,
-			Required:    false,
+			Required:    true,
```
lovincyrus, line 64: As we want to disable the submit button when "Enter connection string" is selected

runtime/drivers/druid/druid.go
```diff
@@ -42,7 +42,6 @@ var spec = drivers.Spec{
 			DisplayName: "Connection string",
 			Placeholder: "https://example.com/druid/v2/sql/avatica-protobuf?authentication=BASIC&avaticaUser=username&avaticaPassword=password",
 			Secret:      true,
-			NoPrompt:    true,
```
lovincyrus: These `dsn` should not be NoPrompt.

ericokuma:
Screenshot 2025-07-15 at 4 08 43 PM
^ probably don't need this

otherwise looks good

lovincyrus:
> Screenshot 2025-07-15 at 4 08 43 PM ^ probably don't need this
> otherwise looks good

We can keep the reference doc until we decide to remove all of them. There are other reference docs in https://github.com/rilldata/rill/blob/main/web-common/src/features/file-explorer/new-files.ts

runtime/drivers/druid/druid.go
ericpgreen2: This affects the CLI too. E.g. see `NoPrompt` in `configure.go`. So it feels like we shouldn't change this.

web-common/src/components/forms/Tabs.svelte
```diff
@@ -12,12 +12,13 @@
 
   const dispatch = createEventDispatcher();
 
-  function handleChange(e: CustomEvent) {
-    dispatch("change", e.detail);
+  function handleValueChange(newValue: string) {
+    value = newValue;
+    dispatch("value", newValue);
```
ericpgreen2: It looks like you're not using this dispatched value anywhere, so can we remove the dispatch altogether?

web-common/src/features/connectors/code-utils.ts
```diff
@@ -11,6 +11,7 @@ import { runtime } from "../../runtime-client/runtime-store";
 export function compileConnectorYAML(
   connector: V1ConnectorDriver,
   formValues: Record<string, unknown>,
+  options?: { skipNoPrompt?: boolean },
```
ericpgreen2: "skipNoPrompt" is a confusing concept. Can you think of a cleaner way?

(I think defining a dynamic form (or multiple forms) in a JSON schema will solve for this complexity.)

ericpgreen2 (dismissed):
UXQA:
1. I'd expect the YAML keys in the preview to be ordered in the same sequence as the form. E.g. `host` and `password` should be at the top, not at the bottom:
    image
2. Nit: debatably, we should omit `managed: false` when it's false.

Here's a couple more findings not related to this precise PR, but I want to note:
1. I'd expect the error message to be above the form, not above the preview.
    image
2. The "optional" text has a different font style between the Text input and the Checkbox input.
    image

lovincyrus:
> 1. I'd expect the error message to be above the form, not above the preview.
>    image

Please refer to https://www.figma.com/design/9OCPgWkDL4mAQzaSQxV7Lt?node-id=26151-6499#1342944104

lovincyrus:
> 2\. The "optional" text has a different font style between the Text input and the Checkbox input.
>     image

Addressed in https://github.com/rilldata/rill/pull/7628/commits/aa54ef0d4d8b31d0141ef39868818dd2fa51a1e1

web-common/src/components/forms/Tabs.svelte
lovincyrus: Addressed

lovincyrus:
RE: https://github.com/rilldata/rill/pull/7628#discussion_r2211294771

> This affects the CLI too. E.g. see NoPrompt in configure.go. So it feels like we shouldn't change this.

I've updated the UI form to exclude NoPrompt. Properties such as clickhouse's `managed` and salesforce `client_id` are now hidden from the UI.

web-common/src/features/connectors/code-utils.ts
lovincyrus: Replaced with `options?: { fieldFilter?: (property: PropertyType) => boolean }` instead

runtime/drivers/druid/druid.go
lovincyrus: I see, if that's the case. I'll revert these NoPrompt: true removal and do not filter out these NoPrompt from the UI.

lovincyrus: https://linear.app/rilldata/issue/APP-209/hide-noprompt-from-ui

lovincyrus:
> 1. I'd expect the YAML keys in the preview to be ordered in the same sequence as the form. E.g. `host` and `password` should be at the top, not at the bottom:

Addressed in https://github.com/rilldata/rill/pull/7628/commits/1400a509e80ee6fd11254257d971bae7d5cde0ba

lovincyrus:
Ready for +1

web-common/src/features/connectors/code-utils.ts
```diff
@@ -20,6 +25,18 @@ type: connector
...
+  let properties =
+    options?.orderedProperties ??
+    connector.configProperties ??
+    connector.sourceProperties ??
+    [];
+
+  // Optionally filter properties
+  if (options?.fieldFilter) {
+    properties = properties.filter(options.fieldFilter);
+  }
```
ericpgreen2, line 38: This complexity still makes me nervous, but let's just take it as motivation to move to the JSON Schema going forward

web-common/src/features/sources/modal/AddDataForm.svelte
```diff
@@ -150,6 +156,46 @@
   // Emit the submitting state to the parent
   $: dispatch("submitting", { submitting });
 
+  // Generate YAML preview from form state
+  $: yamlPreview = (() => {
+    if (connector.name === "clickhouse") {
```
ericpgreen2, line 165: It's not immediately obvious to me why the logic would be different for ClickHouse than it would be for other connectors. Could you add a comment explaining why? Is this a short-term thing?

ericpgreen2 (approved):
Approving with a couple comments for you to merge at your discretion.

UXQA: When I touch an optional field, input some text, then delete that text, the YAML key remains in the YAML preview. The key has no value (e.g. `output_location: ""`) and, upon submitting the form, the generated YAML also has no value. In the case I tested ([see Jam](https://jam.dev/c/05407435-c76e-442e-a25c-78d5ae7c5a6f)), it caused the generated file to fail reconciliation.

web-common/src/features/connectors/code-utils.ts
lovincyrus: Of course, this will be refactored when we have https://linear.app/rilldata/issue/ENG-814/form-generation-using-template-api

web-common/src/features/sources/modal/AddDataForm.svelte
lovincyrus: I understand your concern. The decision to separate Clickhouse from Data Form was made to address the distinct responsibilities of their respective UI rendering processes. This change was primarily driven by the need to accommodate the conditional rendering of the "managed" field.﻿

ericpgreen2: Okay, again sounds like a JSON Schema might help us out here

ericokuma:
I guess if we want to keep the ref doc, i would rather have it be located in the help section

lovincyrus:
> I guess if we want to keep the ref doc, i would rather have it be located in the help section

We should and can address this in a separate branch! https://linear.app/rilldata/issue/APP-231/clean-up-connector-preview-reference-doc

PR #7632, opened 2025-07-15
feat: editing report rows and columns

AdityaHegde:
1. Adds a section to edit rows and columns of reports.
2. Add a scroll and match with alerts height. With rows/columns height can exceed screen height and cut off the cancel/create buttons.

Closes APP-195

**Checklist:**
- [x] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-common/src/features/dashboards/aggregation-request-builder.ts
+++ b/web-common/src/features/dashboards/aggregation-request-builder.ts
@@ -1,4 +1,4 @@
-import { getAggregationDimensionFromTimeDimension } from "@rilldata/web-common/features/dashboards/aggregation-request/dimension-utils.ts";
+import { getAggregationDimensionFromFieldName } from "@rilldata/web-common/features/dashboards/aggregation-request/dimension-utils.ts";
 import { mergeDimensionAndMeasureFilters } from "@rilldata/web-common/features/dashboards/filters/measure-filters/measure-filter-utils.ts";
 import {
   COMPARISON_DELTA,
@@ -89,8 +89,8 @@ export const aggregationRequestWithRowsAndColumns = ({
   return (aggregationRequest: V1MetricsViewAggregationRequest) => {
     const allFields = new Set<string>([...rows, ...columns]);
     const isFlat = rows.length === 0;
-    const pivotOn: string[] = [];
 
+    // Get measures defined as columns. We do allow adding measures as rows so need to check it.
     const measures = columns
       .filter((col) => exploreSpec.measures?.includes(col))
       .flatMap((measureName) => {
@@ -105,27 +105,29 @@ export const aggregationRequestWithRowsAndColumns = ({
 
         return group;
       });
-    const dimensions: V1MetricsViewAggregationDimension[] = rows.map((d) =>
-      getAggregationDimensionFromTimeDimension(d, selectedTimezone),
+
+    // Get dimensions defined as rows
+    const dimensionsFromRows: V1MetricsViewAggregationDimension[] = rows.map(
+      (row) => getAggregationDimensionFromFieldName(row, selectedTimezone),
     );
-    columns
+
+    // Get dimensions defined as columns
+    const dimensionsFromColumns: V1MetricsViewAggregationDimension[] = columns
       .filter((col) => !exploreSpec.measures?.includes(col))
-      .forEach((col) => {
-        if (exploreSpec.dimensions?.includes(col)) {
-          dimensions.push({ name: col });
-          if (!isFlat) pivotOn.push(col);
-          return;
-        }
+      .map((col) =>
+        getAggregationDimensionFromFieldName(col, selectedTimezone),
+      );
 
-        const dimension = getAggregationDimensionFromTimeDimension(
-          col,
-          selectedTimezone,
-        );
-        dimensions.push(dimension);
-        if (!isFlat) pivotOn.push(dimension.alias ?? dimension.name!);
-      });
+    // only add column dimensions as pivot if it is a non-flat view
+    const pivotOn = !isFlat
+      ? dimensionsFromColumns.map((d) => d.alias ?? d.name!)
+      : [];
 
-    const sort = getUpdatedAggregationSort(
+    // Get the full list of dimensions
+    const dimensions = [...dimensionsFromRows, ...dimensionsFromColumns];
+
+    // Get the updated sort based on the new measures and dimensions
+    const updatedAggregationSort = getUpdatedAggregationSort(
       aggregationRequest,
       measures,
       dimensions,
@@ -138,7 +140,7 @@ export const aggregationRequestWithRowsAndColumns = ({
       measures,
       dimensions,
       pivotOn: !pivotOn.length ? undefined : pivotOn,
-      sort,
+      sort: updatedAggregationSort,
     };
   };
 };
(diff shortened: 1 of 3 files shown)


ericpgreen2 (changes requested):
UXQA:
1. @mindspank, can we consider some sort of section description or informational tooltip? I feel I'm a bit puzzled looking at "Rows" and "Columns". Maybe I'm just used to the current application, but, in going through this flow, I intend to export what I see on the screen. So, without context, I'm a bit thrown off.
2. @jkhwu, may I please request a design pass? There's a lot of font variance. In the following stretch of sections... Row Limit, Include metadata, Filters, Rows, Columns... I count 5 sections and 4 different font styles.
    image
3. There should be some sort of message when there are no rows (or columns) selected. Rather than a blank space.
4. From the Dimension Detail Table, I scheduled a report in which I added a Time dimension. When I opened the report sent to me via email, it brought me to this view that felt broken:
    image
5. ... a continuation of 4 (and might explain it). When I clicked "download CSV" in the email, I hit this error:
    image

mindspank:
@ericpgreen2 agree, I'll defer to @jkhwu on design changes

mindspank:
4,5 seems unrelated to this PR? (But something we should fix)

ericpgreen2:
> 4,5 seems unrelated to this PR? (But something we should fix)

Sorry, I edited my message to clarify: I added the time dimension _as a "Row"_ in the Scheduled Report form.

medriscoll:
I've added comments in the related Linear issue.

web-common/src/features/scheduled-reports/BaseScheduledReportForm.svelte
```diff
@@ -30,119 +32,136 @@
   $: ({ instanceId } = $runtime);
 
   $: hasSlackNotifier = getHasSlackConnection(instanceId);
+
+  let didSubmit = false;
```
ericpgreen2: At first, I thought we should get this state from the `superForm` stores. Then I checked the docs and saw the below message about the `posted` store. Maybe you saw it too. Could you add a comment to say that we're intentionally tracking `didSubmit` manually (rather than using SuperForms)?

image

web-common/src/features/scheduled-reports/ScheduledReportDialog.svelte
```diff
@@ -145,6 +152,14 @@
       values.timeOfDay,
       values.dayOfMonth,
     );
+    const updatedQueryJson = getUpdatedAggregationRequest(
```
ericpgreen2: This function name confuses me a bit. What about `updateAggregationRequestWithUserSelections`, or something along those lines?

web-common/src/features/scheduled-reports/rows-and-columns.spec.ts
```diff
@@ -0,0 +1,530 @@
...
+    },
+    {
+      showTimeComparison: false,
+      selectedTimezone: exploreState.selectedTimezone,
+    },
+    updatedRows,
+    updatedColumns,
+    AD_BIDS_EXPLORE,
+  );
+  const cleanedNewRequest = cleanAggregationRequestForAssertion(newRequest);
```
ericpgreen2: Why is this needed? Ideally it's not needed, but at least can you add a comment?

web-common/src/features/scheduled-reports/rows-and-columns.spec.ts
```diff
@@ -0,0 +1,530 @@
...
+        runTest(testCase, (exploreState, timeControlState) =>
+          getDimensionTableAggregationRequestForTime(
+            AD_BIDS_METRICS_NAME,
+            exploreState,
+            {
+              start: timeControlState.timeStart,
+              end: timeControlState.timeEnd,
+            },
+            undefined,
+            "",
```
ericpgreen2: It's hard to know what these parameters are supposed to represent. Maybe this function should take a parameter object?

web-common/src/features/scheduled-reports/utils.ts
```diff
@@ -151,8 +170,10 @@ export function getUpdatedAggregationRequest(
   aggregationRequest: V1MetricsViewAggregationRequest,
   filtersArgs: FiltersState,
   timeControlArgs: TimeControlState,
+  rows: string[],
+  columns: string[],
   exploreSpec: V1ExploreSpec,
-) {
+): V1MetricsViewAggregationRequest {
```
ericpgreen2: Given all the parameters, maybe this should function take a parameter object too?

ericpgreen2: Additionally/alternatively, it may be a nice pipeline if this function were implemented as:
```ts
let aggRequest: ...;
aggRequest = updateAggregationRequestWithTimeRange(aggRequest, ...)
aggRequest = updateAggregationRequestWithRowsAndColumns(aggRequest, ...)
aggRequest = updateAggregationRequestWithFilters(aggRequest, ...)
```
Something like that would be easier to follow.

jkhwu:
1. @aditya.hegde in my mocks I use the same pill-shaped '+' button as in the filter bar: [https://www.figma.com/design/9OCPgWkDL4mAQzaSQxV7Lt/Rill-Design-System-2?node-id=26302-58284&t=07y1l4bXro2qUMhF-11](https://www.figma.com/design/9OCPgWkDL4mAQzaSQxV7Lt/Rill-Design-System-2?node-id=26302-58284&t=07y1l4bXro2qUMhF-11)
2. Agree, the cancel or X buttons should suffice for closing out of this dialog.
3. no comment from design, sounds like a bug
4. Happy to support whatever addition or clarification you decide on @mindspank

web-common/src/features/scheduled-reports/BaseScheduledReportForm.svelte
AdityaHegde: Switching to using validation errors instead. Setting validationMethod to "auto" gives the behaviour we want.

web-common/src/features/scheduled-reports/rows-and-columns.spec.ts
AdityaHegde: These were old functions. Updated all these aggregation request builders to take object parameter.

web-common/src/features/scheduled-reports/rows-and-columns.spec.ts
AdityaHegde: Added a comment.

web-common/src/features/scheduled-reports/utils.ts
AdityaHegde: Good idea. I have been thinking of adding a builder for aggregation request for a while. I guess this is a good time to start. We can refactor other places to use this builder so that all the code is in a common place.

mindspank:
> 4. Happy to support whatever addition or clarification you decide on @mindspank. 

Good feedback, I would keep it as is for now as we are re-thinking Report and Alert creation.

AdityaHegde:
1. Sure @jkhwu updating to have the same look as the add filter button.
2. Updated.
3. @ericpgreen2 was there data for the time you ran the report for? I am using a project that has future data to avoid running into this. Otherwise make sure you pulled the changes and restarted the servers.

ericpgreen2:
> 3\. @ericpgreen2 was there data for the time you ran the report for? I am using a project that has future data to avoid running into this. Otherwise make sure you pulled the changes and restarted the servers.

Ah! Yes, I was running a report on the OpenRTB dataset that doesn't have current data.

web-common/src/features/dashboards/aggregation-request-builder.ts
```diff
@@ -0,0 +1,180 @@
...
+
+        const dimension = getAggregationDimensionFromTimeDimension(
+          col,
+          selectedTimezone,
+        );
+        dimensions.push(dimension);
+        if (!isFlat) pivotOn.push(dimension.alias ?? dimension.name!);
+      });
+
+    const sort = getUpdatedAggregationSort(
```
ericpgreen2: Naming nit: `updateAggregationSort`?

web-common/src/features/dashboards/aggregation-request-builder.ts
```diff
@@ -0,0 +1,180 @@
...
+      undefined,
+    );
+    return {
+      ...aggregationRequest,
+      where: whereFilter,
+    };
+  };
+};
+
+export const aggregationRequestWithRowsAndColumns = ({
```
ericpgreen2, line 76: A few comments on this function:
- Sectional comments would help
- We should prefer using `.map` rather than `.forEach` for data transformations
- I think it'd be more clear to have 1 `isFlat` conditional rather than 2

web-common/src/features/scheduled-reports/utils.ts
```diff
@@ -147,34 +143,38 @@ export function getFiltersAndTimeControlsFromAggregationRequest(
...
-    timeControlArgs.selectedComparisonTimeRange,
-    timeControlArgs.showTimeComparison,
-    timeRange,
-  );
+  const pivotedOn = new Set<string>(aggregationRequest.pivotOn ?? []);
+  const rows: string[] = [];
+  const columns: string[] = [];
+  const isFlat = aggregationRequest.pivotOn === undefined;
+
+  aggregationRequest.dimensions?.forEach((dimension) => {
```
ericpgreen2: Here too we should prefer `.map` over `.forEach`

mindspank:
Approved from product - good to cherry pick.

PR #7633, opened 2025-07-15
feat (MCP): Add `open_url` to the `MetricsViewQuery` tool result

ericpgreen2:
In progress.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/admin/deployments.go
+++ b/admin/deployments.go
@@ -540,6 +540,7 @@ func (s *Service) NewDeploymentAnnotations(org *database.Organization, proj *dat
 		orgID:              org.ID,
 		orgName:            org.Name,
 		orgBillingPlanName: orgBillingPlanName,
+		orgCustomDomain:    org.CustomDomain,
 		projID:             proj.ID,
 		projName:           proj.Name,
 		projProdSlots:      fmt.Sprint(proj.ProdSlots),
@@ -552,6 +553,7 @@ type DeploymentAnnotations struct {
 	orgID              string
 	orgName            string
 	orgBillingPlanName string
+	orgCustomDomain    string
 	projID             string
 	projName           string
 	projProdSlots      string
@@ -560,13 +562,14 @@ type DeploymentAnnotations struct {
 }
 
 func (da *DeploymentAnnotations) ToMap() map[string]string {
-	res := make(map[string]string, len(da.projAnnotations)+7)
+	res := make(map[string]string, len(da.projAnnotations)+8)
 	for k, v := range da.projAnnotations {
 		res[k] = v
 	}
 	res["organization_id"] = da.orgID
 	res["organization_name"] = da.orgName
 	res["organization_plan"] = da.orgBillingPlanName
+	res["organization_custom_domain"] = da.orgCustomDomain
 	res["project_id"] = da.projID
 	res["project_name"] = da.projName
 	res["project_prod_slots"] = da.projProdSlots
--- a/runtime/server/mcp_server.go
+++ b/runtime/server/mcp_server.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"net/url"
 	"time"
 
 	"github.com/mark3labs/mcp-go/mcp"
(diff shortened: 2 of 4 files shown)


web-common/src/runtime-client/gen/resolvers/metrics/schema.ts
ericpgreen2, line 1: I like this. Could you just make sure to add the script (and any necessary dependencies) to the repo?

web-common/src/features/queries/query-types.ts
```diff
@@ -1,41 +1,21 @@
...
-}
-
 /**
  * Validates and normalizes a raw query object into a proper Query type.
  * Throws an error if the query is invalid.
  */
-export function validateQuery(rawQuery: any): Query {
+export function validateQuery(
+  rawQuery: MetricsResolverQuery,
+): MetricsResolverQuery {
```
ericpgreen2: Given your work to generate the proper `MetricsResolverQuery` type from the beginning, do we even need to do any normalization (aka type conversion) in this function? I suppose I now wouldn't expect to see this function return anything– I'd only expect it to throw an error, if any.

web-common/src/features/feature-flags.ts
```diff
@@ -58,7 +58,7 @@ class FeatureFlags {
   reports = new FeatureFlag("user", true);
   darkMode = new FeatureFlag("user", false);
   chat = new FeatureFlag("user", false);
-  dashboardChat = new FeatureFlag("user", false);
+  dashboardChat = new FeatureFlag("user", true);
```
ericpgreen2: Note: this'll need to be reverted before merge

web-common/src/features/queries/open-query.ts
```diff
@@ -260,3 +230,125 @@ async function convertQueryToExploreState(
 
   return partialExploreState;
 }
+
+function mapResolverTimeRangeToDashboardControls(
```
ericpgreen2: Thinking out loud– I wonder if we should move these mappings to be colocated with our other state mappings

web-common/src/features/queries/query-types.ts
AdityaHegde: Ya I only focused on mapping part. Only updated this to not break some things. I agree we should eventually get rid of it.

web-common/src/features/queries/open-query.ts
AdityaHegde: True `features/queries` is not the correct place for any of this IMO. I have not focused on that yet.

web-common/src/features/queries/query-types.ts
AdityaHegde: Removed this all together. This mainly validates data type which already happens in backend when the resolver is called. So we can just skip it. Explore validity will happen downstream so no need for that either.

web-common/src/runtime-client/gen/resolvers/metrics/schema.ts
AdityaHegde: This is a bit tricky. I have to manually edit the generated type to not break. Lets do a follow up

web-common/src/lib/map-utils.ts
ericpgreen2, line 1: Nit: kebab case for file name

web-common/src/runtime-client/gen/resolvers/metrics/schema.ts
ericpgreen2: Hmm, so this file is 1. Generated then 2. Edited? That makes me a bit nervous (and contradicts the comment at the top of the file). At least can you add documentation for the process? 

image

AdityaHegde: Updated the comments. While it is autogenerated it is slightly edited.

ericpgreen2: Thanks. Can you include in the comment the exact command that you ran?

admin/deployments.go
```diff
@@ -533,13 +565,14 @@ type DeploymentAnnotations struct {
...
 func (da *DeploymentAnnotations) ToMap() map[string]string {
-	res := make(map[string]string, len(da.projAnnotations)+7)
+	res := make(map[string]string, len(da.projAnnotations)+8)
 	for k, v := range da.projAnnotations {
 		res[k] = v
 	}
 	res["organization_id"] = da.orgID
 	res["organization_name"] = da.orgName
 	res["organization_plan"] = da.orgBillingPlanName
+	res["organization_custom_domain"] = da.orgCustomDomain
```
begelundmuller: Is it necessary to pass this? Or is the `FrontendUrl` sufficient? 

Each annotation gets added to every log/trace that runtime emits, so worth minimizing the number of these

runtime/server/mcp_server.go
```diff
@@ -364,17 +365,78 @@ Example: Get the top 10 demographic segments (by country, gender, and age group)
...
+	// Get instance to access the configured frontend URL
+	instance, err := s.runtime.Instance(ctx, instanceID)
+	if err != nil {
+		return "", fmt.Errorf("failed to get instance: %w", err)
+	}
+
+	// Use the instance's configured frontend URL (which includes org/project and custom domain for cloud, or the correct localhost URL)
+	if instance.FrontendURL == "" {
+		return "", fmt.Errorf("instance %s has no frontend URL configured", instanceID)
+	}
```
begelundmuller: Probably better to handle this gracefully (i.e. not add an open URL). It might happen in test cases or if people use the MCP server during rollout before a frontend URL has been populated on all the runtimes.

admin/deployments.go
ericpgreen2: Oh, no, I think this may have been a relic from my first iteration. Removed.

runtime/server/mcp_server.go
ericpgreen2: Done 👍

web-common/src/features/queries/open-query.ts
ericpgreen2: Agree, I've moved the file to `features/explore-mappers`

PR #7645, opened 2025-07-17
dashboards as mv

pjain1:
Fixes https://linear.app/rilldata/issue/PLAT-10/dashboard-as-metrics-view-implementation

override - (add explore inline) - explore emitted by default, disable using `skip: true` under `explore: `
```yaml
type: metrics_view
parent: ad_bids
display_name: Ad bids derived 

smallest_time_grain: day
timeseries: timestamp

parent_dimensions: '*'

parent_measures:
  - count
  - revenue

explore:
  ...
```

**Checklist:**
- [x] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-local/tests/metrics-editor.spec.ts
+++ b/web-local/tests/metrics-editor.spec.ts
@@ -88,7 +88,7 @@ test.describe("Metrics editor", () => {
     // the editor should show a validation error
     await expect(
       page.getByText(
-        'must set a value for either the "model" field or the "table" field',
+        'must set a value for either the "model", "table" or "parent" field',
       ),
     ).toBeVisible();
 


mindspank:
Dont forget to update: https://github.com/rilldata/rill/blob/main/runtime/parser/schema/project.schema.yaml

pjain1:
> Dont forget to update: https://github.com/rilldata/rill/blob/main/runtime/parser/schema/project.schema.yaml

Sure will do after initial review

proto/rill/runtime/v1/resources.proto
```diff
@@ -185,6 +185,7 @@ message MetricsView {
 }
 
 message MetricsViewSpec {
+  string parent = 29; // Parent metrics view, if this is a child metrics view.
```
begelundmuller: Nits:
1. Can you move this below the nested type definitions? E.g. around line 247
2. Can you add some more info about what this entails? E.g. a) that if its set, certain other fields like `dimensions` and `model` and others will only be populated in the `valid_spec`?

proto/rill/runtime/v1/resources.proto
```diff
@@ -270,6 +271,10 @@ message MetricsViewSpec {
   repeated Dimension dimensions = 6;
   // Measures in the metrics view
   repeated Measure measures = 7;
+  // Dynamic selector for dimensions from parent metrics view. Will be processed during validation, so it will always be empty in `state.valid_spec`. Can only be used if parent is set.
+  FieldSelector dimensions_selector = 30;
+  // Dynamic selector for measures from parent metrics view. Will be processed during validation, so it will always be empty in `state.valid_spec`. Can only be used if parent is set.
+  FieldSelector measures_selector = 31;
```
begelundmuller: The names sound like they may not be parent-specific. Maybe call these `parent_dimensions_selector` and `parent_measures_selector`? To clarify that it specifically applies to parent dimensions/measures when `parent` is set.

runtime/parser/parse_explore.go
```diff
@@ -19,9 +19,9 @@ type ExploreYAML struct {
 	Description          string                 `yaml:"description"`
 	Banner               string                 `yaml:"banner"`
 	MetricsView          string                 `yaml:"metrics_view"`
-	Dimensions           *FieldSelectorYAML     `yaml:"dimensions"`
-	Measures             *FieldSelectorYAML     `yaml:"measures"`
-	Theme                yaml.Node              `yaml:"theme"` // Name (string) or inline theme definition (map)
+	Dimensions           *FieldSelectorYAML     `yaml:"dimensions"` // deprecated while explore is defined inline, rely on mv dimensions selector
+	Measures             *FieldSelectorYAML     `yaml:"measures"`   // deprecated while explore is defined inline, rely on mv measures selector
```
begelundmuller: nit: The comment here makes it look like these fields are deprecated, but they aren't, they are still fully supported for stand-alone resources. And conversely are not supported for inline definitions in metrics views ("deprecated" means supported but not advised to use, which is something else).

runtime/parser/parse_metrics_view.go
```diff
@@ -81,6 +84,8 @@ type MetricsViewYAML struct {
 		KeySQL  string `yaml:"key_sql"`
 		KeyTTL  string `yaml:"key_ttl"`
 	} `yaml:"cache"`
+	ExploreYAML yaml.Node `yaml:"explore"`
+	NoExplore   bool      `yaml:"no_explore"`
```
begelundmuller: 1. These are listed under `// DEPRECATED FIELDS`, but they are not deprecated
2. Instead of two keys that are mutually exclusive, can it just be one key like `explore: false` to disable explores? A `yaml.Node` can be parsed as any type of value, there are some examples for that.
3. nit: Call the property `Explore` instead of `ExploreYAML` so it aligns with the property name (all the other property names align with their YAML property name)

runtime/parser/parse_metrics_view.go
```diff
@@ -237,7 +242,7 @@ func (p *Parser) parseMetricsView(node *Node) error {
 	if tmp.Table != "" && tmp.Model != "" {
 		return fmt.Errorf(`cannot set both the "model" field and the "table" field`)
 	}
-	if tmp.Table == "" && tmp.Model == "" {
+	if tmp.Table == "" && tmp.Model == "" && tmp.Parent == "" {
 		return fmt.Errorf(`must set a value for either the "model" field or the "table" field`)
```
begelundmuller: Update error message to mention `parent` also

runtime/parser/parse_metrics_view.go
```diff
@@ -63,7 +64,9 @@ type MetricsViewYAML struct {
 		ValidPercentOfTotal bool           `yaml:"valid_percent_of_total"`
 		TreatNullsAs        string         `yaml:"treat_nulls_as"`
 	}
-	Security *SecurityPolicyYAML
+	DimensionsSelector *FieldSelectorYAML `yaml:"dimensions_selector"`
+	MeasuresSelector   *FieldSelectorYAML `yaml:"measures_selector"`
```
begelundmuller: I don't think "selector" is a term we have exposed elsewhere in the YAML so far. It has just been an internal type name. It feels a bit weird because in the explore resource, they're just called `dimensions:` and `measures:`. But of course those names are already taken here (and we probably don't want to overload them as it'll be hard to explain the docs).

What do you think about also using names like `parent_dimensions:` and `parent_measures:` here instead? Another option, but that perhaps is not quite so clear, is to nest parent like this:
```yaml
parent:
  metrics_view: parent_name
  dimensions: '*'
  measures:
    - revenue
    - profits
```

runtime/parser/parse_metrics_view.go
```diff
@@ -260,6 +265,32 @@ func (p *Parser) parseMetricsView(node *Node) error {
 		}
 	}
 
+	if tmp.Parent != "" {
+		if len(tmp.Dimensions) > 0 || len(tmp.Measures) > 0 {
+			return fmt.Errorf("cannot define dimensions or measures in a derviced metrics view, use dimension_selector and measure_selector to select from parent %q", tmp.Parent)
```
begelundmuller: nit: misspelled `derviced`

runtime/parser/parse_metrics_view.go
```diff
@@ -260,6 +265,32 @@ func (p *Parser) parseMetricsView(node *Node) error {
...
+		}
+		if tmp.Database != "" || tmp.DatabaseSchema != "" || tmp.Table != "" || tmp.Model != "" {
+			return fmt.Errorf("cannot set data source in a derived metrics view (parent %q)", tmp.Parent)
+		}
+		if tmp.Watermark != "" {
+			return fmt.Errorf("cannot set watermark in a derived metrics view (parent %q)", tmp.Parent)
+		}
+		if tmp.Security != nil && (len(tmp.Security.Include) > 0 || len(tmp.Security.Exclude) > 0) {
+			return fmt.Errorf("cannot set includes/excludes in derived metrics view security, use dimension selectors to inherit subset from (parent %q)", tmp.Parent)
+		}
```
begelundmuller: Why not? I wonder if it could be useful in e.g. a child dashboard for a specific partner to show different dims to different users

runtime/parser/parse_metrics_view.go
```diff
@@ -628,13 +660,45 @@ func (p *Parser) parseMetricsView(node *Node) error {
...
+			Paths:    node.Paths,
+			YAML:     &tmp.ExploreYAML,
+			YAMLRaw:  yml,
+			YAMLPath: node.YAMLPath,
+		}
+
+		// parse the explore YAML node
+		err = p.parseExplore(exploreNode, node.Name)
+
+		return err
```
begelundmuller: 1. It's not safe to return an error after a resource has been inserted. See this comment above: `// NOTE: After calling insertResource, an error must not be returned. Any validation should be done before calling it.`
2. It also feels a little hacky to re-serialize the YAML and use `Node` to shim it into the `parseExplore` function.

I wonder if it'd be better to either:
1. Since there are some behavioral differences (like not supporting several properties), maybe just duplicate the relevant logic from `parse_explore.go` in this file directly? Generally I think a bit of duplication of simple code is better than complex code or hacky workarounds.
2. Or pull out the shared parts into a partial that gets called from both places? E.g. see `parse_partial_data.go` as an example.

runtime/metricsview/executor_validate.go
```diff
@@ -148,6 +150,185 @@ func (e *Executor) ValidateAndNormalizeMetricsView(ctx context.Context) (*Valida
...
+
+	// resolve dim names to actual dimension spec
+	for _, dimName := range dimNames {
+		dim, ok := allDims[dimName]
+		if !ok {
+			return nil, fmt.Errorf("dimension %q not found in parent metrics view %q", dimName, e.metricsView.Parent)
+		}
+		e.metricsView.Dimensions = append(e.metricsView.Dimensions, dim)
+	}
+	newSpec.Dimensions = e.metricsView.Dimensions
```
begelundmuller: I wonder if this logic can be simplified. Particularly, why not use `fieldselectorpb` to parse the field selector instead of accessing its internal structure?

Could it be something like this?

```suggestion
	names := make([]string, 0, len(newSpec.Dimensions))
	for _, d := range newSpec.Dimensions {
		names = append(names, d.Name)
	}
	names = fieldselectorpb.Resolve(e.metricsView.DimensionsSelector, names)
	slices.DeleteFunc(newSpec.Dimensions, func(d *runtimev1.MetricsViewSpec_Dimension) bool {
		return slices.Contains(names, d.Name)
	})
```

runtime/metricsview/executor_validate.go
```diff
@@ -148,6 +150,185 @@ func (e *Executor) ValidateAndNormalizeMetricsView(ctx context.Context) (*Valida
...
+
+	allMeasures := make(map[string]*runtimev1.MetricsViewSpec_Measure, len(newSpec.Measures))
```
begelundmuller: Similar comment as above for the handling of measures

runtime/metricsview/executor_validate.go
```diff
@@ -148,6 +150,185 @@ func (e *Executor) ValidateAndNormalizeMetricsView(ctx context.Context) (*Valida
...
+		securityRules = append(securityRules, parentAccess...)
+	}
+
+	// append field access rules from the metrics view and parent metrics view, adding field access in derived metrics view yaml is not allowed so this can only be rules added by us in the code if any
+	if len(fieldAccess) > 0 {
+		securityRules = append(securityRules, fieldAccess...)
+	}
+	if len(parentFieldAccess) > 0 {
+		securityRules = append(securityRules, parentFieldAccess...)
+	}
```
begelundmuller: Related to an earlier comment – are there some valid use cases for the derived metrics view having different field access rules? Not sure.

runtime/metricsview/executor_validate.go
```diff
@@ -148,6 +150,185 @@ func (e *Executor) ValidateAndNormalizeMetricsView(ctx context.Context) (*Valida
...
+	if len(rowFilter) > 0 {
+		// If the metrics view has a row filter, we need to AND the row filter with parent row filter
+		if len(parentRowFilter) > 0 {
+			rowFilter[0].GetRowFilter().Sql = fmt.Sprintf("(%s) AND (%s)", parentRowFilter[0].GetRowFilter().Sql, rowFilter[0].GetRowFilter().Sql)
+		}
+		securityRules = append(securityRules, rowFilter[0])
+	} else if len(parentRowFilter) > 0 {
+		// If the metrics view does not have a row filter, we need to inherit the parent row filter
+		securityRules = append(securityRules, parentRowFilter...)
+	}
```
begelundmuller: What if there are multiple row access rules? And can it not just add all of them to the output rules? Doesn't the the security engine already evaluate them separately and AND them together: https://github.com/rilldata/rill/blob/96d18d51466b108ab3be0c33957b0fcc61e81dc5/runtime/security.go#L593-L597

runtime/metricsview/executor_validate.go
```diff
@@ -148,6 +150,185 @@ func (e *Executor) ValidateAndNormalizeMetricsView(ctx context.Context) (*Valida
...
+		}
+		newSpec.FirstDayOfWeek = e.metricsView.FirstDayOfWeek
+	}
+	// If the metrics view has a first month of year, override the parent metrics view first month of year
+	if e.metricsView.FirstMonthOfYear > 0 {
+		if e.metricsView.FirstMonthOfYear < 1 || e.metricsView.FirstMonthOfYear > 12 {
+			return nil, fmt.Errorf("invalid first month of year %d in metrics view %q, must be between 1 and 12", e.metricsView.FirstMonthOfYear, e.metricsView.Parent)
+		}
+		newSpec.FirstMonthOfYear = e.metricsView.FirstMonthOfYear
+	}
```
begelundmuller: Aren't these validated in the parser already? Or should they be?

runtime/metricsview/executor_validate.go
```diff
@@ -148,6 +150,185 @@ func (e *Executor) ValidateAndNormalizeMetricsView(ctx context.Context) (*Valida
 	return res, nil
 }
 
+// ReconcileWithParentMetricsView resolves the parent metrics view and inherits all its dimensions and measures unless they are overridden in the current metrics view.
+func (e *Executor) ReconcileWithParentMetricsView(ctx context.Context) (*runtimev1.MetricsViewSpec, error) {
```
begelundmuller: Can it avoid the word "reconcile" here since that word has a very loaded meaning in our application? E.g. `ResolveParentMetricsView`?

runtime/reconcilers/metrics_view.go
```diff
@@ -108,6 +120,16 @@ func (r *MetricsViewReconciler) Reconcile(ctx context.Context, n *runtimev1.Reso
...
+	reconciled, err := e.ReconcileWithParentMetricsView(ctx)
+	if err != nil {
+		return runtime.ReconcileResult{Err: fmt.Errorf("failed to normalize parent metrics view: %w", err)}
+	}
+	if reconciled != nil {
+		// If the parent metrics view was normalized, we update the spec to match.
+		mv.Spec = reconciled
+	}
+
 	validateResult, validateErr := e.ValidateAndNormalizeMetricsView(ctx)
```
begelundmuller: Something feels a little weird here:
1. The executor was created with `mv.Spec`, not the rewritten spec, so you'd expect to have to create a new executor for the rewritten spec before calling `ValidateAndNormalizeMetricsView`.
2. Would it make sense to put parent metrics view handling into `ValidateAndNormalizeMetricsView`? Since it already says it "normalizes" the metrics view, and handling parents is really just a kind of normalization? Or are there advantages to decoupling it?

runtime/reconcilers/metrics_view.go
```diff
@@ -132,6 +154,10 @@ func (r *MetricsViewReconciler) Reconcile(ctx context.Context, n *runtimev1.Reso
 		return runtime.ReconcileResult{Err: validateErr}
 	}
 
+	// set selectors to nil now that they are resolved
+	mv.Spec.DimensionsSelector = nil
+	mv.Spec.MeasuresSelector = nil
```
begelundmuller: Shouldn't this be part of the rewriting that happens in `ReconcileWithParentMetricsView`?

runtime/reconcilers/metrics_view.go
```diff
@@ -132,6 +154,10 @@ func (r *MetricsViewReconciler) Reconcile(ctx context.Context, n *runtimev1.Reso
...
+
 	// Capture the spec, which we now know to be valid.
 	mv.State.ValidSpec = mv.Spec
 	// If there's no internal ref, we assume the metrics view is based on an externally managed table and set the streaming state to true.
```
begelundmuller: I can't put a comment here, but it looks like the `mv.State.DataRefreshedOn` does not take the parent's model into account.

runtime/reconcilers/explore.go
```diff
@@ -132,6 +132,22 @@ func (r *ExploreReconciler) validateAndRewrite(ctx context.Context, self *runtim
...
+			}
+			spec.DimensionsSelector = nil
+		}
+		if len(spec.Measures) == 0 && spec.MeasuresSelector == nil {
+			for _, m := range mv.Measures {
+				spec.Measures = append(spec.Measures, m.Name)
+			}
+			spec.MeasuresSelector = nil
+		}
+	}
```
begelundmuller: Hmm, this feels like it should not be necessary. Can't this just be solved by setting `spec.DimensionsSelector` and `spec.MeasuresSelector` to `*` in the parser, which is already the default in the explore parser? See here: https://github.com/rilldata/rill/blob/96d18d51466b108ab3be0c33957b0fcc61e81dc5/runtime/parser/parse_partial_field_selector.go#L96-L99

begelundmuller (changes requested):

runtime/parser/parse_explore.go
pjain1: Since going forward we do not recommend creating stand-alone explore so thats why deprecated

pjain1: anyways changing to `not supported when explore is defined inline, use dimensions selector on metrics view`

runtime/parser/parse_metrics_view.go
pjain1: 1. I added to end after `Cache` which is also not deprecated so I may just move all including `Cache` above deprecated fields.
2. okay
3. okay

runtime/parser/parse_metrics_view.go
pjain1: Do you mean to mention that `must set a value for either the "model" field or the "table" field when "parent" is not defined`? If parent is defined then there will be no error and if its not set then the error will indicate to set table/model like it is now, mentioning parent seems redundant.

runtime/parser/parse_metrics_view.go
pjain1: Yeah was not sure what to call them, we can call them `parent_dimensions` and `parent_measures`

runtime/parser/parse_metrics_view.go
pjain1: I can add support if we want and know of a use case but AFAIR as per discussion [here](https://www.notion.so/rilldata/Dashboards-as-Metrics-Views-1d5ba33c8f5780a5af3aff2e739d92c4?source=copy_link#222ba33c8f5780b890d0fd51207bad76) we just wanted to rely on mv security policies and selectors for explore fields.

runtime/parser/parse_metrics_view.go
pjain1: Oh yeah I missed that, let me move it around so that we can still return meaningful parse errors for inline explore.

Agree I can duplicate code. However, re-serializing the YAML was not just to shim `Node` into `parseExplore` its also so that I can use `knownFields` property of yaml decoder to throw parse error for unknown fields which is not possible with decoding inline explore's `yaml.Node` to `ExploreYAML` which is present inside decoded `MetricsViewYAML`. Let me know if you have a good solution for that.

runtime/metricsview/executor_validate.go
pjain1: Thanks will use `fieldselectorpb.Resolve`. `slices. DeleteFunc` leaves nil entries in slice causing panic later.

runtime/metricsview/executor_validate.go
pjain1: yeah not validated, not sure why. I can add that

runtime/parser/parse_metrics_view.go
begelundmuller: Hmm, not a big deal, but in theory maybe the user wanted to set `parent` but forgot it? So was thinking something like:
```suggestion
	if tmp.Table == "" && tmp.Model == "" && tmp.Parent == "" {
		return fmt.Errorf(`must set a value for either the "model", "table" or "parent" field`)
```

runtime/parser/parse_metrics_view.go
begelundmuller: Oh is this related to explore fields? I thought this was related to the derived metrics view's security policy (I agree that we don't need separate explore security policies).

I'm not sure if it's an important use case, but what I'm thinking is that you may want to create a derived metrics view where non-admins can only see a subset of fields. Is that supported already?

runtime/parser/parse_metrics_view.go
begelundmuller: I see, yeah that maybe gets a bit tricky. Maybe it's simplest just to have it as an inline struct? Then it won't support `explore: {true,false}`, but maybe a nested property like `explore.skip: {true,false}` is fine? Like this:
```go
type MetricsViewYAML struct {
  // ... current fields

  Explore struct {
    Skip bool `yaml:"skip"`
    Banner string `yaml:"banner"`
    ...
  } `yaml:"explore"`
}

```
Just a thought, not sure if it accounts for everything.

runtime/reconcilers/metrics_view.go
pjain1: Moved resolution of parent metrics view inside `ValidateAndNormalizeMetricsView`

runtime/parser/parse_metrics_view.go
pjain1: May be someone can ask for it I guess just appending the parent and child field access rules should do the correct thing or more sophisticated deduplication required?

runtime/metricsview/executor_validate.go
pjain1: https://github.com/rilldata/rill/pull/7645#discussion_r2230934937

runtime/reconcilers/metrics_view.go
pjain1: yes can be done there now

runtime/metricsview/executor_validate.go
pjain1: ok interesting then that would simplify the logic here

runtime/reconcilers/explore.go
pjain1: makes sense, can be done

runtime/reconcilers/metrics_view.go
pjain1: handled now

runtime/parser/parse_metrics_view.go
pjain1: ok makes sense

runtime/parser/parse_metrics_view.go
begelundmuller: Yeah maybe appending is enough. Not sure if there's an edge case where the parent's field rules include a field that is not covered in the child metrics view's dimension/measure selector?

pjain1: Sorry for the confusion you are correct this refer to dimensions/measures in derived metrics view not explore, I removed this check. Now include/exclude should just work.

runtime/metricsview/executor_validate.go
pjain1: fixed see the above thread

runtime/metricsview/executor_validate.go
```diff
@@ -148,6 +155,120 @@ func (e *Executor) ValidateAndNormalizeMetricsView(ctx context.Context) (*Valida
...
+		return fmt.Errorf("failed to resolve parent dimensions selector: %w", err)
+	}
+	filteredDims := make([]*runtimev1.MetricsViewSpec_Dimension, 0, len(parent.Dimensions))
+	for _, d := range parent.Dimensions {
+		if slices.Contains(names, d.Name) {
+			filteredDims = append(filteredDims, d)
+		} else {
+			return fmt.Errorf("cannot find dimension %q in parent metrics view %q", d.Name, e.metricsView.Parent)
+		}
+	}
```
begelundmuller: Isn't the last error off here? It would just mean that `d.Name` has been filtered out right, i.e. not a validation error?

runtime/metricsview/executor_validate.go
```diff
@@ -148,6 +155,120 @@ func (e *Executor) ValidateAndNormalizeMetricsView(ctx context.Context) (*Valida
...
+		return fmt.Errorf("failed to resolve parent measures selector: %w", err)
+	}
+	filteredMeasures := make([]*runtimev1.MetricsViewSpec_Measure, 0, len(parent.Measures))
+	for _, m := range parent.Measures {
+		if slices.Contains(names, m.Name) {
+			filteredMeasures = append(filteredMeasures, m)
+		} else {
+			return fmt.Errorf("cannot find measure %q in parent metrics view %q", m.Name, e.metricsView.Parent)
+		}
+	}
```
begelundmuller: Same question about the error here

runtime/metricsview/executor_validate.go
```diff
@@ -148,6 +155,120 @@ func (e *Executor) ValidateAndNormalizeMetricsView(ctx context.Context) (*Valida
...
+		if rule.GetAccess() != nil && hasAccessRule {
+			continue // skip access rules if already defined in the metrics view
+		}
+		e.metricsView.SecurityRules = append(e.metricsView.SecurityRules, rule)
+	}
+
+	// If the metrics view has a time dimension, override the parent metrics view time dimension
+	if e.metricsView.TimeDimension == "" {
+		e.metricsView.TimeDimension = parent.TimeDimension
+	}
```
begelundmuller: Comment is wrong?
```suggestion
	// If the metrics view doesn't have a time dimension, use the parent's time dimension
	if e.metricsView.TimeDimension == "" {
		e.metricsView.TimeDimension = parent.TimeDimension
	}
```

runtime/metricsview/executor_validate.go
```diff
@@ -148,6 +155,120 @@ func (e *Executor) ValidateAndNormalizeMetricsView(ctx context.Context) (*Valida
...
+	// If the metrics view has a smallest time grain, override the parent metrics view smallest time grain
+	if e.metricsView.SmallestTimeGrain != runtimev1.TimeGrain_TIME_GRAIN_UNSPECIFIED {
+		if e.metricsView.SmallestTimeGrain < parent.SmallestTimeGrain {
+			return fmt.Errorf("invalid smallest time grain %s in metrics view %q, must be greater than or equal to parent metrics view smallest time grain %s", e.metricsView.SmallestTimeGrain, e.metricsView.Parent, parent.SmallestTimeGrain)
+		}
+	} else {
+		e.metricsView.SmallestTimeGrain = parent.SmallestTimeGrain
+	}
+	// If the metrics view has ai instructions, override the parent metrics view ai instructions
```
begelundmuller: Same comment about the comments here

runtime/parser/parse_explore.go
```diff
@@ -19,9 +19,9 @@ type ExploreYAML struct {
 	Description          string                 `yaml:"description"`
 	Banner               string                 `yaml:"banner"`
 	MetricsView          string                 `yaml:"metrics_view"`
-	Dimensions           *FieldSelectorYAML     `yaml:"dimensions"`
-	Measures             *FieldSelectorYAML     `yaml:"measures"`
-	Theme                yaml.Node              `yaml:"theme"` // Name (string) or inline theme definition (map)
+	Dimensions           *FieldSelectorYAML     `yaml:"dimensions"` // not supported when explore is defined inline, use dimensions selector on metrics view
+	Measures             *FieldSelectorYAML     `yaml:"measures"`   // not supported when explore is defined inline, use measures selector on metrics view
```
begelundmuller: I think these comment should be removed now since the `ExploreYAML` type is not used inside metrics views anymore.

runtime/parser/parse_explore.go
```diff
@@ -36,7 +36,7 @@ type ExploreYAML struct {
 	Embeds struct {
 		HidePivot bool `yaml:"hide_pivot"`
 	} `yaml:"embeds"`
-	Security *SecurityPolicyYAML `yaml:"security"`
+	Security *SecurityPolicyYAML `yaml:"security"` // not supported when explore is defined inline, rely on mv security rules
```
begelundmuller: Same comment

runtime/parser/parse_metrics_view.go
```diff
@@ -1,12 +1,14 @@
...
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 	"time"
 
 	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
 	"github.com/rilldata/rill/runtime/pkg/rilltime"
+	"golang.org/x/exp/maps"
```
begelundmuller, line 11: I think you can just import `"maps"`

runtime/parser/parse_metrics_view.go
```diff
@@ -260,6 +286,32 @@ func (p *Parser) parseMetricsView(node *Node) error {
...
+	if tmp.Parent != "" {
+		if len(tmp.Dimensions) > 0 || len(tmp.Measures) > 0 {
+			return fmt.Errorf("cannot define dimensions or measures in a derived metrics view, use dimension_selector and measure_selector to select from parent %q", tmp.Parent)
+		}
+		if tmp.Database != "" || tmp.DatabaseSchema != "" || tmp.Table != "" || tmp.Model != "" {
+			return fmt.Errorf("cannot set data source in a derived metrics view (parent %q)", tmp.Parent)
+		}
+		if tmp.Watermark != "" {
+			return fmt.Errorf("cannot set watermark in a derived metrics view (parent %q)", tmp.Parent)
+		}
```
begelundmuller: If we allow changing the time dimension, we should probably also allow changing the watermark expression (they are related).

runtime/parser/parse_metrics_view.go
```diff
@@ -260,6 +286,32 @@ func (p *Parser) parseMetricsView(node *Node) error {
...
+		if tmp.Cache.Enabled != nil || tmp.Cache.KeySQL != "" || tmp.Cache.KeyTTL != "" {
+			return fmt.Errorf("cannot set cache in a derived metrics view (parent %q)", tmp.Parent)
+		}
```
begelundmuller: Mayne I missed it, but I don't think watermark or cache setting get inherited in `resolveParentMetricsView`?

runtime/parser/parse_metrics_view.go
```diff
@@ -582,13 +644,25 @@ func (p *Parser) parseMetricsView(node *Node) error {
...
+		return fmt.Errorf("failed to parse inline explore: %w", err)
+	}
+
+	// insert metrics view resource immediately after parsing the inline explore as it inserts the explore resource so we should not return an error now
 	r, err := p.insertResource(ResourceKindMetricsView, node.Name, node.Paths, node.Refs...)
 	if err != nil {
+		// If we fail to insert the metrics view, we must delete the inline explore if it was created.
+		if exploreRes != nil {
+			p.deleteResource(exploreRes)
+		}
```
begelundmuller: Deleting like this is not quite safe – instead use `insertDryRun` inside `parseInlineExplore` to validate. See this code from inline components in canvas for an example: https://github.com/rilldata/rill/blob/7810a980fef44e7242e36f95f0ea6bf8781d40db/runtime/parser/parse_canvas.go#L248-L322

runtime/reconcilers/metrics_view.go
```diff
@@ -137,6 +152,9 @@ func (r *MetricsViewReconciler) Reconcile(ctx context.Context, n *runtimev1.Reso
...
 	mv.State.Streaming = !hasInternalRef
 	// We copy the underlying model's refreshed_on timestamp to the metrics view state since dashboard users may not have access to the underlying model resource.
+	if parentDataRefreshedOn != nil {
+		dataRefreshedOn = parentDataRefreshedOn
+	}
 	mv.State.DataRefreshedOn = dataRefreshedOn
```
begelundmuller: Won't the second one just overwrite the first one?

Maybe remove the `parentDataRefreshedOn` and just assign into `dataRefreshedOn` in the `if mv.Spec.Parent != "" {` if statement?

begelundmuller:
Also copying this request from @mindspank :
> If we want to be able to non-breaking migrate old projects to this syntax, then I think supporting a custom name would be best:
```
type: metrics_view
...
explore:
  name: custom_name
```

runtime/metricsview/executor_validate.go
pjain1: Oh yeah I switched from cloning parent mv and editing its properties to handling derived metrics view properties to make it robust.

runtime/parser/parse_metrics_view.go
pjain1: Yeah I saw that but it is same as in [parse_explore](https://github.com/rilldata/rill/blob/main/runtime/parser/parse_explore.go#L11) and standard `maps.Keys` returns a `iter.Seq[K]` instead of `[]string`, I think thats why it was used in other places in first place.

runtime/parser/parse_metrics_view.go
pjain1: makes sense

runtime/parser/parse_metrics_view.go
pjain1: yeah its missing added

runtime/reconcilers/metrics_view.go
pjain1: correct

runtime/parser/schema/project.schema.yaml
```diff
@@ -1289,6 +1330,82 @@ definitions:
...
+          additionalProperties: false
+          embeds:
+            type: object
+            description: Configuration options for embedded explore views.
+            properties:
+              hide_pivot:
+                type: boolean
+                description: When true, hides the pivot table view in embedded mode.
+            additionalProperties: false
+      additionalProperties: false
```
begelundmuller: I think the first `additionalProperties: false` here is nested incorrectly. Can you also run `make docs.generate` and check the changed Markdown docs looks good?

runtime/parser/schema/project.schema.yaml
```diff
@@ -1240,13 +1243,51 @@ definitions:
...
             $ref: '#/definitions/security_policy_properties'
             description: Defines a security policy for the dashboard
+          explore:
+            $ref: '#/definitions/explore_properties'
+            description: Defines an inline explore view for the metrics view
         required:
           - type
       - $ref: '#/definitions/common_properties' 
     definitions:
+      fields_selector:
```
begelundmuller: Does this duplicate `field_selector_properties` from line 1057? Should it perhaps be pulled out as a shared type instead?

runtime/parser/parse_metrics_view.go
```diff
@@ -704,6 +786,168 @@ func (p *Parser) parseMetricsView(node *Node) error {
 	return nil
 }
 
+// parseInlineExplore parses and validates the inline explore definition in a metrics view YAML. It returns true if automatic explore emission should be skipped, false otherwise.
+func (p *Parser) parseInlineExplore(tmp *MetricsViewYAML, mvName string, mvPaths []string) (bool, *Resource, error) {
```
begelundmuller: can we call this `parseAndInsertInlineExplore`? Since insertion is pretty sensitive, would be good to make it clear

runtime/parser/parse_metrics_view.go
begelundmuller: Calling `deleteResource` is not safe, it should only be called from the main parser loop. 

It needs to return the error if a dry run was not done earlier. If a dry run was already done (if explore was not skipped), I think it should panic because it's in an inconsistent state then.

begelundmuller (changes requested):

runtime/parser/parse_metrics_view.go
pjain1: Forgot to change this one, already doing dry insert for mv

runtime/parser/schema/project.schema.yaml
```diff
@@ -1289,6 +1271,76 @@ definitions:
...
+          banner:
+            type: string
+            description: Custom banner displayed at the header of the explore view.
+          theme:
+            oneOf:
+              - type: string
+                description: Name of an existing theme to apply to the explore view.
+              - $ref: '#/definitions/theme/definitions/theme_properties'
+                description: Inline theme configuration.
+            description: Name of the theme to use. Only one of theme and embedded_theme can be set.
```
mindspank: Whats `embedded_theme`?

pjain1: I just copied it over from canvas and other places. But this can be confusing change it to `Name of the theme to use or define a theme inline. Either theme name or embedded theme can be set.`

runtime/parser/schema/project.schema.yaml
pjain1: ok

runtime/parser/schema/project.schema.yaml
mindspank: I would avoid the term "embedded" since it can also mean a embedded dashboard inside of a iframe

pjain1: changed to inline

mindspank:
Deferring review to @nishantmonu51 which is the PM for this effort

begelundmuller (approved):

PR #7651, opened 2025-07-17
Changes to Reference

royendo:
New order of Reference (in prep to move Connectors to Docs)
  - Project FIles
  - OLAP Engines
  - Connectors
  - CLI Usage
Rill Extensions

- Reorderd the Project Files with Connectors First (first item in the experience). 
- Added a deprecated note in source and have them look at model YAML instead
- Model YAML has a snippet on why YAML vs SQL 

Didn't add:
- MotherDuck as an OLAP engine, but it makes sense to and will do it in the future

Notes on PR:
Took the chance to recreate our **connector YAML** in preparation for the new connector experience and the importance to have this clearly defined and allow user to easily copy and paste the sample connector (or use the UI). There's a lot of parameters that we dont have documented anywhere and the **reference** should be the place for it. (not docs or relying on them to read github code)

Driver Types are defined with a in-page link to the corresponding sample YAML. This way a user can just copy and paste then insert their own creds. There is a note on each that references that they should not plain text their creds here and instead use a .env.......

### DIFF
--- a/docs/docs/reference/project-files/advanced-models.md
+++ b/docs/docs/reference/project-files/advanced-models.md
@@ -7,7 +7,7 @@ hide_table_of_contents: true
 
 :::tip
 
-Both regular models and source models can use the Model YAML specification described on this page. While [SQL models](model) are perfect for simple transformations, Model YAML files provide advanced capabilities for complex data processing scenarios.
+Both regular models and source models can use the Model YAML specification described on this page. While [SQL models](./models) are perfect for simple transformations, Model YAML files provide advanced capabilities for complex data processing scenarios.
 
 **When to use Model YAML:**
 - **Partitions** - Optimize performance with data partitioning strategies
--- a/docs/docs/reference/project-files/advanced-models.md
+++ b/docs/docs/reference/project-files/advanced-models.md
@@ -1,11 +1,24 @@
 ---
-title: Advanced Model YAML
-sidebar_label: Advanced Model YAML
-sidebar_position: 35
+title: Model YAML
+sidebar_label: Model YAML
+sidebar_position: 20
 hide_table_of_contents: true
 ---
 
-In some cases, advanced models will be required when implementing advanced features such as incremental partitioned models or staging models. 
+:::tip
+
+Both regular models and source models can use the Model YAML specification described on this page. While [SQL models](model) are perfect for simple transformations, Model YAML files provide advanced capabilities for complex data processing scenarios.
+
+**When to use Model YAML:**
+- **Partitions** - Optimize performance with data partitioning strategies
+- **Incremental models** - Process only new or changed data efficiently
+- **Pre/post execution hooks** - Run custom logic before or after model execution
+- **Staging** - Create intermediate tables for complex transformations
+- **Output configuration** - Define specific output formats and destinations
+
+Model YAML files give you fine-grained control over how your data is processed and transformed, making them ideal for production workloads and complex analytics pipelines.
+
+:::
 
 
 ## Properties
--- a/docs/docs/reference/project-files/connectors.md
+++ b/docs/docs/reference/project-files/connectors.md
@@ -39,7 +39,7 @@ When connecting to a data source, you can either explicitly define the connectio
 <!-- - **[local_file](#local_file)** - Local files (CSV, Parquet, etc.) -->
 - **[Salesforce](#salesforce)** - Salesforce data
 - **[Slack](#slack)** - Slack data
-- **[Google Sheets](#googlesheets) - Public Google Sheets
+- **[Google Sheets](#google-sheets)** - Public Google Sheets
 
 ### OLAP Engine Drivers
 When connecting to your own OLAP engine (e.g., ClickHouse, Druid, or Pinot), Rill will automatically generate the corresponding connector file and add the `olap_connector` parameter to your `rill.yaml` file. This will change the behavior of your Rill Developer slightly as not all features are supported across engines. Please see our documentation about [olap-engines](/reference/olap-engines/) for more information.
@@ -229,6 +229,13 @@ type: connector                                  # Must be `connector` (required
 driver: sqlite                                   # Must be `sqlite` _(required)_
 
 dsn: "./data/database.db"                        # SQLite connection DSN _(required)_
+```
+
+### Google Sheets
+```yaml
+type: model                                      # Slightly different than the others `model`
+connector: "duckdb"                              # connector will use default DuckDB engine
+sql: "select * from read_csv_auto('https://docs.google.com/spreadsheets/d/<SPREADSHEET_ID>/export?format=csv&gid=<SHEET_ID>', normalize_names=True)"                           # Fill in the parameters with your public Google Sheet.
 ```
 
 ## OLAP Engine Parameters
--- a/docs/docs/reference/project-files/connectors.md
+++ b/docs/docs/reference/project-files/connectors.md
@@ -2,16 +2,13 @@
 title: Connector YAML
 sidebar_label: Connector YAML 
 sidebar_position: 00
-hide_table_of_contents: true
+toc_max_heading_level: 4
+
 ---
 
 Connector YAML files define how Rill connects to external data sources and OLAP engines. Each connector specifies a driver type and its required connection parameters. 
 
 For newly created projects, Rill automatically generates a `/connectors/duckdb.yaml` file. This file allows you to configure custom database connections and SQL initialization commands through the `init_sql` parameter. This approach provides transparency and control over how Rill connects to your data sources, making it easier to understand and customize your setup.
-
-:::warning Security Best Practice
-**Never store plain text credentials in your YAML files!** Instead, use environment variables and reference them with `{{.env.<connector_type>.<parameter_name>}}` syntax. See our [credentials documentation](/build/credentials/) for detailed setup instructions.
-:::
 
 ## Required Properties
 
@@ -19,12 +16,10 @@ For newly created projects, Rill automatically generates a `/connectors/duckdb.y
 
 **`driver`** - The type of connector (required)
 
-## Driver Types
-
 ### Data Source Drivers
 When connecting to a data source, you can either explicitly define the connection in a YAML file or allow Rill to automatically infer this for you based on the file type.
 
-**Data Warehouse**
+#### _Data Warehouse_
 - **[Snowflake](#snowflake)** - Snowflake data warehouse
 - **[BigQuery](#bigquery)** - Google BigQuery
 - **[Redshift](#redshift)** - Amazon Redshift
@@ -34,16 +29,17 @@ When connecting to a data source, you can either explicitly define the connectio
 - **[sqlite](#sqlite)** - SQLite databases
 - **[MotherDuck](#motherduck-source)** - MotherDuck cloud data warehouse
 
-**Cloud Storage**
+#### _Cloud Storage_
 - **[GCS](#gcs)** - Google Cloud Storage
 - **[S3](#s3)** - Amazon S3 storage
 - **[Azure](#azure)** - Azure Blob Storage
 
-**Other**
-<!-- - **[https](#https)** - Public files via HTTP/HTTPS
-- **[local_file](#local_file)** - Local files (CSV, Parquet, etc.) -->
+#### _Other_
+- **[https](#https)** - Public files via HTTP/HTTPS
+<!-- - **[local_file](#local_file)** - Local files (CSV, Parquet, etc.) -->
 - **[Salesforce](#salesforce)** - Salesforce data
-- **[slack](#slack)** - Slack data
+- **[Slack](#slack)** - Slack data
+- **[Google Sheets](#googlesheets) - Public Google Sheets
 
 ### OLAP Engine Drivers
 When connecting to your own OLAP engine (e.g., ClickHouse, Druid, or Pinot), Rill will automatically generate the corresponding connector file and add the `olap_connector` parameter to your `rill.yaml` file. This will change the behavior of your Rill Developer slightly as not all features are supported across engines. Please see our documentation about [olap-engines](/reference/olap-engines/) for more information.
@@ -58,17 +54,14 @@ When connecting to your own OLAP engine (e.g., ClickHouse, Druid, or Pinot), Ril
 [Contact us](/contact) if you need support for a specific data source or OLAP engine!
 :::
 
-## Connection Parameters
 
-Each driver type has specific connection parameters. Click on any driver above to see its detailed configuration options, or refer to the [complete connector reference](/hidden/yaml/connector) for all available parameters.
+## Data Source Parameters
 
-:::tip Security Recommendation
+:::warning Security Recommendation
 For all credential parameters (passwords, tokens, keys), use environment variables with the syntax `{{.env.<connector_type>.<parameter_name>}}`. This keeps sensitive data out of your YAML files and version control. See our [credentials documentation](/build/credentials/) for complete setup instructions.
 :::
 
-### Data Source Parameters
-
-#### Athena
+### Athena
 ```yaml
 type: connector                                  # Must be `connector` (required)
 driver: athena                                   # Must be `athena` _(required)_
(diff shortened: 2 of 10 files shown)


docs/docs/reference/project-files/advanced-models.md
```diff
@@ -1,11 +1,24 @@
...
+Both regular models and source models can use the Model YAML specification described on this page. While [SQL models](./models) are perfect for simple transformations, Model YAML files provide advanced capabilities for complex data processing scenarios.
+
+**When to use Model YAML:**
+- **Partitions** - Optimize performance with data partitioning strategies
+- **Incremental models** - Process only new or changed data efficiently
+- **Pre/post execution hooks** - Run custom logic before or after model execution
+- **Staging** - Create intermediate tables for complex transformations
+- **Output configuration** - Define specific output formats and destinations
+
+Model YAML files give you fine-grained control over how your data is processed and transformed, making them ideal for production workloads and complex analytics pipelines.
```
ericpgreen2, line 19: This is great

docs/docs/reference/project-files/sources.md
```diff
@@ -1,11 +1,21 @@
...
 hide_table_of_contents: true
 ---
 
-In your Rill project directory, create a `<source_name>.yaml` file in the `sources` directory containing a `type` and location (`uri` or `path`). Rill will automatically detect and ingest the source next time you run `rill start`.
+:::warning Deprecated Feature
+**Sources have been deprecated** and are now considered "source models." While sources remain backward compatible, we recommend migrating to the new source model format for access to the latest features and improvements.
+
+**Next steps:**
+- Continue using sources if needed (backward compatible)
+- Migrate to source models via the `type:model` parameter for new projects
```
ericpgreen2: Nit: _new_ projects don't need to "migrate" because the application will create models (not sources) by default

ericpgreen2 (changes requested):
On the Connectors YAML page:
- I notice we don't have a right-hand side table-of-contents on this page. It'd be useful to quickly scan for the various connector types.
- It feels like the hierarchy isn't totally clear. Here's one example– there are two equivalent headers stacked on top of one another:
    image
    It might be worth asking Cursor to clean up the section hierarchies and make the sections "MECE" :).
- The "Never store plain text credentials in your YAML files" advice is great, but it is communicated a few too many times :). What if we either showed it at the top-level, or only showed it per-connector (which I think I favor)?

royendo:
~~> I notice we don't have a right-hand side table-of-contents on this page. It'd be useful to quickly scan for the various connector types.~~

~~Added. ~~

~~> It feels like the hierarchy isn't totally clear. Here's one example– there are two equivalent headers stacked on top of one another:~~

~~I agree, only so many headers in markdown to be linkable... I added a fourth (#####) and maybe add italics. Removed a section as it was referring to our 'hidden' docs. This makes it a bit easier on the eyes. ~~


~~> It might be worth asking Cursor to clean up the section hierarchies and make the sections "MECE" :).~~
~~> The "Never store plain text credentials in your YAML files" advice is great, but it is communicated a few too many times :). What if we either showed it at the top-level, or only showed it per-connector (which I think I favor)?``~~

~~def overkill haha but ill add it under data source parameters and OLAP engine parameters headers~~

docs/docs/reference/project-files/connectors.md
```diff
@@ -1,68 +1,339 @@
...
+
+For newly created projects, Rill automatically generates a `/connectors/duckdb.yaml` file. This file allows you to configure custom database connections and SQL initialization commands through the `init_sql` parameter. This approach provides transparency and control over how Rill connects to your data sources, making it easier to understand and customize your setup.
+
+## Required Properties
+
+**`type`** - Must be `connector` (required)
+
+**`driver`** - The type of connector (required)
+
+### Data Source Drivers
```
ericpgreen2: To keep the terminology consistent on this page, I'd recommend "Connector Drivers"

docs/docs/reference/project-files/connectors.md
```diff
@@ -1,68 +1,339 @@
...
+When connecting to a data source, you can either explicitly define the connection in a YAML file or allow Rill to automatically infer this for you based on the file type.
```
ericpgreen2: > or allow Rill to automatically infer this for you based on the file type

This is a little hard to follow

docs/docs/reference/project-files/connectors.md
```diff
@@ -1,68 +1,339 @@
...
+
+#### _Other_
+- **[https](#https)** - Public files via HTTP/HTTPS
+<!-- - **[local_file](#local_file)** - Local files (CSV, Parquet, etc.) -->
+- **[Salesforce](#salesforce)** - Salesforce data
+- **[Slack](#slack)** - Slack data
+- **[Google Sheets](#google-sheets)** - Public Google Sheets
+
+### OLAP Engine Drivers
+When connecting to your own OLAP engine (e.g., ClickHouse, Druid, or Pinot), Rill will automatically generate the corresponding connector file and add the `olap_connector` parameter to your `rill.yaml` file. This will change the behavior of your Rill Developer slightly as not all features are supported across engines. Please see our documentation about [olap-engines](/reference/olap-engines/) for more information.
```
ericpgreen2: > When connecting to your own OLAP engine

This could be more clear that you mean "when using the Rill Developer UI..."

docs/docs/reference/project-files/connectors.md
```diff
@@ -1,68 +1,339 @@
...
+
+:::tip Looking for a different connector?
+[Contact us](/contact) if you need support for a specific data source or OLAP engine!
+:::
 
-:::tip Did you know?
 
-Starting from Rill 0.46, you can directly create OLAP engines from the UI! 
-Select + Add -> Data -> Connect an OLAP engine
+## Data Source Parameters
```
ericpgreen2: For consistency: "Driver-specific properties"

ericpgreen2: Ah, now I see that the connectors are split up into "Data Source" connectors and "OLAP engine" connectors. For this reference page, I think we should not differentiate between connector types. For example, one may use MotherDuck as an OLAP engine or as a "source" to ingest into a different OLAP engine. So, I think we should list all connectors _in alphabetical order_ in one flat section.

docs/docs/reference/project-files/connectors.md
royendo: Unless explicitly defined, Rill will [automatically detect connection details](/build/credentials/#setting-credentials-for-rill-developer) from your .env file, credentials passed with the rill start command, or CLI-based configurations.

docs/docs/reference/project-files/connectors.md
```diff
@@ -1,68 +1,337 @@
...
-hide_table_of_contents: true
+sidebar_position: 00
+toc_max_heading_level: 4
+
 ---
 
+Connector YAML files define how Rill connects to external data sources and OLAP engines. Each connector specifies a driver type and its required connection parameters. 
 
-When you add olap_connector to your rill.yaml file, you will need to set up a `<connector_name>.yaml` file in the 'connectors' directory. This file requires the following parameters,`type` and `driver` (see below for more parameter options). Rill will automatically test the connectivity to the OLAP engine upon saving the file. This can be viewed in the connectors tab in the UI.
+For newly created projects, Rill automatically generates a `/connectors/duckdb.yaml` file. This file allows you to configure custom database connections and SQL initialization commands through the `init_sql` parameter. This approach provides transparency and control over how Rill connects to your data sources, making it easier to understand and customize your setup.
```
ericpgreen2: This feels a little TMI for this reference page– feels like it more belongs in the Connect section of the docs. In general, the reference pages should focus less on context and more on just the YAML options.

docs/docs/reference/project-files/connectors.md
```diff
@@ -1,68 +1,337 @@
...
+
+Unless explicitly defined, Rill will [automatically detect connection details](/build/credentials/#setting-credentials-for-rill-developer) from your .env file, credentials passed with the rill start command, or CLI-based configurations.
+
+#### _Data Warehouse_
+- **[Snowflake](#snowflake)** - Snowflake data warehouse
+- **[BigQuery](#bigquery)** - Google BigQuery
+- **[Redshift](#redshift)** - Amazon Redshift
+- **[PostgreSQL](#postgresql)** - PostgreSQL databases
+- **[Athena](#athena)** - Amazon Athena
+- **[mysql](#mysql)** - MySQL databases
```
ericpgreen2: Nit: "mysql" -> "MySQL"

docs/docs/reference/project-files/connectors.md
```diff
@@ -1,68 +1,337 @@
...
+**`type`** - Must be `connector` (required)
 
-:::
+**`driver`** - The type of connector (required)
+
+### Connector Drivers
+
+Unless explicitly defined, Rill will [automatically detect connection details](/build/credentials/#setting-credentials-for-rill-developer) from your .env file, credentials passed with the rill start command, or CLI-based configurations.
+
+#### _Data Warehouse_
```
ericpgreen2: Postgres, MySQL, and SQLite aren't traditionally thought of as data warehouses. We could have another section titled "Databases"?

docs/docs/reference/project-files/connectors.md
```diff
@@ -1,68 +1,337 @@
...
 
+#### _Cloud Storage_
+- **[GCS](#gcs)** - Google Cloud Storage
+- **[S3](#s3)** - Amazon S3 storage
+- **[Azure](#azure)** - Azure Blob Storage
 
-## Properties
+#### _Other_
+- **[https](#https)** - Public files via HTTP/HTTPS
+<!-- - **[local_file](#local_file)** - Local files (CSV, Parquet, etc.) -->
```
ericpgreen2: This caused a gap in the list:
image

docs/docs/reference/project-files/connectors.md
```diff
@@ -1,68 +1,337 @@
...
-- _`pinot`_ link to[ Pinot documentation](https://docs.pinot.apache.org/)
+- **[DuckDB](#duckdb)** - Embedded DuckDB engine (default)
+- **[Clickhouse](#clickhouse)** - ClickHouse analytical database
+- **[MotherDuck](#motherduck)** - MotherDuck cloud database
+- **[Druid](#druid)** - Apache Druid
+- **[Pinot](#pinot)** - Apache Pinot
+
+:::tip Looking for a different connector?
+[Contact us](/contact) if you need support for a specific data source or OLAP engine!
+:::
```
ericpgreen2: This too feels like it belongs in the docs' Connect section, not here.

docs/docs/reference/project-files/connectors.md
```diff
@@ -1,68 +1,337 @@
...
-## Properties
+#### _Other_
+- **[https](#https)** - Public files via HTTP/HTTPS
+<!-- - **[local_file](#local_file)** - Local files (CSV, Parquet, etc.) -->
+- **[Salesforce](#salesforce)** - Salesforce data
+- **[Slack](#slack)** - Slack data
+- **[Google Sheets](#google-sheets)** - Public Google Sheets
 
-**`type`** - refers to the resource type and must be 'connector'
+### OLAP Engine Drivers
```
ericpgreen2: 1. This heading should also be #### – same level as the other subcategories (Data Warehouse, Databases, Cloud Storage, etc.).
2. I'd call it just "OLAP Engines"
3. Given it's arguably the most important subcategory, we could move it up to the top, above Data Warehouses?

docs/docs/reference/project-files/connectors.md
```diff
@@ -1,68 +1,337 @@
...
-Starting from Rill 0.46, you can directly create OLAP engines from the UI! 
-Select + Add -> Data -> Connect an OLAP engine
+**`type`** - Must be `connector` (required)
 
-:::
+**`driver`** - The type of connector (required)
+
+### Connector Drivers
+
+Unless explicitly defined, Rill will [automatically detect connection details](/build/credentials/#setting-credentials-for-rill-developer) from your .env file, credentials passed with the rill start command, or CLI-based configurations.
```
ericpgreen2: This feels a little out-of-place. I think we can remove it and just rely on the "Security Recommendation" section's link-to-docs?

docs/docs/reference/project-files/connectors.md
```diff
@@ -1,68 +1,337 @@
...
+```yaml
+type: connector                                  # Must be `connector` (required)
+driver: sqlite                                   # Must be `sqlite` _(required)_
+
+dsn: "./data/database.db"                        # SQLite connection DSN _(required)_
+```
+
+:::warning Security Recommendation
+For all credential parameters (passwords, tokens, keys), use environment variables with the syntax `{{.env.<connector_type>.<parameter_name>}}`. This keeps sensitive data out of your YAML files and version control. See our [credentials documentation](/build/credentials/) for complete setup instructions.
+:::
```
ericpgreen2: I'm torn on whether we need this repeated down here. Ideally we just have it in one place. For one, if we ever make edits to the copy, it'd be likely that the two instances of the same message will get out-of-sync.

ericpgreen2 (changes requested):

docs/docs/reference/project-files/connectors.md
royendo: most important being an arguement for another day but done! 🤣

docs/docs/reference/project-files/connectors.md
royendo: no hard feelings for me so ill remove it for now ( can always re add any time)

ericpgreen2 (changes requested):
Okay, I think these are my last change requests!

1. It looks a bit off that these _connector types_ are listed underneath the "Required properties" heading. I think they should be their own section with a header "Connector types" (or something like that).
    image
2. Are you positive that we're supposed to define the MotherDuck connection differently in these two cases?
    image

royendo:
1. done.
2. Yeah we have two different ways of working with MD (might need to sync with Platform to see if we deprecate legacy, but for now we support both.

Legacy is ingesting MD data into Rill via the #as-a-source and this will result in data in our Rill Embed DuckDB 
vs. 
Live connector where we'd use the user's MD (like CH and Druid) and run all of the compute on the user's side.

docs/docs/reference/project-files/connectors.md
```diff
@@ -1,68 +1,314 @@
...
 ---
 
+Connector YAML files define how Rill connects to external data sources and OLAP engines. Each connector specifies a driver type and its required connection parameters. 
 
-When you add olap_connector to your rill.yaml file, you will need to set up a `<connector_name>.yaml` file in the 'connectors' directory. This file requires the following parameters,`type` and `driver` (see below for more parameter options). Rill will automatically test the connectivity to the OLAP engine upon saving the file. This can be viewed in the connectors tab in the UI.
+## Connector Types
 
-:::tip Did you know?
+#### _OLAP Engines_
+When connecting to your own OLAP engine using Rill Developer(e.g., ClickHouse, Druid, or Pinot), Rill will automatically generate the corresponding connector file and add the `olap_connector` parameter to your `rill.yaml` file. This will change the behavior of your Rill Developer slightly as not all features are supported across engines. Please see our documentation about [olap-engines](/reference/olap-engines/) for more information.
```
ericpgreen2: missing a space in "Rill Developer (e.g., ClickHouse,...)"

docs/docs/reference/project-files/connectors.md
```diff
@@ -1,68 +1,314 @@
...
+## Required Properties
 
-**`driver`** - refers to the [OLAP engine](../olap-engines/multiple-olap.md)
-- _`clickhouse`_ link to[ ClickHouse documentation](https://clickhouse.com/docs/en/intro)
-- _`druid`_ link to[ Druid documentation](https://druid.apache.org/docs/latest/design/)
-- _`pinot`_ link to[ Pinot documentation](https://docs.pinot.apache.org/)
+**`type`** - Must be `connector` (required)
 
-:::tip A note on OLAP engines
+**`driver`** - The type of connector, see [connector types](#connector-types) (required)
```
ericpgreen2: This sort of reads like the driver could be `driver: olap_engine` or `driver: database`. One solution could be to reframe the "Connector type" section as "Available connectors"? Then the link "see available connectors" would read better.

docs/docs/reference/project-files/connectors.md
ericpgreen2: And looking back at this, I'm not sure if this block really belongs in the reference page. It feels like it more belongs in the main documentation.

royendo: haha was waiting for that one... i also thought it felt out of place (copied from original)

PR #7655, opened 2025-07-18
use describe query to get schema for clickhouse

pjain1:
Fixes https://linear.app/rilldata/issue/PLAT-129/time-dimension-validation-failing-for-ch-projects-in-067

CH does not return schema if limit is 0,  this fixes it by using describe query

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/runtime/drivers/clickhouse/olap.go
+++ b/runtime/drivers/clickhouse/olap.go
@@ -207,40 +207,39 @@ func (c *Connection) Query(ctx context.Context, stmt *drivers.Statement) (res *d
 	return res, nil
 }
 
-func (c *Connection) QuerySchema(ctx context.Context, stmt *drivers.Statement) (*runtimev1.StructType, error) {
-	// ClickHouse does not return schema with LIMIT 0, so we need to wrap query inside DESCRIBE to explicitly get the schema.
-	stmt.Query = fmt.Sprintf("DESCRIBE (%s)", stmt.Query)
+func (c *Connection) QuerySchema(ctx context.Context, query string, args []any) (*runtimev1.StructType, error) {
+	// ClickHouse does not return schema with LIMIT 0, so we need to wrap query inside DESCRIBE to explicitly get the schema. describe_compact_output returns only name and type.
+	query = fmt.Sprintf("DESCRIBE (%s) SETTINGS describe_compact_output=1", query)
 
-	res, err := c.Query(ctx, stmt)
+	if c.config.LogQueries {
+		c.logger.Info("clickhouse query", zap.String("sql", c.Dialect().SanitizeQueryForLogging(query)), zap.Any("args", args))
+	}
+
+	conn, release, err := c.acquireOLAPConn(ctx, 0)
 	if err != nil {
 		return nil, err
 	}
-	defer res.Close()
+	defer func() { _ = release() }()
 
-	cols := make([]any, len(res.Schema.Fields))
-	colPtrs := make([]any, len(cols))
-	var nameIdx, typeIndex int
-	for i := range cols {
-		colPtrs[i] = &cols[i]
-		if strings.EqualFold(res.Schema.Fields[i].Name, "name") {
-			nameIdx = i
-		}
-		if strings.EqualFold(res.Schema.Fields[i].Name, "type") {
-			typeIndex = i
-		}
+	ctx, cancelFunc := context.WithTimeout(ctx, drivers.DefaultQuerySchemaTimeout)
+	defer cancelFunc()
+
+	rows, err := conn.QueryxContext(ctx, query, args...)
+	if err != nil {
+		return nil, err
 	}
+	defer rows.Close()
 
 	schema := &runtimev1.StructType{}
-	for res.Next() {
-		if err = res.Scan(colPtrs...); err != nil {
+	var name, cType string
+	for rows.Next() {
+		if err = rows.Scan(&name, &cType); err != nil {
 			return nil, fmt.Errorf("failed to scan schema: %w", err)
 		}
-		name := cols[nameIdx].(string)
-		dtype := cols[typeIndex].(string)
 		// Convert ClickHouse data type to runtimev1.StructType_Field_Type
-		t, err := databaseTypeToPB(dtype, false)
+		t, err := databaseTypeToPB(cType, false)
 		if err != nil {
-			return nil, fmt.Errorf("failed to convert clickHouse type '%s': %w", dtype, err)
+			return nil, fmt.Errorf("failed to convert clickHouse type '%s': %w", cType, err)
 		}
 		schema.Fields = append(schema.Fields, &runtimev1.StructType_Field{
 			Name: name,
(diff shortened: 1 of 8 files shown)


runtime/metricsview/executor.go
```diff
@@ -270,6 +275,58 @@ func (e *Executor) Schema(ctx context.Context) (*runtimev1.StructType, error) {
 	return res.Schema, nil
 }
 
+func (e *Executor) resolveSchemaForClickhouse(ctx context.Context, sql string, args []any) (*runtimev1.StructType, error) {
+	schema := &runtimev1.StructType{}
```
k-anshul: Nit: Initialize closer to the usage i.e. near to the for loop

runtime/metricsview/executor.go
```diff
@@ -270,6 +275,58 @@ func (e *Executor) Schema(ctx context.Context) (*runtimev1.StructType, error) {
...
+	})
+	if err != nil {
+		return nil, err
+	}
+	defer res.Close()
+
+	cols := make([]any, len(res.Schema.Fields))
+	colPtrs := make([]any, len(cols))
+	var nameIdx, typeIndex int
+	for i := range cols {
```
k-anshul: Is this necessary ? Shouldn't the driver/db always return the columns in stable order ?

runtime/drivers/clickhouse/olap.go
```diff
@@ -323,21 +323,21 @@ func rowsToSchema(r *sqlx.Rows) (*runtimev1.StructType, error) {
 	return &runtimev1.StructType{Fields: fields}, nil
 }
 
-// databaseTypeToPB converts Clickhouse types to Rill's generic schema type.
+// DatabaseTypeToPB converts Clickhouse types to Rill's generic schema type.
 // Refer the list of types here: https://clickhouse.com/docs/en/sql-reference/data-types
-func databaseTypeToPB(dbt string, nullable bool) (*runtimev1.Type, error) {
+func DatabaseTypeToPB(dbt string, nullable bool) (*runtimev1.Type, error) {
```
k-anshul: This is sort of a very internal detail to the driver and feels hacky to expose to the resolver package. Do you think we should explore one of the following solutions?
1.  Create a view/temporary table and do a `InformationSchema.Lookup`. Obviously need to think of cleaning (in case of views) and distributed constructs.
2. Have `ResolveSchemaForClickhouse` be part of clickhouse driver. `clickhouse.Connection` is exported.

runtime/metricsview/executor.go
pjain1: That was causing confusion more with `res.Schema` so I put it above.

runtime/metricsview/executor.go
pjain1: this is to guard against the case if in future `describe` sql starts retuning more columns or changes order of columns then it will just start breaking as we do not always test with every clickhouse release.

runtime/drivers/clickhouse/olap.go
pjain1: Yeah this is tricky, I thought of exposing a method on `InformationSchema` but that was not much useful for other dialects. Creation of view/table seems too much, I would rather do `limit 1` in that case, user may not always have write privileges also. Let try moving the method to driver.

runtime/metricsview/executor.go
k-anshul: We can use `describe (SQL) settings describe_compact_output = 1`.
https://clickhouse.com/docs/operations/settings/settings#describe_compact_output

Clickhouse always has a setting LOL.

runtime/drivers/clickhouse/olap.go
k-anshul: Yeah good point about not having write privileges. Maybe let's just create exported function in clickhouse driver then.

pjain1: I created a query schema method on olap

runtime/metricsview/executor.go
pjain1: Ah nice however since I am exposing `QuerySchema` method on olap and reusing Query method so that I don't need to handle timeouts etc. I cannot add this setting. If we are very sure that column order and names won't change I can just make the row scanning static as there are only 7 columns and remove dynamic handling here.

pjain1: anyways did some refactoring have a look again

k-anshul: >  I cannot add this setting.

Why is that ? 

If that is because of settings than we can set settings multiple times in clickhouse query : `select 1 SETTINGS aggregate_functions_null_for_empty = 1 SETTINGS some_other_setting = 0`

>  If we are very sure that column order and names won't change I can just make the row scanning static as there are only 7 columns and remove dynamic handling here.

I think clickhouse can add more columns but it should be fair to assume that they will not change the ordering so we can keep the list variable depending upon schema but assume the position to be 0 and 1.

runtime/metricsview/executor_validate.go
```diff
@@ -285,17 +285,19 @@ func (e *Executor) validateTimeDimension(ctx context.Context, t *drivers.OlapTab
 			res.TimeDimensionErr = fmt.Errorf("failed to validate time dimension %q: %w", e.metricsView.TimeDimension, err)
 			return
 		}
-		// Validate time dimension type with a query
-		rows, err := e.olap.Query(ctx, &drivers.Statement{
-			Query: fmt.Sprintf("SELECT %s FROM %s LIMIT 0", expr, dialect.EscapeTable(t.Database, t.DatabaseSchema, t.Name)),
-		})
+
+		query := fmt.Sprintf("SELECT %s FROM %s LIMIT 0", expr, dialect.EscapeTable(t.Database, t.DatabaseSchema, t.Name))
+		schema, err := e.olap.QuerySchema(ctx, query, nil)
```
k-anshul, line 290: This seems somewhat weird. If I understand correctly the finally query that will be generated(for duckdb and others) will be `SELECT * FROM (SELECT ... FROM tbl LIMIT 0) LIMIT 0`

My vote will be to take one of the following approaches:
1. Remove `limit 0` from here.
2. Keep the previous logic but assume the position of `type` and `name` will not change and simplify the scanning logic. The number of fields can change.
3. Using the `describe_compact_output` setting to simplify the scanning in previous implementation.

k-anshul (approved):

runtime/metricsview/executor_validate.go
pjain1: The final query generated will depend on the olap engine - for Clickhouse it will use describe, for others it will wrap it in `SELECT * FROM (<query>) LIMIT 0`, yeah `LIMIT 0` not needed here but kept for extra safety does not harm as reconcile runs on very keystroke. For 2nd and 3rd point check [here](https://github.com/rilldata/rill/pull/7655/files#diff-ed7ae0556af79ff8926e28d6369db70f2ea35489a0eae7a9510372f8a7ea157dR210) it has simple logic and uses the setting you mentioned.

PR #7661, opened 2025-07-23
Add metrics view used in metrics_sql to canvas resolver

grahamplata:
Extract metrics views from components metrics sql. This will enable the canvas resolver to return a full list of metrics view referenced by all the components

Using the `Open RTB Ads` sample

```yaml
type: canvas
display_name: Test
defaults:
  time_range: PT24H
  comparison_mode: time
rows:
  - items:
      - kpi:
          metrics_sql: SELECT app_site_name FROM auction_metrics
        width: 12
    height: 40px
  - items:
      - bar_chart:
          metrics_view: bids_metrics
          color: hsl(246, 66%, 50%)
          x:
            type: temporal
            field: __time
            sort: -y
            limit: 20
          y:
            type: quantitative
            field: video_completion_rate
            zeroBasedOrigin: true
        width: 12

```

Sample canvas resolver 

```json
{
    "canvas": {},
    "resolvedComponents": {},
    "referencedMetricsViews": {
        "auction_metrics": {},
        "bids_metrics": {}
    }
}
```

**Checklist:**
- [x] Covered by tests
- [x] [Linked](https://linear.app/rilldata/issue/PLAT-126/add-metrics-view-used-in-metrics-sql-to-canvas-resolver) the issues it closes
- [ ] I'm proud of this work!

### DIFF
--- a/runtime/server/canvases.go
+++ b/runtime/server/canvases.go
@@ -3,8 +3,6 @@ package server
 import (
 	"context"
 	"errors"
-	"fmt"
-	"strings"
 
 	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
 	"github.com/rilldata/rill/runtime"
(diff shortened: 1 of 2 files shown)


runtime/server/canvases.go
```diff
@@ -143,84 +232,84 @@ func (s *Server) ResolveCanvas(ctx context.Context, req *runtimev1.ResolveCanvas
...
+				observability.AddRequestAttributes(ctx,
+					attribute.String("sql_parse_error.component", cmpName),
+					attribute.String("sql_parse_error.message", err.Error()),
+				)
+				// Don't fail the entire canvas resolution if one component has an invalid SQL
+				continue
+			}
+			if mvName != "" {
+				mvNames = append(mvNames, mvName)
+			}
```
begelundmuller: It's not a good idea to add the same observability attributes multiple times, and this function is anyway just optimistic and not responsible for validation, so just ignore a parse error

```suggestion
			if err == nil {
				mvNames = append(mvNames, mvName)
			}
```

runtime/server/canvases.go
```diff
@@ -143,84 +232,84 @@ func (s *Server) ResolveCanvas(ctx context.Context, req *runtimev1.ResolveCanvas
...
-	return &runtimev1.ResolveComponentResponse{
-		RendererProperties: rendererProps,
-	}, nil
+	if len(sql) > 10000 {
+		return "", fmt.Errorf("SQL query too large")
+	}
+
+	if !strings.HasPrefix(strings.ToUpper(sql), "SELECT") {
+		return "", fmt.Errorf("only SELECT statements are allowed")
+	}
```
begelundmuller: Don't do validation here, the metrics SQL parser is responsible for doing its own validation (otherwise you'd have to duplicate validation everywhere that the library is invoked)

runtime/server/canvases.go
```diff
@@ -17,6 +20,103 @@ import (
...
+
+	components, err := s.resolveCanvasComponents(ctx, ctrl, canvas, templateData)
+	if err != nil {
+		return nil, err
+	}
+
+	metricsViews, err := s.resolveMetricsViews(ctx, ctrl, req.InstanceId, components)
+	if err != nil {
+		return nil, err
+	}
```
begelundmuller: I know it's common to try to refactor functions as they grow big, but I think that's not a good idea here because the logic is quite inter-dependent as evidenced by the need to return and pass through several different variables (like `ctrl`, `canvas`, `templateData`, `req.InstanceID`, etc.). Breaking it out has made it more difficult to understand/review as I have to jump back and forth a lot to understand the context. 

The functions are also not reused (or likely to be reused) elsewhere, which is particularly problematic in a large scope such as `s *Server`, which already has probably 100s of functions (one of our largest packages). In large scopes, keeping things self-contained particularly helps readability.

For context, we generally favor readability/reviewability over strict adherence to principles like DRY or short functions. It's okay for a function to be long as long as it easy to read/review, and even okay to have light duplication if the alternative is code that becomes hard to read or unnecessarily coupled.

grahamplata: Fair enough, that was my original thinking. I have unfurled them and will clean up

runtime/server/canvases.go
```diff
@@ -107,47 +108,60 @@ func (s *Server) ResolveCanvas(ctx context.Context, req *runtimev1.ResolveCanvas
...
-					break
+		if validSpec == nil || validSpec.RendererProperties == nil {
+			continue
+		}
+
+		for k, v := range validSpec.RendererProperties.Fields {
+			switch k {
+			case "metrics_view":
+				if name := v.GetStringValue(); name != "" {
+					mvNames = append(mvNames, name)
```
begelundmuller: Could it just do `metricsViews[name] = true` directly here instead of having a separate `mvNames`?

begelundmuller (approved):

PR #7664, opened 2025-07-24
Allow more configuration in openai connector

k-anshul:
closes: https://linear.app/rilldata/issue/PLAT-139/openai-connector-make-baseurl-configurable

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/admin/server/ai.go
+++ b/admin/server/ai.go
@@ -7,7 +7,6 @@ import (
 
 	adminv1 "github.com/rilldata/rill/proto/gen/rill/admin/v1"
 	aiv1 "github.com/rilldata/rill/proto/gen/rill/ai/v1"
-	"github.com/rilldata/rill/runtime/pkg/ai"
 	"github.com/rilldata/rill/runtime/pkg/observability"
 	"go.opentelemetry.io/otel/attribute"
 )
@@ -33,7 +32,7 @@ func (s *Server) Complete(ctx context.Context, req *adminv1.CompleteRequest) (*a
 	}
 
 	// Pass messages and tools to the AI service
-	msg, err := s.admin.AI.Complete(ctx, messages, req.Tools, ai.CompletionOptions{})
+	msg, err := s.admin.AI.Complete(ctx, messages, req.Tools)
 	if err != nil {
 		return nil, err
 	}


runtime/drivers/openai/openai.go
```diff
@@ -55,13 +88,24 @@ func (d driver) Open(instanceID string, config map[string]any, st *storage.Clien
...
+	opts := &ai.Options{}
+	if conf.BaseURL != "" {
+		opts.BaseURL = conf.BaseURL
+	}
+	if conf.APIType != "" {
+		opts.APIType = openaidriver.APIType(conf.APIType)
+	}
+	if conf.APIVersion != "" {
+		opts.APIVersion = conf.APIVersion
+	}
```
begelundmuller: nit: Do you need the `if` checks? If the values are empty, setting them on a new struct value is just a noop
```suggestion
	opts := &ai.Options{
		opts.BaseURL: conf.BaseURL,
		opts.APIType: openaidriver.APIType(conf.APIType),
		opts.APIVersion: conf.APIVersion,
	}
```

runtime/pkg/ai/ai.go
```diff
@@ -6,6 +6,10 @@ import (
 	aiv1 "github.com/rilldata/rill/proto/gen/rill/ai/v1"
 )
 
+type CompletionOptions struct {
+	Model string
+}
+
 type Client interface {
-	Complete(ctx context.Context, msgs []*aiv1.CompletionMessage, tools []*aiv1.Tool) (*aiv1.CompletionMessage, error)
+	Complete(ctx context.Context, msgs []*aiv1.CompletionMessage, tools []*aiv1.Tool, opts CompletionOptions) (*aiv1.CompletionMessage, error)
```
begelundmuller: Why does it need to pass it here? Isn't the config on the constructor (`NewOpenAI`) sufficient?

In theory, the config options could be completely different for different LLM backends, so ideally config is only passed in the constructor.

begelundmuller:
Something else – you may also want to expose access to configuring the `temperature`

runtime/pkg/ai/ai.go
k-anshul: Yeah that makes sense. I overlooked this.

PR #7665, opened 2025-07-24
Validate or infer smallest_time_grain based on the time dimension data type

begelundmuller:
This PR adds validation and inference for the `smallest_time_grain` metrics view property based on the underlying data type.

The logic is rather simple – any `DATE` time dimension must have at least `day` grain, and any `TIMESTAMP` time dimension must have at least `second` grain. This PR does not handle sub-second data types since that would require a bigger refactor (that info is not currently propagated by the drivers).

If `smallest_time_grain` is not set, it will be populated using the above logic in the `state.valid_spec.smallest_time_grain` property.

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!

### DIFF
--- a/web-admin/tests/bookmarks.spec.ts
+++ b/web-admin/tests/bookmarks.spec.ts
@@ -283,7 +283,7 @@ test.describe("Bookmarks", () => {
         // make sure the url has the correct params
         assertUrlParams(
           adminPage,
-          `tr=7D+as+of+latest%2FD%2B1D&f=app_site_domain IN ('Not Available') AND pub_name IN ('Not Available')`,
+          `tr=7D+as+of+latest%2FD%2B1D&grain=day&f=app_site_domain IN ('Not Available') AND pub_name IN ('Not Available')`,
         );
       });
 


runtime/metricsview/executor/executor_validate.go
```diff
@@ -145,6 +145,39 @@ func (e *Executor) ValidateAndNormalizeMetricsView(ctx context.Context) (*Valida
...
+		if d.DataType == nil {
+			continue // Skip if data type discovery failed.
+		}
+		switch d.DataType.Code {
+		case runtimev1.Type_CODE_TIMESTAMP:
+			if smallestPossibleGrain == runtimev1.TimeGrain_TIME_GRAIN_UNSPECIFIED {
+				smallestPossibleGrain = runtimev1.TimeGrain_TIME_GRAIN_SECOND // Default for TIMESTAMP type.
+			}
+		case runtimev1.Type_CODE_DATE:
+			smallestPossibleGrain = runtimev1.TimeGrain_TIME_GRAIN_DAY
```
pjain1, line 194: Not sure how to handle this but there can be date type columns in dimension list but user may not intend to use them as primary time dimension so automatically limiting smallest grain as per this can be problematic ? BTW isn't smallest time grain millis, but it does not make much sense for dashboards.

briangregoryholmes: I don't think it should be based on _any_ time dimension, but rather the primary time dimension set by the `timeseries` key in a metrics view.

pjain1: @briangregoryholmes with multiple time dimension support, any timestamp or date type column can be used as the time dimension column for making aggregation API calls. Application will have a dropdown to select which time dimension to use.

begelundmuller: @briangregoryholmes As Parag says, with the new support for multiple time dimensions, we're trying not to give special meaning to the primary time dimension except as the default for when another time dimension is not explicitly provided.

@pjain1 I agree that this might be too restrictive. I think adding a `smallest_time_grain` option for each time dimension might be a better way to address this. What do you think?

pjain1: you mean storing mapping of `timestamp` -> `SmallestTimeGrain` instead of storing a single `SmallestTimeGrain` in the mv valid spec?

begelundmuller: I mean adding (and inferring by default) `smallest_time_grain` as a property on each timestamp dimension as well, like:
```yaml
# Default
timeseries: __time
smallest_time_grain: hour

# Per time dimension
dimensions:
- column: created_on
  smallest_time_grain: day
- column: updated_on
  smallest_time_grain: day
```

pjain1: yeah this makes sense I think

begelundmuller: Done

runtime/metricsview/executor/executor_validate.go
```diff
@@ -158,6 +158,54 @@ func (e *Executor) ValidateAndNormalizeMetricsView(ctx context.Context) (*Valida
...
 
+	// Validate or infer the smallest time grains for the default time dimension and for any additional time dimensions.
+	// We require the smallest time grain to be at least at second grain.
+	// If any time dimension has DATE type, we require the smallest time grain to be at least at day grain.
+	if e.metricsView.TimeDimension != "" {
+		// Find the smallest possible grain
+		var smallestPossibleGrain runtimev1.TimeGrain
+		col, ok := cols[strings.ToLower(e.metricsView.TimeDimension)]
+		if ok && col.Type.Code != runtimev1.Type_CODE_DATE {
+			smallestPossibleGrain = runtimev1.TimeGrain_TIME_GRAIN_SECOND
```
pjain1, line 170: in `time_grain.proto` smallest time grain possible is TIME_GRAIN_MILLISECOND, that means user can set it and then the validation later `mv.SmallestTimeGrain < smallestPossibleGrain` will fail. Since millis does not make much sense for dashboards, should we add separate check for it and just say smallest grain possible is seconds or is it fine?

begelundmuller: Good point, added explicit validation for this at parse time

briangregoryholmes (approved):

PR #7667, opened 2025-07-24
Do not hide actions cell for non-reconciling resources

lovincyrus:
This PR changes the actions cell visibility logic in the project resources table. Previously, the actions cell was hidden for the entire table when any resource was reconciling. Now, it only hides the actions cell for individual rows where that specific resource is reconciling. In addition to that, when you run an incremental refresh on one resource and open another resource's actions dropdown, both the dropdown and any subsequent dialog will stay open even when the first resource finishes running and updates the table data.

Closes https://linear.app/rilldata/issue/APP-232/project-status-page-dont-hide-model-refresh-button-when-a-different

https://github.com/user-attachments/assets/1bd7ce83-94bb-4c48-9068-28f71f58a3d2

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-admin/src/features/projects/status/ActionsCell.svelte
+++ b/web-admin/src/features/projects/status/ActionsCell.svelte
@@ -8,7 +8,7 @@
   export let resourceKind: string;
   export let resourceName: string;
   export let canRefresh: boolean;
-  export let onOpenRefreshDialog: (
+  export let onClickRefreshDialog: (
     resourceName: string,
     resourceKind: string,
     refreshType: "full" | "incremental",
@@ -29,7 +29,7 @@
         <DropdownMenu.Item
           class="font-normal flex items-center"
           on:click={() => {
-            onOpenRefreshDialog(resourceName, resourceKind, "full");
+            onClickRefreshDialog(resourceName, resourceKind, "full");
           }}
         >
           <div class="flex items-center">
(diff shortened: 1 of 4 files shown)


web-admin/src/features/projects/status/ActionsCell.svelte
```diff
@@ -3,59 +3,22 @@
...
-    }
-
-    await queryClient.invalidateQueries({
-      queryKey: getRuntimeServiceListResourcesQueryKey(
-        $runtime.instanceId,
-        undefined,
-      ),
-    });
-  }
+  export let onOpenRefreshDialog: (
```
ericpgreen2: It feels like this prop name is incorrect given this function is called `on:click` not `on:open`.

web-admin/src/features/projects/status/ProjectResourcesTable.svelte
```diff
@@ -96,13 +168,23 @@
       accessorKey: "actions",
       header: "",
       cell: ({ row }) => {
-        if (!isReconciling) {
+        // Only hide actions for reconciling rows
```
ericpgreen2, line 171: UX: this is fine but, going forward, rather than _hiding_ elements like this, preferably we:
- _Show and disable_ the element(s)
- Include an explanation for _why_ the element(s) are disabled

ericpgreen2 (approved):
Approving with a couple small comments

web-admin/src/features/projects/status/ActionsCell.svelte
lovincyrus: https://github.com/rilldata/rill/pull/7667/commits/c77e8a3f7df9827a60533278924f1bec2a553863

PR #7672, opened 2025-07-25
Git dev deployments

k-anshul:
https://linear.app/rilldata/issue/PLAT-94/git-support-for-dev-deployments

TODO: Cleanup git APIs in CLI when UI migrates to API in runtime

**Checklist:**
- [x] Covered by tests
- [ ] I'm proud of this work!


runtime/drivers/admin/repo_git.go
```diff
@@ -145,9 +149,38 @@ func (r *gitRepo) pullInner(ctx context.Context, force bool) error {
...
+		// Hard reset to remote branch (this discards all local changes)
+		err = resetToRemoteTrackingBranch(repo, worktree, r.editBranch)
+		if err != nil {
+			return fmt.Errorf("failed to reset to remote tracking branch %q: %w", r.editBranch, err)
+		}
+	}
 
+	// merge default branch into edit branch
+	if force {
+		// Maybe instead of merge with the "theirs" strategy should we just reset to the default branch?
```
k-anshul: When doing a force merge of default branch to edit branch should we just reset to default branch ?

runtime/drivers/admin/repo_git.go
```diff
@@ -167,15 +200,34 @@ func (r *gitRepo) root() string {
 
 // commitToEditBranch auto-commits any current changes to the edit branch of the repository.
 // This is done to checkpoint progress when the handle is closed.
-// If there are conflicts, it should drop any local changes.
+// If there are conflicts, it should drop any local changes. Why ??
```
k-anshul, line 203: Usually there should not be any changes in edit branch that exist on remote but doesn't exist locally but even if there are it should drop those and commit local since this will be triggered on `close` or explicit user input

cli/pkg/local/app.go
```diff
@@ -249,7 +249,12 @@ func NewApp(ctx context.Context, opts *AppOptions) (*App, error) {
 	connectors = append(connectors, olapConnector)
 
 	// The repo connector is the local project directory
-	repoConfig, err := structpb.NewStruct(map[string]any{"dsn": projectPath})
+	repoConfig, err := structpb.NewStruct(map[string]any{
+		"dsn":          projectPath,
+		"admin_url":    opts.Ch.AdminURL(),
+		"access_token": opts.Ch.AdminToken(),
+		"org":          opts.Ch.Org,
```
k-anshul: This causes an issue where the org is cached in the runtime permanently and user can end up changing org via CLI. However the same problem exists even today in local server APIs.

If required we can directly fetch the org from the `~/.rill/config.yaml` on each API call.

begelundmuller: I'm wondering if it's possible to avoid providing the `admin` info in the `file` repo driver altogether. Specifically, since this is a special case for Rill-managed Git running locally, what if the `local.NewApp` instead:
1. Checked if Rill-managed Git is used
2. If it is Rill-managed Git, fetches a fresh token and updates the remote
3. Optionally starts a background job that refreshes the remote/token every 30 mins

proto/rill/runtime/v1/api.proto
```diff
@@ -296,6 +296,32 @@ service RuntimeService {
   rpc AnalyzeVariables(AnalyzeVariablesRequest) returns (AnalyzeVariablesResponse) {
     option (google.api.http) = {get: "/v1/instances/{instance_id}/variables/analyze"};
   }
+
+  // Git APIs
```
begelundmuller, line 317: Asking here, but this can be taken up in follow-up PRs if relevant – does it also need a merge RPC? To support merging changes in a dev deployment to `main` when using Rill-managed Git (where you don't have access to PRs).

runtime/drivers/repo.go
```diff
@@ -10,6 +10,8 @@ import (
 	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
 )
 
+var ErrRemoteAhead = fmt.Errorf("remote ahead of local state, please pull first")
+
 // RepoStore is implemented by drivers capable of storing project code files.
 // All paths start with '/' and are relative to the repo root.
 type RepoStore interface {
```
begelundmuller, line 17: In order to support @nishantmonu51's request to support auto-committing and reverting from inside Rill Developer, I think some more changes are needed here:
1. Split `CommitAndPush` into separate functions, so we can auto-commit without pushing.
2. Support reverting a specific commit (not resetting, but doing a new revert commit)

Or do you prefer to handle it differently? Regardless of approach, this can probably be taken up in a separate PR.

runtime/drivers/admin/repo.go
```diff
@@ -694,6 +730,7 @@ func (r *repo) checkHandshake(ctx context.Context, force bool) error {
 		r.git.defaultBranch = meta.GitBranch
 		r.git.editBranch = meta.GitEditBranch
```
begelundmuller: Now that dev/preview deployments have an explicit branch and separate `editable` flag, I think we should remove the `editBranch` notion and have just a single branch.

And then instead have a dedicated merge RPC that the UI can call for merging from a dev deployment to the primary branch.

What do you think? I think that will also simplify some of the logic in the implementation.

proto/rill/runtime/v1/api.proto
k-anshul: Yeah I am doing that in a follow up PR.

runtime/drivers/repo.go
k-anshul: Yeah those are in a follow up PR : https://github.com/rilldata/rill/pull/8575

cli/pkg/local/app.go
k-anshul: Yeah sounds good to me. 
However we need to do it for all projects and not just rill-managed git projects. 
We saw issues with using user's git credentials for self hosted git. Some users did not have credentials properly set on their machines. 
@nishantmonu51 explicitly requested to use short lived token obtained from admin in all cases.

runtime/drivers/admin/repo.go
k-anshul: Yeah sounds good. I refactored the `pullInner` a bit as well. 
I have already added the merge RPC in the other branch.

cli/pkg/local/app.go
k-anshul: Actually on another thought, I am somewhat unconvinced about this. Keeping it in `file` repo driver ensures the token is generated only on use whereas using a background job will keep on generating token till runtime is up. A user might just keep Rill running in their terminal and it will keep generating these tokens.

k-anshul: How about modifying the interface to return a `GitRepoStore` that implements git related functionalities ?

```
type RepoStore interface {
  // the implementation in file driver will fetch git configs using the provider, admin will ignore it since it already has the configs
  AsGitRepoStore(provider GitCredentialProvider) GitRepoStore
}

type GitCredentialProvider interface {
	GetCredentials(ctx context.Context) (*GitConfig, error)
}

type GitConfig struct {
	Remote            string
	Username          string
	Password          string
	PasswordExpiresAt time.Time
	DefaultBranch     string
	Subpath           string
	ManagedRepo       bool
}

type GitRepoStore interface {
	// Status returns the current status of the repository.
	Status(ctx context.Context) (*RepoStatus, error)
	// Pull synchronizes local and remote state.
	// If discardChanges is true, it will discard any local changes made using Put/Rename/etc. and force synchronize to the remote state.
	// If forceHandshake is true, it will re-verify any cached config. Specifically, this should be used when external config changes, such as the Git branch or file archive ID.
	Pull(ctx context.Context, opts *PullOptions) error
        ...
        ...
}
```

begelundmuller: Okay, I guess it makes sense to make it lazy and not run it as a background job. 

I'm good with just passing the properties in the way it currently happens. Just try to make the logic as separate/self-contained as possible (maybe a separate file if possible?), and add clear comments about why it's there (since it's quite confusing that the `file` driver has an `admin` dependency).

begelundmuller (approved):

PR #7673, opened 2025-07-25
feat: measure annotations backend

AdityaHegde:
Adds annotation support in metrics view.
1. Adds `annotations` key in metrics view yaml.
```
annotations:
- name: annotation # optional - defaults to model
  model: annotation_source # required
  measures: * # optional - defaults to all. supports selector syntax similar to explore config
- table: annotation_source # alternative way to specify a table directly
  connector: clickhouse # connector for the table
  database: annotation # database for the table
```
2. Adds `annotation` resolver that queries the underlying table. Takes metrics view, annotation, time start/end, optionally a time grain.
3. `MetricsViewAnnotations` api for UI to query the annotation. Takes the same args as `annotation`

Ref PRD: https://www.notion.so/rilldata/Measure-Annotations-9242c8d90da648b88c3b789ae3c67983?d=23bba33c8f5780c6974a001c74bc9f67#23aba33c8f5780a38834c6cb6ee39e4d

Closes PLAT-114

**Checklist:**
- [x] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/runtime/metricsview/executor.go
+++ b/runtime/metricsview/executor.go
@@ -580,14 +580,14 @@ func (e *Executor) Search(ctx context.Context, qry *SearchQuery, executionTime *
 }
 
 type AnnotationsQuery struct {
-	MetricsView string              `mapstructure:"metrics_view"`
-	Measures    []string            `mapstructure:"measures"`
-	TimeRange   *TimeRange          `mapstructure:"time_range"`
-	Limit       *int64              `mapstructure:"limit"`
-	Offset      *int64              `mapstructure:"offset"`
-	TimeZone    string              `mapstructure:"time_zone"`
-	TimeGrain   runtimev1.TimeGrain `mapstructure:"time_grain"`
-	Priority    int                 `mapstructure:"priority"`
+	MetricsView string     `mapstructure:"metrics_view"`
+	Measures    []string   `mapstructure:"measures"`
+	TimeRange   *TimeRange `mapstructure:"time_range"`
+	Limit       *int64     `mapstructure:"limit"`
+	Offset      *int64     `mapstructure:"offset"`
+	TimeZone    string     `mapstructure:"time_zone"`
+	TimeGrain   TimeGrain  `mapstructure:"time_grain"`
+	Priority    int        `mapstructure:"priority"`
 }
 
 func (q *AnnotationsQuery) AsMap() (map[string]any, error) {
@@ -876,9 +876,9 @@ END) as time_grain`)
 		args = append(args, start, end)
 	}
 
-	if annotation.HasGrain && qry.TimeGrain != runtimev1.TimeGrain_TIME_GRAIN_UNSPECIFIED {
+	if annotation.HasGrain && qry.TimeGrain != TimeGrainUnspecified {
 		b.WriteString(" AND (time_grain == 0 OR time_grain <= ?)")
-		args = append(args, int(qry.TimeGrain))
+		args = append(args, int(qry.TimeGrain.ToTimeutil()))
 	}
 
 	b.WriteString(" ORDER BY time")
--- a/runtime/parser/schema/project.schema.yaml
+++ b/runtime/parser/schema/project.schema.yaml
@@ -1251,7 +1251,7 @@ definitions:
                   description: A stable identifier for the annotation. Defaults to model or table names when not specified
                 model:
                   type: string
-                  description: Refers to the model powering the annotation (either table or model is required)
+                  description: Refers to the model powering the annotation (either table or model is required). The model must have 'time' and 'description' columns. Optional columns include 'time_end' for range annotations and 'grain' to specify when the annotation should appear based on dashboard grain level.
                 database:
                   type: string
                   description: Refers to the database to use in the OLAP engine (to be used in conjunction with table). Otherwise, will use the default database or schema if not specified
--- a/runtime/resolvers/annotations.go
+++ b/runtime/resolvers/annotations.go
@@ -104,21 +104,24 @@ func (r *annotationsResolver) Validate(ctx context.Context) error {
 }
 
 func (r *annotationsResolver) ResolveInteractive(ctx context.Context) (runtime.ResolverResult, error) {
-	tsRes, err := resolveTimestampResult(ctx, r.runtime, r.instanceID, r.query.MetricsView, r.mv.TimeDimension, r.claims, r.query.Priority)
-	if err != nil {
-		return nil, err
-	}
-
-	if r.query.TimeRange == nil || r.query.TimeRange.IsZero() {
-		r.query.TimeRange = &metricsview.TimeRange{
-			Start: tsRes.Min,
-			End:   tsRes.Max,
+	// Only resolve time stamps if an absolute time range is not specified.
+	if r.query.TimeRange == nil || r.query.TimeRange.Start.IsZero() || r.query.TimeRange.End.IsZero() {
+		tsRes, err := resolveTimestampResult(ctx, r.runtime, r.instanceID, r.query.MetricsView, r.mv.TimeDimension, r.claims, r.query.Priority)
+		if err != nil {
+			return nil, err
 		}
-	}
 
-	err = r.executor.BindAnnotationsQuery(ctx, r.query, tsRes)
-	if err != nil {
-		return nil, err
+		if r.query.TimeRange == nil || r.query.TimeRange.IsZero() {
+			r.query.TimeRange = &metricsview.TimeRange{
+				Start: tsRes.Min,
+				End:   tsRes.Max,
+			}
+		}
+
+		err = r.executor.BindAnnotationsQuery(ctx, r.query, tsRes)
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	res, err := r.executor.Annotations(ctx, r.query)
--- a/runtime/server/queries_metrics.go
+++ b/runtime/server/queries_metrics.go
@@ -512,9 +512,18 @@ func (s *Server) MetricsViewAnnotations(ctx context.Context, req *runtimev1.Metr
 		limit = &req.Offset
 	}
 
-	tr := metricsview.TimeRange{}
+	qry := metricsview.AnnotationsQuery{
+		MetricsView: req.MetricsViewName,
+		Measures:    req.Measures,
+		Limit:       limit,
+		Offset:      offset,
+		TimeZone:    req.TimeZone,
+		TimeGrain:   metricsview.TimeGrainFromProto(req.TimeGrain),
+		Priority:    int(req.Priority),
+	}
+
 	if req.TimeRange != nil {
-		tr = metricsview.TimeRange{
+		qry.TimeRange = &metricsview.TimeRange{
 			Start:         req.TimeRange.Start.AsTime(),
 			End:           req.TimeRange.End.AsTime(),
 			Expression:    req.TimeRange.Expression,
@@ -525,19 +534,16 @@ func (s *Server) MetricsViewAnnotations(ctx context.Context, req *runtimev1.Metr
 		}
 	}
 
+	props, err := qry.AsMap()
+	if err != nil {
+		return nil, err
+	}
+
 	res, err := s.runtime.Resolve(ctx, &runtime.ResolveOptions{
-		InstanceID: req.InstanceId,
-		Resolver:   "annotations",
-		ResolverProperties: map[string]any{
-			"metrics_view": req.MetricsViewName,
-			"measures":     req.Measures,
-			"priority":     req.Priority,
-			"time_range":   tr,
-			"time_grain":   req.TimeGrain,
-			"limit":        limit,
-			"offset":       offset,
-		},
-		Claims: auth.GetClaims(ctx).SecurityClaims(),
+		InstanceID:         req.InstanceId,
+		Resolver:           "annotations",
+		ResolverProperties: props,
+		Claims:             auth.GetClaims(ctx).SecurityClaims(),
 	})
 	if err != nil {
 		return nil, err
--- a/docs/docs/hidden/yaml/metrics-view.md
+++ b/docs/docs/hidden/yaml/metrics-view.md
@@ -163,6 +163,36 @@ _[array of object]_ - Used to define the numeric aggregates of columns from your
 
   - **`treat_nulls_as`** - _[string]_ - used to configure what value to fill in for missing time buckets. This also works generally as COALESCING over non empty time buckets. 
 
+### `annotations`
+
+_[array of object]_ - Used to define annotations that can be displayed on charts 
+
+  - **`name`** - _[string]_ - A stable identifier for the annotation. Defaults to model or table names when not specified 
+
+  - **`model`** - _[string]_ - Refers to the model powering the annotation (either table or model is required). The model must have 'time' and 'description' columns. Optional columns include 'time_end' for range annotations and 'grain' to specify when the annotation should appear based on dashboard grain level. 
+
+  - **`database`** - _[string]_ - Refers to the database to use in the OLAP engine (to be used in conjunction with table). Otherwise, will use the default database or schema if not specified 
+
+  - **`database_schema`** - _[string]_ - Refers to the schema to use in the OLAP engine (to be used in conjunction with table). Otherwise, will use the default database or schema if not specified 
+
+  - **`table`** - _[string]_ - Refers to the table powering the annotation, should be used instead of model for annotations from external OLAP tables (either table or model is required) 
+
+  - **`connector`** - _[string]_ - Refers to the connector to use for the annotation 
+
+  - **`measures`** - _[anyOf]_ - Specifies which measures to apply the annotation to. Applies to all measures if not specified 
+
+    - **option 1** - _[string]_ - Simple field name as a string.
+
+    - **option 2** - _[array of anyOf]_ - List of field selectors, each can be a string or an object with detailed configuration.
+
+      - **option 1** - _[string]_ - Shorthand field selector, interpreted as the name.
+
+      - **option 2** - _[object]_ - Detailed field selector configuration with name and optional time grain.
+
+        - **`name`** - _[string]_ - Name of the field to select. _(required)_
+
+        - **`time_grain`** - _[string]_ - Time grain for time-based dimensions. 
+
 ### `security`
 
 _[object]_ - Defines security rules and access control policies for resources 
--- a/docs/docs/reference/cli/project/status.md
+++ b/docs/docs/reference/cli/project/status.md
@@ -13,6 +13,7 @@ rill project status [<project-name>] [flags]
 ### Flags
 
 ```
+      --local            Target locally running Rill
       --path string      Project directory (default ".")
       --project string   Project Name
 ```
--- a/runtime/resolvers/testdata/annotations.yaml
+++ b/runtime/resolvers/testdata/annotations.yaml
@@ -111,7 +111,7 @@ tests:
       time_range:
         start: 2024-01-01T00:00:00Z
         end: 2024-03-31T00:00:00Z
-      time_grain: 5
+      time_grain: "day"
     result:
       - time: "2024-01-05T00:00:00Z"
         grain: "day"


proto/rill/runtime/v1/queries.proto
```diff
@@ -837,6 +844,29 @@ message MetricsViewTimeRangesResponse {
   repeated TimeRange time_ranges = 1;
 }
 
+message MetricsViewAnnotationsRequest {
+  string instance_id = 1;
+  string metrics_view_name = 2 [(validate.rules).string.min_len = 1];
+  string annotation_name = 3 [(validate.rules).string.min_len = 1];
```
begelundmuller: Did you consider passing a list of measure names, and then returning all relevant annotations for those measures? Not sure if that's desirable, just thought that would more map to the use case.

proto/rill/runtime/v1/queries.proto
```diff
@@ -837,6 +844,29 @@ message MetricsViewTimeRangesResponse {
...
+  int32 limit = 8 [(validate.rules).int32.gte = 0];
+  // Optional
+  int64 offset = 9 [(validate.rules).int64.gte = 0];
+}
+
+message MetricsViewAnnotationsResponse {
+  // Not optional, not null
+  StructType schema = 1;
+  // Not optional, not null
+  repeated google.protobuf.Struct data = 2;
```
begelundmuller: Could it be a fixed schema for annotations?

proto/rill/runtime/v1/resources.proto
```diff
@@ -243,6 +243,27 @@ message MetricsViewSpec {
...
+    string table = 5;
+    // Name of the model that source of annotation. Either table or model should be set.
+    string model = 6;
+    // Measures to apply the annotation to. If `measures_selector` is set, this will only be set in `state.valid_spec`.
+    repeated string measures = 7;
+    // Dynamic selector for `measures`. Will be processed during validation, so it will always be empty in `state.valid_spec`.
+    FieldSelector measures_selector = 8;
+    bool global = 9;
+    bool has_time_end = 10;
+    bool has_grain = 11;
```
begelundmuller: Would be nice with comments that describe what these mean

runtime/metricsview/executor.go
```diff
@@ -567,6 +567,73 @@ func (e *Executor) Search(ctx context.Context, qry *SearchQuery, executionTime *
 	return searchResult, nil
 }
 
+func (e *Executor) Annotations(ctx context.Context, annotation *runtimev1.MetricsViewSpec_Annotation, tr *TimeRange, grain runtimev1.TimeGrain) (*drivers.Result, error) {
```
begelundmuller: Consider accepting an annotation name and doing the lookup in `e.metricsView` inside the function, i.e. take `annotationName string` instead of `annotation *runtimev1.MetricsViewSpec_Annotation`.

runtime/metricsview/executor.go
```diff
@@ -567,6 +567,73 @@ func (e *Executor) Search(ctx context.Context, qry *SearchQuery, executionTime *
...
+		if e.security.CanAccessField(measure) {
+			accessibleMeasures++
+		}
+	}
+	// None of the measures are accessible, so annotation is not accessible either
+	if accessibleMeasures == 0 && !annotation.Global {
+		return nil, runtime.ErrForbidden
+	}
+
+	olap, olapCloser, err := e.rt.OLAP(ctx, e.instanceID, annotation.Connector)
```
begelundmuller: nit to align with terminology elsewhere: call it `release` instead of `olapCloser`

runtime/metricsview/executor.go
```diff
@@ -567,6 +567,73 @@ func (e *Executor) Search(ctx context.Context, qry *SearchQuery, executionTime *
...
+	if err != nil {
+		return nil, err
+	}
+	defer olapCloser()
+
+	err = e.resolveTimeRange(ctx, tr, time.UTC, nil)
```
begelundmuller: Maybe pass time zone as input from frontend instead of UTC? Not sure, but could it otherwise lead to some edge cases when using local time zones?

runtime/metricsview/executor_validate.go
```diff
@@ -125,6 +127,7 @@ func (e *Executor) ValidateAndNormalizeMetricsView(ctx context.Context) (*Valida
 		// One or more dimension/measure expressions failed to validate. We need to check each one individually to provide useful errors.
 		e.validateIndividualDimensionsAndMeasures(ctx, t, mv, cols, res)
 	}
+	e.validateAnnotations(ctx, mv, res)
```
begelundmuller: nit: doesn't match the comment above, so maybe move to a new block
```suggestion
	}

	// Check and rewrite annotations
	e.validateAnnotations(ctx, mv, res)
```

runtime/metricsview/executor_validate.go
```diff
@@ -267,6 +270,90 @@ func (e *Executor) validateIndividualDimensionsAndMeasures(ctx context.Context,
 	slices.SortFunc(res.MeasureErrs, func(a, b IndexErr) int { return a.Idx - b.Idx })
 }
 
+func (e *Executor) validateAnnotations(ctx context.Context, mv *runtimev1.MetricsViewSpec, res *ValidateMetricsViewResult) {
```
begelundmuller: Since this also rewrites the annotations, maybe call it `validateAndNormalizeAnnotations`?

runtime/metricsview/executor_validate.go
```diff
@@ -267,6 +270,90 @@ func (e *Executor) validateIndividualDimensionsAndMeasures(ctx context.Context,
...
+	allMeasures := make([]string, 0, len(mv.Measures))
+	for _, m := range mv.Measures {
+		allMeasures = append(allMeasures, m.Name)
+	}
+
+	// Get the controller used for getting the annotation's model
+	ct, err := e.rt.Controller(ctx, e.instanceID)
+	if err != nil {
+		res.OtherErrs = append(res.OtherErrs, fmt.Errorf("failed to get controller: %w", err))
+		return
```
begelundmuller: Better to return this error – it's not really a validation error, more an execution error (most likely a `ctx` cancellation).

runtime/metricsview/executor_validate.go
begelundmuller: Also might be nice with a docstring here that explains high level what it expects/rewrites.

proto/rill/runtime/v1/resources.proto
```diff
@@ -243,6 +243,27 @@ message MetricsViewSpec {
     // The data type of the measure. Only populated in ValidSpec.
     Type data_type = 15;
   }
+  // Annotations that can be applied to measures
+  message Annotation {
```
begelundmuller: Would be good to explicitly document the expected/supported underlying columns for annotations here

runtime/parser/parse_metrics_view.go
```diff
@@ -63,6 +63,15 @@ type MetricsViewYAML struct {
...
 	}
+	Annotations []*struct {
+		Name           string             `yaml:"name"`
+		Model          string             `yaml:"model"`
+		Database       string             `yaml:"database"`
+		DatabaseSchema string             `yaml:"database_schema"`
+		Table          string             `yaml:"table"`
+		Connector      string             `yaml:"connector"`
+		Measures       *FieldSelectorYAML `yaml:"measures"`
+	} `yaml:"annotations"`
```
begelundmuller, line 74: These new properties also need to be documented in `rill/runtime/parser/schema/project.schema.yaml`

runtime/parser/parse_metrics_view.go
```diff
@@ -568,6 +597,22 @@ func (p *Parser) parseMetricsView(node *Node) error {
 		node.Refs = append(node.Refs, ResourceName{Kind: ResourceKindTheme, Name: tmp.DefaultTheme})
 	}
 
+	// Add annotations as refs to the end of the metrics view.
+	for _, annotation := range tmp.Annotations {
```
begelundmuller, line 581: There are three loops over `tmp.Annotations` – looks like the first two loops can be combined into one here

runtime/resolvers/annotations.go
```diff
@@ -0,0 +1,123 @@
...
+	var annotation *runtimev1.MetricsViewSpec_Annotation
+	for _, specAnnotation := range mv.Annotations {
+		if specAnnotation.Name == props.Annotation {
+			annotation = specAnnotation
+			break
+		}
+	}
+	if annotation == nil {
+		return nil, fmt.Errorf("annotation %q not found in metrics view %q", props.Annotation, props.MetricsView)
+	}
```
begelundmuller: Mentioned this elsewhere also, but I suggest moving this lookup into `executor.Annotations`

runtime/resolvers/annotations.go
```diff
@@ -0,0 +1,123 @@
...
+func (r *annotationsResolver) CacheKey(ctx context.Context) ([]byte, bool, error) {
+	return nil, false, nil
+}
+
+func (r *annotationsResolver) Refs() []*runtimev1.ResourceName {
+	return []*runtimev1.ResourceName{
+		{Kind: runtime.ResourceKindMetricsView, Name: r.metricsView},
+		{Kind: runtime.ResourceKindModel, Name: r.annotation.Model},
+	}
+}
```
begelundmuller: I think it's enough to ref the metrics view. The metrics view in turn has refs to the models.

runtime/resolvers/annotations.go
```diff
@@ -0,0 +1,123 @@
...
+
+func (r *annotationsResolver) Close() error {
+	r.executor.Close()
+	return nil
+}
+
+func (r *annotationsResolver) CacheKey(ctx context.Context) ([]byte, bool, error) {
+	return nil, false, nil
+}
```
begelundmuller: Might be nice to add caching for annotations

runtime/server/queries_metrics.go
```diff
@@ -484,6 +489,67 @@ func (s *Server) MetricsViewTimeRanges(ctx context.Context, req *runtimev1.Metri
...
+				break
+			}
+			return nil, err
+		}
+		rowStruct, err := pbutil.ToStruct(row, res.Schema())
+		if err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
+		data = append(data, rowStruct)
+	}
```
begelundmuller: Mentioned this earlier, but would it make sense to parse into a more strictly typed response proto here? Or do you prefer it like this?

begelundmuller (changes requested):

proto/rill/runtime/v1/queries.proto
AdityaHegde: Mapping one backend query to a API request made sense to me. I feel like something like this would be achieved with query batching if and when we integrate it in UI. Open to other opinions.

runtime/metricsview/executor.go
AdityaHegde: Ya it should be taken from request. Got missed while adding this resolution.

runtime/metricsview/executor_validate.go
AdityaHegde: Updated

proto/rill/runtime/v1/resources.proto
AdityaHegde: This is more like the annotation definition. Is this the right place for that description?

proto/rill/runtime/v1/queries.proto
AdityaHegde: Updated to use fixed schema. Unknown fields will be put into `additional_fields`

proto/rill/runtime/v1/resources.proto
begelundmuller: For end-user descriptions, the YAML docs are a better place, but for (internal) info about how the resource works/validates/should be consumed, the protos are probably the best place.

proto/rill/runtime/v1/queries.proto
begelundmuller: Hmm, so the way I see it, the choice of having one or many underlying models is a bit of an implementation detail. We have also discussed supporting inline annotations in the future, so then there may be many more.

Since annotations are small data (we should ensure there aren't millions of annotations), I just wonder if it would make more sense to return them in one go instead of requiring to iterate over each annotation name in the metrics view spec.

For example, if you put yourself in the place of an LLM using MCP or a customer integrating with our APIs, wouldn't they just expect to make one call (maybe with some filters) and get a list of all the annotations to render in the current context?

AdityaHegde: Ok, one sense then. We dont really need annotations metadata outside of getting annotation name per measure.

proto/rill/runtime/v1/resources.proto
AdityaHegde: Sounds good, lemme update.

proto/rill/runtime/v1/queries.proto
```diff
@@ -837,6 +844,55 @@ message MetricsViewTimeRangesResponse {
...
+
+  message Measure {
+    // Not optional, not null
+    string name = 1;
+    // Not optional, not null
+    repeated MeasureAnnotation annotations = 2;
+  }
+
+  repeated Measure measures = 1;
+}
```
begelundmuller: What do you think about something like:
```suggestion
message MetricsViewAnnotationsResponse {
  message Annotation {
    // Description of the annotation. Maps to `description` column from the table.
    string description = 3;
    // Time when the annotation applies. Maps to `time` column from the table.
    google.protobuf.Timestamp time = 1;
    // Optional. Time when the annotation ends. Only present if the underlying table has the `time_end` column.
    optional google.protobuf.Timestamp time_end = 2;
    // Minimum grain this annotation is displayed for. Maps to `grain` column from the table.
    optional string grain = 4;
    // Any other fields are captured here. Will be used in predicates in the future.
    google.protobuf.Struct additional_fields = 5;
    // List of measure names that this annotation applies to. If empty, no restrictions apply.
    repeated string only_measures = 6;    
  }
  repeated Annotation rows = 1;
}
```

AdityaHegde: Ya works, updated.

runtime/server/queries_metrics.go
```diff
@@ -484,6 +491,92 @@ func (s *Server) MetricsViewTimeRanges(ctx context.Context, req *runtimev1.Metri
 	}, nil
 }
 
+func (s *Server) MetricsViewAnnotations(ctx context.Context, req *runtimev1.MetricsViewAnnotationsRequest) (*runtimev1.MetricsViewAnnotationsResponse, error) {
+	observability.AddRequestAttributes(ctx,
+		attribute.String("args.instance_id", req.InstanceId),
+		attribute.String("args.metric_view", req.MetricsViewName),
+		attribute.String("args.annotation", strings.Join(req.Measures, ",")),
```
begelundmuller: Key doesn't match values here

runtime/server/queries_metrics.go
```diff
@@ -484,6 +491,92 @@ func (s *Server) MetricsViewTimeRanges(ctx context.Context, req *runtimev1.Metri
...
+		return nil, err
+	}
+
+	// If there are no requested measures, return all measures of the metrics view.
+	reqMeasures := req.Measures
+	if len(reqMeasures) == 0 {
+		for _, mes := range mv.ValidSpec.Measures {
+			reqMeasures = append(reqMeasures, mes.Name)
+		}
+	}
```
begelundmuller: The logic of resolving which annotations to fetch for which measures should ideally happen inside the resolver or `metricsview` package. I.e. the server function should just be a thin wrapper. This enables easier re-use for invocations from other places (e.g. MCP tools).

Can it pass the measures in the `metricsview.AnnotationsQuery` and delegate all that handling to the metrics view package? Since the goal of the `metricsview` package is that it contains the logic related to how to consume metrics view specs.

runtime/resolvers/annotations.go
```diff
@@ -0,0 +1,152 @@
...
+
+type annotationsResolverArgs struct {
+	Priority   int                           `mapstructure:"priority"`
+	TimeRange  *metricsview.TimeRange        `mapstructure:"time_range"`
+	TimeGrain  runtimev1.TimeGrain           `mapstructure:"time_grain"`
+	TimeZone   string                        `mapstructure:"time_zone"`
+	Limit      *int64                        `mapstructure:"limit"`
+	Offset     *int64                        `mapstructure:"offset"`
+	Timestamps *metricsview.TimestampsResult `mapstructure:"timestamps"`
+}
```
begelundmuller: Could it take a `metricsview.AnnotationsQuery` as the properties? Similar to the `metrics` resolver: https://github.com/rilldata/rill/blob/9eab6594ca43b7e4c3edd96a2a004ca3c0dbe925/runtime/resolvers/metrics.go#L38-L42

runtime/resolvers/annotations.go
```diff
@@ -0,0 +1,152 @@
...
+
+type annotationsResolverArgs struct {
+	Priority   int                           `mapstructure:"priority"`
+	TimeRange  *metricsview.TimeRange        `mapstructure:"time_range"`
+	TimeGrain  runtimev1.TimeGrain           `mapstructure:"time_grain"`
+	TimeZone   string                        `mapstructure:"time_zone"`
+	Limit      *int64                        `mapstructure:"limit"`
+	Offset     *int64                        `mapstructure:"offset"`
+	Timestamps *metricsview.TimestampsResult `mapstructure:"timestamps"`
```
begelundmuller: Can it handle timestamps the same way as the `metricsResolver`? I.e. look them up inside `ResolveInteractive` instead of requiring them to be passed in.

The idea of the resolvers is that they provide a key-value based interface that can be queried/cached directly from many places, such as `data:` sections in alerts/reports/custom APIs, from LLM tool calls, from reconcilers, API handlers, etc.

So internal/computed values should ideally be resolved inside the resolver, not passed in the input.

runtime/metricsview/executor.go
```diff
@@ -577,6 +579,90 @@ func (e *Executor) Search(ctx context.Context, qry *SearchQuery, executionTime *
...
 }
 
+type AnnotationsQuery struct {
+	MetricsView string              `mapstructure:"metrics_view"`
+	Measures    []string            `mapstructure:"measures"`
+	TimeRange   *TimeRange          `mapstructure:"time_range"`
+	Limit       *int64              `mapstructure:"limit"`
+	Offset      *int64              `mapstructure:"offset"`
+	TimeZone    string              `mapstructure:"time_zone"`
+	TimeGrain   runtimev1.TimeGrain `mapstructure:"time_grain"`
```
begelundmuller: Can it use the `metricsview` package's `TimeGrain` type here? You can use `metricsview.TimeGrainFromProto` to convert.

https://github.com/rilldata/rill/blob/6cdbe945affb9d3ce01f10acb2a3113c9769093b/runtime/metricsview/query.go#L262-L275

runtime/metricsview/executor.go
```diff
@@ -732,6 +818,113 @@ func (e *Executor) timeColumnOrExpr(timeDim string) (string, error) {
...
+		b.WriteString(" AND (time_grain == 0 OR time_grain <= ?)")
+		args = append(args, int(qry.TimeGrain))
+	}
+
+	b.WriteString(" ORDER BY time")
+
+	if qry.Limit != nil {
+		b.WriteString(" LIMIT ?")
+		args = append(args, *qry.Limit)
+	}
```
begelundmuller, line 889: So basically the limit gets applied to each annotation call instead of the combined query? That could lead to issues if multiple annotation models have many rows.

But it's a bit tricky to fix (probably requires a combined SQL query or a pagination token with more state). I'm okay with keeping this for now and fixing when/if this becomes an issue.

runtime/parser/schema/project.schema.yaml
```diff
@@ -1240,6 +1240,33 @@ definitions:
                   type: string
                   description: used to configure what value to fill in for missing time buckets. This also works generally as COALESCING over non empty time buckets.
             minItems: 1
+          annotations:
```
begelundmuller, line 1243: Need to also run `make docs.generate` for the auto-generated reference docs that consume these

runtime/parser/schema/project.schema.yaml
```diff
@@ -1240,6 +1240,33 @@ definitions:
...
+            description: Used to define annotations that can be displayed on charts
+            items:
+              type: object
+              properties:
+                name:
+                  type: string
+                  description: A stable identifier for the annotation. Defaults to model or table names when not specified
+                model:
+                  type: string
+                  description: Refers to the model powering the annotation (either table or model is required)
```
begelundmuller: Can you describe the required and optional columns in the model here? That will be convenient to have something to link to.

The property descriptions here are also what we'll be feeding into LLMs for code generation in the upcoming code agent work.

runtime/resolvers/annotations.go
```diff
@@ -0,0 +1,134 @@
...
+		r.query.TimeRange = &metricsview.TimeRange{
+			Start: tsRes.Min,
+			End:   tsRes.Max,
+		}
+	}
+
+	err = r.executor.BindAnnotationsQuery(ctx, r.query, tsRes)
+	if err != nil {
+		return nil, err
+	}
```
begelundmuller: If the time range already has a fixed start and end (which I think it often will since the frontend resolves time ranges before making other calls), then it doesn't need to get the timestamps dynamically, right? So should this be moved into a nested case somehow? 

Looking in the `metricsResolver`, seems like a similar issue may be there. Not sure if I'm missing something.

runtime/server/queries_metrics.go
```diff
@@ -484,6 +490,114 @@ func (s *Server) MetricsViewTimeRanges(ctx context.Context, req *runtimev1.Metri
...
+		Resolver:   "annotations",
+		ResolverProperties: map[string]any{
+			"metrics_view": req.MetricsViewName,
+			"measures":     req.Measures,
+			"priority":     req.Priority,
+			"time_range":   tr,
+			"time_grain":   req.TimeGrain,
+			"limit":        limit,
+			"offset":       offset,
+		},
```
begelundmuller: Consider using `metricsview.AnnotationsQuery{}` and `.AsMap()` for type safety (in case we rename properties one place and forget elsewhere)

begelundmuller (approved):

runtime/metricsview/executor.go
AdityaHegde: For the time being we wont have too much of data. Since we query for the selected time range in dashboard this should fine for now.

runtime/resolvers/annotations.go
AdityaHegde: Ya we can skip it here. But since dashboard always calls `MetricsViewTimeRange` this wont give use too much.

Wont change `metricsResolver` in this PR, dont know if it could lead to issues.

PR #7676, opened 2025-07-28
feat: add color mapping for canvas chart fields

djbarnwal:
- Extract color utility functions to separate util.ts file
- Add labelFirst prop to ColorInput component for inline display
- Integrate color mapping UI into field config inspector
- Enable per-value color customization for chart data series

TODOs - 
- [ ] fix reset functionality
- [ ] fix initial color mapping

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


jkhwu (changes requested):
A bug + a non-blocking request: https://www.loom.com/share/f4461bbb7c0244acb3548c32278d3595?sid=95590ab9-f5bd-4bff-ac34-0e555abd6246

Some more non-blocking requests: https://www.loom.com/share/6c63cd94e6594f2dabc2fc2875ef4a82?sid=989686bc-5153-4200-b695-a06191588728

djbarnwal:
> A bug + a non-blocking request: https://www.loom.com/share/f4461bbb7c0244acb3548c32278d3595?sid=95590ab9-f5bd-4bff-ac34-0e555abd6246

Fixed the bug and update other color pickers to put the label first.

> Some more non-blocking requests: https://www.loom.com/share/6c63cd94e6594f2dabc2fc2875ef4a82?sid=989686bc-5153-4200-b695-a06191588728

Moved the "Reset to default" button at the bottom. Put it on the right side since we have "see more" on the left. 
With respect to the color input field, it expect CSS colors. As such `fff` is invalid compared to `#fff`. We can work something out for this probably later. Also, empty colors are not saved. If you refresh, you will see the previous selected color in the input field.

web-common/src/features/canvas/components/charts/types.ts
```diff
@@ -41,8 +41,15 @@ export type ChartDataResult = {
...
   fields: ChartFieldsMap;
   error?: HTTPError | null;
+  metadata?: ChartMetadata;
 };
 
+export interface ChartMetadata {
+  xValues?: string[];
+  colorValues?: string[];
+  yValues?: string[];
+}
```
ericpgreen2: Nit: I think this seems more like "config" than "metadata"? For example, I don't think of "metadata" as being something that the developer can "reset".

ericpgreen2 (changes requested):
UXQA:
1. It's confusing that changing the color changes the x-axis: [Jam](https://jam.dev/c/b687ec6f-2b22-471a-af48-0a9d4b8d2b13). In one case, it even changed the bin on my bar graph's buckets, so it appeared like it caused the data to change.
2. When I submit an invalid hex code, there's a mismatch between the color preview (black) and the resultant color (blank). Can we include some input validation to make it clear when the provided hex code is invalid?
    image

web-common/src/features/canvas/components/charts/types.ts
djbarnwal: Updated it to `domainValues` and `ChartDomainValues` as is uses data visualization terminology where "domain" refers to the set of possible values for a scale

ericpgreen2 (approved):
LGTM!

jkhwu (approved):
Looks good to me, thank you!

PR #7678, opened 2025-07-28
feat (AI): Add project-level chatbot

ericpgreen2:
Closes [APP-227](https://linear.app/rilldata/issue/APP-227/project-level-chatbot)

TODO:
- [ ] Design review
- [ ] Improve system prompt

image

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!


web-admin/src/routes/[organization]/[project]/+page.ts
```diff
@@ -0,0 +1,16 @@
+import { featureFlags } from "@rilldata/web-common/features/feature-flags.js";
+import { redirect } from "@sveltejs/kit";
+import { get } from "svelte/store";
+
+export const load = async ({ params: { organization, project }, parent }) => {
+  // Wait for the feature flags to load
+  await parent();
+
+  // NOTE: in the future, we'll use user-level `ai` permissions to determine access to the chat page.
+  const chatEnabled = get(featureFlags.chat);
```
Comment (AdityaHegde) [inline web-admin/src/routes/[organization]/[project]/+page.ts:10]: Would we want some kind of user setting as well? I doubt every user would want to land on chat if admin enables this feature.

web-common/src/features/chat/core/chat.ts
```diff
@@ -0,0 +1,596 @@
...
+              getRuntimeServiceGetConversationQueryOptions(
+                instanceId,
+                response.conversationId,
+              ),
+            );
+
+            // Add the new conversation to the `listConversations` TanStack Query Cache
+            const listQueryKey =
+              getRuntimeServiceListConversationsQueryKey(instanceId);
+            queryClient.setQueryData(listQueryKey, (oldList: any) => {
```
AdityaHegde: Could you add type to `oldList` and `conv` in this block? Hard to follow otherwise why we merge conversations field and the object above separately.

web-common/src/features/chat/core/chat.ts
```diff
@@ -0,0 +1,596 @@
...
+      ),
+  );
+  const getConversationQuery = createQuery(getConversationQueryOptionsStore);
+
+  // Derived state
+  const listConversationsData = derived(
+    listConversationsQuery,
+    ($query) => $query.data,
+  );
+  const getConversationData = derived(
```
AdityaHegde: `getConversationData` is not used anywhere. Feels like over kill to have this explanatory variable. Would not be hard understand if `currentConversation` has `$query.data?.conversation || null`.

AdityaHegde (approved):

web-admin/src/routes/[organization]/[project]/+page.ts
Comment (ericpgreen2) [reply web-admin/src/routes/[organization]/[project]/+page.ts:10]: That's an interesting idea. I'll make sure to get more feedback on what landing page people want. For now I'll keep as is since it's behind the feature flag.

web-common/src/features/chat/core/chat.ts
ericpgreen2: Good call– updated

web-common/src/features/chat/core/chat.ts
ericpgreen2: Agreed– updated

PR #7681, opened 2025-07-28
Table Discovery for all supported connectors

lovincyrus:
This PR extends table discovery functionality from OLAP-only connectors to all supported connector types (OLAP + SQL). It introduces a hybrid architecture that automatically detects connector capabilities and uses the appropriate API for table browsing. Fixes https://linear.app/rilldata/issue/APP-234/table-discovery-for-these-connectors

---

To check supported connectors from https://linear.app/rilldata/issue/PLAT-14/platform-support-for-testing-connections
- [x] Postgres (ImplementsSQLStore: true,)
- [x] Snowflake (ImplementsWarehouse: true,)
- [x] Athena (ImplementsWarehouse: true,)
- [x] Redshift (ImplementsWarehouse: true,)
- [x] Bigquery (ImplementsWarehouse: true,)
- [x] MySQL (ImplementsSQLStore: true,)

POSTGRES

CleanShot 2025-07-28 at 09 43 28@2x

SNOWFLAKE

CleanShot 2025-07-28 at 14 03 55@2x

ATHENA

CleanShot 2025-07-28 at 13 39 34@2x

REDSHIFT

CleanShot 2025-07-28 at 13 59 16@2x

BIGQUERY

CleanShot 2025-07-28 at 14 02 11@2x

MYSQL

CleanShot 2025-07-28 at 14 32 07@2x

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!


web-common/src/features/connectors/ConnectorEntry.svelte
```diff
@@ -55,6 +61,10 @@
 
         {#if isOlapConnector}
           <Tag height={16} class="ml-auto">OLAP</Tag>
+        {:else if implementsWarehouse}
+          <Tag height={16} class="ml-auto">WAREHOUSE</Tag>
+        {:else if implementsSqlStore}
+          <Tag height={16} class="ml-auto">SQL</Tag>
```
ericpgreen2: It's nuanced, but these tags are serving a different purpose than the "OLAP" tag. The "OLAP" tag denotes which is your currently active _default_ OLAP engine. Let's say you have 2 OLAP engines connected, only 1 of them (the default one) will have the "OLAP" tag. Currently, the OLAP tag is not denoting which connectors `implementsOlap`.

For now, both for consistency and to keep the prominence of the "OLAP" tag (which is the most important one), I think the simplest thing to do is remove the "Warehouse" and "SQL" tags.

web-common/src/features/connectors/olap/DatabaseSchemaEntry.svelte
ericpgreen2, line 1: Given this is now used by more connectors than just the `olap` connectors, looks like this also should be pulled up one directory (i.e. `web-common/src/features/connectors/DatabaseSchemaEntry.svelte`)

web-common/src/features/connectors/olap/TableEntry.svelte
ericpgreen2, line 1: Similarly, this should be moved up one directory

web-common/src/features/connectors/olap/DatabaseSchemaEntry.svelte
ericpgreen2: Though, another option could be to create a dedicated feature folder (e.g. either `features/connectors/explorer` or `features/connector-explorer`)

web-common/src/features/connectors/selectors.ts
```diff
@@ -0,0 +1,194 @@
...
+          ),
+        )
+      : // For hierarchical structures, use database as the primary level
+        Array.from(
+          new Set($query.data?.map((schema) => schema.database ?? "") ?? []),
+        ).filter(Boolean);
+
+    return { isLoading: false, data: databases, error: undefined };
+  });
+}
```
ericpgreen2: This could be simpler if it leaned into the TanStack Query's `select` parameter, rather than using a derived store. With the `derived` store approach, you have to synchronize the `isLoading` and `error` states– better to let TanStack do it.

web-common/src/features/connectors/selectors.ts
```diff
@@ -0,0 +1,194 @@
...
+ * Fetches schemas for a specific database
+ */
+export function useSchemasForDatabase(
+  instanceId: string,
+  connector: string,
+  database: string,
+) {
+  const databaseSchemasQuery = useDatabaseSchemas(instanceId, connector);
+
+  return derived([databaseSchemasQuery], ([$query]) => {
```
ericpgreen2: Similarly here, I'd recommend using TanStack's native `select` functionality

web-common/src/features/connectors/selectors.ts
```diff
@@ -0,0 +1,194 @@
...
+          databaseSchema !== undefined,
+      },
+    },
+  );
+}
+
+/**
+ * COMPATIBILITY LAYER
+ * These adapt the new APIs to match the old API shapes for easier migration
+ */
```
ericpgreen2: Are the below functions used? I'm not immediately seeing them used anywhere.

ericpgreen2 (dismissed):
Left some code comments. Additionally, would you mind creating a private Notion doc with the `connector.yaml` files that you used to test each of these connectors?

web-common/src/features/connectors/olap/DatabaseSchemaEntry.svelte
lovincyrus: consolidated the selectors in `features/connectors` https://github.com/rilldata/rill/pull/7681/commits/2af260555b6b9dfa403098a3c945921cc2ee2c45

lovincyrus:
> Additionally, would you mind creating a private Notion doc with the `connector.yaml` files that you used to test each of these connectors?

Added the connectors to https://www.notion.so/rilldata/connector-yaml-23fba33c8f5780ba8272c25d1d968fce

lovincyrus:
Re-tested all the connectors after all the feedback changes.

web-common/src/features/connectors/DatabaseExplorer.svelte
ericpgreen2, line 1: It looks like you moved _some_ of the code to `features/connectors/explorer`, but not all of it. I like the `explorer` directory, but it looks like explorer-related code still exists at the top-level in `features/connectors` and also in the `features/connectors/olap` directory. All code related to this panel should be colocated in the `explorer` directory. I think I'd even consider the `TableWorkspace` a part of the `explorer`.

ericpgreen2 (changes requested):
UXQA:
1. If you click on one of these "Generate" buttons, you might notice that they fail due to Postgres (or other SQL or Warehouse connector) not being a valid OLAP connector. In this PR, we should remove these menu items. Either in this PR or in a follow-up, we need to add a "Create model" menu item. See below for what the "Create model" generated code should look like.
    image

    ```yaml
    # filename: bookmarks.yaml
    connector: postgres
    sql: select * from bookmarks
    ```

lovincyrus:
> 2\. Either in this PR or in a follow-up, we need to add a "Create model" menu item. The button should generate code like this:
>     ```yaml
>     # filename: bookmarks.yaml
>     connector: postgres
>     sql: select * from bookmarks
>     ```

We should address this in https://linear.app/rilldata/issue/ENG-854/create-connector-files-not-model-files

lovincyrus:
> 1. If you click on one of these "Generate" buttons, you might notice that they fail due to Postgres (or other SQL or Warehouse connector) not being a valid OLAP connector. In this PR, we should remove these menu items for connectors that are not OLAP engines.
>    image

Addressed in https://github.com/rilldata/rill/pull/7681/commits/901cfdd9f974b7cc32d3f96f9bb8fd180bd75814

web-common/src/features/connectors/explorer/TableEntry.svelte
```diff
@@ -5,23 +5,27 @@
   import MoreHorizontal from "@rilldata/web-common/components/icons/MoreHorizontal.svelte";
   import CaretDownIcon from "../../../components/icons/CaretDownIcon.svelte";
   import TableIcon from "../../../components/icons/TableIcon.svelte";
-  import TableMenuItems from "./TableMenuItems.svelte";
-  import TableSchema from "./TableSchema.svelte";
-  import UnsupportedTypesIndicator from "./UnsupportedTypesIndicator.svelte";
+  import TableMenuItems from "../olap/TableMenuItems.svelte";
+  import TableSchema from "../olap/TableSchema.svelte";
+  import UnsupportedTypesIndicator from "../olap/UnsupportedTypesIndicator.svelte";
```
ericpgreen2: The goal is for this `explorer` directory to encapsulate all the feature's code. These imports signal that these files should be moved to the `explorer` directory too.

web-common/src/features/connectors/explorer/TableEntry.svelte
```diff
@@ -30,24 +34,30 @@
 
   const { allowContextMenu, allowNavigateToTable, allowShowSchema } = store;
 
+  $: ({ instanceId: runtimeInstanceId } = $runtime);
+  $: isModelingSupportedForConnector = useIsModelingSupportedForConnector(
+    runtimeInstanceId,
+    connector,
+  );
+  $: isModelingSupported = $isModelingSupportedForConnector.data;
```
ericpgreen2, line 42: This is fine temporarily, but note that this isn't really what we want. This effectively says "if Postgres doesn't support modeling (it doesn't), then we won't show the `Create model` item". But we _want_ to show `Create model` for Postgres. Our generated model code will be a "source model" that ingests the data into the default OLAP engine, so **the determinant should be whether or not the default OLAP engine supports modeling**.

web-common/src/features/connectors/explorer/TableEntry.svelte
lovincyrus: https://github.com/rilldata/rill/pull/7681/commits/c8f751f2c57aa8d450ad8b17e2755b31b619e8d1

web-common/src/features/connectors/explorer/TableEntry.svelte
lovincyrus: These changes aren't for https://linear.app/rilldata/issue/ENG-856/add-a-create-model-menu-item-for-connector-files

Please refer to [these scoped changes](https://github.com/rilldata/rill/pull/7681/commits/901cfdd9f974b7cc32d3f96f9bb8fd180bd75814) that are meant for https://github.com/rilldata/rill/pull/7681#issuecomment-3141208706

PR #7685, opened 2025-07-29
feat (OLAP connectors): Replace `clickhouseModeling` feature flag with a dynamic check for modeling support

ericpgreen2:
This PR removes the `clickhouseModeling` feature flag in favor of the application's own assessment of whether or not the OLAP connector supports modeling. Modeling is currently supported in these cases:
1. The OLAP engine is DuckDB
2. The OLAP engine is Rill-managed (e.g. Rill-managed ClickHouse)
3. The OLAP engine is running in `readwrite` mode
 
Addresses [this Slack request](https://rilldata.slack.com/archives/C02T907FEUB/p1753697369166259)

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!

### DIFF
--- a/web-common/src/features/connectors/olap/selectors.ts
+++ b/web-common/src/features/connectors/olap/selectors.ts
@@ -33,7 +33,7 @@ function createModelingSupportQueryOptions(
     enabled: !!instanceId && !!connectorName,
     select: (data: V1GetResourceResponse) => {
       const spec = data?.resource?.connector?.spec;
-      if (!spec) return false;
+      if (!spec) return connectorName === "duckdb";
 
       // Modeling is supported if:
       // - DuckDB (embedded database with full SQL support)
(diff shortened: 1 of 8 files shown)


web-common/src/features/connectors/olap/selectors.ts
```diff
@@ -1,63 +1,85 @@
...
+    }),
+    queryFn: () =>
+      runtimeServiceGetResource(instanceId, {
+        "name.kind": ResourceKind.Connector,
+        "name.name": connectorName,
+      }),
+    enabled: !!instanceId && !!connectorName,
+    select: (data: V1GetResourceResponse) => {
+      const spec = data?.resource?.connector?.spec;
+      if (!spec) return false;
```
begelundmuller, line 54: Maybe this is handled elsewhere, but note that some legacy projects don't have an explicit `duckdb` connector resource. So if it's not found, maybe fall back to:
```suggestion
      const spec = data?.resource?.connector?.spec;
      if (!spec) return connectorName === "duckdb";
```

ericpgreen2: Ah, yeah some tests were failing due to this exact reason. Fixed.

PR #7687, opened 2025-07-29
PLAT-132 support local flag in rill project status

NamanMahor:
[PLAT-132: Support `--local` flag in `rill project status`](https://linear.app/rilldata/issue/PLAT-132/support-local-flag-in-rill-project-status)

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!

### DIFF
--- a/cli/cmd/project/status.go
+++ b/cli/cmd/project/status.go
@@ -35,12 +35,12 @@ func StatusCmd(ch *cmdutil.Helper) *cobra.Command {
 				if len(args) > 0 {
 					return fmt.Errorf("the --local flag cannot be used with <project-name> positional argument")
 				}
-			}
-
-			if !local && !cmd.Flags().Changed("project") && len(args) == 0 && ch.Interactive {
-				name, err = ch.InferProjectName(cmd.Context(), ch.Org, path)
-				if err != nil {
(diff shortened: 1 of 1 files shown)


cli/cmd/project/status.go
```diff
@@ -29,67 +29,67 @@ func StatusCmd(ch *cmdutil.Helper) *cobra.Command {
 				name = args[0]
 			}
```
NamanMahor:
```suggestion
			if local {
				if cmd.Flags().Changed("project") || cmd.Flags().Changed("path") {
					return fmt.Errorf("the --local flag cannot be used with --project or --path")
				}
				if len(args) > 0 {
					return fmt.Errorf("the --local flag cannot be used with <project-name> positional argument")
				}
			}

```

NamanMahor: should we add this check too. because it ignore the project name in local mode? we can fail if other args are passed.

begelundmuller: Yeah that would be a nice check to add

begelundmuller:
The current changes look good to me, but will wait for this until I approve:
> should we add this check too. because it ignore the project name in local mode? we can fail if other args are passed.

begelundmuller (approved):

PR #7689, opened 2025-07-29
feat: measure annotations

AdityaHegde:
Adds support for annotations in measures. Ref design: https://www.notion.so/rilldata/Measure-Annotations-9242c8d90da648b88c3b789ae3c67983?source=copy_link#238ba33c8f578053aa95d6b88375b416

Sample models for annotations: https://www.notion.so/rilldata/Measure-Annotations-9242c8d90da648b88c3b789ae3c67983?source=copy_link#14fba33c8f5780a58944e5fbc246f664

Sample yaml changes: https://www.notion.so/rilldata/Measure-Annotations-9242c8d90da648b88c3b789ae3c67983?source=copy_link#14fba33c8f5780cf924bc2cb2441db83

Builds on top of https://github.com/rilldata/rill/pull/7673

TODO list
- [x] Minor tweaks to match the design.
- [x] Get a proper idea on how truncation will work with annotations. Right now they are only truncated if the underlying model has `grain` column.
- [x] Tweak the hover detection behaviour.
- [ ] Tests (Will do a follow up)

Filter and predicate support will come in future versions.

Closes APP-84

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


web-common/src/components/icons/EllipsisVertical.svelte
```diff
@@ -0,0 +1,34 @@
+<script>
```
djbarnwal: This new icon component is probably not needed. We already have the `ThreeDot` icon. Can be used as 
```
<ThreeDot className="rotate-90" size="16px" />
```

web-common/src/features/dashboards/time-controls/range-formatting.ts
```diff
@@ -0,0 +1,72 @@
+import { DateTime, type Interval } from "luxon";
+
+/**
+ * Formats a time range to a human readable time range.
+ * TODO: update time picker to use this
+ */
+export function formatRange(interval: Interval) {
```
djbarnwal: Nit: We have an existing time formatter for human readable time range `prettyFormatTimeRange`. Would be better to use that for consistency.

PR #7690, opened 2025-07-29
Support separate read and write handles in ClickHouse Connector

grahamplata:
Support separate read and write handles in ClickHouse Connector.

Relates to: https://github.com/rilldata/rill-helm-charts/pull/220

**Checklist:**
- [x] Covered by tests
- [x] Linked the [issues](https://linear.app/rilldata/issue/PLAT-131/support-separate-read-and-write-handles-in-clickhouse) it closes and [name](https://linear.app/rilldata/issue/PLAT-155/fix-clickhouse-provisioner-database-names)
- [ ] I'm proud of this work!


runtime/drivers/clickhouse/clickhouse.go
```diff
@@ -61,12 +60,32 @@ var spec = drivers.Spec{
...
-			Required:    true,
+			Required:    false,
 			DisplayName: "Connection string",
 			Placeholder: "clickhouse://localhost:9000?username=default&password=password",
 			Secret:      true,
 			NoPrompt:    true,
 		},
+		{
+			Key:         "read_dsn",
+			Type:        drivers.StringPropertyType,
```
nishantmonu51: This is an implementation detail for how we setup clickhouse, our users don't setup CH the same way. This should be an internal detail - user just configure `managed: true` and we configure these in the project config. User never configures these in the UI.

nishantmonu51: Other way to state the same is that - 
- YAML artifacts just contain dsn and no read/write_dsn
- The parsed catalog object can have these fields which is an implementation detail.

admin/provisioner/clickhousestatic/provisioner.go
```diff
@@ -31,6 +31,21 @@ type Spec struct {
 	// ENV variable that contains the clickhouse DSN.
 	// This is an alternative to specifying the DSN directly, which can be useful for injecting secrets
 	DSNEnv string `json:"dsn_env"`
+
+	// ReadDSN with admin permissions for read operations (optional separate read replica)
+	ReadDSN string `json:"read_dsn,omitempty"`
+	// ReadDSNPath with admin permissions for read operations (optional separate read replica)
+	ReadDSNPath string `json:"read_dsn_path,omitempty"`
+	// ReadDSNEnv with admin permissions for read operations (optional separate read replica)
+	ReadDSNEnv string `json:"read_dsn_env,omitempty"`
```
begelundmuller: Do we use both `dsn_path` and `dsn_env`? I think it's just `dsn_env` now, right? I think we should just remove the one we don't use to keep things simple.

admin/provisioner/clickhousestatic/provisioner.go
```diff
@@ -69,7 +84,48 @@ func New(specJSON []byte, _ database.DB, logger *zap.Logger) (provisioner.Provis
...
+	}
+
+	provDSN := spec.DSN
+	if provDSN == "" && spec.WriteDSN != "" {
+		provDSN = spec.WriteDSN
+	}
+
+	if provDSN == "" {
+		return nil, fmt.Errorf("no DSN available - must specify either 'dsn' or 'write_dsn'")
+	}
```
begelundmuller: Could the logic be simplified if you just have `DSN` (required) and `WriteDSN` (optional) options? Then it uses `WriteDSN` for writes if set, otherwise always falls back to `DSN`.

It avoids some of these edge cases where `ReadDSN` and `WriteDSN` combined are mutually exclusive of `DSN`.

runtime/drivers/clickhouse/clickhouse.go
```diff
@@ -150,6 +149,10 @@ type configProperties struct {
 	Provision bool `mapstructure:"provision"`
 	// DSN is the connection string. Either DSN can be passed or the individual properties below can be set.
 	DSN string `mapstructure:"dsn"`
+	// ReadDSN is the connection string for read operations. Takes precedence over DSN for reads.
+	ReadDSN string `mapstructure:"read_dsn"`
+	// WriteDSN is the connection string for write operations. Takes precedence over DSN for writes.
+	WriteDSN string `mapstructure:"write_dsn"`
```
begelundmuller: Same comment as above, wondering if that would simplify things a bit.

runtime/drivers/clickhouse/clickhouse.go
```diff
@@ -447,14 +446,32 @@ func (c *Connection) Config() map[string]any {
...
+		if errEmbed != nil {
+			errors = append(errors, fmt.Errorf("stopping embedded ClickHouse: %w", errEmbed))
+		}
 	}
 
-	return errors.Join(errDB, errEmbed)
+	if len(errors) > 0 {
+		return fmt.Errorf("multiple errors during close: %v", errors)
+	}
+	return nil
```
begelundmuller: Why this change? Using `%v` to list the errors breaks the ability to use `errors.Is` to handle errors (e.g. context cancellations).

begelundmuller (changes requested):

admin/provisioner/clickhousestatic/provisioner.go
grahamplata: Happy to do so

admin/provisioner/clickhousestatic/provisioner.go
```diff
@@ -22,15 +22,11 @@ func init() {
 }
 
 type Spec struct {
-	// DSN with admin permissions for a Clickhouse service.
-	// This will be used to create a new (virtual) database and access-restricted user for each provisioned resource.
-	DSN string `json:"dsn"`
-	// Path to a file that we should load the DSN from.
-	// This is an alternative to specifying the DSN directly, which can be useful for secrets management.
-	DSNPath string `json:"dsn_path"`
```
begelundmuller: I think only `dsn_path` can be removed, not `dsn`. The `dsn` setting is used in the local dev environments:
```
"clickhouse": {"type": "clickhouse-static", "spec": {"dsn": "clickhouse://default:default@localhost:9000"}}
```

admin/provisioner/clickhousestatic/provisioner.go
```diff
@@ -53,32 +52,40 @@ func New(specJSON []byte, _ database.DB, logger *zap.Logger) (provisioner.Provis
...
-	if spec.DSNPath != "" {
-		dsn, err := os.ReadFile(spec.DSNPath)
-		if err != nil {
-			return nil, fmt.Errorf("failed to read DSN file: %w", err)
-		}
-		spec.DSN = strings.TrimSpace(string(dsn))
+	// Get primary DSN from the specified environment variable (required)
+	dsn := os.Getenv(spec.DSNEnv)
+	if dsn == "" {
+		return nil, fmt.Errorf("environment variable %s is not set or empty", spec.DSNEnv)
```
begelundmuller: Use `%q` so errors due to empty strings are easier to notice
```suggestion
		return nil, fmt.Errorf("environment variable %q is not set or empty", spec.DSNEnv)
```

admin/provisioner/clickhousestatic/provisioner.go
```diff
@@ -53,32 +52,40 @@ func New(specJSON []byte, _ database.DB, logger *zap.Logger) (provisioner.Provis
...
-	if spec.DSNEnv != "" && spec.DSN == "" && spec.DSNPath == "" {
-		dsn := os.Getenv(spec.DSNEnv)
-		if dsn == "" {
-			return nil, fmt.Errorf("environment variable %s is not set or empty", spec.DSNEnv)
+	// Get optional write DSN
+	var writeDSN string
+	if spec.WriteDSNEnv != "" {
+		writeDSN = os.Getenv(spec.WriteDSNEnv)
+		if writeDSN == "" {
+			return nil, fmt.Errorf("environment variable %s is not set or empty", spec.WriteDSNEnv)
```
begelundmuller:
```suggestion
			return nil, fmt.Errorf("environment variable %q is not set or empty", spec.WriteDSNEnv)
```

admin/provisioner/clickhousestatic/provisioner.go
```diff
@@ -53,32 +52,40 @@ func New(specJSON []byte, _ database.DB, logger *zap.Logger) (provisioner.Provis
...
 	ch := clickhouse.OpenDB(opts)
 
 	return &Provisioner{
-		spec:   spec,
-		logger: logger,
-		ch:     ch,
+		spec:     spec,
+		logger:   logger,
+		readDB:   ch,
+		writeDB:  ch,
```
begelundmuller: Doesn't really make sense to have two handles if they are the same? Seems like the old `p.ch` is sufficient int the provisioner (read/write separation for handles only really matter in `runtime/drivers/clickhouse`, not in the provisioner)

admin/provisioner/clickhousestatic/provisioner.go
```diff
@@ -101,13 +123,12 @@ func (p *Provisioner) Provision(ctx context.Context, r *provisioner.Resource, op
 		return nil, err
 	}
 
-	// If the config has already been populated, do a health check and exit early (currently there's nothing to update).
+	// If the config has already been populated, do a health check and exit early
 	if cfg.DSN != "" {
 		err := p.pingWithResourceDSN(ctx, cfg.DSN)
```
begelundmuller, line 116: If it has a write DSN, can should it also ping that to check both?

admin/provisioner/resources.go
```diff
@@ -91,11 +92,13 @@ func NewClickhouseConfig(cfg map[string]any) (*ClickhouseConfig, error) {
...
 	res := make(map[string]any)
-	err := mapstructure.Decode(r, &res)
-	if err != nil {
-		panic(err)
+	if c.DSN != "" {
+		res["dsn"] = c.DSN
+	}
+	if c.WriteDSN != "" {
+		res["write_dsn"] = c.WriteDSN
 	}
```
begelundmuller: Maybe keep the `mapstructure` and add `omitempty` in the struct tags instead? https://pkg.go.dev/github.com/mitchellh/mapstructure#hdr-Omit_Empty_Values

runtime/drivers/clickhouse/clickhouse.go
begelundmuller: Seems like there's still three DSNs here? Shouldn't it just be `DSN` and `WriteDSN` now?

begelundmuller (changes requested):

runtime/drivers/clickhouse/clickhouse.go
```diff
@@ -308,91 +306,99 @@ func (d driver) Open(instanceID string, config map[string]any, st *storage.Clien
 		if err != nil {
 			return nil, err
 		}
-		opts, err = embed.start()
+
+		opts, err := embed.start()
 		if err != nil {
+			if stopErr := embed.stop(); stopErr != nil {
+				logger.Warn("failed to stop embedded ClickHouse after startup failure", zap.Error(stopErr))
+			}
```
begelundmuller: Why this change? If it failed to start, it shouldn't call stop

runtime/drivers/clickhouse/clickhouse.go
```diff
@@ -308,91 +306,99 @@ func (d driver) Open(instanceID string, config map[string]any, st *storage.Clien
...
-		opts.MaxIdleConns = conf.MaxIdleConns
+		readOpts, writeOpts = opts, opts
+	case conf.Host != "":
+		// Build from individual host configuration
+		opts := buildOptsFromOptions(conf)
+		readOpts, writeOpts = opts, opts
+	case conf.WriteDSN != "":
+		return nil, errors.New("when not providing a 'dsn', both 'dsn' and 'write_dsn' must be specified for separate read/write connections")
+	default:
+		return nil, errors.New("must specify either 'dsn' for single connection or both 'dsn' and 'write_dsn' for dual connections")
```
begelundmuller: The `write_dsn` is an advanced internal setting, the old error message was more useful considering what properties end users are likely to set:
```
return nil, errors.New("no clickhouse connection configured: 'dsn', 'host' or 'managed: true' must be set")
```

runtime/drivers/clickhouse/clickhouse.go
```diff
@@ -308,91 +306,99 @@ func (d driver) Open(instanceID string, config map[string]any, st *storage.Clien
...
-		err := db.Ping()
-		if err != nil {
-			return nil, err
+	} else {
+		// Separate read and write connections
+		readDB = sqlx.NewDb(otelsql.OpenDB(clickhouse.Connector(readOpts)), "clickhouse")
+		writeDB = sqlx.NewDb(otelsql.OpenDB(clickhouse.Connector(writeOpts)), "clickhouse")
+
+		readDB.SetMaxOpenConns(conf.MaxOpenConns)
+		readDB.SetMaxIdleConns(conf.MaxIdleConns)
```
begelundmuller: There's a lot of duplication in the case handling here. Seeing as read and write handles are configured identically, a utility function makes sense here.

runtime/drivers/clickhouse/clickhouse.go
```diff
@@ -308,91 +306,99 @@ func (d driver) Open(instanceID string, config map[string]any, st *storage.Clien
...
+		}
 	}
 
 	// group by positional args are supported post 22.7 and we use them heavily in our queries
-	row := db.QueryRow(`
-		WITH
-			splitByChar('.', version()) AS parts,
-			toInt32(parts[1]) AS major,
-			toInt32(parts[2]) AS minor
-		SELECT (major > 22) OR ((major = 22) AND (minor >= 7)) AS is_supported
```
begelundmuller: Why change tabs to spaces?

runtime/drivers/clickhouse/clickhouse.go
```diff
@@ -308,91 +306,99 @@ func (d driver) Open(instanceID string, config map[string]any, st *storage.Clien
...
-	}
-	if opts.ReadTimeout == 0 { // Apply an increased default to reduce the chance of dropped connections with scaled-to-zero ClickHouse.
-		opts.ReadTimeout = time.Second * 300
-	}
 
-	db := sqlx.NewDb(otelsql.OpenDB(clickhouse.Connector(opts)), "clickhouse")
-	err = db.Ping()
-	if err != nil {
-		if !strings.Contains(err.Error(), "unexpected packet") && !strings.Contains(err.Error(), "i/o timeout") {
-			return nil, err
```
begelundmuller, line 361: It seems like this protocol discovery / error handling has been removed?

runtime/drivers/clickhouse/clickhouse.go
```diff
@@ -308,91 +306,99 @@ func (d driver) Open(instanceID string, config map[string]any, st *storage.Clien
...
+		if conf.ConnMaxLifetime != "" {
+			if d, err := time.ParseDuration(conf.ConnMaxLifetime); err == nil {
+				readDB.SetConnMaxLifetime(d)
+			}
 		}
-		opts.ReadTimeout = d
-	}
-	if opts.ReadTimeout == 0 { // Apply an increased default to reduce the chance of dropped connections with scaled-to-zero ClickHouse.
-		opts.ReadTimeout = time.Second * 300
-	}
```
begelundmuller, line 355: It seems like this default has been removed?

runtime/drivers/clickhouse/clickhouse.go
```diff
@@ -308,91 +306,99 @@ func (d driver) Open(instanceID string, config map[string]any, st *storage.Clien
...
+		err = applyCommonConfig(writeOpts, conf)
 		if err != nil {
-			return nil, fmt.Errorf("failed to parse dial_timeout: %w", err)
+			return nil, err
 		}
-		opts.DialTimeout = d
-	}
-	if opts.DialTimeout == 0 { // Apply an increased default to reduce the chance of dropped connections with scaled-to-zero ClickHouse.
-		opts.DialTimeout = time.Second * 60
 	}
```
begelundmuller: Also seems like this default has disappeared?

runtime/drivers/clickhouse/clickhouse.go
```diff
@@ -491,14 +511,26 @@ func (c *Connection) Config() map[string]any {
 func (c *Connection) Close() error {
 	c.cancel()
 
-	errDB := c.db.Close()
+	var errReadDB error
+	if c.readDB != nil {
```
begelundmuller: It's not possible to have an open connection that doesn't have this set. Lots of code relies on the `readDB` and `writeDB` not being nil, so it's weird to see checks like that here.

begelundmuller (approved):

PR #7691, opened 2025-07-29
Fixes for ClickHouse namespacing to honour the database specified

nishantmonu51:
Two improvements - 
- In readwrite mode, create database if it doesn't exist
- When a database is specified in connector config, use that for all crud operations. 
Fixes the error below - 
```
failed to create model: code: 60, message: Table prod.__rill_tmp_model_revenue does not exist. Maybe you meant default.__rill_tmp_model_revenue?: Insertion status:
Wrote 0 blocks and 0 rows on shard 0 replica 0 .... 
```

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/runtime/drivers/clickhouse/clickhouse.go
+++ b/runtime/drivers/clickhouse/clickhouse.go
@@ -731,12 +731,12 @@ func (c *Connection) ensureDatabaseExists(ctx context.Context) error {
 	var query string
 	if c.config.Cluster != "" {
 		// Use ON CLUSTER for clustered setup
-		query = fmt.Sprintf("CREATE DATABASE IF NOT EXISTS %s ON CLUSTER %s", 
-			safeSQLName(c.config.Database), 
+		query = fmt.Sprintf("CREATE DATABASE IF NOT EXISTS %s ON CLUSTER %s",
+			safeSQLName(c.config.Database),
 			safeSQLName(c.config.Cluster))
 	} else {
 		// Regular database creation
-		query = fmt.Sprintf("CREATE DATABASE IF NOT EXISTS %s", 
+		query = fmt.Sprintf("CREATE DATABASE IF NOT EXISTS %s",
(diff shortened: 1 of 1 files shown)


runtime/drivers/clickhouse/clickhouse.go
```diff
@@ -709,3 +717,34 @@ func (c *Connection) checkBillingTableExists(ctx context.Context, cluster string
...
+	}
+
+	// Add timeout to prevent hanging
+	ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
+	defer cancel()
+
+	var query string
+	if c.config.Cluster != "" {
+		// Use ON CLUSTER for clustered setup
+		query = fmt.Sprintf("CREATE DATABASE IF NOT EXISTS %s ON CLUSTER %s",
```
k-anshul, line 734: This does not create a replicated database. A replicated database requires more parameters like shard, zookeeper path, replica name etc. 
I think it is best to just throw an error right now.

runtime/drivers/clickhouse/crud.go
```diff
@@ -657,3 +660,11 @@ func tempTableForDictionary(name string) string {
 func safeSQLString(name string) string {
 	return drivers.DialectClickHouse.EscapeStringValue(name)
 }
+
+// qualifiedTableName returns a properly qualified table name using the database if configured
+func (c *Connection) qualifiedTableName(name string) string {
```
k-anshul, line 665: This is not required. The default database in the connection is set to the database set in configs : https://github.com/rilldata/rill/blob/b4f6eabcbb4cced92cab0423864dcff8febf6e44/runtime/drivers/clickhouse/clickhouse.go#L250-L251

k-anshul (changes requested):

nishantmonu51:
closing after testing with @k-anshul

PR #7695, opened 2025-07-30
validate driver configs and disallow individual props if DSN is set

k-anshul:
Otherwise leads to hard to detect bugs where individual properties are silently ignored.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


runtime/drivers/clickhouse/clickhouse.go
```diff
@@ -191,6 +191,20 @@ type configProperties struct {
 	ReadTimeout string `mapstructure:"read_timeout"`
 }
 
+func (c *configProperties) validate() error {
+	if c.DSN != "" {
+		if c.Host != "" || c.Port != 0 || c.Username != "" || c.Password != "" || c.Database != "" {
+			return errors.New("either 'dsn' or 'host', 'port', 'username', 'password', and 'database' must be set, but not both")
```
nishantmonu51: Can the error be a bit more custom, i.e. only show things that are configured. 
Add all non nulls in an array and then - 
Only one of DSN or [host, port] can be set.

runtime/drivers/pinot/pinot.go
```diff
@@ -118,6 +118,20 @@ type configProperties struct {
...
+func (c *configProperties) validate() error {
+	if c.DSN != "" && (c.BrokerHost != "" || c.ControllerHost != "") {
+		var set []string
+		if c.BrokerHost != "" {
+			set = append(set, "broker_host")
+		}
+		if c.ControllerHost != "" {
+			set = append(set, "controller_host")
+		}
+		return fmt.Errorf("pinot: Only one of 'dsn' or [%s] can be set", set)
```
begelundmuller: Missing `strings.Join(set, ", ")`?

runtime/drivers/postgres/postgres.go
```diff
@@ -78,6 +77,39 @@ type ConfigProperties struct {
...
+		dsn = c.DSN
+	} else {
+		dsn = c.DatabaseURL
+	}
+
+	if dsn != "" && (c.Host != "" || c.Port != "" || c.User != "" || c.Password != "" || c.DBname != "" || c.SSLMode != "") {
+		var set []string
+		if c.Host != "" {
+			set = append(set, "host")
+		}
```
begelundmuller: I think the outer `if` is not needed? In `clickhouse.go` you don't have it and just do:
```go
	if c.DSN != "" {
		if len(set) > 0 {
			return fmt.Errorf("only one of 'dsn' or [%s] can be set", strings.Join(set, ", "))
		}
	}
```

runtime/drivers/snowflake/snowflake.go
```diff
@@ -83,43 +83,78 @@ type configProperties struct {
...
-		return cp.DSN, nil
+func (c *configProperties) validate() error {
+	if c.DSN != "" && (c.Account != "" || c.User != "" || c.Database != "" || c.Password != "" || c.Schema != "" || c.Warehouse != "" || c.Role != "" || c.Authenticator != "" || c.PrivateKey != "") {
+		var set []string
+		if c.Account != "" {
+			set = append(set, "account")
+		}
+		if c.User != "" {
+			set = append(set, "user")
+		}
```
begelundmuller: Same comment as above – each property is checked twice, consider if can be simplified

runtime/drivers/postgres/postgres.go
k-anshul: In clickhouse I had to check the set twice so built it earlier. Here no need to build the set again so built it later but led to checking the code twice. 
But can make similar changes here as well

begelundmuller (approved):

PR #7700, opened 2025-07-30
use uniq alias for computed time dim in clickhouse comparison query

pjain1:
Clickhouse does not return correct results when using date function on a column name and using the same name in alias so use a different alias in this case.

A workaround till this is merged would be to define primary time dimension in dimension list with a different dim name than the column name and use the new name in `timeseries`.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


runtime/metricsview/executor_wrap_clickhouse_compare_time_dim.go
```diff
@@ -0,0 +1,73 @@
...
+// This does not work correctly in ClickHouse and it just does not subtract the interval from the time dimension and return "TIME_DIM" as it is.
+func (e *Executor) wrapClickhouseCompareTimeDim(ast *AST) {
+	if e.olap.Dialect() != drivers.DialectClickHouse || ast.query.ComparisonTimeRange == nil {
+		return
+	}
+
+	e.wrapClickhouseCompareTimeDimWalk(ast, ast.Root)
+}
+
+func (e *Executor) wrapClickhouseCompareTimeDimWalk(a *AST, n *SelectNode) {
```
pjain1: I think we can just assume that `comparison` node will always be present at the Root node of the AST as `JoinComparisonSelect`, that will simplify things and we  won't need this recursive walk. @begelundmuller what do you think?

begelundmuller: I think doing that may have some edge cases for advanced measures. I don't remember though, maybe in practice it's fine.

Would it be difficult to implement a walk similar to here: https://github.com/rilldata/rill/blob/7810a980fef44e7242e36f95f0ea6bf8781d40db/runtime/metricsview/executor_rewrite_druid_groups.go#L19

runtime/metricsview/executor_wrap_clickhouse_compare_time_dim.go
```diff
@@ -0,0 +1,73 @@
...
+	// find node with alias "comparison"
+	if n == nil {
+		return
+	}
+	if n.Alias == "comparison" {
```
begelundmuller: Instead of coupling to a specific alias, it might be more future proof to rely on `n.JoinComparisonSelect != nil`

i.e. do the operation at one higher level of recursion, like:
```go
if n.JoinComparisonSelect != nil {
  target := n.JoinComparisonSelect
  // ... rewrite target ...
}
```

runtime/metricsview/executor_wrap_clickhouse_compare_time_dim.go
```diff
@@ -0,0 +1,73 @@
...
+	"hash/crc32"
+	"strings"
+
+	"github.com/rilldata/rill/runtime/drivers"
+)
+
+// wrapClickhouseCompareTimeDim wraps the comparison AST if it has computed time dimension to have different alias than the time column name.
+// Example, in comparison queries we have expression like this: (date_trunc('day', "TIME_DIM") - INTERVAL (DATEDIFF('day', base_time_start, compare_time_start)) day) AS "TIME_DIM"
+// This does not work correctly in ClickHouse and it just does not subtract the interval from the time dimension and return "TIME_DIM" as it is.
+func (e *Executor) wrapClickhouseCompareTimeDim(ast *AST) {
```
begelundmuller: Does this issue only apply to comparisons, or could it also be a problem for where expressions? For example, what about:
```sql
SELECT date_trunc('day', event_time) AS event_time, ...
FROM tbl
WHERE event_time >= '2025-08-01T17:00:00Z'::TIMESTAMP
```

In this example, will the `WHERE` clause check against the raw `event_time` (expected) or the `date_trunc` result (not expected)?

For other dimensions/measures, we require that the name doesn't overlap with an underlying column – I wonder if we need something similar for the time dimension (maybe by emitting it with a fixed name `__time_dim` from the parser?): https://github.com/rilldata/rill/blob/7810a980fef44e7242e36f95f0ea6bf8781d40db/runtime/metricsview/executor_validate.go#L100-L120

pjain1: You are correct its using `date_trunc` results so that is unexpected. Emitting a fixed name can be problematic - do we overwrite the already defined `timeseries` name in metrics view, also user will have to be aware of this and use this specific name in metrics_sql api.

begelundmuller: My idea above would be that if the user specifies `timeseries: event_time`, we emit a time dimension `{name: "__time_dim", column: "event_time}`. Yes it might break APIs and metrics SQL where the time dimension is referenced by name. We'd need to check if any projects have that (`rill sudo project dump-resources` + `grep`). I think reports and stateful URLs generally don't store the default time dimension name, so I think it might be backwards compatible there, but would also have to check.

Another idea perhaps is that for Clickhouse, we wrap the innermost select to add an alias, e.g.:
```sql
SELECT date_trunc('day', __time_dim) AS event_time, ...
-- FROM underlying
FROM (SELECT event_time AS __time_dim, * FROM underlying)
```

Can you think of other ideas?

pjain1: But users can now use other timestamp type columns in the API calls for which we may not emit this time dimension, also chances of collision if a column named `__time_dim` already defined. I did not fully understood your other example but overall idea looks like whenever there is a computed time dim field and the name is same as underlying column we should have a uniq alias for it in the inner query doing the work and wrap them in an outer query which just returns results from inner query with the expected alias which is same as time dim name, is this correct understanding?

pjain1:
Should we also add validation if timestamp types is defined in the dimension list then it should have different name than the column name, that way in future we may not even need wrapping if all timestamp types (including `timeseries`) is defined in the dim list. But will need to check it should not break any existing mv.

begelundmuller (approved):

begelundmuller:
> Should we also add validation if timestamp types is defined in the dimension list then it should have different name than the column name, that way in future we may not even need wrapping if all timestamp types (including timeseries) is defined in the dim list. But will need to check it should not break any existing mv.

Yeah that might be good, but probably only needed for Clickhouse. Maybe it could go here? https://github.com/rilldata/rill/blob/042142a6b2bb5a538fe71337d9a03d205a964985/runtime/metricsview/executor_validate.go#L100-L120

pjain1:
> Yeah that might be good, but probably only needed for Clickhouse.

That can go [here](https://github.com/rilldata/rill/pull/7709). Merging this now.

PR #7707, opened 2025-07-31
no default limit for timeseries request

pjain1:
Fixes partial results for timeseries request because of default limit 

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/runtime/queries/metricsview_aggregation.go
+++ b/runtime/queries/metricsview_aggregation.go
@@ -412,10 +412,6 @@ func (q *MetricsViewAggregation) rewriteToMetricsViewQuery(export bool) (*metric
 
 	if q.Limit != nil {
 		if *q.Limit == 0 {
-			if q.shouldAddLimit() {
-				tmp := int64(100)
-				q.Limit = &tmp
-			}
 			q.Limit = nil
 		}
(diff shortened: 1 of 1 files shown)


runtime/queries/metricsview_aggregation.go
```diff
@@ -525,6 +528,14 @@ func (q *MetricsViewAggregation) getDisplayName(ctx context.Context, rt *runtime
 	return resourceName.Name, nil
 }
 
+// timeseries request will not have any limit set, so we should not set any default limit if time range is specified otherwise timeseries graph will be partial
+func (q *MetricsViewAggregation) shouldAddLimit() bool {
```
pjain1: This is a very simple fix, should we do better limit calculation in metricsview package after time range is resolved and look for computed time dimension in the query and use its grain to calculate the limit.

begelundmuller (commented):
I think it’s maybe a little weird to conditionally add the 100 row default. Maybe we should just remove the 100 row limit and let it stay unlimited by default? There’s still some protection from the underlying rill.interactive_sql_row_limit setting (default 10k, errors if more rows than that are returned).

begelundmuller (approved):

PR #7712, opened 2025-08-01
PLAT-28 : Disable enable_external_access for duckdb for finer controls

k-anshul:
INSERT DESCRIPTION HERE

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/runtime/drivers/duckdb/duckdb.go
+++ b/runtime/drivers/duckdb/duckdb.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"maps"
 	"net/url"
+	"path/filepath"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


runtime/drivers/duckdb/duckdb.go
```diff
@@ -497,8 +498,30 @@ func (c *connection) reopenDB(ctx context.Context) error {
 	// We want to set preserve_insertion_order=false in hosted environments only (where source data is never viewed directly). Setting it reduces batch data ingestion time by ~40%.
 	// Hack: Using AllowHostAccess as a proxy indicator for a hosted environment.
 	if !c.config.AllowHostAccess {
+		extensionDir, err := extensions.ExtensionsDir()
+		if err != nil {
+			return err
+		}
+
+		// Find the instance dir for the data dir
```
k-anshul: This is not ideal but at present there is no way to get instance dir from storage. The storage driver does not have concept of instance.

k-anshul:
Please hold review on this. Looks like there are edge cases in our test setups that we need to handle.

runtime/drivers/duckdb/duckdb.go
```diff
@@ -544,9 +545,42 @@ func (c *connection) reopenDB(ctx context.Context) error {
...
+			fmt.Sprintf("SET secret_directory=%s", safeSQLString(secretDir)),
 		)
+		// Find the instance dir for the data dir
+		// other drivers always write to a path which is a subdirectory of the instance directory
+		tempDir, err := c.storage.TempDir()
+		if err != nil {
+			return err
+		}
+		if strings.Contains(tempDir, c.instanceID) {
+			for tempDir != "" && tempDir != string(filepath.Separator) && tempDir != "." {
```
begelundmuller: I'm a bit worried this breaks later if we change some paths. Also about edge cases or future bugs in this for loop that might cause an infinite loop.

I think we should either add a concept of instances to the storage client (e.g. `storage.InstanceRoots()`), or else just skip the `allowed_directories` setting, which I'm also okay with at this stage (I don't trust the DuckDB sandboxing that much anyway, and `secret_directory` seems like the most important setting as it avoids accidents).

PR #7728, opened 2025-08-05
add time dimension to resolver cache key

pjain1:
Add time dimension to `metricsview_time_range` resolver cache key as alternate time dimensions can be used now to query APIs. Mostly affects mcp client doing mv time range query with alternate time dim.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/runtime/resolvers/metricsview_time_range.go
+++ b/runtime/resolvers/metricsview_time_range.go
@@ -100,7 +100,12 @@ func (r *metricsViewTimeRangeResolver) Close() error {
 }
 
 func (r *metricsViewTimeRangeResolver) CacheKey(ctx context.Context) ([]byte, bool, error) {
-	return cacheKeyForMetricsView(ctx, r.runtime, r.instanceID, r.mvName, r.args.Priority)
+	key, ok, err := cacheKeyForMetricsView(ctx, r.runtime, r.instanceID, r.mvName, r.args.Priority)
+	if err != nil {
+		return nil, false, err
+	}
+	key = append(key, []byte(r.args.TimeDimension)...)
+	return key, ok, nil
 }
 
 func (r *metricsViewTimeRangeResolver) Refs() []*runtimev1.ResourceName {


runtime/resolvers/metrics.go
```diff
@@ -97,8 +97,12 @@ func (r *metricsResolver) Close() error {
 }
 
 func (r *metricsResolver) CacheKey(ctx context.Context) ([]byte, bool, error) {
+	timeDim := "" // mv default time dimension will be used if empty
+	if r.query.TimeRange != nil && r.query.TimeRange.TimeDimension != "" {
+		timeDim = r.query.TimeRange.TimeDimension
+	}
```
begelundmuller: What is the use case for this? Even if the metrics view supports multiple time dimensions, I wonder if we shouldn't still rely on the primary time dimension (i.e. the watermark) to determine if there's new data or not.

For example, if caching is enabled and the primary time dimension is `event_time`, then any queries (regardless of whether they use normal dimensions or other time dimensions) will stay cached until `event_time` advances.

Or am I not understanding this correctly?

pjain1: Yes that was my thinking too and did not add time dim to cache key for metrics resolver. However, @k-anshul has some arguments about having to do it for each time dimension, for example when alternate time dimension had lower grain than primary one and primary might not advance on adding more rows. Even in that case I think its on user to make sure give correct cache key sql or lowest grain time dim for watermark. So more of defensive approach I think.

I find it confusing that metrics cache key depends on time dimension used to query, conceptually it should be independent. All queries and other resolvers already add time dim in their cache key along side mv cache key.

What do you think?

begelundmuller: > Even in that case I think its on user to make sure give correct cache key sql or lowest grain time dim for watermark

I probably agree that it's a user concern

> I find it confusing that metrics cache key depends on time dimension used to query, conceptually it should be independent. 

Agree

PR #7729, opened 2025-08-05
Refresh cookie on accessing v1/users/current

NamanMahor:
[PLAT-120: Pentest: Implement definitive session timeout](https://linear.app/rilldata/issue/PLAT-120/pentest-implement-definitive-session-timeout)

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!

### DIFF
--- a/admin/server/auth/middleware.go
+++ b/admin/server/auth/middleware.go
@@ -90,6 +90,21 @@ func (a *Authenticator) HTTPMiddlewareLenient(next http.Handler) http.Handler {
 	return a.httpMiddleware(next, true)
 }
 
+// HTTPMiddlewareWithCookieRefresh is a middleware that refreshes the auth cookie
+func (a *Authenticator) HTTPMiddlewareWithCookieRefresh(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		sess := a.cookies.Get(r, cookieName)
+		if authToken, ok := sess.Values[cookieFieldAccessToken].(string); ok && authToken != "" {
+			// Re-save the cookie to refresh its expiration
+			if err := sess.Save(r, w); err != nil {
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+				return
+			}
+		}
+		next.ServeHTTP(w, r)
+	})
+}
+
 // httpMiddleware is the actual implementation of HTTPMiddleware and HTTPMiddlewareLenient.
 func (a *Authenticator) httpMiddleware(next http.Handler, lenient bool) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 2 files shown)


admin/server/auth/middleware.go
```diff
@@ -90,6 +90,21 @@ func (a *Authenticator) HTTPMiddlewareLenient(next http.Handler) http.Handler {
 	return a.httpMiddleware(next, true)
 }
 
+// HTTPMiddlewareWithCookieRefresh is a middleware that refreshes the auth cookie
+func (a *Authenticator) HTTPMiddlewareWithCookieRefresh(next http.Handler) http.Handler {
```
begelundmuller: 1. Can you name this `CookieRefreshHTTPMiddleware`? It sounds like this is similar to the `HTTPMiddleware` above, but that is not the case – this middleware does something different.
2. Maybe add a bit extra info in the docstring, e.g.:
```suggestion
// CookieRefreshMiddleware is a middleware that refreshes the auth cookie.
// This enables us to do rolling cookie refreshes so we can have a relatively short cookie max age.
// Note that it does not update the auth token encrypted inside the cookie.
func (a *Authenticator) CookieRefreshMiddleware(next http.Handler) http.Handler {
```

admin/server/server.go
```diff
@@ -185,6 +185,10 @@ func (s *Server) HTTPHandler(ctx context.Context) (http.Handler, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed to create transcoder: %w", err)
 	}
+
+	// Add specific route for GetCurrentUser with cookie refresh
+	mux.Handle("/v1/users/current", s.authenticator.HTTPMiddlewareWithCookieRefresh(transcoder))
```
begelundmuller: Would be useful with some extra context:
```suggestion
	// Add auth cookie refresh specifically for the GetCurrentUser RPC.
	// This is sufficient to refresh the cookie on each page load without unnecessarily refreshing cookies in each API call.
	mux.Handle("/v1/users/current", s.authenticator.HTTPMiddlewareWithCookieRefresh(transcoder))
```

begelundmuller (commented):
Lastly, can you also:
1. Check that this actually works? (E.g. by artificially reducing the cookie age to a lower time and testing it manually.)
2. Check that replay attacks with the expired cookie is not possible? I.e. that the backend checks the cookie payload itself for expiration, so we don't just rely on browsers to enforce the expiration.

NamanMahor:
> Lastly, can you also:
> 
> 1. Check that this actually works? (E.g. by artificially reducing the cookie age to a lower time and testing it manually.)

Yes it updating the cookie expire time on calling v1/user/current. 

> 2. Check that replay attacks with the expired cookie is not possible? I.e. that the backend checks the cookie payload itself for expiration, so we don't just rely on browsers to enforce the expiration.

403 error is coming when we try with expired cookie.

PR #7733, opened 2025-08-05
Use infinite query when listing org users

lovincyrus:
This PR resolves the pagination issue where user search bars in group management dialogs only showed the first 20 users, preventing users from finding and adding users beyond that limit in their organization.

Closes https://linear.app/rilldata/issue/APP-252/user-search-bar-does-not-paginate-beyond-first-20-users

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


ericpgreen2 (approved):
Code LGTM

PR #7734, opened 2025-08-05
Increasing grpc received message size

NamanMahor:
Fixes [PLAT-134: Fix `rill project logs` exceeding gRPC message size](https://linear.app/rilldata/issue/PLAT-134/fix-rill-project-logs-exceeding-grpc-message-size)

Tested by pointing rill project log to local runtime.
Before fix failed with
```
Error: failed to get logs: rpc error: code = ResourceExhausted desc = grpc: received message larger than max (15733798 vs. 4194304) (ResourceExhausted)
```
After fix working fine.

No need to change server side send message size as its already int max by default.

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!

### DIFF
--- a/cli/cmd/project/logs.go
+++ b/cli/cmd/project/logs.go
@@ -11,6 +11,7 @@ import (
 	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
 	runtimeclient "github.com/rilldata/rill/runtime/client"
 	"github.com/spf13/cobra"
+	"google.golang.org/grpc"
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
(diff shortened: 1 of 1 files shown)


cli/cmd/project/logs.go
```diff
@@ -101,7 +102,9 @@ func LogsCmd(ch *cmdutil.Helper) *cobra.Command {
 				return context.Cause(ctx)
 			}
 
-			res, err := rt.GetLogs(cmd.Context(), &runtimev1.GetLogsRequest{InstanceId: depl.RuntimeInstanceId, Ascending: true, Limit: int32(tail), Level: lvl})
+			res, err := rt.GetLogs(cmd.Context(),
+				&runtimev1.GetLogsRequest{InstanceId: depl.RuntimeInstanceId, Ascending: true, Limit: int32(tail), Level: lvl},
```
pjain1, line 106: Do we have any limits on message size at nginx proxy level or grpc gateway otherwise this will not be much useful.

NamanMahor: since this exception is coming at client side. we know that client have received the message from server and proxies in between.

pjain1: makes sense, approved

pjain1 (approved):

PR #7735, opened 2025-08-05
Implement mode `readwrite` for Motherduck and DuckLake

grahamplata:
Implement `mode: readwrite` for MotherDuck and DuckLake

**Checklist:**
- [x] Covered by tests
- [x] Linked the [issues it closes](https://linear.app/rilldata/issue/PLAT-152/implement-mode-readwrite-for-motherduckducklake-similar-to-clickhouse)
- [x] I'm proud of this work!


runtime/drivers/duckdb/duckdb.go
```diff
@@ -127,6 +127,16 @@ var motherduckSpec = drivers.Spec{
 			Placeholder: "my_new_source",
 			Required:    true,
 		},
+		{
```
k-anshul: I think this should be part of config properties and not source properties. The source properties apply when we are creating a source that extracts data from motherduck.

avaitla:
Will this support querying MotherDuck views (not just tables)? Also can we point one source to use Rill's embedded Duckdb and one to use motherduck connection? The reason for this is to be able to incrementally move each source from local duckdb to md without having to move them all (and if there are issues revert).

runtime/drivers/duckdb/duckdb.go
```diff
@@ -155,6 +165,18 @@ func (d Driver) Open(instanceID string, cfgMap map[string]any, st *storage.Clien
...
+	if cfg.Mode == "" {
+		// The default mode depends on the connection type:
+		// - For generic connections (Motherduck/DuckLake with Path/Attach), default to "read"
+		// - For connections using the embedded DuckDB, default to "readwrite" to maintain compatibility
+		if cfg.Path != "" || cfg.Attach != "" {
+			cfg.Mode = modeReadOnly
+		} else {
+			cfg.Mode = modeReadWrite
+		}
+	}
```
begelundmuller: The logic here looks good, but consider moving it into `newConfig` since that function currently both does validation and applies defaults.

begelundmuller (commented):
Two other requests:
1. Can you document the new config key in `runtime/parser/schema/project.schema.yaml`
2. Can you check if this impacts any prod projects (i.e. projects that use external DuckDB AND use modeling)? And if it does, preemptively add the `mode: readwrite` to them (or reach out to the maintainers)

runtime/drivers/duckdb/duckdb.go
```diff
@@ -355,6 +412,9 @@ func (c *connection) AsObjectStore() (drivers.ObjectStore, bool) {
 
 // AsModelExecutor implements drivers.Handle.
 func (c *connection) AsModelExecutor(instanceID string, opts *drivers.ModelExecutorOptions) (drivers.ModelExecutor, error) {
+	if c.config.Mode != modeReadWrite {
+		return nil, fmt.Errorf("model execution is disabled. To enable modeling on this ClickHouse database, set 'mode: readwrite' in your connector configuration. WARNING: This will allow Rill to create and overwrite tables in your database")
```
NamanMahor, line 416: message should be for duckdb.

PR #7739, opened 2025-08-06
fix: github connection always throwing a popup

AdityaHegde:
We were throwing the github auth popup even if user already had access.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


AdityaHegde:
It was intended to use the other method because of the way our flow works. This will be changed to create a repo through rill in a future PR.

PR #7741, opened 2025-08-06
Managed database name generation

grahamplata:
Edge cases found while testing managed database name generations.

**Checklist:**
- [x] Covered by tests
- [x] [Linked the issues it closes](https://linear.app/rilldata/issue/PLAT-155/fix-clickhouse-provisioner-database-names)
- [x] Checked if the docs need to be updated. If so, create a separate Linear DOCS issue. No Need internal
- [ ] I'm proud of this work!

### DIFF
--- a/admin/provisioner/clickhousestatic/provisioner.go
+++ b/admin/provisioner/clickhousestatic/provisioner.go
@@ -338,5 +338,5 @@ func generateDatabaseName(resourceID string, annotations map[string]string) stri
 		name = name[:63]
 	}
 	name = strings.TrimRight(name, "_")
-	return name
+	return strings.ToLower(name)
 }
--- a/admin/provisioner/clickhousestatic/provisioner.go
+++ b/admin/provisioner/clickhousestatic/provisioner.go
@@ -130,9 +130,16 @@ func (p *Provisioner) Provision(ctx context.Context, r *provisioner.Resource, op
 	}
 
 	// Prepare for creating the schema and user.
-	id := sanitizeName(r.ID)
+	id := strings.ReplaceAll(r.ID, "-", "")
 	user := fmt.Sprintf("rill_%s", id)
-	dbName := generateDatabaseName(id, opts.Annotations)
+	dbName := fmt.Sprintf("rill_%s", id)
+
+	// Use org and project names to create a more human-readable database name.
+	orgName := sanitizeName(getAnnotationValue(opts.Annotations, "organization_name"))
+	projectName := sanitizeName(getAnnotationValue(opts.Annotations, "project_name"))
+	if orgName != "" && projectName != "" {
+		dbName = fmt.Sprintf("rill_%s_%s_%s", orgName, projectName, id)
+	}
 
 	password := newPassword()
 	annotationsJSON, err := json.Marshal(opts.Annotations)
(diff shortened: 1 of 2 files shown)


admin/provisioner/clickhousestatic/provisioner.go
```diff
@@ -148,67 +141,41 @@ func (p *Provisioner) Provision(ctx context.Context, r *provisioner.Resource, op
 	}
 
 	// Idempotently create the schema
-	_, err = p.ch.ExecContext(ctx, fmt.Sprintf("CREATE DATABASE IF NOT EXISTS %s %s COMMENT ?", dbName, p.onCluster()), string(annotationsJSON))
+	_, err = p.ch.ExecContext(ctx, fmt.Sprintf("CREATE DATABASE IF NOT EXISTS `%s` %s COMMENT ?", dbName, p.onCluster()), string(annotationsJSON))
```
begelundmuller: For our own sanity, can we avoid using names that require escaping? Just filter them out when generating the DB name.

And if escaping is needed, use a utility function around the parameter instead of doing it in the SQL statement itself. For example, if `dbName` contains a backtick, this query would break.

admin/provisioner/clickhousestatic/provisioner.go
```diff
@@ -148,67 +141,41 @@ func (p *Provisioner) Provision(ctx context.Context, r *provisioner.Resource, op
...
-			DROP TABLE,
-			DROP DICTIONARY,
-			DROP VIEW,
-			TRUNCATE,
-			OPTIMIZE,
-			SHOW DICTIONARIES,
-			dictGet
-		ON %s.* TO %s
-	`, p.onCluster(), dbName, user))
+	_, err = p.ch.ExecContext(ctx, fmt.Sprintf(`GRANT %s SELECT, INSERT, ALTER, CREATE TABLE, CREATE DICTIONARY, CREATE VIEW, DROP TABLE, DROP DICTIONARY, DROP VIEW, TRUNCATE, OPTIMIZE, SHOW DICTIONARIES, dictGet ON `+"`%s`"+`.* TO `+"`%s`", p.onCluster(), dbName, user))
```
begelundmuller: This feels like readability took a hit vs. the previous code. See comment above – use a utility function like `sqlEscapeName(dbName)` if needed, but in this case, let's avoid special characters in internally generated names altogether.

admin/provisioner/clickhousestatic/provisioner.go
```diff
@@ -357,3 +324,36 @@ func sanitizeName(name string) string {
 
 	return strings.ToLower(result.String())
 }
+
+// generateDatabaseName creates a predictable, length-controlled database name
+func generateDatabaseName(resourceID string, annotations map[string]string) string {
```
begelundmuller, line 327: Feels weird to have both this and `sanitizeName`. And we should really avoid special characters in the DB names. There's a unique ID appended so no need to be super exact here, it's just some helpful context.

Why not just something like:
```suggestion
var nonAlphanumericRegexp := regexp.MustCompile(`[^a-zA-Z0-9]+`)

func generateDatabaseName(resourceID string, annotations map[string]string) string {
  name := "rill"
  if org, ok := annotations["organization_name"]; ok {
    name += "_" + nonAlphanumericRegexp.ReplaceAllString(org, "")
  }
  if proj, ok := annotations["project_name"]; ok {
    name += "_" + nonAlphanumericRegexp.ReplaceAllString(proj, "")
  }
  name += "_" + nonAlphanumericRegexp.ReplaceAllString(resourceID, "")
  return name
}
```

PR #7743, opened 2025-08-07
chore: keep only relevant examples

mindspank:
Only keeps the example projects that are actually used in the UI to reduce file size.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,16 +10,20 @@ cli-only:
 cli: cli.prepare
 	go build -o rill cli/main.go 
 
+KEEP_DIRS := rill-openrtb-prog-ads rill-github-analytics rill-cost-monitoring
+
 .PHONY: cli.prepare
 cli.prepare:
-	rm -rf runtime/pkg/examples/embed/dist || true
-	mkdir -p runtime/pkg/examples/embed/dist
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -12,15 +12,19 @@ cli: cli.prepare
...
 .PHONY: cli.prepare
 cli.prepare:
+	rm -rf runtime/pkg/examples/embed/dist || true
+	mkdir -p runtime/pkg/examples/embed/dist
+	git clone --quiet https://github.com/rilldata/rill-examples.git runtime/pkg/examples/embed/dist
+	find runtime/pkg/examples/embed/dist -mindepth 1 -maxdepth 1 -type d \
+	! -name 'rill-openrtb-prog-ads' \
+	! -name 'rill-github-analytics' \
+	! -name 'rill-cost-monitoring' \
+	-exec rm -rf {} +
```
begelundmuller: I think selecting just a few examples is fine. But instead of `rm -rf {}` could we make it clone `rill-examples` to `$TMPDIR` and then copy over just these example projects into the embed directory? 

Mainly I'm a bit scared about any `rm -rf` call that is not very very explicit. But also want to make completely sure we don't e.g. copy `.git` or other non-important files.

begelundmuller (approved):

PR #7745, opened 2025-08-07
fix: Disable comparison with rill time in alerts and reports

AdityaHegde:
We cant really do correct comparison without some additional rill time syntax support. So temporarily disabling comparisons in alerts an reports.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


briangregoryholmes (approved):

briangregoryholmes:
What do we think the ETA is for re-enabling this?

AdityaHegde:
Hmmm we would need to add support for `offset 2M` like support in syntax. Not too much effort, could be squeezed in for next release, not tomorrow's release though.

PR #7747, opened 2025-08-07
Allow add data source when OLAP is duckdb

lovincyrus:
INSERT DESCRIPTION HERE

Closes https://linear.app/rilldata/issue/APP-250/make-sure-we-are-generating-the-correct-snippets-for-add-data-source

CleanShot 2025-08-07 at 12 28 00@2x

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-common/src/features/entity-management/AddAssetButton.svelte
+++ b/web-common/src/features/entity-management/AddAssetButton.svelte
@@ -45,7 +45,7 @@
 
   $: ({ instanceId } = $runtime);
 
-  // Get current OLAP connector to check if it's not duckdb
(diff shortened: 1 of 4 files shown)


lovincyrus:
> Unfortunately, this approach removes the UI for someone to add another OLAP engine. Let's say someone started with Druid, then wanted to move to ClickHouse– the modal wouldn't be available for them. So, instead, can we just remove/disable the "Sources" section of the "Add Data" modal, so we can retain the "Connect an OLAP engine" section?

https://linear.app/rilldata/issue/APP-250/make-sure-we-are-generating-the-correct-snippets-for-add-data-source#comment-1c041dfd

PR #7748, opened 2025-08-07
[DOCS] Custom 404 page

royendo:
ScreenShot Tool -20250807134842

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/docs/src/theme/NotFound/Content/index.tsx
+++ b/docs/src/theme/NotFound/Content/index.tsx
@@ -43,7 +43,7 @@ export default function NotFoundContent({
                         color: "#1a1a1a",
                         fontSize: "1.125rem",
                         fontWeight: 600,
-                        lineHeight: "0.875rem",
+                        lineHeight: "1.5rem",
                         marginBottom: "1rem",
                     }}
                 >


royendo:
@ericpgreen2  thoughts on an auto redirect to home?

ericpgreen2:
> @ericpgreen2 thoughts on an auto redirect to home?

In its current form:
1. There's too much on the page to read, so it's frustrating that the page redirects before I get a chance to read it. Especially the fact that the content contains a URL, I get this feeling of FOMO that I might've preferred to click whatever URL was presented to me.
2. I'd expect the "Redirecting automatically in 5 seconds" would count down "5, 4, 3, 2, 1". Then the redirect would feel like less of a surprise.
3. "The page you're looking for has moved" – this copy implies that their used to be a page there. But when I navigate to `/page-never-existed`, the copy is misleading.

I also asked Claude about it and agree with the answer:
image

That being said, we could definitely do better than our current 404 page, which copy ("Please contact the owner of the site that linked you to the original URL and let them know their link is broken.") sounds like it's the Docusaurus default copy. But IMO we don't need to do much more than make the copy sound like it's not the Docusaurus placeholder. I like the friendly tone of the copy that you have in this PR.

royendo:
sounds good! ill remove the auto direct and just make it a "better" 404

Di7design:
> A few design nits:
> 
> 1. The "Oops! The page you're looking for has moved" text seems a touch too large?
> 2. The 3 paragraphs seem to have funky spacing between them. E.g. there's more spacing between the 2nd and 3rd paragraphs than between the 1st and 2nd. And the spacing doesn't look like typical spacing between paragraphs.
> 3. "Go to Home Page" -> "Return to docs". We'll at least want to use sentence case, not title case.
> 
> @Di7design, perhaps you might have further design feedback on this page?
 
Good call. I can provide a design for this page. What's the priority for this issue?

ericpgreen2:
> Good call. I can provide a design for this page. What's the priority for this issue?

We shouldn't spend too much time on it– we should do an incremental improvement that's quick.

Di7design:
Check the design [here.](https://www.figma.com/design/JtG3sbaopjO0xQlyeCjmho/RILL-WIP?node-id=266-138466&t=tb0wUWHTsX8MRgTI-1)
Screenshot 2025-10-02 at 11 52 17 AM

royendo:
Screenshot 2025-10-06 at 13 25 05

royendo:
oops

ericpgreen2 (changes requested):
I'm still observing that the font size is not matching Di's designs. 

For example, the "The page you're looking for..." text:
- In her design, I see 16px.
- In this PR, I'm seeing 1.25rem, which by default is 1.25 * 16px = 20px.

royendo:
hmm weird maybe its my monitor/aspect ratios that looked more closer to what was in the figma. Adjusted fontsize, lineheight, to px

royendo:
typescript is an easy fix, i have no idea why docusuarus is fighting against a single file tho

royendo:
retested ejecting the NotFound component from Docusuarus, and it defaults to two files, IMO best to keep it in default format :)

ericpgreen2:
## Line Height Issue in Heading

There's a likely typo at `docs/src/theme/NotFound/Content/index.tsx:46`:

```tsx
lineHeight: "0.875rem",  // 14px
```

This line height (14px) is **smaller** than the font size (18px), which will cause text clipping and rendering issues. Line height should always be greater than or equal to font size.

**Suggested fix:**
```tsx
lineHeight: "1.5rem",  // or "1.75rem"
```

This appears to be a typo where the leading "1" was accidentally omitted.

PR #7763, opened 2025-08-11
use regexp like for druid

pjain1:
To fix search on multi valued dimension column

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/runtime/metricsview/astexpr.go
+++ b/runtime/metricsview/astexpr.go
@@ -378,6 +378,46 @@ func (b *sqlExprBuilder) writeILikeCondition(left, right *Expression, leftOverri
 		if b.ast.dialect.RequiresCastForLike() {
 			b.writeString("::TEXT")
 		}
+	} else if b.ast.dialect.SupportsRegexMatch() {
+		if not {
+			b.writeString(" NOT ")
+		}
+		b.writeString(b.ast.dialect.GetRegexMatchFunction())
+		b.writeByte('(')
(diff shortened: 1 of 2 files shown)


runtime/metricsview/astexpr.go
```diff
@@ -658,6 +698,17 @@ func (b *sqlExprBuilder) sqlForName(name string) (expr string, unnest bool, look
...
+func convertLikeExpressionToRegexExpression(like *Expression) (*Expression, error) {
+	val, ok := like.Value.(string)
+	if !ok {
+		return nil, fmt.Errorf("the pattern expression for regex match function must be a string value, got %T", like.Value)
+	}
+	// convert pattern to a case insensitive regex match pattern, e.g. "%foo%" becomes "^(?i).*foo.*$"
+	pattern := strings.ReplaceAll(val, "%", ".*")
+	pattern = fmt.Sprintf("^(?i)%s$", pattern)
+	return &Expression{Value: pattern}, nil
+}
```
begelundmuller, line 710: Would this fit better in the dialect? E.g. 
```go
...
} else if newExpr, ok :=b.ast.dialect.RewriteLikeExpressionToRegex(leftExpr, rightExpr); ok {
  // use newExpr
```
And then the dialect can return `"", false` if it should not be rewritten?

begelundmuller: Not sure, just a thought

PR #7767, opened 2025-08-12
[DOCS] Add navigational side bar

royendo:
onhover: when hovering the side bar illuminates

Screenshot 2025-08-12 at 11 21 06

onClick: on select the text bolds, and illuminates disappears
Screenshot 2025-08-12 at 11 20 59

This way is a bit less noisy on the side bar. 

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/docs/src/css/_sidebar.scss
+++ b/docs/src/css/_sidebar.scss
@@ -26,7 +26,7 @@ li.theme-doc-sidebar-item-category-level-1 {
   }
   ul.menu__list {
     margin-left: 0.8rem;
-    border-left: 1px solid #ccc;
+    border-left: 1px solid  var(--ifm-menu-sidebar-color);
   }
   .menu__list-item:hover {
       border-left: 1px solid var(--ifm-color-primary);
@@ -60,6 +60,11 @@ li.theme-doc-sidebar-item-category-level-2 .menu__list-item-collapsible {
   &--active {
     font-weight: 600;
   }
+  &:hover {
+    color: var(--ifm-color-primary);
+    font-weight: 600;
+  }
+
 }
 
 .menu__caret:before,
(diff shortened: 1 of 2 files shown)


docs/src/css/_sidebar.scss
```diff
@@ -22,13 +22,30 @@ li.theme-doc-sidebar-item-category-level-1 {
...
   }
   &:not(:first-of-type) .menu__list-item-collapsible {
-    margin-top: var(--space-3); /* Standardized spacing */
+    margin-top: var(--space-3);
   }
-  &:last-of-type {
-    display: block;
+  ul.menu__list {
+    margin-left: 0.8rem;
+    border-left: 1px solid #ccc;
```
ericpgreen2: `#ccc` should pull from our parameterized border colors so that it looks good in dark mode too

ericpgreen2 (changes requested):
1. Can the `on:hover` highlighted border only apply to the immediate level, not to the parent level? Here, when highlighting "GCS", it's a bit distracting that the whole "Data Sources" level gets highlighted
    image
2. Could we add an additional hover state when hovering over a menu item? Some options: 
    - the font could bold
    - the font could change to a "link" color
    - the text could be underlined

    With some text interaction on hover, it may be the case that we don't additionally need the guideline to change color (though it still might be nice).

docs/src/css/_sidebar.scss
royendo: done

royendo:
1. trying to figure out how to do that, so many nested classes generated by docusaurus


2. BQ is open, GCS is hovered
Screenshot 2025-08-12 at 19 53 21

commented out hover in latest commit

royendo:
yeah, no effect on the side bar line and just hover / active for the text seems enoiugh

ericpgreen2 (approved):
Nice, the hover text especially feels great now!

PR #7779, opened 2025-08-14
[APP-156] feat: canvas default filter support

briangregoryholmes:
This PR adds support for enumerating default filters on Canvas dashboards via the YAML file. It includes the following capabilities:

- Specifying filters as being `locked`
- Displaying temporary measures or dimensions without preselected values
- Specifying whether filters are `removable`
- Preventing filters from displaying in the ui via `hidden`
- Setting a `limit` on the number of values in a dimension filter
- Hide the add filter button via `allow_filter_add`

```
defaults:
  time_range: PT24H
  comparison_mode: time
  filters:
    measures:
      - measure: impressions
        locked: true
        hidden: true
        by_dimension: adomain
        operator: gt # (gt, lt, gte, lte, bt, nbt, eq, neq)
        values:
          - 10
    dimensions:
      - dimension: adomain
      - dimension: advertiser_name
        removable: false
      - dimension: app_site_name
        locked: true
        limit: 1
        values:
          - Amerigo
        
 ```

Outstanding items:

- [x] Display list of values in tooltip when a filter is locked

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!

### CODE REVIEW

**Comment by ericokuma** [comment]
@mindspank, should we go ahead with approving this?

File: web-common/src/features/canvas/stores/canvas-entity.ts [hunk 1]
```diff
@@ -80,6 +81,133 @@ export class CanvasEntity {
...
+      }
+    });
+
+    if (this.fileArtifact) {
+      const yamlContent = get(this.fileArtifact.remoteContent);
+
+      const parsed = parseDocument(yamlContent ?? "");
+
+      parsed.setIn(["defaults", "filters", "dimensions"], yamlDimensions);
+      parsed.setIn(["defaults", "filters", "measures"], yamlMeasures);
```
djbarnwal: Nit: check if arrays have any value before setting to avoid pollution of keys in YAML.

djbarnwal (approved):

proto/rill/runtime/v1/resources.proto
```diff
@@ -797,6 +798,33 @@ message CanvasPreset {
...
+
+message CanvasMeasureFilter {
+  string measure = 1;
+  repeated uint32 values = 2;
+  bool removable = 3;
+  string by_dimension = 4;
+  string operator = 5;
+  bool locked = 6;
+  bool hidden = 7;
+}
```
begelundmuller, line 823: I don't have the full context for this feature, so these may not be good suggestions, but what do you think about:
1. Using metrics SQL to parse expressions and propagating the raw expressions here
    - The backend can validate that each filter only addresses one dimension/measure
    - The frontend can use the metrics view to check if a name is a dimension or measure
2. Using a single `mode` field instead of separate properties for `removable`, `locked`, and `hidden`?

Then the protos could perhaps become:

```proto
message CanvasFilter {    
  Expression expression = 1; // Validation guarantees it only filters on one dimension or measure per filter
  CanvasFilterMode mode = 2;
}

enum CanvasFilterMode {
  CANVAS_FILTER_MODE_UNSPECIFIED = 0;
  CANVAS_FILTER_MODE_REMOVABLE = 1;
  CANVAS_FILTER_MODE_LOCKED = 2;
  CANVAS_FILTER_MODE_HIDDEN = 3;
}
```

The YAML could look something like this:

```yaml
type: canvas
...
defaults:
  filters:
    - expression: country NOT IN ('Denmark', 'United States')
      mode: locked
```

begelundmuller (changes requested):

proto/rill/runtime/v1/resources.proto
briangregoryholmes: I'm not against this. I would prefer to not have an escape hatch. This is how the URL captures a measure filter: `sites_domain having (ctr GT 24)`. Maybe this isn't valid Metrics SQL (which I think is what you're saying), but I don't see why we couldn't just co-opt that structure here.

begelundmuller: This PR adds support for measure filters in metrics SQL: https://github.com/rilldata/rill/pull/8010. With that, I don't think we need the escape hatch – updated the proposal above.

mindspank: `mode` would need to be an array as we support combination of values

briangregoryholmes: I do think they're mutually exclusive:

`hidden` is already `locked` and not `removable`
`locked` is already not `removable`

So, in my view, it doesn't make sense to ever pair two or more of them together.

mindspank: I would imagine that hidden could be overwritten by something like a bookmark or "clear" operations though

briangregoryholmes:
Closing in favor of #8391

PR #7782, opened 2025-08-15
fix: sorting issues in reports when not sorting by direct measure or dimension

AdityaHegde:
We were either not persisting or adding invalid field when sorting by a time dimension or comparison measure.

**Checklist:**
- [x] Covered by tests
- [ ] I'm proud of this work!


ericpgreen2 (approved):
Code looks reasonable. I'll also tag @mindspank in case he wants to do a QA pass.

PR #7787, opened 2025-08-18
API for inferring cloud projects for a local project

k-anshul:
INSERT DESCRIPTION HERE

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/cli/pkg/cmdutil/helper.go
+++ b/cli/pkg/cmdutil/helper.go
@@ -22,6 +22,7 @@ import (
 	adminv1 "github.com/rilldata/rill/proto/gen/rill/admin/v1"
 	runtimeclient "github.com/rilldata/rill/runtime/client"
 	"github.com/rilldata/rill/runtime/pkg/activity"
+	"github.com/rilldata/rill/runtime/pkg/fileutil"
 	"go.uber.org/zap"
 	"golang.org/x/sync/errgroup"
 	"google.golang.org/grpc/codes"
@@ -36,6 +37,8 @@ const (
 	telemetryIntakeUser     = "data-modeler"
 	telemetryIntakePassword = "lkh8T90ozWJP/KxWnQ81PexRzpdghPdzuB0ly2/86TeUU8q/bKiVug==" // nolint:gosec // secret is safe for public use
 )
+
+var ErrNoMatchingProject = fmt.Errorf("no matching project found")
 
 type Helper struct {
 	*printer.Printer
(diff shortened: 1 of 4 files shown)


admin/database/postgres/postgres.go
```diff
@@ -282,6 +282,42 @@ func (c *connection) FindProjectsForUser(ctx context.Context, userID string) ([]
...
+		SELECT p.* FROM projects p
+		WHERE p.id IN (
+			SELECT upr.project_id FROM users_projects_roles upr WHERE upr.user_id = $1
+			UNION
+			SELECT ugpr.project_id FROM usergroups_projects_roles ugpr JOIN usergroups_users ugu ON ugpr.usergroup_id = ugu.usergroup_id WHERE ugu.user_id = $1
+		)
+		AND (
+			(p.directory_name != '' AND p.directory_name = $2)
+			OR
+			(p.git_remote != '' AND p.git_remote = $3)
```
nishantmonu51: instead of looking for both directory_name and git_remote, we should aim to simplify and just look based on git remote.

k-anshul: Just `git_remote` is not sufficient for rill managed projects.
Currently there is a 1:1 relationship between rill managed repo and rill projects and that makes that code path somewhat simpler.
If we want to just use `git_remote` then we need to change that relationship to allow one rill managed repo to be associated with many rill projects.

nishantmonu51: In practice, when a project is linked to multiple rill managed repos, the directory name will be same as well.

-

k-anshul: Yeah that's where doing a check on both `directory_name` or `git_remote` will help. Doing a check just on `git_remote` will only list search for one project.

admin/database/postgres/postgres.go
```diff
@@ -282,6 +282,42 @@ func (c *connection) FindProjectsForUser(ctx context.Context, userID string) ([]
...
 }
 
+func (c *connection) FindProjectsForUserAndFingerprint(ctx context.Context, userID, directoryName, githubURL, afterID string, limit int) ([]*database.Project, error) {
+	// Shouldn't happen, but just to be safe and not return all projects.
+	if directoryName == "" && githubURL == "" {
+		return nil, nil
+	}
+
+	args := []any{userID, directoryName, githubURL}
+	qry := `
```
k-anshul, line 296: Okay I think this is problematic for monorepos. Due to the `git_remote` or `directory_name` condition it will return all projects matching the monorepo remote.

k-anshul: What do you think about following solution for finding matching projects ?
1. Query all rill managed projects using `directory = directory_name_arg or git_remote = git_remote_arg`.
2. Query all self managed projects using `remote, sub_path=directory_name_arg` or `remote, sub_path = NULL`. 

Alternatively we can still do the subpath filtering at client side but from an API POV doesn't seem right to return all projects deployed using monorepo as candidate projects. But I am open to this solution if this feels simpler.

k-anshul: Basically the idea is that for self managed repos (which can have monorepos) we remain strict on `remote, sub_path` and ignore `directory_name` or `remote, sub_path` acts as a fingerprint.

AdityaHegde: We could rename `directory` column as `fingerprint`. `remote, sub_path` can be added to it, with room of adding more.

k-anshul: Yes to simplify the query it can also be done. Basically during project creation/updation:
1. Set fingerprint to `directory_name` for rill managed repos.
2. Set fingerprint to `remote, subpath` for self managed repos.

During query:
1. Fetch managed using `fingerprint = directory_name or remote = remote_arg(for bwd compat when fingerpint is not set)`.
2. Fetch self managed using fingerprint = `remote, subpath`.

PR #7726, opened 2025-08-05
Create connector files

lovincyrus:
This PR changes the add data form behavior to create connector files instead of model files. There are also some form modifications to simplify the process of adding a new connector, e.g., removing `sql` and `name` when providing a connection string in `postgres`.

- [x] `type: connector`
- [x] [use ConfigProperties, nuke SourceProperties](https://linear.app/rilldata/issue/ENG-864/form-modifications-for-connector-files) except for the maybeRewriteToDuckdb connectors
- [x] clean up yupSchemas

Closes https://linear.app/rilldata/issue/APP-242/create-connector-files-not-model-files

**Checklist:**
- [ ] Covered by tests
- [x] Ran it and it works as intended
- [x] Reviewed the diff before requesting a review
- [ ] Checked for unhandled edge cases
- [x] Linked the issues it closes
- [x] Checked if the docs need to be updated. If so, create a separate Linear DOCS issue
- [ ] Intend to cherry-pick into the release branch
- [x] I'm proud of this work!

PR #7803, opened 2025-08-20
DuckDB Connectors Preview

lovincyrus:
INSERT DESCRIPTION HERE

CleanShot 2025-08-20 at 10 52 23@2x

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-common/src/features/sources/modal/AddDataForm.svelte
+++ b/web-common/src/features/sources/modal/AddDataForm.svelte
@@ -16,7 +16,11 @@
     type SuperValidated,
   } from "sveltekit-superforms";
   import { yup } from "sveltekit-superforms/adapters";
-  import { inferSourceName } from "../sourceUtils";
+  import {
+    inferSourceName,
+    maybeRewriteToDuckDb,
+    compileSourceYAML,
+  } from "../sourceUtils";
   import { humanReadableErrorMessage } from "../errors/errors";
   import {
     submitAddConnectorForm,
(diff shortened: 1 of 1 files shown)


web-common/src/features/sources/modal/AddDataForm.svelte
```diff
@@ -208,13 +205,46 @@
...
+            }
+          }
+        }
+      }
+
+      // Apply DuckDB rewrite logic for preview
+      const [rewrittenConnector, rewrittenFormValues] = maybeRewriteToDuckDb(
+        connector,
+        previewValues,
+      );
```
ericpgreen2: It's not good that this exists in two places (here and in `submitAddSourceForm`). It seems like the form data is processed differently for the preview than it is for the form submission.

Could we keep _one_ source-of-truth for the data processing logic?

ericpgreen2 (approved):
Approving with a couple comments.

(UXQA) The YAML preview for the DuckDB connector shows a `sql` key, but the form does not accept any `sql` and I don't see any `sql` in the generated file.
image

lovincyrus:
> (UXQA) The YAML preview for the DuckDB connector shows a `sql` key, but the form does not accept any `sql` and I don't see any `sql` in the generated file.
> image

These are the remnants of the issue reported [here](https://github.com/rilldata/rill/pull/7726#discussion_r2271449022).

web-common/src/features/sources/modal/AddDataForm.svelte
lovincyrus: https://github.com/rilldata/rill/pull/7803/commits/ee991868045fc32a1dbeab9c7153130260013225

ericpgreen2:
> These are the remnants of the issue reported [here](https://github.com/rilldata/rill/pull/7726#discussion_r2271449022).

Oh, in that PR you mentioned you'd revert to the Source properties. But it appears that the DuckDB form is currently generated via the Connector properties (which is a broken flow), not the Source ones (which is the working flow that's live  in `release-0.69`).

lovincyrus:
> > These are the remnants of the issue reported [here](https://github.com/rilldata/rill/pull/7726#discussion_r2271449022).
> 
> Oh, in that PR you mentioned you'd revert to the Source properties. But it appears that the DuckDB form is currently generated via the Connector properties (which is a broken flow), not the Source ones (which is the working flow that's live in `release-0.69`).

Pushed some fixes

PR #7806, opened 2025-08-20
Support parameters for Postgres

lovincyrus:
This PR adds the parameters form for the PostgreSQL connector. Closes https://linear.app/rilldata/issue/APP-19/postgres-connector-improvement

CleanShot 2025-08-20 at 16 05 52@2x

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!


runtime/drivers/postgres/postgres.go
```diff
@@ -28,15 +28,59 @@ var spec = drivers.Spec{
...
 			DocsURL:     "https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING",
 			Placeholder: "postgresql://postgres:postgres@localhost:5432/postgres",
-			Hint:        "Can be configured here or by setting the 'connector.postgres.database_url' environment variable (using '.env' or '--env')",
+			Hint:        "Can be configured here or by setting the 'connector.postgres.database_url' environment variable (using '.env' or '--env').",
 			Secret:      true,
 		},
+		{
+			Key:         "host",
+			Type:        drivers.StringPropertyType,
+			DisplayName: "Host/Socket",
```
ericpgreen2: Can we call this just "Host" to match the key?

runtime/drivers/postgres/postgres.go
```diff
@@ -28,15 +28,59 @@ var spec = drivers.Spec{
...
 		{
-			Key:         "database_url",
+			Key:         "dsn",
 			Type:        drivers.StringPropertyType,
 			DisplayName: "Postgres Connection String",
-			Required:    true,
 			DocsURL:     "https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING",
 			Placeholder: "postgresql://postgres:postgres@localhost:5432/postgres",
-			Hint:        "Can be configured here or by setting the 'connector.postgres.database_url' environment variable (using '.env' or '--env')",
+			Hint:        "Can be configured here or by setting the 'connector.postgres.database_url' environment variable (using '.env' or '--env').",
```
ericpgreen2: Should `database_url` be renamed to `dsn` here too?

runtime/drivers/postgres/postgres.go
```diff
@@ -28,15 +28,59 @@ var spec = drivers.Spec{
...
 			Secret:      true,
 		},
+		{
+			Key:         "host",
```
ericpgreen2, line 40: When I don't provide any keys, I see that "host", "user" and "dbname" are required. So seems like we should require them in the form too.
image

runtime/drivers/postgres/postgres.go
```diff
@@ -28,15 +28,59 @@ var spec = drivers.Spec{
...
+			Placeholder: "postgres",
+			Hint:        "Name of the Postgres database to connect to",
+		},
+		{
+			Key:         "sslmode",
+			Type:        drivers.StringPropertyType,
+			DisplayName: "SSL Mode",
+			Placeholder: "require",
+			Default:     "require",
+			Hint:        "SSL mode (disable, require, verify-ca, verify-full). Default is 'require' for security.",
```
ericpgreen2: > Default is 'require' for security

This confuses me because when no `sslmode` is provided, the implicit default value is NOT `require`. I believe the default is `disable`.

ericpgreen2: So, instead, I'd recommend:
- We remove the `default: "require"`
- The hint reads: "Options include `disable`, `require`, `verify-ca`, and `verify-full`"

Though, if these are truly the only options, another option is to use a select component with the fixed set of options.

ericpgreen2 (changes requested):

runtime/drivers/postgres/postgres.go
lovincyrus: https://github.com/rilldata/rill/pull/7806/commits/f9363710c0fd3911b3350963c4edef6c0c4b6f55

runtime/drivers/postgres/postgres.go
lovincyrus: https://github.com/rilldata/rill/pull/7806/commits/3509cbd9d999fa960de6c55007d8556756a655a5

ericpgreen2 (approved):
Looks good! To double-check one thing: @k-anshul would you mind reviewing whether or not we've marked the correct form fields "required"?

k-anshul:
Yeah I think the fields are correctly marked.

PR #7810, opened 2025-08-21
Support model execution retry configurations

grahamplata:
```yaml
# models/my_model.sql
retry:
  attempts: 5
  delay: 10s
  exponential_backoff: true
  if_error_matches:
    - ".*OvercommitTracker.*"
    - ".*Timeout.*"
    - ".*Bad Gateway.*"

# Or minimal config with defaults
retry:
  attempts: 3  # Uses default delay=5s, exponential_backoff=true
```

**Checklist:**
- [x] Covered by tests
- [x] [Linked the issues it closes](https://linear.app/rilldata/issue/PLAT-176/support-model-execution-retries)
- [x] Checked if the docs need to be updated. If so, [create a separate Linear DOCS issue](https://linear.app/rilldata/issue/DOC-32/model-execution-retry-configuration)
- [x] I'm proud of this work!


proto/rill/runtime/v1/resources.proto
```diff
@@ -121,6 +121,8 @@ message ModelSpec {
   google.protobuf.Struct stage_properties = 17;
   string output_connector = 1;
   google.protobuf.Struct output_properties = 12;
+  // retry is optional.
+  Retry retry = 26;
```
begelundmuller: Nesting is nice in the YAML, which is user facing, but in the internal protobuf definitions, we generally prefer flat objects unless the type is reused many places or needs to be `repeated`. (Main reasons being keeping the number of types in the very large `runtimev1` package manageable, as well as having less nil-handling bugs to worry about).

So for the proto (but not the YAML parser), I suggest flattening like this:
```
  uint32 retry_attempts = 26;
  uint32 retry_delay = 27;
  bool retry_exponential_backoff = 28;
  repeated string retry_if_error_matches = 29;
```

runtime/parser/parse_partial_retry.go
```diff
@@ -0,0 +1,75 @@
...
+type RetryYAML struct {
+	// Retry attempts
+	Attempts *uint32 `yaml:"attempts" mapstructure:"attempts"`
+	// Delay between retries
+	Delay *string `yaml:"delay"`
+	// Enable exponential backoff
+	ExponentialBackoff *bool `yaml:"exponential_backoff" mapstructure:"exponential_backoff"`
+	// Errors to match
+	IfErrorMatches []string `yaml:"if_error_matches" mapstructure:"if_error_matches"`
+}
```
begelundmuller: Since this is likely specific to models, I suggest inlining the type similar to e.g. `ModelYAML.Tests`

runtime/reconcilers/model.go
```diff
@@ -1256,42 +1311,55 @@ func (r *ModelReconciler) executeSingle(ctx context.Context, executor *wrappedMo
...
+					cleanupErr := executor.stageResultManager.Delete(ctx, stageResult)
+					if cleanupErr != nil {
+						r.C.Logger.Warn("Failed to clean up staged model output after final failure",
+							zap.String("model", self.Meta.Name.Name),
+							zap.Error(cleanupErr),
+							observability.ZapCtx(ctx))
+					}
+				}
+				return nil, err
+			}
```
begelundmuller: The error handling in these cases seems to get a bit hairy. Consider pushing the retry logic into a util function such that the existing execution logic remains unchanged. E.g.:
```go
func (r *ModelReconciler) executeSingle(...) {
  // setup
  ...

  return r.executeRetry(ctx, mdl, func() (*drivers.ModelResult, error) {
    // same execution logic as before here
  })
}
```

runtime/reconcilers/model.go
```diff
@@ -1256,42 +1311,55 @@ func (r *ModelReconciler) executeSingle(ctx context.Context, executor *wrappedMo
...
+						zap.Error(cleanupErr),
+						observability.ZapCtx(ctx))
+				}
+			}
 
-		// Drop the stage result after the final step has executed.
-		// We do this using the same ctx, which means we may leak data in the staging connector in case of context cancellations.
-		// This is acceptable since the staging connector is assumed to be configured for temporary data.
-		defer func() {
+			time.Sleep(r.calculateRetryDelay(attempt, mdl.Spec.Retry))
```
begelundmuller: This will not be responsive to `ctx` cancellations – I think you need a `select` statement with `time.After` for the sleep

runtime/parser/parse_model.go
```diff
@@ -227,6 +234,42 @@ func (p *Parser) parseModel(ctx context.Context, node *Node) error {
...
+		r.ModelSpec.RetryAttempts = *tmp.Retry.Attempts
+	} else {
+		r.ModelSpec.RetryAttempts = 3
+	}
+
+	if tmp.Retry.Delay != nil {
+		duration, err := time.ParseDuration(*tmp.Retry.Delay)
+		if err != nil {
+			return fmt.Errorf(`invalid retry delay: %w`, err)
+		}
```
begelundmuller: Note line 231:
```
```

runtime/parser/parse_model_test.go
```diff
@@ -51,35 +51,47 @@ output:
...
-				ChangeMode:      runtimev1.ModelChangeMode_MODEL_CHANGE_MODE_RESET,
+				RefreshSchedule:         &runtimev1.Schedule{RefUpdate: true},
+				InputConnector:          "duckdb",
+				InputProperties:         must(structpb.NewStruct(map[string]any{"sql": "SELECT 1"})),
+				OutputConnector:         "duckdb",
+				ChangeMode:              runtimev1.ModelChangeMode_MODEL_CHANGE_MODE_RESET,
+				RetryAttempts:           3,
+				RetryDelay:              5,
+				RetryExponentialBackoff: true,
+				RetryIfErrorMatches:     []string{".*OvercommitTracker.*", ".*Bad Gateway.*", ".*Timeout.*"},
```
begelundmuller: I'm not sure this is a consistent opinion, but it feels annoying that these defaults will get persisted for each resource. I can also imagine wanting to have OLAP-specific defaults at some point.

What do you think about implementing the defaults at execution time inside the model reconciler instead? Similar to how many other default settings are only applied at execution time. In practice, I think you could achieve this by making the retry proto fields `optional`, leaving them empty when not configured in the YAML, and then in the model reconciler use the defaults if they are `nil` (unset) in the proto.

runtime/reconcilers/model.go
```diff
@@ -1256,41 +1301,34 @@ func (r *ModelReconciler) executeSingle(ctx context.Context, executor *wrappedMo
...
+				zap.Int("attempt", attempt),
+				zap.String("error", err.Error()),
+				zap.Strings("retry_patterns", mdl.Spec.RetryIfErrorMatches),
+				zap.Int("retry_action", int(action)),
+				observability.ZapCtx(ctx))
+
+			if action == retrier.Retry {
+				// Calculate the actual backoff duration for this attempt
+				actualBackoff := backoff
+				if mdl.Spec.RetryExponentialBackoff {
```
begelundmuller: I'm still confused by the changes to the actual execution implementation. I don't think the `retrier` package is helping here, as it's still mixing retry logic into the execution logic. It also seems like some unrelated changes have happened, like not logging the stage model output clean up error anymore?

Given the actual retry loop is going to be very simple, I think just wrapping the existing logic with a model-specific retry util function should be straightforward. But maybe I'm missing something. What did you think about my suggestion here? https://github.com/rilldata/rill/pull/7810#discussion_r2307434579

begelundmuller (changes requested):

runtime/controller_test.go
```diff
@@ -1218,12 +1217,12 @@ measures:
...
 }
 
-func newSource(name, path, localFileHash string) (*runtimev1.Model, *runtimev1.Resource) {
+func newSource(t *testing.T, rt *runtime.Runtime, id string, name, path, localFileHash string) (*runtimev1.Model, *runtimev1.Resource) {
 	source := &runtimev1.Model{
 		Spec: &runtimev1.ModelSpec{
 			InputConnector:   "local_file",
 			OutputConnector:  "duckdb",
-			InputProperties:  must(structpb.NewStruct(map[string]any{"path": path, "local_files_hash": localFileHash})),
+			InputProperties:  must(structpb.NewStruct(map[string]any{"path": "data/foo.csv", "local_files_hash": localFileHash})),
```
begelundmuller: What is the point of any of the changes in this file?

runtime/parser/parse_model.go
```diff
@@ -216,6 +223,38 @@ func (p *Parser) parseModel(ctx context.Context, node *Node) error {
 		node.Refs = append(node.Refs, refs...)
 	}
 
+	// Parse retry options with reasonable defaults
+	var retryAttempts *uint32
+	if tmp.Retry.Attempts != nil {
+		retryAttempts = tmp.Retry.Attempts
+	}
```
begelundmuller: This doesn't really do anything – just use `tmp.Retry.Attempts` directly

proto/rill/runtime/v1/resources.proto
```diff
@@ -121,6 +121,11 @@ message ModelSpec {
   google.protobuf.Struct stage_properties = 17;
   string output_connector = 1;
   google.protobuf.Struct output_properties = 12;
+  // retry is optional.
+  optional uint32 retry_attempts = 26;
+  optional uint32 retry_delay = 27;
```
begelundmuller: I suggest calling this `retry_delay_seconds` (not in the YAML, but in the internal proto)

proto/rill/runtime/v1/resources.proto
```diff
@@ -121,6 +121,11 @@ message ModelSpec {
   google.protobuf.Struct stage_properties = 17;
   string output_connector = 1;
   google.protobuf.Struct output_properties = 12;
+  // retry is optional.
+  optional uint32 retry_attempts = 26;
```
begelundmuller: The comment doesn't really add anything here given there's a keyword saying `optional`

runtime/parser/parse_model.go
```diff
@@ -216,6 +223,38 @@ func (p *Parser) parseModel(ctx context.Context, node *Node) error {
...
+			return fmt.Errorf(`invalid retry delay: %w`, err)
+		}
+		delay := uint32(duration.Seconds())
+		retryDelay = &delay
+	}
+
+	var retryExponentialBackoff *bool
+	if tmp.Retry.ExponentialBackoff != nil {
+		retryExponentialBackoff = tmp.Retry.ExponentialBackoff
+	}
```
begelundmuller: Same comment

runtime/parser/parse_model.go
```diff
@@ -216,6 +223,38 @@ func (p *Parser) parseModel(ctx context.Context, node *Node) error {
...
+
+	var retryIfErrorMatches []string
+	if len(tmp.Retry.IfErrorMatches) > 0 {
+		// Validate regex patterns
+		for _, pattern := range tmp.Retry.IfErrorMatches {
+			if _, err := regexp.Compile(pattern); err != nil {
+				return fmt.Errorf("invalid regex pattern '%s': %w", pattern, err)
+			}
+		}
+		retryIfErrorMatches = tmp.Retry.IfErrorMatches
```
begelundmuller: Same comment

runtime/parser/parser_test.go
```diff
@@ -354,11 +354,9 @@ schema: default
...
 			Paths: []string{"/custom/c2.sql"},
 			ModelSpec: &runtimev1.ModelSpec{
-				RefreshSchedule: &runtimev1.Schedule{RefUpdate: true},
-				InputConnector:  "duckdb",
-				InputProperties: must(structpb.NewStruct(map[string]any{
-					"sql": strings.TrimSpace(files["custom/c2.sql"]),
-				})),
+				RefreshSchedule:  &runtimev1.Schedule{RefUpdate: true},
+				InputConnector:   "duckdb",
+				InputProperties:  must(structpb.NewStruct(map[string]any{"sql": strings.TrimSpace(files["custom/c2.sql"])})),
```
begelundmuller: Don't really see the point of any of the changes in this file either

runtime/reconcilers/model.go
```diff
@@ -1233,20 +1234,48 @@ func (r *ModelReconciler) executeSingle(ctx context.Context, executor *wrappedMo
...
 
-	// Execute the stage step if configured
-	var stageDuration time.Duration
-	if executor.stage != nil {
-		// Also resolve templating in the stage props
-		stageProps, err := r.resolveTemplatedProps(ctx, self, incrementalState, partitionData, mdl.Spec.StageConnector, mdl.Spec.StageProperties.AsMap())
-		if err != nil {
-			return nil, err
+	return r.executeWithRetry(ctx, self, mdl, func(ctx context.Context) (*drivers.ModelResult, error) {
+		var finalResult *drivers.ModelResult
```
begelundmuller: There doesn't seem to be any reason to declare this separately, can just use `:=` when it's needed like the old code. 

I also notice many comments are slightly changed. Is it AI edits? It's kind of annoying to review with all these small edits in code that the PR doesn't really need to change.

runtime/reconcilers/model.go
```diff
@@ -1258,40 +1287,97 @@ func (r *ModelReconciler) executeSingle(ctx context.Context, executor *wrappedMo
...
 			return nil, err
 		}
-		stageDuration = stageResult.ExecDuration
+		finalResult.ExecDuration += stageDuration
+
+		return finalResult, nil
+	})
+}
 
-		// We change the inputProps to be the result properties of the stage step
```
begelundmuller, line 1263: E.g. this comment seems to me like it added some useful context to explain what's happening

runtime/reconcilers/model.go
```diff
@@ -1258,40 +1287,97 @@ func (r *ModelReconciler) executeSingle(ctx context.Context, executor *wrappedMo
...
+	attempts := int(*retryAttempts)
+	if attempts == 0 {
+		attempts = 1
+	}
+
+	// maxRetries is not used, so remove its assignment
+	backoff := time.Duration(*retryDelay) * time.Second
+	if backoff == 0 {
+		backoff = 5 * time.Second
+	}
```
begelundmuller: This default is already applied above. This case only runs if the user explicitly set it to 0 – if they did that, isn't the intention just to keep it 0, i.e. retry immediately?

runtime/controller_test.go
grahamplata: It was initially bawking at a previous change when some fields were not optional. I can drop the changes

runtime/reconcilers/model.go
grahamplata: Agreed, I will be mindful of that.

PR #7811, opened 2025-08-21
Fix - Managed mode should take priority over other connector props

grahamplata:
INSERT DESCRIPTION HERE

**Checklist:**
- [x] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/runtime/drivers/clickhouse/clickhouse.go
+++ b/runtime/drivers/clickhouse/clickhouse.go
@@ -195,47 +195,18 @@ type configProperties struct {
 }
 
 func (c *configProperties) validate() error {
-	var set []string
-	if c.Host != "" {
-		set = append(set, "host")
-	}
-	if c.Port != 0 {
-		set = append(set, "port")
-	}
-	if c.Username != "" {
-		set = append(set, "username")
-	}
-	if c.Password != "" {
-		set = append(set, "password")
-	}
-	if c.Database != "" {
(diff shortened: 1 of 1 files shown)


mindspank (approved):

runtime/drivers/clickhouse/clickhouse.go
```diff
@@ -195,39 +195,20 @@ type configProperties struct {
...
-	}
 	if c.Managed {
-		if c.DSN != "" {
-			set = append(set, "dsn")
-		}
-		if len(set) > 0 {
-			return fmt.Errorf("managed ClickHouse does not support setting [%s] properties", strings.Join(set, ", "))
-		}
+		// In managed mode, clear all connection-related properties
+		c.DSN = ""
```
begelundmuller, line 200: This breaks the managed Clickhouse support. When setting `managed: true`, the provisioner provides the DSN. This code clears that DSN making it not able to connect. 

Relevant code: 
- Properties provided by the provisioner are added to the connector here: https://github.com/rilldata/rill/blob/8bec21fa48607085fae370bc2e059e9871558e2e/runtime/connection_cache.go#L137-L145
- Config keys provided by the Clickhouse provisioner: https://github.com/rilldata/rill/blob/8bec21fa48607085fae370bc2e059e9871558e2e/admin/provisioner/resources.go#L80-L84

PR #7813, opened 2025-08-21
Move MotherDuck from sources to OLAP engines and add schema_name config

lovincyrus:
This PR moves MotherDuck from the sources section to the OLAP engine section. In addition to the display changes, we've also added `schema_name` as a required configuration property for MotherDuck connections.

Fixes https://linear.app/rilldata/issue/APP-261/move-motherduck-from-the-source-section-to-the-connect-an-olap-engine & https://linear.app/rilldata/issue/APP-304/add-schema-name-input-field-to-motherduck

CleanShot 2025-08-25 at 15 20 50@2x

https://github.com/user-attachments/assets/b557ae41-7896-40f0-b39f-c25b453963a1

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!

### DIFF
--- a/runtime/drivers/duckdb/duckdb.go
+++ b/runtime/drivers/duckdb/duckdb.go
@@ -41,10 +41,10 @@ var spec = drivers.Spec{
 	DocsURL:     "https://docs.rilldata.com/connect/olap/duckdb",
 	SourceProperties: []*drivers.PropertySpec{
 		{
-			Key:         "path",
+			Key:         "db",
 			Type:        drivers.StringPropertyType,
 			Required:    true,
-			DisplayName: "Path",
+			DisplayName: "DB",
 			Description: "Path to DuckDB database",
 			Placeholder: "/path/to/duckdb.db",
 		},
--- a/web-common/src/features/sources/modal/yupSchemas.ts
+++ b/web-common/src/features/sources/modal/yupSchemas.ts
@@ -36,8 +36,8 @@ export const getYupSchema = {
   }),
 
   duckdb: yup.object().shape({
-    path: yup.string().required("Path is required"),
-    sql: yup.string().required("SQL is required"),
+    db: yup.string().required("db is required"),
+    sql: yup.string().required("sql is required"),
     name: yup
       .string()
       .matches(VALID_NAME_PATTERN, INVALID_NAME_MESSAGE)


lovincyrus:
Latest recording with `schema_name` addition:
 
https://github.com/user-attachments/assets/b41d2991-8a7a-46c3-a5af-42444bfb7798

runtime/drivers/duckdb/duckdb.go
```diff
@@ -84,50 +84,32 @@ var motherduckSpec = drivers.Spec{
...
+			DisplayName: "MotherDuck Connection String",
+			Placeholder: "md:my_db",
+		},
+		{
+			Key:         "schema_name",
+			Type:        drivers.StringPropertyType,
+			Required:    true,
+			DisplayName: "Schema name",
+			Placeholder: "your_schema_name",
+			Hint:        "Set the default schema used by the DuckDB database",
```
lovincyrus: Thoughts on the hint? @ericokuma

ericokuma:
@mindspank, does this hint for Schema Name make sense?

mindspank:
@ericokuma schema name makes sense but perhaps we should say MotherDuck instead of DuckDb?

ericpgreen2 (dismissed):
I've tried this flow with a purposefully incorrect access token:
1. I've entered a dummy access token "abc"
2. When I click "Connect", the MotherDuck OAuth page pops up. I authenticate.
3. The UI's "test connection" flow completes successfully, the file gets created.
4. When the file gets created, the MotherDuck OAuth page gets triggered a second time. I authenticate again.
5. I see my dummy access token in my `.env` file.

A few issues:
1. If we're supporting the MotherDuck OAuth flow, then the form shouldn't require that an `access_token` is provided. 
2. When using the MotherDuck OAuth flow, I don't see the resultant token persisted anywhere in my code files (e.g. in my `motherduck.yaml` or `.env`), so it feels like a Deploy-to-Rill-Cloud will run into authentication issues.
3. When using the MotherDuck OAuth flow, the OAuth flow should not be triggered twice, only once.
4. If an access token is provided, then we shouldn't trigger the OAuth flow.

@lovincyrus, I suggest you reach out to the Platform team to understand the OAuth behavior. (I don't know how it gets triggered myself.)

CC @ericokuma @k-anshul for awareness

lovincyrus:
> I've tried this flow with a purposefully incorrect access token:
> 
> 1. I've entered a dummy access token "abc"
> 2. When I click "Connect", the MotherDuck OAuth page pops up. I authenticate.
> 3. The UI's "test connection" flow completes successfully, the file gets created.
> 4. When the file gets created, the MotherDuck OAuth page gets triggered a second time. I authenticate again.
> 5. I see my dummy access token in my `.env` file.
> 
> A few issues:
> 
> 1. If we're supporting the MotherDuck OAuth flow, then the form shouldn't require that an `access_token` is provided.
> 2. When using the MotherDuck OAuth flow, I don't see the resultant token persisted anywhere in my code files (e.g. in my `motherduck.yaml` or `.env`), so it feels like a Deploy-to-Rill-Cloud will run into authentication issues.
> 3. When using the MotherDuck OAuth flow, the OAuth flow should not be triggered twice, only once.
> 4. If an access token is provided, then we shouldn't trigger the OAuth flow.
> 
> @lovincyrus, I suggest you reach out to the Platform team to understand the OAuth behavior. (I don't know how it gets triggered myself.)
> 
> CC @ericokuma @k-anshul for awareness

Let's bring the discussion to https://linear.app/rilldata/issue/APP-304/add-schema-name-input-field-to-motherduck#comment-cce82f43

lovincyrus:
> @ericokuma schema name makes sense but perhaps we should say MotherDuck instead of DuckDb?

Makes sense. Addressed!

lovincyrus:
Update: to land this PR after https://github.com/rilldata/rill/pull/7842

lovincyrus:
> Update: to land this PR after #7842

We no longer see the motherduck oauth page after [this change](https://github.com/rilldata/rill/pull/7842)! Ready for 🚀

runtime/drivers/duckdb/duckdb.go
```diff
@@ -66,6 +66,7 @@ var spec = drivers.Spec{
 			DisplayName: "SQL",
 			Description: "Query to extract data from DuckDB.",
 			Placeholder: "select * from table;",
+			NoPrompt:    true,
```
ericpgreen2: Why this change? The user needs to provide `sql` for DuckDB-as-a-source (but not for MotherDuck-as-an-OLAP-connector).

Here's what I see for the DuckDB form:
image

Here's what we have on the latest release and what I'd expect:
image

lovincyrus: Addressed the configProperties in duckdb spec

lovincyrus:
> When I submit the DuckDB form, it creates a Connector file, but on `v70` it creates a Model file. Can you please test the DuckDB form to make sure it works on an actual DuckDB database file?

ah, duckdb spec was incorrectly marked as `ImplementsOLAP`. The connector belongs to the SOURCES. We're creating connector files based on the spec type. Updated.

Update 1: Hmm, it seems that DuckDB is an OLAP mentioned in the sources section, but it is presented differently in the documentation ﻿https://rilldata.slack.com/archives/C07R1EP2CCV/p1757538295573349

Update 2: Re-adding `ImplementsOLAP` to the duckdb spec, but preventing duckdb from creating connector file.

lovincyrus:
Marked as "draft" because I noticed the SQL property appearing in MotherDuck's form. I also observed that the "How to connect to DuckDB?" section contains a hyperlink leading to the MotherDuck doc.

lovincyrus:
> Marked as "draft" because I noticed the SQL property appearing in MotherDuck's form. I also observed that the "How to connect to DuckDB?" section contains a hyperlink leading to the MotherDuck doc.

DuckDB

CleanShot 2025-09-11 at 12 35 30@2x

CleanShot 2025-09-11 at 12 35 41@2x

MotherDuck

CleanShot 2025-09-11 at 12 38 12@2x

ericpgreen2:
I'm having trouble getting a DuckDB source to work, and, per your screenshot, it looks like you are too. Would you be able to setup a working example & track down any issues (i.e. coordinate w/ Platform), if any?

Update: I went back to version `v0.70.0` and see that the correct parameter seems to be `db` not `path`:
image

(And we appear to have had a regression in `v0.71.0` – the DuckDB-as-a-source flow is currently broken.)

lovincyrus:
> I'm having trouble getting a DuckDB source to work, and, per your screenshot, it looks like you are too. Would you be able to setup a working example & track down any issues (i.e. coordinate w/ Platform), if any?
> 
> Update: I went back to version `v0.70.0` and see that the correct parameter seems to be `db` not `path`: image
> 
> (And we appear to have had a regression in `v0.71.0` – the DuckDB-as-a-source flow is currently broken.)

The root cause is https://github.com/rilldata/rill/pull/7726/files#diff-ee46cf7550ad45c31d45ff651b46d525a8685389eb1d823d879c7ec88806bef9R39

lovincyrus:
Latest:

CleanShot 2025-09-11 at 14 34 06@2x

lovincyrus:
Reverted so DuckDB now creates a model file with `db` key

PR #7815, opened 2025-08-22
use transitive access policy for resources

pjain1:
INSERT DESCRIPTION HERE

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


admin/auth_token.go
```diff
@@ -175,20 +175,27 @@ func (t *magicAuthToken) OwnerID() string {
...
+	TTL                      *time.Duration
+	CreatedByUserID          *string
+	Attributes               map[string]any
+	FilterJSON               string
+	Fields                   []string
+	State                    string
+	DisplayName              string
+	Internal                 bool
+	Resources                []database.ResourceName
+	TransitiveAccessResource *database.ResourceName // Optional, used for transitive access
```
begelundmuller: Could we grant transitive access to all resources in `Resources`, so we don't need a separate field?

admin/auth_token.go
```diff
@@ -175,20 +175,27 @@ func (t *magicAuthToken) OwnerID() string {
...
 }
 
 // IssueMagicAuthToken generates and persists a new magic auth token for a project.
 func (s *Service) IssueMagicAuthToken(ctx context.Context, opts *IssueMagicAuthTokenOptions) (AuthToken, error) {
+	if opts.TransitiveAccessResource != nil {
+		if len(opts.Resources) > 0 || opts.FilterJSON != "" || len(opts.Fields) > 0 {
```
begelundmuller: I think we still need filter JSON support for explores/canvases (although not reports since the filter is available in the report spec)? E.g. if you create a public URL for a canvas with filters on it, then we need to capture the filters and push them down to the transitive access to the underlying metrics views.

admin/server/projects.go
```diff
@@ -247,94 +247,112 @@ func (s *Server) GetProject(ctx context.Context, req *adminv1.GetProjectRequest)
...
-		// All themes
-		condition.WriteString(fmt.Sprintf("'{{.self.kind}}'='%s'", runtime.ResourceKindTheme))
-
 		var c *client.Client
+		if mdl.TransitiveAccessResource != nil {
+			c, err = s.admin.OpenRuntimeClient(depl)
+			if err != nil {
+				return nil, status.Errorf(codes.Internal, "could not open runtime client: %s", err.Error())
+			}
+			defer c.Close() // nolint:gocritic // client is created only once
```
begelundmuller: The client appears unused in this case?

admin/server/projects.go
```diff
@@ -247,94 +247,112 @@ func (s *Server) GetProject(ctx context.Context, req *adminv1.GetProjectRequest)
...
+							return nil, status.Errorf(codes.NotFound, "resource for magic token not found (name=%q, type=%q)", r.Name, r.Type)
+						}
+						return nil, fmt.Errorf("could not get resource for magic token: %w", err)
 					}
-					defer c.Close() // nolint:gocritic // client is created only once
+
+					spec := resp.Resource.GetExplore().State.ValidSpec
+					if spec != nil {
+						condition.WriteString(fmt.Sprintf(" OR ('{{.self.kind}}'='%s' AND '{{lower .self.name}}'=%s)", runtime.ResourceKindMetricsView, duckdbsql.EscapeStringValue(strings.ToLower(spec.MetricsView))))
+					}
```
begelundmuller: This handling would not be needed anymore if we always give transitive access to all the `Resources`, right? I.e. old public URLs and report tokens would just get the metrics access through the transitive rule instead, right?

proto/rill/runtime/v1/resources.proto
```diff
@@ -341,6 +342,10 @@ message SecurityRuleRowFilter {
   Expression expression = 3; // Regular query expression referencing dimension names
 }
 
+message SecurityRuleTransitiveAccess {
+  ResourceName resource = 1; // this rules resolves to rules that provides access to whatever is needed to access this resource. Currently only used for reports and alerts.
```
begelundmuller: Would be nice also to support explores initially. (Explores are quite simply and would ensure backwards compatibility with old public URLs, but canvases we can do in a separate PR as they don't currently have public URLs.)

runtime/resolver.go
```diff
@@ -46,6 +46,10 @@ type Resolver interface {
 	ResolveInteractive(ctx context.Context) (ResolverResult, error)
 	// ResolveExport resolve data for export (e.g. downloads or reports).
 	ResolveExport(ctx context.Context, w io.Writer, opts *ResolverExportOptions) error
+	// MetricsViewSecurityFields returns the list of mv fields that will be used in the resolver, these can be used by the security engine to enforce field-level security.
+	MetricsViewSecurityFields() []string
+	// SecuredRowFilter returns the row filter for the resolver, which can be used by the security engine to enforce row-level security.
+	SecuredRowFilter() string
```
begelundmuller: It feels weird to have this logic on the `Resolver` interface since it's not metrics-view specific. Could it be implemented directly in the security engine itself? It currently has different rules for different resource types.

If we wanted to move the security handling of individual resource types out of the security engine, I think the reconcilers might be a more generic way to do it (since there is one reconciler per resource type). But I think it's okay to keep individual resource handling inside the security engine at the moment.

begelundmuller: If the problem is circular dependencies, consider if e.g. a `metricsviewutil` package might help instead for util functions for metrics view parsing.

runtime/resolvers/metrics.go
```diff
@@ -125,6 +125,38 @@ func (r *metricsResolver) Validate(ctx context.Context) error {
...
+	return fields
+}
+
+func (r *metricsResolver) SecuredRowFilter() string {
+	// Convert the where expression to SQL if present
+	if r.query.Where != nil {
+		// For now, return empty string as converting Expression to SQL is complex
+		// This can be enhanced later to convert the expression to SQL
+		return ""
+	}
```
begelundmuller: There is actually support for passing an `Expression` to the security engine instead of a raw SQL filter: https://github.com/rilldata/rill/blob/b1eac643ac790c36c5f4ba1dfe82994bc314257d/proto/rill/runtime/v1/resources.proto#L341

and: https://github.com/rilldata/rill/blob/b1eac643ac790c36c5f4ba1dfe82994bc314257d/runtime/security.go#L600-L607

admin/auth_token.go
pjain1: I kept it so that public urls can still be created in same way as earlier. But looks like we want to use transitive access even for public urls? 

I thought we would use transitive access for resources like reports, alerts and canvases which are built on top of basic resources like metrics view. For public url, transitive access resource would be explore? and thus it will resolve to mv access and filters/fields will be added to magic token.

We want transitive access for all resource types in Rill then from mv to canvases and anything in future?

admin/auth_token.go
pjain1: see above [comment](https://github.com/rilldata/rill/pull/7815#discussion_r2298609478)

admin/server/projects.go
pjain1: yes forgot to remove this

admin/server/projects.go
pjain1: This is old code, just moved around and yes won't be needed when transitive access is added for the explore. However, we will still need for old magic tokens to keep backwards compatibility?

runtime/resolver.go
pjain1: I also felt this is a little odd. 

The problem is not circular dependency, problem is that for resources like reports/alerts we need to expand transitive access rule to add mv access, field access and row filter rules, however, this information is present in the query args json. Since resolvers (legacy resolver in case of report) already parses properties to create appropriate query and thus has this information, I thought it would be convenient to expose it here otherwise for example for reports will need to parse the properties and use `case` statement to create appropriate query to get this information which is already done while initializing the resolver.

runtime/resolvers/metrics.go
pjain1: Sure focus was on legacy resolver, this is just a miss and stub is copilot generated :)

proto/rill/runtime/v1/resources.proto
pjain1: sure

admin/auth_token.go
begelundmuller: Yes I was thinking we would also use this for public URLs on both canvas and explores. I thought it would simplify both implementations (no need to know about underlying metrics views on the control plane side). Do you disagree? Maybe I'm missing something.

admin/server/projects.go
begelundmuller: > we will still need for old magic tokens to keep backwards compatibility?

You will have to check, I don't remember, but I think the old magic tokens have the explore name in the `resources` already, so transitive access on that would make them continue working without resolving the metrics view here in the control plane?

runtime/resolver.go
begelundmuller: I see. Could it be made non-metrics-specific with a generic function like:
```go
// InferRequiredSecurityRules attempts to infer the security rules that are required to be able to execute the resolver for the currently configured properties.
InferRequiredSecurityRules() []*runtimev1.SecurityRule
```
Otherwise I'm also okay with hard-coding the cases in the security engine itself since it's specific to metrics queries for alerts/reports only. Up to you.

admin/database/database.go
```diff
@@ -197,6 +197,11 @@ type DB interface {
 	FindReportTokenForMagicAuthToken(ctx context.Context, magicAuthTokenID string) (*ReportToken, error)
 	InsertReportToken(ctx context.Context, opts *InsertReportTokenOptions) (*ReportToken, error)
 
+	FindAlertTokens(ctx context.Context, alertName string) ([]*AlertToken, error)
+	FindAlertTokensWithSecret(ctx context.Context, alertName string) ([]*AlertTokenWithSecret, error)
+	FindAlertTokenForMagicAuthToken(ctx context.Context, magicAuthTokenID string) (*AlertToken, error)
+	InsertAlertToken(ctx context.Context, opts *InsertAlertTokenOptions) (*AlertToken, error)
```
begelundmuller: Is there a difference between the alert and report tokens or could it make sense to merge them into one concept, e.g. a "notification token"?
```go
FindNotificationTokens(ctx context.Context, resourceKind, resourceName string) ([]*NotificationToken, error)
FindNotificationTokensWithSecret(ctx context.Context, resourceKind, resourceName string) ([]*NotificationTokenWithSecret, error)
FindNotificationTokenForMagicAuthToken(ctx context.Context, magicAuthTokenID string) (*NotificationToken, error)
InsertNotificationToken(ctx context.Context, opts *InsertNotificationTokenOptions) (*NotificationToken, error)
```

admin/database/postgres/postgres.go
```diff
@@ -3101,9 +3149,10 @@ func (c *connection) magicAuthTokenFromDTO(dto *magicAuthTokenDTO, fetchSecret b
 // magicAuthTokenWithUserDTO wraps database.MagicAuthTokenWithUser, using the pgtype package to handly types that pgx can't read directly into their native Go types.
 type magicAuthTokenWithUserDTO struct {
 	*database.MagicAuthTokenWithUser
-	Attributes pgtype.JSON      `db:"attributes"`
-	Fields     pgtype.TextArray `db:"fields"`
-	Resources  pgtype.JSONB     `db:"resources"`
+	Attributes               pgtype.JSON      `db:"attributes"`
+	Fields                   pgtype.TextArray `db:"fields"`
+	Resources                pgtype.JSONB     `db:"resources"`
+	TransitiveAccessResource *pgtype.JSONB    `db:"transitive_access_resource"`
```
begelundmuller: Should this field be removed?

admin/server/alerts.go
```diff
@@ -73,9 +76,54 @@ func (s *Server) GetAlertMeta(ctx context.Context, req *adminv1.GetAlertMetaRequ
...
+			if err == nil {
+				ownerEmail = owner.Email
+			}
+		}
+
+		// Create magic tokens for all recipients
+		emailTokens, err := s.createMagicTokensAlert(ctx, proj.ID, req.Alert, req.OwnerId, recipients, attr)
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "failed to issue magic auth tokens: %s", err.Error())
+		}
```
begelundmuller, line 103: So it creates a new token for each report run? I actually thought for reports it re-used existing tokens with `FindReportTokensWithSecret`, but seems that's not the case. 

What do you think here? On the one hand, it's simpler to create new tokens every time, but can also end up with rather many tokens over time, even with expiration. But perhaps it's not really that bad. 

If you stick with creating every time, I guess we don't need `FindAlertTokens` and `FindAlertTokensWithSecret` (and similar for reports)?

admin/urls.go
```diff
@@ -328,15 +328,27 @@ func (u *URLs) ReportEdit(org, project, report string) string {
 }
 
 // AlertOpen returns the URL for opening an alert in the frontend.
-func (u *URLs) AlertOpen(org, project, alert string) string {
-	return urlutil.MustJoinURL(u.Frontend(), org, project, "-", "alerts", alert, "open")
+func (u *URLs) AlertOpen(org, project, alert, token string) string {
+	openURL := urlutil.MustJoinURL(u.Frontend(), org, project, "-", "alerts", alert, "open")
+	if token != "" {
+		openURL += fmt.Sprintf("?token=%s", token)
```
begelundmuller: Use `MustWithQuery` like the other utils in the file?

proto/rill/runtime/v1/resources.proto
begelundmuller: Comment outdated? Now it's also used for public URLs for explores, right?

runtime/reconcilers/alert.go
```diff
@@ -724,23 +737,33 @@ func (r *AlertReconciler) popCurrentExecution(ctx context.Context, self *runtime
...
+								return fmt.Errorf("failed to build recipient open url: %w", err)
+							}
+							msg.OpenLink = openLink
+							msg.EditLink = recipientURLs.EditURL
+							msg.UnsubscribeLink = recipientURLs.UnsubscribeURL
+						} else {
+							// Fallback to default URLs (for internal users)
+							msg.OpenLink = ""
+							msg.EditLink = ""
+							msg.UnsubscribeLink = ""
```
begelundmuller, line 905: "for internal users" – what does that mean? Before there was a comment about outside of cloud:
```
// Note: adminMeta may not always be available (if outside of cloud). In those cases, we leave the links blank (no clickthrough available).
```

runtime/resolvers/annotations.go
```diff
@@ -135,3 +135,7 @@ func (r *annotationsResolver) ResolveInteractive(ctx context.Context) (runtime.R
 func (r *annotationsResolver) ResolveExport(ctx context.Context, w io.Writer, opts *runtime.ResolverExportOptions) error {
 	return errors.New("not implemented")
 }
+
+func (r *annotationsResolver) InferRequiredSecurityRules() []*runtimev1.SecurityRule {
+	panic("not implemented")
```
begelundmuller: Maybe be safe and just return `nil`? The docstring says "attempts" so means it's not guaranteed.

runtime/resolvers/glob.go
```diff
@@ -215,6 +215,10 @@ func (r *globResolver) ResolveExport(ctx context.Context, w io.Writer, opts *run
 	return errors.New("not implemented")
 }
 
+func (r *globResolver) InferRequiredSecurityRules() []*runtimev1.SecurityRule {
+	panic("not implemented")
```
begelundmuller: Same comment here, applies a few places

runtime/resolvers/legacy_metrics.go
```diff
@@ -75,12 +82,20 @@ func newLegacyMetrics(ctx context.Context, opts *runtime.ResolverOptions) (runti
 		return nil, fmt.Errorf("failed to build query: %w", err)
 	}
 
+	// Extract fields and row filter from the query using the queries.SecurityFromQuery helper
+	rowFilter, fields, err := queries.SecurityFromQuery(props.QueryName, props.QueryArgsJSON)
+	if err != nil {
+		return nil, fmt.Errorf("failed to extract accessible fields: %w", err)
+	}
```
begelundmuller: Consider moving to `InferRequiredSecurityRules` since it's only used there (then it doesn't need new struct vars). Would require support for returning an `error` from `InferRequiredSecurityRules` – would that be a problem?

runtime/resolvers/sql.go
```diff
@@ -172,6 +172,24 @@ func (r *sqlResolver) ResolveExport(ctx context.Context, w io.Writer, opts *runt
...
+					Condition: fmt.Sprintf("'{{.self.kind}}'='%s' AND '{{lower .self.name}}'=%s", ref.Kind, duckdbsql.EscapeStringValue(strings.ToLower(ref.Name))),
+					Allow:     true,
+				},
+			},
+		})
+	}
+	// NOTE - not adding field level or row security rules since it requires full sql parsing logic
+	// and also direct columns in the underlying table can be referenced in the sql which may not be defined in the metrics view to which security applies
+	return rules
+}
```
begelundmuller: Not sure I follow this? This is the regular SQL resolver, not metrics SQL. So the only refs would be to models, which don't have security policies / access checks.

runtime/security.go
```diff
@@ -197,8 +201,14 @@ func (p *securityEngine) resolveSecurity(instanceID, environment string, vars ma
 		return ResolvedSecurityOpen, nil
 	}
 
+	rules, err := p.expandRules(context.Background(), instanceID, claims)
```
begelundmuller: Probably need to propagate the source `ctx` instead of using background. Otherwise some of these calls might hang for cancelled requests or during graceful shutdown.

runtime/security.go
```diff
@@ -197,8 +201,14 @@ func (p *securityEngine) resolveSecurity(instanceID, environment string, vars ma
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to expand security rules: %w", err)
+	}
+	claims.AdditionalRules = rules
+
 	// Combine rules with any contained in the resource itself
-	rules := p.resolveRules(claims, r)
+	rules = p.resolveRules(claims, r)
```
begelundmuller: Feels a little dirty to overwrite the claims in-place, which is an unexpected side effect that maybe could lead to some edge cases in upstream logic. Could this be handled in a functional way?

runtime/security.go
```diff
@@ -609,6 +619,372 @@ func (p *securityEngine) applySecurityRuleRowFilter(res *ResolvedSecurity, _ *ru
...
+		if resName == nil {
+			return nil, fmt.Errorf("transitive access rule has no resource")
+		}
+		ctr, err := p.rt.Controller(ctx, instanceID)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get controller: %w", err)
+		}
+		res, err := ctr.Get(ctx, resName, false)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get resource %q of kind %q: %w", resName.Name, resName.Kind, err)
```
begelundmuller, line 651: Thinking about the end user experience here, imagine someone opens a report or magic link, but the resource has been deleted. They would probably expect a 404/"not found" error, not a "failed to get resource". If the resource is not found, the `ctr` would return `drivers.ErrResourceNotFound`, maybe we could add handling for that here: https://github.com/rilldata/rill/blob/7caad7ca93be914fc318727805bba718c6f1ff33/runtime/server/server.go#L334

runtime/security.go
```diff
@@ -609,6 +619,372 @@ func (p *securityEngine) applySecurityRuleRowFilter(res *ResolvedSecurity, _ *ru
...
+						Allow:     true,
+					},
+				},
+			})
+			// add to deny condition
+			denyCondition.WriteString(fmt.Sprintf(" OR ('{{.self.kind}}'=%s AND '{{lower .self.name}}'=%s)", duckdbsql.EscapeStringValue(ref.Kind), duckdbsql.EscapeStringValue(strings.ToLower(ref.Name))))
+			if ref.Kind == ResourceKindMetricsView {
+				mvName = ref.Name
+			}
+		}
```
begelundmuller: Can you think of any ways we could simplify some of this? Maybe with new security rules for resource access and/or for fallback access?

admin/server/alerts.go
pjain1: We were reusing earlier but then switched to creating new tokens everytime [here](https://github.com/rilldata/rill/pull/6216/files#diff-c2bcafed81fcbf8eb2837df22cb63d055009c3c64d5e9d354fbc41659e7b36aeL558). I also remember discussing this and estimating how many reports are currently being sent today and we decided to create on every run and expire after 6 months then changed it to expire every 3 months.

runtime/reconcilers/alert.go
pjain1: yeah comment is wrong, changed

runtime/resolvers/annotations.go
pjain1: yeah did it for safety reasons so that its missed not unintentionally, what do you think?

runtime/resolvers/sql.go
pjain1: Got it

runtime/security.go
pjain1: In `resolveRules` we are already appending to the slice in few places like 
```
rules := claims.AdditionalRules
...
rules = append(rules, allowAccessRule)
```
wouldn't that reflect in the backing slice in some cases like if the slice was not resized? I will not pass expanded rules slice returned from  `p.expandRules` which creates a new slice and does not point to the slice pointed by `claims.AdditionalRules`. Thus this will not append even the `builtInAlertSecurityRule` to `claims.AdditionalRules`, I hope it should be fine.

runtime/security.go
pjain1: You mean a way to not have a deny condition in addition to allow rule? That would be tricky because we add allow all access rule in `resolveRules` mainly to allow access to resource like mv, explore during usual access when there are no security rules defined inline in the resource. May be I can change allow all to just allow access to that resource that way we may not need the deny condition.

runtime/resolvers/annotations.go
begelundmuller: Yeah it's a balance, but IMO I think returning `nil` feels safer here . Or now that it returns an `error`, perhaps just a `"security rule inference not implemented"` error?

runtime/security.go
```diff
@@ -11,6 +11,9 @@ import (
 	"sync"
 
 	"github.com/hashicorp/golang-lru/simplelru"
+	tidbparser "github.com/pingcap/tidb/pkg/parser"
+	"github.com/pingcap/tidb/pkg/parser/ast"
+	"github.com/pingcap/tidb/pkg/parser/mysql"
```
begelundmuller: Can you think of a way we can leverage the `metricssqlparser` package here (maybe create a `runtime/security` package to avoid cyclic dependencies)? There's a bunch of complexity/edge cases in the SQL parsing, and we add new features and fixes with some regularity. So having to maintain SQL parsing in two separate place feels fragile.

The normal metrics SQL parser also gets quite a lot of QA from canvas usage, so we discover issues faster, but the logic here will not be hit as often in development, so easier to overlook a discrepancy during QA.

runtime/security.go
```diff
@@ -609,6 +623,579 @@ func (p *securityEngine) applySecurityRuleRowFilter(res *ResolvedSecurity, _ *ru
 	return nil
 }
 
+func (p *securityEngine) expandRules(ctx context.Context, instanceID string, claims *SecurityClaims) ([]*runtimev1.SecurityRule, error) {
```
begelundmuller: nit: would be nice with a docstring for this function. Also, `expandRules` and `resolveRules` sounds quite similar. Do you think they should be combined, or alternatively, can you think of more descriptive names for each function?

runtime/security.go
```diff
@@ -609,6 +623,579 @@ func (p *securityEngine) applySecurityRuleRowFilter(res *ResolvedSecurity, _ *ru
...
+	}
+	// explicitly allow access to the report itself
+	rules = append(rules, &runtimev1.SecurityRule{
+		Rule: &runtimev1.SecurityRule_Access{
+			Access: &runtimev1.SecurityRuleAccess{
+				Condition: fmt.Sprintf("'{{.self.kind}}'='%s' AND '{{lower .self.name}}'=%s", ResourceKindReport, duckdbsql.EscapeStringValue(strings.ToLower(res.Meta.Name.Name))),
+				Allow:     true,
+			},
+		},
+	})
```
begelundmuller: Since there's quite a lot of logic like this, which gets a little complex (and you have to think about typos and correct escaping in the expressions), I wonder if things could be simplified by refactoring the security rules to access explicit resource kind/name conditions, so we don't need to use the templated expressions (which are also kind of expensive to evaluate), e.g. something like this:
```proto
message SecurityRuleAccess {
  // The condition under which this rule applies.
  // It is ANDed together with the condition_kinds and condition_resources.
  string condition_expression = 1;
  // The resource kinds the rule applies to. If empty, it defaults to all resource kinds.
  repeated string condition_kinds = 3;
  // The resources the rule applies to. If empty, it defaults to all resources in scope covered by `resource_kinds`.
  // It is ORed together with the condition_kinds.
  repeated ResourceName condition_resources = 4;
  // Whether to allow or deny access to the resources covered by the conditions.
  bool allow = 2;
  // If true, any resource not covered by the conditions will explicitly get the opposite permission (e.g. will be denied if `allow` is true).
  bool exclusive = 5; 
}
```

Something similar could be added for the row filter and fields rules as well to simplify those. What do you think?

runtime/security.go
```diff
@@ -609,6 +623,579 @@ func (p *securityEngine) applySecurityRuleRowFilter(res *ResolvedSecurity, _ *ru
...
+				if dimFilter, ok := rendererProps["dimension_filters"].(string); ok && dimFilter != "" {
+					mvSecInfo.rowFilters = append(mvSecInfo.rowFilters, fmt.Sprintf("(%s)", dimFilter))
+
+					// Extract fields from dimension_filters SQL expression
+					dimFilterFields := extractFieldsFromSQLFilter(dimFilter)
+					for _, field := range dimFilterFields {
+						mvSecInfo.fieldSet[field] = true
+					}
+				}
+			}
```
begelundmuller: In theory, one component could reference multiple metrics views (none do today AFAIK, but there may be new/custom components that do). So would be good to decouple from the `componentRes.Meta.Refs`. Perhaps we can just not add rules for the refs, and only rely on the analysis of the renderer properties to identify metrics views?

runtime/security.go
```diff
@@ -609,6 +623,579 @@ func (p *securityEngine) applySecurityRuleRowFilter(res *ResolvedSecurity, _ *ru
...
+				fields, err := extractFieldsFromRendererProperties(componentSpec.Renderer, rendererProps)
+				if err != nil {
+					return nil, fmt.Errorf("failed to extract fields from renderer properties for component %q: %w", componentName, err)
+				}
+				for _, field := range fields {
+					mvSecInfo.fieldSet[field] = true
+				}
+
+				// Parse dimension_filters and create row filter rule
+				if dimFilter, ok := rendererProps["dimension_filters"].(string); ok && dimFilter != "" {
```
begelundmuller: The canvas components are rapidly developing at the moment. So any logic tied to specific renderer properties should ideally be handled in a single util function that is easy for the applications team to update. Having it spread out / hard-coded in different places in the file like this will probably not be easy for them to maintain.

runtime/security.go
```diff
@@ -638,3 +1225,158 @@ func computeCacheKey(instanceID, environment string, claims *SecurityClaims, r *
...
+	for _, key := range objectFields {
+		if obj, ok := rendererProps[key].(map[string]interface{}); ok {
+			if field, ok := obj["field"].(string); ok && field != "" {
+				fields = append(fields, extractDimension(field))
+			}
+		}
+	}
+
+	return fields, nil
+}
```
begelundmuller: Can you think of some util functions that would let this be unified with handling of properties like `metrics_view` and `dimension_filters`, and be easy for the applications team to maintain?

Just a quick suggestion, but maybe some util functions like this (kind of pseudocode):
```go
type rendererRefs struct {
  metricsViews map[string]bool
  metricsViewFields map[string]map[string]bool
  metricsViewRowFilters map[string][]string
}

func (r *rendererRefs) metricsView(mv any) {
  if mv, ok := mv.(string); ok {
    r.metricsViews[mv] = true
  }
}

func (r *rendererRefs) metricsViewFields(mv any, fields any) {
  mv, ok1 := mv.(string)
  fields, ok2 := mv.([]string)
  if ok1 && ok2 {
    for _, f := range fields {
      r.metricsViewFields[mv][f] = true
    }
  }
}

// etc.
```

Perhaps combined with a map accessor, something like this: https://chatgpt.com/share/68cab7d6-0244-8008-ab79-b280804f5f6d

So you could have:
```go
func extractFieldsFromRendererProperties(res *rendererRefs, renderer string, props map[string]any) error {
  switch renderer {
  case "leaderboard":
    res.metricsView(get(props, "metrics_view"))
    res.metricsViewFields(get(props, "metrics_view"), get(props, "dimensions"))
  case "bar_chart":
    res.metricsView(get(props, "metrics_view"))
    res.metricsViewFields(get(props, "metrics_view"), get(props, "color", "field"))
  // etc.
  }
}

```

begelundmuller (changes requested):

runtime/security.go
pjain1: This code is not doing full sql parsing but just used to extract field names from the `where` clause so its fairly simple and contained, but let me see if it can be commonized. I have tried to avoid it but was causing unnecessary refactoring as the only way to fix is to pull the data types shared between the two packages into a new package and make both sides depend on that instead of each other.

admin/server/reports.go
```diff
@@ -73,10 +71,13 @@ func (s *Server) GetReportMeta(ctx context.Context, req *adminv1.GetReportMetaRe
...
-		// in this mode, tokens are used only for unsubscribe links, so no access to resources or owner attributes
+		// This is the default mode for existing reports, this also implies that reports will break for users who don't have access to the project.
+		// But we agree this is acceptable and report owner needs to change to creator mode if they want to share with users who don't have access.
+		// In this mode, tokens are used only for unsubscribe links, so no access to resources or owner attributes
 		tokens, ownerEmail, err = s.createMagicTokens(ctx, proj.OrganizationID, proj.ID, req.Report, "", "", nil, recipients, nil)
 	} else {
-		tokens, ownerEmail, err = s.createMagicTokens(ctx, proj.OrganizationID, proj.ID, req.Report, req.OwnerId, req.WhereFilterJson, req.AccessibleFields, recipients, req.Resources)
+		// not setting whereFilterJSON and accessibleFields as its only needed for creator mode which is not the default mode for existing reports, so no backwards compatibility needed during rollout
+		tokens, ownerEmail, err = s.createMagicTokens(ctx, proj.OrganizationID, proj.ID, req.Report, req.OwnerId, "", nil, recipients, req.Resources)
```
begelundmuller: Why are we not setting the filter JSON and fields now? Is it because it's handled by the transitive access logic now?

proto/rill/runtime/v1/resources.proto
```diff
@@ -324,27 +324,58 @@ message SecurityRule {
...
+  // The resource kinds the rule applies to. If empty, it defaults to all resource kinds.
+  repeated string condition_kinds = 5;
+  // The resources the rule applies to. If empty, it defaults to all resources in scope covered by `resource_kinds`.
+  // It is ORed together with the condition_kinds.
+  repeated ResourceName condition_resources = 6;
   bool allow = 2;
   repeated string fields = 3;
   bool all_fields = 4;
+  // If true, all other fields not explicitly listed will get the opposite permission (e.g. will be denied if `allow` is true).
+  bool exclusive_fields = 7;
```
begelundmuller: nit: maybe just call this `exclusive`, to be consistent with the same property in `SecurityRuleAccess`.

Maybe also move it up below `bool allow` since it's quite related to that (it basically turns the allow into an exclusive allow).

runtime/reconcilers/alert.go
```diff
@@ -213,6 +214,133 @@ func (r *AlertReconciler) Reconcile(ctx context.Context, n *runtimev1.ResourceNa
...
+		if path, ok := spec.Annotations["web_open_path"]; ok {
+			// parse path, extract explore name, it will be like /explore/{explore}
+			if strings.HasPrefix(path, "/explore/") {
+				explore = path[9:]
+				if explore[len(explore)-1] == '/' {
+					explore = explore[:len(explore)-1]
+				}
+			}
+		}
+		// still not found, use mv name as explore name // TODO does this harm anything? as some alerts may not have any explore like those based on sql resolvers
```
begelundmuller: I think it's alright, since it's just an explore resource, which doesn't have any data access by itself

runtime/resolvers/legacy_metrics.go
```diff
@@ -186,6 +189,48 @@ func (r *legacyMetricsResolver) ResolveExport(ctx context.Context, w io.Writer,
...
+
+	var rules []*runtimev1.SecurityRule
+
+	if rowFilter != "" {
+		expr := &runtimev1.Expression{}
+		err := protojson.Unmarshal([]byte(rowFilter), expr)
+		if err != nil {
+			panic(status.Errorf(codes.Internal, "failed to parse row filter expression: %v", err))
```
begelundmuller: 1. Maybe return the error instead of `panic`? 
2. Avoid using the `status` package outside of the `runtime/server` package – try the normal `fmt.Errorf`

runtime/resolvers/metrics.go
```diff
@@ -158,6 +158,51 @@ func (r *metricsResolver) ResolveExport(ctx context.Context, w io.Writer, opts *
...
+	// Extract accessible fields from the query
+	var fields []string
+	// Add dimensions
+	for _, dim := range r.query.Dimensions {
+		fields = append(fields, dim.Name)
+	}
+	// Add measures
+	for _, meas := range r.query.Measures {
+		fields = append(fields, meas.Name)
+	}
```
begelundmuller: Not sure if this matters, but should it skip computed dimensions and measures (or add any underlying dimension/measure instead)?

runtime/resolvers/metrics.go
```diff
@@ -158,6 +158,51 @@ func (r *metricsResolver) ResolveExport(ctx context.Context, w io.Writer, opts *
...
+	if r.query.TimeRange != nil && r.query.TimeRange.TimeDimension != "" {
+		fields = append(fields, r.query.TimeRange.TimeDimension)
+	}
+
+	if len(fields) > 0 {
+		rules = append(rules, &runtimev1.SecurityRule{
+			Rule: &runtimev1.SecurityRule_FieldAccess{
+				FieldAccess: &runtimev1.SecurityRuleFieldAccess{
+					ConditionResources: []*runtimev1.ResourceName{{Kind: runtime.ResourceKindMetricsView, Name: r.query.MetricsView}},
+					ConditionKinds:     []string{runtime.ResourceKindExplore, runtime.ResourceKindCanvas}, // prevents field name leakage
```
begelundmuller: Why does it add all explores and canvases here? They are not required for calling the `metrics` resolver. Should this maybe be pushed into the transitive logic upstream instead?

runtime/security.go
```diff
@@ -16,6 +18,9 @@ import (
 	"github.com/rilldata/rill/runtime/pkg/pbutil"
 	"go.uber.org/zap"
 	"google.golang.org/protobuf/encoding/protojson"
+
+	// need to import parser driver as well
+	_ "github.com/pingcap/tidb/pkg/parser/test_driver"
```
begelundmuller: I think this can be removed now, right?

runtime/security.go
```diff
@@ -609,6 +630,93 @@ func (p *securityEngine) applySecurityRuleRowFilter(res *ResolvedSecurity, _ *ru
...
+			resolvedRules, err = c.reconciler(ResourceKindReport).ResolveTransitiveAccess(ctx, claims, res)
+		case *runtimev1.Resource_Alert:
+			resolvedRules, err = c.reconciler(ResourceKindAlert).ResolveTransitiveAccess(ctx, claims, res)
+		case *runtimev1.Resource_Explore:
+			resolvedRules, err = c.reconciler(ResourceKindExplore).ResolveTransitiveAccess(ctx, claims, res)
+		case *runtimev1.Resource_Canvas:
+			resolvedRules, err = c.reconciler(ResourceKindCanvas).ResolveTransitiveAccess(ctx, claims, res)
+		default:
+			return nil, fmt.Errorf("transitive access rule for resource kind %q is not supported", res.Meta.Name.Kind)
+		}
```
begelundmuller: Instead of hard-coding the types here, can it do this?
```go
resolvedRules, err := c.reconciler(res.Meta.Name.Kind).ResolveTransitiveAccess(ctx, claims, res)
```

If you feel the `"transitive access rule for resource kind %q is not supported"` error is important, then I think it should be moved to the `ResolveTransitiveAccess` functions on the reconcilers that don't support it.   But I'm not sure returning an error really makes sense – transitive access that just grants access to the current resoruce would be correct for other resources too, right?

admin/server/reports.go
pjain1: Actually after [this](https://github.com/rilldata/rill/pull/8054) hotfix, will need to keep backwards compatibility during runtime rollout after admin is upgraded. Once full upgrade is done this can be removed in next version. So will these parameters back.

proto/rill/runtime/v1/resources.proto
pjain1: This is directly linked to fields rather than application of conditions but can make it same.

runtime/resolvers/legacy_metrics.go
pjain1: sure

runtime/resolvers/metrics.go
pjain1: Correct I think we should extract base dim/measures for computed

runtime/resolvers/metrics.go
pjain1: Yes we can move these upstream in reports/alerts

runtime/security.go
pjain1: We can do `c.reconciler(res.Meta.Name.Kind).ResolveTransitiveAccess(ctx, claims, res)`

Other reconcilers than these 4 are currently errors from `ResolveTransitiveAccess`. Do you mean to say for other reconcilers they should just return a `SecurityRuleAccess` with just self `ResourceName` in `condition_resources`?

begelundmuller: Yeah that's what I mean :)

pjain1: added

runtime/controller.go
```diff
@@ -83,6 +83,19 @@ func RegisterReconcilerInitializer(resourceKind string, initializer ReconcilerIn
 	ReconcilerInitializers[resourceKind] = initializer
 }
 
+func SelfAdnRefsAllowRuleAccess(res *runtimev1.Resource) *runtimev1.SecurityRule_Access {
+	var conditionRes []*runtimev1.ResourceName
+	conditionRes = append(conditionRes, res.Meta.Name)
+	conditionRes = append(conditionRes, res.Meta.Refs...)
```
begelundmuller: Why add the refs here? Shouldn't it just be the resource itself?

E.g. if you have transitive access to `mv1`, then you need only access `mv1` in order to make metrics queries, you don't need access to the underlying model (which could even be dangerous since it would bypass other security rules like row filters).

runtime/controller.go
```diff
@@ -83,6 +83,19 @@ func RegisterReconcilerInitializer(resourceKind string, initializer ReconcilerIn
 	ReconcilerInitializers[resourceKind] = initializer
 }
 
+func SelfAdnRefsAllowRuleAccess(res *runtimev1.Resource) *runtimev1.SecurityRule_Access {
```
begelundmuller: nit: typo, should be `And`, not `Adn`

runtime/resolvers/metrics.go
```diff
@@ -251,3 +255,41 @@ func fieldsFromQuery(spec *runtimev1.MetricsViewSpec, q *metricsview.Query) []ma
...
+	}
+
+	if dim.Compute.TimeFloor != nil {
+		return dim.Compute.TimeFloor.Dimension
+	}
+
+	panic("could not find dimension name")
+}
+
+func getMeasureName(m metricsview.Measure) string {
```
begelundmuller: nit: consider moving this logic to the `metricsview` package, just to keep metrics-view related logic contained in that package when possible, e.g.:
```go
fields := metricsview.AnalyzeQueryFields(q)
```

begelundmuller (changes requested):
Also note that CI is still failing

begelundmuller (approved):

PR #7778, opened 2025-08-14
feat: improved github connection flow in rill cloud

AdityaHegde:
Adds a new flow to github connection.
1. Allows creating a repo directly through rill in a selected github org.
2. Also removes disconnect from github. Once a project is connected to a particular repo, we do not allow changing it.

Backend code cleanup will be in a follow up once we decide we wont support other flows as a fallback through cli.

Closes APP-277

**Checklist:**
- [ ] Covered by tests
- [x] Ran it and it works as intended
- [x] Reviewed the diff before requesting a review
- [x] Checked for unhandled edge cases
- [x] Linked the issues it closes
- [ ] Checked if the docs need to be updated. If so, create a separate Linear DOCS issue
- [ ] Intend to cherry-pick into the release branch
- [ ] I'm proud of this work!

PR #7828, opened 2025-08-26
feat: Move `Add Option` within component for select input

AdityaHegde:
To better support https://github.com/rilldata/rill/pull/7778 we need some improvements to the select input.
1. Generalise the `Add Option` section and add support for it in `Input` component as well.
2. Add loading state when options will be loading.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-common/src/components/forms/Input.svelte
+++ b/web-common/src/components/forms/Input.svelte
@@ -46,7 +46,7 @@
     | { value: string; label: string; type?: string }[]
     | undefined = undefined;
   export let optionsLoading: boolean = false;
-  export let onAddOption: () => void | null = null;
+  export let onAddOption: (() => void) | null = null;
   export let addOptionLabel: string | null = null;
   export let additionalClass = "";
   export let onInput: (
--- a/web-common/src/components/forms/Input.svelte
+++ b/web-common/src/components/forms/Input.svelte
@@ -45,6 +45,9 @@
   export let options:
     | { value: string; label: string; type?: string }[]
     | undefined = undefined;
+  export let optionsLoading: boolean = false;
+  export let onAddOption: () => void | null = null;
+  export let addOptionLabel: string | null = null;
   export let additionalClass = "";
   export let onInput: (
     newValue: string,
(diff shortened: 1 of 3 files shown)


web-common/src/components/forms/Select.svelte
```diff
@@ -22,6 +25,9 @@
     disabled?: boolean;
     tooltip?: string;
   }[];
+  export let optionsLoading: boolean = false;
+  export let onAddOption: (() => void) | null = null;
+  export let addOptionLabel: string | null = null;
```
ericpgreen2: I asked Claude about this and like the feedback. At least, I like reframing it as "addNew" rather than "addOption".
image

web-common/src/components/forms/Input.svelte
```diff
@@ -45,6 +45,9 @@
   export let options:
     | { value: string; label: string; type?: string }[]
     | undefined = undefined;
+  export let optionsLoading: boolean = false;
+  export let onAddOption: (() => void) | null = null;
+  export let addOptionLabel: string | null = null;
```
ericpgreen2: Can we remove these props from this component? In general, it feels like an anti-pattern that this `Input` element routes to a `Select` component under-the-hood, so I'd rather not encourage it. In those cases, I'd prefer we just use the `Select` component directly. What do you think?

web-common/src/components/forms/Select.svelte
AdityaHegde: I will change it `addNew`. I was planning to keep it separate but we needed to pass down `close` to the caller. So feels like it should be an implementation detail handled by Select component. The only use case for this option is for add so lets keep it in the select component.

web-common/src/components/forms/Input.svelte
AdityaHegde: Hmmm I am not sure why it was added in the 1st place, so opted to add these as well. I agree that we should use `Select` component directly, but I am guessing it is because of the label and other part reuse. We can revisit this during the form component refactor for new design.

PR #7829, opened 2025-08-26
Tests for rill deploy targeting actual github APIs

k-anshul:
closes https://linear.app/rilldata/issue/PLAT-18/add-tests-for-code-targeting-github-apis

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!

### DIFF
--- a/cli/cmd/deploy/deploy_test.go
+++ b/cli/cmd/deploy/deploy_test.go
@@ -53,6 +53,9 @@ func TestManagedDeploy(t *testing.T) {
 
 	// cleanup repo
 	t.Cleanup(func() {
+		// delete github repo
+		// currently github repos are deleted by a background job
+		// but for test cleanup we will delete it here directly
 		owner, repo, ok := gitutil.SplitGithubRemote(resp.Project.GitRemote)
 		require.True(t, ok, "invalid github remote: %s", resp.Project.GitRemote)
 		_, err = ghClient.Repositories.Delete(context.Background(), owner, repo)


cli/cmd/deploy/deploy_test.go
```diff
@@ -0,0 +1,204 @@
...
+	result = u1.Run(t, "deploy", "--interactive=false", "--org=github-test", "--project=rill-mgd-deploy", "--skip-deploy=true", "--path="+tempDir)
+	require.Equal(t, 0, result.ExitCode, result.Output)
+
+	// verify changes are pushed to Github repo
+	verifyGithubRepoContents(t, ghClient, resp.Project.GitRemote)
+}
+
+// This test require a Github personal access token and refresh token to work
+// Those can be generated using command `rill devtool gh-token` command
+// The refresh token can only be used once to generate a new personal access token and refresh token so the command should be executed everytime before running this test
```
k-anshul: The requirement to always generate a new token before running github tests is frustrating but unfortunately required because a refresh token changes after first use.

Few alternatives that can be considered with their cons:

1. Directly run against admin service running locally but the tests will not use a clean slate.
2. Admin service saves the updated refresh token obtained from Github. We can dump it to `.github_env` after exiting tests so that next run can use it. There will be some minimal state between different tests. It can still fail in some edge cases.

admin/testadmin/testadmin.go
```diff
@@ -50,6 +53,23 @@ type Fixture struct {
...
+	}
+	err := godotenv.Load("../../../.env")
+	require.NoError(t, err, "failed to load env file. Run `rill devtool dotenv refresh` to create one")
+
+	githubAppID, err := strconv.ParseInt(os.Getenv("RILL_ADMIN_GITHUB_APP_ID"), 10, 64)
+	require.NoError(t, err)
+
+	github, err := admin.NewGithub(t.Context(), githubAppID, os.Getenv("RILL_ADMIN_GITHUB_APP_PRIVATE_KEY"), os.Getenv("RILL_ADMIN_GITHUB_MANAGED_ACCOUNT"), zap.Must(zap.NewDevelopment()))
+	require.NoError(t, err)
+	return github
```
begelundmuller: Should we have test-scoped names for these? E.g. `RILL_ADMIN_TEST_GITHUB_APP_PRIVATE_KEY` and `RILL_ADMIN_TEST_GITHUB_MANAGED_ACCOUNT`. 

Just to keep tests configured separately. Maybe I'm wrong, but I don't believe any tests currently use the normal env variables.

admin/testadmin/testadmin.go
```diff
@@ -50,6 +53,23 @@ type Fixture struct {
 	ServerOpts *server.Options
 }
 
+// NewGithub creates a new Github client. In short testing mode this is a mock client which has no-op implementations of all methods.
+// Otherwise it creates a real implementation that makes real API calls to Github.
+func NewGithub(t *testing.T) admin.Github {
```
begelundmuller: 1. Can you move this later in the file? It breaks the style guide on function ordering: https://github.com/uber-go/guide/blob/master/style.md#function-grouping-and-ordering
2. Since it's only used inside this package, should it be `newGithub`?

cli/cmd/deploy/deploy.go
```diff
@@ -83,5 +83,10 @@ func DeployCmd(ch *cmdutil.Helper) *cobra.Command {
 	deployCmd.MarkFlagsMutuallyExclusive("managed", "archive", "subpath")
 	deployCmd.MarkFlagsMutuallyExclusive("managed", "archive", "github")
 
+	deployCmd.Flags().BoolVar(&opts.SkipDeploy, "skip-deploy", false, "Skip the runtime deployment step (for testing only)")
+	err = deployCmd.Flags().MarkHidden("skip-deploy")
+	if err != nil {
+		panic(err)
+	}
```
begelundmuller: Wrap this with `if !ch.IsDev() {` – so we show all flags in dev builds

cli/cmd/deploy/deploy_test.go
```diff
@@ -0,0 +1,204 @@
...
+	require.NoError(t, err)
+	ghClient := adm.Admin.Github.InstallationClient(installationID, nil)
+
+	// cleanup repo
+	t.Cleanup(func() {
+		owner, repo, ok := gitutil.SplitGithubRemote(resp.Project.GitRemote)
+		require.True(t, ok, "invalid github remote: %s", resp.Project.GitRemote)
+		_, err = ghClient.Repositories.Delete(context.Background(), owner, repo)
+		require.NoError(t, err, "failed to delete github repo %s/%s: %w", owner, repo, err)
+	})
```
begelundmuller, line 63: Can you use `DeleteProject` instead, and rely on that to do cleanup? Would make for a better integration test since it would mirror what end users might do.

cli/cmd/deploy/deploy_test.go
```diff
@@ -0,0 +1,204 @@
...
+
+	// redeploy the same project with changes
+	err = os.Mkdir(filepath.Join(tempDir, "models"), 0755)
+	require.NoError(t, err)
+	err = os.WriteFile(filepath.Join(tempDir, "models/model.sql"), []byte(`SELECT 1 AS one`), 0644)
+	require.NoError(t, err)
```
begelundmuller: nit: consider using a map of files and having a util function similar to `testruntime.PutFiles`: https://github.com/rilldata/rill/blob/68143db83dfb6aa9062fa8471a3af30087e262ab/runtime/testruntime/reconcile.go#L18

cli/cmd/deploy/deploy_test.go
```diff
@@ -0,0 +1,204 @@
...
+
+func verifyGithubRepoContents(t *testing.T, client *github.Client, remote string) {
+	owner, repo, ok := gitutil.SplitGithubRemote(remote)
+	require.True(t, ok, "invalid github remote: %s", remote)
+
+	con, _, _, err := client.Repositories.GetContents(t.Context(), owner, repo, "models/model.sql", nil)
+	require.NoError(t, err)
+	contents, err := con.GetContent()
+	require.NoError(t, err)
+	require.Equal(t, "SELECT 1 AS one", contents)
```
begelundmuller: Instead of hard-coding one file here, if you have a map of files as per my comment above, you could pass that map in here and iterate over it, validating that each file is in the repo.

cli/cmd/deploy/deploy_test.go
begelundmuller: Yes this is very annoying, it would be very nice to avoid this `.github_env` file and associated hacks here.

Did you consider using `gh auth token` to get the current user's token? If it fails, you can error and say the test requires that `gh` is installed and authenticated as a personal user.

cli/cmd/project/deploy.go
```diff
@@ -276,6 +279,12 @@ func DeployCmd(ch *cmdutil.Helper) *cobra.Command {
 		}
 	}
 
+	deployCmd.Flags().BoolVar(&opts.SkipDeploy, "skip-deploy", false, "Skip the runtime deployment step (for testing only)")
+	err := deployCmd.Flags().MarkHidden("skip-deploy")
+	if err != nil {
+		panic(err)
+	}
```
begelundmuller: Same comment as earlier here

cli/pkg/gitutil/gitcmdwrapper.go
```diff
@@ -193,9 +194,9 @@ func isGitIgnored(repoRoot, subpath string) (bool, error) {
 				return false, nil
 			}
 			// any other exit code is an error
-			return false, fmt.Errorf("git check-ignore failed: %s", string(execErr.Stderr))
+			return false, fmt.Errorf("git check-ignore failed for path %q, subpath: %q : %s", repoRoot, subpath, string(execErr.Stderr))
 		}
-		return false, fmt.Errorf("git check-ignore failed: %w", err)
+		return false, fmt.Errorf("git check-ignore failed for path %q, subpath: %q : %w", repoRoot, subpath, err)
```
begelundmuller: nit: try the format `path %q, subpath %q: %w` – colons should be used for different levels of context, not for key-value pairs; and there should not be a space before a colon.

cli/cmd/deploy/deploy_test.go
k-anshul: The repo is not deleted on project delete but deleted via a background worker.

begelundmuller: Can you check if River has an easy way to test background job execution? Like a way to await all jobs to finish processing.

If there's not an easy way, you can keep this code, but then can you add a comment that it's there since we don't have test support for background jobs yet.

PR #7832, opened 2025-08-26
Pull project environment variables for authenticated users on start

grahamplata:
The `rill start` will automatically pull the env after the repository has been cloned. Assuming the use has the proper privileges (e.g. admin rights to the project)

**Checklist:**
- [ ] Covered by tests
- [x] [Linked](https://linear.app/rilldata/issue/PLAT-173/rill-start-github-repo-should-also-execute-a-rill-env-pull) the issues it closes
- [x] I'm proud of this work!

### DIFF
--- a/cli/cmd/start/start.go
+++ b/cli/cmd/start/start.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	"github.com/joho/godotenv"
+	"github.com/rilldata/rill/cli/cmd/env"
 	"github.com/rilldata/rill/cli/pkg/cmdutil"
 	"github.com/rilldata/rill/cli/pkg/gitutil"
 	"github.com/rilldata/rill/cli/pkg/local"
@@ -47,6 +48,15 @@ func StartCmd(ch *cmdutil.Helper) *cobra.Command {
 					}
 
 					projectPath = repoName
+
+					if ch.IsAuthenticated() {
+						if _, err := os.Stat(filepath.Join(projectPath, "rill.yaml")); err == nil {
+							err := env.PullVars(cmd.Context(), ch, projectPath, "", environment, true)
+							if err != nil {
+								ch.PrintfWarn("Warning: failed to pull environment credentials: %v\n", err)
+							}
+						}
+					}
 				}
 			} else if !cmdutil.HasRillProject(".") {
 				if !ch.Interactive {


cli/cmd/start/start.go
```diff
@@ -47,6 +48,15 @@ func StartCmd(ch *cmdutil.Helper) *cobra.Command {
 					}
 
 					projectPath = repoName
+
+					if ch.IsAuthenticated() {
```
k-anshul: Why does this happen only when starting rill using github remote path?
Shouldn't it happen in all cases i.e. when starting rill in a local directory connected to a rill project ?

k-anshul (changes requested):

cli/cmd/start/start.go
grahamplata: Very true, made an adjustment

cli/cmd/start/start.go
```diff
@@ -48,6 +49,15 @@ func StartCmd(ch *cmdutil.Helper) *cobra.Command {
 
 					projectPath = repoName
 				}
+
+				if ch.IsAuthenticated() {
```
k-anshul: I may be missing something but shouldn't this be done before `// Parse log format` section otherwise this does not work for cases when I `cd` to a rill project directory and do `rill start` without specifying any path.

cli/cmd/start/start.go
```diff
@@ -102,6 +103,16 @@ func StartCmd(ch *cmdutil.Helper) *cobra.Command {
 				projectPath = "."
 			}
 
+			// Always attempt to pull env for any valid Rill project (after projectPath is set)
+			if ch.IsAuthenticated() {
+				if _, err := os.Stat(filepath.Join(projectPath, "rill.yaml")); err == nil {
+					err := env.PullVars(cmd.Context(), ch, projectPath, "", environment, true)
```
k-anshul:
```suggestion
					err := env.PullVars(cmd.Context(), ch, projectPath, "", environment, false)
```
`warnForNoVars` should be false I think.

cli/cmd/start/start.go
```diff
@@ -102,6 +103,16 @@ func StartCmd(ch *cmdutil.Helper) *cobra.Command {
 				projectPath = "."
 			}
 
+			// Always attempt to pull env for any valid Rill project (after projectPath is set)
+			if ch.IsAuthenticated() {
+				if _, err := os.Stat(filepath.Join(projectPath, "rill.yaml")); err == nil {
```
k-anshul:
```suggestion
				if local.IsProjectInit(projectPath) {
```

cli/cmd/start/start.go
```diff
@@ -102,6 +103,16 @@ func StartCmd(ch *cmdutil.Helper) *cobra.Command {
...
+					err := env.PullVars(cmd.Context(), ch, projectPath, "", environment, true)
+					if err != nil {
```
k-anshul:
```suggestion
					if err != nil && !errors.Is(err, cmdutil.ErrNoMatchingProject) {
```

cli/cmd/start/start.go
begelundmuller: I think there's a util for this: https://github.com/rilldata/rill/blob/1804470df9ae354293711d6b4a5c20f32106dcb2/cli/pkg/local/app.go#L471

k-anshul (approved):

PR #7840, opened 2025-08-26
ClickHouse Cloud Option

lovincyrus:
This PR adds a third option (ClickHouse Cloud) in the connector type dropdown for the ClickHouse connector.

Closes https://linear.app/rilldata/issue/APP-272/add-a-third-option-in-the-connection-type-dropdown-for-clickhouse

CleanShot 2025-08-26 at 16 17 52@2x

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-common/src/features/sources/modal/AddClickHouseForm.svelte
+++ b/web-common/src/features/sources/modal/AddClickHouseForm.svelte
@@ -381,7 +381,9 @@
                   bind:checked={$paramsForm[propertyKey]}
                   label={property.displayName}
                   hint={property.hint}
-                  optional={!property.required}
+                  optional={connectorType === "clickhouse-cloud" && isSSLField
+                    ? false
+                    : !property.required}
                   disabled={connectorType === "clickhouse-cloud" && isSSLField}
                 />
               {:else if property.type === ConnectorDriverPropertyType.TYPE_INFORMATIONAL}


lovincyrus:
Ready for UXQA @ericokuma

ericokuma:
* What happened to the connection string option? 
* Also the snippet below ClickHouse Cloud "connect to your clickhouse…" isn't necessary. 
* Don't think we should hide the SSL option. We should just enable and lock the option. In the tooltip, we'll just mention that SSL is automatically enabled for ClickHouse Cloud connections

lovincyrus:
> * What happened to the connection string option?

I didn't see the connection string as part of the requirement in https://linear.app/rilldata/issue/APP-272/add-a-third-option-in-the-connection-type-dropdown-for-clickhouse. Adding this.

ericokuma:
Ah my apologies

ericokuma:
Okay looks good with some minor adjustments:

1. The Port doesn't look disabled. Also are we sure 8443 is the only port for ClickHouse Cloud? @NamanMahor? I heard from Mike that it's possible to use 9440 as well?
2. Move the "port is fixed to 8443 for Clickhouse Cloud (HTTPS) to the tooltip and replace what was there previously
3. Move the "SSL is always enabled for ClickHouse Cloud" into the tooltip and replace what was there previously

lovincyrus:
> * Move the "port is fixed to 8443 for Clickhouse Cloud (HTTPS) to the tooltip and replace what was there previously
> * Move the "SSL is always enabled for ClickHouse Cloud" into the tooltip and replace what was there previously

Removed these two messages. There is only one single truth for these properties: https://github.com/rilldata/rill/blob/main/runtime/drivers/clickhouse/clickhouse.go#L86 We can't have dynamic descriptions for the time being.

ericokuma:
oh…in that case, we should have the more generic messages in the tooltips if we can't do dynamic descriptions…

lovincyrus:
> oh…in that case, we should have the more generic messages in the tooltips if we can't do dynamic descriptions…

This is what we have for `port`:

CleanShot 2025-09-10 at 16 38 29@2x

and `ssl`:

CleanShot 2025-09-10 at 16 43 44@2x

lovincyrus:
Disabled input styles in Clickhouse Cloud:

CleanShot 2025-09-10 at 16 50 44@2x

ericokuma:
LGTM. One thing I missed is removing the "optional" flag from SSL in ClickHouse Cloud. But other than that, looks good!

lovincyrus:
> removing the "optional" flag from SSL in ClickHouse Cloud

CleanShot 2025-09-11 at 10 24 11@2x

PR #7842, opened 2025-08-27
Install motherduck extension and set token when connector is motherduck

NamanMahor:
We should auto install motherduck when connector is motherduck

https://linear.app/rilldata/issue/APP-304/add-schema-name-input-field-to-motherduck#comment-c6cec213

Below format will work now.
```
type: connector

driver: motherduck
token: '{{ .env.connector.motherduck.token }}'
path: "md:sample_data"
schema_name: kaggle
```

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!

### DIFF
--- a/runtime/drivers/duckdb/config.go
+++ b/runtime/drivers/duckdb/config.go
@@ -41,6 +41,8 @@ type config struct {
 	// Attach allows user to pass a full ATTACH statement to attach a DuckDB database.
 	// Example YAML syntax : attach: "'ducklake:metadata.ducklake' AS my_ducklake(DATA_PATH 'datafiles1')"
 	Attach string `mapstructure:"attach"`
+	// Token is the authentication token used for MotherDuck.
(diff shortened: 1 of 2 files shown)


runtime/drivers/duckdb/config.go
```diff
@@ -41,6 +41,8 @@ type config struct {
 	// Attach allows user to pass a full ATTACH statement to attach a DuckDB database.
 	// Example YAML syntax : attach: "'ducklake:metadata.ducklake' AS my_ducklake(DATA_PATH 'datafiles1')"
 	Attach string `mapstructure:"attach"`
+	// Token is the authentication token used for MotherDuck.
+	Token string `mapstructure:"token"`
```
k-anshul, line 45: It is okay to keep the YAML field as token but probably the internally Go field should be `MotherDuckToken` since the config is generic and token is specific to a implementation.

PR #7846, opened 2025-08-28
update connector reconciler state on failures as well

k-anshul:
Since the instance connector list is updated before connectors are tested so we should update the state even if the reconciler fails.
Otherwise it can lead to following situation:
1. Initially the connector is working state as below:
```
type: connector
driver: duckdb
init_sql: |
    SET threads = 2;
```
2. User makes an incorrect change like below. At this point connector definition has been updated in instance but state  still points to definition 1 above.
```
type: connector
driver: duckdb
init_sql: |
    SET threads = 2; SET invalid_prop = 1;
```
3. User reverts to previous working version. The change is ignored because the state is same as current version leading to the connector in incorrect state till a runtime restart.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/runtime/reconcilers/connector.go
+++ b/runtime/reconcilers/connector.go
@@ -94,13 +94,10 @@ func (r *ConnectorReconciler) Reconcile(ctx context.Context, n *runtimev1.Resour
 
 	// Test the connector configuration
 	err = r.testConnector(ctx, self.Meta.Name.Name)
-	if err != nil {
-		return runtime.ReconcileResult{Err: fmt.Errorf("validation failed: %w", err)}
-	}
-
+	// update state even if test fails because the instance connectors have already been updated
 	t.State.SpecHash = specHash
 
-	err = r.C.UpdateState(ctx, self.Meta.Name, self)
+	err = errors.Join(err, r.C.UpdateState(ctx, self.Meta.Name, self))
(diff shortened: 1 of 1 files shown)


runtime/reconcilers/connector.go
```diff
@@ -94,13 +94,10 @@ func (r *ConnectorReconciler) Reconcile(ctx context.Context, n *runtimev1.Resour
...
 	err = r.testConnector(ctx, self.Meta.Name.Name)
-	if err != nil {
-		return runtime.ReconcileResult{Err: fmt.Errorf("validation failed: %w", err)}
-	}
-
+	// update state even if test fails because the instance connectors have already been updated
 	t.State.SpecHash = specHash
 
-	err = r.C.UpdateState(ctx, self.Meta.Name, self)
+	err = errors.Join(err, r.C.UpdateState(ctx, self.Meta.Name, self))
```
begelundmuller, line 102: If we make this change, then after the next runtime restart, I think the validation error may not be reported anymore. Because it will return no error in the earlier case on line 85:
```
	if specHash == t.State.SpecHash {
		return runtime.ReconcileResult{}
	}
```

begelundmuller (approved):

PR #7848, opened 2025-08-28
[APP-157] feat: secondary time column selection on explore

briangregoryholmes:
**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


ericpgreen2:
I believe I missed the demo of this feature when I was OOO, so I'm a little out-of-the-loop. What's the status on this?

briangregoryholmes:
This is ready. Just need #7779 and #7857 to merge first.

proto/rill/runtime/v1/resources.proto
```diff
@@ -448,6 +448,7 @@ message ExplorePreset {
   optional string timezone = 12;
   optional string time_grain = 13;
   optional string select_time_range = 14;
+  optional string time_column = 33;
```
begelundmuller: nit: call this `time_dimension` to reflect the terminology we use elsewhere ("dimension" basically meaning something the metrics layer can aggregate by, which may not be an actual column in a database table).

ericpgreen2:
@mindspank, can you please do the Product review? For one, can you ensure all views (e.g. leaderboards, time dimension detail, pivot) are consistent?

@AdityaHegde, can you please do the code review?

web-common/src/features/dashboards/filters/Filters.svelte
```diff
@@ -206,6 +217,39 @@
...
   }
 
+  async function onTimeDimensionSelect(column: string) {
+    // Capture time range name before any state changes
+    const timeRangeName = selectedTimeRange?.name;
+
+    metricsExplorerStore.setTimeDimension($exploreName, column);
+
+    // Re-resolve the time range with the new time dimension
+    if (timeRangeName) {
```
AdityaHegde: nit: invert if to avoid additional block

web-common/src/features/dashboards/filters/Filters.svelte
```diff
@@ -206,6 +217,39 @@
     setMeasureFilter(dimension, filter);
   }
 
+  async function onTimeDimensionSelect(column: string) {
+    // Capture time range name before any state changes
+    const timeRangeName = selectedTimeRange?.name;
+
+    metricsExplorerStore.setTimeDimension($exploreName, column);
```
AdityaHegde, line 229: nit: cancel queries before updating the state.

It also need not be inside the `if (timeRangeName)` right?

royendo:
doing some QA; overall great! I've noticed a few things that aren't blockers but nits

1. the setup wasn't self-explanatory. I had to check to see the metrics YAML in this PR to get it working. There's no indicator here to have the user enable `type: time`.
Screenshot 2026-01-16 at 17 21 53

2. Not caused by this PR but if i add a time dimension then remove it from metrics view, im stuck in forever loop. unless i edit URL directly
Screenshot 2026-01-16 at 17 23 07

3. im not personally much of a fan of "Stick it in the dropdown"... its getting long.. Not that I have a great suggestion though.

briangregoryholmes:
@royendo Added a type selection dropdown to dimensions and resolved point 2. Will merge as is. Do you have docs handled?

ericpgreen2:
@mindspank, have you reviewed this? Is this ready-to-merge from the Product perspective?

mindspank:
@ericpgreen2 roy has reviewed it, good to go if he says so

royendo:
Screenshot 2026-01-20 at 15 00 01

Not sure what the best wording here is but, it seems like you can put any columns as a "time" type but this bugs out the UI.

https://www.loom.com/share/11b34b689532406fae22cc528d1c22eb

briangregoryholmes:
Checking with @begelundmuller via Slack about this. We should be able to infer the type so that this is not required and I can also maybe see a small reason for not limiting the valid types.

briangregoryholmes:
@royendo After running this by @begelundmuller and @medriscoll , I'm going to leave this largely as is for now, though what I've added is a confirmation modal whenever you try to assign type `time` to a non-TIMESTAMP/DATE column and vice versa. We may add stricter validation in the future. Apparently automatically inferring DATE columns was causing issues, so it will still be necessary to use this control manually, which should be documented.

Screenshot 2026-01-23 at 6 12 46 PM


Just give me a thumbs up that this works for you and I'll merge.

PR #7850, opened 2025-08-29
Change auto login flow and assume open flow

NamanMahor:
[PLAT-137: Pentest: Change `/auth/with-token` to use short-lived tokens](https://linear.app/rilldata/issue/PLAT-137/pentest-change-authwith-token-to-use-short-lived-tokens).

Authflow change.
 1. UI requests <custom-domain>/auth/login?redirect=<redirect>
    - Set a cookie with state=XXX
    - If a redirect URL is provided, store it in the cookie
 2. Redirect to <canonical-domain>/auth/login?redirect=https://<custom-domain>/auth/custom-domain-callback?state=XXX
    - Set a cookie with state=YYY
 3. Redirect to Auth0 /login?state=YYY
    - User completes login
 4. Redirect to <canonical-domain>/auth/callback?state=YYY&<oauth-params>
    - Verify cookie state matches YYY
    - Issue a noune token
 5. Redirect to <custom-domain>/auth/custom-domain-callback?state=XXX&noune_token=<short-lived-token>
    - Verify cookie state matches XXX
    - Exchange short-lived token for a long-lived token
    - Save the long-lived token in a cookie
    - Redirect to the saved redirect URL (or the frontend root if none was provided)
 
 
Assume user open changes:
The CLI no longer sends the token in the URL when opening as another user. Instead, it sends only the user’s email to the admin API, which issues a new token.
The unassume action now deletes all tokens that the current user created for representing the specified user.
 

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!

### DIFF
--- a/admin/urls.go
+++ b/admin/urls.go
@@ -112,11 +112,7 @@ func (u *URLs) IsCustomDomain(domain string) bool {
 	if err != nil {
 		panic(fmt.Errorf("failed to parse external domain %q: %w", u.external, err))
 	}
-	domainURL, err := url.Parse(domain)
-	if err != nil {
-		panic(fmt.Errorf("failed to parse provided domain %q: %w", domain, err))
-	}
-	return !strings.EqualFold(externalURL.Host, domainURL.Host)
+	return !strings.EqualFold(externalURL.Host, domain)
 }
 
 // External returns the external URL for the admin service.


cli/cmd/sudo/user/open.go
```diff
@@ -10,27 +12,37 @@ import (
...
-			qry := map[string]string{"token": ch.AdminToken()}
-			withTokenURL, err := urlutil.WithQuery(withTokenURI, qry)
+			representingUser, err := ch.DotRill.GetRepresentingUser()
 			if err != nil {
 				return err
 			}
-
-			_ = browser.Open(withTokenURL)
+			if representingUser == "" {
+				return errors.New("no representing user configured, please assume a user first")
```
begelundmuller: Let's remove the "representing" part, might just confuse some people:

```suggestion
				return errors.New("please assume a user first")
```

admin/server/users.go
```diff
@@ -333,6 +333,35 @@ func (s *Server) RevokeUserAuthToken(ctx context.Context, req *adminv1.RevokeUse
 	return &adminv1.RevokeUserAuthTokenResponse{}, nil
 }
 
+func (s *Server) RevokeRepresentativeAuthTokens(ctx context.Context, req *adminv1.RevokeRepresentativeAuthTokensRequest) (*adminv1.RevokeRepresentativeAuthTokensResponse, error) {
+	claims := auth.GetClaims(ctx)
+
+	if !claims.Superuser(ctx) {
+		return nil, status.Error(codes.PermissionDenied, "only superusers can search users by email")
+	}
```
begelundmuller, line 341: Error message doesn't match the RPC:
```suggestion
	if !claims.Superuser(ctx) {
		return nil, status.Error(codes.PermissionDenied, "only superusers can manage representative auth tokens")
	}
```

admin/server/auth/handlers.go
```diff
@@ -140,7 +167,16 @@ func (a *Authenticator) authStart(w http.ResponseWriter, r *http.Request, signup
 		return
 	}
 
-	// Redirect to auth provider
+	// Redirect to <canonical-domain>/auth/login (custom domain flow)
+	if a.admin.URLs.IsCustomDomain(r.Host) {
```
begelundmuller, line 178: This checks probably won't work – the `IsCustomDomain` implementation checks against `u.custom`, but that's only set when you use `WithCustomDomain`: https://github.com/rilldata/rill/blob/7caad7ca93be914fc318727805bba718c6f1ff33/admin/server/organizations.go#L474

You may need to use `a.admin.DB.FindOrganizationByCustomDomain` to check the domain here instead.

admin/server/auth/handlers.go
```diff
@@ -140,7 +167,16 @@ func (a *Authenticator) authStart(w http.ResponseWriter, r *http.Request, signup
...
+		// Redirect to canonical domain with custom domain callback as redirect
+		customCallbackURL := a.admin.URLs.External() + "/auth/custom-domain-callback?state=" + state
```
begelundmuller: It should not do URL building using string concats, it should do it with a new func on `admin.URLs` and use the proper URL building utils.

admin/server/auth/handlers.go
```diff
@@ -261,45 +302,107 @@ func (a *Authenticator) authLoginCallback(w http.ResponseWriter, r *http.Request
 		return
 	}
 
+	redirectURL, err := url.Parse(redirect)
+	if err != nil {
+		http.Error(w, fmt.Sprintf("failed to parse redirect URL: %s", err), http.StatusInternalServerError)
+		return
+	}
+
+	if a.admin.URLs.IsCustomDomain(redirectURL.Hostname()) {
```
begelundmuller: See comment above, not sure this will work.

If in the first call to `/auth/login` you notice it's a custom domain, consider storing that info in the cookie (e.g. `sess.Values[cookieFieldCustomDomainRedirect]`), so you don't need to do another lookup of that here. Might also make some of the URL handling simpler (no need to parse with `url.Parse`).

admin/server/auth/handlers.go
```diff
@@ -324,6 +427,97 @@ func (a *Authenticator) authWithToken(w http.ResponseWriter, r *http.Request) {
...
+	validated, err := a.admin.ValidateAuthToken(ctx, authToken)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	claims := newAuthTokenClaims(validated, a.admin)
+	if claims.OwnerType() != OwnerTypeUser {
+		http.Error(w, "not authenticated as a user", http.StatusUnauthorized)
+		return
+	}
```
begelundmuller, line 450: If you wrap this handler with the auth middleware, I think you can simplify this similar to here: https://github.com/rilldata/rill/blob/7caad7ca93be914fc318727805bba718c6f1ff33/admin/server/auth/device_code.go#L129

admin/server/auth/handlers.go
```diff
@@ -324,6 +427,97 @@ func (a *Authenticator) authWithToken(w http.ResponseWriter, r *http.Request) {
...
+	mdl, ok := validated.TokenModel().(*database.UserAuthToken)
+	if !ok {
+		http.Error(w, "invalid user auth token model", http.StatusBadRequest)
+		return
+	}
+	if u.ID == mdl.UserID {
+		http.Error(w, "representing user cannot represent yourself", http.StatusBadRequest)
+		return
+	}
```
begelundmuller: See comment above, could just use `claims.OwnerID()` for this check

admin/server/auth/handlers.go
NamanMahor: using u.external for matching.

admin/urls.go
```diff
@@ -106,6 +106,15 @@ func (u *URLs) IsHTTPS() bool {
...
 }
 
+// IsCustomDomain returns true if the given domain is a custom domain.
+func (u *URLs) IsCustomDomain(domain string) bool {
+	externalURL, err := url.Parse(u.external)
+	if err != nil {
+		panic(fmt.Errorf("failed to parse external domain %q: %w", u.external, err))
+	}
+	return !strings.EqualFold(externalURL.Host, domain)
+}
```
begelundmuller, line 116: Do you think it would make sense to also parse the domain just to be sure there's no scheme or trailing `/`? 

```suggestion
// IsCustomDomain returns true if the given domain is a custom domain.
func (u *URLs) IsCustomDomain(domain string) bool {
	externalURL, err := url.Parse(u.external)
	if err != nil {
		panic(fmt.Errorf("failed to parse external domain %q: %w", u.external, err))
	}
	domainURL, err := url.Parse(domain)
	if err != nil {
		panic(fmt.Errorf("failed to parse provided domain %q: %w", domain, err))
	}
	return !strings.EqualFold(externalURL.Host, domainURL.Host)
}
```

admin/server/auth/handlers.go
```diff
@@ -134,13 +162,27 @@ func (a *Authenticator) authStart(w http.ResponseWriter, r *http.Request, signup
 		sess.Values[cookieFieldRedirect] = redirect
 	}
 
+	customDomainFlow := r.URL.Query().Get("custom_domain_flow")
```
begelundmuller, line 166: Maybe a small comment to explain:

```suggestion
	// If this is part of the custom domain login flow, save that info in the cookie since we need that info when handling the auth callback.
	customDomainFlow := r.URL.Query().Get("custom_domain_flow")
```

admin/server/auth/handlers.go
```diff
@@ -261,45 +308,103 @@ func (a *Authenticator) authLoginCallback(w http.ResponseWriter, r *http.Request
 		return
 	}
 
+	customDomainFlow, ok := sess.Values[cookieFieldCustomDomainFlow].(bool)
+	delete(sess.Values, cookieFieldRedirect)
```
begelundmuller: I think this should be:
```suggestion
	delete(sess.Values, cookieFieldCustomDomainFlow)
```

admin/server/auth/handlers.go
```diff
@@ -261,45 +308,103 @@ func (a *Authenticator) authLoginCallback(w http.ResponseWriter, r *http.Request
 		return
 	}
 
+	customDomainFlow, ok := sess.Values[cookieFieldCustomDomainFlow].(bool)
```
begelundmuller, line 313: A small comment here would be helpful as well:

```suggestion
	// If it's part of a custom domain login flow, redirect back to the custom domain with a short-lived access token for the user.
	customDomainFlow, ok := sess.Values[cookieFieldCustomDomainFlow].(bool)
```

begelundmuller (approved):

PR #7834, opened 2025-08-26
feat: add combo bar chart

djbarnwal:
Adds combo bar chart

https://www.notion.so/rilldata/Multiple-Metrics-Dual-Axis-Support-in-Canvas-Charts-255ba33c8f5780ffa3fbccf4309104a4?source=copy_link#255ba33c8f57801bb6d1dba2d2540493

Closes https://linear.app/rilldata/issue/APP-291/support-for-combo-chart-type-with-dual-y-axes

**Checklist:**
- [ ] Covered by tests
- [x] Ran it and it works as intended
- [x] Reviewed the diff before requesting a review
- [x] Checked for unhandled edge cases
- [x] Linked the issues it closes
- [ ] Checked if the docs need to be updated. If so, create a separate Linear DOCS issue
- [ ] Intend to cherry-pick into the release branch
- [ ] I'm proud of this work!

PR #7860, opened 2025-08-31
refac: enable multi metric charts from cartesian charts

djbarnwal:
Adds ability to enable multi metric charts in existing charts
https://linear.app/rilldata/issue/APP-290/support-for-a-ux-path-to-activate-multiple-measuresmetrics-in-a-subset

Have also merged https://github.com/rilldata/rill/pull/7834 into this branch

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


ericokuma:
@djbarnwal, is there any way to persist multiple metric selection when user navigates from a supported cartesian chart to a non-supported (like Donut Chart) and then back to a supported chart?<br><br>e.g. 

1. I select multiple measures in Bar Chart
2. I click on Donut chart
3. I click back to Bar Chart 

Currently, we clear the multiple measures selection

djbarnwal:
> @djbarnwal, is there any way to persist multiple metric selection when user navigates from a supported cartesian chart to a non-supported (like Donut Chart) and then back to a supported chart?e.g.

Moving from one chart type to a different one (where the field types are different) will lose some data. We try to preserve fields which have the same key name. We can build something to keep track of existing field but it is not likely a very small solution. It would probably require another PR.

ericokuma:
Gotcha, okay. I'll create a separate ticket for this!

PR #7863, opened 2025-09-02
do not wrap prev time in any_value as its not a measure

pjain1:
Fixes https://linear.app/rilldata/issue/PLAT-158/fix-invalid-date-issue-on-druid-backed-projects

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/runtime/metricsview/ast.go
+++ b/runtime/metricsview/ast.go
@@ -530,6 +530,10 @@ func (a *AST) resolveMeasure(qm Measure, visible bool) (*runtimev1.MetricsViewSp
 			Expression:  expr,
 			Type:        runtimev1.MetricsViewSpec_MEASURE_TYPE_TIME_COMPARISON,
 			DisplayName: fmt.Sprintf("Comparison time for %s", qd.Name),
+			// adding data type - needed for proper handling while adding this field to ast as a dimension
+			DataType: &runtimev1.Type{
+				Code: runtimev1.Type_CODE_TIMESTAMP,
+			},
 		}, nil
 	}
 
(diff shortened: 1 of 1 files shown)


runtime/metricsview/ast.go
```diff
@@ -1106,6 +1110,18 @@ func (a *AST) addTimeComparisonMeasure(n *SelectNode, m *runtimev1.MetricsViewSp
 	if err != nil {
 		return err
 	}
+
+	// This is mainly required for getting the correct data type for druid as it wraps all measures in ANY_VALUE in rewriteDruidGroups.
```
pjain1, line 1114: Another way is to just solve this for druid by looking for presence of time column computation pattern in `rewriteDruidGroups` and not wrapping that measure but that seems fragile.

runtime/metricsview/ast.go
```diff
@@ -1106,6 +1110,18 @@ func (a *AST) addTimeComparisonMeasure(n *SelectNode, m *runtimev1.MetricsViewSp
...
+	// However, wrapping timestamp in ANY_VALUE return unix epoch as string, anyway computed time is actually a dimension, but it's sent as a measure for convenience
+	if m.DataType != nil && m.DataType.Code == runtimev1.Type_CODE_TIMESTAMP {
+		n.DimFields = append(n.DimFields, FieldNode{
+			Name:        m.Name,
+			DisplayName: m.DisplayName,
+			Expr:        expr,
+		})
```
begelundmuller, line 1121: It feels a bit weird to add it to the `DimFields` since we group by the dim fields. I'm worried having extra group dimensions would make the DB pick less optimal code paths. (E.g. in Druid, would it still be a timeseries query or would it become a groupby?)

I'm also worried it might cause issues in code that relies on the number of query dimensions mapping to the number of `DimFields` (which I believe has been a guarantee so far?).

pjain1:
You re right, fixing it on druid side will make more sense, so closing it here.

PR #7867, opened 2025-09-02
Test connection for connectors

lovincyrus:
This PR supports the ability to test the connection for connectors. Closes https://linear.app/rilldata/issue/APP-23/add-test-connection-feature-to-all-connectors-ui and https://linear.app/rilldata/issue/APP-275/add-our-standard-loading-animation-icon-into-the-button-for-test-and

Another notable change is that we now test olap connectors using resource-level reconciliation.

Postgres `ImplementsSQLStore` (happy path)

https://github.com/user-attachments/assets/25bb9e61-e21a-490d-ba6f-a9672e936a49

MySQL `ImplementsSQLStore` (unhappy path)

https://github.com/user-attachments/assets/210fab27-7091-4f8e-a70c-924254456e99

Redshift `ImplementsWarehouse` (unhappy path)

https://github.com/user-attachments/assets/30fbad3c-8192-43ca-a7c3-76f001f0b0b6

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!


lovincyrus:
Ready for UXQA @ericokuma

ericokuma:
**Snowflake feedback:**

[Rill Developer | snowflake_1.yaml - 4 September 2025](https://www.loom.com/share/4e3142775842472c9ed79abeec23de89?sid=e02f832b-7fa5-4b32-86c3-f78cddf4f47b)

lovincyrus:
> **Snowflake feedback:**
> 
> [Rill Developer | snowflake_1.yaml - 4 September 2025](https://www.loom.com/share/4e3142775842472c9ed79abeec23de89?sid=e02f832b-7fa5-4b32-86c3-f78cddf4f47b)

Addressed - Marking `database` as required in Snowflake based on [this check](https://github.com/rilldata/rill/blob/main/runtime/drivers/snowflake/snowflake.go#L156-L158)

web-common/src/features/sources/modal/constants.ts
```diff
@@ -11,24 +11,25 @@ export const CONNECTION_TAB_OPTIONS: { value: string; label: string }[] = [
   { value: "dsn", label: "Enter connection string" },
 ];
 
+export const OLAP_ENGINES = ["clickhouse", "druid", "pinot"];
+
+// pre-defined order for sources
 export const SOURCES = [
```
ericpgreen2, line 21: Did you intend to change the order? Previously we _loosely_ structured it as 1. Object storage, 2. Warehouses, 3. Databases, 4. Other. But I'm not wed to that order, I just want to make sure we pick an order with intention CC @ericokuma.

web-common/src/features/sources/modal/submitAddDataForm.ts
```diff
@@ -193,6 +191,25 @@ export async function submitAddConnectorForm(
         details: result.details,
       };
     }
+  } else {
+    // Poll for the non-OLAP connectors resource status
+    // If the resource status is not reconciled, rollback the changes
+    const result = await pollConnectorResource(
+      instanceId,
+      connector.name as string,
+    );
```
ericpgreen2: It looks like this `pollConnectorResource` function includes logic that's redundant to what exists above in this `submitAddConnectorForm` function. 

For example:
- `runtimeServicePutFileAndWaitForReconciliation` waits for reconciliation of the resource, so we shouldn't need the polling logic in `pollConnectorResource`, right?
- `fileArtifacts.checkFileErrors` checks for resource reconciliation errors, so we additionally shouldn't need to check again in `pollConnectorResource`, right?

ericpgreen2: In general, as much as possible, we should use the same code path for all connectors (both non-OLAP and OLAP).

ericpgreen2 (changes requested):

web-common/src/features/sources/modal/constants.ts
lovincyrus: Synced offline with @ericokuma to arrange them in ascending order and end with https + local_file

web-common/src/features/sources/modal/submitAddDataForm.ts
lovincyrus: I've updated the implementation to use the `ListDatabaseSchemas` API for the non-OLAP connectors.

ericpgreen2: As discussed in our check-in, for non-OLAP connectors, I don't think it's necessary to call `ListDatabaseSchemas` to test the connection client-side. Per [Naman's PR](https://github.com/rilldata/rill/pull/7426), the runtime's reconciliation process should be testing the connection server-side. You can look at that PR to see which connectors are covered by his work. For example, here's the Athena logic:

image

lovincyrus: Addressing

lovincyrus: Reverted the calling `ListDatabaseSchemas` API changes. I think we should still [use a polling](https://linear.app/rilldata/issue/APP-23/add-test-connection-feature-to-all-connectors-ui#comment-df1c7978) to check for the server's reconcile status. I tried the approach by just relying on `runtimeServicePutFileAndWaitForReconciliation`. It will continue to process the resource in the server side but it won't reflect the reconcile status changes to the UI.

lovincyrus: I retried and verified that `waitForProjectParserVersion` continues to create the connector file without verifying the resource status. To address this, we need [custom polling](https://github.com/rilldata/rill/pull/7867/commits/457ae5471f16049ed3c86148588f712fb0b93e77) to manually verify the reconcile status. Without it, users will only encounter the reconcile error within the editor.﻿

Custom polling was also an approach mentioned [here](https://linear.app/rilldata/issue/APP-23/add-test-connection-feature-to-all-connectors-ui#comment-df1c7978).

ericpgreen2: > To address this, we need [custom polling](https://github.com/rilldata/rill/pull/7867/commits/457ae5471f16049ed3c86148588f712fb0b93e77) to manually verify the reconcile status.

Ok, makes sense. However, can we use the same polling logic for both the non-OLAP connectors and the OLAP connectors? I.e. remove the `runtimeServicePutFileAndWaitForReconciliation` function. I'd like to avoid code branches if they're not absolutely necessary.

Further, as a nice-to-have, it'd be nice if your `runtimeServicePutConnectorFileAndWaitForReconciliation` function were not connector-specific but rather could be re-used for any resource. E.g. a function named `runtimeServicePutFileAndWaitForResourceReconciliation(...)`.

lovincyrus: We don't need to apply the same polling logic to OLAP connectors since we fire `ListTables` API for a response. Since they are on a separate code path, i'd very much prefer to isolate them. One is using ListTables API, another one requires custom polling to check for the resource's reconcile status.

---

Non-OLAP Connectors:
1. Create connector file with runtimeServicePutConnectorFileAndWaitForReconciliation
2. Wait for reconciliation + connection testing during reconciliation
3. Create .env file with runtimeServicePutFileAndWaitForReconciliation

OLAP Connectors:
1. Create connector file with simple runtimeServicePutFile
2. Create .env file with runtimeServicePutFileAndWaitForReconciliation
3. Let reconciliation happen async on server side
4. Test connection via ListTables API (testOLAPConnector)

lovincyrus:
> 2 lines export const line 20, 40 https://github.com/rilldata/rill/blob/8e8e2cb3e6280b73e285c4c65c5c066b650b0b7d/web-common/src/features/sources/modal/constants.ts

https://github.com/rilldata/rill/pull/7867/commits/02a56a3ee772391ae73550fb8c572f18c067e098

ericpgreen2 (changes requested):
I tried to connect a local Postgres instance, and this error makes it feel like something is wrong:

image

I believe it's because the code is currently testing the connection _before_ writing the environment variables to the `.env` file.

An e2e test would be helpful to give us confidence whether or not things are working as expected.

web-common/src/features/sources/modal/submitAddDataForm.ts
ericpgreen2: > Non-OLAP Connectors:
> 1. Create connector file with runtimeServicePutConnectorFileAndWaitForReconciliation
> 2. Wait for reconciliation + connection testing during reconciliation
> 3. Create .env file with runtimeServicePutFileAndWaitForReconciliation

As mentioned in my latest review, we'll need to create the `.env` file _before_ testing the connection.

ericpgreen2: In general, the `submitAddConnectorForm` currently requires a very careful read to understand the difference between OLAP connectors and non-OLAP connectors. To strive toward one consistent codepath, we could request that Platform tests the OLAP connectors server-side during reconciliation – just like they do for the non-OLAP connectors. Do you think that would help?

royendo (approved):
Experience pass: 👍 

Have some follow up, will create sep tickets

lovincyrus:
> I tried to connect a local Postgres instance, and this error makes it feel like something is wrong:
> 
> image
> I believe it's because the code is currently testing the connection _before_ writing the environment variables to the `.env` file.
> 
> An e2e test would be helpful to give us confidence whether or not things are working as expected.

Fixed and added a couple of e2e.

PR #7869, opened 2025-09-03
Adding pagination in ListDatabaseSchemas and ListTables

NamanMahor:
Adding Pagination API for ListDatabaseSchemas and ListTables for [PLAT-161: Support pagination in connector table listings](https://linear.app/rilldata/issue/PLAT-161/support-pagination-in-connector-table-listings)

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!

### DIFF
--- a/runtime/server/controller.go
+++ b/runtime/server/controller.go
@@ -321,14 +321,14 @@ func (s *Server) GetModelPartitions(ctx context.Context, req *runtimev1.GetModel
 	}
 	defer release()
 
-	_defaultPageSize := 100
+	defaultPageSize := 100
 	opts := &drivers.FindModelPartitionsOptions{
 		ModelID:          partitionsModelID,
 		WherePending:     req.Pending,
 		WhereErrored:     req.Errored,
 		BeforeExecutedOn: beforeExecutedOn,
 		AfterKey:         afterKey,
-		Limit:            pagination.ValidPageSize(req.PageSize, _defaultPageSize),
+		Limit:            pagination.ValidPageSize(req.PageSize, defaultPageSize),
 	}
 
 	partitions, err := catalog.FindModelPartitions(ctx, opts)
@@ -337,7 +337,7 @@ func (s *Server) GetModelPartitions(ctx context.Context, req *runtimev1.GetModel
 	}
 
 	var nextPageToken string
-	if len(partitions) == pagination.ValidPageSize(req.PageSize, _defaultPageSize) {
+	if len(partitions) == pagination.ValidPageSize(req.PageSize, defaultPageSize) {
 		last := partitions[len(partitions)-1]
 		nextPageToken = pagination.MarshalPageToken(last.Index, last.Key)
 	}


runtime/drivers/athena/information_schema.go
```diff
@@ -48,49 +54,66 @@ func (c *Connection) ListDatabaseSchemas(ctx context.Context) ([]*drivers.Databa
...
+		}
+	}
 	q := fmt.Sprintf(`
 	SELECT
 		table_name,
 		table_type
 	FROM %s.information_schema.tables
-	WHERE table_schema = %s
-	`, sqlSafeName(database), escapeStringValue(databaseSchema))
+	WHERE table_schema = %s
```
k-anshul: In this rests of the pagination implementation we use `WHERE table_name > %s LIMIT %s` constructs. 
1. This is somewhat more performant than `offset` because the database does not need to compute results to calculate offset but rather can directly seek to the results.
2. Will prevent skipping/duplicating rows if data is added or removed.

Also see `marshalPageToken` and `unmarshalPageToken` for reference.

runtime/drivers/athena/information_schema.go
```diff
@@ -210,6 +233,36 @@ func (c *Connection) listSchemasForCatalog(ctx context.Context, client *athena.C
 	return res, nil
 }
 
+func paginateSchemas(res []*drivers.DatabaseSchemaInfo, pageSize uint32, pageToken string) ([]*drivers.DatabaseSchemaInfo, string, error) {
```
k-anshul, line 234: Why can't we directly paginate in the query ?

runtime/drivers/snowflake/information_schema.go
```diff
@@ -44,28 +45,59 @@ func (c *connection) ListDatabaseSchemas(ctx context.Context) ([]*drivers.Databa
...
+	}
+	offset := 0
+	if pageToken != "" {
+		var err error
+		offset, err = strconv.Atoi(pageToken)
+		if err != nil {
+			return nil, "", fmt.Errorf("invalid page token: %w", err)
+		}
+	}
 	q := fmt.Sprintf(`
```
k-anshul, line 69: Same as athena

runtime/drivers/redshift/information_schema.go
```diff
@@ -11,99 +12,135 @@ import (
...
+	q := fmt.Sprintf(`
 	SELECT 
 		database_name, 
 		schema_name 
 	FROM svv_all_tables 
 	WHERE schema_name NOT IN ('information_schema', 'pg_catalog') OR schema_name = current_schema()
 	GROUP BY database_name, schema_name 
 	ORDER BY database_name, schema_name
-	`
+	LIMIT %d
```
k-anshul, line 34: Same comment as in other drivers.

runtime/drivers/postgres/information_schema.go
```diff
@@ -2,63 +2,96 @@ package postgres
...
 import (
 	"context"
+	"fmt"
+	"strconv"
 
 	"github.com/rilldata/rill/runtime/drivers"
 )
 
-func (c *connection) ListDatabaseSchemas(ctx context.Context) ([]*drivers.DatabaseSchemaInfo, error) {
+func (c *connection) ListDatabaseSchemas(ctx context.Context, pageSize uint32, pageToken string) ([]*drivers.DatabaseSchemaInfo, string, error) {
```
k-anshul, line 11: Same comment as in other drivers.

runtime/drivers/athena/information_schema.go
NamanMahor: @k-anshul we are doing multiple query in parallel and wait for completion in this case that why we have to do this

k-anshul: Okay sounds good.

runtime/server/controller.go
```diff
@@ -320,13 +321,14 @@ func (s *Server) GetModelPartitions(ctx context.Context, req *runtimev1.GetModel
 	}
 	defer release()
 
+	_defaultPageSize := 100
```
k-anshul: Nit: `defaultPageSize`. `_` is typically prefixed for private constants.

PR #7870, opened 2025-09-03
fix: total user count

mindspank:
Total user count from the payload instead of counting rows visible in the table.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


mindspank:
> If showing total users rather than filtered users is the intended behavior, then :shipit:

Ye I can see it both ways but folks were confused the number incremented as we scrolled the table. Maybe we can add " X total users - showing X users". What do you think?

lovincyrus:
> > If showing total users rather than filtered users is the intended behavior, then :shipit:
> 
> Ye I can see it both ways but folks were confused the number incremented as we scrolled the table. Maybe we can add " X total users - showing X users". What do you think?

I'd use totalUsers to display "X total users" and filteredUsers to display "X total users - showing X users"

PR #7875, opened 2025-09-03
Fix: Auto-convert Snowflake private keys to correct DSN format

ericokuma:
## Summary
- Fixes Snowflake DSN authentication failures when using private keys
- Automatically converts private keys to the required URL-safe base64 format without padding
- Resolves the issue reported in Slack where multiple customers encountered "error decoding private key"

## Problem
When using Snowflake DSN connections with keypair authentication, users were getting "error decoding private key" errors. The issue was that:
- Individual connection properties support various private key formats
- DSN connections require a specific format: DER-encoded PKCS#8 in URL-safe base64 **without padding**
- The documentation command (`cat rsa_key.p8 | base64 | tr '+/' '-_' | tr -d '\n'`) was incorrect for DSN usage

## Solution
This PR adds automatic private key format detection and conversion:

1. **Enhanced `parseRSAPrivateKey`**: Now supports `base64.RawURLEncoding` (URL-safe base64 without padding)
2. **New `fixDSNPrivateKey` function**: Intercepts DSN strings and automatically converts private keys
3. **Helper functions**: Convert between different key formats transparently

## Changes
- Added support for RawURLEncoding in the private key parser
- Added automatic DSN private key format conversion
- No breaking changes - existing connections continue to work

## Test plan
- [ ] Test with private key in standard base64 format
- [ ] Test with private key in URL-safe base64 with padding
- [ ] Test with private key in URL-safe base64 without padding (correct format)
- [ ] Test with raw PEM format
- [ ] Verify individual connection properties still work
- [ ] Verify DSN connections now work with various key formats

## Related Issues
- Slack thread: Multiple customers (Angel Studios, BrainForge) encountering this issue
- The correct DSN format requires: `openssl pkcs8 -topk8 -inform PEM -outform DER -in rsa_key.pem -nocrypt | base64 | tr '+/' '-_' | tr -d '='`

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/runtime/drivers/snowflake/snowflake.go
+++ b/runtime/drivers/snowflake/snowflake.go
@@ -150,7 +150,12 @@ func (c *configProperties) validate() error {
 
 func (c *configProperties) resolveDSN() (string, error) {
 	if c.DSN != "" {
-		return c.DSN, nil
+		// Parse and fix DSN if it contains a private key
+		fixedDSN, err := fixDSNPrivateKey(c.DSN)
+		if err != nil {
+			return "", fmt.Errorf("failed to parse DSN: %w", err)
+		}
+		return fixedDSN, nil
 	}
 
 	if c.Account == "" || c.User == "" || c.Database == "" {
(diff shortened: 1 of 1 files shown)


NamanMahor:
@ericokuma A DSN is typically expressed as a URL and must already be URL-safe. We should avoid additional manipulation. User should provide url-safe dsn

PR #7882, opened 2025-09-05
Updated deploy flow from CLI

k-anshul:
1. Implements restrictions from PRD : https://www.notion.so/rilldata/Rill-Deployment-and-Git-architecture-25bba33c8f5780e1b0f2ea74cc7c4b35#25bba33c8f5780e1b0f2ea74cc7c4b35
2. One notable change from PRD is that `__rill_remote` is only set for rill managed repositories. Setting it for user managed git repos leads to various edge cases.
3. Both rill managed and self managed projects now get matched with `git_remote`. The `directory_name` is still kept for matching archive projects which are used in testing.

Future work:
1. Update `ConnectProjectToGithub` API to mark repo as transfered.
2. Update `rill project connect-github` to migrate a project from rill managed project to self managed project similar to UI. Currently it throws an error.
3. Update `create org` flow to ask input from user similar to UI.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/cli/cmd/project/connect_github.go
+++ b/cli/cmd/project/connect_github.go
@@ -120,7 +120,7 @@ func ConnectGithubFlow(ctx context.Context, ch *cmdutil.Helper, opts *DeployOpts
 		return err
 	}
 	if !ok {
-		ch.PrintfBold("You can run `rill project connect-github` again when you have pushed your local changes to the remote.\n")
+		ch.PrintfBold("You can run `rill deploy` again when you have pushed your local changes to the remote.\n")
 		return nil
 	}
 


cli/cmd/project/deploy.go
```diff
@@ -66,58 +78,168 @@ func (o *DeployOpts) ValidatePathAndSetupGit(ch *cmdutil.Helper) error {
...
-	if err != nil && !errors.Is(err, gitutil.ErrGitRemoteNotFound) {
+	// extract remote and check if project already exists
+	err = o.detectGitRemoteAndProject(ctx, ch, repoRoot, subpath)
+	if err != nil {
 		return err
 	}
-	if remote.URL == "" {
-		// no remote configured
+
+	// if there is a project already connected to this repo+subpath offer to push changes to it
```
AdityaHegde, line 110: This check only seem to be on client side. Shouldn't there be an equivalent one in CreateProject/UpdateProject?

cli/cmd/project/deploy.go
```diff
@@ -66,58 +78,168 @@ func (o *DeployOpts) ValidatePathAndSetupGit(ch *cmdutil.Helper) error {
...
 	if subpath != "" {
 		ch.PrintfBold("Project location within repo: ")
 		ch.Printf("%s\n", subpath)
 	}
-	confirmed, err := cmdutil.ConfirmPrompt("Enable automatic deploys to Rill Cloud from GitHub?", "", true)
-	if err != nil {
-		return err
+	if o.Managed {
+		// if user explicitly wants managed deploys confirm if they want to really skip github connection
+		ok, err := cmdutil.ConfirmPrompt("Do you want to skip connecting to GitHub and use Rill managed deploys? (Note: Subsequent deploys/push from Rill will not push changes to your GitHub repo)", "", true)
```
AdityaHegde, line 161: We have opted to not allow this in UI to stop users from hitting edge cases. Do not want the same in cli? cc @ericokuma

admin/server/projects.go
```diff
@@ -155,20 +156,41 @@ func (s *Server) ListProjectsForFingerprint(ctx context.Context, req *adminv1.Li
...
 	if err != nil {
 		return nil, err
 	}
 
-	nextToken := ""
-	if len(projects) >= pageSize {
-		nextToken = marshalPageToken(projects[len(projects)-1].ID)
+	if len(projects) == 0 {
+		// if no project is found check if there is project user doesn't have access to
+		projects, err = s.admin.DB.FindProjectsByGitRemote(ctx, normalizeGitRemote(req.GitRemote))
```
AdityaHegde, line 177: Shouldnt there be a check to make sure `GitRemote` is not empty? A `ListProjectsForFingerprint` call with just `RillMgdGitRemote` and no matching projects would match every project without a git remote right?

admin/server/projects.go
```diff
@@ -155,20 +156,41 @@ func (s *Server) ListProjectsForFingerprint(ctx context.Context, req *adminv1.Li
...
+		for _, p := range projects {
+			if p.Subpath != req.SubPath {
+				continue
+			}
+			org, err := s.admin.DB.FindOrganization(ctx, p.OrganizationID)
+			if err != nil {
+				return nil, err
+			}
+			return &adminv1.ListProjectsForFingerprintResponse{
+				UnauthorizedProject: fmt.Sprintf("%s/%s", org.Name, p.Name),
```
AdityaHegde, line 190: Should we return the org and project name here? Or are there leaks from other places that this wont matter?

k-anshul: I was also somewhat split here. The good thing is that it gives a hint to the user that this is where this project is deployed. The bad thing is that it exposes org/project name.
I think a user will anyways not be able to know private/internal repo names so this will not matter. For public repos it seems okay to give the org/project details. But if you prefer can hide this detail.
We can also explore exposing UUIDs instead of names but also need to provide access to search for projects/orgs using UUIDs (given they have access to it).

cli/cmd/project/deploy.go
k-anshul: I think the option should still be available to the user to skip github connect. 

This is especially helpful for cases when users may not have access to install github repo in their org and they want to deploy a project. I remember this used to be blocker during POC calls where users will not have access to install repos. Though this was before we had managed project support and only option was to create a git repo.

I think it is fine to be somewhat opinionated on UI and take a call that if user has decided to create a git repo we should prompt the installation but CLI can allow such escape hatches. 
Please let me know if you guys think otherwise then I can remove this.

cli/cmd/project/deploy.go
k-anshul: Good point. I will skip the API checks and add a unique index on `git_remote, subpath` (Will confirm if such projects exist in production already so that index does not fail).

admin/server/projects.go
AdityaHegde: User can use this org/project to land on request access page. Since a lot of times the github org name and rill org name are same, most probably same with project name this feels ok to do.
UUID will have the same issue if there a search no? Feels unnecessary if so.

cli/cmd/project/deploy.go
ericokuma: If we allow this option to skip github connect when there's a self-managed git detected, then how exactly would a customer sync to that git repo in RC after the fact? Since we are not allowing users to Sync to existing repos in RC?

k-anshul: User will have an option to change remote of an existing project using `rill project edit` CLI command (will raise a PR for this soon).
If we want we can extend this to UI in future as well.

admin/server/projects.go
k-anshul: > Since a lot of times the github org name and rill org name are same, most probably same with project name this feels ok to do.

Yeah makes sense.

> UUID will have the same issue if there a search no? 

I mean we can prevent search from returning a name if the user does not have access. 

But I think it is fine to leave names as noted above.

cli/cmd/project/connect_github.go
```diff
@@ -69,122 +66,72 @@ func GitPushCmd(ch *cmdutil.Helper) *cobra.Command {
...
 	}
 
-	// We now have a gitRemote.
+	// Error if the repository is not in sync with the remote
+	ok, err := repoInSyncFlow(ch, localGitPath, opts.ProdBranch, opts.RemoteName)
+	if err != nil {
+		return err
+	}
+	if !ok {
+		ch.PrintfBold("You can run `rill project connect-github` again when you have pushed your local changes to the remote.\n")
```
AdityaHegde: Should we omit `rill project connect-github` or use `rill deploy` instead? I would assume `deploy` sub command is the most used one.

cli/cmd/project/deploy.go
```diff
@@ -66,58 +78,168 @@ func (o *DeployOpts) ValidatePathAndSetupGit(ch *cmdutil.Helper) error {
...
+		}
+		o.pushToProject = p
+		o.remoteURL = p.GitRemote
+		// do not return yet, there might be a managed project
+		// this is not possible with new flow but keeping it for consistency
+	}
+
+	if req.RillMgdGitRemote != "" {
+		// if no project found remove `__rill_remote`. This can happen if managed repo was provisioned but project was not created/failed.
+		err = removeRemote(repoRoot, "__rill_remote")
```
AdityaHegde, line 232: This doesnt fully work in a mono repo. While the remote is gone, `.git` remains. So repo detection fails after deploying a rill-managed project and deleting it. We should cleanup `.git` as well.

cli/cmd/project/connect_github.go
k-anshul: Since this is

cli/cmd/project/deploy.go
AdityaHegde: Honestly this is not a blocker for this PR.

AdityaHegde (approved):

PR #7885, opened 2025-09-05
chore(deps): upgrade arrow-go

zeroshade:
An unintentional breaking change was introduced into v18.4.1 of arrow-go which requires adding a new method to any concrete types implementing `array.RecordReader` called `RecordBatch`. Rather than letting people encounter the issue themselves, it made sense to help the small number of consumers that will actually be affected by it upgrade appropriately.

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!


k-anshul (approved):

k-anshul:
Hey @zeroshade 

Thank you so much for flagging and fixing the breaking change.
Apologies for delay in reviewing this.

PR #7890, opened 2025-09-05
Pin below the fold dimension filters

lovincyrus:
This PR implements a "below the fold" strategy for dimension filters, similar to the existing leaderboard pattern. This ensures that selected values always appear in filter results, even if they're not in the top 250 results returned by the main query

Closes https://linear.app/rilldata/issue/APP-339/dashboard-filter-bug

https://github.com/user-attachments/assets/47838547-edb1-48cc-b437-dd0b5e60c91a

Reviewer, you can:
1. Setting main query limit to "5"
2. Opening a dimension filter
3. Selecting 10+ values
4. Closing and reopening
5. Verifying all selected values appear

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-common/src/features/dashboards/filters/dimension-filters/dimension-filter-values.ts
+++ b/web-common/src/features/dashboards/filters/dimension-filters/dimension-filter-values.ts
@@ -4,7 +4,6 @@ import {
   createInExpression,
   createLikeExpression,
   createAndExpression,
-  sanitiseExpression,
 } from "@rilldata/web-common/features/dashboards/stores/filter-utils";
 import { queryClient } from "@rilldata/web-common/lib/svelte-query/globalQueryClient";
 import {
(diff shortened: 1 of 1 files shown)


web-common/src/features/dashboards/filters/dimension-filters/dimension-filter-values.ts
```diff
@@ -67,14 +72,64 @@ export function useDimensionSearch(
...
-  return getCompoundQuery(queries, (responses) => {
-    const values = responses
-      .filter((r) => !!r?.data)
-      .map((r) => r!.data!.map((i) => i[dimensionName]))
-      .flat();
-    const dedupedValues = new Set(values);
-    return [...dedupedValues] as string[];
-  });
+  // Below the fold query: Ensure selected values are included (for both InList and Select modes)
+  const belowFoldQueries =
```
AdityaHegde: Why do we need to query here? In leaderboard we need measure values so we needed the query, here we just need the values no?

lovincyrus: Fixed

PR #7892, opened 2025-09-08
fix: Move embedded dashboard navigation to route based from param based

AdityaHegde:
Navigation in embedded pages are based on url params. Different resources are switched in `+page.svelte`.

This moves it to route based with dedicated routes for resources and home.

Closes APP-324

**Checklist:**
- [x] Covered by tests
- [ ] I'm proud of this work!


web-admin/src/features/embeds/embed-store.ts
ericpgreen2, line 1: Can you please add documentation to describe this class?

web-admin/src/features/embeds/embed-store.ts
```diff
@@ -0,0 +1,25 @@
...
+    this._instance = new EmbedStore(url);
+  }
+
+  public static getInstance() {
+    return this._instance;
+  }
+
+  private constructor(url: URL) {
+    this.instanceId = url.searchParams.get("instance_id") ?? "";
+    this.runtimeHost = url.searchParams.get("runtime_host") ?? "";
```
ericpgreen2, line 35: Can you add error handling in the case that `instance_id` or `runtime_host` aren't found?

web-admin/src/features/embeds/TopNavigationBarEmbed.svelte
```diff
@@ -55,7 +53,6 @@
     if (!resource) {
       throw new Error(`Resource not found: ${name}`);
     }
-    dispatch("select-resource", resource.meta.name);
```
ericpgreen2, line 58: Now it seems like this `onSelectResource` function has no purpose. Can we delete it? Does the two-tiered navigation still work?

web-admin/src/features/dashboards/listing/DashboardsTable.svelte
```diff
@@ -125,13 +118,7 @@
       </svelte:fragment>
     </NoResourceCTA>
   {:else}
-    <Table
-      kind="dashboard"
-      data={dashboardsData}
-      {columns}
-      {columnVisibility}
-      on:click-row={handleClickRow}
```
ericpgreen2, line 133: It looks like we can now delete the dispatch of `click-row` from the `<Table />` component

web-admin/src/features/embeds/embed-store.ts
```diff
@@ -0,0 +1,25 @@
+export class EmbedStore {
+  public readonly instanceId: string;
+  public readonly runtimeHost: string;
+  public readonly accessToken: string;
+  public readonly navigationEnabled: boolean;
+
+  public readonly exploreSeen = new Set<string>();
```
ericpgreen2: Documentation for why we need this would particularly help too

ericpgreen2: Nit: `visitedExplores` might be slightly more clear (for one, we should use the plural "explores" not singular)

web-admin/src/routes/-/embed/explore/[name]/+page.ts
```diff
@@ -0,0 +1,22 @@
+import { EmbedStore } from "@rilldata/web-admin/features/embeds/embed-store.ts";
+import { EmbedStorageNamespacePrefix } from "@rilldata/web-admin/features/embeds/constants.ts";
+import { clearExploreSessionStore } from "@rilldata/web-common/features/dashboards/state-managers/loaders/explore-web-view-store.ts";
+
+export const load = ({ params }) => {
+  const exploreName = params.name;
+
+  const embedStore = EmbedStore.getInstance();
```
Comment (ericpgreen2) [inline web-admin/src/routes/-/embed/explore/[name]/+page.ts]: Instead of using `getInstance()`, perhaps it'd be more SvelteKit conventional to pass it down through the route context?

ericpgreen2 (approved):
Great cleanup! Approving with some comments.

PR #7897, opened 2025-09-08
rename settings_override to query_settings_override in docs

k-anshul:


runtime/parser/schema/project.schema.yaml
```diff
@@ -257,9 +257,9 @@ definitions:
             log_queries:
               type: boolean
               description: Controls whether to log raw SQL queries
-            settings_override:
+            query_settings_override:
               type: string
-              description: override the default settings used in queries. example `readonly = 1, session_timezone = 'UTC'`
+              description: "override the default settings used in queries. Changing the default settings can lead to incorrect query results and is generally not recommended. If you need to add settings, append them to the defaults - `cast_keep_nullable = 1, join_use_nulls = 1, session_timezone = 'UTC', prefer_global_in_and_join = 1, insert_distributed_sync = 1, <your additional settings>`"
```
NamanMahor, line 262: Not related to this. but can we raise the ticket so that user does not need to add all the defaults. because we might add new default in code in future and forgot to add them here.

NamanMahor (approved):

runtime/parser/schema/project.schema.yaml
k-anshul: Yeah agreed. That is already in my backlog.
Actually we had an internal use case for `query_settings_override` to run rill on clickhouse playground which does not allow modifying session settings. This was not supposed to be exposed but it seems it is in docs now and some users have picked this up so added more clarifications. 

Will add `additional_query_settings` for adding extra settings.

PR #7906, opened 2025-09-08
[APP-33] fix: add delete confirmation dialog to canvas

briangregoryholmes:
The original issue is not practical to fully solve in the short term as it is an issue with an older version of a library we are using that can't be updated until we migrate to Svelte 5. This PR puts in a tangential fix by displaying a confirmation dialog when trying to delete a component/widget with the backspace/delete key.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


ericokuma (approved):
LGTM!

PR #7912, opened 2025-09-09
fix: show/hide dropdown not working as expected when filtered

AdityaHegde:
Hide action in show/hide dropdown for measures and dimensions is based on index from the filtered array, but uses the index on the original array. So when there is a filter this breaks by hiding incorrect item.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


lovincyrus (approved):
tested locally and +1

PR #7925, opened 2025-09-10
adding pagination api in olap list tables

NamanMahor:
Pagination in Olap list table api [PLAT-161: Support pagination in connector table listings](https://linear.app/rilldata/issue/PLAT-161/support-pagination-in-connector-table-listings)

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!

### DIFF
--- a/runtime/pkg/rduckdb/information_schema.go
+++ b/runtime/pkg/rduckdb/information_schema.go
@@ -43,8 +43,12 @@ func (d *db) Schema(ctx context.Context, ilike, name string, pageSize uint32, pa
 
 	// Add pagination filter
 	if pageToken != "" {
+		var startAfterName string
+		if err := pagination.UnmarshalPageToken(pageToken, &startAfterName); err != nil {
+			return nil, "", fmt.Errorf("invalid page token: %w", err)
+		}
 		whereClause += " AND t.table_name > ?"
-		args = append(args, pageToken)
+		args = append(args, startAfterName)
 	}
 
 	// the database schema changes with every ingestion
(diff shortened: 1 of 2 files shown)


runtime/drivers/snowflake/information_schema.go
```diff
@@ -58,9 +58,19 @@ func (c *connection) ListDatabaseSchemas(ctx context.Context, pageSize uint32, p
 	}
 	end := start + limit
 	if end >= len(res) {
-		return res[start:], "", rows.Err()
+		end = len(res)
 	}
-	return res[start:end], fmt.Sprintf("%d", end), rows.Err()
+
+	if start >= len(res) {
+		return []*drivers.DatabaseSchemaInfo{}, "", rows.Err()
```
k-anshul: It will probably be cleaner to check for `rows.Err()` after the for loop.

runtime/pkg/rduckdb/information_schema.go
```diff
@@ -57,13 +66,17 @@ func (d *db) Schema(ctx context.Context, ilike, name string) ([]*Table, error) {
 			AND t.table_schema = current_schema()
 			%s
 		GROUP BY ALL
-		ORDER BY ALL
+		ORDER BY t.table_name
+		LIMIT ?
```
k-anshul, line 74: Isn't this problematic that this limit is applied after join?
Shouldn't this be similar to clickhouse where limit is applied before join ?

NamanMahor: > Isn't this problematic that this limit is applied after join? Shouldn't this be similar to clickhouse where limit is applied before join ?

Test are working fine, it is not required because we are doing group by all and it have arra_agg for cols

runtime/drivers/snowflake/information_schema.go
NamanMahor: sure.

PR #7929, opened 2025-09-10
fix: fill missing implementation

briangregoryholmes:
Setting fillMissing to true triggers a path in the platform code that checks that there is exactly one time dimension being used for the time spine. When creating a chart _without_ a time dimension, this results in a 500 error being returned. This PR only sets fillMissing true when using a time dimension for the x axis.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-common/src/features/canvas/components/charts/cartesian-charts/CartesianChart.ts
+++ b/web-common/src/features/canvas/components/charts/cartesian-charts/CartesianChart.ts
@@ -286,7 +286,6 @@ export class CartesianChartComponent extends BaseChart<CartesianChartSpec> {
             where: topNWhere,
             timeRange,
             limit: limit?.toString(),
-            // fillMissing: config.x?.type === "temporal",
           },
           {
             query: {


web-common/src/features/canvas/components/charts/cartesian-charts/CartesianChart.ts
```diff
@@ -415,7 +414,7 @@ export class CartesianChartComponent extends BaseChart<CartesianChartSpec> {
             sort: xAxisSort ? [xAxisSort] : undefined,
             where: combinedWhere,
             timeRange,
-            fillMissing: true,
+            fillMissing: config.x?.type === "temporal" ? true : false,
```
djbarnwal: Ternary operator not needed here

web-common/src/features/canvas/components/charts/cartesian-charts/CartesianChart.ts
```diff
@@ -286,7 +286,7 @@ export class CartesianChartComponent extends BaseChart<CartesianChartSpec> {
             where: topNWhere,
             timeRange,
             limit: limit?.toString(),
-            fillMissing: true,
+            fillMissing: config.x?.type === "temporal" ? true : false,
```
djbarnwal: Ternary operator not needed here

web-common/src/features/canvas/components/charts/cartesian-charts/CartesianChart.ts
```diff
@@ -286,7 +286,7 @@ export class CartesianChartComponent extends BaseChart<CartesianChartSpec> {
             where: topNWhere,
             timeRange,
             limit: limit?.toString(),
-            fillMissing: true,
+            fillMissing: config.x?.type === "temporal",
```
djbarnwal: This is not needed as `topNXQuery` is only enabled for nominal fields.

djbarnwal:
I am also seeing another error -

```
{
    "code": 2,
    "message": "spine time range is only supported with a single time dimension",
    "details": []
}
```

We need to disable time spine when there is a color dimension attached. So for most parts, time spine will only work for a chart where there is no "split by" dimension. We also need to account for that.

briangregoryholmes:
How are you getting it into an error state? Situation below is what needs to be accounted for. Works fine for me?

That's the original error this PR was trying to resolve.

Screenshot 2025-09-10 at 11 21 49 PM

djbarnwal:
> How are you getting it into an error state? Situation below is what needs to be accounted for. Works fine for me?
> 
> That's the original error this PR was trying to resolve.

Ah yes, I checked again seems to be working fine. Probably a bad state.

PR #7932, opened 2025-09-11
feat: return a reconcile error when model execution is disabled for external clickhouse in read mode

k-anshul:
**Checklist:**
- [x] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/runtime/drivers/drivers.go
+++ b/runtime/drivers/drivers.go
@@ -22,8 +22,6 @@ var ErrStorageLimitExceeded = fmt.Errorf("connectors: exceeds storage limit")
 // ErrNotNotifier indicates the driver cannot be used as a Notifier.
 var ErrNotNotifier = errors.New("driver: not a notifier")
 
-var ErrCannotExecuteModels = errors.New("driver: connector is not capable of executing models")
-
 // Drivers is a registry of drivers.
 var Drivers = make(map[string]Driver)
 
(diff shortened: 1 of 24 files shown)


runtime/drivers/drivers.go
```diff
@@ -22,6 +22,8 @@ var ErrStorageLimitExceeded = fmt.Errorf("connectors: exceeds storage limit")
 // ErrNotNotifier indicates the driver cannot be used as a Notifier.
 var ErrNotNotifier = errors.New("driver: not a notifier")
 
+var ErrCannotExecuteModels = errors.New("driver: connector is not capable of executing models")
```
begelundmuller: Since this is returned from `AsModelExecutor`, consider naming it `ErrNotModelExecutor`. However, see comment above – do we actually need a separate error or could we just use `ErrNotImplemented`?

runtime/drivers/clickhouse/clickhouse.go
```diff
@@ -504,24 +504,23 @@ func (c *Connection) AsObjectStore() (drivers.ObjectStore, bool) {
...
 	if opts.InputHandle.Driver() == "s3" || opts.InputHandle.Driver() == "gcs" {
-		return &objectStoreToSelfExecutor{opts.InputHandle, c}, true
+		return &objectStoreToSelfExecutor{opts.InputHandle, c}, nil
 	}
 	if opts.InputHandle.Driver() == "local_file" {
-		return &localFileToSelfExecutor{opts.InputHandle, c}, true
+		return &localFileToSelfExecutor{opts.InputHandle, c}, nil
 	}
-	return nil, false
+	return nil, drivers.ErrNotImplemented
```
begelundmuller, line 523: Some places it uses `drivers.ErrCannotExecuteModels`, other places `drivers.ErrNotImplemented`. Is there a difference?

k-anshul: Bad oversight.

runtime/drivers/drivers.go
k-anshul: Yeah I can use `ErrNotImplemented`

PR #7943, opened 2025-09-12
feat: GCS glob resolver to work with HMAC keys as well

k-anshul:
closes https://linear.app/rilldata/issue/PLAT-195/gcs-glob-resolver-should-also-work-with-hmac-keys

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/runtime/drivers/gcs/gcs.go
+++ b/runtime/drivers/gcs/gcs.go
@@ -139,9 +139,6 @@ type Connection struct {
 	config  *ConfigProperties
 	storage *storage.Client
 	logger  *zap.Logger
-
-	// s3Conn is used when connecting to GCS in S3 compatible mode
-	s3Conn drivers.ObjectStore
 }
 
 var _ drivers.Handle = &Connection{}
--- a/runtime/drivers/gcs/api.go
+++ b/runtime/drivers/gcs/api.go
@@ -16,8 +16,6 @@ import (
 
 const defaultPageSize = 20
 
-// TODO: These APIs are not used right now.
-// Transfer these functions to drivers.ObjectStore interface when needed and use gcs/s3compatible connection accordingly.
 func (c *Connection) ListBuckets(ctx context.Context, req *runtimev1.GCSListBucketsRequest) ([]string, string, error) {
 	credentials, err := gcputil.Credentials(ctx, c.config.SecretJSON, c.config.AllowHostAccess)
 	if err != nil {


runtime/drivers/gcs/gcs.go
```diff
@@ -86,6 +87,35 @@ func (d driver) Open(instanceID string, config map[string]any, st *storage.Clien
...
+		if !ok {
+			return nil, fmt.Errorf("internal error: s3 compatible connection does not implement object store")
+		}
+		conn := &Connection{
+			config:  conf,
+			storage: st,
+			logger:  logger,
+			s3Conn:  store,
+		}
+		return conn, nil
```
NamanMahor, line 115: HI @k-anshul 
Adding my point of view here. 
if we could return s3 handler with endpoint and key, secret . we can clean lots of code in. gcs manager, object store executor, gcs ping(which i need to implement in soon for hmac key and value). 
this might need some change in config(need to make it common).

k-anshul: I actually evaluated that as well. 
The problem is that returning a `s3` connection means we loose the context completely that it actually talks to a GCS service. Having the original driver name is helpful in certain cases like [here](https://github.com/rilldata/rill/blob/05d55bd01bd9abef935734f5d8f02d16e57cfe89/runtime/drivers/duckdb/model_executor_objectstore_self.go#L178-L179)

> we can clean lots of code in. gcs manager, object store executor, gcs ping(

Given we are now wrapping s3 under GCS wouldn't we just be able to call `Ping` and other methods directly on the `s3Conn` like 
```
if c.s3Conn != nil {
    return c.s3Conn.Ping(ctx)
}
```

It is still not a strong opinion and I was also somewhat split on this. If you/@begelundmuller think that directly returning a s3 connection is better than I can also evaluate that and maybe fine workarounds for the issue I describe above (May just need a external service name to be set on the s3 driver).

NamanMahor: we could use endpoint == "storage.googleapis.com" we can also extend this for R2 Cloudflare too.

runtime/drivers/gcs/object_store.go
```diff
@@ -12,6 +13,9 @@ import (
 
 // ListObjects implements drivers.ObjectStore.
 func (c *Connection) ListObjects(ctx context.Context, path string) ([]drivers.ObjectStoreEntry, error) {
+	if c.s3Conn != nil {
+		return c.s3Conn.ListObjects(ctx, rewriteToS3Path(path))
+	}
```
begelundmuller: I'm worried this is easy to forgot if we make changes or add new functions to the interface. 

This may have too any negative side effects, but did you consider creating an `s3CompatibleConnection` in this package that uses the S3 conn? Then we'd get type errors if it didn't implement the required interfaces. 

The alternative of returning the S3 driver that you discussed above also makes sense to me, but agree that the entanglement/context loss might be annoying. As Naman points out, it could make sense to check `endpoint == "storage.googleapis.com"` inside the S3 driver to know if the backend is GCS (or R2) and adapt errors and paths accordingly.

I don't have a strong opinion on which path here, what do you think? But preventing individual case management in each interface function seems like it would be nice.

k-anshul: I evaluated returning a s3 connection directly but saw following cons:
1. For configurations where both credentials JSON and keys are present returning s3 connection means native credentials can not be used at all.
2. For these configuration we need to handle two cases for GCS in model executors when driver is s3 and endpoint is `storage.googleapis.com` and driver is GCS then extract key_id and secret.
3. If duckdb/clickhouse starts supporting native credentials then returning s3 connection means we will not be able to support native creds at all.
4. Though not tested but there are chances that connectors listed for the project returns s3 and not GCS. I user would expect to see GCS when configuring `driver: gcs` and s3 when configuring `driver: s3, endpoint: ...`

I implemented a `s3CompatibleConnection` in GCS instead that is just a wrapper over s3 and calls s3 APIs. Please let me know what do you think.

cc : @NamanMahor

runtime/drivers/gcs/s3compatible.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+// AsModelExecutor implements drivers.Handle.
+func (s *s3CompatibleConn) AsModelExecutor(instanceID string, opts *drivers.ModelExecutorOptions) (drivers.ModelExecutor, error) {
+	return nil, drivers.ErrNotImplemented
+}
+
+// AsModelManager implements drivers.Handle.
+func (s *s3CompatibleConn) AsModelManager(instanceID string) (drivers.ModelManager, bool) {
+	return nil, false
```
NamanMahor:
```suggestion
	return s, true
```

runtime/drivers/gcs/s3compatible.go
```diff
@@ -0,0 +1,164 @@
...
+func (s *s3CompatibleConn) ListObjects(ctx context.Context, path string) ([]drivers.ObjectStoreEntry, error) {
+	return s.s3Conn.ListObjects(ctx, rewriteToS3Path(path))
+}
+
+func rewriteToS3Path(s string) string {
+	if after, ok := strings.CutPrefix(s, "gs://"); ok {
+		return "s3://" + after
+	}
+	if after, ok := strings.CutPrefix(s, "gcs://"); ok {
+		return "s3://" + after
```
NamanMahor: not sure this is required. because in s3 api we usually pass bucket name after path parsing. so schema might not being used any where.

NamanMahor: if above is not required then only the that we are doing here is Driver() returning "gcs" which i think we can be done in s3 driver it self based on endpoint ?

runtime/drivers/gcs/gcs.go
```diff
@@ -86,6 +87,35 @@ func (d driver) Open(instanceID string, config map[string]any, st *storage.Clien
 		return nil, err
 	}
 
+	if conf.SecretJSON == "" && conf.KeyID != "" && conf.Secret != "" {
+		// open s3 connection to be used in case of S3 compatible mode
+		s3Config := s3.ConfigProperties{
+			AccessKeyID:     conf.KeyID,
+			SecretAccessKey: conf.Secret,
+			Endpoint:        "storage.googleapis.com",
```
NamanMahor, line 95: also should we pass allow host accesss false here.

k-anshul: Good catch. It should ideally pass whatever is set in GCS configs.

runtime/drivers/gcs/s3compatible.go
k-anshul: There is a check for schema in s3 here : https://github.com/rilldata/rill/blob/8b8bff0a353ab2a43d77b215851e4bcad96e1f01/runtime/drivers/s3/object_store.go#L80-L82

k-anshul: I wouldn't say returning s3 driver directly is not a viable option but as noted in point 2 here : https://github.com/rilldata/rill/pull/7943#discussion_r2374451872
We will need to do duplicate handling for GCS and s3 as GCS. Sadly we choose different config properties for s3 and GCS as s3.

runtime/drivers/gcs/gcs.go
```diff
@@ -110,6 +141,9 @@ type Connection struct {
 	config  *ConfigProperties
 	storage *storage.Client
 	logger  *zap.Logger
+
+	// s3Conn is used when connecting to GCS in S3 compatible mode
+	s3Conn drivers.ObjectStore
```
NamanMahor: not getting used any where

runtime/drivers/gcs/s3compatible.go
```diff
@@ -0,0 +1,164 @@
...
+	"strings"
+
+	"github.com/mitchellh/mapstructure"
+	"github.com/rilldata/rill/runtime/drivers"
+	"github.com/rilldata/rill/runtime/drivers/s3"
+)
+
+// s3CompatibleConn is a wrapper over s3.Connection.
+// It is implemented to provide operations for GCS driver that is configured using S3 compatible credentials only.
+type s3CompatibleConn struct {
```
NamanMahor, line 9: Not being used anywhere.

runtime/drivers/gcs/s3compatible.go
NamanMahor: can we remove s3 schema check?

Also I meant 
```
type GCSHMACConnection struct {
    *Connection
}

func (g *GCSHMACConnection) Driver() string {
    return "gcs"
}
```
```
gcsConn := &GCSConnection{Connection: s3conn}
```

NamanMahor (changes requested):

runtime/drivers/gcs/s3compatible.go
k-anshul: Ahh thanks. I missed pushing a stashed change :)

runtime/drivers/gcs/s3compatible.go
k-anshul: > can we remove s3 schema check?

Maybe not. A lot of driver inference logic depends on `s3` scheme being set in URIs. Given that is an existing check I wouldn't want to remove it unless there is a specific reason.

k-anshul: > Also I meant

Do you expect to create that in s3 driver ?
How will the logic here work in cases when both native credentials and HMAC keys are set for GCS ? https://github.com/rilldata/rill/blob/4178526715055c7f59e59ada3ac11e9591a87abf/runtime/drivers/clickhouse/model_executor_objectstore_self.go#L128-L149

PR #7944, opened 2025-09-12
Admin API organization

grahamplata:
Updates to openapi protos for admin and runtime resources

**Checklist:**
- [x] Covered by tests
- [x] Linked the issues it closes - [Harmonize the usage of organization and orgs across API paths](https://linear.app/rilldata/issue/PLAT-150/api-clean-up)
- [ ] I'm proud of this work!


mindspank:
From a product pov, looks good

proto/rill/admin/v1/api.proto
```diff
@@ -545,15 +545,15 @@ service AdminService {
...
   }
 
   // GetIFrame returns the iframe URL for the given project
   rpc GetIFrame(GetIFrameRequest) returns (GetIFrameResponse) {
     option (google.api.http) = {
-      post: "/v1/organizations/{organization}/projects/{project}/iframe",
+      post: "/v1/orgs/{organization}/projects/{project}/iframe",
       body: "*"
     };
   }
```
begelundmuller, line 600: This API is currently already documented in our docs. It's still good to rename, but can you add a manual alias in `admin/server/server.go` from the old path to the new path?

proto/rill/admin/v1/api.proto
```diff
@@ -1015,7 +1015,7 @@ message ListOrganizationsResponse {
 }
 
 message GetOrganizationRequest {
-  string name = 1;
+  string organization = 1;
```
begelundmuller: If we're using `orgs` in the path, shouldn't we name all these params `org` (instead of `organization`)?

proto/rill/runtime/v1/api.proto
```diff
@@ -190,6 +190,29 @@ service RuntimeService {
...
+      body: "*"
+    };
+  }
+
+  // DeleteResource deletes a resource from the catalog
+  rpc DeleteResource(DeleteResourceRequest) returns (DeleteResourceResponse) {
+    option (google.api.http) = {
+      delete: "/v1/instances/{instance_id}/resources/{kind}/{name}"
+    };
+  }
```
begelundmuller: What's the motivation for these RPCs? It's not really safe to support outright creation/edits of resources, all edits have to go through YAML files, which internally are reconciled with the catalog.

begelundmuller (changes requested):
Try searching the frontend code and docs for `/v1/organizations/` – it needs to be updated in a few places

grahamplata:
Screenshot 2025-09-30 at 12 14 08 PM

docs/docs/guides/api-usage.md
```diff
@@ -0,0 +1,307 @@
+---
```
begelundmuller: 1. Some of the paths in this docs page are broken, e.g. still uses the old `/v1/organizations` prefix
2. Can you raise this docs page as a separate PR? I would like to merge the refactors here ASAP since otherwise we'll get lots of merge conflicts with a diff this big.

proto/rill/admin/v1/api.proto
```diff
@@ -2,131 +2,177 @@ syntax = "proto3";
...
   // Ping returns information about the server
   rpc Ping(PingRequest) returns (PingResponse) {
+    option (google.api.method_visibility).restriction = "INTERNAL";
     option (google.api.http) = {get: "/v1/ping"};
   }
 
   // ListOrganizations lists all the organizations currently managed by the admin
   rpc ListOrganizations(ListOrganizationsRequest) returns (ListOrganizationsResponse) {
-    option (google.api.http) = {get: "/v1/organizations"};
+    option (google.api.method_visibility).restriction = "PUBLIC";
```
begelundmuller, line 45: What is the default visibility? It would be nice if we only need to mark internal ones, not public ones (or vice versa)

begelundmuller (commented):
I would like to merge this ASAP to avoid too many merge conflicts for others, which gets very likely with a diff this big.

Can you pull the changes in `docs` (including the auto-generated docs) out into a separate PR? I think we'll want to polish that some more before rolling it out. 

Then we can merge this PR now.

PR #7945, opened 2025-09-12
Use unique env key based on connector instance name

lovincyrus:
This PR uses connector instance names instead of driver names for environment variable keys, ensuring each connector gets unique credentials. Fixes https://linear.app/rilldata/issue/APP-273/make-sure-that-unique-passwords-are-stored-as-separate-options-in-env

CleanShot 2025-09-12 at 15 30 37@2x

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!

### DIFF
--- a/web-common/src/features/connectors/code-utils.ts
+++ b/web-common/src/features/connectors/code-utils.ts
@@ -222,13 +222,13 @@ export function deleteEnvVariable(
 }
 
 export function makeDotEnvConnectorKey(
-  connectorName: string,
+  driverName: string,
   key: string,
   connectorInstanceName?: string,
 ) {
-  // Note: The connector instance name is used when provided, otherwise fall back to connector name.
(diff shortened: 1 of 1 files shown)


royendo:
did a pass through, looks good (specifically for .env naming)

royendo (approved):
FR: will doc on linear, removing connectors rmeoves the corresponding .env? or at least prompts

lovincyrus:
> FR: will doc on linear, removing connectors rmeoves the corresponding .env? or at least prompts

We do not currently track "removing connectors rmeoves the corresponding .env"

ericpgreen2 (approved):
Approving as this is a good incremental improvement. A follow-up may be to implement [Benjamin's proposal here](https://rilldata.slack.com/archives/C088Q884YF7/p1755532502291189).

PR #7947, opened 2025-09-15
change docs for special character in password and handler issue in special character handling.

NamanMahor:
[PLAT-183: Document expectations around special characters in MySQL/Postgres passwords](https://linear.app/rilldata/issue/PLAT-183/document-expectations-around-special-characters-in-mysqlpostgres)

1. Adding documentation for postgres and mysql.
2. Fixed port decoding issue in duckdb model for mysql and postgres by change Decode to WeakDecode
3. Changed mysql `ssl_mode` to `ss-mode` matching to official documentation and now it is same as in dsn param
4. Changed `mysql.Config`to `mysql.NewConfig` because `mysql.Config ` was adding some default properties which was causing issue to foodbot.
5. added extra encoding handling in individual properties which was causing parsing issue in duckdb.
6. Raised the [PR in duckdb mysql](https://github.com/duckdb/duckdb-mysql/pull/159) because some of the character decoding is not proper in duckdb-mysql 

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!

### DIFF
--- a/runtime/drivers/mysql/mysql.go
+++ b/runtime/drivers/mysql/mysql.go
@@ -109,6 +109,15 @@ func (c *ConfigProperties) ResolveDSN() (string, error) {
 		return c.DSN, nil
 	}
 
+	var userInfo *url.Userinfo
+	if c.User != "" {
+		if c.Password != "" {
+			userInfo = url.UserPassword(c.User, c.Password)
+		} else {
+			userInfo = url.User(c.User)
+		}
+	}
+
 	host := c.Host
 	if host == "" {
 		host = "localhost"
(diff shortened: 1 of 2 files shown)


runtime/drivers/mysql/mysql.go
```diff
@@ -79,11 +79,11 @@ var spec = drivers.Spec{
 			Hint:        "Name of the MySQL database to connect to",
 		},
 		{
-			Key:         "sslmode",
+			Key:         "ssl-mode",
```
k-anshul, line 82: Please verify there are no existing projects using this if not done already

PR #7948, opened 2025-09-15
[PLAT-187] restrict using same name as time col in ch

pjain1:
Although it won't be an issue because of [this](https://github.com/rilldata/rill/pull/7700), but just restrict it to prevent query rewrites and thus keeping query simple. Note that if primary time dim defined under `timeseries` key then it will still cause query rewriting.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/runtime/metricsview/executor_validate.go
+++ b/runtime/metricsview/executor_validate.go
@@ -152,6 +152,16 @@ func (e *Executor) ValidateAndNormalizeMetricsView(ctx context.Context) (*Valida
 		}
 	}
 
+	// now we have datatypes for dims populated, check time/date types for clickhouse
+	if e.olap.Dialect() == drivers.DialectClickHouse {
+		for _, d := range mv.Dimensions {
+			if d.DataType != nil && (d.DataType.Code == runtimev1.Type_CODE_TIMESTAMP || d.DataType.Code == runtimev1.Type_CODE_DATE) && strings.EqualFold(d.Name, d.Column) {
+				res.OtherErrs = append(res.OtherErrs, fmt.Errorf("invalid dimension %q: timestamp or date type dimensions cannot have the same name as a column in the underlying table when backed by clickhouse", d.Name))
+				continue
+			}
+		}
+	}
+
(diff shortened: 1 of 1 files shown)


runtime/metricsview/executor_validate.go
```diff
@@ -152,6 +152,16 @@ func (e *Executor) ValidateAndNormalizeMetricsView(ctx context.Context) (*Valida
 		}
 	}
 
+	// now we have datatypes for dims populated, check time/date types for clickhouse
+	if e.olap.Dialect() == drivers.DialectClickHouse {
+		for _, d := range mv.Dimensions {
+			if d.DataType != nil && (d.DataType.Code == runtimev1.Type_CODE_TIMESTAMP || d.DataType.Code == runtimev1.Type_CODE_DATE) && strings.EqualFold(d.Name, d.Column) {
+				res.OtherErrs = append(res.OtherErrs, fmt.Errorf("invalid dimension %q: timestamp or date type dimensions cannot have the same name as a column in the underlying table when backed by clickhouse", d.Name))
```
begelundmuller, line 159: I thought we didn't need this check anyway? 

The behavior of `timeseries: name` is to use the same `name` and `column`, so it feels problematic to not support that for extra time dimensions.

pjain1: Yeah we don't need it, we create a unique name in case there is a name clash. We cannot put this restriction for timeseries key as it will break existing projects but going forward we can try to avoid this clash if user defines time dimension or AI generates the mv file. This way we will avoid query rewriting, making queries simpler. This is my thought, but let me know if you think its confusing.

begelundmuller: I'm worried this is too big an intrusion in the developer experience. I think people expect they can just add a column as a dimension and not worry further about that. So probably better to stay with the internal rewrites for now (hopefully Clickhouse fixes the underlying issue here at some point).

pjain1: Ok makes sense, no issues.

PR #7961, opened 2025-09-16
feat: add templating to project feature flags

AdityaHegde:
Adds support for templating in project's feature flag.
```
features:
  dashboardChat: '{{not (.user.embed)}}'
  reports: |
    '{{.user.domain}}' = 'rilldata.com'
```

Closes PLAT-197

**Checklist:**
- [x] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/runtime/drivers/registry.go
+++ b/runtime/drivers/registry.go
@@ -208,7 +208,7 @@ var DefaultFeatureFlags = map[string]string{
 	// Controls visibility of theme switching between light/dark modes
 	"dark_mode": "false",
 	// Controls visibility of project-level chat functionality
-	"chat": "true",
+	"chat": "false",
 	// Controls visibility of dashboard-level chat functionality
-	"dashboard_chat": "true",
+	"dashboard_chat": "false",
 }


runtime/drivers/registry.go
```diff
@@ -184,3 +184,32 @@ func (i *Instance) Config() (InstanceConfig, error) {
 
 	return res, nil
 }
+
+var defaultFeatureFlags = map[string]string{
```
mindspank: Missing dark mode, https://github.com/rilldata/rill/blob/main/web-common/src/features/feature-flags.ts#L59

admin/server/deployment.go
```diff
@@ -693,6 +693,12 @@ func (s *Server) GetIFrame(ctx context.Context, req *adminv1.GetIFrameRequest) (
 		}
 	}
 
+	// Mark the token as embed. Can be used in security policy or feature flags by `.user.embed`
+	if attr == nil {
+		attr = make(map[string]any)
+	}
+	attr["embed"] = true
```
begelundmuller: Probably only add it if it's not already in `attr`. So any user-provided custom attribute takes precedence.

runtime/drivers/registry.go
```diff
@@ -184,3 +184,33 @@ func (i *Instance) Config() (InstanceConfig, error) {
...
+	"dimensionSearch":     "false",
+	"twoTieredNavigation": "false",
+	"rillTime":            "true",
+	"hidePublicUrl":       "{{.user.embed}}",
+	"exportHeader":        "false",
+	"alerts":              "true",
+	"reports":             "true",
+	"darkMode":            "false",
+	"chat":                "true",
+	"dashboardChat":       "true",
```
begelundmuller: All our YAML syntax uses snake_case except for these feature flags. While we're making changes here, how do you feel about changing these to snake_case? For backwards compatibility, we can add a little mapper in the parser that does automatic `camelCase` to `snake_case` conversion.

runtime/drivers/registry.go
```diff
@@ -184,3 +184,33 @@ func (i *Instance) Config() (InstanceConfig, error) {
...
+	}
+	for f, v := range defaultFeatureFlags {
+		if _, ok := flags[f]; ok {
+			continue
+		}
+		flags[f] = v
+	}
+
+	i.FeatureFlags = flags
+}
```
begelundmuller: I think this means they'll be stored in the DB? It would be better to apply the defaults at resolve time so we don't have potential mismatch of defaults in code vs. defaults previously captured in the DB

runtime/drivers/sqlite/registry.go
```diff
@@ -114,9 +114,17 @@ func (c *connection) findInstances(_ context.Context, whereClause string, args .
 			return nil, err
 		}
 
-		i.FeatureFlags, err = mapFromJSON[bool](featureFlags)
+		featureFlagBools, err := mapFromJSON[bool](featureFlags)
```
begelundmuller, line 120: Can you add a comment that this is for backwards compatibility (previously stored as bools, now stored as strings)

runtime/registry.go
```diff
@@ -148,6 +149,41 @@ func (r *Runtime) TempDir(instanceID string, elem ...string) (string, error) {
...
+	featureFlags := make(map[string]bool)
+	for f, v := range inst.FeatureFlags {
+		rv, err := parser.ResolveTemplate(v, templateData, false)
+		if err != nil {
+			return nil, fmt.Errorf("failed to resolve feature flag %q with template %q: %w", f, v, err)
+		}
+		rv = strings.TrimSpace(rv)
+		if rv == "" || rv == "<no value>" {
+			continue
+		}
```
begelundmuller, line 187: Should it explicitly set it to `false` in this case? So all flags either become `true` or `false`, and we don't also have to handle an `undefined` case

runtime/runtime.go
```diff
@@ -174,7 +174,7 @@ func (r *Runtime) UpdateInstanceWithRillYAML(ctx context.Context, instanceID str
 		vars[k] = v
 	}
 	inst.ProjectVariables = vars
-	inst.FeatureFlags = rillYAML.FeatureFlags
+	inst.SetFeatureFlags(rillYAML.FeatureFlags)
```
begelundmuller: Just echoing my previous comment – I think it might be better to keep this as before, and apply the defaults at resolve time

begelundmuller (changes requested):

runtime/drivers/registry.go
AdityaHegde: Hmmm I think this should be ok. We need to convert back to `camelCase` when returning the instance in the API, else these will be the only keys with `snake_case`. Any objection to that?

mindspank:
Looks great to me, good stuff!
One ask would be to get a embed e2e test in place, it will be a property that's going to be heavily used

runtime/registry.go
```diff
@@ -148,6 +149,45 @@ func (r *Runtime) TempDir(instanceID string, elem ...string) (string, error) {
...
+	}
+
+	templateData := parser.TemplateData{
+		Environment: inst.Environment,
+		User:        attrs,
+		Variables:   vars,
+	}
+
+	featureFlags := make(map[string]bool)
+	mergedFeatureFlags := drivers.MergeDefaultFeatureFlags(inst.FeatureFlags)
```
begelundmuller: Consider just exposing `drivers.DefaultFeatureFlags` as an exported const. Then you can replace the `drivers.MergeDefaultFeatureFlags` function with just a couple lines here:
```go
mergedFeatureFlags := maps.Clone(drivers.DefaultFeatureFlags)
maps.Copy(mergedFeatureFlags, inst.FeatureFlags)
```

runtime/server/instances.go
```diff
@@ -27,9 +28,15 @@ func (s *Server) ListInstances(ctx context.Context, req *runtimev1.ListInstances
...
 
+	claims := auth.GetClaims(ctx).SecurityClaims()
+
 	pbs := make([]*runtimev1.Instance, len(instances))
 	for i, inst := range instances {
-		pbs[i] = instanceToPB(inst, true)
+		featureFlags, err := runtime.ResolveFeatureFlags(inst, claims)
+		if err != nil {
+			return nil, status.Error(codes.Internal, err.Error())
+		}
```
begelundmuller, line 38: Is this always an internal error? Or could it happen if a user provided a broken template string? If it could be user caused, we should not use the `Internal` code

runtime/server/instances.go
```diff
@@ -65,8 +72,13 @@ func (s *Server) GetInstance(ctx context.Context, req *runtimev1.GetInstanceRequ
 		return nil, status.Error(codes.InvalidArgument, err.Error())
 	}
 
+	featureFlags, err := runtime.ResolveFeatureFlags(inst, auth.GetClaims(ctx).SecurityClaims())
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
```
begelundmuller: Same comment as above here

runtime/server/instances.go
```diff
@@ -104,8 +116,13 @@ func (s *Server) CreateInstance(ctx context.Context, req *runtimev1.CreateInstan
...
```
begelundmuller: Same comment as above here

runtime/server/instances.go
```diff
@@ -189,8 +206,13 @@ func (s *Server) EditInstance(ctx context.Context, req *runtimev1.EditInstanceRe
 		s.logger.Error("failed to acquire repo after editing instance", zap.String("instance_id", req.InstanceId), zap.Error(err), observability.ZapCtx(ctx))
 	}
 
+	featureFlags, err := runtime.ResolveFeatureFlags(inst, auth.GetClaims(ctx).SecurityClaims())
+	if err != nil {
+		return nil, status.Error(codes.Internal, err.Error())
```
begelundmuller: Same comment as above here

runtime/drivers/registry.go
```diff
@@ -184,3 +185,35 @@ func (i *Instance) Config() (InstanceConfig, error) {
...
+	"dimensionSearch":     "false",
+	"twoTieredNavigation": "false",
+	"rillTime":            "true",
+	"hidePublicUrl":       "{{.user.embed}}",
+	"exportHeader":        "false",
+	"alerts":              "true",
+	"reports":             "true",
+	"darkMode":            "false",
+	"chat":                "true",
+	"dashboardChat":       "true",
```
begelundmuller: It would be nice with a one line docstring for each property about what the feature is

runtime/drivers/registry.go
begelundmuller: Yeah I know the UI uses different casing, I'm mainly just thinking about this from a YAML consistency perspective. Since everything we do in YAML is snake case, I just feel it's worth being consistent here. E.g. you can imagine our AI code agent prompt says "always use snake case", and it would be nice if we didn't have exceptions then.

AdityaHegde: Ya I am ok with this, we can convert to `camelCase` before returning in API.

@mindspank any objections to this? We want to use `snake_case` to keep all fields in yaml consistent.

mindspank: If we maintain backwards compatibility then no objections

runtime/server/instances.go
AdityaHegde: Updated to `InvalidArgument`. It will always be because of broken template variable used. But most of them will be caught in parsing.

begelundmuller (approved):

runtime/server/instances.go
mindspank: How will this behave if:

1. Customer has a flag on
2. We remove the flag because the feature is now launched

Will it break in parsing? I would expect it to warn but not fail to parse

AdityaHegde: We do not do a strict parsing. So any additional keys are parsed and returned but UI wont use it.

PR #7969, opened 2025-09-17
Streaming completions API

begelundmuller:
This PR adds a `CompleteStreaming` RPC to the runtime that streams completion messages as they are generated. It is backed by our current AI implementation in `runtime/completion.go`, but will work with the upcoming new internals without change to the API signature.

Notes:
- Similar to `WatchFiles` and `WatchRepo`, it has an SSE shim for use in the browser that requires manual handling in the frontend (not handled by Orval).
- The messages it streams back do not map 1:1 to the results obtained from `GetConversation`, but contain the same information. Notable differences:
  - The streaming messages always contain only one content block
  - The streaming message IDs are random and not persisted

Example usage:

```bash
curl --request POST \
  --url 'http://localhost:9009/v1/instances/default/ai/complete/stream?stream=messages' \
  --header 'Accept: text/event-stream' \
  --header 'Content-Type: application/json' \
  --data '{"prompt": "How many auctions have there been?"}'
```


proto/rill/runtime/v1/api.proto
```diff
@@ -238,9 +238,12 @@ service RuntimeService {
...
-  // ListConversations lists all AI chat conversations for an instance.
-  rpc ListConversations(ListConversationsRequest) returns (ListConversationsResponse) {
-    option (google.api.http) = {get: "/v1/instances/{instance_id}/ai/conversations"};
+  // CompleteStreaming runs an AI-powered chat completion, optionally invoking agents or tool calls available in Rill.
+  rpc CompleteStreaming(CompleteStreamingRequest) returns (stream CompleteStreamingResponse) {
+    option (google.api.http) = {
+      post: "/v1/instances/{instance_id}/ai/complete/stream",
+      body: "*"
+    };
   }
```
ericpgreen2, line 247: Was it intended to delete the `ListConversations` endpoint?

begelundmuller: WTF no, seems like it somehow happened when I cherry-picked the file from the other PR. Sorry I didn't notice! Fixed now.

PR #7974, opened 2025-09-17
Gplata/plat 212 single

grahamplata:
When running  commands like`rill project refresh --model MODEL --partition PARTITION_ID` the current model reconciler not only refreshes the specified partition, but also runs a normal incremental refresh. That's usually not a problem, but has caused some confusion and can be an unnecessary expense in certain cases.

My proposed changes add an additional field to the modelState in the form of "triggered partitions" which are used to track specific partitions to be executed. It will return early, preventing the normal incremental refresh of all partitions.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


proto/rill/runtime/v1/resources.proto
```diff
@@ -152,6 +152,8 @@ message ModelState {
   string test_hash = 27;
   // test_errors contains the results of the model's tests.
   repeated string test_errors = 28;
+  // repeated string of partition keys that were explicitly triggered
+  repeated string triggered_partitions = 29;
```
begelundmuller: Since this could potentially be a big value, it'd be better to add a column for explicit triggers in the `model_partitions` table instead of in the model resource. The state here is returned as part of many API calls, so need to be kept somewhat lean.

runtime/reconcilers/model.go
```diff
@@ -988,6 +988,49 @@ func (r *ModelReconciler) executeAll(ctx context.Context, self *runtimev1.Resour
...
+		model.State.TriggeredPartitions = nil
+		err = r.C.UpdateState(ctx, self.Meta.Name, self)
+		if err != nil {
+			return "", nil, false, err
+		}
+
+		r.C.Logger.Info("Completed targeted partition refresh", zap.Strings("partitions", triggeredPartitions), observability.ZapCtx(ctx))
+
+		return executor.finalConnector, result, incrementalRun, nil
+	}
```
begelundmuller: 1. What if this coincides with a regular incremental refresh (e.g. if it still hasn't completed the first full ingestion, or if it coincides with the refresh cron)? Then it can't return like this. Not sure, you may also need to add a `bool trigger_individual` property in the model spec to handle this correctly.
2. Rather than have separate handling for individually triggered partitions, consider incorporating this handling as filters in the `FindModelPartitions` calls made in the existing partitioned processing logic below.

runtime/drivers/sqlite/catalog.go
```diff
@@ -334,6 +340,34 @@ func (c *catalogStore) UpdateModelPartitionsPendingIfError(ctx context.Context,
...
 }
 
+func (c *catalogStore) UpdateModelPartitionsExplicitlyTriggered(ctx context.Context, modelID string, partitionKeys []string, triggered bool) error {
+	if len(partitionKeys) == 0 {
+		return nil
+	}
+
+	var qry strings.Builder
+	var args []any
+	qry.WriteString("UPDATE model_partitions SET explicitly_triggered=? WHERE instance_id=? AND model_id=? AND key IN (")
```
begelundmuller: Additionally it needs `SET executed_on=NULL` since that's our indicator for it being pending

runtime/reconcilers/model.go
```diff
@@ -1026,7 +1027,18 @@ func (r *ModelReconciler) executeAll(ctx context.Context, self *runtimev1.Resour
 	defer release()
 
 	// First step is to resolve and sync the partitions.
-	err = r.resolveAndSyncPartitions(ctx, self, model, incrementalState)
+	if !incrementalRun || model.Spec.Trigger || model.Spec.TriggerFull || model.Spec.TriggerPartitions {
```
begelundmuller: What does this new check accomplish?

In general I think it's fine to sync partitions every time, just not to execute them if only a specific partition was triggered.

runtime/reconcilers/model.go
```diff
@@ -1026,7 +1027,18 @@ func (r *ModelReconciler) executeAll(ctx context.Context, self *runtimev1.Resour
...
+		if err != nil {
+			return "", nil, false, fmt.Errorf("failed to sync partitions: %w", err)
+		}
+	}
+
+	// Check if we have explicitly triggered partitions after resolution
+	explicitlyTriggeredPartitions, err := catalog.FindModelPartitions(ctx, &drivers.FindModelPartitionsOptions{
+		ModelID:                  model.State.PartitionsModelId,
+		WhereExplicitlyTriggered: true,
+	})
```
begelundmuller: Is it necessary to have a separate call here? Instead of a separate call, could this be incorporated into the existing `FindModelPartitions` call? E.g. pass something like `WherePending: incrementalRun, WhereExplicitlyTriggered: model.Spec.TriggerPartitions` in that call

runtime/reconcilers/model.go
```diff
@@ -1035,19 +1047,31 @@ func (r *ModelReconciler) executeAll(ctx context.Context, self *runtimev1.Resour
...
+	explicitlyTargetedRefresh := len(explicitlyTriggeredPartitions) > 0
+	partitionFilter := &drivers.FindModelPartitionsOptions{
+		ModelID: model.State.PartitionsModelId,
+	}
+
+	if model.Spec.TriggerPartitions {
+		partitionFilter.WhereExplicitlyTriggered = true
+	} else {
+		partitionFilter.WherePending = true
+	}
```
begelundmuller: Needs to handle the case where a partition refresh coincides with an incremental refresh. I.e. these two cases are not mutually exclusive.

runtime/reconcilers/refresh_trigger.go
```diff
@@ -129,6 +129,35 @@ func (r *RefreshTriggerReconciler) Reconcile(ctx context.Context, n *runtimev1.R
...
+				triggeredPartitions = mt.Partitions
+			} else if mt.AllErroredPartitions {
+				erroredPartitions, err := catalog.FindModelPartitionsWithErrors(ctx, modelID)
+				if err != nil {
+					return runtime.ReconcileResult{Err: fmt.Errorf("failed to find errored partitions for model %s: %w", mt.Model, err)}
+				}
+				for _, p := range erroredPartitions {
+					triggeredPartitions = append(triggeredPartitions, p.Key)
+				}
+			}
```
begelundmuller: Can this be addressed by setting `ExplicitlyTriggered` to `true` in `UpdateModelPartitionsPendingIfError`? Instead of the different handling here.

runtime/drivers/catalog.go
```diff
@@ -47,8 +47,12 @@ type CatalogStore interface {
 	UpdateModelPartition(ctx context.Context, modelID string, partition ModelPartition) error
 	UpdateModelPartitionPending(ctx context.Context, modelID, partitionKey string) error
 	UpdateModelPartitionsPendingIfError(ctx context.Context, modelID string) error
+	UpdateModelPartitionsExplicitlyTriggered(ctx context.Context, modelID string, partitionKeys []string, triggered bool) error
```
begelundmuller: Feels like these two funcs can be combined, maybe:
```
UpdateModelPartitionsTriggered(ctx context.Context, modelID string, wherePartitionKeyIn []string, whereErrored bool) error
```

runtime/drivers/catalog.go
```diff
@@ -46,9 +46,12 @@ type CatalogStore interface {
 	InsertModelPartition(ctx context.Context, modelID string, partition ModelPartition) error
 	UpdateModelPartition(ctx context.Context, modelID string, partition ModelPartition) error
 	UpdateModelPartitionPending(ctx context.Context, modelID, partitionKey string) error
-	UpdateModelPartitionsPendingIfError(ctx context.Context, modelID string) error
+	UpdateModelPartitionsTriggered(ctx context.Context, modelID string, wherePartitionKeyIn []string, whereErrored bool, triggered bool) error
 	DeleteModelPartitions(ctx context.Context, modelID string) error
 
+	FindModelPartitionByKey(ctx context.Context, modelID, partitionKey string) (ModelPartition, error)
+	FindModelPartitionsWithErrors(ctx context.Context, modelID string) ([]ModelPartition, error)
```
begelundmuller: Seems like these are not used/needed now – remove them?

runtime/drivers/sqlite/catalog.go
```diff
@@ -320,13 +326,44 @@ func (c *catalogStore) UpdateModelPartitionPending(ctx context.Context, modelID,
...
 }
 
-func (c *catalogStore) UpdateModelPartitionsPendingIfError(ctx context.Context, modelID string) error {
-	_, err := c.db.ExecContext(
-		ctx,
-		"UPDATE model_partitions SET executed_on=NULL WHERE instance_id=? AND model_id=? AND error != ''",
-		c.instanceID,
-		modelID,
-	)
+func (c *catalogStore) UpdateModelPartitionsTriggered(ctx context.Context, modelID string, wherePartitionKeyIn []string, whereErrored, triggered bool) error {
```
begelundmuller: Is it necessary to pass `triggered` here, or can it not always just be set to `true`? Even if the trigger was `whereErrored` (`rill project refresh --errored-partitions ...`), that's kind of still an explicit trigger (compared to a scheduled trigger).

runtime/drivers/catalog.go
```diff
@@ -46,9 +46,12 @@ type CatalogStore interface {
 	InsertModelPartition(ctx context.Context, modelID string, partition ModelPartition) error
 	UpdateModelPartition(ctx context.Context, modelID string, partition ModelPartition) error
 	UpdateModelPartitionPending(ctx context.Context, modelID, partitionKey string) error
-	UpdateModelPartitionsPendingIfError(ctx context.Context, modelID string) error
+	UpdateModelPartitionsTriggered(ctx context.Context, modelID string, wherePartitionKeyIn []string, whereErrored bool, triggered bool) error
```
begelundmuller: I think you can remove the old `UpdateModelPartitionPending` function now – seems like `UpdateModelPartitionsTriggered` can replace it as a more generic alternative?

runtime/drivers/sqlite/catalog.go
```diff
@@ -320,13 +326,44 @@ func (c *catalogStore) UpdateModelPartitionPending(ctx context.Context, modelID,
...
+		for i, k := range wherePartitionKeyIn {
+			if i == 0 {
+				qry.WriteString("?")
+			} else {
+				qry.WriteString(",?")
+			}
+			args = append(args, k)
+		}
+		qry.WriteString(")")
+	}
```
begelundmuller: 1. Feels like this can be simplified into two separate cases instead of three if-else cases
2. In the combined case, should the operator between them be `OR` instead of `AND`?

runtime/reconcilers/model.go
```diff
@@ -1035,23 +1036,43 @@ func (r *ModelReconciler) executeAll(ctx context.Context, self *runtimev1.Resour
...
+		ModelID: model.State.PartitionsModelId,
+	}
+
+	if model.Spec.TriggerPartitions {
+		// When partitions are explicitly triggered, process both explicitly triggered partitions AND any pending partitions in the same run
+		partitionFilter.WhereExplicitlyTriggered = true
+		partitionFilter.WherePending = true
+	} else {
+		partitionFilter.WherePending = true
+	}
```
begelundmuller: Can be simplified:
```suggestion
	// When partitions are explicitly triggered, process both explicitly triggered partitions AND any pending partitions in the same run
	partitionFilter.WherePending = true
	partitionFilter.WhereExplicitlyTriggered = model.Spec.TriggerPartitions
```

runtime/reconcilers/model.go
```diff
@@ -1159,6 +1183,19 @@ func (r *ModelReconciler) executeAll(ctx context.Context, self *runtimev1.Resour
...
+		var partitionKeys []string
+		for _, p := range explicitlyTriggeredPartitions {
+			partitionKeys = append(partitionKeys, p.Key)
+		}
+		err = catalog.UpdateModelPartitionsTriggered(ctx, model.State.PartitionsModelId, partitionKeys, false, false)
+		if err != nil {
+			return "", nil, false, fmt.Errorf("failed to clear triggered flag: %w", err)
+		}
+		r.C.Logger.Info("Completed targeted partition refresh", zap.Int("partitions_processed", len(explicitlyTriggeredPartitions)), observability.ZapCtx(ctx))
+	}
```
begelundmuller: This redundantly adds an extra DB operation – just set `ExplicitlyTriggered = false` in all cases when finishing a partition in `executePartition` here: https://github.com/rilldata/rill/blob/c92a41f03b205dad72e765ef979d8a848a685a3f/runtime/reconcilers/model.go#L1208-L1216

runtime/reconcilers/model.go
```diff
@@ -1030,32 +1031,51 @@ func (r *ModelReconciler) executeAll(ctx context.Context, self *runtimev1.Resour
 	defer release()
 
 	// First step is to resolve and sync the partitions.
-	err = r.resolveAndSyncPartitions(ctx, self, model, incrementalState)
-	if err != nil {
-		return "", nil, false, fmt.Errorf("failed to sync partitions: %w", err)
+	// Skip this if we're doing an explicit partition refresh (TriggerPartitions), as we only want to execute the explicitly triggered partitions
+	if !model.Spec.TriggerPartitions {
```
begelundmuller: You can have a case where e.g. both `.TriggerFull` _and_ `.TriggerPartitions` are true (for example, the user may first refresh a partition, then before that operation is done, trigger a full refresh). So this check is not sufficient to determine whether to discover new partitions or not.

runtime/reconcilers/refresh_trigger.go
```diff
@@ -129,18 +129,24 @@ func (r *RefreshTriggerReconciler) Reconcile(ctx context.Context, n *runtimev1.R
...
-			if mt.AllErroredPartitions {
-				err := catalog.UpdateModelPartitionsPendingIfError(ctx, modelID)
-				if err != nil {
-					return runtime.ReconcileResult{Err: fmt.Errorf("failed to update partitions for model %s: %w", mt.Model, err)}
-				}
+			// Handle specific partitions vs all errored partitions
+			if len(mt.Partitions) > 0 {
+				// Set the triggered flag on specific partitions
+				err = catalog.UpdateModelPartitionsTriggered(ctx, modelID, mt.Partitions, false)
+			} else if mt.AllErroredPartitions {
```
begelundmuller: These cases are not necessarily mutually exclusive

runtime/reconcilers/model.go
```diff
@@ -1030,32 +1031,51 @@ func (r *ModelReconciler) executeAll(ctx context.Context, self *runtimev1.Resour
...
+		err = r.resolveAndSyncPartitions(ctx, self, model, incrementalState)
```
begelundmuller: You can have a case where e.g. both `.TriggerFull` _and_ `.TriggerPartitions` are true (for example, the user may first refresh a partition, then before that operation is done, trigger a full refresh). So this check is not sufficient to determine whether to discover new partitions or not.

runtime/reconcilers/model.go
```diff
@@ -1030,32 +1031,51 @@ func (r *ModelReconciler) executeAll(ctx context.Context, self *runtimev1.Resour
...
 
+	var explicitlyTriggeredPartitions []drivers.ModelPartition
+	explicitlyTargetedRefresh := model.Spec.TriggerPartitions
+	partitionFilter := &drivers.FindModelPartitionsOptions{
+		ModelID: model.State.PartitionsModelId,
+	}
+
+	// When partitions are explicitly triggered, process both explicitly triggered partitions AND any pending partitions in the same run
+	partitionFilter.WherePending = true
+	partitionFilter.WhereExplicitlyTriggered = model.Spec.TriggerPartitions
```
begelundmuller: Looking over this, why does it even need the `explicitly_triggered` field and related queries/checks? It should never _not_ run pending partitions. It just should not call `resolveAndSyncPartitions` (which discovers new partitions) when only a specific partition has been triggered (i.e. marked pending).

begelundmuller (approved):

PR #7984, opened 2025-09-18
fix(rilltime): ceil legacy iso instead of truncating

AdityaHegde:
Legacy iso truncated by the smallest grain in iso only. This is how old ISO resolution worked.

We are changing this behaviour to instead be a add a unit before truncating.

**Checklist:**
- [x] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/runtime/pkg/rilltime/rilltime.go
+++ b/runtime/pkg/rilltime/rilltime.go
@@ -332,7 +332,11 @@ func (e *Expression) Eval(evalOpts EvalOptions) (time.Time, time.Time, timeutil.
 		tg := timeutil.TimeGrainUnspecified
 		if e.Grain != nil {
 			tg = grainMap[*e.Grain]
+
+			// ISO durations are mapped to `ref-iso to ref as of watermark/grain+1grain`
+			isoStart = timeutil.OffsetTime(isoStart, tg, 1, e.tz)
 			isoStart = timeutil.TruncateTime(isoStart, tg, e.tz, evalOpts.FirstDay, evalOpts.FirstMonth)
+			isoEnd = timeutil.OffsetTime(isoEnd, tg, 1, e.tz)
 			isoEnd = timeutil.TruncateTime(isoEnd, tg, e.tz, evalOpts.FirstDay, evalOpts.FirstMonth)
 		}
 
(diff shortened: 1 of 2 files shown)


runtime/pkg/rilltime/rilltime.go
```diff
@@ -332,7 +332,11 @@ func (e *Expression) Eval(evalOpts EvalOptions) (time.Time, time.Time, timeutil.
 		tg := timeutil.TimeGrainUnspecified
 		if e.Grain != nil {
 			tg = grainMap[*e.Grain]
+
+			// ISO durations are mapped to `ref-iso to ref as of watermark/grain+1grain`
+			isoStart = timeutil.OffsetTime(isoStart, tg, 1, e.tz)
```
begelundmuller, line 337: Does this mean the `rilltime` package is no longer backwards compatible with old alerts/reports? Ideally we want to move alerts/reports to use `rilltime` as well, I think the only reason we didn't do it in the first PR was to allow time for QA first.

The watermark used for alerts/reports is the exclusive end of the range they are expected to check. E.g. for `expr=P1D` and `watermark=2025-09-19`, an alert/report would check `time >= 2025-09-18 AND time < 2025-09-19`.

AdityaHegde: Yes, if we ever change [this path](https://github.com/rilldata/rill/blob/main/runtime/metricsview/executor_rewrite_time.go#L67) to go through `rilltime.Parse` then it will no longer be backwards compatible.

It honestly depends on what we would do with old `IsoDuration` and `IsoOffset`. We could send them as `<IsoOffset> - <IsoOffset> as of watermark/<grain>` to keep backwards compatibility.

begelundmuller: > If we ever change [this path](https://github.com/rilldata/rill/blob/main/runtime/metricsview/executor_rewrite_time.go#L67) to go through rilltime.Parse then it will no longer be backwards compatible.

Yeah that's the change I was thinking of. We only kept the old code temporarily while Rill time went through QA, but we should change it to use the `rilltime` library now. We don't want to maintain two slightly different time range implementations.

AdityaHegde: This PR is needed for the release, so will do this refactor as a follow up.

PR #7985, opened 2025-09-18
fix: display ranges as inclusive

briangregoryholmes:
When displaying ranges in their compact forms, we prefer displaying inclusive endings at day precision or lower.

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


web-common/src/lib/time/ranges/formatter.ts
```diff
@@ -41,7 +39,14 @@ function formatDatePartOfTimeRange(interval: Interval, grain: V1TimeGrain) {
 
   if (showDay) format.day = "numeric";
 
-  return interval.toLocaleString(format);
+  const displayAsInclusiveEnd = grainOrder >= dayGrainOrder;
+
+  return displayAsInclusiveEnd
+    ? Interval.fromDateTimes(
```
AdityaHegde, line 46: This needs a check to make sure start and end are not same. The test failures are because of that.

AdityaHegde (approved):
Approved with a small comment

PR #7987, opened 2025-09-18
upgrade s3 api

NamanMahor:
Closes# PLAT-228  [PLAT-228: Upgrade s3 driver to aws-sdk-go-v2](https://linear.app/rilldata/issue/PLAT-228/upgrade-s3-driver-to-aws-sdk-go-v2)
**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!

### DIFF
--- a/runtime/drivers/s3/object_store.go
+++ b/runtime/drivers/s3/object_store.go
@@ -66,8 +66,12 @@ func (c *Connection) parseBucketURL(path string) (*globutil.URL, error) {
 
 func (c *Connection) openBucket(ctx context.Context, bucket string, anonymous bool) (*blob.Bucket, error) {
 	var s3client *s3.Client
+	var err error
 	if anonymous {
-		s3client = getAnonymousS3Client(c.config)
+		s3client, err = getAnonymousS3Client(ctx, c.config, bucket)
+		if err != nil {
+			return nil, err
+		}
 	} else {
 		var err error
 		s3client, err = getS3Client(ctx, c.config, bucket)
--- a/runtime/drivers/s3/s3.go
+++ b/runtime/drivers/s3/s3.go
@@ -310,16 +310,27 @@ func bucketRegionFromConfig(ctx context.Context, cfg aws.Config, confProp *Confi
 	return region, nil
 }
 
-func getAnonymousS3Client(confProp *ConfigProperties) *s3.Client {
+func getAnonymousS3Client(ctx context.Context, confProp *ConfigProperties, bucket string) (*s3.Client, error) {
 	cfg := aws.Config{
 		Credentials: aws.AnonymousCredentials{},
+	}
+	region := confProp.Region
+	// If the region is not explicitly provided in the config,
+	// try to automatically detect it from the bucket
+	if region == "" && bucket != "" {
+		var err error
+		region, err = bucketRegionFromConfig(ctx, cfg, confProp, bucket)
+		if err != nil {
+			return nil, fmt.Errorf("failed to detect bucket region: %w", err)
+		}
 	}
 	return s3.NewFromConfig(cfg, func(o *s3.Options) {
 		if confProp.Endpoint != "" {
 			o.BaseEndpoint = aws.String(confProp.Endpoint)
 			o.UsePathStyle = true
 		}
-	})
+		o.Region = region
+	}), nil
 }
 
 func getS3Client(ctx context.Context, confProp *ConfigProperties, bucket string) (*s3.Client, error) {


runtime/drivers/s3/s3.go
```diff
@@ -284,81 +269,138 @@ func (c *Connection) AsNotifier(properties map[string]any) (drivers.Notifier, er
...
+	if err != nil {
+		return aws.Config{}, fmt.Errorf("failed to load AWS config: %w", err)
+	}
+	return cfg, nil
+}
+
 // newCredentials returns credentials for connecting to AWS.
-func (c *Connection) newCredentials() (*credentials.Credentials, error) {
-	// If a role ARN is provided, assume the role and return the credentials.
+func (c *Connection) newCredentials(ctx context.Context) (aws.CredentialsProvider, error) {
```
k-anshul: Nit: `newCredentialsProvider` ?

runtime/drivers/s3/s3.go
```diff
@@ -284,81 +269,138 @@ func (c *Connection) AsNotifier(properties map[string]any) (drivers.Notifier, er
...
+
+	if c.config.AccessKeyID != "" && c.config.SecretAccessKey != "" {
+		provider = credentials.NewStaticCredentialsProvider(
+			c.config.AccessKeyID,
+			c.config.SecretAccessKey,
+			c.config.SessionToken,
+		)
+	} else if c.config.AllowHostAccess {
+		// Allow host-based credentials (env vars, shared config, etc.)
+		cfg, err := config.LoadDefaultConfig(ctx)
```
k-anshul, line 414: The earlier implementation explicitly created lookup chain to not allow remote credentials lookup. Is that not to be handled now ? 
```
// The chain used here is a duplicate of defaults.CredProviders(), but without the remote credentials lookup (since they resolve too slowly).
```

runtime/drivers/s3/s3.go
```diff
@@ -284,81 +269,138 @@ func (c *Connection) AsNotifier(properties map[string]any) (drivers.Notifier, er
...
+		// Fall back to anonymous creds
+		provider = aws.AnonymousCredentials{}
+	}
+
+	if _, err := provider.Retrieve(ctx); err != nil {
+		return aws.AnonymousCredentials{}, nil
 	}
 
-	return creds, nil
+	// Fallback to anonymous
```
k-anshul: Misplaced comment ?

runtime/drivers/s3/object_store.go
```diff
@@ -84,86 +79,28 @@ func (c *Connection) parseBucketURL(path string) (*globutil.URL, error) {
 }
 
 func (c *Connection) openBucket(ctx context.Context, bucket string, anonymous bool) (*blob.Bucket, error) {
-	var creds *credentials.Credentials
+	region := c.config.Region
+	if c.config.Endpoint == "" && region == "" {
+		if r, err := c.BucketRegion(ctx, bucket); err == nil && r != "" {
```
k-anshul: Shouldn't we throw error if resolving bucket region fails ?

runtime/drivers/s3/s3.go
NamanMahor: i did't see any slowest tested in local. also there is no good way to do it in v2. Also its default so I think it should not be problem.

runtime/drivers/s3/object_store.go
NamanMahor: done

runtime/drivers/duckdb/model_executor_objectstore_self.go
```diff
@@ -147,9 +146,9 @@ func objectStoreSecretSQL(ctx context.Context, opts *drivers.ModelExecuteOptions
 			if err != nil {
 				return "", fmt.Errorf("failed to parse path %q: %w", optionalBucketURL, err)
 			}
-			reg, err := conn.BucketRegion(ctx, uri.Host)
+			reg, err := s3.BucketRegion(ctx, *s3Config, uri.Host)
```
k-anshul: Any particular reason for not accepting pointer in function definition and dereferencing in many calls below ?

runtime/drivers/s3/s3.go
```diff
@@ -156,14 +158,15 @@ var _ drivers.Handle = &Connection{}
...
 func (c *Connection) Ping(ctx context.Context) error {
-	cfg, err := c.GetAWSConfig(ctx)
-	if err != nil {
-		return err
-	}
-	stsClient := c.GetSTSClient(cfg)
-	_, err = stsClient.GetCallerIdentity(ctx, &sts.GetCallerIdentityInput{})
-	if err != nil {
-		return fmt.Errorf("GetCallerIdentity failed: %w", err)
+	if c.config.Endpoint == "" {
```
k-anshul, line 161: Will you be implementing `Ping` for compatible services later?

runtime/drivers/s3/s3.go
k-anshul: I will let @begelundmuller also take a look at this. He added this initially so can comment how was this identified. 
I think it can resolve slowly on VMs/pods running on AWS.

runtime/drivers/duckdb/model_executor_objectstore_self.go
NamanMahor: Yes previously it was bug. we were not using config *s3Config which is defined(override in model). `conn.BucketRegion(ctx, uri.Host)` was using connect config

runtime/drivers/s3/s3.go
NamanMahor: Yes there is sperate ticket for handling it.
https://linear.app/rilldata/issue/PLAT-184/getcalleridentity-in-s3-ping-failed-for-cloudflare-r2

runtime/drivers/duckdb/model_executor_objectstore_self.go
k-anshul: I understand that we need to pass s3Config as well but why do we need to derefrence the pointer as well?

NamanMahor: no point in making it pointer since only thing it uses is config which we are passing. and also it make confusion in function to which config should be use c.config or config passed in params.

runtime/drivers/s3/s3.go
begelundmuller: I forgot the exact context, but yes I believe it resolved slowly either in VMs/containers, or maybe it was when there wasn't a fast network connection (since in some cases, I believe it tries to load credentials from the network, and it can then take some time for it to discover that that's not possible).

NamanMahor: @begelundmuller I think we should retest it on test cluster since we have upgrade to v2

runtime/drivers/s3/api.go
```diff
@@ -2,54 +2,23 @@ package s3
...
-	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/aws/aws-sdk-go-v2/service/s3"
 	runtimev1 "github.com/rilldata/rill/proto/gen/rill/runtime/v1"
 	"gocloud.dev/blob"
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
 func (c *Connection) ListBuckets(ctx context.Context) ([]string, error) {
-	creds, err := c.newCredentials()
+	client, err := getS3Client(ctx, *c.config, "")
```
begelundmuller: These dereferencing arg passes `*c.config` look wrong – I agree with Anshul that we should be consistent and either use values or pointers everywhere, not mix and match.

runtime/drivers/duckdb/model_executor_objectstore_self.go
NamanMahor: done

runtime/drivers/s3/api.go
NamanMahor: done

PR #7988, opened 2025-09-18
feat:add color range config to heatmap

djbarnwal:
Add a scheme and range toggle under Color selector for heatmap charts.

The scheme tab has a list of preset color schemes. The range tab has a start and end color value which the user can tweak.

https://linear.app/rilldata/issue/APP-295/support-custom-chart-colors-for-heatmaps-and-pie-charts

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-common/src/features/canvas/components/charts/builder.ts
+++ b/web-common/src/features/canvas/components/charts/builder.ts
@@ -16,6 +16,7 @@ import {
   isDomainStringArray,
   isFieldConfig,
   mergedVlConfig,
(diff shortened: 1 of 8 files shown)


djbarnwal:
> Only feedback is:
> 
> * Is `range` standard terminology? If not, perhaps `Custom` might make more sense?

`range` is the standard terminology for output mapping. Perhaps `gradient` works here?

PR #7994, opened 2025-09-19
Correct DuckDB/Motherduck connector and source properties

begelundmuller:
Note that I only described the important config properties, not all the available advanced options.

### DIFF
--- a/runtime/drivers/duckdb/duckdb.go
+++ b/runtime/drivers/duckdb/duckdb.go
@@ -49,6 +49,14 @@ var spec = drivers.Spec{
 			Placeholder: "/path/to/main.db",
 		},
 		{
+			Key:         "attach",
+			Type:        drivers.StringPropertyType,
+			Required:    false,
+			DisplayName: "Attach",
+			Description: "Attach to an existing DuckDB database. This is an alternative to `path` that supports attach options.",
+			Placeholder: "'ducklake:metadata.ducklake' AS my_ducklake(DATA_PATH 'datafiles')",
+		},
+		{
 			Key:         "mode",
 			Type:        drivers.StringPropertyType,
 			Required:    false,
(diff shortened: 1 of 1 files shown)


runtime/drivers/duckdb/duckdb.go
```diff
@@ -113,15 +108,16 @@ var motherduckSpec = drivers.Spec{
...
 			Type:        drivers.StringPropertyType,
 			Required:    true,
-			DisplayName: "MotherDuck Connection String",
-			Placeholder: "md:motherduck.db",
+			DisplayName: "Schema name",
+			Placeholder: "your_schema_name",
+			Hint:        "Set the default schema used by the MotherDuck database",
 		},
+	},
+	SourceProperties: []*drivers.PropertySpec{
```
k-anshul, line 120: Given motherduck as source connector is not to be supported from now should this be an empty array instead ?

begelundmuller: The `SourceProperties` is kind of a weird name since it's actually model properties now. Since we'll still support modeling on Motherduck (not as a source, but inside the Motherduck), I think the `sql` property should still be documented here?

Or what do you think? These abstractions are a bit out of date, so maybe there's something I'm not thinking of.

k-anshul: Yeah somewhat difficult to comprehend but sounds good.

k-anshul (approved):

runtime/drivers/duckdb/duckdb.go
lovincyrus: > The `SourceProperties` is kind of a weird name since it's actually model properties now. Since we'll still support modeling on Motherduck (not as a source, but inside the Motherduck)

This was part of the reason for the change. I was struggling to find the source of truth for the properties to be displayed appropriately in the form.

With the latest changes, the form now includes only the `sql` field.

CleanShot 2025-09-19 at 10 46 40@2x

lovincyrus: I believe we need to display `path` from ConfigProperties and `sql` from SourceProperties for DuckDB form.

lovincyrus: https://rilldata.slack.com/archives/C088Q884YF7/p1758305286090269

PR #7675, opened 2025-07-25
Add pinning behavior to dimensions filter

lovincyrus:
This pull request enhances the DimensionFilter component by implementing a "pinning" behavior for selected items in
  Select mode. The changes improve UX by showing checked items at the top of the dropdown list, similar to
  SelectionDropdown behavior.

Closes https://linear.app/rilldata/issue/APP-25/add-pinning-behavior-to-dimension-filter-pill-dropdown

https://github.com/user-attachments/assets/46244a5f-3d23-4426-b4f2-f743d4443fdd

**Checklist:**
- [ ] Covered by tests
- [x] Ran it and it works as intended
- [x] Reviewed the diff before requesting a review
- [ ] Checked for unhandled edge cases
- [x] Linked the issues it closes
- [ ] Checked if the docs need to be updated. If so, create a separate Linear DOCS issue
- [ ] Intend to cherry-pick into the release branch
- [x] I'm proud of this work!

PR #7995, opened 2025-09-19
[APP-389] fix: exclude mode toggle

briangregoryholmes:
INSERT DESCRIPTION HERE

**Checklist:**
- [ ] Covered by tests
- [x]  Reviewed the diff before requesting a review
- [ ] I'm proud of this work!

### DIFF
--- a/web-common/src/features/dashboards/filters/dimension-filters/DimensionFilter.svelte
+++ b/web-common/src/features/dashboards/filters/dimension-filters/DimensionFilter.svelte
@@ -277,8 +277,6 @@
 
   function handleToggleExcludeMode() {
     curExcludeMode = !curExcludeMode;
-    const shouldToggleImmediately = mode === curMode;
-    if (shouldToggleImmediately) onToggleFilterMode();
   }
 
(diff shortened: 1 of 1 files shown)


AdityaHegde:
This changes the way exclude works. Before https://github.com/rilldata/rill/pull/7675 changes in select mode were instant. Any action in the bottom bar was instant in all modes including the exclude toggle.

But now, after #7675 selections only apply after clicking out of the dropdown. Looks like exclude mode was not considered for this change. After this PR exclude toggle also does not apply instantly.

@mindspank Are we ok with making exclude not instant? Please check how it works in all modes. It doesnt feel like good UX.

briangregoryholmes:
My thinking was that everything in the dropdown should follow the same behavior: instant or not instant. I prefer instant and am not sure why this was changed in #7675, but matching the delayed apply behavior was the quickest way to fix the underlying issue.

mindspank:
It's fine by me, I can see some benefits to not having instant filters as well especially for customer managed olaps where you might want to reduce query load and not fire off queries for every single applied filter

PR #7979, opened 2025-09-18
Add Data Modal Layout

lovincyrus:
This PR tweaks the layout of add data modal from 3 to 4 columns.

Closes https://linear.app/rilldata/issue/APP-381/add-source-modal-is-currently-3-column-grid-layout-needs-4-column-grid

CleanShot 2025-09-17 at 19 17 20@2x

https://github.com/user-attachments/assets/140eb385-07fd-4541-b9b0-7155a9d206c4

**Checklist:**
- [ ] Covered by tests
- [ ] Ran it and it works as intended
- [ ] Reviewed the diff before requesting a review
- [ ] Checked for unhandled edge cases
- [ ] Linked the issues it closes
- [ ] Checked if the docs need to be updated. If so, create a separate Linear DOCS issue
- [ ] Intend to cherry-pick into the release branch
- [ ] I'm proud of this work!

PR #7998, opened 2025-09-19
Create a connector file for DuckDB connector

lovincyrus:
This PR tweaks [the max w and padding of AddDataForm](https://github.com/rilldata/rill/pull/7979#issuecomment-3312209622) and enforces the duckdb form to create a `type: connector` file. The DuckDB form will now have a required `path` and an optional `attach`.

https://linear.app/rilldata/issue/APP-403/add-path-to-duckdb-form

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-common/src/features/sources/modal/AddDataForm.svelte
+++ b/web-common/src/features/sources/modal/AddDataForm.svelte
@@ -65,8 +65,11 @@
           (property) => property.key !== "dsn",
         )) ?? [];
 
-  // FIXME: APP-209
-  const filteredParamsProperties = properties;
+  // Filter properties based on connector type
+  const filteredParamsProperties = (() => {
+    // For other connectors, filter out noPrompt properties
+    return properties.filter((property) => !property.noPrompt);
+  })();
   const schema = yup(getYupSchema[connector.name as keyof typeof getYupSchema]);
   const initialFormValues = getInitialFormValuesFromProperties(properties);
   const {
--- a/web-common/src/features/sources/modal/AddDataModal.svelte
+++ b/web-common/src/features/sources/modal/AddDataModal.svelte
@@ -122,10 +122,8 @@
   $: isModelingSupported = $isModelingSupportedForDefaultOlapDriver.data;
 
   // FIXME: excluding salesforce until we implement the table discovery APIs
-  // NOTE: DuckDB should not create connector files, only sources
   $: isConnectorType =
-    (selectedConnector?.implementsOlap &&
-      selectedConnector?.name !== "duckdb") ||
+    selectedConnector?.implementsOlap ||
     selectedConnector?.implementsSqlStore ||
     (selectedConnector?.implementsWarehouse &&
       selectedConnector?.name !== "salesforce");
@@ -144,8 +142,8 @@
   >
     <Dialog.Content
       class={cn(
-        "overflow-hidden",
-        step === 2 ? "max-w-4xl p-0 gap-0" : "p-6 gap-4",
+        "overflow-hidden max-w-4xl",
+        step === 2 ? "p-0 gap-0" : "p-6 gap-4",
       )}
       noClose={step === 1}
     >
@@ -163,7 +161,7 @@
                     on:click={() => goToConnectorForm(connector)}
                     class="connector-tile-button size-full min-w-24 min-h-16 h-20"
                   >
-                    <div class="connector-wrapper px-6 py-4 md:px-4 md:py-2">
+                    <div class="connector-wrapper px-6 py-4">
                       <svelte:component this={ICONS[connector.name]} />
                     </div>
                   </button>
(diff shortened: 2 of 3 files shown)


NamanMahor:
@lovincyrus it's either `attach` or `path` not both.

royendo:
Hold up, this completely breaks the duckdb flow.
```
type: connector

driver: duckdb 
path: "/Users/royendo/Desktop/GitHub/docs-explore/data/another_db.db"
mode: "readwrite"
```

Even when the connector is create successfully, trying to create a model
if read mode
`model execution is disabled. To enable modeling on this ClickHouse database, set 'mode: readwrite' in your connector configuration. WARNING: This will allow Rill to create and overwrite tables in your database`

if readwrite
`cannot execute model: input connector "duckdb_1" and output connector "duckdb" are not compatible`

for DuckDB we should still prompt for DB/path, SQL and source_name.

```
type: model

driver: duckdb 
db: "/Users/royendo/Desktop/GitHub/docs-explore/data/another_db.db"

sql: select * from main.my_table
```

royendo:
If we want to do the Ducklake as a live connector we'll need to test that much more before its an icon in the UI. We can market it as such too, DuckLake instead of DuckDB.

ericokuma:
the sizing still feels a bit off. This is what I'm seeing:
Screenshot 2025-09-19 at 2 07 45 PM


This is what it used to look like:
Screenshot 2025-09-19 at 2 06 20 PM

The aspect ratio of the previous buttons is what we should strive for

grahamplata (approved):
Confirmed it works on my end. I will defer ui changes :)

royendo:
Naman confirmed that db works in main so lets revert ui to model generation 👍 

duckdb as a connector is out of scope

lovincyrus:
Marked as draft as we're waiting for https://rilldata.slack.com/archives/C088Q884YF7/p1758320572058039?thread_ts=1758314464.072089&cid=C088Q884YF7

lovincyrus:
Latest flow:

```
type: connector
driver: duckdb
path: /Users/cyrusgoh/my_database.duckdb
mode: read
```

https://github.com/user-attachments/assets/182f882b-5a7d-4de3-a242-ef1c4a40946a

royendo:
@lovincyrus how big is the lift to allow duckdb to modify the `olap_connector` in the `rill.yaml`? If its a lot then we'll move that to a fast follow.

Can you also fix MD connector UI?

Some changes to `token` config.
Screenshot 2025-09-22 at 10 15 44

let me know if platform needs to fix that with its current implementation in duck.go

lovincyrus:
> how big is the lift to allow duckdb to modify the olap_connector in the rill.yaml? If its a lot then we'll move that to a fast follow.

We should do this right from a proper design input in a follow-up.

lovincyrus:
> Can you also fix MD connector UI?

On it

runtime/drivers/duckdb/duckdb.go
```diff
@@ -94,9 +94,13 @@ var motherduckSpec = drivers.Spec{
...
 		},
 		{
-			Key:    "token",
-			Type:   drivers.StringPropertyType,
-			Secret: true,
+			Key:         "token",
+			Type:        drivers.StringPropertyType,
+			Secret:      true,
+			Required:    true,
+			DisplayName: "Token",
```
lovincyrus, line 101: Noticed that we have removed these in the recent changes. Were they intentional? @grahamplata

grahamplata: c.c. @begelundmuller

lovincyrus: Without the properties, they look like this.

CleanShot 2025-09-22 at 19 15 57@2x

royendo (approved):
in terms of UI fixes, looks good!

thanks for the fixes here.

runtime/drivers/duckdb/duckdb.go
begelundmuller: Is about model properties for Motherduck? My earlier comment applies to Motherduck as well – using it as a source is deprecated, going forward we should only support it as an OLAP engine, similar to DuckDB (basically, it’s DuckDB with a different connection string). I.e. where you create a `duckdb.yaml` file with `path:` being a Motherduck path, and set `olap_connector: duckdb` in `rill.yaml`. 

But if time is tight and you want to keep the deprecated support for Motherduck as an external source that copies data into the embedded DuckDB, I think you’ll need to add the properties I had in this PR to the `motherduckSpec`: https://github.com/rilldata/rill/pull/8006/files

lovincyrus: Cool, I was on the right path. To clarify, we'll create a connector file for MotherDuck. e.g.

```
type: connector

driver: motherduck
token: "{{ .env.connector.motherduck.token }}"
path: "md:sample_data"
schema_name: kaggle
```

I was asking about re-adding some missing property spec for `token` in the ConfigProperties (a.k.a Connector Properties)

```
{
	Key:         "token",
	Secret:      true,
	Required:    true,
	DisplayName: "Token",
	Description: "MotherDuck token",
	Placeholder: "your_motherduck_token",
},
```

The recent changes were below:

```
{
	Key:    "token",
	Type:   drivers.StringPropertyType,
	Secret: true,
},
```

I'll proceed to include these missing property spec to the token key in the connector config.

PR #8000, opened 2025-09-19
feat (AI chatbot): Streaming responses

ericpgreen2:
**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-common/src/features/chat/core/conversation.ts
+++ b/web-common/src/features/chat/core/conversation.ts
@@ -1,21 +1,18 @@
-import { page } from "$app/stores";
 import { queryClient } from "@rilldata/web-common/lib/svelte-query/globalQueryClient";
 import {
   getRuntimeServiceGetConversationQueryKey,
   getRuntimeServiceGetConversationQueryOptions,
-  getRuntimeServiceListConversationsQueryKey,
-  runtimeServiceComplete,
   type RpcStatus,
-  type V1CompleteResponse,
+  type V1CompleteStreamingResponse,
   type V1GetConversationResponse,
-  type V1ListConversationsResponse,
   type V1Message,
 } from "@rilldata/web-common/runtime-client";
+import { runtime } from "@rilldata/web-common/runtime-client/runtime-store";
+import { SSEFetchClient } from "@rilldata/web-common/runtime-client/sse-fetch-client";
 import { createQuery, type CreateQueryResult } from "@tanstack/svelte-query";
 import { derived, get, writable, type Readable } from "svelte/store";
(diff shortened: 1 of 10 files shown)


web-common/src/runtime-client/sse-fetch-client.ts
```diff
@@ -0,0 +1,202 @@
+import { runtime } from "@rilldata/web-common/runtime-client/runtime-store";
+import { get } from "svelte/store";
+
+/**
+ * A clean, reusable client for handling Server-Sent Events (SSE) streams.
+ *
+ * This client properly handles the SSE protocol with "data: " prefixes and provides
+ * a simple event-based interface for consuming streaming responses.
+ */
+export class SSEFetchClient<T> {
```
AdityaHegde, line 10: This seems like a very verbose version of `streamingFetchWrapper`. Any way to use that here?

I agree it is a major refactor to incorporate it so lets take it up as a follow up.

web-common/src/features/chat/core/conversation.ts
```diff
@@ -70,227 +79,404 @@ export class Conversation {
...
+        !this.hasReceivedFirstUserMessage
+      ) {
+        this.hasReceivedFirstUserMessage = true;
+        return;
+      }
 
-      // Reset the `GetNewConversation` query cache
-      queryClient.removeQueries({ queryKey: getNewConversationQueryKey });
+      // Check if this is a tool result message
+      const toolResult = response.message.content?.[0]?.toolResult;
```
AdityaHegde, line 248: Are we a 100% sure there will only ever be a single entry inside content?

web-common/src/runtime-client/sse-fetch-client.ts
ericpgreen2: I did look into using `streamingFetchWrapper`, but it doesn't handle SSE. For example, it appears the `WatchRequestClient` doesn't use the `streamingFetchWrapper` for SSE either. 

As a follow-up to this PR, I'd actually like to use this `SSEFetchClient` in the `WatchRequestClient` as well. This would give us consistent SSE handling across the codebase and let us add features like custom headers (which EventSource doesn't support well).

web-common/src/features/chat/core/conversation.ts
ericpgreen2: Yes, [per this code](https://github.com/rilldata/rill/blob/7e51871f96aa759cecbf50ab46a151d4461192af/runtime/server/chat.go#L190-L207)

PR #8004, opened 2025-09-22
feat: Add offset rill time syntax support

AdityaHegde:
With rill time we need comparison support for things like `previous period` in a single expression.

This PR adds an offset support.
1. Adding `offset -1P` to a time range, `7d as of watermark/d offset -1P` will give 7 day period offset by 7 days.
2. Adding `offset -1M` to a time range, `7d as of watermark/d offset -1M` will give 7 day period offset by 1 month.

Also updates the way execution time is sent.

Closes APP-406

**Checklist:**
- [x] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/runtime/pkg/rilltime/rilltime.go
+++ b/runtime/pkg/rilltime/rilltime.go
@@ -25,7 +25,7 @@ var (
 		{"Now", "now"},
 		{"Latest", "latest"},
 		{"Watermark", "watermark"},
-		{"PreviousPeriod", "(?i)pp"},
+		{"PreviousPeriod", "(?i)p"},
 		{"Offset", `(?i)offset`},
 		// this needs to be after Now and Latest to match to them
 		{"PeriodToGrain", `[sSmhHdDwWqQMyY]TD`},
@@ -50,6 +50,7 @@ var (
 	rillTimeParser = participle.MustBuild[Expression](
 		participle.Lexer(rillTimeLexer),
 		participle.Elide("Whitespace"),
+		participle.UseLookahead(2), // Needed to disambiguate `offset -1P` vs `offset -1M`
 	)
 	daxNotations = map[string]string{
 		// Mapping for our old rill-<DAX> syntax
@@ -218,8 +219,13 @@ type ISOPointInTime struct {
 }
 
 type Offset struct {
-	Grain          *GrainPointInTimePart `parser:"( @@"`
-	PreviousPeriod bool                  `parser:"| @PreviousPeriod)"`
+	PreviousPeriod *PreviousPeriod       `parser:"( @@"`
+	Grain          *GrainPointInTimePart `parser:"| @@)"`
+}
+
+type PreviousPeriod struct {
+	Prefix string `parser:"@Prefix"`
+	Num    int    `parser:"@Number PreviousPeriod"`
 }
 
 type Ordinal struct {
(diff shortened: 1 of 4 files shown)


runtime/metricsview/executor_rewrite_time.go
```diff
@@ -67,7 +67,7 @@ func (e *Executor) resolveTimeRange(ctx context.Context, tr *TimeRange, tz *time
 	if tr.Expression == "" {
 		return e.resolveISOTimeRange(ctx, tr, tz, executionTime)
 	}
-	if !tr.Start.IsZero() || !tr.End.IsZero() || tr.IsoDuration != "" || tr.IsoOffset != "" || tr.RoundToGrain != TimeGrainUnspecified {
+	if !tr.Start.IsZero() || tr.IsoDuration != "" || tr.IsoOffset != "" || tr.RoundToGrain != TimeGrainUnspecified {
 		return errors.New("other fields are not supported when expression is provided")
```
begelundmuller, line 71: I don't think it's a good idea to let `tr.End` overwrite the watermark. We already pass `execution_time` separately/contextually for that use case.

runtime/pkg/rilltime/rilltime.go
```diff
@@ -214,6 +217,11 @@ type ISOPointInTime struct {
 	tg timeutil.TimeGrain
 }
 
+type Offset struct {
+	Grain          *GrainPointInTimePart `parser:"( @@"`
+	PreviousPeriod bool                  `parser:"| @PreviousPeriod)"`
+}
```
begelundmuller, line 229: Did you consider supporting expressions in `as of` instead of a new keyword? I.e. instead of:
```
7d as of watermark/d offset -1M
```
could you have:
```
7d as of watermark/d - 1M
```
?

AdityaHegde: @briangregoryholmes wanted this syntax. I am open to anything.

nishantmonu51:
IMHO, instead of introducing new syntax the ref should be offsetted based on which service it is to be used. 
Based on prev discussions with @briangregoryholmes and @medriscoll, ref is a way to include offsets. 

e.g. `7d as of ref/d` 
- ref = watermark for latest 
- ref = watermark - 1D for offset by 1 day 
- ref = 2025-01-01 to run an alert 

The key part is that the syntax remain same and consistent, ref is offsetted based on the where its to be used.

runtime/pkg/rilltime/rilltime.go
briangregoryholmes: I have no problem with supporting expressions in `as of`, but it is technically different than what `offset` does. Allowing for calendar math and snapping means that deriving a reference point using an expression and then deriving a range based on that reference point does not always yield the same output as deriving a range and then applying an offset to the start and end of that range.

It also captures the idea of comparison directly in the syntax, so `7d as of watermark/d offset -1M` captures a primary range and a comparison range. `7d as of watermark/d - 1M` does not. Having an additional keyword also allows for multiple offsets, so I could say something like `7d as of watermark/d offset -1M, -2M, -3M` to get three comparison periods.

briangregoryholmes:
My only note is that period offsets should support integers. So, it should be `offset -1P` rather than `offset PP`

PR #8006, opened 2025-09-22
Re-add deprecated DuckDB source properties

begelundmuller:
If we merge this now, we should revert it again once is able to create connector files.


lovincyrus:
Closing this as we can create a connector file for DuckDB without adding these deprecated properties.

PR #7793, opened 2025-08-19
Support additional connectors when initializing an Empty project

grahamplata:
Support additional connectors when initializing an Empty project

**Checklist:**
- [x] Covered by tests
- [x] Ran it and it works as intended
- [ ] Reviewed the diff before requesting a review
- [ ] Checked for unhandled edge cases
- [x] [Linked](https://linear.app/rilldata/issue/PLAT-178/during-a-fresh-project-start-when-choosing-to-connect-to-clickhouse-we) the issues it closes
- [ ] Checked if the docs need to be updated. If so, create a separate Linear DOCS issue
- [x] Intend to cherry-pick into the release branch
- [ ] I'm proud of this work!

PR #8009, opened 2025-09-22
Add connector YAML creation

grahamplata:
Relates to fix: https://github.com/rilldata/rill/pull/7793#discussion_r2368872926

### DIFF
--- a/runtime/parser/init.go
+++ b/runtime/parser/init.go
@@ -46,6 +46,28 @@ mock_users:
 		return err
 	}
 
+	// Determine the driver based on olap
+	var driver string
+	switch olap {
+	case "", "duckdb":
+		driver = "duckdb"
+	case "clickhouse":
+		driver = "clickhouse"
+	default:
+		// Default to duckdb for unknown olap
+		driver = "duckdb"
+	}
+
+	// Create the connector YAML
+	connectorYAML := fmt.Sprintf(`type: connector
+driver: %s
+`, driver)
+
+	err = repo.Put(ctx, fmt.Sprintf("connectors/%s.yaml", driver), strings.NewReader(connectorYAML))
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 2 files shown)


runtime/parser/init.go
```diff
@@ -46,6 +46,28 @@ mock_users:
...
+	// Determine the driver based on olap
+	var driver string
+	switch olap {
+	case "", "duckdb":
+		driver = "duckdb"
+	case "clickhouse":
+		driver = "clickhouse"
+	default:
+		// Default to duckdb for unknown olap
+		driver = "duckdb"
```
begelundmuller: This default also happens on line 26 – should probably only have one variable instead of both `driver` and `olap`

runtime/parser/init.go
```diff
@@ -46,6 +46,28 @@ mock_users:
...
+	}
+
+	// Create the connector YAML
+	connectorYAML := fmt.Sprintf(`type: connector
+driver: %s
+`, driver)
```
begelundmuller: Can you customize this for each connector? E.g. for Clickhouse, let's add `managed: true`. In the future, we may also want to add some commented out options or relevant docs links

begelundmuller (changes requested):

runtime/parser/init.go
grahamplata: Definitely, that makes sense.

runtime/parser/init.go
```diff
@@ -22,7 +22,10 @@ func InitEmpty(ctx context.Context, repo drivers.RepoStore, instanceID, displayN
 	}
 
 	// If olap is not specified we can default to duckdb
-	if olap == "" {
+	switch olap {
+	case "duckdb":
+	case "clickhouse":
+	default:
```
begelundmuller: The previous `if olap == ""` seems like a less verbose way to express the same logic?

```suggestion
	if olap == "" {
```

PR #8015, opened 2025-09-23
deploy improvements followups

k-anshul:
closes https://linear.app/rilldata/issue/PLAT-206/new-deploy-journey-follow-ups

Updates:
- Update create org flow to ask input from user similar to UI.
- Allow users to set branches for rill managed projects as well when deploying projects
- Throw errors if prod branch does not match the current branch when doing local deploys

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


admin/server/projects.go
```diff
@@ -2061,10 +2061,10 @@ func (s *Server) githubOptsForRemote(ctx context.Context, orgID, branch string,
 			return nil, nil, nil, "", status.Error(codes.InvalidArgument, "failed to get github repo")
 		}
 
-		if prodBranch != "" && prodBranch != safeStr(ghRepo.DefaultBranch) {
-			return nil, nil, nil, "", fmt.Errorf("branch should be empty or default repo branch for rill managed git repos")
+		if branch == "" {
```
AdityaHegde, line 2064: Are the named return values still needed for this method? Probably better to not have it to keep things simple.

cli/pkg/gitutil/gitutil.go
```diff
@@ -239,6 +239,19 @@ func CommitAndForcePush(ctx context.Context, projectPath string, config *Config,
...
+	headRef, err := repo.Head()
+	if err != nil {
+		return err
+	}
+	if !headRef.Name().IsBranch() {
+		return fmt.Errorf("detached HEAD state detected. Checkout a branch")
+	}
+	branch := headRef.Name().Short()
+	if headRef.Name().Short() != config.DefaultBranch {
+		return fmt.Errorf("current branch %q does not match deployed branch %q", branch, config.DefaultBranch)
```
AdityaHegde, line 252: Hmmm these errors would show up in UI as well right? We should have some UI components to either direct user on how to switch or have a button in UI for this.

admin/server/projects.go
k-anshul: Yeah sort of. It would have been best to take opts as input and directly update them. But this is being used in create and update both which have different opts.

PR #8017, opened 2025-09-23
Update InitEmpty to always create an OLAP connector file

grahamplata:
Update InitEmpty and remove default OLAP connector assignment

**Checklist:**
- [x] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/runtime/parser/init.go
+++ b/runtime/parser/init.go
@@ -58,6 +58,12 @@ driver: duckdb
 driver: clickhouse
 managed: true
 `
+	default:
+		connectorYAML = fmt.Sprintf(`type: connector
+driver: %s
+
+# TODO: Configure the connection.
+`, olap)
 	}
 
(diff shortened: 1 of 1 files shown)


ericpgreen2:
Note: as discussed, this should only be merged along with @lovincyrus's frontend changes.

grahamplata:
> Can you explain why this is needed?
> 
> From my perspective, our new goal is to always be explicit about which OLAP is configured. This PR seems to just revert to the legacy behavior of DuckDB being implicitly configured as the OLAP (since if no olap_connector: is present in rill.yaml, DuckDB is the fallback default OLAP connector).

Spoke offline.

begelundmuller (approved):

lovincyrus:
We should revert this until the frontend changes are in. cc: https://github.com/rilldata/rill/pull/8017#issuecomment-3325568723

PR #8019, opened 2025-09-24
feat: pull from existing github repo

AdityaHegde:
Closes APP-401

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


web-admin/src/features/projects/github/GithubAccessManager.ts
```diff
@@ -71,5 +74,14 @@ export class GithubAccessManager {
...
     this.githubConnectionFailed.set(false);
+
+    if (!this.reSelectingRepos) {
+      this.reSelectingRepos = false;
+      return;
+    }
+
+    await queryClient.resetQueries({
+      queryKey: getAdminServiceListGithubUserReposQueryKey(),
+    });
```
ericpgreen2: `resetQueries` makes me pause. Here's Claude's impression (which may or may not be right!):

image

AdityaHegde: Hmmm I am not sure what the issue is here. `GithubUserStatus` contains the list of github orgs and `ListGithubUserRepos` contains just the repos. I can change it to `resetQueries` with `await ...refetch` to make things clearer. 
1. We need the refetch for `ListGithubUserRepos`.
2. User access check is just an escape hatch.
3. We have to refetch here since we wont know if the user changes selections or not without actually refetching repos/orgs.
4. Github auth vs repo access are different. `hasAccess` will remain the same even if new repos are added.

AdityaHegde:
Addressing claude's concerns.
1. Good catch. Fixed.
2. We are getting repo from the list passed to Select. The null checks are just type safety.
3. We would need to make the form schema reactive based on form values. Feels like circular dependancy. Lets keep it as is.
4. This was the suggestion from a previous PR. I can revert back to the simplified version.
5. No need for try-catch. We use error from tanstack query.
6. Hmmm I dont think we are doing this anywhere else. We only ever add aria label for testing.
7. Done.

Thanks for the review. But I feel like an image is counter intuitive. Could you post it as text next time?

mindspank (approved):

web-admin/src/features/projects/github/GithubAccessManager.ts
ericpgreen2: Okay, I was implying/wondering if we just needed `refetchQueries` (merely fires another network request), not `refetchQueries` (which clears all cached state).

web-admin/src/features/projects/github/GithubAccessManager.ts
```diff
@@ -61,15 +64,22 @@ export class GithubAccessManager {
...
 
   private async refetch() {
-    await queryClient.refetchQueries({
+    await queryClient.resetQueries({
       queryKey: getAdminServiceGetGithubUserStatusQueryKey(),
     });
-    await waitUntil(() => !get(this.userStatus).isFetching);
 
-    if (!get(this.userStatus).data?.hasAccess) {
+    const refetched = await get(this.userStatus).refetch();
```
ericpgreen2: This might lead to 2 network requests?

ericpgreen2 (approved):

web-admin/src/features/projects/github/GithubAccessManager.ts
AdityaHegde: Ya updated this to `refetchQueries`

web-admin/src/features/projects/github/GithubAccessManager.ts
AdityaHegde: Ya, I removed the `resetQueries`

PR #8022, opened 2025-09-24
Support Motherduck paths when using `driver: duckdb`

begelundmuller:
- Loads the Motherduck extension and validates that `token` is set when the DuckDB path starts with `md:`
- Updates the deprecated Motherduck-to-DuckDB executor to use the `token` and `path` from `motherduck.yaml`
- Updates the integration test to test Motherduck both as an OLAP and as a source

**Checklist:**
- [x] Covered by tests
- [x] I'm proud of this work!

### DIFF
--- a/runtime/drivers/duckdb/config.go
+++ b/runtime/drivers/duckdb/config.go
@@ -1,6 +1,7 @@
 package duckdb
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 
@@ -82,6 +83,11 @@ func newConfig(cfgMap map[string]any) (*config, error) {
 		}
 	}
 
+	// Validate Motherduck token
+	if cfg.isMotherduck() && cfg.Token == "" {
+		return nil, errors.New("you must configure the 'token' connector property when using Motherduck")
+	}
+
 	// Set pool size
 	poolSize := cfg.PoolSize
 	if poolSize == 0 && cfg.CPU != 0 {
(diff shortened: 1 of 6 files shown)


runtime/drivers/duckdb/config.go
```diff
@@ -82,6 +83,11 @@ func newConfig(cfgMap map[string]any) (*config, error) {
 		}
 	}
 
+	// Validate Motherduck token
+	if cfg.isMotherduck() && cfg.Token == "" {
```
k-anshul: What about the case when `token` is set via `init_sql` ?
This is what we suggested to users earlier:
```
type: connector
driver: duckdb

path: "md:my_db"

init_sql: |
  INSTALL 'motherduck';
  LOAD 'motherduck';
  SET motherduck_token= {{ .env.motherduck_token }} 
```

runtime/reconcilers/model.go
```diff
@@ -1528,31 +1528,34 @@ func (r *ModelReconciler) acquireExecutorInner(ctx context.Context, opts *driver
...
-		}
+		// Try the other connector
 		executorName = opts.OutputConnector
-		e, err = oc.AsModelExecutor(r.C.InstanceID, opts)
-		if err != nil {
-			ir()
-			or()
-			if errors.Is(err, drivers.ErrNotImplemented) {
-				return "", nil, nil, fmt.Errorf("cannot execute model: input connector %q and output connector %q are not compatible", opts.InputConnector, opts.OutputConnector)
+		inputErr := err
```
k-anshul: Nit: Why not call it `inputErr` at line 1537 itself ?

runtime/drivers/duckdb/config.go
begelundmuller: Good point. Removing it for now. I'm also not sure if there are ways the token may have been configured in e.g. `~/.duckdb`.

runtime/reconcilers/model.go
begelundmuller: Good idea :)

PR #8023, opened 2025-09-24
installing motherduck based on path

NamanMahor:
>2. Using driver: duckdb instead of driver: motherduck to be consistent with what we do with ClickHouse

based on this we need to install motherduck based on path

**Checklist:**
- [ ] Covered by tests
- [x] I'm proud of this work!

### DIFF
--- a/runtime/drivers/duckdb/duckdb.go
+++ b/runtime/drivers/duckdb/duckdb.go
@@ -476,7 +476,13 @@ func (c *connection) reopenDB(ctx context.Context) error {
 		connInitQueries []string
 	)
 
-	if strings.HasPrefix(strings.TrimSpace(c.config.Path), "md:") {
+	normalize := func(s string) string {
+		s = strings.TrimSpace(s)
+		s = strings.Trim(s, `"'`)
+		return s
+	}
+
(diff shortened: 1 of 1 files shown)


runtime/drivers/duckdb/duckdb.go
```diff
@@ -476,7 +476,13 @@ func (c *connection) reopenDB(ctx context.Context) error {
...
-	if c.driverName == "motherduck" {
+	normalize := func(s string) string {
+		s = strings.TrimSpace(s)
+		s = strings.Trim(s, `"'`)
+		return s
+	}
+
+	if strings.HasPrefix(normalize(c.config.Path), "md:") || strings.HasPrefix(normalize(c.config.Attach), "md:") {
 		dbInitQueries = append(dbInitQueries,
 			"INSTALL 'motherduck'",
```
NamanMahor, line 487: we can remove install here but token need to be set. because install is auto my duckdb mentioned [here](https://motherduck.com/docs/getting-started/interfaces/connect-query-from-duckdb-cli/#connect-to-motherduck)

begelundmuller:
I've already implemented this here: https://github.com/rilldata/rill/pull/8022

PR #8024, opened 2025-09-24
Add managed to default duckdb.yaml on project init

grahamplata:
Add managed to default duckdb.yaml on project init

**Checklist:**
- [x] Covered by tests
- [] Reviewed the diff before requesting a review
- [x] [Linked the issues it closes](https://linear.app/rilldata/issue/PLAT-239/add-managed-true-to-default-duckdbyaml-on-project-init)
- [x] I'm proud of this work!

### DIFF
--- a/runtime/parser/init.go
+++ b/runtime/parser/init.go
@@ -51,10 +51,13 @@ mock_users:
 	switch olap {
 	case "duckdb":
 		connectorYAML = `type: connector
+
 driver: duckdb
+managed: true
 `
 	case "clickhouse":
 		connectorYAML = `type: connector
+
 driver: clickhouse
 managed: true
 `
--- a/runtime/parser/init_test.go
+++ b/runtime/parser/init_test.go
@@ -61,6 +61,7 @@ func TestInitWithDuckDB(t *testing.T) {
 	require.NoError(err)
 	require.Contains(duckdbYAML, "type: connector")
 	require.Contains(duckdbYAML, "driver: duckdb")
+	require.Contains(duckdbYAML, "managed: true")
 }
 
 func TestInitWithClickHouse(t *testing.T) {
@@ -90,6 +91,7 @@ func TestInitWithClickHouse(t *testing.T) {
 	require.NoError(err)
 	require.Contains(clickhouseYAML, "type: connector")
 	require.Contains(clickhouseYAML, "driver: clickhouse")
+	require.Contains(clickhouseYAML, "managed: true")
 
 	// Verify that duckdb.yaml is NOT created for ClickHouse projects
 	_, err = repo.Get(t.Context(), "connectors/duckdb.yaml")


runtime/parser/parse_connector.go
```diff
@@ -49,6 +49,14 @@ func (p *Parser) parseConnector(node *Node) error {
 		default:
 			return fmt.Errorf("invalid type for 'managed': expected boolean or map of args")
 		}
+	} else if tmp.Driver == "duckdb" {
+		// For backwards compatibility with DuckDB projects that don't have the 'managed' key,
+		// default to true unless path or attach are explicitly configured
+		_, hasPath := tmp.Defaults["path"]
+		_, hasAttach := tmp.Defaults["attach"]
+		if !hasPath && !hasAttach {
+			provision = true
```
begelundmuller: Does this actually work if you deploy into a cloud dev environment?

I think it will end up calling this, which I think should fail because the `admin` service doesn't have a DuckDB provisioner: https://github.com/rilldata/rill/blob/adc6744737e70b7c59f40a233ffab1fcf2517d05/runtime/connection_cache.go#L137

grahamplata: Sorry I took that literally. I will check in the connection_cache and short circuit there

runtime/drivers/duckdb/config.go
```diff
@@ -73,12 +76,12 @@ func newConfig(cfgMap map[string]any) (*config, error) {
...
 	if cfg.Mode == "" {
 		// The default mode depends on the connection type:
-		// - For generic connections (Motherduck/DuckLake with Path/Attach), default to "read"
-		// - For connections using the embedded DuckDB, default to "readwrite" to maintain compatibility
-		if cfg.Path != "" || cfg.Attach != "" {
-			cfg.Mode = modeReadOnly
-		} else {
+		// - For managed/embedded DuckDB, default to "readwrite" to maintain compatibility
+		// - For external connections (Path/Attach), default to "read"
+		if cfg.Managed || (cfg.Path == "" && cfg.Attach == "") {
```
begelundmuller: It should error if `managed: true` is combined with `path:` or `attach:`

runtime/parser/parse_connector.go
begelundmuller: I'm not sure this has been addressed

runtime/drivers/duckdb/config.go
```diff
@@ -40,6 +40,9 @@ type config struct {
 	Secrets string `mapstructure:"secrets"`
 	// Mode specifies the mode in which to open the database.
 	Mode string `mapstructure:"mode"`
+	// Managed indicates whether to use Rill-managed embedded DuckDB (true) or external DuckDB (false).
+	// When true, ignores Path and Attach settings and uses embedded database.
+	Managed bool `mapstructure:"managed"`
```
begelundmuller: I still don't think this works. The `managed:` property is parsed separately in the `parse_connector.go` and I don't think it will propagate to here. See: https://github.com/rilldata/rill/blob/c92a41f03b205dad72e765ef979d8a848a685a3f/runtime/parser/parse_connector.go#L15

Likewise, I think using `managed: true` will still end up setting `provision` to `true`, which might cause issues in Rill Cloud. See my previous comment here: https://github.com/rilldata/rill/pull/8024#discussion_r2376564396

begelundmuller (changes requested):

runtime/parser/parse_connector.go
```diff
@@ -69,6 +76,15 @@ func (p *Parser) parseConnector(node *Node) error {
...
 	r.ConnectorSpec.Provision = provision
 	r.ConnectorSpec.ProvisionArgs = provisionArgsPB
+
+	// Add the managed field to properties so drivers can access it
+	if managed != "" {
+		if r.ConnectorSpec.Properties == nil {
+			r.ConnectorSpec.Properties = make(map[string]string)
+		}
+		r.ConnectorSpec.Properties["managed"] = managed
+	}
```
begelundmuller: I don't think this is necessary, it already injects it when resolving the properties here: https://github.com/rilldata/rill/blob/aa74969f90ca4a2f42be160f257a49548c75455e/runtime/connection_cache.go#L122

runtime/drivers/duckdb/config.go
begelundmuller: > It should error if `managed: true` is combined with `path:` or `attach:`

This comment was resolved, but I don't see the corresponding code?

grahamplata: Sorry about that, addressed now

runtime/connection_cache.go
```diff
@@ -121,8 +121,12 @@ func (r *Runtime) openAndMigrate(ctx context.Context, cfg cachedConnectionConfig
 			cfg.config = maps.Clone(cfg.config)
 			cfg.config["managed"] = true
 
-			if inst.AdminConnector == "" {
-				// Provisioning has been requested, but the instance does not have an admin connector.
+			// For DuckDB with managed: true, we don't need admin provisioning.
+			// Skip admin provisioning to avoid errors in cloud environments where the admin service doesn't have a DuckDB provisioner.
+			skipAdminProvisioning := cfg.driver == "duckdb" && cfg.config["managed"] == true
```
begelundmuller: Doesn't make sense to check `cfg.config["managed"] == true` here, it's always going to be `true` per line 122

runtime/parser/parse_connector.go
```diff
@@ -69,6 +69,19 @@ func (p *Parser) parseConnector(node *Node) error {
...
+	if !tmp.Managed.IsZero() {
+		if r.ConnectorSpec.Properties == nil {
+			r.ConnectorSpec.Properties = make(map[string]string)
+		}
+		if provision {
+			r.ConnectorSpec.Properties["managed"] = "true"
+		} else {
+			r.ConnectorSpec.Properties["managed"] = "false"
+		}
+	}
```
begelundmuller: As per my previous comment, I don't think this is needed: https://github.com/rilldata/rill/pull/8024#discussion_r2432823376

PR #8030, opened 2025-09-25
feat: Automatically configure duckdb secrets for object stores

k-anshul:
https://linear.app/rilldata/issue/PLAT-217/automatically-create-duckdb-secrets-for-compatible-connectors

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/runtime/drivers/duckdb/model_executor_self.go
+++ b/runtime/drivers/duckdb/model_executor_self.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"net/url"
+	"slices"
 	"strings"
 	"time"
 
@@ -97,9 +98,26 @@ func (e *selfToSelfExecutor) Execute(ctx context.Context, opts *drivers.ModelExe
 	}
 
 	// Add PreExec statements that create temporary secrets for object store connectors.
-	for _, connector := range e.c.config.secretConnectors() {
+	secretConnectors := e.c.config.secretConnectors()
+	if len(secretConnectors) == 0 {
+		// if nothing is configured then configure every applicable connector
+		for _, connector := range opts.Env.Connectors {
+			switch connector.Type {
+			case "s3", "azure", "gcs":
+				secretConnectors = append(secretConnectors, connector.Name)
+			}
+		}
+	}
+	for _, connector := range secretConnectors {
 		secretSQL, err := objectStoreSecretSQL(ctx, opts, connector, "", nil)
 		if err != nil {
+			if !slices.Contains(e.c.config.secretConnectors(), connector) {
+				// if user has not explicitly configured this secret container then do not fail
+				continue
+			}
+			if errors.Is(err, errGCSUsesNativeCreds) {
+				continue
+			}
 			return nil, fmt.Errorf("failed to create secret for connector %q: %w", connector, err)
 		}
 		if inputProps.PreExec == "" {
--- a/runtime/drivers/registry.go
+++ b/runtime/drivers/registry.go
@@ -184,3 +184,38 @@ func (i *Instance) Config() (InstanceConfig, error) {
 
 	return res, nil
 }
+
+func (i *Instance) ResolveConnectors() []*runtimev1.Connector {
+	var res []*runtimev1.Connector
+	res = append(res, i.Connectors...)
+	res = append(res, i.ProjectConnectors...)
+	// implicit connectors
+	vars := i.ResolveVariables(true)
+	for k := range vars {
+		if !strings.HasPrefix(k, "connector.") {
+			continue
+		}
+
+		parts := strings.Split(k, ".")
+		if len(parts) <= 2 {
+			continue
+		}
+
+		// Implicitly defined connectors always have the same name as the driver
+		name := parts[1]
+		found := false
+		for _, c := range res {
+			if c.Name == name {
+				found = true
+				break
+			}
+		}
+		if !found {
+			res = append(res, &runtimev1.Connector{
+				Type: name,
+				Name: name,
+			})
+		}
+	}
+	return res
+}
--- a/runtime/resolvers/testdata/s3_connector.yaml
+++ b/runtime/resolvers/testdata/s3_connector.yaml
@@ -36,7 +36,8 @@ project_files:
     type: model
     connector: duckdb
     materialize: true
-    pre_exec: "CREATE SECRET secret1 (TYPE s3, KEY_ID '{{.env.connector.s3.aws_access_key_id}}', SECRET '{{.env.connector.s3.aws_secret_access_key}}', REGION 'us-east-1')"
+    # commenting the `pre_exec` below to test that automatic s3 secret creation in duckdb works 
+    # pre_exec: "CREATE SECRET secret1 (TYPE s3, KEY_ID '{{.env.connector.s3.aws_access_key_id}}', SECRET '{{.env.connector.s3.aws_secret_access_key}}', REGION 'us-east-1')"
     partitions:
       glob:
         path: s3://integration-test.rilldata.com/glob_test/y=202*/a*.csv
--- a/go.mod
+++ b/go.mod
@@ -43,7 +43,6 @@ require (
 	github.com/fatih/color v1.16.0
 	github.com/fsnotify/fsnotify v1.7.0
 	github.com/getkin/kin-openapi v0.132.0
-	github.com/go-git/go-billy/v5 v5.6.2
 	github.com/go-git/go-git/v5 v5.13.2
 	github.com/go-jose/go-jose/v3 v3.0.4
 	github.com/go-logr/zapr v1.2.4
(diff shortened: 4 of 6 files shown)


runtime/drivers/duckdb/model_executor_self.go
```diff
@@ -97,9 +98,26 @@ func (e *selfToSelfExecutor) Execute(ctx context.Context, opts *drivers.ModelExe
 	}
 
 	// Add PreExec statements that create temporary secrets for object store connectors.
-	for _, connector := range e.c.config.secretConnectors() {
+	secretConnectors := e.c.config.secretConnectors()
+	if len(secretConnectors) == 0 {
+		// if nothing is configured then configure every applicable connector
+		for _, connector := range opts.Env.Connectors {
+			switch connector.Type {
+			case "s3", "azure", "gcs":
```
k-anshul: Few considerations here:
1. If a user define two types of s3 connectors then we will create secrets for both. This means it is not guaranteed if this will work. DuckDB picks just one and it may work. In this case a user will be expected to create a secret with a `scope` defined to path.
  - I think in this case it is okay to create secrets.
3. We can create secrets for `postgres`/`mysql` as well but a user will be still required to define a `pre_exec` that runs attach. Though it will be much shorter. `ATTACH '' AS postgres_db`.
  - I think it is fine to skip creating secrets in this case since users will anyways need a `pre_exec`.

docs/docs/reference/project-files/models.md
```diff
@@ -438,6 +438,10 @@ _[string]_ - refers to SQL queries to run before the main query, available for D
 
 _[string]_ - refers to a SQL query that is run after the main query, available for DuckDB-based models. _(optional)_. Ensure `post_exec` queries are idempotent. Use `IF EXISTS` statements when applicable. 
 
+### `secrets`
+
+_[string]_ - Comma-separated list of connector names to create temporary secrets for. If this is set then secrets configured on connector is ignored.
```
begelundmuller: Can you describe this in `runtime/parser/schema` and regenerate the docs?

runtime/drivers/duckdb/model_executor_self.go
```diff
@@ -21,7 +23,10 @@ type selfToSelfExecutor struct {
 	c *connection
 }
 
-var _ drivers.ModelExecutor = &selfToSelfExecutor{}
+var (
+	_                 drivers.ModelExecutor = &selfToSelfExecutor{}
+	createSecretRegex                       = regexp.MustCompile(`(?i)\bcreate\b(?:\s+\w+)*?\s+secret\b`)
+)
```
begelundmuller: nit: since these two declarations, combining them just causes confusion:
```suggestion
var _ drivers.ModelExecutor = &selfToSelfExecutor{}

var createSecretRegex = regexp.MustCompile(`(?i)\bcreate\b(?:\s+\w+)*?\s+secret\b`)
```

Also consider moving `createSecretRegex` further down in the file since it's an internal util that's only related to one function

runtime/drivers/duckdb/model_executor_self.go
```diff
@@ -395,3 +399,23 @@ func rewriteLocalPaths(ast *duckdbsql.AST, basePath string, allowHostAccess bool
 
 	return ast.Format()
 }
+
+func secretConnectors(modelSecrets string, duckdbSecrets []string, allConnectors []*runtimev1.Connector) ([]string, bool) {
```
begelundmuller: nit: can you add a docstring here explaining the intended behavior? I.e. that model secrets overwrite connector-level secrets, and that connector-level secrets default to all compatible connectors in the project

runtime/drivers/duckdb/model_manager.go
```diff
@@ -16,6 +16,9 @@ type ModelInputProperties struct {
 	InitQueries string `mapstructure:"init_queries"`
 	PreExec     string `mapstructure:"pre_exec"`
 	PostExec    string `mapstructure:"post_exec"`
+	// Secrets is a comma-separated list of connector names to create temporary secrets for before executing models.
+	// If this is set then connector specific secrets are ignored.
+	Secrets string `mapstructure:"secrets"`
```
begelundmuller: Both here and for the connector-level config, what do you think about adding support for `*` to mean all compatible connectors? And then make `*` the default at the connector level.

Just to have an explicit syntax for saying "all" – so we have the option to include it in YAML boilerplate if we want:
```yaml
type: connector
driver: duckdb
managed: true # Optional, it's the default
secrets: '*'  # Optional, it's the default
```

runtime/drivers/registry.go
```diff
@@ -184,3 +184,38 @@ func (i *Instance) Config() (InstanceConfig, error) {
...
+				found = true
+				break
+			}
+		}
+		if !found {
+			res = append(res, &runtimev1.Connector{
+				Type: name,
+				Name: name,
+			})
+		}
```
begelundmuller, line 218: Should it also add the properties for the implicit connectors?

Btw I'm also okay with not supporting implicit connectors here if that's actually easier

runtime/resolvers/testdata/s3_connector.yaml
```diff
@@ -36,7 +36,8 @@ project_files:
     type: model
     connector: duckdb
     materialize: true
-    pre_exec: "CREATE SECRET secret1 (TYPE s3, KEY_ID '{{.env.connector.s3.aws_access_key_id}}', SECRET '{{.env.connector.s3.aws_secret_access_key}}', REGION 'us-east-1')"
+    # commenting the `pre_exec` below to test that automatic s3 secret creation in duckdb works 
+    # pre_exec: "CREATE SECRET secret1 (TYPE s3, KEY_ID '{{.env.connector.s3.aws_access_key_id}}', SECRET '{{.env.connector.s3.aws_secret_access_key}}', REGION 'us-east-1')"
```
begelundmuller: Should/can this be removed now?

runtime/drivers/duckdb/model_executor_self.go
k-anshul: The linter complained and asked to combine variables. I can move this further down the file.

runtime/drivers/duckdb/model_manager.go
k-anshul: I would say we should keep the YAML boilerplate as minimal as possible. The same thing can be achieved with a commented code block as well.

But I can add `*` as well. Currently I did not return error if generating a secret SQL returned an error and nothing was explicitly configured. I am extending the same for `*` as well.

runtime/resolvers/testdata/s3_connector.yaml
k-anshul: I added this comment to make it explicit that creating secrets is not required anymore if someone takes a look at tests.

runtime/drivers/registry.go
k-anshul: Yeah it can be somewhat tricky because we also allow setting `aws_secret_key` directly without the prefix. Let's skip for now :)

runtime/drivers/duckdb/model_executor_self.go
begelundmuller: Okay, stupid linter...

runtime/drivers/duckdb/model_manager.go
begelundmuller: Agree on keeping the YAML boilerplate limited. I just mean to have a structured way to indicate "all", which would be the internal default. I.e. if the internal default is `*`, then if you explicitly put `secrets: ""`, then it will overwrite the internal default and disable secrets as expected.

runtime/resolvers/testdata/s3_connector.yaml
begelundmuller: Hmm, if you want to make that explicit, maybe duplicate the test case and have one with the `pre_exec` and one without it? And use the test case names to explain the difference – e.g. `all_datatypes_duckdb_model_partitions_explicit_secret` and `all_datatypes_duckdb_model_partitions_implicit_secret`.

runtime/drivers/duckdb/model_executor_self.go
```diff
@@ -400,17 +397,26 @@ func rewriteLocalPaths(ast *duckdbsql.AST, basePath string, allowHostAccess bool
...
+// A user can also specify the list in the duckdb driver configuration which will be used for all models using duckdb driver. This is only used if the model specific configuration is not set.
+// If neither is specified then we auto detect connectors of type s3, azure, gcs from the list of all connectors in the project.
+// The boolean return value is true if the returned list was auto-detected.
+func secretConnectors(modelSecrets, duckdbSecrets string, allConnectors []*runtimev1.Connector) ([]string, bool) {
+	var configuredSecret string
+	if modelSecrets != "" && modelSecrets != "*" { // * means all applicable connectors
+		configuredSecret = modelSecrets
+	} else if duckdbSecrets != "" && duckdbSecrets != "*" {
+		configuredSecret = duckdbSecrets
+	}
```
begelundmuller: If you put `secrets: '*'` in the model, but have `secrets: s3` in the `duckdb.yaml`, then won't this resolve to only `s3` instead of all secrets? Seem like that's wrong since the model config should take precedence here.

begelundmuller (approved):

runtime/drivers/duckdb/model_executor_self.go
k-anshul: Good catch :)

PR #8043, opened 2025-09-29
fix: leaderboard bar lengths with multiple measures

AdityaHegde:
Updates the leaderboard and dimension table measure bars based on whether the measure is a valid percent of total or not.
1. For valid percent of total, use the total value as max width.
2. For non valid percent of total, use the max visible value as max width.

Also, the leaderboard bars use the full table width when only one measure is present. Uses the measure column width when there are multiple.

Closes APP-17

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


mindspank:
@djbarnwal would make sense but it's a pretty big change from how it is today, we can start here and ease in that change over time.

I'm also slightly concerned that proportions will be off if we also span delta columns exaggerating small changes

PR #8046, opened 2025-09-29
fix: creating empty project leading to 500 error

AdityaHegde:
This error is hit somewhat frequently in our web-local E2E. It is always when using an empty project.
In the logs we see `Context cancelled` for the refetch of list of files. This PR tries to avoid this by throttling the refetch a bit.

Closes APP-345

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


web-common/src/features/entity-management/WatchFilesClient.ts
```diff
@@ -12,9 +13,15 @@ import { runtime } from "@rilldata/web-common/runtime-client/runtime-store";
 import { WatchRequestClient } from "@rilldata/web-common/runtime-client/watch-request-client";
 import { get } from "svelte/store";
 
+const REFETCH_LIST_FILES_THROTTLE_MS = 200;
```
ericpgreen2: Can we make this a little smaller? 50? 100?

web-common/src/features/entity-management/WatchFilesClient.ts
```diff
@@ -72,10 +79,11 @@ export class WatchFilesClient {
       }
     }
     if (isNew || res.event === V1FileEvent.FILE_EVENT_DELETE) {
-      // TODO: should this be throttled?
-      void queryClient.refetchQueries({
-        queryKey: getRuntimeServiceListFilesQueryKey(instanceId),
-      });
+      this.refetchListFilesThrottle.throttle(() =>
```
ericpgreen2, line 83: Mind adding a comment to explain why this is throttled?

ericpgreen2: I'm also thinking that another option would be to manipulate the TanStack Query cache directly, rather than fire an API call.

AdityaHegde: Sure will add a comment. I dont know if that is a good idea. The query is not too heavy, so lets keep it as is. We can look into changing it when we have cloud editing and see issues with it.

web-common/src/features/entity-management/WatchFilesClient.ts
AdityaHegde: Lets do 100. Hopefully it is not too low.

PR #8055, opened 2025-09-30
feat: add resize API to embedded dashboards

AdityaHegde:
Adds a param to the embed url `dynamic_height=true` to signify that the content height will be based on content rather than the iframe size.
Also adds `resized` event to the embed API that emits the content size.

Code wise, this updates layout to not have `height:100%` when `dynamicHeight` is true. Also fixes an issue with vega lite tooltip to not break the layout by setting top & left to 0.

Use this to test embed scenario: https://github.com/AdityaHegde/rill-embed-sample.
Update `.env` and [the route](https://github.com/AdityaHegde/rill-embed-sample/blob/main/public/src/routes/dynamic-height/%2Bpage.svelte#L10-L11) to your deployed org, project and dashboards.

Closes APP-378

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-admin/src/features/embeds/init-embed-public-api.ts
+++ b/web-admin/src/features/embeds/init-embed-public-api.ts
@@ -51,7 +51,6 @@ export default function initEmbedPublicAPI(): () => void {
     RESIZE_THROTTLE_TIMEOUT,
   );
   function onResize(event: PageContentResized) {
-    console.log("onResize", event);
     // Throttle the resize event.
     // This avoids too many events being fired when size changes quickly, especially when page is loading.
     resizeThrottler.throttle(() => {


web-admin/src/features/embeds/init-embed-public-api.ts
```diff
@@ -43,7 +46,31 @@ export default function initEmbedPublicAPI(): () => void {
     });
   });
 
-  return unsubscribe;
+  const resizeThrottler = new Throttler(
+    RESIZE_THROTTLE_TIMEOUT,
+    RESIZE_THROTTLE_TIMEOUT,
+  );
+  function onResize(event: PageContentResized) {
+    console.log("onResize", event);
```
djbarnwal: Leftover log

PR #8057, opened 2025-10-01
adding support map<string,any> in connector

NamanMahor:
INSERT DESCRIPTION HERE

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!


cli/pkg/local/app.go
```diff
@@ -119,12 +120,17 @@ func NewApp(ctx context.Context, opts *AppOptions) (*App, error) {
 		logger.Info("Dropping old stage.db file and rebuilding project")
 	}
 
+	metastoreProp, err := structpb.NewStruct(map[string]any{"dsn": "file:rill?mode=memory&cache=shared"})
+	if err != nil {
+		return nil, err
+	}
+
 	// Create a local runtime with an in-memory metastore
```
begelundmuller: nit: the code is related to the comment here, so should come after it, not before

```suggestion
	// Create a local runtime with an in-memory metastore
	metastoreProp, err := structpb.NewStruct(map[string]any{"dsn": "file:rill?mode=memory&cache=shared"})
	if err != nil {
		return nil, err
	}
```

cli/pkg/local/app.go
```diff
@@ -194,41 +200,58 @@ func NewApp(ctx context.Context, opts *AppOptions) (*App, error) {
...
 
+	olapProp, err := structpb.NewStruct(map[string]any{
+		"pool_size":   "4", // Default pool size for DuckDB
+		"log_queries": strconv.FormatBool(opts.Debug),
+	})
+	if err != nil {
+		return nil, err
+	}
+
 	// Add default OLAP connector
```
begelundmuller: Same problem as above. Seems this is a problem throughout the file – can you go over and update?

As a general note about comments in Rill, in longer functions, we often use "section header" style comments that explain what each section/block of related code does. E.g. if you have 10 lines to do one thing (like add a default OLAP connector), the goal is not to explain every line or the last line, the goal is to have a comment at the _start_ that briefly explains what the block is about.

Think about it from the perspective of the reviewer or the next person who's going to make a change – you want to know what the goal of the next few lines is before reading the actual lines in detail. So having comments at the start of each section makes it easier to understand what's happening without reading each line in depth. This also makes it easier to skim/skip through the code to find what you're looking for. 

For example, this is not so useful:
```go
olapProp, err := ...
if err != nil { ... }
olapConnector := &runtimev1.Connector{...}
// Add OLAP connector
connectors = append(connectors, olapConnector)
```

Because at the point where you get to the comment, you already read those lines and it's kind of obvious what's happening. On the other hand, this is useful because you can read the comment quickly to determine whether you need to read the code in detail or not:

```go
// Add OLAP connector
olapProp, err := ...
if err != nil { ... }
olapConnector := &runtimev1.Connector{...}
connectors = append(connectors, olapConnector)
```

runtime/connections.go
```diff
@@ -340,37 +329,14 @@ func (c *ConnectorConfig) Resolve() map[string]any {
...
-	n := len(c.Preset) + len(c.Project) + len(c.Env)
-	if n == 0 {
-		return nil
-	}
-
-	cfg := make(map[string]string, n)
-	for k, v := range c.Project {
-		cfg[strings.ToLower(k)] = v
+	for k, v := range lowerKeys(c.Project) {
+		cfg[k] = v
```
begelundmuller: This modifies `c.Project` in-place, doesn't it? Seems weird to have side effects in this place. 

Maybe just keep the old code?
```suggestion
	for k, v := range c.Project {
		cfg[strings.ToLower(k)] = v
```

runtime/connections.go
```diff
@@ -340,37 +329,14 @@ func (c *ConnectorConfig) Resolve() map[string]any {
...
 	}
 	for k, v := range c.Env {
 		cfg[strings.ToLower(k)] = v
 	}
-	for k, v := range c.Preset {
-		cfg[strings.ToLower(k)] = v
-	}
+	cfg = deepMerge(cfg, lowerKeys(c.Preset))
```
begelundmuller: What is the use case for a deep merge here? I guess I would expect the variables to overwrite, not merge

runtime/connections.go
```diff
@@ -381,7 +347,51 @@ func (c *ConnectorConfig) setPreset(k, v string, force bool) {
...
+		case []any:
+			res[lk] = lowerKeysInSlice(val)
+		default:
+			res[lk] = val
+		}
+	}
+	return res
+}
+
+func lowerKeysInSlice(s []any) []any {
```
begelundmuller: What is the use case for lowercasing nested keys? I think maybe just lowercasing the variable name is all that's expected.

proto/rill/runtime/v1/api.proto
```diff
@@ -348,9 +348,7 @@ message Connector {
   // Name of the connector
   string name = 2;
   // Config for the connector
-  map<string, string> config = 3;
```
begelundmuller, line 353: I think the `runtimev1.Connector` type is persisted in the database, so I'm not sure this change is safe. Can you check that? https://github.com/rilldata/rill/blob/c92a41f03b205dad72e765ef979d8a848a685a3f/runtime/drivers/sqlite/registry.go#L97-L100

Maybe keep the name here `config`. Also, are there any other places you think compatibility could be a problem?

runtime/connections.go
```diff
@@ -290,31 +287,23 @@ func (r *Runtime) ConnectorConfig(ctx context.Context, instanceID, name string)
...
-	res := maps.Clone(c.Config)
+func resolveConnectorProperties(environment string, vars map[string]string, c *runtimev1.Connector) (map[string]any, error) {
+	res := c.Properties.AsMap()
 	if res == nil {
-		res = make(map[string]string)
+		res = make(map[string]any)
 	}
-
-	// Resolve templating in properties that use it
-	for _, k := range c.TemplatedProperties {
```
begelundmuller, line 302: The change here drops this optimization. Is there an easy way we can keep it, at least for root-level props? This code is very hot path (almost every API call), so nice not to run the template engine redundantly

cli/pkg/local/app.go
NamanMahor: done

runtime/connections.go
NamanMahor: lowerKeys return the copy with lower keys.

cli/cmd/runtime/start.go
```diff
@@ -221,6 +222,12 @@ func StartCmd(ch *cmdutil.Helper) *cobra.Command {
 
 			// Create ctx that cancels on termination signals
 			ctx := graceful.WithCancelOnTerminate(context.Background())
+			metastoreConfig, err := structpb.NewStruct(map[string]any{
+				"dsn": conf.MetastoreURL,
+			})
```
begelundmuller, line 228: Comment doesn't match what happens here

runtime/connections.go
```diff
@@ -381,7 +375,20 @@ func (c *ConnectorConfig) setPreset(k, v string, force bool) {
...
 	}
 	c.Preset[k] = v
 }
+
+func lowerKeys(m map[string]any) map[string]any {
+	res := make(map[string]any, len(m))
+	for k, v := range m {
+		lk := strings.ToLower(k)
+		if nested, ok := v.(map[string]any); ok {
+			res[lk] = lowerKeys(nested)
```
begelundmuller: Why does it need to lowercase nested objects? I don't understand the use case for that. Same question as here: https://github.com/rilldata/rill/pull/8057#discussion_r2399171246

runtime/connections.go
```diff
@@ -290,29 +287,47 @@ func (r *Runtime) ConnectorConfig(ctx context.Context, instanceID, name string)
...
+		// if value is root level string properties is must have already be resolved using TemplatedProperties
+		if _, ok := v.(string); !ok {
+			continue
+		}
+		v, err := parser.ResolveTemplateRecursively(v, td, true)
+		if err != nil {
+			return nil, fmt.Errorf("failed to resolve template: %w", err)
+		}
+		res[k] = v
+	}
```
begelundmuller: These `v.(string)` feel hacky and tightly coupled to the behavior of how `TemplatedProperties` are parsed. I would expect that templating is applied for any property in `TemplatedProperties`, and not for any other.

So instead of doing `v.(string)` checks here, consider if you can refactor the parser such that it also adds fields that have nested templating to `TemplatedProperties`.

runtime/drivers/https/https.go
```diff
@@ -49,17 +49,21 @@ var spec = drivers.Spec{
 type driver struct{}
 
 type ConfigProperties struct {
+	Headers map[string]string `mapstructure:"headers"`
```
begelundmuller, line 52: Does this impact the docs and/or JSON schema in `runtime/parser/schema/project.schema.yaml`? Looking in `project.schema.yaml`, it currently says `path:` is supported in the `type: connector` file, but I don't think that's correct.

runtime/parser/parse_connector.go
```diff
@@ -51,11 +51,17 @@ func (p *Parser) parseConnector(node *Node) error {
 		}
 	}
 
-	// Find out if any properties are templated
```
begelundmuller, line 54: Why remove this comment? If the behavior changed, add a different section comment

runtime/parser/parse_connector.go
```diff
@@ -65,19 +71,23 @@ func (p *Parser) parseConnector(node *Node) error {
...
-	r.ConnectorSpec.Properties = tmp.Defaults
+	r.ConnectorSpec.Properties = propertiesPB
 	r.ConnectorSpec.TemplatedProperties = templatedProps
 	r.ConnectorSpec.Provision = provision
 	r.ConnectorSpec.ProvisionArgs = provisionArgsPB
 
 	return nil
 }
 
-// analyzeTemplatedProperties returns a slice of map keys that have a value which contains templating tags.
```
begelundmuller, line 76: Why remove this comment? Ideally every non-trivial function, especially in large packages, should have a docstring

runtime/connections.go
NamanMahor: also not sure why we need it for root properties follow the same convention to lower the key for sub properties. let me know if we want this only at root properties. will remove this.

begelundmuller: We need to do it for root properties for backwards compatibility because once some users used uppercase properties like `AWS_ACCESS_KEY_ID: '{{ .env.AWS_ACCESS_KEY_ID }}'. But I definitely don't think this should be done for nested properties.

NamanMahor: removed lowerKeys.

begelundmuller (approved):

PR #8063, opened 2025-10-01
Update copy of generate metrics for external table

lovincyrus:
CleanShot 2025-10-01 at 09 54 08@2x

Closes https://linear.app/rilldata/issue/APP-444/update-copy-of-the-external-table-option

**Checklist:**
- [ ] Covered by tests
- [ ] I'm proud of this work!

### DIFF
--- a/web-common/src/features/connectors/explorer/TableMenuItems.svelte
+++ b/web-common/src/features/connectors/explorer/TableMenuItems.svelte
@@ -159,7 +159,7 @@
   <NavigationMenuItem on:click={handleGenerateMetricsAndExplore}>
     <ExploreIcon slot="icon" />
     <div class="flex gap-x-2 items-center">
-      Generate an Explore dashboard
+      Generate metrics
       {#if $ai}
         with AI
         <WandIcon class="w-3 h-3" />


web-common/src/features/connectors/explorer/TableMenuItems.svelte
```diff
@@ -159,7 +159,7 @@
   <NavigationMenuItem on:click={handleGenerateMetricsAndExplore}>
     <ExploreIcon slot="icon" />
     <div class="flex gap-x-2 items-center">
-      Generate an Explore dashboard
+      Generate metrics
```
ericpgreen2: Currently, to be consistent with the menu options for the OLAP connectors:
1. This option should read "Generate dashboard" (because it generates an Explore file)
2. We should add another entry that only generates a Model and Metrics View file. No Explore file. It should be named "Generate metrics".

In the future, when our Metrics View files also generate Explore resources, then we can drop the "Generate dashboard" option.

ericpgreen2 (changes requested):

web-common/src/features/connectors/explorer/TableMenuItems.svelte
lovincyrus: resolved

